PR #721, opened 2017-01-26
Constants for supported language codes for Speech API.

jskeet:
We'll need to regenerate this periodically - the generator is extremely simple,
but seems to do its job.


apis/Google.Cloud.Speech.V1Beta1/Google.Cloud.Speech.V1Beta1/LanguageCodes.cs
```diff
@@ -0,0 +1,518 @@
+﻿// Copyright 2016, Google Inc. All rights reserved.
```
chrisdunelm: nit: 2017

PR #724, opened 2017-01-27
Use the right deadline and cancellation token when polling

jskeet:
Needs tests, along with a new GAX release, but should give context to the GAX changes

### DIFF
--- a/apis/Google.LongRunning/Google.LongRunning/OperationsClientPartial.cs
+++ b/apis/Google.LongRunning/Google.LongRunning/OperationsClientPartial.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using Google.Api.Gax;
+using Google.Api.Gax.Grpc;
 using System;
 
 namespace Google.LongRunning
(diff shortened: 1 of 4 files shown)


apis/Google.LongRunning/Google.LongRunning/OperationsClientPartial.cs
```diff
@@ -34,6 +35,19 @@ public virtual IScheduler Scheduler
...
         }
+
+        // It would be nice to make this internal, but then anyone creating their
+        // own subclass for testing would be broken.
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="callSettings"></param>
+        /// <returns></returns>
```
evildour: Should this have docs or are they intentionally empty?

apis/Google.LongRunning/Google.LongRunning/OperationsClientPartial.cs
```diff
@@ -34,6 +35,19 @@ public virtual IScheduler Scheduler
         {
             get { throw new NotImplementedException(); }
         }
+
+        // It would be nice to make this internal, but then anyone creating their
+        // own subclass for testing would be broken.
```
evildour: I'm not sure I understand exactly why that would break libraries with their own subclasses, but would protected internal work as well?

jskeet: Hmm... protected internal should work, so long as mocking libraries are able to cope with it. Basically, we want users to be able to test with their own code, so they've got to be able to override this to avoid polling operations going bang.

apis/Google.LongRunning/Google.LongRunning/OperationsClientPartial.cs
jskeet: Docs will be there when I write tests :) Basically we need to get a GAX release done before I can finish this PR.

apis/Google.LongRunning/Google.LongRunning/OperationsClientPartial.cs
evildour: Makes sense. Thanks

apis/Google.LongRunning/Google.LongRunning/OperationsClientPartial.cs
evildour: Ok, SGTM.

jskeet:
This has now been updated - PTAL.

PR #727, opened 2017-01-31
Make MetadataResult constructor public

chrisdunelm:
To allow faking/mocking of the API calls that return a MetadataResult.

### DIFF
--- a/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1/MetadataResult.cs
+++ b/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1/MetadataResult.cs
@@ -33,10 +33,15 @@ namespace Google.Cloud.Metadata.V1
         /// </summary>
         public string Content { get; }
 
-        internal MetadataResult(string content, string etag)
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1/MetadataResult.cs
```diff
@@ -33,7 +33,12 @@ public sealed class MetadataResult
         /// </summary>
         public string Content { get; }
 
-        internal MetadataResult(string content, string etag)
+        /// <summary>
+        /// Construct a result.
+        /// </summary>
+        /// <param name="content">The content of the server response.</param>
+        /// <param name="etag">The ETag header from the server response.</param>
+        public MetadataResult(string content, string etag)
```
evildour: I think my original case was wrong. It should have been `eTag` I think.

chrisdunelm: Done

PR #729, opened 2017-01-31
Use ulong instance ID in test

chrisdunelm:
Exposes problem with parsing ulong within json.net

### DIFF
--- a/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1.IntegrationTests/Emulator/initial_data.json
+++ b/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1.IntegrationTests/Emulator/initial_data.json
@@ -14,7 +14,7 @@
       }
     ],
     "hostname": "name.project.google.com.internal",
-    "id": 67890,
+    "id": 13512546227574112017,
     "image": "",
     "licenses": [
       {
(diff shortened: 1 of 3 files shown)


chrisdunelm:
The instanceID I've used in the integration test is from a real GCE instance.

apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1.IntegrationTests/Emulator/initial_data.json
```diff
@@ -14,7 +14,7 @@
       }
     ],
     "hostname": "name.project.google.com.internal",
-    "id": 67890,
+    "id": 13512546227574112017,
```
evildour: These locations should also be updated: [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1.IntegrationTests/Emulator/metadata_server_test.py#L265) and [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Metadata.V1/Google.Cloud.Metadata.V1.IntegrationTests/Emulator/metadata_server_test.py#L561).

chrisdunelm: Done

PR #730, opened 2017-01-31
Make the PollSettings client-specific, configurable by settings

jskeet:
This PR includes generated code - we'd need to change the codegen
to include these partial methods.

(We can decide whether to submit as-is and update codegen before we next generate LRO, or hold of on this change. The codegen would need to change to generate the right OperationsSettings too.)

### DIFF
--- a/apis/Google.LongRunning/Google.LongRunning.Tests/OperationTest.cs
+++ b/apis/Google.LongRunning/Google.LongRunning.Tests/OperationTest.cs
@@ -237,6 +237,7 @@ namespace Google.LongRunning.Tests
             private readonly CallSettings _baseCallSettings;
             public override IClock Clock => FakeScheduler.Clock;
             public override IScheduler Scheduler => FakeScheduler;
+            public override PollSettings DefaultPollSettings => null;
             public FakeClock FakeClock => FakeScheduler.Clock;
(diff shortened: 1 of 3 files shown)


apis/Google.LongRunning/Google.LongRunning/OperationsClient.cs
```diff
@@ -54,6 +54,8 @@ private OperationsSettings(OperationsSettings existing) : base(existing)
             DeleteOperationSettings = existing.DeleteOperationSettings;
         }
 
+        partial void OnCopy(OperationsSettings existing);
```
chrisdunelm: This needs calling.

jskeet: Whoops - done.

jskeet:
Now that #790 is in, we can expand OperationSettings appropriately. That leaves the question of what will generate appropriate defaults...

jskeet:
This has now been updated - looks good to me (obviously) but let's wait for platform-wide discussion before merging.

jskeet:
Revisiting this, I think it's still the right way to go, even if we don't currently generate any code that sets it. @chrisdunelm could you re-review and we can merge?

PR #745, opened 2017-02-06
Remove the ISizer interface in favor of a function Func<T, int>.

### LINKED ISSUE #738: Remove ISizer interface
The ISizer interface can easily be replaced with Func<T, int> and will make alternative implementations simpler.

cc @jskeet

iantalarico:
Fixes #738


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ErrorReporting/ReportedErrorEventSizer.cs
```diff
@@ -16,17 +16,9 @@
 
 namespace Google.Cloud.Diagnostics.Common
 {
-    /// <summary>
-    /// An <see cref="ISizer{T}"/> for <see cref="ReportedErrorEvent"/>s.
-    /// </summary>
-    internal sealed class ReportedErrorEventSizer : ISizer<ReportedErrorEvent>
+    internal static class ReportedErrorEventSizer
```
jskeet: We don't need this and the TraceSizer. We could just have `MessageSizer<T> where T : IMessage`, then you could refer to `MessageSizer<ReportedErrorEvent>.GetSize` etc. It would be nice not to have this "bounce" at all, but I don't think there's a very simple way of creating an open delegate :(

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Utils.cs
```diff
@@ -43,16 +43,10 @@ public static string GetTestId()
             return Guid.NewGuid().ToString();
         }
 
-        /// <summary>A sizer for ints.</summary>
-        public class IntSizer : ISizer<int>
+        public static class IntSizer
```
jskeet: Possibly `ConstantSizer`? (That could be generic too, if you ever want to get a constant size for other types.)

iantalarico: Done.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ErrorReporting/ReportedErrorEventSizer.cs
iantalarico: This is a much better solution, done!

PR #746, opened 2017-02-07
Support customer-supplied encryption keys for storage

jskeet:
Draft of the changes to support customer-suppied encryption keys.
I haven't performed *any* testing yet, and this relies a change in the REST support libraries

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UrlSignerTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UrlSignerTest.cs
@@ -179,49 +179,20 @@ namespace Google.Cloud.Storage.V1.IntegrationTests
             var data = _fixture.SmallContent;
             string url = null;
 
-            string key;
-            string hash;
-            using (var rand = RandomNumberGenerator.Create())
-            using (var sha256 = SHA256.Create())
+            EncryptionKey key = EncryptionKey.Generate();
+
+            Func<HttpRequestMessage> createGetRequest = () =>
             {
-                var keyBytes = new byte[32];
-                rand.GetBytes(keyBytes);
-                key = Convert.ToBase64String(keyBytes);
-                hash = Convert.ToBase64String(sha256.ComputeHash(keyBytes));
-            }
-            
-            Func<HttpRequestMessage> createGetRequest = () => new HttpRequestMessage()
-            {
-                Method = HttpMethod.Get,
-                Headers = {
-                    { "x-goog-encryption-algorithm", "AES256" },
-                    { "x-goog-encryption-key", key },
-                    { "x-goog-encryption-key-sha256", hash }
-                }
+                var request = new HttpRequestMessage { Method = HttpMethod.Get };
+                key.ModifyRequest(request);
+                return request;
             };
 
             _fixture.RegisterDelayTest(_duration,
                 beforeDelay: async duration =>
                 {
-                    // TODO: Replace this with calls to the StorageClient if/when it supports customer-supplied encryption keys.
-                    var putRequest = new HttpRequestMessage()
-                    {
-                        Content = new ByteArrayContent(data),
-                        Method = HttpMethod.Put,
-                        Headers = {
-                            { "x-goog-encryption-algorithm", "AES256" },
-                            { "x-goog-encryption-key", key },
-                            { "x-goog-encryption-key-sha256", hash }
-                        }
-                    };
-
-                    putRequest.RequestUri = new Uri(_fixture.UrlSigner.Sign(
-                        bucket,
-                        name,
-                        expiration: null,
-                        request: putRequest));
-                    var response = await _fixture.HttpClient.SendAsync(putRequest);
-                    Assert.True(response.IsSuccessStatusCode);
+                    var encryptingClient = StorageClient.Create(encryptionKey: key);
+                    encryptingClient.UploadObject(bucket, name, "application/octet-stream", new MemoryStream(data));
 
                     // Make sure the encryption succeeded.
                     await Assert.ThrowsAsync<GoogleApiException>(
@@ -232,7 +203,7 @@ namespace Google.Cloud.Storage.V1.IntegrationTests
                     request.RequestUri = new Uri(url);
 
                     // Verify that the URL works initially.
-                    response = await _fixture.HttpClient.SendAsync(request);
+                    var response = await _fixture.HttpClient.SendAsync(request);
                     var result = await response.Content.ReadAsByteArrayAsync();
                     Assert.Equal(data, result);
                 },
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/EncryptionKeyTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/EncryptionKeyTest.cs
@@ -1,0 +1,128 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using Xunit;
+
+namespace Google.Cloud.Storage.V1.Tests
+{
+    public class EncryptionKeyTest
+    {
+        [Fact]
+        public void Generate()
+        {
+            EncryptionKey key1 = EncryptionKey.Generate();
+            EncryptionKey key2 = EncryptionKey.Generate();
+
+            Assert.NotNull(key1.Base64Key);
+            Assert.NotNull(key1.Base64Hash);
+
+            Assert.NotEqual(key1.Base64Key, key2.Base64Key);
+            Assert.NotEqual(key1.Base64Hash, key2.Base64Hash);
+        }
+
+        [Fact]
+        public void Create()
+        {
+            EncryptionKey key = EncryptionKey.Create(new byte[32]);
+            Assert.Equal("AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=", key.Base64Key);
+            Assert.Equal("Zmh6rfhivXdsj8GLjp+OIAiXFIVu4jOzkCpZHQ1fKSU=", key.Base64Hash);
+        }
+
+        [Fact]
+        public void Create_Invalid()
+        {
+            Assert.Throws<ArgumentNullException>(() => EncryptionKey.Create(null));
+            Assert.Throws<ArgumentException>(() => EncryptionKey.Create(new byte[33]));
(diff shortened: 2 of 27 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.cs
```diff
@@ -67,13 +69,18 @@ public static string ApplicationName
...
         /// <summary>
         /// Constructs a new client wrapping the given <see cref="StorageService"/>.
         /// </summary>
         /// <param name="service">The service to wrap. Must not be null.</param>
-        public StorageClientImpl(StorageService service)
+        /// <param name="encryptionKey">Optional <see cref="EncryptionKey"/> to use for all object-based operations by default. May be null.</param>
+        public StorageClientImpl(StorageService service, EncryptionKey encryptionKey = null)
         {            
             Service = GaxPreconditions.CheckNotNull(service, nameof(service));
+            EncryptionKey = encryptionKey;
```
evildour: What do you think about having this fallback to `EncryptionKey.None` so the `EncryptionKey` property is always non-null?

jskeet: I wondered about that. Yes, that's probably worth doing.

jskeet:
The removals of ModifyRequest are just bugs left over from a previous attempt - will fix :)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/EncryptionKey.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            using (var rng = RandomNumberGenerator.Create())
+            {
+                byte[] key = new byte[32];
+                rng.GetBytes(key);
+                return Create(key);
+            }
+        }
+
+        internal void ModifyRequest(HttpRequestMessage request)
```
evildour: It might be helpful to expose this for use with the signed URL uploading: for the `ModifySessionInitiationRequest` of the options passed to `SignedUrlResumableUpload.Create`, one can just create or reuse an `EncryptionKey` instance and delegate off to its ability to modify the request.

jskeet: Yup, that sounds fine to me.

jskeet:
Ready for actual review now - @evildour if you could have a second look to check the changes I made in the URL signer tests, that would be handy too.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UrlSignerTest.cs
```diff
@@ -179,41 +179,24 @@ private void GetWithCustomerSuppliedEncryptionKeysTest_InitDelayTest()
...
-                }
+                var request = new HttpRequestMessage { Method = HttpMethod.Get };
+                key.ModifyRequest(request);
+                return request;
             };
 
             _fixture.RegisterDelayTest(_duration,
                 beforeDelay: async duration =>
                 {
-                    // TODO: Replace this with calls to the StorageClient if/when it supports customer-supplied encryption keys.
```
evildour: Actually, the intent was to replace everything from here down to `Assert.True(response.IsSuccessStatusCode);` with a call up UploadObjectAsync. This test isn't meant to test uploading with a signed url, it was just the easiest way to get it up there so I could test downloading (although now that I think of it, I should probably add a test for this as well: non-resumable uploading with encryption keys. I'll add this once this PR is merged).

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/EncryptionKeyTest.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        [Fact]
+        public void ModifyRequest_None()
+        {
+            var request = new HttpRequestMessage();
+            EncryptionKey.None.ModifyRequest(request);
+            IEnumerable<string> values;
+            Assert.False(request.Headers.TryGetValues(EncryptionKey.AlgorithmHeader, out values));
+            Assert.False(request.Headers.TryGetValues(EncryptionKey.KeyHeader, out values));
+            Assert.False(request.Headers.TryGetValues(EncryptionKey.KeyHashHeader, out values));
```
evildour: It might be good to verify all headers are missing, to protect against copy/paste errors in future maintenance.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/EncryptionKeyTest.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        [Fact]
+        public void ModifyRequestForRewriteSource_None()
+        {
+            var request = new HttpRequestMessage();
+            EncryptionKey.None.ModifyRequestForRewriteSource(request);
+            IEnumerable<string> values;
+            Assert.False(request.Headers.TryGetValues(EncryptionKey.CopySourceAlgorithmHeader, out values));
+            Assert.False(request.Headers.TryGetValues(EncryptionKey.CopySourceKeyHeader, out values));
+            Assert.False(request.Headers.TryGetValues(EncryptionKey.CopySourceKeyHashHeader, out values));
```
evildour: Here too

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CopyObjectOptions.cs
```diff
@@ -98,6 +98,18 @@ public sealed class CopyObjectOptions
...
+        /// The encryption key to use for this operation. If this property is null, the <see cref="StorageClient.EncryptionKey"/>
+        /// will be used instead. Use <see cref="EncryptionKey.None"/> to remove encryption headers from this request.
+        /// </summary>
+        public EncryptionKey EncryptionKey { get; set; }
+
+        /// <summary>
+        /// The encryption key to use for the source of the copy. The <see cref="EncryptionKey"/> and
+        /// <see cref="StorageClient.EncryptionKey"/> properties are used for the target of the copy.
+        /// </summary>
+        public EncryptionKey SourceEncryptionKey { get; set; }
```
evildour: Should this default to using the client's encryption key as well? It seems like a copy should be analogous to a download then upload using the client, but a download then upload wouldn't require specifying a key if the object were already encrypted with the client's key whereas a copy would.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/EncryptionKey.cs
```diff
@@ -0,0 +1,121 @@
...
+using System.Net.Http;
+using System.Security.Cryptography;
+
+namespace Google.Cloud.Storage.V1
+{
+    /// <summary>
+    /// An AES-256 key passed to the Google Cloud Storage servers via headers to allow objects to be encrypted at
+    /// rest using a client-supplied key rather than a server-supplied default key.
+    /// </summary>
+    public sealed class EncryptionKey
```
evildour: This seems like it should provide `Equals` and `GetHashCode` implementations.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.cs
```diff
@@ -54,18 +54,26 @@ public abstract partial class StorageClient
...
+
+        /// <summary>
         /// Asynchronously creates a <see cref="StorageClient"/>, using application default credentials if
         /// no credentials are specified.
         /// </summary>
         /// <remarks>
         /// The credentials are scoped as necessary.
         /// </remarks>
         /// <param name="credential">Optional <see cref="GoogleCredential"/>.</param>
+        /// <param name="encryptionKey">Optional <see cref="EncryptionKey"/> to use for all object-based operations by default.</param>
```
evildour: Nit: may want to add "May be null" to match the async version.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/EncryptionKeyTest.cs
jskeet: Do you mean the CopySource headers as well?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UrlSignerTest.cs
jskeet: Right - I wondered, but wasn't sure. Will see how simple that turns out to be :)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CopyObjectOptions.cs
jskeet: Hmm. Really unsure, to be honest. It could easily be a copy from somewhere else, at which point things would break if we didn't have anything. Given that it's easy enough to specify none, I'll make that change...

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.cs
jskeet: Done.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/EncryptionKey.cs
jskeet: I guess. It's certainly easy enough to do.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UrlSignerTest.cs
jskeet: Hmm... I need to go out in 5 minutes and I can't quite see how to do this at the moment. I wonder whether we should have an option to make UrlSigner accept an EncryptionKey. What do you think?

jskeet:
I think I've addressed all the issues other than the UrlSignerTest, which I'll try to look at tomorrow. (Or we could merge as-is, maybe with a TODO...)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UrlSignerTest.cs
evildour: Adding the EncryptionKey as a parameter to the Sign methods on UrlSigner seems like a good idea. I can add that in a separate PR.

As for my comment about using UploadObjectAsync, I simply mean that everything from here down to line 204 should be replaced by this:
```csharp
await _fixture.Client.UpdateObjectAsync(bucket, name, new UploadObjectOptions { EncryptionKey = key });
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/EncryptionKeyTest.cs
evildour: Yes

jskeet: Great, done. (All tests check all 6 headers now.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UrlSignerTest.cs
jskeet: Right, I'm with you. Done.

jskeet:
Thanks for the review and feedback. All done and squashed into one commit. Will merge on green.

PR #748, opened 2017-02-07
Pass a label with the agent and version to all log entries

iantalarico:
This will allow the agent that is sending logs to be recorded.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/CommonUtils.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/CommonUtils.cs
@@ -19,17 +19,17 @@ namespace Google.Cloud.Diagnostics.Common
 {
     internal static class CommonUtils
     {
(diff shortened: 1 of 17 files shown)


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/CommonUtils.cs
```diff
@@ -20,5 +21,22 @@ internal static class CommonUtils
     {
         /// <summary>A completed <see cref="Task"/>.</summary>
         internal readonly static Task CompletedTask = Task.FromResult(false);
+
+        /// <summary>The name of the this agent.</summary>
+        internal const string AgentName = "google-cloud-csharp-diagnostics";
```
evildour: It might be better to keep the constants above all the statics, but I haven't been paying close enough attention to know what the convention is.

iantalarico: I think that is the convention, fixed.

Issue #704, opened 2017-01-23
Publish ASP.NET Error Reporting MVC Filter


iantalarico:
The code was submitted in PR #697 but needs to be pushed to nuget.

Expected to be out by EOW

iantalarico:
@csells FYI

csells:
We'll need the How-To docs updated when this PR goes through.

csells:
And we need to update the ER docs to show how to enable the ER API or no exceptions can happen once the app is deployed (and it will throw exceptions).

jskeet:
Will update https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.ErrorReporting.V1Beta1/index.html when I've pushed the release (just to show the difference between MVC and Web API) but that won't include the API enabling or any other How-To docs. Ian, are you happy to own that?

iantalarico:
I will add the API enabling and other how to docs.

iantalarico:
I got side tracked by some related work to push events through the logs api to the error reporting api.  I will make updates to the ER doc about the change, how to use the error reporting api if desired (and in that case that the api must be enabled).

PR #749, opened 2017-02-07
Update the Error Reporting documentation to note the Logging API is used by default.

iantalarico:
The change also shows how to use the Error Reporting API directly and how to enable the Error Reporting API.

This adds the needed documentation required in issue #704 

cc @steren


apis/Google.Cloud.Diagnostics.AspNet/docs/index.md
```diff
@@ -26,12 +26,26 @@ installed, run the following command in a Google Cloud SDK Shell:
 
 ## Registering Error Reporting (Web Api)
 
+Reports error events to the Stackdriver Error Reporting API via the Stackdriver Logging API (default
+and suggested method).  For more information see
+["Formatting Log Error Messages"](https://cloud.google.com/error-reporting/docs/formatting-error-messages).
```
evildour: "Stackdriver Error Reporting API via the Stackdriver Logging API" is a little confusing, especially since the alternative also mentions "Stackdriver Error Reporting API".

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.Snippets/AspNetSnippets.cs
```diff
@@ -25,28 +25,61 @@ namespace Google.Cloud.Diagnostics.AspNet.Snippets
 {
     public class AspNetSnippets
     {
-        // Sample: RegisterExceptionLoggerWebApi
-        public static void Register(HttpConfiguration config)
+        private class ErrorReporting_LoggingApi
```
jskeet: Yes, this should be fine, I think. The processing is pretty crude :)

iantalarico:
After thinking about this more I (updated) and moved the documentation into the code where the options are.  Most users probably will not need to use this option or need to change anything.  I think adding the docs where I did would just be more confusing.

PTAL

iantalarico:
Thanks Mike!

PR #764, opened 2017-02-08
Factor out Common Code for Error Reporting Unit Tests.

iantalarico:


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ErrorReporting/ErrorReportingMatching.cs
```diff
@@ -0,0 +1,133 @@
...
+{
+    internal class ErrorReportingMatching
+    {
+        internal const int ConflictStatusCode = 409;
+        internal const string ServiceName = "SomeService";
+        internal const string VersionName = "1.0.0";        
+        internal const string UserAgentValue = "user-agent-1.0";
+        internal static readonly Uri GoogleUri = new Uri("https://www.google.com");
+        internal static readonly string DeleteMethod = "DELETE";
+        internal static readonly string ExceptionMessage = "some exception message";
```
evildour: These two can be const

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ErrorReporting/ErrorReportingMatching.cs
```diff
@@ -0,0 +1,133 @@
...
+        ///   - The HttpMethod matches the default http method
+        ///   - The Url matches the default url
+        ///   - The UserAgent is empty or null
+        ///   - The ResponseStatusCode is "0"
+        ///   - The LineNumber is "0"
+        ///   - The FilePath is empty or null
+        ///   - The FunctionName is empty or null
+        ///   - The Service is <see cref="ServiceName"/>
+        ///   - The Version is <see cref="VersionName"/>
+        /// </summary>
```
evildour: Not important here since this isn't public, but just for future reference, making bulleted lists like this won't work when this gets turned into docs. I think the dashes will just wrap like any other character. You'd need to do something like this instead:

```csharp
/// <summary>
/// Something:
/// <list type="bullet">
/// <term>A</term>
/// <term>B</term>
/// <term>C</term>
/// </list>
/// </summary>
```

jskeet: Actually using docfx, it recognizes Markdown and does the right thing - which is handy as our API docs in the protos are in Markdown and copied verbatim. But yes, using a list is probably a better idea :)

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.Tests/ErrorReporting/ErrorReportingExceptionFilterTest.cs
```diff
@@ -51,98 +44,43 @@ private ExceptionContext CreateSimpleContext()
             return new ExceptionContext
             {
                 HttpContext = mockContext.Object,
-                Exception = new Exception()
+                Exception = new Exception(ErrorReportingMatching.ExceptionMessage)
```
jskeet: Can we not call CreateException because that would include a stack trace?

PR #765, opened 2017-02-08
Move AspNet Trace Integration Tests and Misc Clean Up.

iantalarico:
Move the AspNet Trace Integration Tests to Common Integration tests and rename to SimpleManagedTracerTest as this is all that is being tested and SimpleManagedTracer is common code.


iantalarico:
Thanks!

PR #775, opened 2017-02-14
Integration tests and updated comments for the TraceHeaderPropagatingHandlerTest

iantalarico:


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/Trace/TraceHeaderPropagatingHandlerTest.cs
```diff
@@ -0,0 +1,144 @@
...
+            tracer.StartSpan(rootSpanName);
+            var traceHeaderHandler = TraceHeaderPropagatingHandler.Create(tracer);
+            using (var httpClient = new HttpClient(traceHeaderHandler))
+            {
+                try
+                {
+                    await httpClient.GetAsync(uri);
+                    Assert.False(shouldThrow);
+                }
+                catch
```
evildour: Do you mean to catch the thrown exception when `shouldThrow` is true? If so, you may want to add that to the param docs above. They currently make it sound like the `TraceOutGoingRequest` call will throw when it is true, but it will not.

iantalarico: Updated docs to be more clear

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/Trace/TraceHeaderPropagatingHandlerTest.cs
```diff
@@ -0,0 +1,144 @@
...
+        /// Creates a <see cref="TraceHeaderPropagatingHandler"/> and traces the sending of a
+        /// GET request to the given uri.  The trace is wrapped in a parent span.
+        /// </summary>
+        /// <param name="tracer">The tracer to trace the request with.</param>
+        /// <param name="rootSpanName">The name of the root span that will wrap the span
+        ///     that traces the request.</param>
+        /// <param name="uri">The uri to request.</param>
+        /// <param name="shouldCatch">True if exceptions from the request to the uri should be caught.</param>
+        private async Task TraceOutGoingRequest(
+            IManagedTracer tracer, string rootSpanName, string uri, bool shouldCatch)
```
evildour: This seems slightly more confusing. You always catch, regardless of the value of `shouldCatch`. But then if `shouldCatch` is false, you re-throw a different exception. However, if `shouldCatch` is true, you force an exception to be thrown which is then caught. Perhaps this is clearer and still in line with what you're trying to do here:

```csharp
try
{
    await httpClient.GetAsync(uri);
}
catch when (shouldCatch)
{
}
```

evildour: Actually, looking closer at how this is being used, it looks like you want as assertion failure for the test if `shouldCatch` is true but no exception is actually thrown by `await httpClient.GetAsync(uri)`, right? In that case, neither the original nor my rewritten code will accomplish this. You want something like this:

```csharp
try
{
    await httpClient.GetAsync(uri);
    Assert.False(shouldCatch);
}
catch (Exception e) when (shouldCatch && !(e is Xunit.Sdk.XunitException))
{
}
```

Also, since I think I have my head around what's going on here a bit better, I think `exceptionExpected` might be a better name for this parameter.

iantalarico: Agreed the naming choice was poor.  I updated the names, comments and exception check

iantalarico:
Thanks Mike, PTAL

PR #776, opened 2017-02-14
Rename enums *ToLocation to *Target.

### LINKED ISSUE #739: Consider Renaming LogTo and ReportEventsTo
LogTo and LogToLocation are used to send logs to either a project or organization.
https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LogTo.cs

ReportEventsTo and ReportEventsToLocation are used to send error events to either the error reporting api or the logging api.
https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ErrorReporting/ReportEventsTo.cs

cc @jskeet @csells

iantalarico:
Fixes #739


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ErrorReporting/ReportEventsTo.cs
```diff
@@ -43,7 +43,7 @@ public sealed class ReportEventsTo
         internal static readonly MonitoredResource GlobalResource = new MonitoredResource { Type = "global" };
 
         /// <summary>The location to send error events to.</summary>
-        public ReportEventsToLocation ReportEventsToLocation { get; private set; }
+        public EventTarget Location { get; private set; }
```
evildour: You may want to change this property name to `Target` to match the new naming, but I don't feel too strongly about it.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/LogTo.cs
```diff
@@ -36,37 +36,37 @@ public enum LogToLocation
     public sealed class LogTo
     {
         /// <summary>The location to log entries to.</summary>
-        public LogToLocation Location { get; }
+        public LogTarget Location { get; }
```
evildour: Same here.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ErrorReporting/ReportEventsTo.cs
```diff
@@ -22,7 +22,7 @@ namespace Google.Cloud.Diagnostics.Common
     /// <summary>
     /// The location error events will be sent to.
     /// </summary>
-    public enum ReportEventsToLocation
+    public enum EventTarget
```
evildour: Should the class be renamed as well? If the enum name was not good for some reason, I'm wondering if the same reasoning applies to the class also.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ErrorReporting/ReportEventsTo.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/LogTo.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ErrorReporting/ReportEventsTo.cs
iantalarico: I don't think the class name is as bad (maybe still not great) as the enum names.

However, I'm open to suggestions and I will change it if you feel strongly about it.

evildour: No, I don't feel too strongly about it. I do like the fluent syntax of something like `ReportEventsTo.Logging(...)` but in other contexts the class name feels a bit awkward. I don't have a better idea though. If @jskeet is ok with it, I am too.

jskeet: I'm not *hugely* happy with it to be honest. It's a more detailed EventTarget, it feels like...  I think I'd personally call the class EventTarget and the enum EventTargetKind or EventTargetType.

How much are end-users going to use this? The degree of bike-shedding that makes sense will depend on that :)

iantalarico: So my rational for not changing the class was like Mike said ReportEventsTo.Logging(...) is easy to parse.

Users will almost exclusivly interact at the class level ReportEventsTo.Logging(...).  The enum more of less internal.  That being said I'd rather have a proper name than not

jskeet: `ReportEventsTo.Logging` sounds imperative though - whereas really the result needs to be used, doesn't it?

```csharp
// Looks okay, but is a bug, because the return value isn't used
ReportEventsTo.Logging(...);

// Correct, but having a value of type "ReportEventsTo" seems weird
var reportEventsTo = ReportEventsTo.Logging(...);
```

If we had `EventTarget` you could have:

```csharp
// Sounds wrong - which is great, because it *is* wrong
EventTarget.ForLogging(...);

// Looks right and is right
var target = EventTarget.ForLogging(...);
```

But maybe I'm missing something?

iantalarico: You aren't missing anything (I was :-) )

iantalarico:
Thanks Jon and Mike! All fixed

iantalarico:
The failure seems to be an issue with mono? 
"Got a SIGABRT while executing native code. This usually indicates
a fatal error in the mono runtime or one of the native libraries 
used by your application."

I ran this again on a ubuntu machine and didn't get any failures.  I'm going to merge and I will revert if this causes issues

PR #735, opened 2017-02-02
[WIP] log4net sourceLocation and platform detection

chrisdunelm:
If location is enabled, puts source location data in the source_location field of the LogEntry proto.

Detects platform (GCE, GAE, & Unknown) and uses the correct ResourceType and resource labels.

PR #777, opened 2017-02-14
Attempt to automatically detect the current platform (monitored resource).

iantalarico:
Heavily based PR #735 by @chrisdunelm (thanks Chris :) )


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/MonitoredResourceUtils.cs
```diff
@@ -0,0 +1,68 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
```
chrisdunelm: This class (or something similar to it) should probably go in Google.Apis.Logging.v2
Then this code and the log4net code can share it.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/MonitoredResourceUtils.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// <summary>The global resource.</summary>
+        public static readonly MonitoredResource GlobalResource = new MonitoredResource { Type = "global" };
+
+        /// <summary>
+        /// Attempts to detect the current <see cref="MonitoredResource"/> this is being run in.
+        /// If the monitored resource cannot be detected the global monitored resource is used.
+        /// </summary>
+        public static MonitoredResource DetectCurrentResource()
+        {
+            var platform = Platform.Instance();
```
chrisdunelm: Just to make sure you're aware that Platform.Instance() can block for up to 1 second the first time it's called.

iantalarico: Yes, this is only ever called at start up.  So I *think* this is worthwhile.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/MonitoredResourceUtils.cs
iantalarico: I think that would be a good idea, not point in duplicating the code.

My only worry is that (IIRC) MonitoredResource isn't a logging specific thing.  I think it's general for many Google APIs.

chrisdunelm: Yes, you're right.
It's actually in GAX ([MonitoredResource.cs](https://github.com/googleapis/gax-dotnet/blob/master/src/Google.Api.CommonProtos/Api/MonitoredResource.cs)), which means this common code can go there.
Perhaps leave this as it is for now, and file an issue to move this code into GAX, which I'll do soon.

chrisdunelm:
Are the test failures anything serious?

iantalarico:
They shouldn't be failing they are passing locally. Investigating now.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/MonitoredResourceUtils.cs
iantalarico: Created an issue and cc'd you

iantalarico:
Looks like auto detection was working :) 

Fixed tests.

PR #778, opened 2017-02-15
First code drop for Google Cloud Spanner APIs

jskeet:
First commit is just codegen (and a tweak to generateapis.sh)

Second commit is just project files, changes to make gRPC classes partial, and add appropriate CreateOperationsClient methods.

There's no significant hand-written code yet, or documentation. We fully expect to add a hand-written abstraction layer over this one over time, but work has not started on that yet.


jskeet:
(AppVeyor failure was unrelated.)

PR #783, opened 2017-02-16
Update the timed buffer to automatically flush buffers after the given wait time.

iantalarico:
1. Update the timed buffer to automatically flush buffers after the given wait time
2. The timed buffer does not wait for a receive call to flush.
3. Use the timed buffer as the default buffer.

These changes will give users an over all better experience. Previously entries could possibly never shown up with the sized buffer (it never filled) or the timed buffer (if another receive never occurred).

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
@@ -21,56 +21,50 @@ using System.Threading;
 namespace Google.Cloud.Diagnostics.Common
 {
     /// <summary>
-    /// A <see cref="IFlushableConsumer{T}"/> that will flush the buffer during a receive
-    /// call if the given amount of time has passed..
+    /// A <see cref="IFlushableConsumer{T}"/> that will automatically flush the buffer after a 
+    /// given amoutn of time.
     /// </summary>
     internal class TimedBufferingConsumer<T> : FlushableConsumerBase<T>
     {
         /// <summary>The consumer to flush to.</summary>
         private readonly IConsumer<T> _consumer;
 
-        /// <summary>The minimum amount of time to wait between automatically flushing the buffer.</summary>
-        private readonly TimeSpan _waitTime;
-
-        /// <summary>A clock for getting the current timestamp.</summary>
-        private readonly IClock _clock;
-
         /// <summary>
         /// The buffered items. This is not readonly as it is replaced when the buffer is flushed.
         /// </summary>
         private List<T> _items = new List<T>();
 
-        /// <summary>The earliest time of the next automatica flush.</summary>
-        private DateTime _nextFlush;
+        /// <summary>The timer to automatically flush the buffer.</summary>
+        private IThreadingTimer _timer;
 
-        private TimedBufferingConsumer(IConsumer<T> consumer, TimeSpan waitTime, IClock clock)
+        internal TimedBufferingConsumer(IConsumer<T> consumer, TimeSpan waitTime, IThreadingTimer timer)
         {
             _consumer = GaxPreconditions.CheckNotNull(consumer, nameof(consumer));
-            _waitTime = waitTime;
-            _clock = GaxPreconditions.CheckNotNull(clock, nameof(clock));
-            _nextFlush = _clock.GetCurrentDateTimeUtc().Add(_waitTime);
+            _timer = GaxPreconditions.CheckNotNull(timer, nameof(timer));
+            // Initialize the timer to flush ever wait time interval.
+            _timer.Initialize((e) => { Flush(); }, waitTime);
+        }
+
+        ~TimedBufferingConsumer()
+        {
+            Flush();
+            _timer.Dispose();
         }
 
         /// <summary>
-        /// Creates a new <see cref="TimedBufferingConsumer{T}"/> that will flush to the
-        /// given <see cref="IConsumer{T}"/> during a <see cref="Receive(IEnumerable{T})"/> call after
-        /// the preset wait time since the last flush has passed.
+        /// Creates a new <see cref="TimedBufferingConsumer{T}"/> that will automatically flush the 
+        /// buffer to the <see cref="IConsumer{T}"/> after the given wait time.
         /// </summary>
         /// <param name="consumer">The consumer to flush to, cannot be null.</param>
-        /// <param name="waitTime">The minimum amount of time between automatic flushes.</param>
-        /// <param name="clock">A clock for getting the current timestamp.</param>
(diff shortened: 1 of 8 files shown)


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
```diff
@@ -21,56 +21,50 @@
...
-            _nextFlush = _clock.GetCurrentDateTimeUtc().Add(_waitTime);
+            _timer = GaxPreconditions.CheckNotNull(timer, nameof(timer));
+            // Initialize the timer to flush ever wait time interval.
+            _timer.Initialize((e) => { Flush(); }, waitTime);
+        }
+
+        ~TimedBufferingConsumer()
+        {
+            Flush();
+            _timer.Dispose();
```
evildour: It would probably be good to implement the [dispose pattern](https://msdn.microsoft.com/en-us/library/b1yfkh5e(v=vs.110).aspx) on this class and then dispose these consumers where they're used (or create them in using statements).

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
```diff
@@ -21,56 +21,50 @@
...
-        private TimedBufferingConsumer(IConsumer<T> consumer, TimeSpan waitTime, IClock clock)
+        internal TimedBufferingConsumer(IConsumer<T> consumer, TimeSpan waitTime, IThreadingTimer timer)
         {
             _consumer = GaxPreconditions.CheckNotNull(consumer, nameof(consumer));
-            _waitTime = waitTime;
-            _clock = GaxPreconditions.CheckNotNull(clock, nameof(clock));
-            _nextFlush = _clock.GetCurrentDateTimeUtc().Add(_waitTime);
+            _timer = GaxPreconditions.CheckNotNull(timer, nameof(timer));
+            // Initialize the timer to flush ever wait time interval.
+            _timer.Initialize((e) => { Flush(); }, waitTime);
```
evildour: Do you think it would be common to have these consumers with nothing waiting to be flushed? If so, maybe `IThreadingTimer` should support stopping (via `_timer.Change(Timeout.Infinite, Timeout.Infinite)`) and re-starting. Then if `FlushWithSemaphoreHeld...` is called when there are no items, the timer would be stopped and then restarted on the next `ReceiveWithSemaphoreHeld...` call. Perhaps it's more trouble than it's worth though.

iantalarico: I think this really depends on the service that is using this, it could be frequent for one and not for another.

I think this is a good idea, I just need to work out how to be sure that Receive calls don't always reset the timer.

I'll do this in a follow up PR if that's okay.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
iantalarico: I had thought about this but these consumer are used in middleware and I'm not quite sure how to properly dispose of things in all cases.

Do you have any insight here?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
evildour: Yeah, I was worried about the same thing so that's why I was thinking to stop the timer on a flush if there are no items. That way the timer would only stop if you went a whole cycle without receiving anything. Then when receiving, you would only restart the timer if it was stopped.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
evildour: I think it usually makes sense to implement the dispose pattern if the class owns anything disposable. I don't know if anything about this situation gives a reason to go against that, but I honestly don't have much experience with ASP.NET, so I'm not sure. @jskeet, do you have any thoughts on this?

jskeet: I generally agree with making it disposable, and I really don't like having a finalizer.
Do we know what the middleware is going to do? Will it ever dispose of this? How many are going to be created?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
jskeet: I'd be tempted not to complicate things, to be honest. I don't think it will add much benefit - a timer ticking when it doesn't need to isn't going to add any real load.

iantalarico: I think this is the best option. 

Will do in a follow up PR as I think this will break a lot of tests :)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
iantalarico: Sounds reasonable.  I'll also do this in a separate PR as this will require changing a few interfaces (which will require a lot of changes)

It might be out of the scope of this PR to talk about how to properly dispose of things.  Hopefully Jon will have some insight here

evildour: [This section](https://msdn.microsoft.com/en-us/library/b1yfkh5e(v=vs.110).aspx#basic_pattern) of the link above shows how to properly dispose of things. Basically, if someone calls Dispose directly, you want to dispose everything you own that's disposable or needs cleanup. You also want to prevent the finalizer from running. If instead the finalizer is run, you only want to cleanup unmanaged resources. Other managed, disposable objects will have their finalizers run as well and that might happen before the current object is disposed.

evildour:
LGTM, assuming the dispose pattern will come later or Jon thinks it's unnecessary.

jskeet:
Sorry, hadn't spotted that I'd failed to submit the review. Would rather get the disposal done sooner rather than later... finalizers make me wince.

iantalarico:
The dispose pattern will come in a PR today. It will just require changing a lot more files.

iantalarico:
Talked with Jon offline.  Merging for now.  PR to come soon.

PR #786, opened 2017-02-16
Fix broken integration tests from a previous refactor.

iantalarico:


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -70,6 +70,19 @@ private void BlockUntilClockTick()
             }
         }
 
+        /// <summary>Create a thrown exception.</summary>
+        internal Exception CreateException()
```
jskeet: Presumably the difference is that this has a stack trace because it was thrown?

iantalarico: Correct, this was me missing this when moving this over to "common"

PR #788, opened 2017-02-16
Use common MonitoredResource builder, and fix log4net appender

chrisdunelm:
The log4net appender was non-deterministic in two situations: (1) When uploading a 'lost message' log entry; and (2) When uploading pre-log log entries. These have now been made deterministic.


chrisdunelm:
@iantalarico there's a test failure that I'll fix tomorrow; I'll probably ask Jon to have a quick look, then merge it. And yes, I'll put a simple comment on the LogQueuePeekResult class. Thanks :)

PR #790, opened 2017-02-20
Regenerate all APIs

jskeet:
As far as I can see, the changes are only:

- Copyright dates
- Lack of _clientHelper field
- Snippets populating requests


chrisdunelm:
Is the error due to #730 not yet being implemented?
Although it looks like it may be worth passing the clientHelper to OnConstruction.
I can put #730 in the codegen today if required.

jskeet:
Whoops, I hadn't seen that there was an error.

Urgh - yes, implementing #730 would fix that, and yes, passing in the client helper would probably be the best approach, to avoid the duplicate logic.

chrisdunelm:
OK, PR coming soon...

chrisdunelm:
https://github.com/googleapis/toolkit/pull/1044

jskeet:
Right, should be fixed now... let's see if it goes green :)

chrisdunelm:
Green :) I'll take a look now.

PR #795, opened 2017-02-20
Regenerate all APIs.

jskeet:
(This brings in changes from codegen for resource names and doc
comment escaping.)


jskeet:
Urgh - we don't want to submit this. Some API docs apparently use actual HTML, whereas others are using HTML in the resource names. Will raise internally...

jskeet:
Actually, we may want to submit this anyway. The doc comments for messages are already HTML-escaped, e.g. https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.ErrorReporting.V1Beta1/api/Google.Cloud.ErrorReporting.V1Beta1.GetGroupRequest.html - it's not nice, but it's probably better than the Spanner docs not being valid XML docs.

chrisdunelm:
Yes, agreed we should merge this as is.

Issue #803, opened 2017-02-24
Properly Dispose of the ErrorReportingExceptionFilter


iantalarico:
We currently do not properly clean up the ErrorReportingExceptionFilter's Resources

iantalarico:
Dropping to P2 as the resources are alive as long as the application

jskeet:
Closing for now having added this to the [backlog](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/BACKLOG.md)

amanda-tarafa:
ErrorReportingExceptionFilter is from the Diagnostics.AspNet package which we don't support (except for critical bug fixes) as per  #4828.
Removing this issue from backlog.

Issue #802, opened 2017-02-24
Properly Dispose of all ASP.NET Core Middleware


iantalarico:
We currently do not properly clean up the ASP.NET Core Middleware's Resources

iantalarico:
This should be easier once the IMiddleware interface is released.

iantalarico:
Dropping to P2 as the resources are alive as long as the application

jskeet:
Is this still an issue? Are we ever likely to address it? If not, let's close it.

iantalarico:
It looks like the interface isn't in 1.0 so it cannot be used.

PR #804, opened 2017-02-24
Dispose of ASP.NET Core bits

iantalarico:
Properly dispose of ASP.NET Core bits.

1. Services registered with HttpConfiguration that implement IDisposable are cleaned up automatically.
2. We dispose of CloudTrace via Event Handlers that are called during dispose.
3. #803 To track proper disposal for ErrorReportingExceptionFilter.
4. #802 To track proper disposal for ASP.NET Core Middleware.


apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet/Trace/CloudTrace.cs
```diff
@@ -110,8 +112,14 @@ public static void Initialize(string projectId, HttpApplication application, Tra
             // Add event handlers to the application.
             application.BeginRequest += trace.BeginRequest;
             application.EndRequest += trace.EndRequest;
+            application.Disposed += trace.Disposed;
```
jskeet: I'd just use a lambda expression or anonymous method here, rather than an extra method.

iantalarico: Done.

iantalarico:
Thanks Jon

PR #806, opened 2017-02-24
Make StartSpan return an IDisposable.

### LINKED ISSUE #791: Make StartSpan return an IDisposable?
Looking at the sample code here:

```csharp
tracer.StartSpan(nameof(TraceHelloWorld));
Console.Out.WriteLine("Hello, World!");
tracer.EndSpan();
```

makes me wonder what will happen if an exception is thrown before I call `EndSpan`... or what would happen if I forget to call `EndSpan`. Would it make sense to have `StartSpan` return a disposable object, so I could use:

```csharp
using (tracer.StartSpan(nameof(TraceHelloWorld))
{
    Console.Out.WriteLine("Hello, World!");
}
```

iantalarico:
This will greatly simplify ending spans especially if an exception is thrown.

Also adds RunInSpan which runs a function in a span and captures a stacktrace if one occurs. The RunInSpan changes includes a change to span options so users can easily pass in labels.

Also fixes an old issue where the simple managed tracer mutex was static and should have been readonly instead.

Fixes #791

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/IManagedTracer.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/IManagedTracer.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 
@@ -27,12 +28,37 @@ namespace Google.Cloud.Diagnostics.Common
         /// </summary>
         /// <param name="name">The name of the span, cannot be null.</param>
         /// <param name="options">The span options to override default values.</param>
-        void StartSpan(string name, StartSpanOptions options = null);
+        /// <returns>
+        /// An <see cref="IDisposable"/> that will end the current span.  If the span cannot
+        /// end in the current scope, disposal can be skipped and <see cref="EndSpan"/> can
+        /// be called at a later time (this is not recommended and should be avoided if possible).
+        /// </returns>
+        IDisposable StartSpan(string name, StartSpanOptions options = null);
 
         /// <summary>
-        /// Ends the current span.
+        /// Ends the current span if not ended by the <see cref="IDisposable"/> from
+        /// <see cref="StartSpan(string, StartSpanOptions)"/>.
         /// </summary>
         void EndSpan();
+
+        /// <summary>
+        /// Runs the function in a span and will add a stacktrace from a thrown
+        /// exception (the exception will be re-thrown) to the span.
+        /// </summary>
+        /// <param name="action">The action to run in a span.</param>
+        /// <param name="name">The name of the span, cannot be null.</param>
+        /// <param name="options">The span options to override default values.</param>
+        void RunInSpan(Action action, string name, StartSpanOptions options = null);
+
+        /// <summary>
+        /// Runs the function in a span and will add a stacktrace from a thrown
+        /// exception (the exception will be re-thrown) to the span.
+        /// </summary>
+        /// <param name="func">The function to run in a span.</param>
+        /// <param name="name">The name of the span, cannot be null.</param>
+        /// <param name="options">The span options to override default values.</param>
+        /// <returns>The result from the call to <paramref name="func"/></returns>
+        T RunInSpan<T>(Func<T> func, string name, StartSpanOptions options = null);
 
         /// <summary>
         /// Annotates the current span with the given labels. 
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/IManagedTracer.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/IManagedTracer.cs
@@ -29,8 +29,8 @@ namespace Google.Cloud.Diagnostics.Common
         /// <param name="name">The name of the span, cannot be null.</param>
         /// <param name="options">The span options to override default values.</param>
         /// <returns>
-        /// An <see cref="IDisposable"/> that will end the current span.  If the span cannot
-        /// end in the current scope, disposal can be skipped and <see cref="EndSpan"/> can
+        /// An <see cref="IDisposable"/> that will end the current span when disposed.  If the span
+        /// cannot end in the current scope, disposal can be skipped and <see cref="EndSpan"/> can
         /// be called at a later time (this is not recommended and should be avoided if possible).
         /// </returns>
         IDisposable StartSpan(string name, StartSpanOptions options = null);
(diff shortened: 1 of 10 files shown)


jskeet:
Will review on Monday.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/IManagedTracer.cs
```diff
@@ -27,14 +28,39 @@ public interface IManagedTracer
         /// </summary>
         /// <param name="name">The name of the span, cannot be null.</param>
         /// <param name="options">The span options to override default values.</param>
-        void StartSpan(string name, StartSpanOptions options = null);
+        /// <returns>
+        /// An <see cref="IDisposable"/> that will end the current span.  If the span cannot
```
jskeet: Nit: "that will end the current span when disposed"?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -28,8 +28,20 @@ namespace Google.Cloud.Diagnostics.Common
...
+        /// <summary>
+        /// A class that represents a running trace span.
+        /// </summary>
+        private class Span : IDisposable
+        {
+            private readonly SimpleManagedTracer _tracer;
+            public Span(SimpleManagedTracer tracer) { _tracer = tracer; }
+
+            /// <summary> Ends the current span.</summary>
+            public void Dispose() =>_tracer.EndSpan();
```
jskeet: Nit: space after `=>`

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -28,8 +28,20 @@ namespace Google.Cloud.Diagnostics.Common
...
+        }
+
         /// <summary>A mutex to protect the rate limiter instance.</summary>
-        private static object _stackLock = new object();
+        private readonly object _stackLock = new object();
```
jskeet: Definitely approve of this being readonly - being non-static is a really big change. Is that deliberate? My guess is that it is, but I thought I'd check. (Also, should the summary comment be updated?)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -90,6 +100,43 @@ public void StartSpan(string name, StartSpanOptions options = null)
...
+        /// <inheritdoc />
+        public void RunInSpan(Action action, string name, StartSpanOptions options = null)
+        {
+            using (StartSpan(name, options))
+            {
+                try
+                {
+                    action();
+                }
+                catch (Exception e)
```
jskeet: It may be better to use an exception filter here:

```
catch (Exception e) when (SetStackTraceAndReturnFalse(e))
{
}

private bool SetStackTraceAndReturnFalse(Exception e)
{
    SetStackTrace(new StackTrace(e, true));
    return false;
}
```

Definitely test carefully before making this change, because that changes the timing of when the stack trace will be acquired, which could change things significantly... but it does make the exception handling more transparent in terms of final stack.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -90,6 +100,43 @@ public void StartSpan(string name, StartSpanOptions options = null)
...
+        /// <inheritdoc />
+        public T RunInSpan<T>(Func<T> func, string name, StartSpanOptions options = null)
+        {
+            using (StartSpan(name, options))
+            {
+                try
+                {
+                    return func();
+                }
+                catch (Exception e)
```
jskeet: Obviously apply the same change here as above, if we do it at all.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/StartSpanOptions.cs
```diff
@@ -22,16 +25,22 @@ public sealed class StartSpanOptions
         /// <summary>Gets the span kind.</summary>
         public SpanKind SpanKind { get; }
 
-        private StartSpanOptions(SpanKind spanKind)
+        /// <summary>Gets the lables to be added to the span.</summary>
```
jskeet: lables => labels

Perhaps this change should be in a separate PR? It's not obvious to me how it relates to the IDisposable aspect.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
iantalarico: I gave this a quick test and seems to be doing the right thing.  Updated.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/StartSpanOptions.cs
iantalarico: Done.

I updated the PR comment. We need this change for the new RunIn functions as otherwise its not as easy to pass in labels to the spans and would require the user to do it in the function/action being traced.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -208,6 +206,17 @@ public string GetCurrentTraceId()
             }
         }
 
+        /// <summary>
+        /// Sets a <see cref="StackTrace"/> on the current span for the given exception and
+        /// returns false.  This is used for exception handling to ensure proper timing for
+        /// recording the stacktrace.
```
jskeet: It's more that it avoids sometimes losing data in the stack trace that's eventually recorded, but the details are somewhat detailed :)

Did you check that this still behaves the way we want it to?

iantalarico: Yes. I updated the comment I also added and extra unit test for this.

iantalarico:
Thanks Jon!

PR #810, opened 2017-02-27
Make DoNothingTracer public and rename to NullManangedTracer

### LINKED ISSUE #809: Make DoNothingTracer public, possibly after a rename
For testing and/or code reuse purposes, it would be useful to be able to create a tracer that doesn't do anything.

We already have the implementation (DoNothingManagedTracer) - we just need to expose it after potentially renaming it, e.g. NullManagedTracer (to match [NullLogger](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/NullLogger.cs)).

iantalarico:
1. Making DoNothingTracer public will allow user to use it for testing.
2. Renaming to NullManagedTracer to match [NullLogger](https://github.com/aspnet/Logging/blob/dev/src/Microsoft.Extensions.Logging.Abstractions/NullLogger.cs).
3. Adding comments to be clear what each function does.

Fixes #809


iantalarico:
Thanks Jon!

PR #812, opened 2017-02-28
Regenerate Spanner API with extra resource names

jskeet:


jskeet:
I fail at git. I keep using git commit -a when I don't mean to...

PR #816, opened 2017-03-01
Regenerate all APIs.

jskeet:
There are effectively two sets of changes here:

- This is the first regeneration with gRPC 1.1, which includes
  more XML comments and makes all classes partial. (The XML comments
  have improved indentation, too.)
- This includes the latest Vision API changes.


jskeet:
(Self-merging when CI is green.)

PR #825, opened 2017-03-01
Regenerated APIs with retry changes

### LINKED ISSUE #822: Idle gRPC channels cause Unavailable exceptions
Currently only seen on GAE Flex and GKE, but without diagnosing why, it's hard to say how wide the effect would be.

Symptoms: after a leaving a Datastore or Logging channel open for a while, make a request which isn't classified as idempotent. It will fail with a status of Unavailable, but a subsequent request will work.

jskeet:
Fixes #822.

(All the changes should be the same, but a once-over would be appreciated just for the first few, to make sure this really is sensible.)


jskeet:
Self-merging after all - have had a close enough look. (And it's all just generated.)

PR #833, opened 2017-03-02
Use the IHttpContextAccessor to get the IManagedTracer.

iantalarico:
We do this as the IHttpContextAccessor is a singleton and will always have the proper HttpContext while IServiceProvider does not.

Without this traces will be incorrectly reported.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
@@ -74,6 +74,9 @@ namespace Google.Cloud.Diagnostics.AspNetCore
     /// </remarks>
     public static class CloudTraceExtension
     {
+        ///<summary>The key to save the tracer under in the <see cref="HttpContext"/></summary> 
+        internal const string TraceKey = "Google.Cloud.Diagnostics.AspNetCore.Trace";
+
         /// <summary>
         /// Uses middleware that will trace time taken for all subsequent delegates to run.
         /// The time taken and metadata will be sent to the Stackdriver Trace API. To be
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -121,7 +124,15 @@ public static void UseGoogleTrace(this IApplicationBuilder app)
         /// </summary>
```
jskeet: Don't we also want to change the AddScoped calls a few lines above to AddSingleton? Basically everything becomes a singleton, via the magic of IHttpContextAccessor?

iantalarico: I don't think so I think we still need the IManagedTracer to be scoped right?  Unless we are going to return Func<IManagedTracer> instead of IManagedTracer?

jskeet: Ah yes, you're right. At least I think you are. I find all of this very confusing.
An alternative would be make IManagedTracer singleton-like and make it find the right trace behind the scenes. We should definitely get other ASP.NET Core folks to look at this...

iantalarico: I think you're right though.  Otherwise it's putting too much on the user to figure out scoping of things.

jskeet: How about we make sure this works, then we can improve it in another release?

iantalarico: Yea that works for me.  This works as is if IManagedTracer is used in a scoped manner.

I think the other issue we are seeing is an issue with the propagation of the headers.

Issue #840, opened 2017-03-02
Add a parameter Func<HttpRequest, bool?> to Google.Cloud.Diagnostics.AspNet.CloudTrace.Initialize


iantalarico:
This will allow users to make decisions about what is traced when a trace header is not set.

jskeet:
So this is the AspNet version of the existing AspNetCore code?

iantalarico:
Correct, I have a few changes made to the AspNetCore bits I needs to apply to the AspNet bits.

iantalarico:
Fixed via #877

PR #841, opened 2017-03-03
Add a parameter Func<HttpRequest, bool?> to the cloud trace set up.

iantalarico:
This will allow users to make decisions about what is traced when a trace header is not set.

I also opened PR #840 for the non core lib

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Trace/CloudTraceExtensionTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Trace/CloudTraceExtensionTest.cs
@@ -19,30 +19,71 @@ using Moq;
 using System;
 using Xunit;
 
+using static Google.Cloud.Diagnostics.AspNetCore.CloudTraceExtension;
+
 namespace Google.Cloud.Diagnostics.AspNetCore.Tests
 {
     public class CloudTraceExtensionTest
     {
-        private const string TraceId = "105445aa7843bc8bf206b12000100f00";
-        private const ulong SpanId = 81237123;
+        private const string _traceId = "105445aa7843bc8bf206b12000100f00";
+        private const ulong _spanId = 81237123;
+
+        /// <summary>
+        /// Create an <see cref="IServiceProvider"/> to be used during tests of 
+        /// <see cref="CloudTraceExtension.CreateTraceHeaderContext(IServiceProvider)"/>.
+        /// It will set up an <see cref="IHttpContextAccessor"/> with the passed in trace header,
+        /// a <see cref="ShouldTraceRequest"/> with the passed in should trace function and a
+        /// <see cref="TraceIdFactory"/>.
+        /// </summary>
+        private IServiceProvider CreateProviderForTraceHeaderContext(
+            string traceHeader, Func<HttpRequest, bool?> shouldTraceFunc = null)
+        {
+            var context = new DefaultHttpContext();
+            var request = new DefaultHttpRequest(context);
+            request.Headers[TraceHeaderContext.TraceHeader] = traceHeader;
+
+            var accessor = new HttpContextAccessor();
+            accessor.HttpContext = context;
+
+            var traceIdFactory = TraceIdFactory.Create();
+
+            Mock<IServiceProvider> mockProvider = new Mock<IServiceProvider>();
+            mockProvider.Setup(p => p.GetService(typeof(IHttpContextAccessor))).Returns(accessor);
+            mockProvider.Setup(p => p.GetService(typeof(ShouldTraceRequest))).Returns(
+                new ShouldTraceRequest { ShouldTrace = shouldTraceFunc });
+            mockProvider.Setup(p => p.GetService(typeof(TraceIdFactory))).Returns(traceIdFactory);
+            return mockProvider.Object;
+        }
 
         [Fact]
         public void CreateTraceHeaderContext()
         {
-            string header = $"{TraceId}/{SpanId};o=1";
-
-            var context = new DefaultHttpContext();
-            var request = new DefaultHttpRequest(context);
-            request.Headers[TraceHeaderContext.TraceHeader] = header;
-
-            var accessor = new HttpContextAccessor();
-            accessor.HttpContext = context;
-            
-            Mock<IServiceProvider> mockProvider = new Mock<IServiceProvider>();
-            mockProvider.Setup(p => p.GetService(typeof(IHttpContextAccessor))).Returns(accessor);
-
-            var headerContext = CloudTraceExtension.CreateTraceHeaderContext(mockProvider.Object);
+            var header = $"{_traceId}/{_spanId};o=1";
+            var provider = CreateProviderForTraceHeaderContext(header);
+            var headerContext = CloudTraceExtension.CreateTraceHeaderContext(provider);
             Assert.Equal(TraceHeaderContext.FromHeader(header).ToString(), headerContext.ToString());
+        }
+
+        [Fact]
+        public void CreateTraceHeaderContext_ForceTrace_False() =>
+            TestCreateTraceHeaderContextForceTrace(false);
+
+        [Fact]
+        public void CreateTraceHeaderContext_ForceTrace_True() =>
+            TestCreateTraceHeaderContextForceTrace(true);
+
+        [Fact]
+        public void CreateTraceHeaderContext_ForceTrace_Null() =>
+            TestCreateTraceHeaderContextForceTrace(null);
+
+        [Fact]
+        public void CreateTraceHeaderContext_ForceTrace_NoHeader()
+        {
+            var provider = CreateProviderForTraceHeaderContext("", ((HttpRequest r) => true));
+            var headerContext = CloudTraceExtension.CreateTraceHeaderContext(provider);
+            Assert.NotNull(headerContext.TraceId);
+            Assert.Equal((ulong)0, headerContext.SpanId);
+            Assert.Equal(true, headerContext.ShouldTrace);
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Trace/CloudTraceExtensionTest.cs
```diff
@@ -19,33 +19,74 @@
...
+        public void CreateTraceHeaderContext()
+        {
+            var header = $"{_traceId}/{_spanId};o=1";
+            var provider = CreateProviderForTraceHeaderContext(header);
+            var headerContext = CloudTraceExtension.CreateTraceHeaderContext(provider);
             Assert.Equal(TraceHeaderContext.FromHeader(header).ToString(), headerContext.ToString());
         }
 
         [Fact]
+        public void CreateTraceHeaderContext_ForceTrace_False() =>
```
jskeet: You can just use a parameterized test for this. See [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/BucketValidationTest.cs#L22) for an example.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -95,25 +105,35 @@ public static void UseGoogleTrace(this IApplicationBuilder app)
...
+        /// <param name="shouldTraceFunc">Optional function to trace requests. If the trace header is not set
+        ///     then this function will be called to determine if a given request should be traced.  This will
+        ///     not override trace headers. If the function returns true the request will be traced, if false
+        ///     is returned the trace will not be traced and if null is returned it will not affect the
+        ///     trace decision.</param>
         public static void AddGoogleTrace(
             this IServiceCollection services, string projectId,
-            TraceConfiguration config = null, TraceServiceClient client = null)
+            TraceConfiguration config = null, TraceServiceClient client = null,
+            Func<HttpRequest, bool?> shouldTraceFunc = null)
```
jskeet: Possibly rename to traceOverridePredicate? Definitely put Predicate in the name to mimic LINQ. (Admittedly the fact that it can return null makes it slightly different, but only slightly.)

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -78,6 +78,16 @@ public static class CloudTraceExtension
         internal const string TraceKey = "Google.Cloud.Diagnostics.AspNetCore.Trace";
 
         /// <summary>
+        /// A class to wrap the "shouldTraceFunc" param in <see cref="AddGoogleTrace"/>.
+        /// This is used instead of injecting the value of "shouldTraceFunc" directly 
+        /// to ensure we do not override a service the user setup.
+        /// </summary>
+        internal class ShouldTraceRequest
+        {
+            internal Func<HttpRequest, bool?> ShouldTrace { get; set; }
```
jskeet: Make this immutable, and a sealed class?
(It's annoying that it's needed. I wonder if there's some other way of marking things for dependency injection using attributes etc...)

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -95,25 +105,35 @@ public static void UseGoogleTrace(this IApplicationBuilder app)
...
                  new GrpcTraceConsumer(client), MessageSizer<TraceProto>.GetSize, config.BufferOptions);
 
             var tracerFactory = new ManagedTracerFactory(projectId, consumer,
-                RateLimitingTraceOptionsFactory.Create(config), TraceIdFactory.Create());
+                RateLimitingTraceOptionsFactory.Create(config), traceIdFactory);
 
             services.AddSingleton<IManagedTracerFactory>(tracerFactory);
             services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
             services.AddSingleton(CreateTraceHeaderPropagatingHandler);
+            services.AddSingleton(new ShouldTraceRequest { ShouldTrace = shouldTraceFunc });
```
jskeet: Thinking about it, we don't really need to register this with DI. We *could* just make anything that needs it know about it by capturing it here. I don't have much opinion about which is better, but we can change it later anyway. (If we captured it, we wouldn't need the extra class.)

We could capture the trace factory as well - not sure about the HttpContextAccessor. (I don't know what happens if you ask for the IHttpContextAccessor immediately after adding it...)

Let me know if you're not sure what I mean by all this - it'll be easier to show in a prototype PR.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -137,12 +157,24 @@ private static TraceHeaderPropagatingHandler CreateTraceHeaderPropagatingHandler
...
+            var shouldTraceRequest = provider.GetServiceCheckNotNull<ShouldTraceRequest>();
+            var traceIdFactory = provider.GetServiceCheckNotNull<TraceIdFactory>();
+
+            string header = accessor.HttpContext?.Request?.Headers[TraceHeaderContext.TraceHeader];
+            var traceHeaderContext = TraceHeaderContext.FromHeader(header);
+            if (traceHeaderContext.ShouldTrace == null && shouldTraceRequest.ShouldTrace != null)
+            {
+                string traceId = traceHeaderContext.TraceId ?? traceIdFactory.NextId();
+                ulong spanId = traceHeaderContext.SpanId ?? 0;
+                bool? shouldTrace = shouldTraceRequest.ShouldTrace(accessor.HttpContext?.Request);
```
jskeet: Perhaps we should only generate a new trace ID if `shouldTrace` returns true? (If it returns false, we're not going to want to trace, so do we need a traceId/spanId anyway? If it returns null, presumably we can just return traceHeaderContext.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -137,12 +157,24 @@ private static TraceHeaderPropagatingHandler CreateTraceHeaderPropagatingHandler
...
+                return new TraceHeaderContext(traceId, spanId, shouldTrace);
+            }
```
jskeet: Just return traceHeaderContext in the line below rather than calling FromHeader again

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
jskeet: Just as one extra thought - I originally suggested taking HttpRequest... can you think of any reason we might want to pass the whole HttpContext into the predicate instead? Just something to consider - I think it would be pretty unusual, to be honest.

iantalarico: I had thought about this initially but couldn't find a compelling reason why we would need the context and thought it may make things a little more confusing.

However due to the "magic of HttpContext" if users really want it they can get it from the Request as it has a HttpContext field.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
iantalarico: While I agree we can probably works this out and make it a little smoother the main issue here is that most these classes are common code and unfortunately HttpContext is asp.net core specific.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
iantalarico: Yea we don't need to generate it otherwise.  Updated.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
jskeet: The change I'm thinking about is *only* in this file. It's probably simpler to show than describe though, so I'll create that as a follow-on PR.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
jskeet: SGTM.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -137,13 +162,29 @@ private static TraceHeaderPropagatingHandler CreateTraceHeaderPropagatingHandler
...
+            var traceHeaderContext = TraceHeaderContext.FromHeader(header);
+            if (traceHeaderContext.ShouldTrace == null && shouldTraceRequest.ShouldTrace != null)
+            {
+                bool? shouldTrace = shouldTraceRequest.ShouldTrace(accessor.HttpContext?.Request);
+                string traceId = traceHeaderContext.TraceId;
+                if (shouldTrace == true && traceHeaderContext.TraceId == null)
+                {
+                    traceId = traceIdFactory.NextId();
+                }
+                ulong spanId = traceHeaderContext.SpanId ?? 0;
```
jskeet: If shouldTrace is null or false, why are we creating a non-null spanId?
If shouldTrace is null, I'd expect us to just return traceHeaderContext - it should be a no-op.

If it's true, we may need to generate IDs. If it's false, we may or may not need any existing trace/span ID - not sure.

jskeet:
Not sure whether you'll have seen the comment I put about the refactoring of the DI part, as it was on an outdated commit, but if you look at the first commit you should be able to see it.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
iantalarico: Yea you are correct. My bad, fixed.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
iantalarico: SGTM, thanks!

PR #843, opened 2017-03-03
Regenerate Datastore API

jskeet:
The parameter flattening for querying has been removed in
the configuration. Later on, there will be flattening with
a oneof type for the query, but for now we can readd the
previous methods using a partial class.


jskeet:
I've had some issues with getting the snippets to work, but if you could review this and trust me to only merge on green (and highlight any *significant* further changes) that would be great :)

PR #855, opened 2017-03-07
Support for async and overlapping traces

evildour:


jskeet:
Before I review the actual implementation, I'd like to understand the intended scope of the change. It feels like we have three things we need to do:

- Remove the potential deadlock we think we've discovered in the current code
- Make sure that spans flow across async contexts
- Work out where a trace tree should split - if two deliberately-parallel execution flows from the same request occur, we want to split into two branches at that point. It's not clear to me how we detect this.

Am I right in saying this PR addresses the first two but not the last? If so, have you had any thoughts about the last bit? I think we at least want to know how it can work before we know whether a fix to the first two is going to be viable.

evildour:
I don't think I was aware of a potential deadlock. If that was the case due to the locks, this has likely fixed it because there are no longer locks. It definitely addresses the 2nd two items. I have a unit test in this PR specifically designed to make sure the split into two branches works correctly. Things fall out of the `AsyncLocal/CallContext` naturally because the `ExecutionContext` is set to copy on write as an async region is _entered_. So if you do something like
- Start span A
- Call async method (which marks the context for copy on write)
- Start span B
- await on something

You will come back out to the caller and the stack of spans will now only contain span A. The copy generated when starting span B would be restored when the awaited _something_ finishes and the callee continues.

jskeet:
Thanks very much for the extra detail. I'm not sure whether in my jetlagged state I'll be able to really reason about async stuff clearly, but I'll see what I can do :)

evildour:
Actually, rather than solving any potential deadlock issues, this seems to have introduced one. Both CI builds failed. Presumably because the test to split into branches is trying to run everything on the same thread. Is this expected when running tests on the CI servers?

evildour:
Perhaps it's because I forced a `Wait()` rather than awaiting for the operations to finish. I'll update things to see if that will fix it.

evildour:
Hmm, well that fixed the deadlock and made travis pass but it made AppVeyor fail. I'll look into this some more.

evildour:
It was an intermittent issue. Looks like I was a little too hasty removing the locks from everything. Although the logical call context for a particular thread can no longer be modified by another at the same time, the `TraceProto` is shared and therefore can. When adding the completed span to the `TraceProto` and possibly calling `Flush`, that should have been synchronized. About to push another commit and hopefully that will fix it.

evildour:
Ok, I think things are good now. PTAL.

iantalarico:
I stared at this and think I've wrapped my head around this. But forgive me if I'm way off here.

I think we will still have some issues with where spans are branched out.  If I understand AsyncLocal correctly we can only ever have the parent of a span be the rootSpanId or the another span in the current thread.  Does that sounds right?

If so I think we want to be able to have spans from other threads possibly be the parents right otherwise we will have spans with their parent be the root span when they should be the span from another thread.

evildour:
Actually, the execution context flows from a parent to a child thread so for something like this:

Thread 1 - start span A
Thread 1 - kick off thread 2
Thread 2 - start span B
Thread 2 - end span B
Thread 1 - end span A

Span B will have span A as a parent. I added a unit test to verify this and it passes, but I don't want to push it just yet because there are some other issues with this:

Issue 1 - For this scenario:

Thread 1 - kick off thread 2
Thread 1 - start span A
Thread 2 - start span B
Thread 2 - end span B
Thread 1 - end span A

Span B will not be a child of span A because the "copy" of the trace stack it received when it was kicked off was empty. I'm not sure if this is a huge issue, but can see it going either way. If you feel strongly about it, I'll see if there's a way to get the parent thread's logical context, but it seems like it might not be possible.

Issue 2 - For this scenario:

Thread 1 - start span A
Thread 1 - kick off thread 2
Thread 1 - end span A
Thread 2 - start span B
Thread 2 - end span B

Things are a bit worse: span B is never flushed! This is because the "copy" of the trace stack it received when it was kicked off had A in it, so B's parent is A, and when B ends, the stack copy it has is non-empty, so it is waiting for A to finish. I think we can solve this though by continually popping off anything form the stack which already has an EndTime set at two points: before starting a span and after ending a span (with some appropriate locking since it will be looking at shared state). This should prevent B becoming a child of A in this scenario and allow it to be flushed. And it will also allow B to be flushed when A ends between B starting and ending.

evildour:
Some more thoughts on this: I don't think Issue 1 is actually a problem. Just because span A overlaps B in time doesn't mean it should be the parent of B. Thread 2 was kicked off from thread 1 before span A was opened and then thread 1 continued doing its own thing, which may not be related to what thread 2 is doing, so I don't think anything else needs to be done there.

As far as Issue 2 goes, my proposed solution feels a bit messy. I was thinking instead that it would just be cleaner to keep a counter of open spans and then (atomically) increment it when a span is opened and decrement it when a span is closed. Flush when it gets to zero. Although I think I would still do the "pop off all ended spans" strategy before starting a span, so we don't make a span a child of an already ended span. Not sure on this though. For a similar reason that I think Issue 1 is ok, one could argue that if thread 2 was kicked off while span A was open, span A should be an ancestor of all spans within thread 2.

Thoughts?

iantalarico:
For issue 1 I think you are correct.  I think in this case a span should only be a parent if it was kicked off during another span.  I think the threads will do the right thing here.

I need to think about Issue 2 a little more but I think this should be true "if thread 2 was kicked off while span A was open, span A should be an ancestor of all spans within thread 2." as span A started thread two but this may make for some odd looking traces depending on how things finish.

evildour:
Yeah, I think they may look a bit odd, but it feels like the right thing to do. Do we know how other libraries handle this?

evildour:
It looks like in Go and NodeJS you need to explicitly get the parent span so that you can open a child span within in (see [here](https://github.com/GoogleCloudPlatform/google-cloud-go/blob/master/trace/trace.go#L141) and [here](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/test/test-span-data.js#L106)). And in Ruby, it appears that spans are relative to a particular thread (see [here](https://github.com/GoogleCloudPlatform/google-cloud-ruby/blob/master/google-cloud-trace/lib/google/cloud/trace.rb#L292)). Same with Java (see [here](https://github.com/GoogleCloudPlatform/cloud-trace-java-instrumentation/blob/master/jdbc/driver/src/main/java/com/google/cloud/trace/jdbc/CloudTraceSdkTraceService.java#L50) and [here](https://github.com/GoogleCloudPlatform/cloud-trace-java-instrumentation/blob/master/jdbc/driver/src/main/java/com/google/cloud/trace/jdbc/Tracers.java#L67)).

iantalarico:
Agreed that it may be odd but seems right. 

I've been working with the Java team on most of this and the general model with Java seems like the best match/model for C# IMO.

I do have an action item to create an unmanaged tracer (similar to nodejs and go) that would give the user more control but would also be more work.  The priority though is to have the managed tracer to GA first.

evildour:
Ok so just to recap: we will keep the stack as is at all times, so issues 1 continues and the part about A owning B from issue 2 continues as well. To avoid the non-flush of span B from issue 2, I will introduce an atomic counter which we will use to determine when the flush needs to take place. I'll make these changes after #857 is merged (since I need it to write some more tests for these scenarios) unless anyone thinks my recap is wrong.

iantalarico:
Sounds good to me.  I'll try and go a full review of the code.  Note I closed #857 as I think you are correct the API taking a TraceProto is poor.  I'll try to get out a new PR today too.

evildour:
Great, thanks Ian. No rush.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -158,16 +159,19 @@ public T RunInSpan<T>(Func<T> func, string name, StartSpanOptions options = null
...
-            lock (_stackLock)
-            {
-                CheckStackNotEmpty();
+            var currentStack = TraceStack;
+            CheckStackNotEmpty(currentStack);
 
-                TraceSpan span = _traceStack.Pop();
-                span.EndTime = Timestamp.FromDateTime(DateTime.Now.ToUniversalTime());
+            TraceSpan span;
+            currentStack = currentStack.Pop(out span);
```
iantalarico: Do we need to lock on this line and the one below?

evildour: I don't believe so. As it is an immutable stack, the Pop call above doesn't nothing to the receiver. Rather it returns a new stack without the top item. The line below does change the stack currently stored in the AsyncLocal/logical call context, but that is a per-thread/per-async block storage location, which means there is no chance of race condition.

iantalarico: Ah okay sgtm

evildour:
Thanks but I'm going to push another PR after your changes to fix the flush issue. I need that to do proper testing of these new threading scenarios. But like I said, no rush.

jskeet:
Okay - will wait for your ping, Mike :)

evildour:
PTAL. For some reason AppVeyor thinks the PR is not mergeable but the merge button is active. Any idea how to get it to refresh itself?

jskeet:
Hmm... have you rebase so that this is just on top of the latest changes?

evildour:
Never mind, I just pushed another commit and that seems to have worked.

evildour:
Yup, I did rebase. And since I rebased, all four commits showed up at the bottom of the history, but your comment was in the middle of the group for me, so perhaps it got into a bad state.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -254,6 +256,203 @@ public void MultipleSpans()
...
+                    IsValidSpan(spans[3], "grandchild-two", parentId: childTwoId) &&
+                    IsValidSpan(spans[4], "root");
+            };
+            mockConsumer.Setup(c => c.Receive(Match.Create(matcher)));
+
+            Func<string, string, Task> op = async (childName, grandchildName) =>
+            {
+                using (tracer.StartSpan(childName))
+                {
+                    await Task.Delay(30);
```
jskeet: Are these just to enforce it not running synchronously? If so, `Task.Yield` might be a better option.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -254,6 +256,203 @@ public void MultipleSpans()
...
+                    }
+                    await Task.Delay(30);
+                }
+            };
+
+            Task t1;
+            Task t2;
+            using (tracer.StartSpan("root"))
+            {
+                t1 = Task.Run(() => op("child-one", "grandchild-one").Wait());
```
jskeet: Why not just `t1 = op(...)` with an `await Task.Yield()` at the start of `op`?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -254,6 +256,203 @@ public void MultipleSpans()
...
+                t2 = Task.Run(() => op("child-two", "grandchild-two").Wait());
+            }
+
+            Predicate<IEnumerable<TraceProto>> childMatcher = (IEnumerable<TraceProto> t) =>
+            {
+                // Verify that even though the child spans were started after the root span was ended, they
+                // still have it as a parent since it was opened when their threads were spawned.
```
jskeet: Nit: I'd change this to "when their tasks were created" - there's no guarantee that it'll be starting actual new threads, nor that any new threads would be running by the time we hit the end of the `using` statement. Unless I've misunderstood something, of course - which would probably be related to the previous comment.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -17,7 +17,13 @@
 using Google.Protobuf.WellKnownTypes;
 using System;
 using System.Collections.Generic;
+using System.Collections.Immutable;
 using System.Diagnostics;
+#if !NETSTANDARD1_5
+// TODO: Remove when we support .NET 4.6.1
```
jskeet: I don't think we'll be supporting *only* net461 (as a lowest version) any time even slightly soon... we could potentially have another target of net461 if that were useful though.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -92,27 +109,20 @@ public IDisposable StartSpan(string name, StartSpanOptions options = null)
...
-                    span.ParentSpanId = _traceStack.Peek().SpanId;
-                }
-                else if (_rootSpanParentId != null)
-                {
-                    span.ParentSpanId = (ulong)_rootSpanParentId;
-                }
+                SpanId = _spanIdFactory.NextId(),
+                Kind = options.SpanKind.Convert(),
+                Name = name,
+                StartTime = Timestamp.FromDateTime(DateTime.Now.ToUniversalTime()),
```
jskeet: Use `DateTime.UtcNow`. I realize this is an extra change to the previous code, but we really shouldn't have DateTime.Now anywhere :)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -168,16 +178,21 @@ public T RunInSpan<T>(Func<T> func, string name, StartSpanOptions options = null
...
+            TraceStack = currentStack;
+            span.EndTime = Timestamp.FromDateTime(DateTime.Now.ToUniversalTime());
 
+            lock (_traceLock)
+            {
                 _trace.Spans.Add(span);
 
-                if (_traceStack.Count == 0)
+                var newOpenSpanCount = Interlocked.Decrement(ref _openSpanCount);
+                Debug.Assert(newOpenSpanCount >= 0, "Invalid open span count");
```
jskeet: Can we use GaxPreconditions.CheckState?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -254,6 +256,204 @@ public void MultipleSpans()
         }
 
         [Fact]
+        public async Task MultipleSpans_Async()
+        {
+            var mockConsumer = new Mock<IConsumer<TraceProto>>();
+            var tracer = SimpleManagedTracer.Create(mockConsumer.Object, ProjectId, TraceId);
+
+            Predicate<IEnumerable<TraceProto>> matcher = (IEnumerable<TraceProto> t) =>
```
evildour: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
evildour: They are to make sure the span starts/ends interleave in time. Done.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
evildour: I wanted to test with child threads and I wanted to make sure the child spans were not started until we setup the 2nd matcher below. Will this happen as well with your proposal?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
evildour: Done. And you're right, the threads might not be started, but the logical call context would have already flowed to the tasks in the using, which is the goal.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
evildour: Acknowledged. I removed the TODOs and changed these to look for NET45 instead so a target of .NET 4.6.1 will automatically pickup the AsyncLocal code.

evildour: I also cleaned up some of the conditional compilation blocks, but let me know if you want me to reorganize anything.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
evildour: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
evildour: This is really a "this should never happen" kind of thing. I don't know if we need the runtime overhead, although small, of checking on every span. Also, I'm a little worried about the debuggability of something like this: if you call end span too many times, you won't know until one of the last open spans are closed, which could happen later and on a different thread.

Personally, I think it would be safer to hide EndSpan and force users to dispose the result of StartSpan, and then perhaps have that disposable object do a GaxPreconditions.CheckState to ensure it is never double-disposed. But that should be for another PR.

If you'd still like me to change this though, let me know and I will.

evildour:
Thanks Jon, PTAL

evildour:
Hold off on reviewing, looks like another intermittent issue got introduced. I'll try to investigate a bit later.

evildour:
Think I figured it out. Looks like the `Yield` calls weren't quite _yielding_ as I'd expected. Once green, PTAL.

jskeet:
Will have to look tomorrow - too tired to think about async tonight, and will be in the air in a couple of hours. But if the changes are minor test ones, I'm happy to check it later if you'd like to merge it now.

evildour:
No problem. I can leave it unmerged until you have a chance to review.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -364,19 +364,29 @@ public void MultipleSpans()
...
                     }
-                    await Task.Yield();
+
+                    do
+                    {
+                        await Task.Yield();
+                        // Don't allow the span to close until both are started
+                        // to make sure we only get a single Receive call below
+                        // instead of two.
+                    } while (startedChildSpans < 2);
```
jskeet: I suspect this should probably be something also using Interlocked to ensure it actually reads from main memory, in theory... but unless we see hanging tests, I'm happy to assume that there's enough going on in async/await to make the JIT compiler not optimize the read away.

evildour: No, that's a good point. Fixed.

evildour:
Thanks, will merge when green.

PR #857, opened 2017-03-08
Ensure spans are cleared out when flushed.

iantalarico:


evildour:
Thinking about this some more, is this ok? One of the receivers is a timed buffer, right? So it will get a trace proto instance which it holds onto for later, but when it later goes to process it, the spans would have already been cleared out and possibly replaced. If that's the case, it almost seems like we would need to release the reference to the trace proto once it is sent to the consumer and create a new one for subsequent spans.

iantalarico:
Yes you are correct, fixed.  Thanks!

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -241,9 +241,9 @@ private void CheckStackNotEmpty()
 
         private void Flush()
         {
-            _consumer.Receive(new[] { _trace });
-            // Clear out the flushed spans.
-            _trace.Spans.Clear();
+            var old = _trace;
+            _trace = new TraceProto { TraceId = old.TraceId };
```
evildour: This looks like it should also copy over a project Id.

iantalarico:
Yea I agree the current API is not ideal.  I think I'll scrap this PR and create a new one with these changes

PR #861, opened 2017-03-10
Make the common extensions internal and not public.

iantalarico:
They are only used internally and we have no reason to expose them.


iantalarico:
Thanks Mike!

PR #862, opened 2017-03-10
Create a HttpContext Based Managed Tracer

### LINKED ISSUE #835: Consider accessing IManagedTracers from the IHttpContextAccessor
Currently for Google.Cloud.Diagnostics.AspNetCore access for IManagedTracers is done by dependency injects and each IManagedTraceris scoped to the current request.  This requires users to have to inject the scope based IManagedTracers.

We should consider passing around instead a singleton Func<IManagedTracer> that relies on IHttpContextAccessor where the user would not have to worry about scoping.

cc: @jskeet

iantalarico:
This PR Creates an IManagedTracer that is a wrapper around another IManagedTracer which it accesses via the current HttpContext.  This allows for a singleton IManagedTracer.

The IManagedTracer is now created in the middleware right before it is used and then set on the HttpContext. The current HttpContext can always be accessed by a singleton IHttpContextAccessor this allows for a singleton IManagedTracer so users do not have to deal with scoping issues of the IManagedTracer.

This change does mean users who try to trace anything before the trace middleware is registered will just be using a null managed tracer and not tracing anything.  I think this is acceptable for now as this seems like it will be a small use case and we would need to be careful about how we handle the spans.

Fixes #835 

This is a similar access model to the AspNet library.


evildour:
There are test failures. Should I hold off on reviewing until they are fixed or is it safe to review?

iantalarico:
It's safe to review.  I just submitted a fix. Thanks Mike!

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceMiddleware.cs
```diff
@@ -29,24 +29,36 @@ namespace Google.Cloud.Diagnostics.AspNetCore
...
         }
 
         /// <summary>
-        /// Invokes the next <see cref="RequestDelegate"/> and trace the time 
+        /// Invokes the next <see cref="RequestDelegate"/> and traces the time 
         /// taken for the next delegate to run, reporting the results to the
         /// Stackdriver Trace API.
         /// </summary>
-        public async Task Invoke(HttpContext httpContext, IManagedTracer tracer)
+        public async Task Invoke(HttpContext httpContext, TraceHeaderContext traceHeaderContext)
```
evildour: Might be good to add a \<param\> tag for this, which also indicates that it cannot be null.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceMiddleware.cs
```diff
@@ -29,24 +29,36 @@ namespace Google.Cloud.Diagnostics.AspNetCore
...
+        private readonly IManagedTracerFactory _tracerFactory;
+        private readonly IHttpContextAccessor _accessor;
 
         /// <summary>
         /// Create a new instance of <see cref="CloudTraceMiddleware"/>.
         /// </summary>
         /// <param name="next">The next request delegate. Cannot be null.</param>
-        public CloudTraceMiddleware(RequestDelegate next)
+        /// <param name="tracerFactory">A factory to create <see cref="IManagedTracer"/>s.</param>
+        /// <param name="accessor">Access to <see cref="HttpContext"/>.</param>
```
evildour: Indicate that these two cannot be null.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextManagedTracer.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Google.Cloud.Diagnostics.Common;
+using Microsoft.AspNetCore.Http;
+using Google.Api.Gax;
+
+namespace Google.Cloud.Diagnostics.AspNetCore
+{
+    /// <summary>
+    /// An <see cref="IManagedTracer"/> that will pull the the current tracer from
```
evildour: Nit: double "the" here

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextTracerManager.cs
```diff
@@ -0,0 +1,47 @@
...
+        public static void SetCurrentTracer(IHttpContextAccessor accessor, IManagedTracer tracer)
+        {
+            if (accessor?.HttpContext?.Items != null)
+            {
+                accessor.HttpContext.Items[TraceKey] = tracer;
+            }
+        }
+
+        /// <summary>
+        /// Gets the current <see cref="IManagedTracer"/> or a <see cref="NullManagedTracer"/> if non exists.
```
evildour: Nit: "non" => "none"

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -138,31 +135,24 @@ public static void UseGoogleTrace(this IApplicationBuilder app)
             var tracerFactory = new ManagedTracerFactory(projectId, consumer,
                 RateLimitingTraceOptionsFactory.Create(config), traceIdFactory);
 
+            services.AddScoped(CreateTraceHeaderContext);
```
jskeet: So this isn't used by anything else, it's just available if you *want* to inject it in a scoped dependency? If we didn't have this line (and the CreateTraceHeaderContext method) would anything else fail?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextManagedTracer.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// the <see cref="HttpContext"/>. The current tracer is set per request in the
+    /// <see cref="CloudTraceMiddleware"/> via the <see cref="ContextTracerManager"/>.
+    /// This allows for this to be a singleton <see cref="IManagedTracer"/> so users
+    /// do not need to worry about scoping of the tracer.
+    /// </summary>
+    internal sealed class ContextManagedTracer : IManagedTracer
```
jskeet: If we find we want a similar class elsewhere, we could have a DelegatingTracer which accepted a `Func<IManagedTracer>`, and made each call just delegate to that and then pass the arguments to it, abstracting the `ContextTracerManager.GetCurrentTracer(_accessor)` part.

Do we already have anything like this in the AspNet code?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextTracerManager.cs
```diff
@@ -0,0 +1,47 @@
...
+    /// <summary>
+    /// Manages access to the current <see cref="IManagedTracer"/>.
+    /// </summary>
+    internal static class ContextTracerManager
+    {
+        ///<summary>The key to save the tracer under in the <see cref="HttpContext"/>.</summary> 
+        internal const string TraceKey = "Google.Cloud.Diagnostics.AspNetCore.Trace";
+
+        /// <summary>
+        /// Sets the current <see cref="IManagedTracer"/>.
```
jskeet: And this is called by the middleware on each request, yes? Maybe add a comment about that.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
iantalarico: This is used in the middleware to create the managed tracer.  We could take that logic and move it here too.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceMiddleware.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceMiddleware.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextManagedTracer.cs
iantalarico: done

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextManagedTracer.cs
iantalarico: We don't have anything yet but I need to apply this general change to the ASP.NET code and will make this change then.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextTracerManager.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextTracerManager.cs
iantalarico: Done

jskeet:
Ah - so `public async Task Invoke(HttpContext httpContext, IManagedTracer tracer)` in `CloudTraceMiddleware` gets the scoped tracer injected automatically? That makes sense. Presumably we *could* use the context-based one there too, but we don't need to.

Okay, LGTM (IIRC - it's hard to keep track of where I was on this).

PR #890, opened 2017-03-16
Allow resumable upload session initiation

evildour:
This allows for scenario 2 from [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/640#issuecomment-268116549).

PR #891, opened 2017-03-16
Reuse trace test servers when possible and fix flakiness of Trace_QPS .

iantalarico:
- Reuses trace test servers when possible.
- Change the logic in Trace_QPS to be more robust.

These changes will reduce test run time and fix some test flakiness.

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.Snippets/DatastoreDbSnippets.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.Snippets/DatastoreDbSnippets.cs
@@ -56,13 +56,19 @@ namespace Google.Cloud.Datastore.V1.Snippets
             Assert.Null(entities[1]);
         }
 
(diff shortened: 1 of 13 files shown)


evildour:
You may need to rebase master. AppVeyor has an error that the UploadObjectWithSessionUri snippet can't be found, which I just added in #890.

iantalarico:
@evildour I'm already merged with the latest bits, I'll take a closer look.

iantalarico:
@evildour it looks like this failure started in #890.  Although looking at the PR I'm not sure why.  It looks correct.

iantalarico:
@jskeet @evildour  Getting off topic but I tried to build the docs locally but wasn't quite sure what to do.  Do we have something to generate docfx.jsons?

evildour:
I haven't tried building them locally, but I think I see what the problem is. About to submit a PR and will merge if it fixes the issue.

iantalarico:
I don't think we will run into interdependence if we reuse the server as long as we clean the work space up.  However, I also don't see a reason to reuse them if we don't have too.

PR #892, opened 2017-03-16
Add an option to allow other integration tests to run even if one fails

iantalarico:
Currently runintegrationtests.sh will fail fast if one integration test from a project fails this option allows for the failure to be delayed until all tests have executed.

@jskeet My bash is weak, sorry if this is horrible :)

### DIFF
--- a/runintegrationtests.sh
+++ b/runintegrationtests.sh
@@ -5,6 +5,7 @@
 set -e
 
 CONTINUE_ARG=
+CONTINUE_ON_FAILURE=false
 
 for arg in "$@"
 do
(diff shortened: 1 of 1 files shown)


iantalarico:
Can we collect a list of failed tests as we go along? Yes.  I can do that.

Should a failed tests end up in the progress file? I'm not sure.

I originally made this PR for the CI but after more thought it was cleaner to just run individual tests.  I don't have a use for this change but am happy to finish it if you see value.

jskeet:
If you have no need for it, let's shelve it for now - maybe leave the branch around, but close the PR?

iantalarico:
SGTM, no point in adding unused code :)

PR #896, opened 2017-03-17
Regenerate all APIs

jskeet:
The changes here are primarily due to codegen changes around required fields.

Will self-merge after running integration tests, assuming CI is green.

cc @chrisdunelm


chrisdunelm:
Looks like a vkit bug causing this error. I'll have a look

chrisdunelm:
gapic.yml error. Fixed in https://github.com/googleapis/googleapis/pull/292

jskeet:
Datastore tests are failing as ReadOptions is marked required and I'm not setting it. Asking internally whether it should actually be optional.

PR #901, opened 2017-03-17
Decrease flakes in Diagnostic integration tests.

iantalarico:
- Use a smaller buffer for trace buffer tests as this depends the stacktrace created being large enough.
- Only check for filepath and line number on windows as we do in other tests


iantalarico:
Thanks Jon!

Issue #903, opened 2017-03-17
Add documentation about commit limits


jskeet:

jskeet:
(Only 500 entities per commit.)

ILMTitan:
I have been thinking about this, and I think we should automatically batch for Insert requests. Insert does not make any promises about transactional logic, and if the user cared strongly, they would be using a transaction.

jskeet:
I'd far rather be consistent, to be honest. It gets complicated *really* quickly in terms of retries and the like. It also means the library doesn't need to be changed if Datastore's limits change, for example.

jskeet:
(Have asked internally for comment from the Datastore team.)

jskeet:
Confirmed that this is really an application concern. (I know that feels like a cop-out, but the number of potential options required and inconsistency between inside/outside transactions is enough.)

PR #910, opened 2017-03-20
First pass at documenting Datastore limits

jskeet:
This is to address #903, but there are pending questions about
whether we should actually be batching within the library.
(We'll need to check this thoroughly, in terms of hitting every
overload of every relevant method...)

Not ready for submission, but creating the PR so that it's easily viewable.

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
@@ -255,6 +255,9 @@ namespace Google.Cloud.Datastore.V1
         /// Call settings are not supported on this call due to restrictions with methods containing a parameter array and optional parameters.
         /// To specify options, create a collection or array explicitly, and call <see cref="AllocateIds(IEnumerable{Key},CallSettings)"/>.
         /// </para>
+        /// <para>Datastore limits the number of keys that can be allocated in a single operation. When allocating a large
+        /// number of keys, partition the allocations into batches. See
+        /// [Datastore limits](https://cloud.google.com/datastore/docs/concepts/limits) for more details on Datastore limits.</para>
         /// </remarks>
         /// <param name="keys">The incomplete keys. Must not be null or contain null elements.</param>
         /// <returns>A collection of complete keys with allocated IDs, in the same order as <paramref name="keys"/>.</returns>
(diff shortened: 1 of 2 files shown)


jskeet:
@cc ILMTitan for first review if possible. I'll have another pass at it myself...

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
```diff
@@ -274,7 +282,12 @@ public virtual IReadOnlyList<Key> AllocateIds(IEnumerable<Key> keys, CallSetting
         /// <summary>
         /// Allocates an ID for a single incomplete key asynchronously.
         /// </summary>
-        /// <remarks>This method simply delegates to <see cref="AllocateIdsAsync(IEnumerable{Key},CallSettings)"/>.</remarks>
+        /// <remarks>
+        /// <para>This method simply delegates to <see cref="AllocateIdsAsync(IEnumerable{Key},CallSettings)"/>.</para>
+        /// <para>Datastore limits the number of keys that can be allocated in a single operation. When allocating a large
```
ILMTitan: I don't think we need this comment here on the single Key method.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
```diff
@@ -566,6 +611,10 @@ public virtual Task<IReadOnlyList<Key>> InsertAsync(IEnumerable<Entity> entities
         /// <remarks>
         /// <para>On success, if entity has an incomplete key it will be updated in memory with the server-allocated keys.</para>
         /// <para>This method simply delegates to <see cref="Upsert(IEnumerable{Entity}, CallSettings)"/>.</para>
+        /// <para>Datastore limits the number of entities that can be modified in a Commit operation. Although this method does not
+        /// use an existing transaction, it still performs all the work in a single Commit operation. When modifying a large
+        /// number of entities, partition the changes into batches. See
+        /// [Datastore limits](https://cloud.google.com/datastore/docs/concepts/limits) for more details on Datastore limits.</para>
```
ILMTitan: Don't need this in the single upsert method.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
jskeet: Done

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
jskeet: Done.

jskeet:
Transaction methods: I've only got them on Lookup, because that's a separate set of limits to the Commit ones. Have added to Commit, as well as DatastoreDb.Update/InsertAsync.

Thanks for checking - will get Chris to have a look too :)

PR #913, opened 2017-03-20
Use netcoreapp1.0 for Google.Cloud.Diagnostics.AspNetCore.Snippets to match other libraries

iantalarico:


iantalarico:
It doesn't need to be but all of the others are and this makes the CI easier :)

iantalarico:
The appveyor failure looks unrelated "'curl' is not recognized as an internal or external command,"

PR #926, opened 2017-03-21
Add aspnet or aspnetcore to the agent version string.

iantalarico:
This will allow us to better see the difference between aspnet and aspnetcore usage.


iantalarico:
I'm a little confused then.  Won't we always be using common aspnet with the aspnet bits and common aspnetcore with aspnetcore bits?

Do you know of a way to properly detect this if this won't work?

jskeet:
ASP.NET Core runs perfectly well on the desktop framework too, which will resolve to net45 unless I've missed something. I would suggest passing something in from the consuming project, but i haven't checked whether that works cleanly.

iantalarico:
Alright I think this may have to be a little messier than I would like.  Thanks Jon!

PR #927, opened 2017-03-22
Translation wrapper changes suggested by the team

jskeet:
- Separate out detected and user-specified source languages
- Fix a typo in snippets
- Improve documentation for target language in ListLanguages
- Give an example with a target language in snippets

@chrisdunelm to review first, then will ask for team review before merging.

### DIFF
--- a/apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationResult.cs
+++ b/apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationResult.cs
@@ -32,10 +32,16 @@ namespace Google.Cloud.Translation.V2
         public string TranslatedText { get; }
 
         /// <summary>
-        /// The source language code, either as provided by the caller
-        /// or detected by the server.
+        /// The source language code detected by the server. This may
+        /// be null, typically due to the source language being supplied
+        /// by the caller.
         /// </summary>
-        public string SourceLanguage { get; }
+        public string DetectedSourceLanguage { get; }
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationResult.cs
```diff
@@ -32,10 +32,16 @@ public sealed class TranslationResult
...
+        /// be null, typically due to the source language being supplied
+        /// by the caller.
         /// </summary>
-        public string SourceLanguage { get; }
+        public string DetectedSourceLanguage { get; }
+
+        /// <summary>
+        /// The source language code supplied by the caller, if any.
+        /// </summary>
+        public string SpecifiedSourceLanguage { get; }
```
evildour: Since there is a `Language` class, which itself has names and codes, these names could cause a bit of confusion. Perhaps suffixing each with `...Code` would be good, although I'm not sure it's worth the extra name length.

jskeet: Definite ack to the tension here. I suspect having "code" in the documentation (and the fact that the type is `string` rather than `Language`) is enough.

evildour: Yeah I figured it probably wasn't worth it, but thought I should bring it up anyway.

jskeet:
Speculatively merging - we'll only be releasing an alpha, so there's still a chance to revise further.

PR #929, opened 2017-03-24
Regenerate all APIs using protobuf 3.2 and gRPC 1.2

jskeet:


jskeet:
(Self-assigning, as this is all autogenerated other than changing the version numbers in the script. Will merge when green.)

PR #937, opened 2017-03-27
Append git revision id label.

Deren-Liao:
The label is appened if the source context file exists and git revision
id is present in the file.


jskeet:
Is the failure just due to it using an old version? We should rebase when we've updated everything - hopefully in the next day or two - and see if that fixes it...

Deren-Liao:
It has dependent on 
https://github.com/googleapis/gax-dotnet/pull/159. 
I am waiting for that PR to be approved.
How (often) does dot net libs pick up changes from GAX?

jskeet:
Ah - I hadn't spotted that. GAX builds tend to happen "as we need them", but we're heading to V1.0 at the moment. I have a few concerns about the GAX PR - will have a closer look and comment appropriately.

PR #943, opened 2017-03-29
Add the Trace Teader to HTTP Response Headers.

iantalarico:
This is required to allow load balancers to make trace decisions properly.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceMiddleware.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceMiddleware.cs
@@ -72,7 +72,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore
                 var updatedHeaderContext = TraceHeaderContext.Create(
                     tracer.GetCurrentTraceId(), tracer.GetCurrentSpanId() ?? 0, true);
                 httpContext.Response.Headers.Add(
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceMiddleware.cs
```diff
@@ -68,6 +68,12 @@ internal sealed class CloudTraceMiddleware
             }
             else
             {
+                // Set the trace updated trace header on the response.
+                var updatedHeaderContext = TraceHeaderContext.Create(
+                    tracer.GetCurrentTraceId(), tracer.GetCurrentSpanId() ?? 0, true);
+                httpContext.Response.Headers.Add(
+                        TraceHeaderContext.TraceHeader, updatedHeaderContext.ToString());
```
jskeet: Nit: outdent by 4, or just put it all on the same line.

jskeet:
Does this leak information that users may not want leaked though? Perhaps it should be configurable? Maybe that could come in a later step - the code looks okay to me at the moment.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceMiddleware.cs
iantalarico: Fixed

iantalarico:
I think we always need this on but should only respond with the header if we received one.  Then we are not giving anything back we did not get (besides the decision to trace or not).

PR #953, opened 2017-03-30
Update dependencies

jskeet:
Easiest to review one commit at a time, but the GAX version changes to 1.0.0 then 1.0.1.


jskeet:
(Any failures in CI may be due to relying on packages which are still being indexed in nuget.org. Will hit retry until it goes green :)

jskeet:
(Appveyor cancelled manually; was still queued. Will make sure we have green appveyor before releasing.)

PR #955, opened 2017-03-30
Update version numbers for *all* packages

jskeet:
Most packages are just bumped up their existing alpha/beta level.
In addition:

- Google.Cloud.Storage.V1 goes to 1.0.0 (GA)
- Google.Cloud.Datastore.V1 goes to 1.0.0 (GA)
- Google.Cloud.Logging.V2 goes to 1.0.0 (GA)
- Google.Cloud.Logging.Type goes to 1.0.0 (GA)
- Google.Cloud.Translation.V2 goes to 1.0.0-alpha01


jskeet:
Not sure why there are 20 files here - would expect 22. Looking...

jskeet:
Ah - no, would be 23 files, but Spanner has no packages. It's fine.

Issue #847, opened 2017-03-03
Add a default filter for tracing to avoid certain traces


iantalarico:
The trace diagnostics library currently traces based on QPS however if running on app engine the health check url (/_ah/health) gets called frequently and causes spam in trace logs.  We should add a default filter for the Func<HttpRequest, bool?> param that will filter out these requests.

cc: @jskeet

iantalarico:
Reopen this as I think we should have a default in general

jskeet:
(Sorry, that comment wasn't intended to close this issue!)

PR #965, opened 2017-04-04
Add a Default Trace Http Request Predicate.

iantalarico:
This will by default not trace health checks ("/_ah/health") which overwhelm the traces on app engine flex.  The predicate is wrapped in a class that allows for easy extension by users.

Fixes issues #847


apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet/Trace/TraceDecisionPredicate.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// be traced and if null is returned it will not affect the trace decision.
+    /// </summary>
+    internal static class TraceDecisionPredicate
+    {
+        /// <summary>
+        /// Default decision predicate to determine if an <see cref="HttpRequest"/>
+        /// should be traced. If the request path is "/_ah/health" false will be
+        /// returned otherwise null is returned.
+        /// </summary>
+        public static bool? Default(HttpRequest request) =>
```
jskeet: As this is in an internal class, it doesn't make much sense for the property to be public. Did you mean to make the class public?

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet/Trace/CloudTrace.cs
```diff
@@ -114,6 +114,7 @@ public sealed class CloudTrace : IDisposable
             // Create the default values if not set.
             client = client ?? TraceServiceClient.Create();
             config = config ?? TraceConfiguration.Create();
+            traceFallbackPredicate = traceFallbackPredicate ?? TraceDecisionPredicate.Default;
```
jskeet: This is awkward. I suspect we do want the ability to trace even health requests - but *implicitly* disabling the default predicate as soon as you provide a different one feels wrong to me. I suspect most people will want it to be additive (in terms of what it filters), basically.

If TraceDecisionPredicate were public, it would at least be able to do that explicitly, although we should document it carefully (and possibly even provide a method to make it easy to build the compound predicate).

iantalarico:
Thanks for the input here Jon.  

Looking at use cases for this:

- I want to filter out health checks only (the default)
  - The user does nothing and it works out of the box.
- I want to filter out health checks and requests to Foo
  -  Make `TraceDecisionPredicate` public and they can add this to their predicate? 
- I want to filter out requests to Foo, but not health checks
  - The user passes in their own predicate. 
- I don't want to filter anything
  - The user passes in `null` - although this seems a little odd.  Maybe have a do nothing predicate to be more explicit?

I think a combination of this with better documentation should help.  Thoughts?

jskeet:
I suspect that the first two use cases are the most common, so I think we should make them the simplest. Building a predicate which does both is feasible but at least slightly fiddly - easy to get wrong, particularly with nullability etc.

We could *potentially* have an extra parameter for "don't filter health checks" - the internals can then act on a single delegate that we create for the user. The downside is that the method already has quite a few parameters. If we think it'll get any more complex than this, I'd be tempted to create a whole class for the trace decision making configuration... with only two aspects, it may not be worth it.

@evildour: Any thoughts on this?

evildour:
Yeah as I was reading through this, I think I had the same thought as you @jskeet: have a class for this. You could make `TraceDecisionPredicate` public and non-static with a factory method (or constructor) which takes a delegate and optional boolean parameter to also filter health checks (defaulted to true). Then you can give it an implicit conversion from the delegate type to the class so someone wanting to still pass in a lambda to the original method can. Or so I thought...

This causes a compilation error because the compiler won't chain user-defined implicit conversions and a lambda to a delegate is an implicit conversion:
```csharp
class Program
{
    static void Main(string[] args)
    {
        X(() => args[0]);
    }

    static void X(Foo y) { }
}

class Foo
{
    public static implicit operator Foo(Func<string> a) { return null; }
}
```

Changing the X call to either this
`X((Foo)(() => args[0]));`
or this
`X((Func<string>)(() => args[0]));`
makes it work, but that's a bit clunky.

So maybe that part won't work, but I still think having the class is the right way to go.

iantalarico:
Thanks for the feedback!

I wrapped the current default predicate in a class with options to ignore health checks and add an additional user defined predicate.

PTAL

iantalarico:
Friendly Ping, the failure is unrelated.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/TraceDecisionPredicateTest.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        [InlineData(null)]
+        public void ShouldTrace_NoHealthCheck_Fallback(bool? returnValue)
+        {
+            var request = CreateRequest("/_ah/health");
+            Func<HttpRequest, bool?> fallback = (HttpRequest r) => { return returnValue; };
```
evildour: The lambda can simply be `r => returnValue;` (and similar below), but up to you. I don't feel too strongly about it.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/TraceDecisionPredicateTest.cs
```diff
@@ -0,0 +1,92 @@
...
+            var predicate = TraceDecisionPredicate.Create(fallback, ignoreHealthChecks: false);
+            Assert.Equal(returnValue, predicate.ShouldTrace(request));
+        }
+
+        [Fact]
+        public void ShouldTrace_HealthCheck_Fallback()
+        {
+            var request = CreateRequest("/_ah/health");
+            Func<HttpRequest, bool?> fallback = (HttpRequest r) => { return false; };
```
evildour: Shouldn't this return true so the verification below can assert that the health check determined whether to trace and not the fallback predicate?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// Creates a <see cref="TraceDecisionPredicate"/>.
+        /// </summary>
+        /// <param name="traceDecisionPredicate">Optional. Used to determine if a request should be traced.
+        ///     Takes an <see cref="HttpRequest"/>.  It should return true if the request should be traced,
+        ///     false if the request should not be traced and null if the decision should be left to other
+        ///     sampling mechanisms. Can be null.</param>
+        /// <param name="ignoreHealthChecks">Optional. True if Google App Engine health check requests should
+        ///     not be traced. Defaults to true.</param>
```
evildour: These param comments (or perhaps remarks) should indicate which one takes precedence: the ignoring of health checks or the callback. You may also want to mention that if ignoreHealthChecks is true and the request is a health check, the predicate won't even be called. Based on the ordering of the parameters, it might be easy for one to assume that the predicate gets first crack at everything even though that is not the case.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// <returns></returns>
```
evildour: Empty returns

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet/Trace/CloudTrace.cs
```diff
@@ -107,15 +107,14 @@ public sealed class CloudTrace : IDisposable
...
         {
             GaxPreconditions.CheckNotNull(projectId, nameof(projectId));
 
             // Create the default values if not set.
             client = client ?? TraceServiceClient.Create();
-            config = config ?? TraceConfiguration.Create();
-
-            _traceFallbackPredicate = traceFallbackPredicate;
+            config = config ?? TraceConfiguration.Create(); 
+            _traceFallbackPredicate = traceFallbackPredicate ?? TraceDecisionPredicate.Create(); ;
```
evildour: Nit: extra semicolon

jskeet:
I'm going to wait until Mike's comments have been addressed, then I'll take another look. Ping me if you want an earlier opinion :)

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet/Trace/CloudTrace.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/TraceDecisionPredicateTest.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/TraceDecisionPredicateTest.cs
iantalarico: Opps yes. Fixed

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
```diff
@@ -47,13 +47,18 @@ public class TraceDecisionPredicate
         /// <summary>
         /// Creates a <see cref="TraceDecisionPredicate"/>.
         /// </summary>
+        /// <remarks>
+        /// Ignoring of health checks (<paramref name="ignoreHealthChecks"/>) takes precedence over the
+        /// predicate (<paramref name="traceDecisionPredicate"/>).  If <paramref name="ignoreHealthChecks"/>
+        /// is true and the request is a health check (<paramref name="traceDecisionPredicate"/>) will not
```
evildour: Remove parens around this mention of the predicate.

iantalarico:
Done, thanks Mike!

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet/Trace/CloudTrace.cs
```diff
@@ -107,15 +107,14 @@ public sealed class CloudTrace : IDisposable
...
         {
             GaxPreconditions.CheckNotNull(projectId, nameof(projectId));
 
             // Create the default values if not set.
             client = client ?? TraceServiceClient.Create();
-            config = config ?? TraceConfiguration.Create();
-
-            _traceFallbackPredicate = traceFallbackPredicate;
+            config = config ?? TraceConfiguration.Create(); 
+            _traceFallbackPredicate = traceFallbackPredicate ?? TraceDecisionPredicate.Create();
```
jskeet: Given that TraceDecisionPredicate is immutable, make this `TraceDecisionPredicate.Default` as a static property?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
```diff
@@ -0,0 +1,86 @@
...
+using Microsoft.AspNetCore.Http;
+#endif
+
+namespace Google.Cloud.Diagnostics.Common
+{
+    /// <summary>
+    /// Used to determine if an <see cref="HttpRequest"/> should be traced. This will not override
+    /// a decision by the trace header. 
+    /// </summary>
+    public class TraceDecisionPredicate
```
jskeet: Sealed?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        private TraceDecisionPredicate(
+            Func<HttpRequest, bool?> traceDecisionPredicate, bool ignoreHealthChecks)
+        {
+            _traceDecisionPredicate = traceDecisionPredicate;
+            _ignoreHealthChecks = ignoreHealthChecks;
+        }
+
+        /// <summary>
+        /// A <see cref="TraceDecisionPredicate"/> that does not have any affect on
```
jskeet: Nit: affect => effect.
And maybe be specific that health-checks are *not* disabled by this?
And maybe make it a property?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// is true and the request is a health check <paramref name="traceDecisionPredicate"/> will not
+        /// be called.
+        /// </remarks>
+        /// <param name="traceDecisionPredicate">Optional. Used to determine if a request should be traced.
+        ///     Takes an <see cref="HttpRequest"/>.  It should return true if the request should be traced,
+        ///     false if the request should not be traced and null if the decision should be left to other
+        ///     sampling mechanisms. Can be null.</param>
+        /// <param name="ignoreHealthChecks">Optional. True if Google App Engine health check requests should
+        ///     not be traced. Defaults to true.</param>
+        public static TraceDecisionPredicate Create(
```
jskeet: If we have a Default property (which ignores health checks) it would then be reasonable to remove the default value for `traceDecisionPredicate`, and validate that it's *not* null here.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/TraceDecisionPredicateTest.cs
```diff
@@ -0,0 +1,92 @@
...
+            var request = CreateRequest("/_ah/health");
+            Func<HttpRequest, bool?> fallback = (HttpRequest r) => true;
+            var predicate = TraceDecisionPredicate.Create(fallback);
+            Assert.False(predicate.ShouldTrace(request));
+        }
+
+        [Fact]
+        public void ShouldTrace_HealthCheck_Fallback_UseFunc()
+        {
+            var request = CreateRequest("/_ah/somepath");
```
jskeet: Possibly make it a more-different path?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
```diff
@@ -0,0 +1,86 @@
...
+            Func<HttpRequest, bool?> traceDecisionPredicate = null, bool ignoreHealthChecks = true) =>
+                new TraceDecisionPredicate(traceDecisionPredicate, ignoreHealthChecks);
+
+        /// <summary>
+        /// Used to determine if a request should be traced.  This will not override a trace header
+        /// decision.
+        /// </summary>
+        /// <param name="request">The http request to check.</param>
+        /// <returns>True if the request should be traced, false if the request should not be traced
+        ///     and null if the decision should be left to other sampling mechanisms.</returns>
```
jskeet: Nit: remove extra spaces.

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet/Trace/CloudTrace.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/TraceDecisionPredicateTest.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceDecisionPredicate.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -152,16 +136,16 @@ private static TraceHeaderPropagatingHandler CreateTraceHeaderPropagatingHandler
...
         {
             var accessor = provider.GetServiceCheckNotNull<IHttpContextAccessor>();
-            var shouldTraceRequest = provider.GetServiceCheckNotNull<ShouldTraceRequest>();
+            var traceDecisionPredicate = provider.GetServiceCheckNotNull<TraceDecisionPredicate>();
 
             string header = accessor.HttpContext?.Request?.Headers[TraceHeaderContext.TraceHeader];
-            Func<bool?> shouldTraceFunc = () => 
-                shouldTraceRequest?.ShouldTrace?.Invoke(accessor.HttpContext?.Request);
+            Func<bool?> shouldTraceFunc = () =>
+                traceDecisionPredicate?.ShouldTrace(accessor.HttpContext?.Request);
```
jskeet: No need for the first `?` here, given that you're checking it's not null?

iantalarico: Done, thanks!

PR #976, opened 2017-04-10
New Google.Cloud.DevTools.Common package

jskeet:
Currently this just contains a proto for SourceContext;
more code - includng hand-written code - to come.

@Deren-Liao: Not sure why I can't select you as an assignee for this, but...


apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/project.json
```diff
@@ -0,0 +1,32 @@
...
+    "Google.Api.CommonProtos": "1.0.0",
+    "Google.Protobuf": "3.2.0",
+    "Google.Protobuf.Tools": {
+      "version": "3.2.0",
+      "type": "build"
+    }
+  },
+  "frameworks": {
+    "net45": {},
+    "netstandard1.3": {}
```
iantalarico: Aren't we using 1.5 in most places?

jskeet: Yes, because gRPC needs that. This is just protos, at least at the moment. We certainly *could* make it 1.5, but we don't need to. If we ever decided to consume this in something using the REST API, it would be useful to have 1.3, potentially...

iantalarico: No reasons too if it's not needed. I'll default to you here :)

jskeet:
Okay, merging - we can still change our minds about the name until we release it as GA :)

PR #980, opened 2017-04-11
[WIP] Pubsub hi-perf development

chrisdunelm:


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,43 @@
...
+{
+    // class name still to be decided, but can't be Publisher
+    public abstract class HiPerfPublisher
+    {
+        // All entries will use defaults if null
+        public sealed class Settings
+        {
+            Func<Channel> ChannelProvider { get; set; }
+            PublisherSettings PublisherSettings { get; set; }
+            BatchingSettings BatchSettings { get; set; }
```
jskeet: Call both BatchingSettings?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,43 @@
...
+            FlowControlSettings FlowControlSettings { get; set; }
+        }
+
+        public static HiPerfPublisher Create(TopicName topicName, Settings settings = null) =>
+            new HiPerfPublisherImpl(topicName, settings ?? new Settings());
+
+        public virtual TopicName TopicName { get { throw new NotImplementedException(); } }
+
+        public Task<string> Publish(PubsubMessage message)
```
jskeet: PublishAsync? (Even if we're only providing an async interface, it's good to make that clear.)

What's the value returned by the completed task?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,69 @@
...
+            TimeSpan? AckExtensionWindow { get; set; }
+            TaskFactory TaskFactory { get; set; }
+        }
+
+        public static HiPerfSubscriber Create(SubscriptionName subscriptionName, Settings settings = null) =>
+            new HiPerfSubscriberImpl(subscriptionName, settings ?? new Settings());
+
+        public SubscriptionName SubscriptionName { get { throw new NotImplementedException(); } }
+
+        // TODO: Should this accept a CancellationToken?
```
jskeet: Everything should if possible, I think.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,69 @@
...
+        public virtual Task StartAsync(Func<PubsubMessage, Task<AckReply>> handler)
+        {
+            throw new NotImplementedException();
+        }
+
+        public virtual void Start(Func<PubsubMessage, AckReply> handler) =>
```
jskeet: This gives one approach that the client can use - is there anything else we'd expose, such as events or Rx?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
jskeet: Can we add some overloads for different common message bodies, e.g. `byte[]`, `ByteString`, `string`, `IMessage`?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -115,10 +117,11 @@ public sealed class Settings
         /// <see cref="Reply.Ack"/> or <see cref="Reply.Nack"/> the handled messages.
         /// If the task fails or the handler throws an exception, then the message will be
         /// <see cref="Reply.Nack"/>ed.</param>
+        /// <param name="cancellationToken">Used to abort this operation.</param>
         /// <returns>A task which completes once the message receiver is fully operational.</returns>
         /// <exception cref="InvalidOperationException">
         /// Thrown if this <see cref="HiPerfSubscriber"/> has already been started.</exception>
-        public virtual Task StartAsync(Func<PubsubMessage, Task<Reply>> handler)
+        public virtual Task StartAsync(Func<PubsubMessage, Task<Reply>> handler, CancellationToken cancellationToken = default(CancellationToken))
```
jskeet: Presumably if the cancellation token is cancelled, that acts like a Stop request, but the final task will be faulted.

chrisdunelm: Yes, I've added an explanation to the param xmldoc.

jskeet:
Side-note: any thoughts about how Seek and snapshots will fit in here?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Tasks/TaskAwaitable.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                _task = task;
+                _taskScheduler = taskScheduler;
+            }
+
+            private Task _task;
+            private TaskScheduler _taskScheduler;
+
+            public void OnCompleted(Action continuation) => _task.ContinueWith(_ => continuation(), _taskScheduler);
+            public void UnsafeOnCompleted(Action continuation) => _task.ContinueWith(_ => continuation(), _taskScheduler);
```
jskeet: I get nervous seeing code like this not in test code. I can't remember all the details about what OnCompleted vs UnsafeOnCompleted are meant to do, but I suspect they're important. Let's chat about this.

(I'm keen on testability, but ideally we shouldn't have any code like this in the prod package.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/InGax.cs
```diff
@@ -1,15 +1,16 @@
...
 using System.Threading.Tasks;
 
 namespace Google.Api.Gax
 {
     public sealed class BatchingSettings
     {
-        long? ElementCountThreshold { get; set; }
-        long? RequestByteThreshold { get; set; }
-        TimeSpan? DelayThreshold { get; set; }
+        public long? ElementCountThreshold { get; set; }
```
jskeet: Is the intention that these can be changed at any time? There's no element of thread safety / atomicity / volatility here.

chrisdunelm: These are just for passing in settings. All values will be copied out during construction/initialisation.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Tasks/TaskAwaitable.cs
chrisdunelm: This specific nested class is only used for testing, so is almost certainly moveable to a test[ing] package.

Unsafe/non-unsafe is about propagating ExecutionContext (unsafe doesn't need to).
As I understand it, because this class implements ICriticalNotifyCompletion (Via ITaskAwaiter) the OnCompleted() method will never be called. But I could be wrong.
If this is moved out of prod, then only the forwarding awaiter needs to remain in prod, which defines no new behaviour.

I'm fairly strongly of the opinion that the testing that this enables is worth this code. It adds ~zero complexity to the actual code.

jskeet: I'd certainly be a lot happier with as little as possible in prod code, yes.

jskeet: And yes, looking at ForwardingAwaiter, that looks safe to me :)

PR #984, opened 2017-04-11
Add SourceContextFile class that reads source-context.json

Deren-Liao:


jskeet:
Ignore the AppVeyor problem for the moment - looks like I need update how we build gRPC docs. I'll take care of that.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContextFile.cs
```diff
@@ -0,0 +1,111 @@
...
+#if NETSTANDARD1_3
+            string root = AppContext.BaseDirectory;
+#else
+            string root = AppDomain.CurrentDomain.BaseDirectory;
+#endif
+            var fullPath = Path.Combine(root, SourceContextFileName);
+            return File.Exists(fullPath) ? fullPath : null;
+        }
+
+        private static bool IsIOException(Exception ex)
```
jskeet: Nit - you can use an expression-bodied method for this and the next one. No actual need to though.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContextFile.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// <summary>
+        /// Gets the custom log label of Stackdriver Logging entry for Git commit id.
+        /// </summary>
+        public const string GitRevisionIdLogLabel = "git_revision_id";
+
+        /// <summary>
+        /// Gets the Git revision id if it is present. 
+        /// Returns null if there is no Git Repo source context found.
+        /// </summary>
+        public static string GitRevisionId => s_gitRevisionId.Value;
```
jskeet: Rather than exposing a static property just for this part, have you considered making this a partial class with `SourceContext`, and just having a `static SourceContext ApplicationContext => s_sourceContext.Value;` property or similar?

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContextFile.cs
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        /// <summary>
+        /// Find source context file and open the content.
+        /// </summary>
+        private static string Read()
+        {
+            try
+            {
+                using (StreamReader sr = File.OpenText(s_filePath.Value))
```
jskeet: `File.ReadAllText` perhaps?

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContextFile.cs
```diff
@@ -0,0 +1,111 @@
...
+#else
+            string root = AppDomain.CurrentDomain.BaseDirectory;
+#endif
+            var fullPath = Path.Combine(root, SourceContextFileName);
+            return File.Exists(fullPath) ? fullPath : null;
+        }
+
+        private static bool IsIOException(Exception ex)
+        {
+            return ex is FileNotFoundException
```
jskeet: This is effectively just `ex is IOException || ex is UnauthorizedAccessException`, as all the others are subclasses of IOException.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContextFile.cs
```diff
@@ -0,0 +1,111 @@
...
+{
+    /// <summary>
+    /// Utilities that reads source-context.json file from app root.
+    /// </summary>
+    public static class SourceContextFile
+    {
+        private const string SourceContextFileName = "source-context.json";
+        private static Lazy<string> s_filePath = new Lazy<string>(GetFilePath);
+        private static Lazy<SourceContext> s_sourceContext = new Lazy<SourceContext>(OpenParseFile);
+        private static Lazy<string> s_gitRevisionId => new Lazy<string>(() => s_sourceContext.Value?.Git.RevisionId);
```
jskeet: `Git?.RevisionId`, so that if there's a context without a git part, we just get null instead of an exception?

Deren-Liao: Yes, Fixed. I think I did the test. Maybe somehow get reverted.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContextFile.cs
Deren-Liao: Done. I think this makes sense.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContextFile.cs
Deren-Liao: Done.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContextFile.cs
Deren-Liao: Done.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContextFile.cs
Deren-Liao: Done.  After using ReadAllText, there are a few more possible exceptions though.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.Partial.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <summary>
+        /// Gets the custom log label of Stackdriver Logging entry for Git commit id.
+        /// </summary>
+        public const string GitRevisionIdLogLabel = "git_revision_id";
+
+        /// <summary>
+        /// Gets the Git revision id if it is present. 
+        /// Returns null if there is no <seealso cref="GitSourceContext"/> in the protobuf.
+        /// </summary>
+        public string GitRevisionId => Git?.RevisionId;
```
jskeet: Sorry, I wasn't suggesting this - I was suggesting that everything in `SourceContextFile` could actually be in here. I suspect we don't actually need `GitRevisionId` at all - unless it's going to be used by a *lot* of callers, it's pretty easy to just navigate...

Deren-Liao:
I made changes on your feedback.   Please take a look again.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.Partial.cs
Deren-Liao: I see your point.  
I was reluctant to do so. I was thinking about others use the SourceContext class in other scenario.  Adding a "ApplicationSourceContext" could be misleading. It seems like everybody would get the file by calling the static property.
I'll make the change again that in reality, I don't see others will use this class . Or if they do, they get a null , which won't do harm.

Deren-Liao: Done.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
```diff
@@ -0,0 +1,104 @@
...
+                return null;
+            }
+
+            try
+            {
+                return JsonParser.Default.Parse<SourceContext>(sourceContext);
+            }
+            catch (Exception ex) when (IsProtobufParserException(ex))
+            {
+                return null;
```
iantalarico: It seems a little odd to swallow so many exceptions. Given that this is a public library would it be worthwhile to let the caller handle the exceptions?

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
```diff
@@ -0,0 +1,104 @@
...
+    /// The class here reads source-context.json file from app root.
+    /// </summary>
+    public sealed partial class SourceContext
+    {
+        private const string SourceContextFileName = "source-context.json";
+        private static Lazy<string> s_filePath = new Lazy<string>(GetFilePath);
+        private static Lazy<SourceContext> s_sourceContext = new Lazy<SourceContext>(OpenParseFile);
+
+        /// <summary>
+        /// Gets the custom log label of Stackdriver Logging entry for Git commit id.
```
iantalarico: Can you add more context here as to why we have this and that it is temporary?

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
```diff
@@ -0,0 +1,104 @@
...
+#else
+            string root = AppDomain.CurrentDomain.BaseDirectory;
+#endif
+            var fullPath = Path.Combine(root, SourceContextFileName);
+            return File.Exists(fullPath) ? fullPath : null;
+        }
+
+        private static bool IsReadFailureException(Exception ex) =>
+            ex is IOException
+            || ex is ArgumentNullException
```
iantalarico: Is ArgumentNullException a read failure? Should you just check for a null value before?

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        private static string GetFilePath()
+        {
+#if NETSTANDARD1_3
+            string root = AppContext.BaseDirectory;
+#else
+            string root = AppDomain.CurrentDomain.BaseDirectory;
+#endif
+            var fullPath = Path.Combine(root, SourceContextFileName);
+            return File.Exists(fullPath) ? fullPath : null;
```
iantalarico: If it doesn't exist should it throw a file not found exception?

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
Deren-Liao: Done.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
Deren-Liao: Your concern makes sense. Thinking about the possible consequence of not doing so,  Log4Net or user code calls this to write log entry. They are not familiar what exception there might be, or they probably don't even realize it. Then exception happens, user application repeatedly crash for they add a logging method.  That does not make sense to them. 
Adding log normally is not critical path to user, if browsing to source feature does not work, it's fine. But if app crashes repeatedly, that's incident in many cases.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
Deren-Liao: File does not exist is valid scenario.  We don't expect all the app have the source context file.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
Deren-Liao: Done.

Deren-Liao:
Thanks for the feedback. Please look again.

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
iantalarico: I understand the concern but this is a library that other may consume and it swallowing a lot of errors.  How will clients ever get this info if all they get back is null?

Log4Net is a client and IMO should handle this case and maybe log the failure.  If the exceptions are documented I think it should be okay.

iantalarico:
No tests still?

Deren-Liao:
I manually tested the code.  How does other library write tests ? Could you
point me to an example so that I can follow the existing pattern.

iantalarico:
We use XUnit.  Most of the bits in the repo have tests.

For example BigQuery:
https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs

apis/Google.Cloud.DevTools.Common/Google.Cloud.DevTools.Common/Source/V1/SourceContext.File.cs
```diff
@@ -104,9 +103,5 @@ private static string GetFilePath()
             || ex is NotSupportedException
```
iantalarico: Is it okay to swallow these?  I understand the not found bits but SecurityException or UnauthorizedAccessException seem useful to see.

Deren-Liao:
I also opened an issue https://github.com/GoogleCloudPlatform/google-cloud-dotnet/issues/988 to write the tests. I may do this some time next week.

PR #987, opened 2017-04-12
ADO.Net Provider skeleton

benwulfe:
Initial checkin.  For now, minor differences for .net 451 are #if'd (that's what msft does anyhow).  If it gets unmanageable, I'll refactor, but I don't think it will based on my prototype code.


jskeet:
Hey Ben, I'm on vacation until Tuesday - I'll look then.

ILMTitan:
FYI, our current C# code uses K&R style braces (open brace on new line) for functions and classes, (but not for array and object initializes).

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerException.cs
```diff
@@ -0,0 +1,40 @@
...
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class SpannerException : Exception
```
jskeet: Extend DbException, like SqlException does? I'm not sure how much baggage that comes with, admittedly, but it's worth considering.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerParameter.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="spannerColumnName"></param>
+        /// <param name="type"></param>
+        public SpannerParameter(string spannerColumnName, SpannerDbType type) : this(spannerColumnName)
```
jskeet: When you do come to implement these fully, I'd try to arrange it for "smaller" constructors to end up calling one "master" constructor which accepts all possible parameters, rather than the other way round. But I understand this isn't a full implementation yet :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerParameter.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="spannerColumnName"></param>
+        /// <param name="value"></param>
+        public SpannerParameter(string spannerColumnName, object value) : this(spannerColumnName)
```
jskeet: I'm tempted to remove this unless we definitely need it. I'd personally prefer everyone to be very specific about the type of parameter they're providing... but I know that does go against what already exists to some extent.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/TimestampBound.cs
```diff
@@ -0,0 +1,97 @@
...
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class TimestampBound
```
jskeet: Would it make sense for this to be immutable?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/TimestampBoundMode.cs
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public enum TimestampBoundMode
```
jskeet: This is to avoid exposing the proto side of things, presumably?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
```diff
@@ -0,0 +1,80 @@
...
+using System.Threading.Tasks;
+using Google.Api.Gax.Grpc;
+using Google.Apis.Auth.OAuth2;
+using Grpc.Auth;
+
+namespace Google.Cloud.Spanner.V1
+{
+    internal static class ClientPool
+    {
+        static readonly ConcurrentDictionary<ITokenAccess, ClientEntry> ClientEntryPool =
```
jskeet: rename to `s_clientEntryPool` - and make explicitly static.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
```diff
@@ -0,0 +1,80 @@
...
+                new ConcurrentDictionary<ITokenAccess, ClientEntry>();
+
+        public static async Task<SpannerClient> AcquireClientAsync(ITokenAccess credentials)
```
jskeet: We probably need to think about what kind of equality we get from `ITokenAccess` in various cases.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            ClientEntry entry = ClientEntryPool.GetOrAdd(credentials ?? ApplicationDefault.Instance, key => new ClientEntry(key));
+            return await entry.AcquireClientFromEntryAsync();
+        }
+
+        public static int Timeout { get; set; }
```
jskeet: TimeSpan again if possible :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// It is possible to obtain a stale client.
+        /// </summary>
+        /// <returns></returns>
+        public static async Task CloseAllAsync()
+        {
+            await SpannerClient.ShutdownDefaultChannelsAsync();
+            ClientEntryPool.Clear();
+        }
+
+        public class ClientEntry
```
jskeet: Presumably *all* the public members here could be internal - I think it would be good to make them explicitly internal, so readers don't need to check the access of the outer class.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
```diff
@@ -0,0 +1,80 @@
...
+                _credential = credential;
+            }
+
+            public async Task<SpannerClient> AcquireClientFromEntryAsync()
+            {
+                if (_client == null)
+                {
+                    if (_credential != null && _credential != ApplicationDefault.Instance)
+                    {
+                        _client = await SpannerClient.CreateAsync(SpannerClient.DefaultEndpoint,
```
jskeet: This buried reference to `SpannerClient.DefaultEndpoint` is interesting - we should consider what will happen if there's ever a Spanner emulator. It may well affect the design in various places, unfortunately :( (It could reasonably be in the connection string.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
```diff
@@ -0,0 +1,80 @@
...
+                            new SpannerSettings
```
jskeet: This will set the call credentials to the specified ones, but it will still establish the gRPC channel with the application default credentials. Is that definitely what we want to do?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ResourceNames.cs
```diff
@@ -337,7 +337,7 @@ public partial class Session
         /// </summary>
         public SessionName SessionName
         {
-            get { return string.IsNullOrEmpty(Name) ? null : Google.Cloud.Spanner.V1.SessionName.Parse(Name); }
+            get { return String.IsNullOrEmpty(Name) ? null : SessionName.Parse(Name); }
```
jskeet: This code is autogenerated, and shouldn't be touched by anything. (We will need to regenerate regularly.) In this case, my guess is that it was R# being over-enthusiastic...

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Google.Apis.Auth.OAuth2;
+
+namespace Google.Cloud.Spanner.V1
+{
+    public partial class Session : IDisposable
+    {
+        private readonly CancellationTokenSource _cancellation = new CancellationTokenSource();
+        private Task _keepAlive;
```
jskeet: readonly?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
```diff
@@ -0,0 +1,175 @@
...
+        private int _lastAccessTime = Environment.TickCount;
```
jskeet: I suspect the code would be easier to understand using `DateTime` here. Are you using `int` for the sake of atomicity?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
```diff
@@ -0,0 +1,175 @@
...
+        private ITokenAccess _credential;
+        private static int _activeSessions;
+        private static readonly object CreateSync = new object();
```
jskeet: All static fields should be `s_fooBar`.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
```diff
@@ -0,0 +1,175 @@
...
+
+        private async Task KeepAlive()
+        {
+            ExecuteSqlRequest request = new ExecuteSqlRequest
+            {
+                SessionAsSessionName = SessionName,
+                Sql = "SELECT 1",
+            };
+
+            while (true)
```
jskeet: Will properly review this logic later, particularly if the types are going to be changed. Note that if we keep using `TickCount`, we need to carefully handle overflow.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
```diff
@@ -0,0 +1,175 @@
...
+            _lastAccessTime = Environment.TickCount;
+        }
+
+        internal SessionName GetSessionAndMark()
+        {
+            MarkUsed();
+            return SessionName;
+        }
+
+        internal static int MaximumActiveSessions { get; set; } = int.MaxValue;
```
jskeet: I assume all of this will be tied into the SessionPoolOptions later?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
```diff
@@ -0,0 +1,174 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Google.Apis.Auth.OAuth2;
+
+namespace Google.Cloud.Spanner.V1
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    internal class SessionPool : IComparable
```
jskeet: Note to self: I haven't reviewed this yet. Too sleepy to do a good job :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data.Tests/Google.Cloud.Spanner.Data.Tests.xproj
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
benwulfe: I'm not a fan of xproj's -- whats the timeline you think?  next quarter?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class ConnectionPoolOptions
```
benwulfe: I think its ok for these properties to change in realtime.  Thread-wise, it will be safe to set individual options without atomicity. The effects will not be immediate if say, maxsessions is reduced with an outstanding count greater than the new value.   But I think thats ok.   I don't see a reason to put a builder pattern here, but happy to discuss further.  Builder pattern always struck me as more java-esque with lower usability.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+    {
+        /// <summary>
+        /// The default instance of connection pool options.
+        /// </summary>
+        public static readonly ConnectionPoolOptions Instance = new ConnectionPoolOptions();
```
benwulfe: Ack on property.  These are global appdomain wide settings not intended to be per-connection.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+        private ConnectionPoolOptions() {
+            
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public static int GrpcChannelCount
```
benwulfe: I've deleted this setting as I've read more about how the channels are managed -- and this doesn't make sense at a spanner level since it will likely only ever use a single channel.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+        {
+            get { throw new NotImplementedException(); }
+            set { throw new NotImplementedException(); }
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public int KeepAliveIntervalMinutes
```
benwulfe: Ack.  SGTM.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+        public ResourcesExhaustedBehavior ResourcesExhaustedBehavior
+        {
+            get { throw new NotImplementedException(); }
+            set { throw new NotImplementedException(); }
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public static int Timeout
```
benwulfe: It's the overall timeout for a command/request.  internally, we will have settings for backoff time, etc -- but won't expose that to start.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ErrorCode.cs
```diff
@@ -0,0 +1,87 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public enum ErrorCode
```
benwulfe: those classes are used in both grpc and apiary?  If so, then that makes sense.  Is there also an exception class that holds the code so I can get rid of SpannerException?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ResourcesExhaustedBehavior.cs
```diff
@@ -0,0 +1,31 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public enum ResourcesExhaustedBehavior
```
benwulfe: Sure

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerBatchOperation.cs
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class SpannerBatchOperation : IDisposable
```
benwulfe: I've actually already done a pass to make things sealed.  just not pushed yet.  next update.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerBatchOperation.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        private SpannerConnection spannerConnection;
```
benwulfe: yes

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerBatchOperation.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public SpannerBatchOperation(SpannerConnection spannerConnection)
+        {
+            this.spannerConnection = spannerConnection;
+        }
+
+        /// <inheritdoc />
```
benwulfe: ack

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -0,0 +1,216 @@
...
+// ReSharper disable UnusedParameter.Local
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public class SpannerCommand : DbCommand
+#if NET451
+        , ICloneable
```
benwulfe: yes -- its expected for these to be cloneable in "full" net4.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -0,0 +1,216 @@
...
+        public SpannerCommand(string commandText, SpannerConnection connection, SpannerTransaction transaction = null,
+            SpannerParameterCollection parameters = null)
+            : this(SpannerCommandTextBuilder.FromCommandText(commandText), connection, transaction, parameters)
+        {
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public SpannerCommandTextBuilder SpannerCommandTextBuilder
```
benwulfe: internally, we use the builder -- but it doesnt need to be public.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -0,0 +1,216 @@
...
+        protected override DbConnection DbConnection
+        {
+            get { throw new NotImplementedException(); }
+            set { throw new NotImplementedException(); }
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public new SpannerParameterCollection Parameters
```
benwulfe: nope.  fixed.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -0,0 +1,108 @@
...
+        /// <param name="type"></param>
+        /// <param name="databaseTable"></param>
+        private SpannerCommandTextBuilder(OperationType type, string databaseTable)
+        {
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public OperationType OperationType
```
benwulfe: yes, already have that queued for my next push.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -0,0 +1,313 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Google.Apis.Auth.OAuth2;
+using Google.Cloud.Spanner.V1;
+
+// ReSharper disable UnusedParameter.Local
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// A SpannerConnection represents a connection to a single Spanner database.
```
benwulfe: A single grpc channel can support multiple 'sessions' to spanner.
A SpannerConnection is an ADO.Net-only concept and is similar to a session, but it will actually use multiple sessions internally if required.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -0,0 +1,313 @@
...
+    /// </summary>
+    public class SpannerConnection : DbConnection
+    {
+        private SpannerBatchOperation _activeBatchOperation;
```
benwulfe: you *can* actually change the connection string or database -- which is kind of a PITA, but in my currently implementation I require that the connection be closed to simplify that.
so short answer is -- not much can be made readonly here, but as I progress resharper will be screaming at me to make those that can be and I will follow suit.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -0,0 +1,313 @@
...
+
+#if NET451
+        /// <inheritdoc />
+        public override void EnlistTransaction(System.Transactions.Transaction transaction)
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        protected override DbProviderFactory DbProviderFactory
```
benwulfe: actually this PR is a bit old as I've reformatted those that can be expression-bodied.  will update today

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -0,0 +1,313 @@
...
+            OpenAsync(CancellationToken.None).Wait();
+        }
+
+        /// <inheritdoc />
+        public override async Task OpenAsync(CancellationToken cancellationToken) {
+            if (IsOpen) {
+                return;
+            }
+            _client = await ClientPool.AcquireClientAsync(_connectionStringBuilder.Credential);
+            _session =
```
benwulfe: the parallel wait won't actually accomplish anything because the session call needs the client as well, but I agree that's not clear and it wouldnt hurt to wait in parallel as a good practice.
but multiple thread access should be guarded here.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+        public string ClientId
+        {
+            get { return (string)this[nameof(ClientId)]; }
+            private set { this[nameof(ClientId)] = value; }
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public string ClientSecret
```
benwulfe: i don't think he fully understands the options I presented.  while clientid and secret are in the connectionstring, there are no refresh tokens or anything else.  It's purely a shortcut for client side auth if you don't use defaultapp credentials.  I do have a proposal for "fixing" auth because right now its either completely obscure magic (default app credentials), or hard to manage (you need to pass around credentials to every api and ensure it has the right scopes).

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -0,0 +1,161 @@
...
+        public async Task<UserCredential> QueryForCredentials() {
+            if (_credential == null)
+            {
+                var clientId = ClientId;
+                var clientSecret = ClientSecret;
+                if (!string.IsNullOrEmpty(clientId)
+                    && !string.IsNullOrEmpty(clientSecret)) {
+                    return
+                        await
+                            GoogleWebAuthorizationBroker.AuthorizeAsync(
```
benwulfe: no, its not from java -- are we going to ignore client access?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -0,0 +1,77 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public enum SpannerDbType
```
benwulfe: These types come from Google.Cloud.Spanner.V1.TypeCode

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerException.cs
benwulfe: Not worth it IMO, it doesnt add anything and unless other parts of the common infrastructure do something special with it, all its doing is exposing HRESULTs to our api -- which dont apply.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerParameter.cs
benwulfe: +1  its going to be a balance of conforming to some practices that have changed over time, but some people still use.  suggest we can revisit

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/TimestampBound.cs
benwulfe: (y)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/TimestampBoundMode.cs
benwulfe: yes.

benwulfe: it'd feel odd to have randomn classes coming from underneath under a different namespace.  Its one thing if that namespace was "x.y.z.Transactions".

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
benwulfe: default resharper naming conventions for statics are cap cased like this.  I've changed it and will reformat all code.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
benwulfe: I'm assuming reference equality which means they'd need to send in the same credential to get the cached connection.  In the most common cases I think it's ok, but will review later.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
benwulfe: ack

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
benwulfe: excellent idea.  we should bubble this up to the string.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ClientPool.cs
benwulfe: no -- I don't have a solution for this yet as I think I need to create my own channel instead of using the channelpool.  I'll mark it as TODO

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ResourceNames.cs
benwulfe: yes.  I've b-slapped r# and it promised not to do it again.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
benwulfe: yes, working on it.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
benwulfe: not initialized directly in the ctor, so nope...

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
benwulfe: fixed overflow handling.  I think tickcount is easier because the timespan is fairly short and I would otherwise need to manage DST.

benwulfe:
I can keep this in my private fork for a while, but figured it might be better to get earlier reviews.
I could develop in fork/featurebranch -- give you the chance to do some spot checks when I merge to fork/master and then we do a full detailed review in the PR from fork/master to orig/master.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data.Tests/Google.Cloud.Spanner.Data.Tests.xproj
jskeet: I'm hoping it'll be this quarter. I don't expect it to cause much disruption to you though... I'll probably do GAX first, then all of google-cloud-dotnet in one go... but upgrading the Spanner part should be simple enough either before or after we merge it into master.

(If you haven't already got VS2017 installed, plan to do that :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// </summary>
+    public sealed class ConnectionPoolOptions
+    {
+        /// <summary>
+        ///     The default instance of connection pool options.
+        /// </summary>
+        public static readonly ConnectionPoolOptions s_instance = new ConnectionPoolOptions();
```
jskeet: This is still public. You don't need it at all as far as I can see - just use

```csharp
public static ConnectionPoolOptions Instance { get; } = new ConnectionPoolOptions();
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
```diff
@@ -0,0 +1,79 @@
...
+// limitations under the License.
+
+using System;
+using Google.Cloud.Spanner.V1;
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// </summary>
+    public sealed class ConnectionPoolOptions
```
jskeet: Looks like responding to previous comments is hard. If this is a singleton with no expectation of it ever implementing an interface or being passed anywhere etc, we might want to consider using static properties instead - that makes it clearer that this really is simple global state. Not urgent, but we can discuss later.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/ErrorCode.cs
```diff
@@ -0,0 +1,85 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// </summary>
+    public enum ErrorCode
```
jskeet: (Responding to previous comment.) No, the common protos are only used by gRPC-based APIs, but that's fine, as this *is* gRPC-based. While there *are* some common exceptions, I think having a SpannerException makes sense.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            if (Credential == null)
+            {
+                var clientId = ClientId;
+                var clientSecret = ClientSecret;
+                if (!string.IsNullOrEmpty(clientId)
+                    && !string.IsNullOrEmpty(clientSecret))
+                    return
+                        await
+                            GoogleWebAuthorizationBroker.AuthorizeAsync(
```
jskeet: I think if we want to do client-side authentication, we shouldn't assume it would be using `GoogleWebAuthorizationBroker`. But let's talk to the spanner folks about what auth options they actually require.

Personally I'd have expected the `ITokenAccess` specified to be the one to use for the channel credentials, and not to use anything for the `CallCredentials`. If the user wants to do client access, they can do the auth dance however they want to, and then provide that as the `ITokenAccess`. If it proves useful to have separate call credentials, we could potentially support that as a second auth token - but it sounds overly complicated for now.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerException.cs
```diff
@@ -0,0 +1,37 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Spanner
+{
+    /// <summary>
+    /// </summary>
+    public sealed class SpannerException : Exception
```
jskeet: Re DbException - my only concern is if code is trying to be db-agnostic, and catching DbException. I don't know how often that happens in reality, to be honest.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
jskeet: Ah - hadn't noticed this was a partial class mixed with the proto.

I think it would be better to separate the proto from the other behaviour, to be honest. Keeping all protos as pure DTOs makes them easy to reason about. We can't easily "hide" this class, but we could have a separate SpannerSession class perhaps?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Session.cs
jskeet: No, there'd be no need to manage DST - just always use `DateTime.UtcNow` and timespans. I really do think it's easier to reason about time-based code using time-based types, even though the .NET ones are poor. (Admittedly using `DateTime.UtcNow` there's the potential for time changing due to NTP syncs etc, but I doubt that that will be significant.)

chrisdunelm: Agree with @jskeet about using time-based types in all possible places, unless there's a *very* good reason not to. We've consistently done this in all existing code in this repo.
I accept that this is more important when publicly exposed, but I'd much rather see time-based types used internally as well.
If code *must* use ints/longs, then please always put the unit in the name.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -45,59 +52,56 @@ internal SpannerDataReader(ReliableStreamReader resultset, SpannerConnection con
...
+                {
+                    int i = 0;
+                    foreach (var field in metadata.RowType.Fields)
+                    {
+                        _fieldIndex[field.Name] = i;
+                        i++;
+                    }
+                }
+            }
+            return _fieldIndex[fieldName];
```
ILMTitan: This looks like it could be replaced with a `new Lazy()`.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -45,59 +52,56 @@ internal SpannerDataReader(ReliableStreamReader resultset, SpannerConnection con
...
         }
+        /// <inheritdoc />
+        public override object this[string name] => this[GetFieldIndexAsync(name, CancellationToken.None).Result];
 
         /// <inheritdoc />
-        public override int RecordsAffected
-        {
-            get { throw new NotImplementedException(); }
-        }
+        public override int RecordsAffected => -1;
```
ILMTitan: Why is this not `0`. Is `-1` a well known or special flag in ADO.NET?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -109,130 +113,202 @@ public override long GetBytes(int i, long fieldOffset, byte[] buffer, int buffer
...
         {
-            throw new NotImplementedException();
+            return _resultset.GetMetadataAsync(CancellationToken.None).Result.RowType.Fields[i].Name;
         }
 
         /// <inheritdoc />
         public override int GetOrdinal(string name)
         {
-            throw new NotImplementedException();
+            var fields = _resultset.GetMetadataAsync(CancellationToken.None).Result.RowType.Fields;
```
ILMTitan: Do we need to get the metadata every time? The `GetMetadataAsync()` method below makes me think not?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerParameter.cs
```diff
@@ -86,108 +59,96 @@ public SpannerParameter(string spannerColumnName, object value) : this(spannerCo
...
 
         /// <inheritdoc />
         public override ParameterDirection Direction
         {
-            get { throw new NotImplementedException(); }
-            set { throw new NotImplementedException(); }
+            get { return _direction; }
+            set
+            {
+                throw new InvalidOperationException("Spanner does not support anything except input parameters.");
```
ILMTitan: "Spanner supports input parameters only ."

I think starting a sentence with "Spanner does not support anything"  reads poorly.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -109,130 +113,202 @@ public override long GetBytes(int i, long fieldOffset, byte[] buffer, int buffer
...
         }
 
         /// <inheritdoc />
         public override bool Read()
         {
-            throw new NotImplementedException();
+            return ReadAsync(CancellationToken.None).Result;
         }
 
+        private async Task<ResultSetMetadata> GetMetadataAsync(CancellationToken cancellationToken)
```
ILMTitan: This method name makes me think it is not caching.

benwulfe:
I will resubmit this PR after doing more work.

PR #992, opened 2017-04-18
Add example of logging in REST-based APIs to the FAQ

jskeet:
(Maybe we need `LoggingEventType.All` to avoid needing `~None`?)


jskeet:
Great - we'll update the FAQ at that point. I'll look into the build failures now.

jskeet:
(Whoops, my global.json change was idiotic. Fixed; should go green. Will wait until AppVeyor has, as that will prove that the docs build too.)

PR #997, opened 2017-04-19
Update DetectLanguage calls

### LINKED ISSUE #995: DetectLanguages should have an overload accepting multiple strings
I'd assumed multiple values were returned in order to handle multiple possible languages, but it looks like they're meant to correspond to multiple inputs.

This will be a breaking API change, so the snippets will need updating too. When I've made the change, I'll build a new release to make that simpler.

cc @SurferJeffAtGoogle

jskeet:
The API call can handle multiple inputs, each of which can supposedly
return multiple results - but in practice, only a single result is
returned per input.

We now offer two methods (DetectLanguage and DetectLanguages) to
detect single or multiple items. (Along with the async versions, of
course.)

Fixes #995.

(I'm still verifying that it really *will* only return a single detection per input item. I'll wait until I've got that confirmation before merging.)

### DIFF
--- a/apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationClientImpl.cs
+++ b/apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationClientImpl.cs
@@ -132,10 +132,6 @@ namespace Google.Cloud.Translation.V2
         }
 
         /// <inheritdoc />
-        public override Detection DetectLanguage(string text) =>
-            DetectLanguages(new[] { GaxPreconditions.CheckNotNull(text, nameof(text)) })[0];
-
-        /// <inheritdoc />
         public override IList<Detection> DetectLanguages(IEnumerable<string> textItems)
         {
             GaxPreconditions.CheckNotNull(textItems, nameof(textItems));
@@ -200,15 +196,6 @@ namespace Google.Cloud.Translation.V2
             return (await request.ExecuteAsync(cancellationToken).ConfigureAwait(false)).Translations
                 .Zip(items, (result, item) => new TranslationResult(result, item, sourceLanguage, targetLanguage))
                 .ToList();
-        }
-
-        /// <inheritdoc />
-        public override async Task<Detection> DetectLanguageAsync(string text, CancellationToken cancellationToken = default(CancellationToken))
-        {
-            GaxPreconditions.CheckNotNull(text, nameof(text));
-            var listTask = DetectLanguagesAsync(new[] { text }, cancellationToken);
-            var list = await listTask.ConfigureAwait(false);
-            return list[0];
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationClientImpl.cs
```diff
@@ -132,12 +132,19 @@ public override IList<TranslationResult> TranslateHtml(IEnumerable<string> htmlI
         }
 
         /// <inheritdoc />
-        public override IList<Detection> DetectLanguage(string text)
+        public override Detection DetectLanguage(string text) =>
```
chrisdunelm: This implementation could be in the abstract base class, as it's just delegating.

apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationClientImpl.cs
```diff
@@ -196,12 +203,26 @@ public override IList<Language> ListLanguages(string target = null)
         }
 
         /// <inheritdoc />
-        public override async Task<IList<Detection>> DetectLanguageAsync(string text, CancellationToken cancellationToken = default(CancellationToken))
+        public override async Task<Detection> DetectLanguageAsync(string text, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            GaxPreconditions.CheckNotNull(text, nameof(text));
```
chrisdunelm: Why not a one-liner, as in the sync case?
And this is delegating, so could be in the base abstract class.

If keeping like this, maybe at least combine the next two lines. It looks rather odd having the task in its own local!

apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationClientImpl.cs
jskeet: I considered that... the async version is sufficiently implementation-laden that I thought it might be better not to. Happy to change it though.

apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationClientImpl.cs
jskeet: I originally had it as a one-liner before adding validation, but by the time you've got:

- The precondition
- ConfigureAwait(false)
- The cancellation token

... it ended up being huge. Happy to combine the two lines around the task though.

apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2/TranslationClientImpl.cs
chrisdunelm: The async implementation is essentially all boiler-plate, it's not adding any behaviour.
So I think it'd be better in the base abstract class.

jskeet:
Have pushed a second commit - if you could take a look, I'll then squash them before merging. (I'm still going to wait for confirmation that this is appropriate before merging...)

PR #998, opened 2017-04-19
First pass at modified LRO to include metadata

jskeet:
This is a simple approach, but the name of the returned type ends up
being very long.

(Ready for review in terms of general approach; more testing work required.)

### DIFF
--- a/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.cs
+++ b/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.cs
@@ -644,7 +644,7 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         /// <returns>
         /// A Task containing the RPC response.
         /// </returns>
-        public virtual Task<Operation<Database>> CreateDatabaseAsync(
+        public virtual Task<Operation<Database, CreateDatabaseMetadata>> CreateDatabaseAsync(
             InstanceName parent,
             string createStatement,
             CallSettings callSettings = null) => CreateDatabaseAsync(
@@ -680,7 +680,7 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         /// <returns>
         /// A Task containing the RPC response.
         /// </returns>
-        public virtual Task<Operation<Database>> CreateDatabaseAsync(
+        public virtual Task<Operation<Database, CreateDatabaseMetadata>> CreateDatabaseAsync(
             InstanceName parent,
             string createStatement,
             CancellationToken cancellationToken) => CreateDatabaseAsync(
@@ -713,7 +713,7 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         /// <returns>
         /// The RPC response.
         /// </returns>
-        public virtual Operation<Database> CreateDatabase(
+        public virtual Operation<Database, CreateDatabaseMetadata> CreateDatabase(
             InstanceName parent,
             string createStatement,
             CallSettings callSettings = null) => CreateDatabase(
@@ -743,7 +743,7 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         /// <returns>
         /// A Task containing the RPC response.
         /// </returns>
-        public virtual Task<Operation<Database>> CreateDatabaseAsync(
+        public virtual Task<Operation<Database, CreateDatabaseMetadata>> CreateDatabaseAsync(
             CreateDatabaseRequest request,
             CallSettings callSettings = null)
         {
@@ -756,9 +756,9 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         /// <param name="operationName">The name of a previously invoked operation. Must not be <c>null</c> or empty.</param>
         /// <param name="callSettings">If not null, applies overrides to this RPC call.</param>
         /// <returns>A task representing the result of polling the operation.</returns>
-        public virtual Task<Operation<Database>> PollOnceCreateDatabaseAsync(
+        public virtual Task<Operation<Database, CreateDatabaseMetadata>> PollOnceCreateDatabaseAsync(
             string operationName,
-            CallSettings callSettings = null) => Operation<Database>.PollOnceFromNameAsync(
+            CallSettings callSettings = null) => Operation<Database, CreateDatabaseMetadata>.PollOnceFromNameAsync(
                 GaxPreconditions.CheckNotNullOrEmpty(operationName, nameof(operationName)),
                 LongRunningOperationsClient,
                 callSettings);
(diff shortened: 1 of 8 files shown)


apis/Google.LongRunning/Google.LongRunning/Operation.cs
```diff
@@ -28,8 +28,11 @@ namespace Google.LongRunning
     /// For simplicity, no methods on this type modify the proto message. Instead, to get up-to-date
     /// information you can use Refresh to obtain a new instance.
     /// </remarks>
-    /// <typeparam name="T">The response message type.</typeparam>
-    public class Operation<T> where T : IMessage<T>, new()
+    /// <typeparam name="TResponse">The response message type.</typeparam>
+    /// <typeparam name="TMetadata">The metata message type.</typeparam>
+    public sealed class Operation<TResponse, TMetadata>
+        where TResponse : IMessage<TResponse>, new()
+        where TMetadata : class, IMessage<TMetadata>, new()
```
chrisdunelm: Just curious, why is TMetadata constrained to be a class?

apis/Google.LongRunning/Google.LongRunning/Operation.cs
```diff
@@ -130,16 +139,32 @@ public T Result
...
         {
             if (IsCompleted)
             {
                 return this;
             }
             callSettings = Client.GetEffectiveCallSettingsForGetOperation(callSettings);
 
-            Func<DateTime?, Operation<T>> pollAction = deadline => PollOnce(callSettings.WithEarlierDeadline(deadline, Client.Clock));
+            Func<DateTime?, Operation<TResponse, TMetadata>> pollAction =
+                deadline =>
```
chrisdunelm: Could create different `pollAction` function depending on whether `metadataCallback` is `null` or not.
Or if `metadataCallback` is `null` then default it to an empty method.

apis/Google.LongRunning/Google.LongRunning/Operation.cs
```diff
@@ -130,16 +139,32 @@ public T Result
...
+                {
+                    var result = PollOnce(callSettings.WithEarlierDeadline(deadline, Client.Clock));
+                    TMetadata metadata;
+                    if (metadataCallback != null && (metadata = result.Metadata) != null)
```
chrisdunelm: Not desperately keen on this line of code ;)

apis/Google.LongRunning/Google.LongRunning/Operation.cs
jskeet: So that we can return null from the Metadata property if there isn't any metadata.

apis/Google.LongRunning/Google.LongRunning/Operation.cs
jskeet: I wasn't either - it avoids nested if statements though, which also aren't terribly nice. Happy to go that way if you'd prefer though.

apis/Google.LongRunning/Google.LongRunning/Operation.cs
jskeet: Defaulting to an empty method means unpacking (deserializing) the metadata even if we're just going to throw it away, which I'd rather not do. We could generate a different poll action, or convert the `Action<TMetadata>` into an `Action<Operation>` that unpacks the metadata... I think it may just be simplest to make it a nested if statement though.

jskeet: Or add a MaybeCallbackWithMetadata method. That's probably simplest, and I can use it from two places :)

jskeet:
Pushed change for the callback.

apis/Google.LongRunning/Google.LongRunning/Operation.cs
chrisdunelm: Yes, that's definitely better :)

apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.cs
```diff
@@ -1930,12 +1930,12 @@ public DatabaseAdminClientImpl(DatabaseAdmin.DatabaseAdminClient grpcClient, Dat
...
         /// </returns>
-        public override async Task<Operation<Empty>> UpdateDatabaseDdlAsync(
+        public override async Task<Operation<Empty, UpdateDatabaseDdlMetadata>> UpdateDatabaseDdlAsync(
             UpdateDatabaseDdlRequest request,
             CallSettings callSettings = null)
         {
             Modify_UpdateDatabaseDdlRequest(ref request, ref callSettings);
-            return new Operation<Empty>(
+            return new Operation<Empty, UpdateDatabaseDdlMetadata>(
                 await _callUpdateDatabaseDdl.Async(request, callSettings), LongRunningOperationsClient);
```
evildour: `ConfigureAwait(false)`? There are 3 other places like this as well.

jskeet: Well spotted - this is in the code generator, so I've filed https://github.com/googleapis/toolkit/issues/1199

jskeet:
@evildour: Any general thoughts about this approach? Other alternatives would involve generating a new type for each long-running call, so we'd have `LongRunningRecognizeOperation` - there are various options for exactly how to do that, but they're all quite involved.

evildour:
That was my first thought when skimming this, but I could go either way. It's kind of borderline-ugly at this point and since `var` is an option in many cases, it might not matter.

Would each new type contain an `Operation<...>` or derive from it? I would lean toward the latter so  callers would not need to do `instance.Operation.Whatever`. But given that the class needs to create new instances of itself, `Operation<...>` would need abstract `Create...` methods or it would have to be parameterized with a 3rd type argument: the derived type, which always seems a bit hacky to me.

jskeet:
@evildour: That was basically my thought process, experimenting for a while. I started off with composition - generally favouring composition over inheritance - then tried inheritance, and in every case it got really complicated.

The downsides of the proposal are only:

- It avoids long return types (which we already have for pagination, mind you)
- If we knew that some services didn't support cancellation or deletion, we could not generate those methods
- If we ever wanted to add extra methods to the generated code, we could do so

I'm leaning more and more to this being the solution.

evildour:
You mean you're leaning towards inheritance being the solution? I guess abstract/virtual creation methods are the cleanest way to do this if that's the way to go.

jskeet:
No, I'm leaning to this PR being the solution. I tried both composition and inheritance, and both were messy. We could easily end up with a type system which is fine so long users _only_ use Intellisense - but which is really confusing (for very little benefit) as soon as they start looking at the documentation.

evildour:
I see, got a little confused there. Ok, this make sense. Yeah, I didn't consider the documentation, but that's a good point.

jskeet:
This is now our plan of record, modulo objections.

PR #1000, opened 2017-04-20
Overhaul of the Translation client

jskeet:
- Validation of results (n inputs => n outputs, 1 detection per input)
- Tests for argument validation
- Integration tests using a large input to check that's okay
  (I was worried we might be putting everything in the URL)
- Implement single-element translations in TranslationClient via
  delegation


apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2.Tests/CodeHealthTest.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright 2016 Google Inc. All Rights Reserved.
```
chrisdunelm: 2017

jskeet: Darn it, I thought I was so careful :)

PR #1002, opened 2017-04-21
Update how translation models are handled

### LINKED ISSUE #1001: Update docs around the translation model
The NMT model is now the default for translation, so do we need a parameter at all?

jskeet:
- We now have an explicit "let the server decide" enum value. This
  is better than overloading the use of null, as otherwise if the
  client specified a particular model, it couldn't be overridden
  by a single method call.
- Documentation is updated to describe the models in more detail,
  including when you might want to explicitly use PBMT.

Fixes #1001. (More extensive than expected...)

### DIFF
--- a/apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2.Snippets/TranslationClientSnippets.cs
+++ b/apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2.Snippets/TranslationClientSnippets.cs
@@ -22,21 +22,21 @@ namespace Google.Cloud.Translation.V2.Snippets
     public class TranslationClientSnippets
     {
         
-        public void TranslateTextNmtDefaultModel()
+        public void TranslateTextPbmtDefaultModel()
         {
-            // Sample: TranslateTextNmtDefaultModel
-            TranslationClient client = TranslationClient.Create(model: TranslationModel.NeuralMachineTranslation);
+            // Sample: TranslateTextPbmtDefaultModel
+            TranslationClient client = TranslationClient.Create(model: TranslationModel.PhraseBasedMachineTranslation);
             TranslationResult result = client.TranslateText("It is raining.", LanguageCodes.French);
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2.Snippets/TranslationClientSnippets.cs
```diff
@@ -22,21 +22,21 @@ namespace Google.Cloud.Translation.V2.Snippets
     public class TranslationClientSnippets
     {
         
-        public void TranslateTextNmtDefaultModel()
+        public void TranslateTextPbmtDefaultModel()
         {
             // Sample: TranslateTextNmtDefaultModel
```
chrisdunelm: Sample: TranslateTextPbmtDefaultModel?

apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2.Snippets/TranslationClientSnippets.cs
```diff
@@ -22,21 +22,21 @@ namespace Google.Cloud.Translation.V2.Snippets
...
+            TranslationClient client = TranslationClient.Create(model: TranslationModel.PhraseBasedMachineTranslation);
             TranslationResult result = client.TranslateText("It is raining.", LanguageCodes.French);
             Console.WriteLine($"Result: {result.TranslatedText}; detected language {result.DetectedSourceLanguage}");
             // End sample
         }
 
-        public void TranslateTextNmtOverrideModel()
+        public void TranslateTextPbmtOverrideModel()
         {
             // Sample: TranslateTextNmtOverrideModel
```
chrisdunelm: Sample: TranslateTextPbmtOverrideModel?

jskeet: Yes indeed, thank you! Hopefully AppVeyor would have caught that - I'm going to let it go red just to prove that, then fix it, push and wait for it to go green.

PR #1007, opened 2017-04-22
Add git revision id

Deren-Liao:
to log entries if the source context file exists and contains git repo
information.

### DIFF
--- a/apis/Google.Cloud.Logging.V2/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
+++ b/apis/Google.Cloud.Logging.V2/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
@@ -15,8 +15,10 @@
 using Google.Api;
 using Google.Api.Gax;
 using Google.Api.Gax.Grpc;
+using Google.Cloud.DevTools.Source.V1;
 using Google.Cloud.Logging.Type;
 using Google.Cloud.Logging.V2;
+using Google.Protobuf;
 using log4net.Appender;
 using log4net.Core;
 using System;
@@ -269,6 +271,7 @@ namespace Google.Cloud.Logging.Log4Net
             {
                 labels.Add(nameof(MetaDataType.Level), loggingEvent.Level?.Name ?? unknown);
             }
+            TryAddGitRevisionId(labels);
             foreach (var customLabel in _customLabels)
             {
                 labels.Add(customLabel.Key, customLabel.Value);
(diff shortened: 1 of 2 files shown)


Deren-Liao:
Hi there,

As I clicked the sign url, it told me I am a *Googlers do not need to sign
a CLA.*

jskeet:
I suspect the reason is that the commit has Charlie-ET as the author, instead of the Deren-Liao user. I suspect you just need to amend the email address used for that commit - and probably change your defaults for your local clone too.

Deren-Liao:
Yes, that's the reason. I did some test with another account.

Deren-Liao:
I fixed the CLA problem.  Some tests failed.  I don't think it's related to my change.  How may I let the test try again?

jskeet:
Hmm... the BigQuery tests appear to have hung. Have restarted it.

apis/Google.Cloud.Logging.V2/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -289,6 +292,23 @@ private LogEntry BuildLogEntry(LoggingEvent loggingEvent)
...
+        private void TryAddGitRevisionId(Dictionary<string, string> labels)
+        {
+            string gitId = null;
+            try
+            {
+                gitId = SourceContext.AppSourceContext?.Git?.RevisionId;
+            }
+            catch (Exception ex) when (ex is System.Security.SecurityException || ex is InvalidProtocolBufferException
+                || ex is InvalidJsonException || ex is UnauthorizedAccessException)
+            { }
```
jskeet: I'd reformat this as one type per line for readability. And could we at least have a comment in the catch body explaining why we're not doing anything?

apis/Google.Cloud.Logging.V2/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -289,6 +292,23 @@ private LogEntry BuildLogEntry(LoggingEvent loggingEvent)
...
+
+            if (!String.IsNullOrWhiteSpace(gitId))
```
jskeet: I'd move this inside the catch block - then you don't need the variable to be declared first. The Add call won't throw any of the caught exceptions, so it won't change behaviour.

apis/Google.Cloud.Logging.V2/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -289,6 +292,23 @@ private LogEntry BuildLogEntry(LoggingEvent loggingEvent)
...
+            {
+                labels.Add(SourceContext.GitRevisionIdLogLabel, gitId);
```
jskeet: I think I missed where `GitRevisionIdLogLabel` was added to `SourceContext`. Is that definitely the right place for it? It feels like a logging concern more than one of `SourceContext` itself...

apis/Google.Cloud.Logging.V2/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
Deren-Liao: Done.

apis/Google.Cloud.Logging.V2/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
Deren-Liao: Done.

apis/Google.Cloud.Logging.V2/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
Deren-Liao: My perception is that It's a log label for source context. It's related to both.  I feel it's acceptable to put it in either here or in the Google.Cloud.Logging.V2.dll.

Deren-Liao:
Thank you Jon for all the help and input.

PR #1008, opened 2017-04-23
Add git revision id

Deren-Liao:
Add git revision id to log entries if the source context file exists and contains git repo information.


Deren-Liao:
Test to see if I can fix CLA problem.  Close.

PR #1015, opened 2017-04-27
Change how external dependencies are handled in docs

jskeet:
This allows us to use a mixture of project.json and csproj projects.

The metadata for external dependencies is now built when the
"external" directory is first created. (That can be recreated by
calling fetchexternal.) The result is an external/api directory
containing all the metadata, broken down by package. The
GenerateDocfxSources code detects the dependencies, and then the
builddocs.sh script copies it into place alongside the other library
metadata, and concatenates the TOC files.

It would have been nice to be able to just refer to the external
dependency metadata from the docfx.json files, but if the files
start in different directories that ends up being tricky.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@ bin
 project.lock.json
 _site
 site
+dependencies
+# Legacy docs/external; now docs/dependencies
 external
 output
 *~
--- a/docs/builddocs.sh
+++ b/docs/builddocs.sh
@@ -2,19 +2,6 @@
 
 set -e
 
-fetch() {
-  # $1: repo (and directory name)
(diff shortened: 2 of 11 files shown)


jskeet:
(Appveyor may well fail to start with - will get on that.)

jskeet:
Urgh. Looks like it's hard to get appveyor to support project.json and csproj at the same time.

Will look into having a separate orphan branch for "dependency-apis" which can be built locally - will speed up the build as well. Suggest reviewing the first commit, but none of the later ones...

jskeet:
This is now completely ready for review, with the idea of the orphan branch already implemented (and the branch is already present)

PR #1024, opened 2017-05-04
Attempt to make BigQuery integration/snippet tests less flaky.

jskeet:
Unified approach to waiting until newly-inserted rows are visible.


jskeet:
Have had a suggestion that checking for a change in last-modified date may work too...

PR #1027, opened 2017-05-04
Fix JSON handling for strings that look like dates

### LINKED ISSUE #1016: DateTimes stored as strings
In our bigquery tables, we store datetime as a timestamp, but we also store the datetime as the original string provided by the ultimate data source.

When attempting to retrieve the string form of the date, an illegal cast exception is thrown:
`   at Google.Cloud.BigQuery.V2.BigQueryRow.ConvertSingleValue(Object rawValue, TableFieldSchema field)
   ...`
 on a statement like:
`    var value = row["dateTimeRaw"];`
It appears that the value has already been internally converted to a DateTime (though it should not have been) and that an attempt is being made to cast that to a string and failing.

jskeet:
This will fix #1016.

Currently the test fails, because I haven't actually fixed it... but once the next release of the REST support libraries has happened, we'll be able to specify our own serializer settings to fix it.


chrisdunelm:
I'll delay reviewing this until the REST library release is properly sorted out.

jskeet:
(Fixed now - ready for review.)

PR #1029, opened 2017-05-08
Changes to Translation models

### LINKED ISSUE #1018: Update model documentation again
We should refer to the "base" model in generic terms, rather than it explicitly being PBMT.

jskeet:
The meaning of "base" may change over time. This PR reverts to using
Base as the model name, and makes the documentation more circumspect
about what it means.

Fixes #1018.


jskeet:
(Will wait for feedback from the Translation API team before merging.)

PR #1031, opened 2017-05-08
Restore nesting for class parts in project files

jskeet:


jskeet:
Nope, it's just that it's not nested by default. It wasn't in old-school csproj either - looks like project.json-based tooling did it automatically.

https://github.com/dotnet/project-system/issues/159 is tracking this in general.

PR #1032, opened 2017-05-08
Porting Log4Net to netstandard 1.5

Jetski5822:
For review. There is one thing left to do around AllAssemblies()


Jetski5822:
I signed it!

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Snippets/Google.Cloud.Logging.Log4Net.Snippets.csproj
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net452</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.1;net452</TargetFrameworks>
```
chrisdunelm: Please target `netcoreapp1.0` to match other projects.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Google.Cloud.Logging.Log4Net.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net452</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.1;net452</TargetFrameworks>
```
chrisdunelm: And `netcoreapp1.0` here too.

jskeet:
(Assigned Chris as I know little about this code...)

Jetski5822:
Darn, failing mutli-threaded test!!

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/LruCacheTest.cs
```diff
@@ -128,7 +128,7 @@ public void MultithreadedUse()
             }
             for (int i = 0; i < threadCount; i++)
             {
-                Assert.True(threads[i].Join(TimeSpan.FromSeconds(5)));
+                Assert.True(threads[i].Join(TimeSpan.FromSeconds(5).Milliseconds));
```
chrisdunelm: Please remove `.Milliseconds`. Using `TimeSpan`s directly is preferable where possible.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.csproj
```diff
@@ -1,8 +1,8 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Version>1.0.0-beta10</Version>
-    <TargetFrameworks>net45</TargetFrameworks>
+    <Version>1.0.0-beta11</Version>
```
chrisdunelm: Please leave at version `beta10`. We'll up-version just before release.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.csproj
```diff
@@ -21,7 +21,12 @@
   <ItemGroup>
     <ProjectReference Include="..\..\Google.Cloud.DevTools.Common\Google.Cloud.DevTools.Common\Google.Cloud.DevTools.Common.csproj" />
     <ProjectReference Include="..\..\Google.Cloud.Logging.V2\Google.Cloud.Logging.V2\Google.Cloud.Logging.V2.csproj" />
-    <PackageReference Include="log4net" Version="2.0.5" />
+    <PackageReference Include="log4net" Version="2.0.8" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.5'">
+    <PackageReference Include="Microsoft.Extensions.DependencyModel">
```
chrisdunelm: Why is this required? Please remove if not required.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -1,502 +1,507 @@
...
+            }
+            string function0 = null;
+            string fullTypeName = loggingEvent.LocationInformation?.ClassName;
+            if (fullTypeName != null)
+            {
+                var type = _typeCache.GetOrAdd(fullTypeName, () =>
+                {
+                    try
+                    {
+#if NET45
```
chrisdunelm: Please use:
```
#if NET45
...
#elif NETSTANDARD1_5
...
#else
#error Platform not supported.
#endif
```

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -1,502 +1,507 @@
...
+                         return AppDomain.CurrentDomain.GetAssemblies()
+                             .SelectMany(a => a.GetTypes())
+                             .FirstOrDefault(t => t.FullName == fullTypeName);
+#endif
+#if NETSTANDARD1_5
+
+                        return null;
```
chrisdunelm: Please add `// TODO: Support type lookup in netstandard`

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/LruCacheTest.cs
Jetski5822: Join doesnt take a TimeSpan in netstandard 1.5, only an integer.

chrisdunelm: Oh, yes, you're right. Fair enough ;)

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
Jetski5822: done

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.csproj
Jetski5822: done

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.csproj
Jetski5822: done

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Google.Cloud.Logging.Log4Net.Tests.csproj
Jetski5822: done

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Snippets/Google.Cloud.Logging.Log4Net.Snippets.csproj
Jetski5822: done

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.csproj
Jetski5822: Seems we do need this as we are targeting netstandard 1.5, this could go if we were using 1.6 I think

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -1,502 +1,507 @@
...
+            {
+                file = loggingEvent.LocationInformation?.FileName;
+            }
+            if (long.TryParse(loggingEvent.LocationInformation?.LineNumber, out long lineNumber))
+            {
+                line = lineNumber;
+            }
+            string function0 = null;
+            string fullTypeName = loggingEvent.LocationInformation?.ClassName;
+#if NET45
```
chrisdunelm: Sorry, I wasn't clear.
Please put this `#if` back to where you had it before, within the `try` block; and use `#elif`, so the contents of the `try` block should look like this:
```
try
{
#if NET45
    return AppDomain.CurrentDomain.GetAssemblies()
        .SelectMany(a => a.GetTypes())
        .FirstOrDefault(t => t.FullName == fullTypeName);
#elif NETSTANDARD1_5
    // TODO: Support type lookup in netstandard
    return null;
#else
#error Unsupported platform.
#endif
}
catch
...
```
This puts the `#if` around the least amount of code.
And explicitly generates a compile-time error if we try change targeted platforms in any way, meaning we won't end up with any nasty surprises.
Thanks.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.csproj
chrisdunelm: Exactly what is required from `Microsoft.Extensions.DependencyModel`? If that `PackageReference` is removed, what doesn't build?

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
Jetski5822: done :)

Jetski5822:
So the only thing I am not understanding is why the test is failing, do you have any ideas?

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Snippets/GoogleStackdriverAppenderSnippets.cs
```diff
@@ -59,7 +60,7 @@ public void Overview()
                 // Resource: log4net-template.xml log4net_template
                 // Sample: Overview
                 // Configure log4net to use Google Stackdriver logging from the XML configuration file.
-                XmlConfigurator.Configure(new FileInfo("log4net.xml"));
+                XmlConfigurator.Configure(LogManager.GetRepository(Assembly.GetEntryAssembly()), new FileInfo("log4net.xml"));
```
chrisdunelm: `Assembly.GetEntryAssembly()` can return null on some platforms. Please use `GetType().GetTypeInfo().Assembly`

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Snippets/GoogleStackdriverAppenderSnippets.cs
```diff
@@ -83,7 +84,7 @@ public void Overview_AspNet()
             // Resource: log4net-aspnet-template.xml log4net_aspnet_template
             // Sample: Overview_AspNet
             // Load log4net configuration from Web.config
-            log4net.Config.XmlConfigurator.Configure();
+            log4net.Config.XmlConfigurator.Configure(LogManager.GetRepository(Assembly.GetEntryAssembly()));
```
chrisdunelm: And `GetType().GetTypeInfo().Assembly` here too please.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
```diff
@@ -38,7 +39,7 @@ public class Log4NetTest
         [MethodImpl(MethodImplOptions.NoInlining)]
         private void LogInfo(IEnumerable<string> messages)
         {
-            ILog log = LogManager.GetLogger("testlogger");
+            ILog log = LogManager.GetLogger(Assembly.GetEntryAssembly(), "testlogger");
```
chrisdunelm: `Assembly.GetEntryAssembly()` returns null on netcore, causing tests to fail. Please use `typeof(Log4NetTest).GetTypeInfo().Assembly`.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
```diff
@@ -98,7 +99,7 @@ private class NoDelayScheduler : IScheduler
             Hierarchy hierarchy = null;
             if (requiresLog4Net)
             {
-                hierarchy = (Hierarchy)LogManager.GetRepository();
+                hierarchy = (Hierarchy)LogManager.GetRepository(Assembly.GetEntryAssembly());
```
chrisdunelm: `Assembly.GetEntryAssembly()` returns null on netcore, causing tests to fail. Please use `typeof(Log4NetTest).GetTypeInfo().Assembly`.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/LruCacheTest.cs
chrisdunelm: This should be `.Join((int)TimeSpan.FromSeconds(5).TotalMilliseconds)`.
`.Milliseconds` returns just the millisecond part, which is 0. This was the cause of the test failure.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -1,502 +1,508 @@
...
+        }
+
+        private LogEntrySourceLocation BuildLogEntryLocation(LoggingEvent loggingEvent)
+        {
+            string file = null;
+            long? line = null;
+            string function = null;
+            if (loggingEvent.LocationInformation?.FileName != null)
+            {
+                file = loggingEvent.LocationInformation?.FileName;
```
chrisdunelm: It appears the latest version of log4net now has a filename of "?", rather than empty/null, when it doesn't know the filename.
Please insert this code:
``` C#
if (file == "?")
{
    file = null;
}
```
following this assignment of `file`

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -1,502 +1,508 @@
...
+                        // Returning null caches the failure.
+                        return null;
+                    }
+                });
+                if (type != null)
+                {
+                    function0 = type.AssemblyQualifiedName;
+                }
+            }
+            string function1 = loggingEvent.LocationInformation?.MethodName;
```
chrisdunelm: In the same way to `file`, `MethodName` also now contains "?" rather than being empty/null.
Please insert this code:
``` C#
if (function1 == "?")
{
    function1 = null;
}
```
following the assignment of `function1`.

chrisdunelm:
Sorry, one more change is required in `Log4NetTest.cs`:
In the method `SingleLogEntryWithLocation()`, the last three `Assert`s need surrounding with an `if` statement, and the line numbers in the `entry0.SourceLocation.Line` checks adjusting if required:
``` C#
if (entry0.SourceLocation != null)
{
    ... The 3 existing Asserts, with possibly modified line numbers in the second Assert ...
}
```

You can run these unit tests locally by going to the `Google.Cloud.Logging.Log4Net.Tests` directory and running:
`dotnet test -f net452` for net45 tests; and
`dotnet test -f netcoreapp1.0` for net core tests

Hope this is all clear, please let me know if not.
Thanks :)

chrisdunelm:
So close...

The final change required (and this is why appeyor is failing), is to update the apis/apis.json file with the newly supported target frameworks.
Lines 165 and 166 (within the `"id": "Google.Cloud.Logging.Log4Net"` section) need updating to include the new frameworks.
Line 165: `"targetFrameworks": "netstandard1.5;net45"`
Line 166: `"testTargetFrameworks": "netcoreapp1.0;net452"`

Once that change has been made, run ./generateprojects.sh (I use a git-bash console to do this on Windows) and then look at the git diff
The only change that I found when I did this was one line-ending change. Commit this, and now appveyor should pass.

Sorry that this has been a somewhat long and drawn-out change. Thanks for your persistence.

apis/apis.json
```diff
@@ -161,8 +161,8 @@
     "id": "Google.Cloud.Logging.Log4Net",
     "version": "1.0.0-beta11",
     "type": "other",
-    "targetFrameworks": "net45",
-    "testTargetFrameworks": "net452",
+    "targetFrameworks": "netstandard1.5;net45"
+    "testTargetFrameworks": "netcoreapp1.0;net452"
```
jskeet: Not that it needs a change now (I don't want to cause more churn than is required :) but these are now the defaults - so we can remove these lines entirely. I'll do that myself in a later commit.

Jetski5822: I can revert this?

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/LruCacheTest.cs
Jetski5822: @jskeet the current code is..

> Assert.True(threads[i].Join((int)TimeSpan.FromSeconds(5).TotalMilliseconds));

Which comment are you replying too?

apis/apis.json
jskeet: No, you either need the lines as per your commit, or change the commit to delete the lines entirely. Basically, the previously-present lines were there to explicitly state targets which *differed* from the default generated by the tool - the new targets *are* the default, so they can either be stated explicitly or not stated at all.

It's probably easiest just to leave it as it is for now though - I can remove it later one with no changes elsewhere.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/LruCacheTest.cs
jskeet: I was reacting to the code in the first commit, which had `TimeSpan.FromSeconds(5).Milliseconds`. If it's using `TotalMilliseconds` now, that's fine, at least for test code. (If we need this in multiple places, I'd consider using an extension method...)

Jetski5822: Okay cool, I thought as much - just wanted to be sure.

apis/apis.json
Jetski5822: Okay, I will leave.

chrisdunelm:
@Jetski5822 What exactly is weird? I saw just the one appveyor test failure, due to the line number change in Log4NetTest.cs
The line number change is due to the addition of the `using System.Reflection` in line 32.
So the 44L and 42L in line 253 now need to be incremented by one to 45L and 43L.

jskeet:
We should be able to avoid hard-coding those line numbers using caller line info. I'm happy to do that in a separate PR though.

Jetski5822:
@chrisdunelm its weird, as I thought these files were originally without the BOM, but now they have a BOM. But maybe I am wrong.

jskeet:
@Jetski5822: It's possibly because this PR was started before #1035.

Jetski5822:
@jskeet that would be it.... I thought I was going crazy.!!

Jetski5822:
GREEN!!! Whoop

chrisdunelm:
@Jetski5822 Thanks :)

jskeet:
My mistake on the target frameworks part - there's no default when the API type is "other". So that doesn't need doing at all.

PR #1034, opened 2017-05-09
Labels API for Storage buckets - no implementation yet

jskeet:
FAO @houglum


houglum:
Looks like a good interface for modifying labels. (I'm assuming there will be <i>Get/Fetch</i> and <i>Clear</i>BucketLabels as well in subsequent commits)

jskeet:
ClearBucketLabels sounds like a good idea. I don't think there's any need for a Get operation, as it's already literally as simple as:

```csharp
var labels = client.GetBucket(bucketName).Labels;
```

jskeet:
Closing this PR as this should go onto a feature branch.

PR #1039, opened 2017-05-12
Implementation and integration tests for Storage labels

jskeet:
Last piece will be the snippets.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.ModifyBucketLabels.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.ModifyBucketLabels.cs
@@ -12,8 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
+using Google.Api.Gax;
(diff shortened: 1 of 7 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.ModifyBucketLabels.cs
```diff
@@ -24,38 +27,151 @@ public sealed partial class StorageClientImpl
...
-
-            // In theory, many label modifications could be performed with just a single call rather than
-            // read-modify-write, but:
-            // - Always doing a read-modify-write simplifies the code
-            // - This way we get to return the previous values as well
-            throw new NotImplementedException();
+            ValidateBucketName(bucket);
+            GaxPreconditions.CheckNotNull(labels, nameof(labels));
+
+            int retries = options?.Retries ?? ModifyBucketLabelsOptions.DefaultRetries;
```
jskeet: I wondered about that. Elsewhere we validate things in a `ModifyRequest`, but we don't have anything like that. I'll check that it's non-negative within the property, but put no upper bound on it... a huge upper bound may not be *sensible*, but it at least has a valid meaning...

PR #1048, opened 2017-05-16
Regenerate all APIs

jskeet:
(Self-assigning, as this is just a regen.)


jskeet:
Validated. This is just:

- Some extra proto fields
- Doc updates
- A couple of `ConfigureAwait(false)` calls now generated by codegen
- Fixed metadata uses now generated by codegen (I used the wrong metadata type before by hand)
- Lots of spaces in type argument lists...

PR #1053, opened 2017-05-17
Set debug symbols and the debug type for asp.net core error reporting tests

iantalarico:
The docs for DebugSymbols says that this determines if the symbols are generated or not*.  With and without this the symbols are generated.  With this set to true it looks like they are used.

* I could only find the docs for VS 2015. See: https://msdn.microsoft.com/en-us/library/bb629394.aspx.

With this changes the tests do pass.


jskeet:
This will fail CI, because you're editing a generated property group. Put it in a new property group instead - only the first property group and the first item group are replaced by the script.

iantalarico:
Done, PTAL

jskeet:
I certainly have no objections to that now - which isn't quite the same as saying I think it'll definitely work :)

iantalarico:
WCPGW? I'll let the CI finished and see what happens :)

PR #1063, opened 2017-05-22
Update all versions

jskeet:
- Update dependencies to GAX 2.0.0-beta01 and Google.Api.CommonProtos 1.1.0
- Add imports for ScopedCredentialProvider (which has moved in GAX v2)
- Update the version number of every package. GA packages have a major version bump and are now beta; everything else just bumps up a beta version.


jskeet:
(Will self-merge on green.)

PR #1067, opened 2017-05-24
Make Log4Net line number test more robust

jskeet:


apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
```diff
@@ -29,24 +29,16 @@
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
-using Xunit;
-
+using Xunit;
+
```
jskeet: This diff is almost certainly just a CRLF vs LF difference. The file was inconsistent before. It's now all-Unix. We may want to change it to all-Windows at some point...

jskeet:
Okay, this doesn't seem to get rid of the AppVeyor part. Closing...

PR #1070, opened 2017-05-24
Update dependencies

jskeet:
- gRPC now uses csproj
- google-api-dotnet-client now uses csproj
- We now use csproj, so for protobuf we need to create a global.json

(I've already pushed the result to the dependencies branch.)

### DIFF
--- a/docs/fetchdependencies.sh
+++ b/docs/fetchdependencies.sh
@@ -12,8 +12,8 @@ git -C dependencies checkout README.md .gitignore
 
 echo "Cloning repositories"
 git clone https://github.com/googleapis/gax-dotnet dependencies/gax-dotnet --quiet --depth=1 -b master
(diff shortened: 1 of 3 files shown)


docs/fetchdependencies.sh
```diff
@@ -31,9 +31,9 @@ cp dependencies-docfx/docfx-google-api-dotnet-client.json dependencies/google-ap
 # TODO: The last of these fails to find System.DateTime in Storage. No idea why,
 # but chances are it'll be fixed when google-api-dotnet-client updates to csproj...
```
chrisdunelm: Is this comment still true?

jskeet: Hmm. Not sure. Will check in the next couple of days.

jskeet: Looks like it's gone now. Edited the TODO out.

PR #1075, opened 2017-05-25
Increase timeout in LruCacheTest

chrisdunelm:


jskeet:
LGTM if it's consistently green :)

chrisdunelm:
That test takes ~2minutes on travis, which is  strangely slow. But it passes :)

PR #1085, opened 2017-05-30
Initial implementation of requester-pays feature for Storage

jskeet:
It's not as bad as it looks - the vast majority of this code is just repeated for each request.


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RequesterPaysTest.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        [Fact]
+        public void UploadObject_WithOption()
+        {
+            if (_skip)
+            {
+                return;
+            }
+
```
jskeet: Thanks, will fix.

PR #1089, opened 2017-06-05
Use a custom property for snippets in docfx

### LINKED ISSUE #1055: Use custom heading for doc snippets
At the moment I suspect the samples are overwriting any Remarks sections in the XML docs. We should use a custom model property and then customize the template appropriately.

jskeet:
Fixes #1055.

### DIFF
--- a/docs/docfx-template/partials/class.tmpl.partial
+++ b/docs/docfx-template/partials/class.tmpl.partial
@@ -1,0 +1,238 @@
+{{!Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE file in the project root for full license information.}}
+
+{{>partials/class.header}}
+{{#children}}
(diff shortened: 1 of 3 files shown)


docs/docfx-template/partials/class.tmpl.partial
```diff
@@ -0,0 +1,238 @@
+{{!Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE file in the project root for full license information.}}
```
chrisdunelm: I assume this file is just copied from MS somewhere?
Should it still have the MS copyright?

jskeet: Good question. I *think* it's probably appropriate to keep it. (It doesn't come out in the generated docs, btw.)

PR #1099, opened 2017-06-06
Cloud spanner testperf

benwulfe:
adds test logging for unit tests and on demand perf log output.
also a minor optimization for choosing an already prewarmed session.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
@@ -25,14 +25,16 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
 {
     public class BindingTests : IClassFixture<TestDatabaseFixture>
     {
+        // ReSharper disable once UnusedParameter.Local
         public BindingTests(TestDatabaseFixture testFixture, ITestOutputHelper outputHelper)
         {
             _testFixture = testFixture;
-            // Uncomment the lines below to enable detailed logging
-            //            SpannerConnection.ConnectionPoolOptions.LogLevel = LogLevel.Debug;
-            //            SpannerConnection.ConnectionPoolOptions.LogPerformanceTraces = true;
-            //            SpannerConnection.ConnectionPoolOptions.PerformanceTraceLogInterval = 1000;
+#if LoggingOn
+            SpannerConnection.ConnectionPoolOptions.LogLevel = LogLevel.Debug;
+            SpannerConnection.ConnectionPoolOptions.LogPerformanceTraces = true;
+            SpannerConnection.ConnectionPoolOptions.PerformanceTraceLogInterval = 1000;
             TestLogger.TestOutputHelper = outputHelper;
+#endif
         }
 
         private readonly TestDatabaseFixture _testFixture;
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Threading.Tasks;
 using Xunit;
+using Xunit.Abstractions;
 
 #endregion
 
@@ -24,7 +25,15 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
 {
     public class BindingTests : IClassFixture<TestDatabaseFixture>
     {
-        public BindingTests(TestDatabaseFixture testFixture) => _testFixture = testFixture;
+        public BindingTests(TestDatabaseFixture testFixture, ITestOutputHelper outputHelper)
+        {
+            _testFixture = testFixture;
+            // Uncomment the lines below to enable detailed logging
+            //            SpannerConnection.ConnectionPoolOptions.LogLevel = LogLevel.Debug;
+            //            SpannerConnection.ConnectionPoolOptions.LogPerformanceTraces = true;
+            //            SpannerConnection.ConnectionPoolOptions.PerformanceTraceLogInterval = 1000;
+            TestLogger.TestOutputHelper = outputHelper;
+        }
 
         private readonly TestDatabaseFixture _testFixture;
 
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
@@ -39,7 +39,8 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         // not cause failures).
         // if you use a scratch database, be sure to comment out the database
         // creation and disposal methods.
-        public string DatabaseName { get; } = "t_" + Guid.NewGuid().ToString("N").Substring(0, 28);
+        public string DatabaseName { get; } = "scratch";
+//            "t_" + Guid.NewGuid().ToString("N").Substring(0, 28);
 
         public string TestTable { get; } = "TestTable";
 
@@ -47,8 +48,8 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
 
         public void Dispose()
         {
-            _databaseAdminClient?.DropDatabase(
-                new DatabaseName(TestProjectName, TestInstanceName, DatabaseName));
+//            _databaseAdminClient?.DropDatabase(
+//                new DatabaseName(TestProjectName, TestInstanceName, DatabaseName));
         }
 
         public async Task<SpannerConnection> GetTestDatabaseConnectionAsync()
@@ -61,12 +62,14 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
 
         private async Task EnsureTestDatabaseImplAsync()
         {
-            await CreateDatabaseAsync();
-            await Task.WhenAll(
-                CreateTableAsync(),
-                CreateTypeTableAsync(),
-                CreateTxTableAsync()).ConfigureAwait(false);
-            await FillSampleData();
+            TestLogger.Install();
+            await Task.Delay(1); // this line exists so the others can be commented.
+//            await CreateDatabaseAsync();
+//            await Task.WhenAll(
+//                CreateTableAsync(),
+//                CreateTypeTableAsync(),
+//                CreateTxTableAsync()).ConfigureAwait(false);
+//            await FillSampleData();
         }
 
         private async Task<Database> CreateDatabaseAsync()
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
@@ -39,8 +39,8 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         // not cause failures).
         // if you use a scratch database, be sure to comment out the database
         // creation and disposal methods.
-        public string DatabaseName { get; } = "scratch";
-//            "t_" + Guid.NewGuid().ToString("N").Substring(0, 28);
+        public string DatabaseName { get; } = //"scratch";
+            "t_" + Guid.NewGuid().ToString("N").Substring(0, 28);
 
         public string TestTable { get; } = "TestTable";
 
@@ -48,8 +48,8 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
 
         public void Dispose()
         {
-//            _databaseAdminClient?.DropDatabase(
-//                new DatabaseName(TestProjectName, TestInstanceName, DatabaseName));
+            _databaseAdminClient?.DropDatabase(
+                new DatabaseName(TestProjectName, TestInstanceName, DatabaseName));
         }
 
         public async Task<SpannerConnection> GetTestDatabaseConnectionAsync()
@@ -64,12 +64,12 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         {
             TestLogger.Install();
             await Task.Delay(1); // this line exists so the others can be commented.
-//            await CreateDatabaseAsync();
-//            await Task.WhenAll(
-//                CreateTableAsync(),
-//                CreateTypeTableAsync(),
-//                CreateTxTableAsync()).ConfigureAwait(false);
-//            await FillSampleData();
+            await CreateDatabaseAsync();
+            await Task.WhenAll(
+                CreateTableAsync(),
+                CreateTypeTableAsync(),
+                CreateTxTableAsync()).ConfigureAwait(false);
+            await FillSampleData();
         }
 
         private async Task<Database> CreateDatabaseAsync()
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestLogger.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestLogger.cs
@@ -1,4 +1,18 @@
-﻿using System;
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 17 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
```diff
@@ -159,6 +172,7 @@ private class HistoryEntry
         public async Task AbortedThrownCorrectly()
         {
             await WriteSampleRowsAsync();
+            LogPerformanceDataBeforeTest();
```
evildour: It may be cleaner to have a disposable helper object which logs performance data in both its constructor and `Dispose` implementation. Then you can wrap all the tests in something like:
```csharp
using (new PerformanceLogHelper())
{
   ...
}
```

This would also prevent missing 'after' logging when returning early in one of the logged regions (although that seems unlikely for a test).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
```diff
@@ -304,12 +322,14 @@ private class HistoryEntry
...
             }
+            LogPerformanceDataAfterTest();
         }
 
         [Fact]
         public async Task ReadExactSingle()
         {
             await WriteSampleRowsAsync();
+            LogPerformanceDataBeforeTest();
             await Task.Delay(6);
```
evildour: Should this delay be logged as part of the test? Seems like the 'before' logging should be done after this delay.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
```diff
@@ -59,8 +59,11 @@ public LogLevel LogLevel
                 {
                     return (LogLevel) underlyingLevel;
                 }
-                // ReSharper disable once NotResolvedInText
-                throw new ArgumentOutOfRangeException("LogLevel");
+                else
+                {
+                    // ReSharper disable once NotResolvedInText
+                    throw new ArgumentOutOfRangeException("LogLevel");
+                }
```
evildour: I don't think you should throw from a getter, but if it is necessary, then you might want to use `GaxPreconditions.CheckEnumValue` if possible.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
```diff
@@ -69,8 +72,11 @@ public LogLevel LogLevel
                 {
                     Logger.LogLevel = (V1.Logging.LogLevel) underlyingLevel;
                 }
-                // ReSharper disable once NotResolvedInText
-                throw new ArgumentOutOfRangeException("LogLevel");
+                else
+                {
+                    // ReSharper disable once NotResolvedInText
+                    throw new ArgumentOutOfRangeException("LogLevel");
+                }
```
evildour: As mentioned above, use `GaxPreconditions.CheckEnumValue` if available.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/TransactionPool.cs
```diff
@@ -115,6 +115,16 @@ public static TransactionOptions GetLastUsedTransactionOptions(this Session sess
...
+        public static bool IsPreWarmedTransactionReady(this Session session)
+        {
+            SessionInfo info;
+            if (s_sessionInfoTable.TryGetValue(session, out info))
+            {
+                return info.HasActiveTransaction;
+            }
+            return false;
+        }
+
```
evildour: With C# 7 this can now be an expression-bodied method:
```csharp
... => s_sessionInfoTable.TryGetValue(session, out var info) && info.HasActiveTransaction;
```
But I don't feel too strongly about it.

jskeet:
Hmm. I'm generally slightly uncomfortable with mixing performance testing and correctness testing - it feels like they could easily come into conflict, and that performance testing typically requires multiple iterations etc to get realistic numbers. I think it would be entirely reasonable to have a separate project for PerformanceTests, covering much the same ground as this but without using the correctness testing (which can impact results too).

benwulfe:
This wasnt intended to be 'the perf tests'.  I saw what I thought was a
perf regression and added that logging to provide me more information -
that's all.

Without these changes, you cannot see any logs (diagnostic or perf) on any
tests, which makes debugging more challenging than it needs to be.

jskeet:
If they're not meant to be particularly measuring perf, could we hook into XUnit to perform diagnostics before and after each test automatically? I think we want to avoid having a start/end call on every single test.

benwulfe:
XUnit has no before/after methods for each test because it doesn't believe
in that design.

jskeet:
Hmm. I suspect there may be a workaround. Are you intending that every integration test has this pair? It really feels like cruft that's going to get in the way significantly :(

benwulfe:
Ahhh, there is BeforeAfterTestAttribute.  The original blog about xunit
said something about this being an antipattern and it looks like they
changed their minds.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -17,14 +17,23 @@
...
     public class BindingTests : IClassFixture<TestDatabaseFixture>
     {
-        public BindingTests(TestDatabaseFixture testFixture) => _testFixture = testFixture;
+        public BindingTests(TestDatabaseFixture testFixture, ITestOutputHelper outputHelper)
+        {
+            _testFixture = testFixture;
+            // Uncomment the lines below to enable detailed logging
+            //            SpannerConnection.ConnectionPoolOptions.LogLevel = LogLevel.Debug;
+            //            SpannerConnection.ConnectionPoolOptions.LogPerformanceTraces = true;
+            //            SpannerConnection.ConnectionPoolOptions.PerformanceTraceLogInterval = 1000;
```
benwulfe: I like it!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
```diff
@@ -39,7 +39,8 @@ public class TestDatabaseFixture : IDisposable
         // not cause failures).
         // if you use a scratch database, be sure to comment out the database
         // creation and disposal methods.
-        public string DatabaseName { get; } = "t_" + Guid.NewGuid().ToString("N").Substring(0, 28);
+        public string DatabaseName { get; } = //"scratch";
```
benwulfe: yes, I prefer to leave it in there because I am always using it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
```diff
@@ -61,6 +62,8 @@ public void Dispose()
 
         private async Task EnsureTestDatabaseImplAsync()
         {
+            TestLogger.Install();
+            await Task.Delay(1); // this line exists so the others can be commented.
```
benwulfe: when I use my scratch database, I usually comment out all the creation bits and dispose.
1 millisecond isn't a big deal, but sure.

benwulfe: its internal...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestLogger.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using Google.Cloud.Spanner.V1.Logging;
+using Xunit.Abstractions;
+
+namespace Google.Cloud.Spanner.Data.IntegrationTests
+{
+    internal class TestLogger : DefaultLogger
+    {
+        private static Lazy<TestLogger> s_Instance = new Lazy<TestLogger>(
```
benwulfe: no reason to create it if it isnt getting installed.

benwulfe: yep, I was just debugging something.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestLogger.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+        }
+
+        public static void Install()
+        {
+            SetLogger(s_Instance.Value);
+        }
+
+        // ReSharper disable once MemberCanBePrivate.Global
+        // ReSharper disable once UnusedAutoPropertyAccessor.Global
```
benwulfe: its only used within the class, so resharper (rightly) tells me to make it private.  But I prefer to keep it public for now and simply want a green light from resharper.  It's not terribly important.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
benwulfe: Switching to use the beforeafter attribute in xunit -- this should clean things up

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
benwulfe: true re: logging.  but I'll just need to account for built in delays in the tests because I am moving the perf logging to be in attributes, so will also include the sample writing.
I'm not looking for regressions on the order of 6ms anyway.  more like ~2 seconds, which would indicate a shared session wasn't used (and a new one was created).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
benwulfe: fair enough re: throwing in general during get.  though this change is just adding some braces to ensure (facepalm) you can actually get the value at all without throwing.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/TransactionPool.cs
benwulfe: overdoing expression bodies has so far caused more work than its worth.  for cases where its probable to add additional logic, I prefer to leave it as statement syntax

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
evildour: Is it possible to have a boolean flag in this class that switches between using/not using a scratch DB? It looks like people need to remember to also comment out other code if using a scratch DB, so using the boolean there to prevent errors in the future would be better too IMO. Seems like it would also remove the need for the dummy await below.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestLogger.cs
evildour: FYI, you can still add a breakpoint with my proposed line by placing the cursor anywhere in the object creation expression and pressing F9 (or whatever your breakpoint shortcut in VS is).

evildour: I believe the singleton will only get created when the class is referenced the first time. Also, since it has no instance fields, it only takes up the heap metadata header's worth of space. It doesn't seem like the savings from avoiding a creation are worth the extra code, but that's just me (actually, the `Lazy<T>` instance itself almost certainly take up more space since it is also a reference type). If you disagree that's fine.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestLogger.cs
evildour: Ok, that makes sense, although I would argue that since the setter is public, the getter should be too. I don't think it should be private. Is the "Unused..." disable necessary?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ConnectionPoolOptions.cs
evildour: It looks like this change is a no-op actually. The throw would only be reached if the if block's condition were false in both old and new versions. Same with the setter.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestLogger.cs
benwulfe: you cannot predict when static constructors run.  they often run when the JIT touches the type for the first time.  relying on that behavior is always bad news.

evildour: I'm definitely being pedantic now, so apologies, but I believe you actually can predict when they will run (see https://msdn.microsoft.com/en-us/library/aa645758(VS.71).aspx). It is static initializers in the absence of a static constructor which cannot be predicted, but since that is the case we're in here anyway, that's probably what you meant.

However, I still don't see the point. The constructor has no side-effects and as I said, the instance uses almost no memory, so at some implementation-defined point, you either (1) allocate the `Lazy<T>` with three reference fields in it which allows delaying the creation of a `TestLogger` or (2) you allocate a `TestLogger` with no fields. Once again, apologies for being pedantic. Feel free to ignore me.

benwulfe:
I've seen static initializers get run as soon as a type is "touched" in a
method -- ie the JIT sees a method using type FOO in some way and then JITs
it and runs the static initializer (ctor or not) whether or not an instance
of FOO gets created or a method is called.

The page you refer to is talking about the order of static field
initializers which is related, but different in that everything that page
talks about is happening after the runtime decides its time to run static
initializers.

But it isn't always very obvious like it is in that example.  You can have
a type reference other types and indirectly cause a JIT/initialization of
another type and alot of this behavior (AFAIK) isn't specified by the cli
ecma or any other runtime documentation  (so it may be slightly different
on say mono).

jskeet:
If a type with a static constructor is initialized before it is actually accessed, that's a bug. Static constructors provide a much stronger guarantee than other general type initialization, which can be remarkably lazy and remarkably eager. I don't think I've ever seen a type with a static constructor get over-eagerly initialized.

benwulfe:
I'm fairly certain I've seen a static ctor get called when a method got
JIT'd -- but the type was never actually "accessed".  It was just the
possibility of it being accessed that caused the initialization.

I'm open to seeing proof to the contrary.

jskeet:
It's hard for me to prove you've never seen it, but I'll happily look up the specs showing you *shouldn't* have seen it :) (In the morning...) If you happen to remember an example, I'd be really interested to see it. (Runtime bugs do happen, of course.)

benwulfe:
You must be correct as I cannot repro it.  Its still not important enough
 to switch -- but I stand corrected.

jskeet:
For any interested parties ('cos I'm interested if no-one else is :) the relevant spec sections are:

C# 5 specification, 10.12:

> The static constructor for a closed class type executes at most once in a given application domain. The execution of a static constructor is triggered by the first of the following events to occur within an application domain:
> 
> - An instance of the class type is created.
> - Any of the static members of the class type are referenced.

(There's a bit more for structs in 11.3.10.)

10.5.5.1 (static field initialization) mentions the difference between a type that includes a static constructor or not:

>  If a static constructor (§10.12) exists in the class, execution of the static field initializers occurs immediately prior to executing that static constructor. Otherwise, the static field initializers are executed at an implementation-dependent time prior to the first use of a static field of that class. 

The C# compiler implements that using `beforefieldinit` which is specified for types *without* a static constructor, but not for types with one. That leads to ECMA-335 section I.8.9.5:

> If not marked BeforeFieldInit then that type’s initializer method is executed at (i.e.,
is triggered by):
>
> - a. first access to any static field of that type, or
> - b. first invocation of any static method of that type, or
> - c. first invocation of any instance or virtual method of that type if it is a value
type or
> - d. first invocation of any constructor for that type.

PR #1100, opened 2017-06-07
Cloud spanner ddl

benwulfe:
Implements DDL as well as the proper resource format (projects/x/instances/y/databases/z)
To do a create database command, you can use a datasource (in your connectionstring) without specifying a database.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
@@ -92,7 +92,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
             catch (SpannerException e)
             {
                 exceptionCaught = true;
-                Debug.WriteLine("TestBadColumnName: Caught error code:" + e.ErrorCode);
+                Debug.WriteLine($"TestBadColumnName: Caught error code:{e.ErrorCode}");
                 Assert.Equal(ErrorCode.InvalidArgument, e.ErrorCode);
                 Assert.False(e.IsTransientSpannerFault());
             }
@@ -104,8 +104,8 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         [Fact]
         public async Task BadDbName()
         {
-            string connectionString = "Data Source=projects/" + _testFixture.TestProjectName + "/instances/"
-                + _testFixture.TestInstanceName + "/databases/" + "badjuju";
+            string connectionString = $"Data Source=projects/{_testFixture.TestProjectName}/instances/"
+                + $"{_testFixture.TestInstanceName}/databases/badjuju";
             // ReSharper disable once RedundantAssignment
             int rowsRead = -1;
             var exceptionCaught = false;
@@ -115,7 +115,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
                 using (var connection = new SpannerConnection(connectionString))
                 {
                     var cmd = connection.CreateSelectCommand(
-                        "SELECT * FROM " + _testFixture.TestTable + " WHERE Key = 'k1'");
+                        $"SELECT * FROM {_testFixture.TestTable} WHERE Key = 'k1'");
                     using (var reader = await cmd.ExecuteReaderAsync())
                     {
                         rowsRead = 0;
@@ -181,7 +181,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
                 using (var connection = await _testFixture.GetTestDatabaseConnectionAsync())
                 {
                     var cmd = connection.CreateSelectCommand(
-                        "SELECT * FROM " + _testFixture.TestTable);
+                        $"SELECT * FROM {_testFixture.TestTable}");
 
                     using (var reader = await cmd.ExecuteReaderAsync())
                     {
@@ -245,8 +245,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
             using (var connection = await _testFixture.GetTestDatabaseConnectionAsync())
             {
                 var cmd = connection.CreateSelectCommand(
-                    "SELECT * FROM " + _testFixture.TestTable
-                    + " WHERE Key = 'k1'");
+                    $"SELECT * FROM {_testFixture.TestTable} WHERE Key = 'k1'");
                 using (var reader = await cmd.ExecuteReaderAsync())
                 {
                     rowsRead = 0;
@@ -270,8 +269,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
             using (var connection = await _testFixture.GetTestDatabaseConnectionAsync())
             {
                 var cmd = connection.CreateSelectCommand(
-                    "SELECT * FROM " + _testFixture.TestTable
-                    + " WHERE Key = 'k99'");
+                    $"SELECT * FROM {_testFixture.TestTable} WHERE Key = 'k99'");
                 using (var reader = await cmd.ExecuteReaderAsync())
                 {
                     rowsRead = 0;
@@ -303,7 +301,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
                 using (var connection = await _testFixture.GetTestDatabaseConnectionAsync())
                 {
                     var cmd = connection.CreateSelectCommand(
-                        "SLECT * FROM " + _testFixture.TestTable);
+                        $"SLECT * FROM {_testFixture.TestTable}");
                     using (var reader = await cmd.ExecuteReaderAsync())
                     {
                         rowsRead = 0;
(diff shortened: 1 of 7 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
```diff
@@ -104,8 +104,8 @@ public class ReadTests : IClassFixture<TestDatabaseFixture>
         [Fact]
         public async Task BadDbName()
         {
-            string connectionString = "Data Source=" + _testFixture.TestProjectName + "/"
-                + _testFixture.TestInstanceName + "/" + "badjuju";
+            string connectionString = "Data Source=projects/" + _testFixture.TestProjectName + "/instances/"
```
jskeet: Possibly use an interpolated string literal? I think that would be easier to understand. You could potentially suck in local variables first:

```csharp
string project = _testFixture.TestProjectName;
string instance = _testFixture.TestInstanceName;

string connectionString = $"Data Source=projects/{project}/instances/{instance}/databases/badjuju";
```

(Ditto for other changes using concatenation.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
```diff
@@ -47,7 +50,10 @@ public class TestDatabaseFixture : IDisposable
 
         public void Dispose()
         {
-            _databaseAdminClient?.DropDatabase(
+            //TODO(benwu):DROP Database is not supported in DDL form yet.
+            var databaseAdminClient = DatabaseAdminClient.CreateAsync().Result;
```
jskeet: Use just `DatabaseAdminClient.Create()` rather than async and Result.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
```diff
@@ -47,7 +50,10 @@ public class TestDatabaseFixture : IDisposable
...
+
+            databaseAdminClient?.DropDatabase(
```
jskeet: Why the `?.`? You'd never end up with a null reference here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
```diff
@@ -69,53 +75,32 @@ public void Dispose()
...
-
-            // Poll until the returned long-running operation is complete
-            var completedResponse = await response.PollUntilCompletedAsync();
-
-            // Retrieve the operation result
-            return completedResponse.Result;
+            using (var connection = new SpannerConnection(NoDbConnectionString))
+            {
+                var createCmd = connection.CreateDdlCommand("CREATE DATABASE " + DatabaseName);
+                await createCmd.ExecuteNonQueryAsync(CancellationToken.None);
```
jskeet: You shouldn't need to specify CancellationToken.None here - the DbCommand implementation will pass that to the SpannerCommand implementation, won't it?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TestDatabaseFixture.cs
```diff
@@ -69,53 +75,32 @@ public void Dispose()
...
-                    new DatabaseName(
-                        TestProjectName,
-                        TestInstanceName,
-                        DatabaseName),
-                    new[] {typeTable});
-
-            // Poll until the returned long-running operation is complete
-            await response.PollUntilCompletedAsync();
+
+            using (var connection = new SpannerConnection(ConnectionString))
```
jskeet: Given the repetition of this block, maybe extract to a method of `ExecuteDdlAsync(string ddl)`?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -222,6 +230,40 @@ public override int ExecuteNonQuery() => ExecuteNonQueryAsync(_synchronousCancel
                     "You can only call ExecuteNonQueryAsync on a Delete, Insert, InsertUpdate, or Update Command");
             }
 
+            if (SpannerCommandTextBuilder.SpannerCommandType == SpannerCommandType.Ddl)
```
jskeet: Conditional expression instead? Personal preference, definitely.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -222,6 +230,40 @@ public override int ExecuteNonQuery() => ExecuteNonQueryAsync(_synchronousCancel
...
+            }
+            else
+            {
+                return ExecuteMutationsAsync(cancellationToken);
+            }
+        }
+
+        private async Task<int> ExecuteDdlAsync(CancellationToken cancellationToken)
+        {
+            var databaseAdminClient = await DatabaseAdminClient.CreateAsync();
```
jskeet: ConfigureAwait (and below).
We should sync up on when we're going to merge all this into master so we can get the automated checking in...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -72,16 +76,36 @@ public SpannerCommandTextBuilder(string commandText)
...
+            // These three form the ddl for spanner.
+            // For reference: https://cloud.google.com/spanner/docs/data-definition-language
+            return string.Equals(commandPart, CreateCommand, StringComparison.OrdinalIgnoreCase)
+                || string.Equals(commandPart, AlterCommand, StringComparison.OrdinalIgnoreCase)
+                || string.Equals(commandPart, DropCommand, StringComparison.OrdinalIgnoreCase);
+        }
+
+        /// <summary>
+        /// </summary>
+        public bool IsCreateDatabaseCommand => CommandText?.ToUpper().StartsWith(CreateDatabaseCommand) ?? false;
```
jskeet: A couple of problems here:

- Culture-sensitive upper-casing; use `string.StartsWith(string, StringComparison)`
- Whitespace: shouldn't "CREATE   DATABASE" with multiple spaces between the words work as well?

Probably fiddly to fix - I'm fine if you want to just raise an issue for now so we can come back to it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -35,7 +36,33 @@ public sealed class SpannerConnectionStringBuilder : DbConnectionStringBuilder
         public string DataSource
         {
             get => GetValueOrDefault("Data Source");
-            private set => this["Data Source"] = value;
+            private set => this["Data Source"] = ValidatedDataSource(value);
+        }
+
+        private string ValidatedDataSource(string dataSource)
```
jskeet: This should basically be a DatabaseName (as in the resource), right? If so, we already have parsing code in there.

There's a difference though - the "projects", "instances" and "databases" part would be case-sensitive. Do we definitely want to make that non-case-sensitive? It's tricky given that we want to be like GCP in some ways and like ADO.NET in others... I strongly suspect that the actual *ID* parts are going to be treated case-sensitively anyway.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
benwulfe: I'll do it right after checking in these two.  Then I just have my in progress doccomments which (you would hope) wont have merge conflicts.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
benwulfe: Yeah, I had thought about the whitespace issue.  The grammar definition for spanner ddl doesn't seem to indicate variable whitespace here, but you never know -- I'll see what I can do -- it is fiddly.  fixed string issue

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
benwulfe: dont think case sensitivity is that important.
hmm good point re:using templates -- except for the case where you do CREATE DATABASE and do not have a database.
i was aiming to allow you to only specify the first two...  Let me see what I can do here.

PR #1102, opened 2017-06-08
Fix all occurences of "TimeStamp" to "Timestamp"

jskeet:
(Checked by searching for meSt in a case-sensitive fashion.)


benwulfe:
LGTM

Issue #1003, opened 2017-04-21
Revisit LRO metadata callbacks


jskeet:
(Tagged as speech as that will be our first LRO-based API to go GA)

If PR #998 is merged, the behaviour of metadata callbacks in PollUntilCompleted needs to be reviewed and better documented. In particular:

- Do we want to make the callback with a value of null if there isn't any metadata? That could be useful to clients in terms of acknowledging that a poll has completed
- Do we want to make the callback if the operation is already complete when PollUntilCompleted is called?

jskeet:
Okay, suggested approach:

- Always call the callback if we have one, even if the metadata is null
- Call the callback if the operation is complete, for consistency
- Document that the callback may be called in any thread. (We can't guarantee it's the initial thread, and in order to guarantee that it's *not* the initial thread, we'd have to use Task.Run, which is pretty ugly.)
- Document that the callback will be called synchronously (next poll will be blocked on the callback completing, and we'll never have multiple callbacks executing at the same time)

@chrisdunelm: Do all of those sound correct to you? If so, I'll implement this weekend.

PR #1107, opened 2017-06-12
Implementation of metadata callback changes

jskeet:
This is to demonstrate the changes proposed in #1003.
Tests will be added if we decide this is the behavior we want.

### DIFF
--- a/apis/Google.LongRunning/Google.LongRunning/Operation.cs
+++ b/apis/Google.LongRunning/Google.LongRunning/Operation.cs
@@ -133,29 +133,23 @@ namespace Google.LongRunning
         }
 
         /// <summary>
-        /// Executes the given callback with the metadata in the RPC message
-        /// if both the callback and the metadata are non-null.
-        /// </summary>
-        private void MaybeFireCallback(Action<TMetadata> metadataCallback)
-        {
-            if (metadataCallback != null && RpcMessage.Metadata != null)
-            {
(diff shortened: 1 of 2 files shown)


jskeet:
@chrisdunelm: Ping for comments...

apis/Google.LongRunning/Google.LongRunning/Operation.cs
```diff
@@ -202,14 +204,15 @@ private void MaybeFireCallback(Action<TMetadata> metadataCallback)
         {
```
chrisdunelm: `Func<TMetadata, Task>` for this async case? Which would be awaited on when invoked.
If a user is writing async-based code, it may be frustrating that this callback is not Task-based.

jskeet: Hmm. Not sure. Maybe stick with the sync version and add an overload later if we get requests for that? Note that we won't have a synchronization context (it could be called from any thread) so at least some of the normal benefits of async aren't present anyway...

chrisdunelm: OK, sgtm

jskeet:
Great. Will write tests and then get those reviewed before merging. Thanks!

jskeet:
Okay, tests now added accordingly (in the single commit). No change to prod code since earlier version. PTAL.

PR #1108, opened 2017-06-12
Allow user to specify failure mode for diagnostic libraries

### LINKED ISSUE #842: Allow user to specify failure mode for diagnostic libraries
This is about the Google.Cloud.Diagnostics.* libraries rather than the underlying API packages, so only blocks the release of those higher level libraries.

If calls to Stackdriver APIs fail, what should we do? Currently, we just let the exception propagate:

- If that happens synchronously within a request, the request will fail. (Not sure when this would occur, but it could be interesting if Error Reporting middleware fails...)
- If that happens asynchronously due to log buffering, it will bring down the process

In *some* cases that may be what the user wants. Other options the user may want:

- Buffer for some time and retry periodically, failing if the diagnostic failure continues for a long time.
- Buffer for some time but then start throwing away log entries and other requests to avoid failing
- Don't even bother buffering - make diagnostics "best effort, but don't waste memory"

iantalarico:
This will allow users to select how they wish to handle failures when making calls to diagnostic libraries.

The default will not retry and swallow exceptions so the diagnostics bits do not crash or interfere with the application (this is the current logic). 

Fixes #842.

Unit/Integration test will be in a follow up PR.
Breaking up the buffer into smaller pieces if too large will be in a follow up PR.


jskeet:
Note: I'm away this week, so won't be able to get to this until next Monday, probably.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ConsumerFactoryTest.cs
```diff
@@ -23,55 +23,61 @@ public class ConsumerFactoryTest
...
 
         [Fact]
         public void GetConsumer_InvalidOptions()
         {
-            BufferOptions options = new BufferOptions((BufferType)5);
+            BufferOptions bufferOptions = new BufferOptions((BufferType)5);
+            RetryOptions retryOptions = RetryOptions.NoRetry();
             Assert.Throws<ArgumentException>(
-                () => ConsumerFactory<int>.GetConsumer(new IntConsumer(), null, options));
+                () => ConsumerFactory<int>.GetConsumer(new IntConsumer(), ConstantSizer<int>.GetSize, bufferOptions, RetryOptions.NoRetry()));
```
evildour: Use retryOptions here or remove the declaration above

iantalarico:
@jskeet No worries/rush.  I figured this may sit for a bit I'm OOO the second half of June.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ISequentialThreadingTimer.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System;
+
+namespace Google.Cloud.Diagnostics.Common
+{
+    /// <summary>
+    /// A timer that will automatically call a callback after the previous call has finished and an
+    /// additional time interval has passed.
+    /// </summary>
+    interface ISequentialThreadingTimer : IDisposable
```
evildour: Add explicit `internal`

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+        None,
+
+        /// <summary>Requests will be retried.</summary>
+        Retry,
+    }
+
+    /// <summary>How to handle a full buffer holding items from calls to be retried.</summary>
+    public enum BufferOverflow
+    {
+        /// <summary>Throws and exception when the buffer is full.</summary>
```
evildour: "Throws and" => "Throw an". Another option is to keep "Throws" and change "Ignore" below to "Ignores" so they have a consistent style.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+        Swallow,
+    }
+
+    /// <summary>
+    /// Options that define retry logic for requests that failed.
+    /// </summary>
+    public sealed class RetryOptions
+    {
+        /// <summary>The default buffer size in bytes. 2^18 = 262144.</summary>
+        private const int DefaultBufferSize = 262144;
```
evildour: For powers of 2, I usually like to define my constants by left shifting 1 by the exponent amount, like this:
```csharp
1 << 18
```
But that's just me. Feel free to ignore.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+        public BufferOverflow BufferOverflow { get;  }
+
+        /// <summary>The size of the buffer in bytes, only used for <see cref="RetryType.Retry"/></summary>
+        public int BufferSizeBytes { get; }
+
+        /// <summary>The number of attempts to retry sending information, only used for <see cref="RetryType.Retry"/></summary>
+        public int RetryAttempts { get; }
+
+        /// <summary>The amount of time to wait between retries, only used for <see cref="RetryType.Retry"/></summary>
+        public TimeSpan RetryInterval { get; }
```
evildour: Nit: end these three with '.'

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+            int? bufferSizeBytes = null, int? retryAttempts = null, TimeSpan? retryInterval = null)
+        {
+            RetryType = GaxPreconditions.CheckEnumValue(retryType, nameof(retryType));
+            ExceptionHandling = GaxPreconditions.CheckEnumValue(exceptionHandling, nameof(exceptionHandling));
+            BufferOverflow = GaxPreconditions.CheckEnumValue(bufferOverflow, nameof(bufferOverflow));
+            BufferSizeBytes = GaxPreconditions.CheckArgumentRange(
+                bufferSizeBytes ?? default(int), nameof(bufferSizeBytes), 0, int.MaxValue);
+            RetryAttempts = GaxPreconditions.CheckArgumentRange(
+                retryAttempts ?? default(int), nameof(retryAttempts), 0, int.MaxValue);
+            RetryInterval = retryInterval ?? default(TimeSpan);
```
evildour: I would probably be explicit about these defaults: `0` and `TimeSpan.Zero`. It makes things just a bit easier to grok at a glance: you don't need to do the mental translation of the default expression to the default value in your head. Then again, I don't feel too strongly about it.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+            GaxPreconditions.CheckArgument(RetryInterval >= TimeSpan.Zero, nameof(retryInterval), 
+                $"{nameof(retryInterval)} must be greater than 0");
+        }
+
+        /// <summary>
+        /// Create a <see cref="RetryOptions"/> for <see cref="RetryType.None"/>.
```
evildour: I would use the descriptive "Creates" here (and below) rather than the imperative "Create" to be consistent with other summaries.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// </summary>
+        /// <param name="exceptionHandling">Optional, How exceptions should be handled.
```
evildour: "How" => "how" (same comment twice below as well)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// Create a <see cref="RetryOptions"/> for <see cref="RetryType.Retry"/>.
+        /// </summary>
+        /// <param name="exceptionHandling">Optional, How exceptions should be handled.
+        ///     Defaults to <see cref="ExceptionHandling.Swallow"/>.</param>
+        /// <param name="bufferOverflow">Optional, How a buffer overflow should be handled.
+        ///     Defaults to <see cref="BufferOverflow.IgnoreNewEntries"/>.</param>
+        /// <param name="bufferSizeBytes">Optional, The size of the buffer to be used in bytes.
+        ///     Defaults to 2^18 = 262144.</param>
+        /// <param name="retryInterval">Optional, The time between retries. Default to 2 seconds.</param>
+        /// <param name="retryAttempts">Optional, The number of retry attempts.  Defaults to 10.</param>
```
evildour: Single-space?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+    public enum ExceptionHandling
+    {
+        /// <summary>
+        /// Let the errors propagate up after they have been retried if they continually fail.
+        /// This may kill the running process.
+        /// </summary>
+        Propagate,
+
+        /// <summary>Swallow and ignore the errors.</summary>
+        Swallow,
```
evildour: Perhaps `Ignore` instead?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+        private static readonly TimeSpan DefaultRetryInterval = TimeSpan.FromSeconds(2);
+
+        /// <summary>The type of retry to make.</summary>
+        public RetryType RetryType { get; }
+
+        /// <summary>How exceptions are handled.</summary>
+        public ExceptionHandling ExceptionHandling { get; }
+
+        /// <summary>How to handle a full buffer.</summary>
+        public BufferOverflow BufferOverflow { get;  }
```
evildour: Nit: extra space here

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,214 @@
...
+            }
+        }
+
+        /// <inheritdoc />
+        public void Dispose() => _timer?.Dispose();
+
+        /// <summary>
+        /// Handles a <see cref="RpcException"/> that occurred during a <see cref="Receive(IEnumerable{T})"/> call.
+        /// </summary>
+        private void HandleException(IEnumerable<T> items, RpcException e)
```
evildour: Maybe have this return a bool and use it in an exception filter where this is needed. Return false when you want to propagate the exception so you don't lose stack trace details.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,214 @@
...
+                switch (_options.ExceptionHandling)
+                {
+                    case ExceptionHandling.Propagate:
+                        throw e;
+                    case ExceptionHandling.Swallow:
+                        Debug.WriteLine($"Swallowed exception: {e}");
+                        return;
+                    default:
+                        throw new InvalidOperationException(
+                            $"Invalid ExceptionHandling option: {_options.ExceptionHandling}");
```
evildour: I'd probably throw this kind of exception when the value is specified when creating options and simply debug assert here (same comment below).

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,214 @@
...
+                return;
+            }
+
+            // Move the current buffer into a new buffer to send.
+            _semaphore.Wait();
+            List<T> toSend = _buffer;
+            int toSendSize = _size;
+            _buffer = new List<T>();
+            _semaphore.Release();
+            _retries++;
```
evildour: This looks like it should be guarded by the semaphore as well. Or at least incremented with `Interlocked`.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,214 @@
...
+        /// Function to create a <see cref="ISequentialThreadingTimer"/> that will start the given
+        /// action based on retry options.
+        /// </summary>
+        private readonly Func<Action, RetryOptions, ISequentialThreadingTimer> _timerFactory;
+
+        /// <summary>The timer running in the background to retry failed requests.</summary>
+        private readonly ISequentialThreadingTimer _timer;
+
+        /// <summary>A semaphore to protect the buffer.</summary>
+        private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1);
```
evildour: Why can't this be a normal object which you lock on when modifying the buffer, size, and retry count?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,214 @@
...
+        internal readonly IConsumer<T> _consumer;
+
+        /// <summary>
+        /// The buffered items. This is not readonly as it is replaced when the buffer is successfully send to
+        /// the base consumer.
+        /// </summary>
+        private List<T> _buffer = new List<T>();
+
+        /// <summary>The size of the buffer.</summary>
+        private int _size = 0;
```
evildour: Size is ambiguous here (it could be confused with the number of items in the buffer). Perhaps `_sizeInBytes`?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,214 @@
...
+            _semaphore.Release();
+        }
+
+        /// <summary>
+        /// Try to flush the buffer if it has any entries.
+        /// </summary>
+        private void FlushBuffer()
+        {
+            // If the buffer is empty do nothing.
+            if (_buffer.Count() == 0)
```
evildour: This should be in the lock as well or you should do double-checked locking. Two threads could each check this and pass it before one of them resets the buffer. Then the second thread will flush an empty buffer.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SequentialThreadingTimer.cs
```diff
@@ -0,0 +1,44 @@
...
+        private readonly CancellationTokenSource _source = new CancellationTokenSource();
+
+        /// <inheritdoc />
+        public void Initialize(Action callback, TimeSpan waitTime)
+        {
+            Task.Run(() =>
+            {
+                while (!_source.Token.IsCancellationRequested)
+                {
+                    Thread.Sleep(waitTime);
```
evildour: I would make this an async lambda and use `await Task.Delay(waitTime)` here.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceConfiguration.cs
```diff
@@ -45,10 +49,12 @@ private TraceConfiguration(double qpsSampleRate, BufferOptions bufferOptions)
         ///     how often requests are automatically traced. Defaults to <see cref="DefaultQpsSampleRate"/>
         /// </param>
         /// <param name="bufferOptions">Optional, the buffer options.  Defaults to a <see cref="BufferType.Timed"/></param>
+        /// <param name="retryOptions"></param>
```
evildour: Missing comment

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,214 @@
...
+            // If the new items would be greater than the size of the buffer drop them or throw.
+            if (size + _size > _options.BufferSizeBytes) 
+            {
+                switch (_options.BufferOverflow)
+                {
+                    case BufferOverflow.IgnoreNewEntries:
+                        Debug.WriteLine($"Ignored {items.Count()} entries.");
+                        return;
+                    case BufferOverflow.ThrowException:
+                        throw new InvalidOperationException("Buffer overflow");
```
evildour: Maybe pass `e` as the inner exception here?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceConfiguration.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SequentialThreadingTimer.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -83,9 +83,9 @@ public void Receive(IEnumerable<T> items)
             {
                 _consumer.Receive(items);
             }
-            catch (RpcException e)
+            catch (RpcException e) when (!HandleException(items, e))
```
evildour: This is reversed. Catch when `HandleException` is true and then don't throw in the catch block (same below).

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -110,76 +109,78 @@ public Task ReceiveAsync(IEnumerable<T> items, CancellationToken cancellationTok
...
             if (_options.RetryType == RetryType.None)
             {
                 switch (_options.ExceptionHandling)
                 {
                     case ExceptionHandling.Propagate:
-                        throw e;
-                    case ExceptionHandling.Swallow:
+                        return false;
+                    case ExceptionHandling.Ignore:
                         Debug.WriteLine($"Swallowed exception: {e}");
```
evildour: Ignored?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
evildour: It may be simpler, but it is also more dangerous. Using an object, you can use a lock statement to lock on it, ensuring the lock is released when exiting the critical section. The semaphore requires manual management of the critical section.

Case in point: in your changes just now in `FlushBuffer`, you moved the first `_semaphore.Wait()` call above a return statement, so if you bail early you will deadlock on the next call to `FlushBuffer`.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,215 @@
...
+            _semaphore.Release();
+            return true;
+        }
+
+        /// <summary>
+        /// Try to flush the buffer if it has any entries.
+        /// </summary>
+        private void FlushBuffer()
+        {
+            _semaphore.Wait();
```
evildour: As mentioned above, this needs to be released when you return below. I would suggest using a lock object with double-checked locking here:

```csharp
if (_buffer.Count() == 0)
{
  return;
}

List<T> toSend;
int toSendSize;
lock (_lockObject)
{
  if (_buffer.Count() == 0)
  {
    return;
  }
  toSend = _buffer;
  toSendSize = _sizeInBytes;
  _buffer = new List<T>();
  _retries++;
}
...
```

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: Fair enough, done.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,222 @@
...
+                        Debug.Fail($"Invalid ExceptionHandling option: {_options.ExceptionHandling}");
+                        return true;
+                }
+            }
+            
+            // Get the size of items to be buffered and retried.
+            int size = items.Sum(item => _sizer(item));
+
+            // If the new items would be greater than the size of the buffer drop them or throw.
+            if (size + _sizeInBytes > _options.BufferSizeBytes)
```
evildour: This if block should be in the lock below. There is a race condition here where two thread can both pass this check and then both add to the size in a way that would exceed the limit.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,222 @@
...
+                }
+            }
+            catch (RpcException ex) 
+            {
+                // If we have reached the maximum number of retries. 
+                if (_retries >= _options.RetryAttempts)
+                {
+                    if (_options.ExceptionHandling == ExceptionHandling.Propagate)
+                    {
+                        throw ex;
```
evildour: Use `throw;` instead here.

evildour: This code path does not remove from `_sizeInBytes` despite the fact that it has effectively removed the items from the buffer, so I think it is a bug. I think it would be safer to always update `_buffer` and `_sizeInBytes` together within a lock so anytime you are outside of a lock, you keep to the invariant that `_sizeInBytes` accurately describes the contents in `_buffer`. This would mean setting `_sizeInBytes` to 0 in the first lock above and only using `toSendSize` once in this method: in the lock below to re-add the size to when the items are re-added to the buffer.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,222 @@
...
+
+            try
+            {
+                _consumer.Receive(toSend);
+
+                // The call succeeded, remove update the buffer size and retries.  
+                lock (_mutex)
+                {
+                    _sizeInBytes -= toSendSize;
+                    _retries = 0;
```
evildour: The management of this `_retries` value and the general approach makes me a bit nervous. Do we need to worry about the order of RPCs here? What happens if thread A increments the count above, clears the buffer, and is about to send the content to the consumer, but then on thread B, another item is added to the buffer and it gets flushed before thread A can actually finish? Should this whole method be in a single lock statement?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: I think this is safe only because the bufferflush is private and is called sequentially and the calls will never overlap.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: Done

iantalarico: Much cleaner, thanks.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -174,38 +174,31 @@ private void FlushBuffer()
...
+            catch (RpcException) 
             {
-                // If we have reached the maximum number of retries. 
                 if (_retries >= _options.RetryAttempts)
                 {
-                    if (_options.ExceptionHandling == ExceptionHandling.Propagate)
+                    lock (_mutex)
                     {
-                        throw ex;
+                        _retries = 0;
```
evildour: This doesn't need to be in a lock anymore

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
evildour: Even so, things can still get out of order: you take things out of the buffer above, then something else queues new items in the buffer in `HandleException`, and then you re-add the original items again below in the catch block. Is this ok?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
evildour: Nope, I'm wrong about that. But you can remove the lock if you change the increment of `_retries` to use `Interlocked.Increment` instead.

iantalarico:
I can't comment on your last comment (probably because I pushed the merge).

But yes I think that is okay.  We are just making an effort to get these entries pushed out.  I don't think the ordering matters.  But I may be missing something.

evildour:
Ok, in that case LGTM

iantalarico:
Thanks Mike, I'm going to wait and see if @jskeet has an comments.

jskeet:
Sorry for not getting to this before - particularly with this comment: why are we reinventing retry when we already have RetrySettings (https://github.com/googleapis/gax-dotnet/blob/master/src/Google.Api.Gax.Grpc/RetrySettings.cs) and you can specify that in a client settings already? I realize that's generally a lower-level abstraction, but I don't think we necessarily need to reinvent the ideas here. Maybe the options should just include a RetrySettings that you can specify when you create the individual gRPC clients?

iantalarico:
I looked at the retry settings before but thought it was more relevant to the low lever of reties as you mentioned.  I don't see how the RetrySettings will allow us to accomplish what we need here.  Can you elaborate?  Might be easier to follow up offline?

jskeet:
Yes, let's talk offline. I may well need to understand more about what we're trying to capture here.

iantalarico:
@jskeet I just merged with master. PTAL.

jskeet:
Hmm... it's still claiming it has conflicts. Maybe squash into a single commit to make the rebase simpler?

iantalarico:
It says "This branch has no conflicts with the base branch" for me.  Can you double check?

jskeet:
"This branch cannot be rebased due to conflicts
Rebasing the commits of this branch on top of the base branch cannot be performed automatically due to conflicts encountered while reapplying the individual commits from the head branch."

Odd.

jskeet:
(Don't have time to look further now, I'm afraid.)

jskeet:
I suspect the problem may be that there are merge commits in there as well. I'm going to try fetching it myself and creating a new (single-commit) PR.

jskeet:
It turns out I'm able to push into your branch :)

Steps:

- Rebase from master, fixing up all the conflicts
- Squash everything down to one commit
- Remove redundant "Service Include=..." bits which the project generator now creates earlier
- Force push back to this branch.

So it's now a single commit which rebases cleanly onto master. Will review...

jskeet:
(And now I've broken the build... will fix.)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <param name="bufferOverflow">Optional, how a buffer overflow should be handled.
+        ///     Defaults to <see cref="BufferOverflow.IgnoreNewEntries"/>.</param>
+        /// <param name="bufferSizeBytes">Optional, the size of the buffer to be used in bytes.
+        ///     Defaults to 2^18 = 262144.</param>
+        /// <param name="retryInterval">Optional, the time between retries. Default to 2 seconds.</param>
+        /// <param name="retryAttempts">Optional, the number of retry attempts. Defaults to 10.</param>
+        public static RetryOptions Retry(
+            ExceptionHandling exceptionHandling = ExceptionHandling.Ignore,
+            BufferOverflow bufferOverflow = BufferOverflow.IgnoreNewEntries,
+            int bufferSizeBytes = DefaultBufferSize,
```
jskeet: How certain are we of these defaults? For the enum values I think it's reasonable, but I'd be tempted to use an `int?` for `bufferSizeBytes` and initialize it like `retryInterval`, turning `DefaultBufferSize` into a `public static readonly int` instead, basically so we can change it later safely. (Ditto DefaultRetryAttempts.)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,209 @@
...
+                        Debug.WriteLine($"Ignored exception: {e}");
+                        return true;
+                    default:
+                        Debug.Fail($"Invalid ExceptionHandling option: {_options.ExceptionHandling}");
+                        return true;
+                }
+            }
+            
+            // Get the size of items to be buffered and retried.
+            int size = items.Sum(item => _sizer(item));
```
jskeet: Just `items.Sum(_sizer)` should be fine, I think.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
jskeet: I would avoid the double-checked locking to be honest - let's not try to be clever; just always lock. (That just means removing these 5 lines of code - you've already got the early out later.)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,209 @@
...
+            catch (RpcException e) when (HandleException(items, e)) { }
+        }
+
+        /// <inheritdoc />
+        public Task ReceiveAsync(IEnumerable<T> items, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            GaxPreconditions.CheckNotNull(items, nameof(items));
+            try
+            {
+                return _consumer.ReceiveAsync(items, cancellationToken);
```
jskeet: One aspect about this that might be a little odd. Suppose we have two entries, E1 and E2. The first fails, so needs to be retried, but the second succeeds. We could easily end up sending E2 before E1, because E1 is only sent on a timer. Is that a problem?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,209 @@
...
+                toSendSize = _sizeInBytes;
+                _sizeInBytes = 0;
+                _buffer = new List<T>();
+            }
+            Interlocked.Increment(ref _retries);
+
+            try
+            {
+                _consumer.Receive(toSend);
+                _retries = 0;
```
jskeet: I'm never sure how safe this is. I tend to try to use variables with Interlocked *only* if I'm using it at all. You can use `Interlocked.Exchange(ref _retries, 0)` as an unconditional write.

(The issue would be if the simple write to 0 is buffered somehow. It's a shame the memory model isn't more clearly documented.)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,209 @@
...
+                {
+                    _retries = 0;
+                    if (_options.ExceptionHandling == ExceptionHandling.Propagate)
+                    {
+                        throw;
+                    }
+                }
+                else
+                {
+                    // The call failed, add the items that were not successfully sent back to the buffer.
```
jskeet: This could mean the buffer is now bigger than it should be, of course... and it could keep growing that way.
I think it might be better to reduce _sizeInBytes by the size sent, and clear out the sent number of items from the list, when the send is successful.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SequentialThreadingTimer.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <inheritdoc />
+        public void Initialize(Action callback, TimeSpan waitTime)
+        {
+            Task.Run(async () =>
+            {
+                while (!_source.Token.IsCancellationRequested)
+                {
+                    await Task.Delay(waitTime).ConfigureAwait(false);
+                    callback();
```
jskeet: What do we want to happen if the callback throws an exception?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceConfiguration.cs
```diff
@@ -0,0 +1,60 @@
...
+
+using Google.Api.Gax;
+using Google.Cloud.Diagnostics.Common;
+
+namespace Google.Cloud.Diagnostics.Common
+{
+    /// <summary>
+    /// Optional configuration to be used when initializing tracing.
+    /// </summary>
+    public sealed class TraceConfiguration
```
jskeet: Is this file meant to be in this PR, or did I mess up somehow? It doesn't seem to be referenced anywhere else.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceConfiguration.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <summary>
+        /// Creates a <see cref="TraceConfiguration"/>.
+        /// </summary>
+        /// <param name="qpsSampleRate">Optional, the qps sample rate.  The sample rate determines
+        ///     how often requests are automatically traced. Defaults to <see cref="DefaultQpsSampleRate"/>.
+        /// </param>
+        /// <param name="bufferOptions">Optional, the buffer options.  Defaults to a <see cref="BufferType.Timed"/>.</param>
+        /// <param name="retryOptions">Optional, the retry options.  Defaults to a <see cref="RetryType.None"/>.</param>
+        public static TraceConfiguration Create(
+            double qpsSampleRate = DefaultQpsSampleRate, BufferOptions bufferOptions = null, RetryOptions retryOptions = null)
```
jskeet: Expression bodied member?

jskeet:
You've merged your local branch with the changes I've made... I suggest that:

- I force push again, bringing it back to where it was before you merged
- You force *pull*, overwriting the long list of commits with just the single commit
- You make any further changes required

Sound good?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SequentialThreadingTimer.cs
iantalarico: I think we should let this propagate and have the user handle exceptions in the callback?  I want to keep this class simple

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ConsumerFactory.cs
```diff
@@ -24,23 +24,35 @@ internal static class ConsumerFactory<T>
...
+            BufferOptions bufferOptions, RetryOptions retryOptions)
         {
             GaxPreconditions.CheckNotNull(consumer, nameof(consumer));
-            GaxPreconditions.CheckNotNull(options, nameof(options));
+            GaxPreconditions.CheckNotNull(sizer, nameof(sizer));
+            GaxPreconditions.CheckNotNull(bufferOptions, nameof(bufferOptions));
+            GaxPreconditions.CheckNotNull(retryOptions, nameof(retryOptions));
 
-            switch (options.BufferType)
+            Func<Action, RetryOptions, ISequentialThreadingTimer> timerFactory =
```
iantalarico: I'm going to leave as is for now unless you have strong opinions otherwise.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -56,13 +56,13 @@ public enum ExceptionHandling
     public sealed class RetryOptions
     {
         /// <summary>The default buffer size in bytes. 2^18 = 262144.</summary>
-        private const int DefaultBufferSize = 262144;
+        public static readonly int DefaultBufferSize = 262144;
```
jskeet: Possibly don't document the current default value, or explicitly warn that it might change?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RetryOptions.cs
```diff
@@ -113,15 +113,15 @@ public static RetryOptions NoRetry(ExceptionHandling exceptionHandling = Excepti
...
+        /// <param name="retryAttempts">Optional, the number of retry attempts. Defaults to <see cref="DefaultRetryAttempts"/>.</param>
         public static RetryOptions Retry(
             ExceptionHandling exceptionHandling = ExceptionHandling.Ignore,
             BufferOverflow bufferOverflow = BufferOverflow.IgnoreNewEntries,
-            int bufferSizeBytes = DefaultBufferSize,
-            TimeSpan? retryInterval = null, int retryAttempts = DefaultRetryAttempts)
+            int? bufferSizeBytes = null, TimeSpan? retryInterval = null, int? retryAttempts = null)
                 => new RetryOptions(RetryType.Retry, exceptionHandling, bufferOverflow,
-                    bufferSizeBytes, retryAttempts, retryInterval ?? DefaultRetryInterval);
+                    bufferSizeBytes ?? DefaultBufferSize, retryAttempts ?? DefaultRetryAttempts,
```
jskeet: Can we reorder either this method or the RetryOptions constructor to keep the parameter order consistent?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -174,21 +168,24 @@ private void FlushBuffer()
...
+                    _retries = 0;
+                    _sizeInBytes -= toSendSize;
+                }
             }
             catch (RpcException) 
             {
                 if (_retries >= _options.RetryAttempts)
                 {
-                    _retries = 0;
+                    Interlocked.Exchange(ref _retries, 0);
```
jskeet: Doesn't this mean we start retrying again?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,205 @@
...
+                {
+                    _retries = 0;
+                    _sizeInBytes -= toSendSize;
+                }
+            }
+            catch (RpcException) 
+            {
+                if (_retries >= _options.RetryAttempts)
+                {
+                    Interlocked.Exchange(ref _retries, 0);
```
jskeet: Why do we reset the number of retries here? Doesn't that mean we'll keep retrying? It's not really clear to me what the behavior should be at this point if we've said we want to swallow exceptions. Am I right in saying that at the moment, it's "drop the contents of the buffer, but retry future buffers"?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SequentialThreadingTimer.cs
jskeet: I'm concerned now about where the exception *is* going to be handled. You're actually using this (I think) from RpcRetryConsumer using FlushBuffer as the callback. So if that throws an exception (due to the option) it's going to just be thrown from Task.Run. Is that exception going to be swallowed by the system, or bring down the process, or something else? Is there any way the user can even see it?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -0,0 +1,209 @@
...
+using System.Threading.Tasks;
+
+
+namespace Google.Cloud.Diagnostics.Common
+{
+    /// <summary>
+    /// An <see cref="IConsumer{T}"/> that will allow for the option to retry RPC failures and buffer
+    /// the data while the retries occur.
+    /// </summary>
+    internal class RpcRetryConsumer<T> : IConsumer<T>
```
iantalarico: Not in this PR but I have them in https://github.com/iantalarico/google-cloud-dotnet/tree/retry-test and will patch up and send a PR for them once this PR is in.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: I missed updating this but yes we reset the retries (and update the buffer size) if we are swallowing things

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SequentialThreadingTimer.cs
iantalarico: Correct.  It will bring down the process if the user is not swallowing processes (after the given retries). The user will not see it if it is being swallowed.

jskeet:
Okay. I think it may be worth being even clearer in the options what `Propagate` means. But other than that, when it's green again (and with the constructor ordering fixed that I mentioned), I'm fine to merge.

iantalarico:
@jskeet @evildour  Thanks for all the feedback!

PR #1110, opened 2017-06-12
Merge Spanner work into master

jskeet:
This has been rebased to master, and then commit merged to avoid small "code review" commits


jskeet:
This is now as clean as I can easily make the commit log, I think.

Ben, if you could pull this and run the tests, that would be great. It at least builds...

jskeet:
AppVeyor failure was due to project file not having ConfigureAwait checker. I'm not sure why that wasn't there, when it's in cloud-spanner.

Abandoning this attempt... I'll try a different merge approach.

PR #1116, opened 2017-06-13
Spanner review source

benwulfe:
For the spanner team, this is the C# spanner library review in its entirety.  This is already in master, but this PR was created for your review.


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolImpl.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        public void ReleaseSessionToPool(SpannerClient client, Session session)
+        {
+            Logger.Debug(() => "Placing session back into the pool and starting the evict timer.");
+            //start evict timer.
+            SessionPoolEntry entry = new SessionPoolEntry(session, new CancellationTokenSource());
+
+            if (SessionPool.UseTransactionWarming)
+            {
+                Task.Run(() => client.PreWarmTransactionAsync(entry.Session));
```
vkedia: Will this be invoked for Readonly transactions as well? If yes, then that is probably not a good idea because ReadOnlyTransaction involves picking a timestamp which should be picked when user explicitly creates a transaction.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
```diff
@@ -0,0 +1,391 @@
...
+    /// To create a session using the sessionpool, use the extension method on SpannerClient.
+    ///   var session = await SpannerClient.CreateSessionFromPoolAsync(project, spannerInstance, database, cancellationtoken);
+    /// To release a session back to the sessionpool:
+    ///   session.ReleaseToPool();
+    /// If you use the SessionPool, you must make sure sessions are properly released back into the pool and are not simply GC'd.
+    /// Allowing a session to GC will incur penalties on the current process as the session will count towards the maximum allowed
+    /// and it will incur a penalty on other Spanner processes because it will be an hour before the server frees the session if 
+    /// it's not properly deleted.
+    /// </summary>
+    internal static class SessionPool
```
vkedia: Is there a Close method on the pool which can be invoked at the time of shutdown which will delete all the sessions that were created in this pool?

vkedia: Do you have a test like this:
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-spanner/src/test/java/com/google/cloud/spanner/SessionPoolStressTest.java

This essentially does lot of operations on the pool concurrently and then asserts that all the invariants are maintained. It mocks out the spanner api so it runs very fast. This would be very useful to catch subtle concurrency bugs.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolImpl.cs
benwulfe: It all runs through the same code path here (to keep all paths the same) until it gets to actually filling in transaction details in TransactionPool.PreWarmAsync -- there you will see this code:
                // for now, we only prewarm readwrite transactions because the read transaction semantics are usually
                // dependent on the time the transaction begins.
                if (options?.ModeCase == TransactionOptions.ModeOneofCase.ReadWrite)
                {
                    Logger.Debug(
                        () => $"Pre-warming session transaction state. Mode={options.ModeCase}");
                    ActiveTransactionOptions = options;
                    PreWarmTask = Task.Run(() => CreateTransactionImplAsync(session));
                }
                else
                {
                    MarkTransactionUsed();
                }

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
benwulfe: SessionPool.ReleaseAll does that and you'll see its called on AppDomain.Unload (which is more or less shutdown in .net speak).

benwulfe: I'm going to be adding unit tests which include this, performance tests, stress tests and type conversion tests (there are quite a lot of scenarios of storing a value of type 'a' which gets stored as 'b' and pulled out as 'c').
There are also a bunch of tests around dataadapter, connectionstrings, transaction scopes.
its quite a list...

benwulfe: oh so the direct answer to your question:  not yet, but coming!

wthsts: Thanks Ben.

benwulfe:
Hey guys this review is a bit old.  Did you want to meet or do anything else to review this code?

PR #1117, opened 2017-06-13
Fix the QPS integration test by resetting the rate limiter before each test.

### LINKED ISSUE #966: Fix Flakey Test - AspNetCore.IntegrationTests.TraceTest.Trace_QPS
This test is continually flakey.

iantalarico:
We also disable palatalization to ensure the rate limiter is correct when each test is run.

Also ensures all test servers and clients are properly disposed of.

@jskeet @evildour Any idea how to handle this in a better way?

Fixes #966

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
@@ -38,7 +38,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
         public TraceTest()
         {
             // The rate limiter instance is static and only set once.  If we do not reset it at the
-            // beginning of each tests the qps will not change.  This is dependent the tests not
+            // beginning of each tests the qps will not change.  This is dependent on the tests not
             // running in parallel.
             RateLimiter._instance = null;
         }
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
@@ -40,7 +40,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
             // The rate limiter instance is static and only set once.  If we do not reset it at the
             // beginning of each tests the qps will not change.  This is dependent on the tests not
             // running in parallel.
-            RateLimiter._instance = null;
+            RateLimiter.Reset();
         }
 
         [Fact]
@@ -205,6 +205,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
                 try
                 {
                     await client.GetAsync($"/Trace/ThrowException/{testId}");
+                    Assert.True(false, "Exception should have been thrown.");
                 }
                 catch
                 {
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -193,24 +199,27 @@ public class TraceTest
...
 
-            var client = _noBufferHighQps.CreateClient();
-            try
+            using (var server = new TestServer(new WebHostBuilder().UseStartup<TraceTestNoBufferHighQpsApplication>()))
+            using (var client = server.CreateClient())
             {
-                await client.GetAsync($"/Trace/ThrowException/{testId}");
+                try
+                {
+                    await client.GetAsync($"/Trace/ThrowException/{testId}");
```
evildour: Not that its something you added here, but when an exception is expected, it's probably a good idea to assert if you get past this line using `Assert.True(false, ...)`

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -35,32 +35,37 @@ public class TraceTest
...
-            new WebHostBuilder().UseStartup<TraceTestNoBufferHighQpsApplication>());
-
-        private readonly TestServer _noBufferLowQps = new TestServer(
-            new WebHostBuilder().UseStartup<TraceTestNoBufferLowQpsApplication>());
+        public TraceTest()
+        {
+            // The rate limiter instance is static and only set once.  If we do not reset it at the
+            // beginning of each tests the qps will not change.  This is dependent on the tests not
+            // running in parallel.
+            RateLimiter._instance = null;
```
evildour: Does this actually reset it for each test? Seems like this would only be run once, but I could be wrong.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/RateLimiter.cs
```diff
@@ -27,7 +27,7 @@ internal sealed class RateLimiter
         private static object _instanceMutex = new object();
 
         /// <summary>The single rate limiter instance.</summary>
-        private static RateLimiter _instance;
+        internal static RateLimiter _instance;
```
evildour: I would probably have a `Reset()` static method or something similar rather than make this internal.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
iantalarico: Yea its fairly confusing using the constructor for a before each, but that how xunit works.

https://xunit.github.io/docs/shared-context.html

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
iantalarico: Fixed

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/RateLimiter.cs
iantalarico: Done.

jskeet:
I haven't looked at this carefully, but does the rate limiter have to be a singleton? Even if it's used as a singleton from the outside, we could presumably have an internal constructor...

iantalarico:
@jskeet  I think we need to have the rate limiter be a singleton to ensure we properly limit qps when throttling traces.

evildour:
LGTM, but I'll let Jon give final approval in case he still wants to discuss singleton vs. internal constructor.

iantalarico:
SGTM, thanks Mike!

jskeet:
I'm fine for there to be a single instance for public consumption, but for the unit tests, I think it would make life easier if we could create a separate instance. As this is an integration test, I'm not sure whether we need to be testing the rate limiting at all, if that's also unit tested everywhere else. Maybe I'm just missing something.

Either way, this is probably better than it was, so if Mike's happy with the change, I'm happy... I assume we know that the constructor *is* called at the start of each test?

jskeet:
What's the status of this? Are we convinced it's better, so should be merged even if we might revisit? I'm nervous of PRs that get old...

iantalarico:
Sorry I was OOO for the last 2 weeks. Merging now!

PR #1121, opened 2017-06-14
Remove the agent log label.

iantalarico:
This is a lot of extra space to store in logs and will be repalced with a header that does not affect log size.


iantalarico:
Thanks Mike!

PR #1123, opened 2017-06-14
Broke dataadapter a few ways:

benwulfe:
1. the change in spannerparameter ctor caused the wrong argument to be set.
2. missing configureawait
3. the commands set on the rowupdating event (vs interface) is too late to be handled by ParameterInput, causing bad values.
4. SourceColumnNullMapping should not be true as it breaks insert commands done through dataadapter.


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataAdapter.cs
```diff
@@ -92,6 +92,17 @@ public string AutoGeneratedCommandTable
             get => _autoGeneratedCommandTable;
             set
             {
+                // Note that we auto build the commands as a feature of the *Data adapter* versus a separate class.
+                // This is done for two reasons:
+                // a) We cannot create a "proper" DbCommandBuilder because DDL is not supported by Spanner.
+                //    This means any "SpannerCommandBuilder" would not live up to expectations and be castable
+                //    to DbCommandBuilder.
+                // b) The code for building these commands is both much simpler than othe providers and also
```
jskeet: Nit: othe => other

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataAdapter.cs
```diff
@@ -211,7 +222,7 @@ protected override void OnRowUpdated(RowUpdatedEventArgs rowUpdatedEventArgs)
                 foreach (var field in readerMetadata.RowType.Fields)
                 {
                     _parsedParameterCollection.Add(
-                        new SpannerParameter(field.Name, SpannerDbType.FromProtoBufType(field.Type), field.Name));
+                        new SpannerParameter(field.Name, SpannerDbType.FromProtoBufType(field.Type), null, field.Name));
```
jskeet: While I see it, didn't we rename FromProtoBufType to FromProtobufType?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataAdapter.cs
```diff
@@ -279,42 +278,6 @@ private SpannerCommand GetBuiltSelectCommand()
         /// <inheritdoc />
         protected override void OnRowUpdating(RowUpdatingEventArgs rowUpdatingEventArgs)
         {
-            if (rowUpdatingEventArgs.Command == null && AutoCreateCommands)
```
jskeet: I don't really understand what this would be doing. Could you edit the PR description to clarify it a bit, just for the record?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataAdapter.cs
benwulfe: missed this one.  fixed.

benwulfe:
When I tested it last week, I made a mistake.  The way dbcommandbuilder works is that it hooks rowupdating and builds dbCommand objects if not specified in the dbdataadapater.  Important in this is that it also sets values (and incidentally rebuilds the command for each row+value).

When I refactored this code into SpannerDataAdapter, values stopped being set (and I actually broke it in multiple ways) because the inner logic inside dbDataAdapter to set values only runs on commands returned by the dataadapter -- *not* on commands returned by the dbCommandBuilders.

The TLDR is that the new way of returning the prebuild commands ensures that the base class will do "its thing" and actually set values into the command from the row.  the other hook I used to do this (rowupdating) bypasses this logic because the commandbuilder does its own value set.

PR #1125, opened 2017-06-15
delete fix for dataadapter.

benwulfe:
More testing for DataAdapter revealed that delete had an error.
More tests coming, but would like to include this fix in the alpha update tonight.


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataAdapter.cs
```diff
@@ -155,6 +161,11 @@ IDbCommand IDbDataAdapter.UpdateCommand
         }
 
         /// <summary>
+        /// The set of primary keys defined for <see cref="AutoGeneratedCommandTable"/>.
+        /// </summary>
+        public HashSet<string> AutoGeneratedCommandPrimaryKeys { get; } = new HashSet<string>();
```
evildour: Does this need to be exposed as a mutable collection? Or exposed at all?

benwulfe: yes for the case where the developer uses the default ctor (which is the designer scenario).
Not sure if we'll be able to get the toolbox working near term, but I don't want to cut off the use case.

ILMTitan: The user does not need to know it is a `HashSet<string>`. The `ISet<string>` interface would do.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataAdapter.cs
```diff
@@ -119,11 +120,16 @@ public string AutoGeneratedCommandTable
...
+        /// <param name="primaryKeys">The set of columns that form the primary key for <paramref name="autoGeneratedCommandTable"/>.</param>
+        public SpannerDataAdapter(SpannerConnection connection, string autoGeneratedCommandTable, params string[] primaryKeys)
         {
             GaxPreconditions.CheckNotNull(connection, nameof(connection));
             SpannerConnection = connection;
             AutoGeneratedCommandTable = autoGeneratedCommandTable;
+            if (primaryKeys != null)
+            {
+                AutoGeneratedCommandPrimaryKeys.UnionWith(primaryKeys);
+            }
```
ILMTitan: It is just a matter of preference, but I tend to think lines like this look better with null coalescing.
```
AutoGeneratedCommandPrimaryKeys.UnionWith(primaryKeys ?? Enumerable.Empty<string>());
```

benwulfe: interesting option.  next time :)  thanks for the suggestion.

PR #1137, opened 2017-06-22
Adjust documentation for WriteDisposition in jobs.

### LINKED ISSUE #1136: CreateJobOptions.WriteDisposition default is incorrectly documented
We claim that the default write disposition is WriteAppend, but it's actually WriteEmpty.

This is just a doc fix, but I should check it for all WriteDisposition properties...

jskeet:
- Fix the default for CreateQuery
- Make all comments have a similar format

Fixes #1136.


jskeet:
(AppVeyor seems to be stuck. Merging without.)

PR #1144, opened 2017-06-22
Tool to tag releases automatically

### LINKED ISSUE #1141: Better release tagging tool
We should be able to compare apis.json with what's in NuGet and tag new versions that way. More reliable than working it out by hand.

(First pass: it could generate a shell script...)

jskeet:
Fixes #1141.

If we find there are situations where we want more flexibility (e.g.
just tagging some releases) we can build that in later.


tools/Google.Cloud.Tools.TagReleases/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine($"Error: {e}");
+                return 1;
+            }
+            return 0;
+        }
+
+        private static async Task MainAsync(string[] args)
```
ILMTitan: I'm excited about C# 7.1's `async Main` when this can just be the main method.

tools/Google.Cloud.Tools.TagReleases/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+        {
+            var client = new GitHubClient(new ProductHeaderValue(ApplicationName))
+            {
+                Credentials = new Credentials(args[0])
+            };
+            var tags = (await client.Repository.GetAllTags(RepositoryOwner, RepositoryName)).Select(tag => tag.Name);
+
+            var apis = LoadApis();
```
ILMTitan: I argue that `tags` and `apis` should use an explicit type. I like to be able to know the type by reading the current method, especially when reading online and not from an IDE.

tools/Google.Cloud.Tools.TagReleases/Program.cs
jskeet: Me too :)

tools/Google.Cloud.Tools.TagReleases/Program.cs
jskeet: Mmm. I'm somewhat torn. I tend to regard quick tools like this as having a slightly lower bar than prod code (just like I use `var` more in test code) - not because it shouldn't be readable, just because the size of scope tends to be smaller. (It's also less likely that you *will* be reading tool code online, I suspect.)

PR #1151, opened 2017-06-28
Hi-perf Publisher, and unit tests

chrisdunelm:


chrisdunelm:
Decision needs about where to put the code in InGax.cs
Subscriber to follow later; integration tests to follow once Subscriber is in.

jskeet:
Looks like you need to run generateprojects.sh (after carefully making sure that won't destroy stuff) to satisfy AppVeyor. Will review code now though.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+                    if (credentials.IsCreateScopedRequired)
+                    {
+                        credentials = credentials.CreateScoped(PublisherClient.DefaultScopes);
+                    }
+                    channelCredentials = credentials.ToChannelCredentials();
+                }
+                // Create the channels and clients, and register shutdown functions for each channel
+                var clients = new PublisherClient[clientCount];
+                var shutdowns = new Func<Task>[clientCount];
+                for (int i = 0; i < clientCount; i += 1)
```
jskeet: i++ instead of i += 1?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+                BatchingSettings = other.BatchingSettings?.Clone();
+                Scheduler = other.Scheduler;
+                BatchMaxElementCount = other.BatchMaxElementCount;
+                BatchMaxByteCount = other.BatchMaxByteCount;
+            }
+
+            /// <summary>
+            /// The clients to use within a <see cref="HiPerfPublisher"/>.
+            /// </summary>
+            public IList<PublisherClient> Clients { get; }
```
jskeet: Thinking about it again, it feels odd for Clients to be part of the settings in the first place. It doesn't feel like the other settings at all...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+
+        /// <summary>
+        /// Publish a message to the topic specified in <see cref="TopicName"/>.
+        /// </summary>
+        /// <param name="message">The <see cref="PubsubMessage"/> to publish.</param>
+        /// <param name="encoding">The encoding for string to byte conversion.
+        /// If <c>null</c>, defaults to <see cref="Encoding.UTF8"/>.</param>
+        /// <returns>A task which completes once the message has been published.
+        /// The task <see cref="Task{String}.Result"/> contains the message ID.</returns>
+        public virtual Task<string> PublishAsync(string message, Encoding encoding = null) =>
```
jskeet: Did the PubSub team sign off on this in the end then?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+        /// Shutdown this <see cref="HiPerfPublisher"/>.
+        /// The returned <see cref="Task"/> completes when all queued messages have been published.
+        /// The returned <see cref="Task"/> cancels if the passed <see cref="CancellationToken"/> is cancelled.
+        /// Cancelling the passed <see cref="CancellationToken"/> aborts publishing, and cancels the returned
+        /// <see cref="Task"/> as quickly as possible.
+        /// </summary>
+        /// <param name="hardStopToken">Cancel this <see cref="CancellationToken"/> to abort publishing queued messages.</param>
+        /// <returns>A <see cref="Task"/> that completes when all queued messages have been published; or cancels if the
+        /// passed <see cref="CancellationToken"/> is cancelled.</returns>
+        public virtual Task ShutdownAsync(CancellationToken hardStopToken) => throw new NotImplementedException();
```
jskeet: Interesting - this is quite a different use of a cancellation token. I simultaneously like it as a clean design, and dislike it for being surprising. Good call on not calling the parameter cancellationToken though.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+    /// Implementation of PubSub publisher that is associated with a specific <see cref="TopicName"/>.
+    /// </summary>
+    public sealed class HiPerfPublisherImpl : HiPerfPublisher
+    {
+        private class Batch
+        {
+            public TaskCompletionSource<IList<string>> BatchCompletion { get; } = new TaskCompletionSource<IList<string>>();
+            public List<PubsubMessage> Messages { get; } = new List<PubsubMessage>();
+            public CancellationTokenSource TimerCts { get; } = new CancellationTokenSource();
+            public long ByteCount { get; private set; }
```
jskeet: So this is the byte count of the messages, not necessarily the byte count of the request itself (which will include other bits) right?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.PubSub.V1
+{
+    /// <summary>
+    /// A PubSub publisher that is associated with a specific <see cref="TopicName"/>.
+    /// </summary>
+    public abstract class HiPerfPublisher
```
jskeet: Is this name mandated by the PubSub team? I'm not terribly keen, although I can live with it. If it's up to us, let's consider alternative.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+        internal HiPerfPublisherImpl(TopicName topicName, Settings settings, Func<Task> shutdown, TaskHelper taskHelper)
+        {
+            TopicName = GaxPreconditions.CheckNotNull(topicName, nameof(topicName));
+            GaxPreconditions.CheckNotNull(settings, nameof(settings));
+            GaxPreconditions.CheckNotNull(settings.Clients, nameof(settings.Clients));
+            settings.Validate();
+            _taskHelper = GaxPreconditions.CheckNotNull(taskHelper, nameof(taskHelper));
+            _shutdown = shutdown;
+
+            // Initialise settings
```
jskeet: Let's revisit this if/when you make BatchingSettings immutable.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+        private TaskCompletionSource<int> _shutdownTcs;
+        private bool _shutdownStarted;
+
+        /// <inheritdoc/>
+        public override TopicName TopicName { get; }
+
+        /// <inheritdoc/>
+        public override async Task<string> PublishAsync(PubsubMessage message)
+        {
+            CheckShutdown();
```
jskeet: Do we make any guarantees about what happens if we shut down after this check but before a lot of the rest of the method executes?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+            int index;
+            int messageByteCount = message.CalculateSize();
+            lock (_lock)
+            {
+                // Update flow-control counts.
+                _queueElementCount += 1;
+                _queueByteCount += messageByteCount;
+                // Queue the current batch if this message would cause the batch to go over-byte-size
+                // (unless this would be the only message in the batch, then it's allowed).
+                QueueCurrentBatchIfRequired(false, messageByteCount);
```
jskeet: Named arguments would help the clarity here, I think.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+        }
+
+        private void CheckShutdown() => GaxPreconditions.CheckState(!_softStopCts.IsCancellationRequested, "Publisher already shutdown, cannot use");
+
+        private void ShutdownIfCompleted()
+        {
+            // Pre-condition: Must be locked if not a hard stop
+            if (_hardStopCts.IsCancellationRequested || (_softStopCts.IsCancellationRequested && _batchesInFlightCount == 0 && _batchesReady.Count == 0))
+            {
+                if (_shutdownLock.Locked(() => {
```
jskeet: This is more confusing to me than:

```csharp
lock (_shutdownLock)
{
    var shutdownAlreadyStarted = _shutdownAlreadyStarted ;
    _shutdownStarted = true;
    if (shutdownAlreadyStarted)
    {
       return;
    }
}
// Rest of code

File: apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs [hunk 11]
```diff
@@ -0,0 +1,626 @@
...
+                }
+            }
+        }
+
+        // Not "async void" to gain standard async exeption handling
+        private async Task CreateBatch()
+        {
+            // Pre-condition: Must be locked - lock active up until first 'await'.
+            _currentBatch = new Batch();
+            var timerToken = _currentBatch.TimerCts.Token;
```
jskeet: This is a bit confusing, being in the same async method. Can we use Task.Run or a scheduler equivalent instead? Happy to hear about the problems of that on a VC.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+                if (!timerToken.IsCancellationRequested)
+                {
+                    // Force queuing of the current batch, whatever the size.
+                    // There will always be at least one message in the batch.
+                    QueueCurrentBatchIfRequired(true, 0);
+                }
+            }
+        }
+
+        private void QueueCurrentBatchIfRequired(bool force, int extraByteCount)
```
jskeet: I think I'd split this into two methods:

- QueueCurrentBatchIfByteCountExceeded (or similar)
- QueueCurrentBatch

The former could call the latter...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+                // Mark that there is no current batch.
+                _currentBatch = null;
+                // Trigger send to server.
+                Task unusedSendTask = TriggerSend();
+            }
+        }
+
+        // Pre-condition: Must be locked
+        // Allow one message that is larger than requestByteThreshold
+        private bool CurrentBatchIsFull(bool isQueuing, int extraByteCount) => isQueuing ?
```
jskeet: I'm finding this very hard to understand. Rather than struggle through it, could you elaborate in a comment, or direct the reader to an existing explanation?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+                _currentBatch.Messages.Count >= _batchMaxElementCount || (_currentBatch.Messages.Count > 0 && _currentBatch.ByteCount >= _batchMaxByteCount) :
+                _currentBatch.Messages.Count >= _batchElementCountThreshold || (_currentBatch.Messages.Count > 0 && _currentBatch.ByteCount >= _batchRequestByteThreshold);
+
+        // Not "async void" to gain standard async exeption handling
+        private async Task TriggerSend()
+        {
+            // Pre-condition: Must be locked - lock active up until first 'await'
+            // Start sending a batch if there's a batch to send, and a client to use to send it.
+            if (_batchesReady.Count > 0 && _idleClients.Count > 0)
```
jskeet: Invert the condition and just return? (Reduces nesting and mental strain...)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+                        else
+                        {
+                            _shutdownTcs.SetResult(0);
+                        }
+                    });
+                }
+            }
+        }
+
+        // Not "async void" to gain standard async exeption handling
```
jskeet: Except that the task returned is never awaited, so exceptions are just getting swallowed, aren't they? Ditto later for TriggerSend...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+        private async Task CreateBatch()
```
jskeet: Possibly make this a local method to make it clear it's only being called from one place? (Ditto TriggerSend.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+        private async Task TriggerSend()
+        {
+            // Pre-condition: Must be locked - lock active up until first 'await'
+            // Start sending a batch if there's a batch to send, and a client to use to send it.
+            if (_batchesReady.Count > 0 && _idleClients.Count > 0)
+            {
+                // Remove client and batch from relevant queues.
+                var client = _idleClients.Dequeue();
+                var batch = _batchesReady.Dequeue();
+                _batchesInFlightCount += 1;
```
jskeet: Any reason for +=1 instead of ++?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+                // Update flow-control counts.
+                _queueElementCount -= batch.Messages.Count;
+                _queueByteCount -= batch.ByteCount;
+                // Perform the RPC to server, catching exceptions.
+                try
+                {
+                    var response = await _taskHelper.ConfigureAwait(client.PublishAsync(TopicName, batch.Messages, CallSettings.FromCancellationToken(_hardStopCts.Token)));
+                    batch.BatchCompletion.SetResult(response.MessageIds);
```
jskeet: Do we not have a PropagateResult method somewhere? I know I've considered it before now...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+                }
+                catch (Exception e)
+                {
+                    batch.BatchCompletion.SetException(e);
```
jskeet: Does this handle cancellation properly?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+                // A client is now idle. Record it, and see if a further batch is ready to send.
+                lock (_lock)
+                {
+                    _batchesInFlightCount -= 1;
+                    _idleClients.Enqueue(client);
+                    ShutdownIfCompleted();
+                    if (_batchesReady.Count > 0)
+                    {
+                        // Already a batch in the ready-queue, so just send it.
+                        Task unusedSendTask = TriggerSend();
```
jskeet: The recursion here is slightly confusing - I worry that something will be keeping track of a logical call stack, and after blasting a lot of messages, it'll get very large. I wonder whether we might want a different approach to the threading for this. Probably best talked through on VC.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/HiPerfPublisherTest.cs
```diff
@@ -0,0 +1,225 @@
...
+        {
+            var topicName = new TopicName("FakeProject", "FakeTopic");
+            var scheduler = new TestScheduler(threadCount);
+            TaskHelper taskHelper = scheduler.TaskHelper;
+            var clients = Enumerable.Range(0, clientCount).Select(_ => new FakePublisher(scheduler, taskHelper)).ToArray();
+            var settings = MakeSettings(clients, scheduler, batchElementCountThreshold: batchElementCountThreshold, batchRequestByteThreshold: 10000);
+            int shutdownCount = 0;
+            var pub = new HiPerfPublisherImpl(topicName, settings, () =>
+            {
+                shutdownCount += 1;
```
chrisdunelm: This *should* only ever be called once, but if there's an error it could be called multiple times. Done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/HiPerfPublisherTest.cs
```diff
@@ -0,0 +1,225 @@
...
+            {
+                var id = await taskHelper.ConfigureAwait(pub.PublishAsync("1"));
+                bool isCancelled = await taskHelper.ConfigureAwaitHideCancellation(pub.ShutdownAsync(new CancellationToken(hardStop)));
+                Assert.Equal(hardStop, isCancelled);
+                Assert.Equal("1", id);
+                Assert.Equal(1, shutdownCount);
+            });
+        }
+
+        [Theory, PairwiseData]
```
chrisdunelm: It ensures every possible pair of values are tested. Perhaps the [nunit documentation](https://github.com/nunit/docs/wiki/Pairwise-Attribute) is better.
It hugely reduces the combinatorial explosion of test-cases compared to using `CombinatorialData`.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/HiPerfPublisherTest.cs
```diff
@@ -0,0 +1,225 @@
...
+            var settings = MakeSettings(clients, scheduler, batchElementCountThreshold: batchElementCountThreshold, batchRequestByteThreshold: 10000);
+            int shutdownCount = 0;
+            var pub = new HiPerfPublisherImpl(topicName, settings, () =>
+            {
+                shutdownCount += 1;
+                return Task.FromResult(0);
+            }, taskHelper);
+            scheduler.Run(async () =>
+            {
+                Task<string>[] tasks = new Task<string>[messageCount];
```
chrisdunelm: Done.
I felt slightly uneasy using the more functional constructs (ie `Select`) with the imperative nature of `PublishAsync`.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/HiPerfPublisherTest.cs
```diff
@@ -0,0 +1,225 @@
...
+                for (int i = 0; i < tasks.Length; i++)
+                {
+                    tasks[i] = pub.PublishAsync(i.ToString());
+                }
+                var ids = new HashSet<string>(await taskHelper.ConfigureAwait(taskHelper.WhenAll(tasks)));
+                var isCancelled = await taskHelper.ConfigureAwaitHideCancellation(pub.ShutdownAsync(new CancellationToken(hardStop)));
+                Assert.Equal(hardStop, isCancelled);
+                var expected = new HashSet<string>(Enumerable.Range(0, tasks.Length).Select(x => x.ToString()));
```
chrisdunelm: Done

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.csproj
```diff
@@ -22,9 +22,9 @@
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\Google.Cloud.Iam.V1\Google.Cloud.Iam.V1\Google.Cloud.Iam.V1.csproj" />
-    <PackageReference Include="ConfigureAwaitChecker.Analyzer" Version="1.0.0-beta4">
+    <!--<PackageReference Include="ConfigureAwaitChecker.Analyzer" Version="1.0.0-beta4">
       <PrivateAssets>All</PrivateAssets>
-    </PackageReference>
+    </PackageReference>--> <!-- TODO: Work out how to add this back -->
```
chrisdunelm: How would I suppress the warning? There doesn't appear to be an warning number than I can use in a `#pragma warning disable`?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
chrisdunelm: No, this name is not final. There should be a TODO about this. Added.
The pubsub team prefer `Publisher`, but we can't really use that...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
chrisdunelm: Yes, mostly agreed. Let's discuss how this can be improved.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+            /// <summary>
+            /// The clients to use within a <see cref="HiPerfPublisher"/>.
+            /// </summary>
+            public IList<PublisherClient> Clients { get; }
+
+            /// <summary>
+            /// The number of <see cref="PublisherClient"/> to create and use within a <see cref="HiPerfPublisher"/> instance.
+            /// If <c>null</c>, defaults to the CPU count on the machine this is being executed on.
+            /// </summary>
+            public int? ClientCount { get; }
```
chrisdunelm: The reason this is nullable, is because if the settings `Clients` is set, then having this set to *anything* doesn't make sense, as it won't be used.
If the way settings/clients is specified changes, then this will also change. Let's discuss.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+            /// </summary>
+            public BatchingSettings BatchingSettings { get; set; }
+
+            /// <summary>
+            /// The <see cref="IScheduler"/> to use.
+            /// If <c>null</c>, defaults to <see cref="SystemScheduler"/>. Usually only useful for testing.
+            /// </summary>
+            public IScheduler Scheduler { get; set; }
+
+            /// <summary>
```
chrisdunelm: No, this is slightly subtle, and needs better docs; or the feature removing completely.
Currently, `BatchingSettings` specifies the thresholds at which a batch is sent to the server.
However, if a batch *cannot* be sent to the server right now (because there are no available clients), then it continues filling the batch past the thresholds in `BatchingSettings`. These two following properties (`BatchMaxElementCount`, `BatchMaxByteCount`) specify the absolute maximum size it'll let a batch get to before queuing for sending.
As soon as a new batch can be sent (ie a client is available), then an 'over-threshold' batch will immediately be sent.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -0,0 +1,626 @@
...
+            /// The number of bytes currently queued.
+            /// </summary>
+            public long ByteCount { get; }
+        }
+
+        // All defaults taken from Java (reference) implementation.
+
+        /// <summary>
```
chrisdunelm: See above comment about `BatchingSettings` thresholds.
Yes, I agree that BatchingSettings would probably be better immutable, and the exact structure of this code could change.
But `BatchingSettings` includes a delay, which isn't relevant here. I dislike have irrelevant settings available.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
chrisdunelm: Done

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
chrisdunelm: They seem to have accepted it as an idiom we'd like to maintain.
I guess it may come up for discussion again during beta...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
chrisdunelm: Yes agreed. Personally I like the cleanness, and would like to keep this over having (say) another method for hard shutdown.
We can discuss.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
chrisdunelm: Correct. It's the number of bytes of memory that are required to locally queue the messages (ignoring all the other C# heap objs that are also required...).

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
chrisdunelm: yes.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
chrisdunelm: This is just the CancellationToken used to cancel the batch being sent on delay, if it's already been sent due to the byte/element-count.
I'm not sure how this should be changed?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.csproj
jskeet: There should be an ID that you can use - the pragma doesn't have to be numeric these days.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
jskeet: Will think on it. "BatchingTopicPublisher" would describe everythig, but be a bit ugly...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
jskeet: But can't those still be represented in a BatchingSettings? It's a sort of "absolute worst case batching settings"...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
jskeet: Ah, that's a fair point. Will ponder further.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
jskeet: Not the cancellation token part really - more that there's an async method which is "create batch and run it later". Possibly bad choice on my part of where to put the comment.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/Google.Cloud.PubSub.V1.Tests.csproj
```diff
@@ -40,6 +40,10 @@
     <Folder Include="Tasks\" />
   </ItemGroup>
 
+  <ItemGroup>
```
jskeet: I'm going to add this to the project generator, to stop it from coming up on random reviews...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/Tasks/TestAwaiter.cs
```diff
@@ -35,7 +35,16 @@ public TestAwaiter(Task task, TaskScheduler taskScheduler)
         public void UnsafeOnCompleted(Action continuation) =>
             _task.ContinueWith(_ => continuation(), CancellationToken.None, TaskContinuationOptions.DenyChildAttach, _taskScheduler);
         public bool IsCompleted => _task.IsCompleted;
-        public void GetResult() => _task.Wait();
+        public void GetResult()
```
jskeet: Is this now equivalent to our `_task.WaitWithUnwrappedExceptions()` method? (Slightly more efficient when an exception is thrown, admittedly - maybe we should change our implementation.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/Tasks/TestAwaiter.cs
```diff
@@ -53,7 +62,16 @@ public TestAwaiter(Task<T> task, TaskScheduler taskScheduler)
             _task.ContinueWith(_ => continuation(), CancellationToken.None, TaskContinuationOptions.DenyChildAttach, _taskScheduler);
         public void UnsafeOnCompleted(Action continuation) =>
             _task.ContinueWith(_ => continuation(), CancellationToken.None, TaskContinuationOptions.DenyChildAttach, _taskScheduler);
-         public bool IsCompleted => _task.IsCompleted;
-        public T GetResult() => _task.Result;
+        public bool IsCompleted => _task.IsCompleted;
+        public T GetResult()
```
jskeet: Ditto, ResultWithUnwrappedExceptions()?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -41,109 +47,117 @@ public abstract class HiPerfPublisher
...
             /// <summary>
-            /// The number of <see cref="PublisherClient"/> to create and use within a <see cref="HiPerfPublisher"/> instance.
-            /// If <c>null</c>, defaults to the CPU count on the machine this is being executed on.
+            /// <see cref="BatchingSettings"/> that control how messages are batched when a batch
+            /// cannot be immediately sent (because all clients are already busy sending batches).
+            /// <see cref="BatchingSettings.DelayThreshold"/> is not relevant in this context.
+            /// If <c>null</c>, defaults to <see cref="ApiMaxBatchingSettings"/>.
             /// </summary>
-            public int? ClientCount { get; }
+            public BatchingSettings MaxBatchingSettings { get; set; }
```
jskeet: QueuedBatchingSettings perhaps? (Definitely bike-shedding here - let's not worry about it for now. Just a thought for later.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -334,10 +327,10 @@ public FlowState(long elementCount, long byteCount)
         public virtual FlowState GetCurrentFlowState() => throw new NotImplementedException();
 
         /// <summary>
-        /// Shutdown this <see cref="HiPerfPublisher"/>.
+        /// Shutdown this <see cref="HiPerfPublisher"/>. Cancelling <paramref name="hardStopToken"/> aborts the
+        /// clean shutdown process, and will leave some locally queued messages unsent.
```
jskeet: will => may, to cover the case where actually it had already finished?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -383,32 +377,36 @@ public int AddMessage(PubsubMessage message, int byteCount)
...
             _shutdown = shutdown;
+            _taskHelper = GaxPreconditions.CheckNotNull(taskHelper, nameof(taskHelper));
 
-            // Initialise settings
-            var batchingSettings = settings.BatchingSettings ?? DefaultBatchingSettings();
-            _batchElementCountThreshold = batchingSettings.ElementCountThreshold ?? ApiMaxRequestElementCount;
-            _batchRequestByteThreshold = batchingSettings.RequestByteThreshold ?? ApiMaxRequestByteCount;
-            _batchDelayThreshold = batchingSettings.DelayThreshold ?? TimeSpan.FromHours(1);
+            // Initialise batching settings. Use ApiMax settings for components not given.
+            var batchingSettings = settings.BatchingSettings ?? DefaultBatchingSettings;
```
jskeet: We might want to consider a Merge method or similar...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -447,28 +445,31 @@ internal HiPerfPublisherImpl(TopicName topicName, Settings settings, Func<Task>
...
                 _queueByteCount += messageByteCount;
                 // Queue the current batch if this message would cause the batch to go over-byte-size
                 // (unless this would be the only message in the batch, then it's allowed).
-                QueueCurrentBatchIfRequired(false, messageByteCount);
+                QueueCurrentBatchIfRequired(messageByteCount);
                 if (_currentBatch == null)
                 {
                     // Create a new batch if this is the first ever batch, or a batch has just been queued.
-                    Task unusedTimeoutTask = CreateBatch();
+                    //Task unusedTimeoutTask = CreateBatch();
```
jskeet: Remove commented out code?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -509,98 +513,148 @@ private void ShutdownIfCompleted()
...
                 {
-                    // Force queuing of the current batch, whatever the size.
-                    // There will always be at least one message in the batch.
-                    QueueCurrentBatchIfRequired(true, 0);
-                }
+                    await _taskHelper.ConfigureAwait(_scheduler.Delay(batchDelayThreshold, timerCancellationToken));
+                    // If batch has already moved to queue, timerToken will have been cancelled.
+                    lock (_lock)
+                    {
+                        // Check for cancellation inside lock to avoid race-condition.
```
jskeet: It may be worth considering making the batch know whether or not it's been queued, and just unconditionally call some "QueueIfNotAlreadyQueued" here - and rather than using _currentBatch, either read _currentBatch before the task, or accept it as a method parameter.

No action required for now, but an alternative to consider for later.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -509,98 +513,148 @@ private void ShutdownIfCompleted()
...
         }
 
-        // Pre-condition: Must be locked
-        // Allow one message that is larger than requestByteThreshold
-        private bool CurrentBatchIsFull(bool isQueuing, int extraByteCount) => isQueuing ?
-                _currentBatch.Messages.Count >= _batchMaxElementCount || (_currentBatch.Messages.Count > 0 && _currentBatch.ByteCount >= _batchMaxByteCount) :
-                _currentBatch.Messages.Count >= _batchElementCountThreshold || (_currentBatch.Messages.Count > 0 && _currentBatch.ByteCount >= _batchRequestByteThreshold);
+        private void QueueCurrentBatch()
+        {
+            // Pre-condition: Must be locked
```
jskeet: I wonder whether we should add validation for these preconditions. I'm not sure how though... there's no Monitor.IsLocked :(

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfPublisher.cs
```diff
@@ -509,98 +513,148 @@ private void ShutdownIfCompleted()
...
+                return;
+            }
+            // Remove client and batch from relevant queues.
+            var client = _idleClients.Dequeue();
+            var batch = _batchesReady.Dequeue();
+            _batchesInFlightCount += 1;
+            // Update flow-control counts.
+            _queueElementCount -= batch.Messages.Count;
+            _queueByteCount -= batch.ByteCount;
+            _taskHelper.Run(async () =>
```
jskeet: Extract this code into a separate method? (Maybe a local method, maybe not - but it's a significant chunk in its own right.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/InGax.cs
```diff
@@ -47,36 +47,31 @@ public static long CheckArgumentRange(long argument, string paramName, long minI
...
+        /// <param name="requestByteThreshold">The byte count above which further processing of a batch will occur.</param>
+        /// <param name="delayThreshold">The batch lifetime above which further processing of a batch will occur.</param>
+        public BatchingSettings(long? elementCountThreshold, long? requestByteThreshold, TimeSpan? delayThreshold)
         {
-            ElementCountThreshold = other.ElementCountThreshold;
-            RequestByteThreshold = other.RequestByteThreshold;
-            DelayThreshold = other.DelayThreshold;
+            ElementCountThreshold = elementCountThreshold;
+            RequestByteThreshold = requestByteThreshold;
+            DelayThreshold = delayThreshold;
```
jskeet: Validation that nothing is negative? Presumably everything can be 0, meaning effectively "send immediately"? (Or maybe the bytes and element counts should be 1 or more...)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Tasks/TaskHelper.cs
```diff
@@ -102,7 +107,7 @@ public static TaskAwaitable<bool> ConfigureAwaitHideCancellation(this TaskHelper
                     await taskHelper.ConfigureAwait(task);
                     return false;
                 }
-                catch (AggregateException e) when (e.IsCancellation())
+                catch (Exception e) when (e.IsCancellation())
```
jskeet: Perhaps we should just catch OperationCanceledException instead?

Issue #1040, opened 2017-05-12
New MVC App + CloudTrace.Initialize() yields exception.


SurferJeffAtGoogle:
I created a new MVC web app.  Added the dependency Google.Cloud.Diagnostics.AspNet.  Then added a call to CloudTrace.Initialize(), and I get this exception when I try to run the application:

```
Could not load file or assembly 'System.Collections.Immutable, Version=1.2.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)
```

Here are the nuget packages pulled in by the app:
```
PM> get-package | select id, version

Id                                                 Version     
--                                                 -------     
Antlr                                              3.4.1.9004  
bootstrap                                          3.0.0       
Google.Api.CommonProtos                            1.0.0       
Google.Api.Gax                                     1.0.1       
Google.Api.Gax.Grpc                                1.0.1       
Google.Apis                                        1.24.1      
Google.Apis.Auth                                   1.24.1      
Google.Apis.Core                                   1.24.1      
Google.Cloud.Diagnostics.AspNet                    1.0.0-beta09
Google.Cloud.Diagnostics.Common                    1.0.0-beta01
Google.Cloud.ErrorReporting.V1Beta1                1.0.0-beta05
Google.Cloud.Logging.Type                          1.0.0       
Google.Cloud.Logging.V2                            1.0.0       
Google.Cloud.Trace.V1                              1.0.0-beta05
Google.Protobuf                                    3.2.0       
Grpc.Auth                                          1.2.2       
Grpc.Core                                          1.2.2       
jQuery                                             1.10.2      
jQuery.Validation                                  1.11.1      
Microsoft.AspNet.Mvc                               5.2.3       
Microsoft.AspNet.Razor                             3.2.3       
Microsoft.AspNet.Web.Optimization                  1.1.3       
Microsoft.AspNet.WebApi.Client                     5.2.3       
Microsoft.AspNet.WebApi.Core                       5.2.3       
Microsoft.AspNet.WebPages                          3.2.3       
Microsoft.CodeDom.Providers.DotNetCompilerPlatform 1.0.0       
Microsoft.jQuery.Unobtrusive.Validation            3.2.3       
Microsoft.Net.Compilers                            1.0.0       
Microsoft.Web.Infrastructure                       1.0.0       
[...]

PR #1152, opened 2017-06-29
Add local analyzers project with omitted argument checker

evildour:
(It is easiest to review this one commit at a time)

This adds a roslyn analyzers project to the repo which we can use to create custom analyzers specific to this repo or general analyzers not available elsewhere which we can take advantage of while trying to get them accepted into other repos (such as dotnet/roslyn-analyzers perhaps).

It also adds a single analyzer (suggested by @jskeet) for warning about omitted arguments if there is a local/parameter in scope which could have been used for it. Among other things, this will help with overloads calling off to other overloads but forgetting to pass along all arguments, like options or cancellation tokens.

Note: I had planned to have the analyzers project added into each .sln file with project dependencies set so each main project depended on the analyzers (note that the main projects would not reference the analyzers project though). That way, when working in any solution, rebuilding would cause the analyzers to get built before the main projects. However, there were a number of issues which made this difficult:
* The analyzers template generates normal MSBuild projects for the analyzers and tests. This would prevent 'dotnet build' from being able to build the solutions.
* I tried to fix the previous issue by changing them to the new style projects, but then the analyzer tests couldn't load `System.Collections.Immutable`. I think this might be related to the problem discovered in #1040, but unfortunately that fix didn't fix this issue.
* There seems to be bug in 'dotnet build' where it incorrectly assumes project dependencies are project references. So regardless of which style project was being used, 'dotnet build' would fail to build the netstandard1.3 versions of the main projects because it thought they were trying to reference a net452 project even though they weren't. I will look into this some more and report it in the dotnet/cli repo if it really is a bug.

So for now, we just have to remember to first open the Google.Cloud.Tools solution and build it if we want the main projects to take advantage of the analyzers. I will continue to research how best to fix this, if possible.


evildour:
Looks like the build.sh script isn't skipping over the analyzer project on Travis like it was on my local machine. Will look into it in the morning. All other files should be reviewable, but I understand if you'd like to hold off until I get things green.

jskeet:
Will try to get it today, but I have Spanner and PubSub reviews to do as well :)

build.sh
```diff
@@ -30,11 +30,20 @@ then
   for api in $apis
   do 
     [[ -d "$api" ]] && apidir=$api || apidir=apis/$api
-    dotnet sln AllProjects.sln add $apidir/*/*.csproj
-  
+    for project in $apidir/*/*.csproj
```
jskeet: Any reason we don't want to build the analyzers here? Aren't they as portable as we'd like?

tools/Google.Cloud.Tools.Analyzers.Tests/Google.Cloud.Tools.Analyzers.Tests.csproj
```diff
@@ -0,0 +1,183 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c)  Microsoft.  All Rights Reserved.  Licensed under the Apache License, Version 2.0.  See License.txt in the project root for license information. -->
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
jskeet: Any reason this is an "old-style" project instead of a new one? I much prefer the new format, and then we don't need to worry about nuget restore etc.

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/CodeFixVerifier.Helper.cs
```diff
@@ -0,0 +1,85 @@
+﻿using Microsoft.CodeAnalysis;
```
jskeet: Copyright statements.

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/CodeFixVerifier.Helper.cs
```diff
@@ -0,0 +1,85 @@
...
+namespace TestHelper
+{
+    /// <summary>
+    /// Diagnostic Producer class with extra methods dealing with applying codefixes
+    /// All methods are static
+    /// </summary>
+    public abstract partial class CodeFixVerifier : DiagnosticVerifier
+    {
+        /// <summary>
+        /// Apply the inputted CodeAction to the inputted document.
```
jskeet: Nit: I'd use "specified" rather than "inputted" here.

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/CodeFixVerifier.Helper.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// Meant to be used to apply codefixes.
+        /// </summary>
+        /// <param name="document">The Document to apply the fix on</param>
+        /// <param name="codeAction">A CodeAction that will be applied to the Document.</param>
+        /// <returns>A Document with the changes from the CodeAction</returns>
+        private static Document ApplyFix(Document document, CodeAction codeAction)
+        {
+            var operations = codeAction.GetOperationsAsync(CancellationToken.None).Result;
+            var solution = operations.OfType<ApplyChangesOperation>().Single().ChangedSolution;
```
jskeet: Doesn't this assume there's only a single change to make? Or does ApplyChangesOperation apply multiple changes?

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/DiagnosticResult.cs
```diff
@@ -0,0 +1,87 @@
...
+
+        public string Path { get; }
+        public int Line { get; }
+        public int Column { get; }
+    }
+
+    /// <summary>
+    /// Struct that stores information about a Diagnostic appearing in a source
+    /// </summary>
+    public struct DiagnosticResult
```
jskeet: Mutable structs (especially containing reference types) make me sad. Any reason this can't be a class?

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/DiagnosticResult.cs
```diff
@@ -0,0 +1,87 @@
...
+            }
+        }
+
+        public DiagnosticSeverity Severity { get; set; }
+
+        public string Id { get; set; }
+
+        public string Message { get; set; }
+
+        public string Path
```
jskeet: Use expression bodied properties to make these three much shorter.

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -0,0 +1,173 @@
...
+            {
+                throw new ArgumentException("Unsupported Language");
+            }
+
+            var project = CreateProject(sources, language);
+            var documents = project.Documents.ToArray();
+
+            if (sources.Length != documents.Length)
+            {
+                throw new SystemException("Amount of sources did not match amount of Documents created");
```
jskeet: Number rather than amount?

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -0,0 +1,173 @@
...
+
+            var solution = new AdhocWorkspace()
+                .CurrentSolution
+                .AddProject(projectId, TestProjectName, TestProjectName, language)
+                .WithProjectParseOptions(projectId,
+                    CSharpParseOptions.Default.WithFeatures(
+                        new[] { new KeyValuePair<string, string>("IOperation", "true") }))
+                .AddMetadataReference(projectId, CorlibReference)
+                .AddMetadataReference(projectId, SystemCoreReference)
+                .AddMetadataReference(projectId, CSharpSymbolsReference)
```
jskeet: Do our test projects usually need to look at Roslyn? That seems a little odd. Then again, I haven't looked at the tests yet...

tools/Google.Cloud.Tools.Analyzers.Tests/InternalOptionalParametersRequiredTest.cs
```diff
@@ -0,0 +1,144 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System;
+using TestHelper;
+using Xunit;
+
+namespace Google.Cloud.Tools.Analyzers.Test
+{
+    public class InternalOptionalParametersRequiredTest : CodeFixVerifier
```
jskeet: Add a test for when the value *is* specified correctly?

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/DiagnosticVerifier.Helper.cs
```diff
@@ -0,0 +1,173 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+
+namespace TestHelper
```
jskeet: I've just looked, and a lot of this code exists elsewhere... can we pull in an existing nuget package instead of repeating things? If not, we'll need to find out what the process is for including code from elsewhere in our own repo. Using a submodule may be a better approach.

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequired.cs
```diff
@@ -0,0 +1,110 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Semantics;
+using System;
+
+namespace Google.Cloud.Tools.Analyzers
+{
+    // Warns about omitted default arguments for calls to externally visible methods which are defined within the same assembly,
```
jskeet: Turns this into an XML comment?

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequired.cs
```diff
@@ -0,0 +1,110 @@
...
+
+                    var nameMatch = convertibleVariables.FirstOrDefault(
+                        symbol => symbol.Name.Equals(arg.Parameter.Name, StringComparison.OrdinalIgnoreCase));
+
+                    // If the parameter's type is String or any primitive type, only show the diagnostic when there is
+                    // some variable with the same (case-insentive) name as the argument. Strings are too commmon, for
+                    // example, to assume that any string in scope should be supplied for a default parameter.
+                    if (parameterType.SpecialType == SpecialType.System_String || parameterType.IsPrimitive())
+                    {
+                        if (nameMatch == null)
```
jskeet: Any reason for not just using && here, instead of a nested if?

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequired.cs
```diff
@@ -0,0 +1,110 @@
...
+                        {
+                            continue;
+                        }
+                    }
+
+                    // Report the missing argument with a suggested variable to be used. Prefer one with a name matching
+                    // the missing parameter's name if available.
+                    var preferrredVariable = nameMatch ?? convertibleVariables[0];
```
jskeet: Nit: remove an r

tools/Google.Cloud.Tools.Analyzers/SymbolExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+                case Accessibility.Public:
+                case Accessibility.Protected:
+                case Accessibility.ProtectedOrInternal:
+                    return symbol.ContainingType?.IsExternallyVisible() ?? true;
+                default:
+                    return false;
+            }
+        }
+
+        internal static bool IsPrimitive(this ITypeSymbol type)
```
jskeet: Just have a set or array of primitives and use Contains?

tools/Google.Cloud.Tools.ProjectGenerator/ApiMetadata.cs
```diff
@@ -130,7 +133,20 @@ private static void WriteProjectFile(string directory, XElement propertyGroup, X
                 doc = XElement.Load(file);
                 doc.Elements("Import").Where(x => (string)x.Attribute("Project") == @"..\..\StripDesktopOnNonWindows.xml").Remove();
                 doc.Elements("PropertyGroup").First().ReplaceWith(propertyGroup);
-                doc.Elements("ItemGroup").First().ReplaceWith(itemGroup);
+                doc.Elements("ItemGroup").First().ReplaceWith(dependenciesItemGroup);
```
jskeet: I think we can simplify this by putting the Exists condition on the Analyzer element instead of the ItemGroup element. Then we can still have a single ItemGroup.

evildour:
@jskeet Apologies, I forgot to mention in the notes at the top that the files in the Helpers and Verifies folders of the test project are from the analyzer template. I did not modify them except to change the `Assert` calls to use the xunit APIs. I wasn't sure if they should be reviewed and/or if the copyright notice should be placed in them. Seems like no for the latter.

jskeet:
Ah, right. That's good to know - let's not add copyright, but just an explanatory comment at the head of each file.

Given that we don't need to build a VSIX from this, I'd still really like to convert the project files to be new-style msbuild ones if we possibly can...

tools/Google.Cloud.Tools.Analyzers.Tests/Google.Cloud.Tools.Analyzers.Tests.csproj
evildour: Well as I said above, it caused the tests to fail to load System.Collections.Immutable, but I played around with it a bit more just now and was able to get things working by removing the binding redirects to the Roslyn libraries which are added by the analyzer template.

build.sh
evildour: It was not possible using the old style projects. Trying to add them caused an error. But now that the new style projects work, we can add them, so I've removed these changes.

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/CodeFixVerifier.Helper.cs
evildour: Added note at top that this came from the template.

tools/Google.Cloud.Tools.Analyzers.Tests/Google.Cloud.Tools.Analyzers.Tests.csproj
jskeet: Whoops - I should really read the PR description before reviewing :)

tools/Google.Cloud.Tools.Analyzers.Tests/Helpers/CodeFixVerifier.Helper.cs
evildour: (ignoring this and other feedback on the template files, but let me know if there's still anything you'd like changed)

tools/Google.Cloud.Tools.Analyzers.Tests/InternalOptionalParametersRequiredTest.cs
evildour: Done

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequired.cs
evildour: Oops, thought I had done that. Thanks.

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequired.cs
evildour: I usually try to avoid changing logical operators in `if` conditions, but this one isn't too bad so I probably would have used a single `if` here. I think I intended to have more logic in here at first but then realized I could just fall through to below and things would work out. I just didn't clean it up after changing course.

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequired.cs
evildour: Done

tools/Google.Cloud.Tools.Analyzers/SymbolExtensions.cs
evildour: Are you concerned about performance or just the code style? If the former, I don't think you can get much better than what's here. The compiler seems to create range checks on blocks of contiguous values:

```il
.method assembly hidebysig static bool  IsPrimitive(class [Microsoft.CodeAnalysis]Microsoft.CodeAnalysis.ITypeSymbol 'type') cil managed
{
  .custom instance void [mscorlib]System.Runtime.CompilerServices.ExtensionAttribute::.ctor() = ( 01 00 00 00 ) 
  // Code size       25 (0x19)
  .maxstack  2
  .locals init (valuetype [Microsoft.CodeAnalysis]Microsoft.CodeAnalysis.SpecialType V_0)
  IL_0000:  ldarg.0
  IL_0001:  callvirt   instance valuetype [Microsoft.CodeAnalysis]Microsoft.CodeAnalysis.SpecialType [Microsoft.CodeAnalysis]Microsoft.CodeAnalysis.ITypeSymbol::get_SpecialType()
  IL_0006:  stloc.0
  IL_0007:  ldloc.0
  IL_0008:  ldc.i4.7
  IL_0009:  sub
  IL_000a:  ldc.i4.s   9
  IL_000c:  ble.un.s   IL_0015
  IL_000e:  ldloc.0
  IL_000f:  ldc.i4.s   18
  IL_0011:  sub
  IL_0012:  ldc.i4.1
  IL_0013:  bgt.un.s   IL_0017
  IL_0015:  ldc.i4.1
  IL_0016:  ret
  IL_0017:  ldc.i4.0
  IL_0018:  ret
} // end of method SymbolExtensions::IsPrimitive
```

However, if it's just about style, I can change it.

jskeet: Oh, only style - it just seems easier to say "If it's one of these things". But I'm not massively bothered, either :)

tools/Google.Cloud.Tools.ProjectGenerator/ApiMetadata.cs
evildour: Good call. Done.

tools/Google.Cloud.Tools.Analyzers/SymbolExtensions.cs
evildour: Done

evildour:
I've been seeing issues trying to run build.sh locally with this PR. Multiple dialogs appear saying "dotnet has stopped working" and many messages appear in the console which say something like
```
C:\Program Files\dotnet\sdk\1.0.4\Roslyn\Microsoft.CSharp.Core.targets(71,5): error MSB6006: "C:\Program Files\dotnet\sdk\1.0.4\Roslyn\RunCsc.cmd" exited with code -2146232797. [C:\Users\mdour\git_projects\google-cloud-dotnet\apis\Google.Cloud.BigQuery.V2\Google.Cloud.BigQuery.V2\Google.Cloud.BigQuery.V2.csproj]
```
This was not happening when the analyzers were built with the old style project, so perhaps it is a problem running the analyzers built from the new style project. I will keep looking into this but for now I'm just going to push a commit which only enables the analyzers when building in the debug configuration.

evildour:
PTAL

tools/Google.Cloud.Tools.Analyzers/Google.Cloud.Tools.Analyzers.csproj
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Version>1.0.0</Version>
+    <TargetFrameworks>net452</TargetFrameworks>
```
jskeet: Once this is in, I might try updating to the 2.x Roslyn packages and target netstandard1.3... then we should be able to build on Linux with no problems.

evildour: Ok. Maybe that will fix these other build issues as well.

evildour:
Thanks Jon. The AppVeyor queue seems to be stuck. Should I wait for it to build before merging?

jskeet:
Nope, don't worry about AppVeyor, at least for now - it's on my list to have a look at what's wrong with it...

evildour:
(submitted issue for the 3rd bullet in the description above: https://github.com/dotnet/cli/issues/7075)

PR #1153, opened 2017-06-29
ProjectGenerator can not add ConfigureAwaitChecker

chrisdunelm:
And remove ConfigureAwaitChecker from pubsub; new pubsub code correctly doesn't use ConfigureAwait directly.


jskeet:
Let's leave this until we see whether there are alternatives. (When Mike's got all the new analyzers up, we may well want to be more configurable in general...)

PR #1154, opened 2017-06-30
Code fix for the omitted argument analyzer

evildour:
Unfortunately, the code fixes don't seem to be available with the new style projects. I believe it is this issue: https://github.com/dotnet/project-system/issues/285 since all symptoms of that bug can be seen in this repo's main projects. It is already fixed, so we just need to wait until that becomes available to take advantage of this code fix.

### DIFF
--- a/tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredAnalyzer.cs
+++ b/tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredAnalyzer.cs
@@ -86,13 +86,13 @@ namespace Google.Cloud.Tools.Analyzers
             }
         }
 
-        internal static List<ISymbol> GetVariablesInScope(SyntaxNode syntaxNode, SemanticModel semanticModel) =>
-            (from symbol in semanticModel.LookupSymbols(syntaxNode.SpanStart)
-             where symbol.Kind == SymbolKind.Local || symbol.Kind == SymbolKind.Parameter
-             select symbol).ToList();
+        internal static IEnumerable<ISymbol> GetVariablesInScope(SyntaxNode syntaxNode, SemanticModel semanticModel) =>
+            from symbol in semanticModel.LookupSymbols(syntaxNode.SpanStart)
+            where symbol.Kind == SymbolKind.Local || symbol.Kind == SymbolKind.Parameter
+            select symbol;
 
         internal static ISymbol TryGetVariableForArgument(
-            IParameterSymbol parameter, Compilation compilation, List<ISymbol> variablesInScope)
+            IParameterSymbol parameter, Compilation compilation, IEnumerable<ISymbol> variablesInScope)
         {
             var parameterType = parameter.Type;
             var convertibleVariables = variablesInScope.Where(
--- a/tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredAnalyzer.cs
+++ b/tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredAnalyzer.cs
@@ -12,14 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System.Collections.Immutable;
-using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Semantics;
 using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
 
 namespace Google.Cloud.Tools.Analyzers
 {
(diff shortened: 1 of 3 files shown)


tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredAnalyzer.cs
```diff
@@ -106,5 +85,38 @@ private static void AnalyzeInvocation(OperationAnalysisContext context)
                 }
             }
         }
+
+        internal static List<ISymbol> GetVariablesInScope(SyntaxNode syntaxNode, SemanticModel semanticModel) =>
```
jskeet: It may be worth removing the ToList call here, so you only materialize the query where you need to. For example, you don't need to in TryGetVariableForArgument. (You may not need to in the previous code either; I haven't checked.)

I know Roslyn analyzers are probably more performance sensitive than other code - I wouldn't normally mention this...

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredCodeFixProvider.cs
```diff
@@ -0,0 +1,115 @@
...
+            var root = await document.GetSyntaxRootAsync(context.CancellationToken).ConfigureAwait(false);
+
+            var argumentListExpression =
+                root.FindToken(context.Span.Start).Parent.AncestorsAndSelf().OfType<ArgumentListSyntax>().FirstOrDefault();
+            if (argumentListExpression == null)
+            {
+                return;
+            }
+
+            var invocation =
```
jskeet:
```csharp
if (!semanticModel.GetOperation(...) is IInvocationExpression invocation)
{
    return;
}
```
? (Honestly not sure which I'd go with here, but it's obviously an option...)

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredCodeFixProvider.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                switch (argument.ArgumentKind)
+                {
+                    case ArgumentKind.Positional:
+                        break;
+                    case ArgumentKind.Named:
+                        useNamedArguments = true;
+                        break;
+                    case ArgumentKind.DefaultValue:
+                        {
```
jskeet: I think I'd normally have all of this unindented, so the { lined up under the c. Was this how VS indented it? (Not really a problem of course.)

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredCodeFixProvider.cs
```diff
@@ -0,0 +1,115 @@
...
+                semanticModel.GetOperation(argumentListExpression, context.CancellationToken) as IInvocationExpression;
+            if (invocation == null)
+            {
+                return;
+            }
+
+            var variablesInScope =
+                InternalOptionalParametersRequiredAnalyzer.GetVariablesInScope(argumentListExpression, semanticModel);
+
+            var omittedParameterVariablePairs = new List<Tuple<IParameterSymbol, ISymbol>>();
```
jskeet: It looks like we only actually use the names... can we use a `Tuple<string, string>` instead?

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredAnalyzer.cs
evildour: Well that just means that if there are multiple omitted arguments per call, we would need to re-iterate over all symbols in scope to find just locals or parameters, but for the common case or 0 or 1 omitted arguments, we would perform the iteration 0 or 1 times, respectively, so yes overall this is much better.

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredCodeFixProvider.cs
evildour: Done

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredCodeFixProvider.cs
evildour: Done

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredAnalyzer.cs
jskeet: But the caller can call ToList if they want to. There's no point in calling ToList when you're then going to pass the result into Where and then call ToList on *that*, which is what you do elsewhere.

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredCodeFixProvider.cs
evildour: Done. Yes, this is how VS formatted it and unfortunately I don't see a formatting option to control this.

evildour:
Thanks! Will merge when green

tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredAnalyzer.cs
evildour: Yes, I agree. My only point was that if there are multiple omitted arguments, we're going to end up calling `TryGetVariableForArgument` multiple times for the same enumerable returned from `GetVariablesInScope`. Each of those calls will iterate over the same large list of symbols rather than the much smaller filtered list. I think it is better this way, I'm just saying that there's a trade off for a small number of cases.

PR #1157, opened 2017-06-30
More version-friendly Operation<,>.

jskeet:
If the API looks right, I'll add tests - so the review at the moment is "does this feel right".

Also raised https://github.com/google/protobuf/issues/3294 to improve Google.Protobuf.

### DIFF
--- a/apis/Google.LongRunning/Google.LongRunning/Operation.cs
+++ b/apis/Google.LongRunning/Google.LongRunning/Operation.cs
@@ -15,23 +15,47 @@
 using Google.Api.Gax;
 using Google.Api.Gax.Grpc;
 using Google.Protobuf;
+using Google.Protobuf.WellKnownTypes;
 using System;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Google.LongRunning
 {
+    // TODO: Update the class documentation with alternatives for handling multiple types when
+    // https://github.com/google/protobuf/issues/3294 is resolved.
+
     /// <summary>
     /// A long-running operation with an associated client, and which knows the expected response type.
     /// </summary>
     /// <remarks>
+    /// <para>
     /// For simplicity, no methods on this type modify the proto message. Instead, to get up-to-date
     /// information you can use Refresh to obtain a new instance.
+    /// </para>
+    /// <para>
+    /// If the operation was created with a different major version of the service API than expected,
+    /// the metadata and response values may not be of the expected type. There are three approaches to handling this:
+    /// <list type="bullet">
+    ///   <item>
+    ///     <description>To fail with an exception if an unexpected type of value is present, use the <see cref="Result"/>
+    ///     and <see cref="Metadata"/> properties.</description>
+    ///   </item>
+    ///   <item>
+    ///     <description>To receive a null reference if an unexpected type of value is present, use the <see cref="GetResultOrNull()"/>
+    ///     and <see cref="GetMetadataOrNull()"/> methods. You can then check the returned value and ignore nulls.</description>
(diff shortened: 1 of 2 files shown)


apis/Google.LongRunning/Google.LongRunning/Operation.cs
```diff
@@ -105,17 +129,37 @@ public Operation(Operation rpcMessage, OperationsClient client)
         /// Retrieves the metadata associated with this operation, or <c>null</c> if there is no
         /// metadata in the underlying response message.
         /// </summary>
-        public TMetadata Metadata => RpcMessage.Metadata?.Unpack<TMetadata>();
+        /// <remarks>
+        /// Only the in-memory representation of the operation (this object) is consulted for its state.
+        /// See the <see cref="Operation"/> documentation for information about dealing with different metadata type versions.
+        /// </remarks>
+        /// <exception cref="InvalidOperationException">Metadata is present, but is not of the expected type.</exception>
```
chrisdunelm: Use a custom exception `IncompatibleMetadataException` (or something), for clarity?

apis/Google.LongRunning/Google.LongRunning/Operation.cs
```diff
@@ -105,17 +129,37 @@ public Operation(Operation rpcMessage, OperationsClient client)
...
+        /// If the operation has completed but the result is not present (for example due to being excluded by
+        /// a field mask) this returns null.
+        /// </summary>
+        /// <remarks>
+        /// Only the in-memory representation of the operation (this object) is consulted for its state.
+        /// See the <see cref="Operation"/> documentation for information about dealing with different response type versions.
         /// </remarks>
         /// <exception cref="OperationFailedException">The operation completed with an error.</exception>
-        /// <exception cref="InvalidOperationException">The operation has not completed yet.</exception>
+        /// <exception cref="InvalidOperationException">The operation has not completed yet, or the result is present but
```
chrisdunelm: Return different Exceptions depending on which condition failed? (Related to above comment)

jskeet: It feels a bit overkilly to me - especially as it's easy enough to check whether or not it *has* completed, if you need to. (You could even check it in a guard clause if you really wanted.)

InvalidOperationException really does feel like the most appropriate one...

jskeet:
Tests are now present, but I had to change the implementation as well. If the result has been masked out, we don't really know whether it completed successfully or whether it was an error that was masked out. So having the Done property *is* a useful thing.

@chrisdunelm if you could review today or Monday that would be great (before I lose context) but I may wait for internal discussions to conclude before merging.

apis/Google.LongRunning/Google.LongRunning/Operation.cs
```diff
@@ -105,34 +126,74 @@ public Operation(Operation rpcMessage, OperationsClient client)
...
             get
             {
-                switch (RpcMessage.ResultCase)
+                // This uses the Done property...
+                if (!IsCompleted)
+                {
+                    throw new InvalidOperationException("Operation has not completed.");
+                }
+
+                // If we have no result case, that's could be because there was an error but the fieldmask
```
chrisdunelm: "that", not "that's"

jskeet: Doh, thanks. Will fix in a minute.

jskeet:
Discussion concluded this is appropriate - we may make Any support even better in protobuf, which could have knock-on effects here later (e.g. fetch `object GetResult(TypeRegistry)`).

PR #1162, opened 2017-07-03
Regenerate APIs and add DLP API

jskeet:
All auto-generated apart from the snippets and the gRPC partial


jskeet:
(Will self-merge when green.)

PR #1163, opened 2017-07-03
Use Roslyn v2 for analyzers

jskeet:
Tested by including failing code, and then removing that commit. This way we get analyzers in CI...

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.csproj
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.csproj
@@ -27,7 +27,7 @@
     </PackageReference>
     <PackageReference Include="Google.Api.Gax.Rest" Version="2.0.0" />
     <PackageReference Include="Google.Apis.Bigquery.v2" Version="1.27.1.885" />
-    <Analyzer Condition="Exists('..\..\..\tools\Google.Cloud.Tools.Analyzers\bin\$(Configuration)\net452\Google.Cloud.Tools.Analyzers.dll') AND $(Configuration) == 'Debug'" Include="..\..\..\tools\Google.Cloud.Tools.Analyzers\bin\$(Configuration)\net452\Google.Cloud.Tools.Analyzers.dll" />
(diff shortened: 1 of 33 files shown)


jskeet:
We have to target Roslyn 2.0 as that's the version that the dotnet CLI we're using also uses. Trying again...

jskeet:
(Using the publish directory may well help in terms of being able to rebuild the tools solution with VS open, btw. It was originally to get around a different problem, but that's a side benefit.)

evildour:
How would the normal development process work now? Using the publish directory means just building the tools solution now is no longer sufficient to get the analyzers to be used when working in one of the projects. Do we have to remember to publish the appropriate build if we want to use them locally?

jskeet:
Yes, I think so - which we can easily put in a shell script, of course. But if we don't update them very often, it's a mostly "one off" build...

evildour:
Ok, yeah I guess if anyone does forget and something slips through, at least the CI build should catch it. Also, I tried using these changes to see if I could go back to including the analyzers in each solution with build dependencies, but the issue about 'dotnet build' assuming build dependencies are project dependencies still occurs, but now it is the net45 target which can't build because it is 'depending' on the netstandard1.3 analyzers.

LGTM.

jskeet:
Thanks for that. I'm definitely keener on the CI catching things but devs maybe not than the other way round...

evildour:
Had another thought about this: we could have the project generator create two analyzer nodes pointing to the same dll - one conditioned on the release build and pointing to the publish dir and the other conditioned on the debug build not pointing to the publish dir. Once the build dependency issue is sorted out, then I think things could work well for both scenarios.

jskeet:
We could do that... but I fear we'll get into the locking situation again. Let's see how we go. I might ask the Roslyn team if they have any recommendations for this sort of thing...

evildour:
What was the locking situation? I don't think I saw any issues with VS open.

jskeet:
I thought you had to reboot because VS had something locked while a build was trying to replace it, or similar?

evildour:
Oh right. I don't remember exactly what the issue was because I was dealing with a lot of various problems there, but for that I do remember two things about that issue: 1) it was happening during a command line build and 2) a reboot didn't help. I think some other error in the build and/or running analyzers process earlier was causing files to get locked later.

PR #1167, opened 2017-07-04
Hi-perf subscriber, and unit tests

chrisdunelm:


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+        private TaskCompletionSource<int> _mainTcs;
+        private CancellationTokenSource _globalSoftStopCts;
+        private CancellationTokenSource _globalHardStopCts;
+
+        /// <inheritdoc />
+        public override Task StartAsync(Func<PubsubMessage, CancellationToken, Task<Reply>> handlerAsync)
+        {
+            lock (_lock)
+            {
+                GaxPreconditions.CheckState(_mainTcs == null, "Can only start an instance once.");
```
jskeet: Gosh, you can't even restart it after shutting it down? That's a bit surprising, particularly if it's creating a load of clients. Maybe it fits in with the common use case though.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+{
+    /// <summary>
+    /// A PubSub subscriber that is associated with a specific <see cref="SubscriptionName"/>.
+    /// </summary>
+    /// <remarks>
+    /// <para>To receive messages, the <see cref="StartAsync"/> method must be called,
+    /// with a suitable message handler.</para>
+    /// <para>The message handler <see cref="Reply"/> states whether to acknowledge the message;
+    /// if acknowledged then (under normal conditions) it will not be received on this subscription again.</para>
+    /// <para>But note that it is always possible to receive duplicate messages. This services
```
jskeet: I wonder whether we're basically encouraging each user to build their own backlog of handled message IDs so that they can dedupe... could we do that instead?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                GaxPreconditions.CheckState(!_globalSoftStopCts.IsCancellationRequested, "Can only stop an instance once.");
+                _globalSoftStopCts.Cancel();
+            }
+            var registration = hardStopToken.Register(() => _globalHardStopCts.Cancel());
+            // Do not register this Task, it completes *after* mainTask
+            _taskHelper.Run(async () => await _taskHelper.ConfigureAwaitWithFinally(_mainTcs.Task, () => registration.Dispose()));
+            return _mainTcs.Task;
+        }
+
+        private class Flow
```
jskeet: A doc comment here would be very useful. I don't really know what this means yet...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+        public override Task StopAsync(CancellationToken hardStopToken)
+        {
+            lock (_lock)
+            {
+                GaxPreconditions.CheckState(_mainTcs != null, "Can only stop a started instance.");
+                GaxPreconditions.CheckState(!_globalSoftStopCts.IsCancellationRequested, "Can only stop an instance once.");
+                _globalSoftStopCts.Cancel();
+            }
+            var registration = hardStopToken.Register(() => _globalHardStopCts.Cancel());
+            // Do not register this Task, it completes *after* mainTask
```
jskeet: I don't really understand this comment, to be honest...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+            private readonly long _maxElementCount;
+            private readonly Action<Task> _registerTaskFn;
+            private readonly TaskHelper _taskHelper;
+
+            private long _byteCount;
+            private long _elementCount;
+
+            private TaskCompletionSource<int> _event = new TaskCompletionSource<int>();
+
+            private bool IsFlowOk() => _byteCount < _maxByteCount && _elementCount < _maxElementCount;
```
jskeet: Feels more like a property than a method...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+
+            public async Task Process(long byteCount, Func<Task> fn)
```
jskeet: Process what? Could you add a doc comment here?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+            private readonly object _lock = new object();
+            private readonly long _maxByteCount;
+            private readonly long _maxElementCount;
+            private readonly Action<Task> _registerTaskFn;
+            private readonly TaskHelper _taskHelper;
+
+            private long _byteCount;
+            private long _elementCount;
+
+            private TaskCompletionSource<int> _event = new TaskCompletionSource<int>();
```
jskeet: What's this for?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                            _byteCount += byteCount;
+                            _elementCount += 1;
+                            break;
+                        }
+                        eventTask = _event.Task;
+                    }
+                    prevEventTask = eventTask;
+                    await _taskHelper.ConfigureAwait(eventTask);
+                }
+                _registerTaskFn(_taskHelper.Run(async () => await _taskHelper.ConfigureAwaitWithFinally(fn(), () =>
```
jskeet: Comments definitely needed here - I've no idea what this is doing...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                            ev.SetResult(0);
+                        }
+                    }
+                })));
+            }
+        }
+
+        private class SingleChannel
+        {
+            public SingleChannel(SubscriptionName subscriptionName,
```
jskeet: If this took a HiPerfSubscriberImpl as a parameter, it could get to a lot of these from that (as it's a nested class). Or we could move them into a separate class.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                            var ev = _event;
+                            _event = new TaskCompletionSource<int>();
+                            ev.SetResult(0);
+                        }
+                    }
+                })));
+            }
+        }
+
+        private class SingleChannel
```
jskeet: Doc comment please :)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                        pull = _client.StreamingPull(CallSettings.FromCancellationToken(streamingPullCts.Token),
+                            new BidirectionalStreamingSettings(10));// TODO: Put back to 1 once bug #188 released in v2.1.0
+                        // Initial call to start subscribe messages arriving.
+                        await _taskHelper.ConfigureAwait(pull.WriteAsync(new StreamingPullRequest
+                        {
+                            SubscriptionAsSubscriptionName = _subscriptionName,
+                            StreamAckDeadlineSeconds = _modifyDeadlineSeconds
+                        }));
+                        // Start Push task which pushes acks and extends.
+                        var pushTask = _taskHelper.Run(() => Push(pull, streamingPullCts.Token));
```
jskeet: Maybe streamingPullCts should be renamed given that it's used for pushing as well?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                        // Read incoming messages whilst not stopped, and not in a fault condition.
+                        var pullTask = _taskHelper.Run(() => Pull(pull, streamingPullCts.Token));
+                        // Wait for either pull or push to complete.
+                        var completedTask = await _taskHelper.ConfigureAwait(_taskHelper.WhenAny(pushTask, pullTask));
+                        // Only cancel streaming if one of the tasks faulted. Otherwise use the non-abortive shutdown.
+                        if (completedTask.IsFaulted)
```
jskeet: What if one finishes normally and the other finished with an exception, both at pretty much the same time? There's a race condition as to which we see as completedTask here. Does that matter?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                            _registerTaskFn(_taskHelper.Run(() => pull.WriteCompleteAsync()));
+                        }
+                    }
+                }
+            }
+
+            private async Task Pull(SubscriberClient.StreamingPullStream pull, CancellationToken streamingPullToken)
+            {
+                var cts = CancellationTokenSource.CreateLinkedTokenSource(streamingPullToken, _softStopCts.Token);
+                while (!cts.IsCancellationRequested && await _taskHelper.ConfigureAwait(pull.ResponseStream.MoveNext(CancellationToken.None)))
```
jskeet: Why are we using CancellationToken.None here? Doesn't that mean we could be blocked waiting for the next item? Or is the point that if something gets cancelled, the response stream will finish?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                {
+                    // If in soft-stop, the overall StreamingPull might not yet be cancelled, but don't process any more msgs.
+                    if (cts.IsCancellationRequested)
```
jskeet: This looks very odd as the first thing in a loop which has a condition that checks that it's *not* requested. Did you do half a refactor here?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                            break;
+                        }
+                        // Returned Task completes once flow-control is satisfied.
+                        await _taskHelper.ConfigureAwait(_flow.Process(msg.CalculateSize(), async () =>
+                        {
+                            if (cts.IsCancellationRequested)
+                            {
+                                return;
+                            }
+                            lock (_qLock)
```
jskeet: Any reason for using a lock for this instead of Interlocked.Increment / Interlocked.Decrement?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                            {
+                                _unAckedMsgCount += 1;
+                            }
+                            var reply = await _taskHelper.ConfigureAwaitHideErrors(_handlerAsync(msg.Message, _hardStopCts.Token), Reply.Nack);
```
jskeet: So this is the code that calls the user-supplied callback, right?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                            TaskCompletionSource<int> qEvent = null;
+                            lock (_qLock)
+                            {
+                                if (reply == Reply.Ack)
```
jskeet: Are we okay with users returning `(Reply) 100` and it being treated as a nack?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+                                        allMsgsHandledCts.Cancel();
+                                    }
+                                    qEvent = _qEvent;
+                                }
+                                else
+                                {
+                                    _unAckedMsgCount -= 1;
+                                }
+                            }
+                            if (qEvent != null)
```
jskeet: This is to unblock the flow, if I'm understanding this correctly?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+            private async Task AutoExtend(HashSet<string> ackIds, CancellationToken allMsgsHandledToken)
+            {
+                // Moves ackIds into the extend-queue as required for auto lease extension
+                while (true)
+                {
+                    await _taskHelper.ConfigureAwait(_scheduler.Delay(_autoExtendInterval, allMsgsHandledToken));
+                    TaskCompletionSource<int> qEvent;
+                    lock (_qLock)
+                    {
+                        _extendQueue.EnqueueAll(ackIds);
```
jskeet: Okay, so this takes a snapshot of the ack IDs that hadn't been processed when the timer went off... which may still be processed, but that's okay?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Done

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: comment added about this.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: The race condition doesn't matter. Comment added.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Comment added.
This won't accept anything else :(
This caused me quite a bit of pain, as I previously got the whole thing working by using a real CancellationToken here, when running against the mock.
The mock subscriber used in tests now emulates the real behaviour.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: This is correct. The comment explains why it's required. It's because we can't cancel the MoveNext when we go into soft-stop, but if it pulls some more messages we don't want to process any more.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Mostly for consistency/sanity. All access to various fields is done under this lock. I don't think there's much advantage in  mix-n-matching.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Yes. Comment added.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Yes. I don't think any other behaviour makes sense.
The only extra thing to do would be to log it, but logging is still TODO.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: To unblock the Pusher, now that there's an ACK to push. Comment added.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Yes, it's OK. Comment added.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/Tasks/TestScheduler.cs
```diff
@@ -334,10 +334,47 @@ public override void Wait(Task task)
...
                     await this.ConfigureAwaitHideErrors(task);
+                    switch (task.Status)
+                    {
+                        case TaskStatus.RanToCompletion:
+                            break;
+                        case TaskStatus.Faulted:
+                            exceptions.AddRange(task.Exception.InnerExceptions);
+                            break;
+                        case TaskStatus.Canceled:
+                            cancelled = true;
```
benwulfe: this will loop and await every task even if the first cancels which appears to result in the same result (a TaskCanceledException).
Seems it might be better to do a WaitAny, loop thru the completed/final tasks removing them from the list, bail if canceled, and then continue with the rest until empty...?

chrisdunelm: This needs to match the specification of [Task.WhenAll()](https://msdn.microsoft.com/en-us/library/hh160384(v=vs.110).aspx), so the returned Task must not complete until all tasks in the argument have completed, whatever the status of their completion (ie even if some cancel/fault, this method must still wait for all tasks to complete).

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -94,7 +94,8 @@ internal Settings(Settings other)
             public TimeSpan? StreamAckDeadline { get; set; }
 
             /// <summary>
-            /// Duration before message expiration.
+            /// Duration before <see cref="StreamAckDeadline"/> at which the message ACK deadline
+            /// is automatically extended.
```
jskeet: Ah, so if this is 5 seconds and the ack deadline is 20 seconds, then in 15 seconds we extend it, yes? (So this is to give enough time for the extension to actually happen.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -0,0 +1,723 @@
...
+            var registeredTasks = new HashSet<Task>();
+            void RegisterTask(Task task)
+            {
+                registeredTasks.Locked(() => registeredTasks.Add(task));
+                void UnregisterTask() => registeredTasks.Locked(() => registeredTasks.Remove(task));
+                _taskHelper.Run(async () => await _taskHelper.ConfigureAwaitWithFinally(task, UnregisterTask));
+            }
+            _globalSoftStopCts = new CancellationTokenSource();
+            _globalHardStopCts = new CancellationTokenSource();
+            Flow flow = new Flow(_flowControlSettings.MaxOutstandingByteCount ?? long.MaxValue,
```
jskeet: Not massively strongly. It's far from the most complex bit of this code :)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -424,11 +428,15 @@ public override Task StopAsync(CancellationToken hardStopToken)
                 _globalSoftStopCts.Cancel();
             }
             var registration = hardStopToken.Register(() => _globalHardStopCts.Cancel());
-            // Do not register this Task, it completes *after* mainTask
+            // Do not register this Task, it completes *after* _mainTcs, and all registered tasks must complete before _mainTcs
```
jskeet: I guess what confuses me is what you mean by "register this task".

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
jskeet: Doesn't the same apply for (say) `List<T>.Count` or `StringBuilder.Length`?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -462,55 +485,70 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
                     {
                         if (IsFlowOk())
                         {
+                            // Flow is OK, so add for this element, and execute.
                             _byteCount += byteCount;
                             _elementCount += 1;
                             break;
                         }
+                        // FLow not OK, prepare to wait until a previous fn completes.
```
jskeet: Nit: FLow => Flow

(But these comments are making it *so* much easier to understand...)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
```diff
@@ -462,55 +485,70 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+            public SingleChannel(HiPerfSubscriberImpl subscriber,
                 SubscriberClient client, Func<PubsubMessage, CancellationToken, Task<Reply>> handlerAsync,
-                TimeSpan autoExtendInterval, int modifyDeadlineSeconds, int maxAckExtendCount, Flow flow,
-                Action<Task> registerTaskFn, IScheduler scheduler, CancellationTokenSource globalSoftStopCts,
-                CancellationTokenSource globalHardStopCts, TaskHelper taskHelper)
+                Flow flow,
+                Action<Task> registerTaskFn)
             {
-                _subscriptionName = subscriptionName;
+                _subscriptionName = subscriber.SubscriptionName;
```
jskeet: I was more meaning getting rid of the fields entirely (they're all immutable, right?) but it's all an implementation detail, so we can shift later. This is a good start in terms of massively reducing the parameter count.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
jskeet: Yikes. That's a gRPC limitation? We should try to get it fixed...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Correct.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Added further clarification in the comment.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: Done

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/HiPerfSubscriber.cs
chrisdunelm: I'd rather keep the fields. Otherwise all access has to indirect through a class/struct, which will make the code slightly messier.
Anyway, as you say, this can be changed later.

jskeet:
I've not looked at the tests in detail, but LGTM in general. (Obviously rebase to fix up the commit history...)

PR #1173, opened 2017-07-05
Comprehensive Type conversion unit tests.

benwulfe:
Introduces a unit test project for spanner ado.net
Adds the first set of unit tests which test storing and retrieving types.
Runtime changes (Type conversion changes):
  Fixes char conversions so it isnt (incorrectly) treated as a long and has its own handling.
  Added custom handling of decimal and ulong which cannot leverage the long codepath
  Fixes sbyte so that it leverages the typeof(long) codepath
  Ensures datetime+timestamp values can roundtrip as strings
  Fixed a bug with structs where the field member wasn't read correctly
  Allows conversion to bool of various types (convert allows it)
  Allows you to call GetFieldValue<List<X>>() where X is the list item type which recursively does conversion.
Also has 2 other typos/nonfunctional changes in PriorityList + ReliableStreamReader.
  ditto with GetFieldValue<X[]>  (typed arrays)
  ditto with Dictionary<, X> (the first generic argument must accept a string)

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/TypeTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/TypeTests.cs
@@ -1,0 +1,623 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using Google.Protobuf;
+using Google.Protobuf.WellKnownTypes;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    //more exhaustive (compared to integration tests) type conversion test cases.
+    public class TypeTests
+    {
+        public enum TestType
+        {
+            //Only test the CLR type -> Value conversion
+            ClrToValue = 0,
+
+            //Only test the Value -> CLR Type conversion
+            ValueToClr = 1,
+
+            //Test both ways.  There can be no loss of information or precision in the ClrToValue conversion.
+            Both = 2
+        }
+
+        private static readonly DateTime s_testDate = new DateTime(2017, 1, 31, 3, 15, 30, 500);
+        private static readonly byte[] s_bytesToEncode = {1, 2, 3, 4};
+        private static readonly string s_base64Encoded = Convert.ToBase64String(s_bytesToEncode);
+
+        private static readonly SpannerDbType s_struct = SpannerDbType.StructOf(
+            new Dictionary<string, SpannerDbType>
+            {
+                {"StringField", SpannerDbType.String},
+                {"Int64Field", SpannerDbType.Int64},
+                {"Float64Field", SpannerDbType.Float64},
+                {"BoolField", SpannerDbType.Bool},
+                {"DateField", SpannerDbType.Date},
+                {"TimestampField", SpannerDbType.Timestamp}
+            });
+
+        //nested complex type support.
+        private static readonly SpannerDbType s_arrayOfStruct = SpannerDbType.ArrayOf(s_struct);
+
+        private static readonly SpannerDbType s_complexStruct = SpannerDbType.StructOf(
+            new Dictionary<string, SpannerDbType>
+            {
+                {"StructField", s_struct},
+                {"ArrayField", SpannerDbType.ArrayOf(SpannerDbType.Int64)}
+            });
+
+        private static string Quote(string s) => $"\"{s}\"";
+
+        private static IEnumerable<string> GetStringsForArray()
+        {
+            yield return "abc";
+            yield return "123";
+            yield return "def";
+        }
+
+        private static IEnumerable<int> GetIntsForArray()
+        {
+            yield return 4;
+            yield return 5;
+            yield return 6;
+        }
+
+        private static IEnumerable<double> GetFloatsForArray()
+        {
+            yield return 1.0;
+            yield return 2.0;
+            yield return 3.0;
+        }
+
+        private static IEnumerable<bool> GetBoolsForArray()
+        {
+            yield return true;
+            yield return false;
+            yield return true;
+        }
+
+        private static IEnumerable<DateTime> GetDatesForArray()
+        {
+            yield return new DateTime(2017, 1, 31);
+            yield return new DateTime(2016, 2, 15);
+            yield return new DateTime(2015, 3, 31);
+        }
+
+        private static IEnumerable<DateTime> GetTimestampsForArray()
+        {
+            yield return new DateTime(2017, 1, 31, 3, 15, 30, 500);
+            yield return new DateTime(2016, 2, 15, 13, 15, 30, 000);
+            yield return new DateTime(2015, 3, 31, 3, 15, 30, 250);
+        }
+
+        // This data serves as inputs to converting from CLR types that a developer
+        // would use in his app to a given json serialized format using the spanner type
+        // to determine how to serialize.
+        // The inputs are used to test both ways (also deseralizing the generated json
+        // to a requested clr type).  However some cases are specified as only one direction
+        // usually because the conversion is by definition lossy.
+        // TODO(benwu):local specific tests (set current culture to ensure conversions use
+        // invariant cultures.
+        public static IEnumerable<object[]> GetValidValueConversions()
+        {
+            //format is:  LocalClrInstance,  SpannerType,  SerializedJsonFromProto, [test one or both ways]
+            //testing can be one way if there is loss of information in the conversion.
+
+            //Spanner type = Float64 tests.
+            yield return new object[] {true, SpannerDbType.Float64, "1"};
+            yield return new object[] {false, SpannerDbType.Float64, "0"};
+            yield return new object[] {(byte) 1, SpannerDbType.Float64, "1"};
+            yield return new object[] {(sbyte) 1, SpannerDbType.Float64, "1"};
+            yield return new object[] {1.5M, SpannerDbType.Float64, "1.5"};
+            yield return new object[] {1.5D, SpannerDbType.Float64, "1.5"};
+            yield return new object[] {1.5F, SpannerDbType.Float64, "1.5"};
+            yield return new object[] {double.NegativeInfinity, SpannerDbType.Float64, Quote("-Infinity")};
+            yield return new object[] {double.PositiveInfinity, SpannerDbType.Float64, Quote("Infinity")};
+            yield return new object[] {double.NaN, SpannerDbType.Float64, Quote("NaN")};
+            yield return new object[] {1, SpannerDbType.Float64, "1"};
+            yield return new object[] {1U, SpannerDbType.Float64, "1"};
+            yield return new object[] {1L, SpannerDbType.Float64, "1"};
+            yield return new object[] {(ulong) 1, SpannerDbType.Float64, "1"};
+            yield return new object[] {(short) 1, SpannerDbType.Float64, "1"};
+            yield return new object[] {(ushort) 1, SpannerDbType.Float64, "1"};
+            yield return new object[] {"1", SpannerDbType.Float64, "1"};
+            yield return new object[] {"1.5", SpannerDbType.Float64, "1.5"};
+            yield return new object[] {null, SpannerDbType.Float64, "null"};
+
+            //Spanner type = Int64 tests.
+            yield return new object[] {true, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {false, SpannerDbType.Int64, Quote("0")};
+            yield return new object[] {(char) 1, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {(byte) 1, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {(sbyte) 1, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {2M, SpannerDbType.Int64, Quote("2")};
+            yield return new object[] {1.5M, SpannerDbType.Int64, Quote("2"), TestType.ClrToValue};
+            yield return new object[] {2D, SpannerDbType.Int64, Quote("2")};
+            yield return new object[] {1.5D, SpannerDbType.Int64, Quote("2"), TestType.ClrToValue};
+            yield return new object[] {2F, SpannerDbType.Int64, Quote("2")};
+            yield return new object[] {1.5F, SpannerDbType.Int64, Quote("2"), TestType.ClrToValue};
+            yield return new object[] {1, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {1U, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {1L, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {(ulong) 1, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {(short) 1, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {(ushort) 1, SpannerDbType.Int64, Quote("1")};
+            yield return new object[] {"1", SpannerDbType.Int64, Quote("1")};
+
+            //Spanner type = Bool tests.
+            yield return new object[] {true, SpannerDbType.Bool, "true"};
+            yield return new object[] {false, SpannerDbType.Bool, "false"};
+            yield return new object[] {(byte) 1, SpannerDbType.Bool, "true"};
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/Google.Cloud.Spanner.Data.Tests.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
benwulfe: +1 -- done in the other PR

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -88,6 +90,34 @@ private SpannerDbType(TypeCode typeCode, RepeatedField<StructType.Types.Field> s
             }
         }
 
+        /// <inheritdoc />
+        public override string ToString()
```
benwulfe: yes, for logs and debugger use.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ValueConversion.cs
```diff
@@ -76,21 +76,16 @@ public static object ConvertToClrType(this Value wireValue, V1.Type spannerType,
                 return Convert.ToUInt16((long) ConvertToClrTypeImpl(wireValue, spannerType, typeof(long)));
             }
 
-            if (targetClrType == typeof(char))
+            if (targetClrType == typeof(sbyte))
```
benwulfe: no char needs special handling and cannot leverage the long codepath.  it was removed on purpose.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -137,14 +167,26 @@ internal V1.Type ToProtobufType()
 
         /// <summary>
         /// Creates an array of the specified type. This can be done on any arbitrary <see cref="SpannerDbType"/>.
+        /// You may use any type that implements IEnumerable as a source for the array.  The type of each
+        /// item is determined by <paramref name="elementType"/>.
+        /// When calling <see cref="SpannerDataReader.GetFieldValue(string)"/> the default type
+        /// is <see cref="ArrayList"/>. You may, however, specify any type that implements IList which
```
benwulfe: not particularly.  I did not want to force a generic on folks by default.  I can see about using List<object>

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ValueConversion.cs
```diff
@@ -76,21 +76,16 @@ public static object ConvertToClrType(this Value wireValue, V1.Type spannerType,
...
             {
-                return Convert.ToChar((long) ConvertToClrTypeImpl(wireValue, spannerType, typeof(long)));
+                return Convert.ToSByte((long)ConvertToClrTypeImpl(wireValue, spannerType, typeof(long)));
```
benwulfe: you mean why use Convert.ToSByte when you can just cast long to sbyte?  I wanted consistency in explaining to developers what to expect (expect convert to 'yourtype' to be called).
I dont see a downside to using Convert...

Agree with the extra helper that does the cast.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ValueConversion.cs
```diff
@@ -180,24 +175,25 @@ public static Value ToValue(object value, SpannerDbType spannerDbType)
                 case TypeCode.Bool:
                     return new Value {BoolValue = Convert.ToBoolean(value)};
                 case TypeCode.String:
+                    if (value is DateTime)
+                    {
+                        // If the value is a DateTime, we always convert using XmlConvert.
+                        // This allows us to convert back to a datetime reliably from the
+                        // resulting string (so roundtrip works properly if the developer uses
+                        // a string as a backing field for a datetime for whatever reason).
+                        return new Value { StringValue = XmlConvert.ToString((DateTime)value, XmlDateTimeSerializationMode.Utc) };
```
benwulfe: I'm using XmlConvert to parse RFC 3339 so I'm using it in both directions for clarity.
I would feel worse about doing say a parseexact and getting a slight variation in the format that XmlConvert might handle.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ValueConversion.cs
```diff
@@ -252,9 +248,37 @@ private static object ConvertToClrTypeImpl(Value wireValue, V1.Type spannerType,
...
                         return wireValue.BoolValue;
+                    case Value.KindOneofCase.NumberValue:
+                        return Convert.ToBoolean(wireValue.NumberValue);
+                    default:
+                        throw new InvalidOperationException(
+                            $"Invalid Type conversion from {wireValue.KindCase} to {targetClrType.FullName}");
+                }
+            }
+
+            if (targetClrType == typeof(char))
```
benwulfe: yes exactly.  to reduce the number of the cases here, earlier up I convert to others (eg long/double) if they can be done without loss of precision.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ValueConversion.cs
```diff
@@ -252,9 +248,37 @@ private static object ConvertToClrTypeImpl(Value wireValue, V1.Type spannerType,
...
+                switch (wireValue.KindCase)
+                {
+                    case Value.KindOneofCase.BoolValue:
+                        return Convert.ToChar(wireValue.BoolValue);
+                    case Value.KindOneofCase.NumberValue:
+                        return Convert.ToChar(wireValue.NumberValue);
+                    case Value.KindOneofCase.NullValue:
+                        return default(char);
+                    case Value.KindOneofCase.StringValue:
+                        if (spannerType.Code == TypeCode.Int64)
```
benwulfe: hopefully this makes more sense now.  int64 spanner types are encoded over the wire as strings.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/TypeTests.cs
```diff
@@ -0,0 +1,627 @@
...
+        private static readonly SpannerDbType s_arrayOfStruct = SpannerDbType.ArrayOf(s_struct);
+
+        private static readonly SpannerDbType s_complexStruct = SpannerDbType.StructOf(
+            new Dictionary<string, SpannerDbType>
+            {
+                {"StructField", s_struct},
+                {"ArrayField", SpannerDbType.ArrayOf(SpannerDbType.Int64)}
+            });
+
+        private static string S(string s) => $"\"{s}\"";
```
benwulfe: sure.  S was shorter.  I'm fine with Quote.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/TypeTests.cs
```diff
@@ -0,0 +1,627 @@
...
+
+        private static IEnumerable<string> GetStringsForArray()
```
benwulfe: i wanted some of the test cases to be a raw enumerable backed by a method to ensure that also worked.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/TypeTests.cs
```diff
@@ -0,0 +1,627 @@
...
+            yield return new object[] {1U, SpannerDbType.Float64, "1"};
+            yield return new object[] {1L, SpannerDbType.Float64, "1"};
+            yield return new object[] {(ulong) 1, SpannerDbType.Float64, "1"};
+            yield return new object[] {(short) 1, SpannerDbType.Float64, "1"};
+            yield return new object[] {(ushort) 1, SpannerDbType.Float64, "1"};
+            yield return new object[] {"1", SpannerDbType.Float64, "1"};
+            yield return new object[] {"1.5", SpannerDbType.Float64, "1.5"};
+            yield return new object[] {null, SpannerDbType.Float64, "null"};
+
+            //Spanner type = Int64 tests.
```
benwulfe: true -- originally some were lossy in precision and I'll add some back and make them one way.

jskeet:
Thanks for all the extra info - I definitely need to take another pass with fresh eyes. That may or may not be today, unfortunately. My main point of concern is still in terms of cultures:

- Converting from CLR to Spanner for things like double to string, there's an argument for using the current culture, but it still makes me shudder
- Converting from Spanner to CLR, I'm pretty sure we should be using the invariant culture everywhere

As an aside, we should have tests for timestamps with nanosecond precision (which should lose information but still parse without error) - just thought about that earlier, and didn't want to forget...

benwulfe:
I'm thinking its much better to always use invariant culture for all of our conversions.
If the developer wants to convert something for display purposes, they can do this on their own using their own conversions after retrieving it in its native format.

I think this falls into the case of not creating a system that allows developers to easily fall into a trap.  And converting using current culture can easily lead them into a situation where on some machines values get stored differently in the database.  This is a far worse situation that forcing the developer to do culture aware conversions on their own.

The alternative would be to allow the developer to pass in an IFormatProvider to use for conversions.  We'd only be able to support this on our own methods (as opposed to some codepaths that ado.net controls).  I think we can do this if demand shows itself -- but the conversion layer is already fairly heavy and its worth getting feedback on it before adding more functionality at this point.

benwulfe:
Did a pass to make conversions invariant culture and added a roundtrip test using nanoseconds.
probably needs another round of looking over/discussion.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ValueConversion.cs
```diff
@@ -180,24 +176,25 @@ public static Value ToValue(object value, SpannerDbType spannerDbType)
...
+                        // If the value is a DateTime, we always convert using XmlConvert.
+                        // This allows us to convert back to a datetime reliably from the
+                        // resulting string (so roundtrip works properly if the developer uses
+                        // a string as a backing field for a datetime for whatever reason).
+                        return new Value { StringValue = XmlConvert.ToString((DateTime)value, XmlDateTimeSerializationMode.Utc) };
+                    }
+                    return new Value { StringValue = Convert.ToString(value, CultureInfo.InvariantCulture) };
                 case TypeCode.Int64:
-                    return new Value {StringValue = value.ToString()};
+                    return new Value {StringValue = Convert.ToInt64(value).ToString(CultureInfo.InvariantCulture) };
```
jskeet: We don't know what type `value` is here, right? If it's a string, we should be using the invariant culture here, too.
(This is one thing I dislike about the `Convert` methods - they make it unclear whether we're performing string parsing, string formatting, or just a numeric/bitwise conversion such as `long` to `byte`.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ValueConversion.cs
```diff
@@ -373,14 +441,15 @@ private static object ConvertToClrTypeImpl(Value wireValue, V1.Type spannerType,
                 }
                 //a bit of recursion here...
                 IDictionary dictionary = (IDictionary)Activator.CreateInstance(targetClrType);
+                var itemType = targetClrType.GetGenericArguments().Skip(1).FirstOrDefault() ?? typeof(object);
```
jskeet: We could run into interesting problems here if what's passed in is an open generic type, but I think it's probably reasonable for that to fail.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PriorityList.cs
```diff
@@ -19,7 +19,8 @@
 namespace Google.Cloud.Spanner.V1
 {
     /// <summary>
-    /// A simple sorted list that allows items to return equality for IComparable.CompareTo
+    /// A simple sorted list that sorts based on the result of CompareTo,
```
jskeet: Is this change meant to be in this PR rather than the PriorityList one?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs
```diff
@@ -34,7 +34,6 @@ namespace Google.Cloud.Spanner.V1 {
     internal sealed class ReliableStreamReader : IDisposable
     {
         private AsyncServerStreamingCall<PartialResultSet> _currentCall;
-        private readonly object _sync = new object();
```
jskeet: Not sure which PR this is meant to be in, but if there's no particularly logic place, I don't mind too much.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/TypeTests.cs
```diff
@@ -0,0 +1,643 @@
...
+        {
+            //There are cases where the serialization (of structs) is nondeterministic
+            //This would cause the unit tests to fail randomly, so if we see a failure
+            //we decompose the expected and actual into their respective pieces and validate
+            //each piece.
+            //we try to decompose the actual and expected in case its an unordered dictionary.
+            expected = StrippedString(expected);
+            actual = StrippedString(actual);
+            var unorderedMatchExpected = new HashSet<string>(expected.Split(','));
+            var unorderedMatchActual = new HashSet<string>(actual.Split(','));
```
jskeet: Did parsing as a struct not work, or is that still TBD? (You could always add a TODO to do that later, to get this in sooner.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ValueConversion.cs
benwulfe: will examine value to see if its a string incoming...

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs
benwulfe: was a random thing I spotted (unused field).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/TypeTests.cs
benwulfe: tbd -- working on it now.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PriorityList.cs
benwulfe: i think I just noticed it when doing this PR.  If its not a big deal, would like to just leave it in here.  If its a big issue, I can create a separate PR.

benwulfe:
ok I squashed this down to a single commit, added the better conversion and a todo for culture specific tests.  Jon, you think we can get this in now?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/TypeTests.cs
```diff
@@ -0,0 +1,633 @@
...
+
+            //Spanner type = String tests.
+            //(all work)
+
+            //Spanner type = Date tests.
+            yield return new object[] {new ToStringClass("hello"), SpannerDbType.Date, ""};
+            yield return new object[] {"badjuju", SpannerDbType.Date, ""};
+        }
+
+        private string StrippedString(string s) => new StringBuilder(s)
```
jskeet: Looks like you can now delete this. (Yay!)

PR #1175, opened 2017-07-05
PriorityList unit tests.

benwulfe:


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
+﻿using System;
```
evildour: Add copyright notice

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    public class PriorityListTests
+    {
+        [Fact]
+        public void CanAddMultipleOfSamePriority()
+        {
+            PriorityList<TestItem> priorityList = new PriorityList<TestItem>();
```
evildour: Nit: I'd use `var` here (and below) where you're declaring and creating the same type.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+            AssertExist(priorityList, item1, item2);
+        }
+
+        [Fact]
+        public void DeleteRemovesCorrectItem()
+        {
+            PriorityList<TestItem> priorityList = new PriorityList<TestItem>();
+            TestItem toRemove;
+            priorityList.Add(toRemove = new TestItem(5));
+            priorityList.Add(new TestItem(5));
```
evildour: I'd add another item with the same priority first, like this:
```csharp
priorityList.Add(new TestItem(5));
priorityList.Add(toRemove = new TestItem(5));
priorityList.Add(new TestItem(5));
```
That way you'll catch implementations in the future which erroneously pick the first or the last item with the specified priority.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+            Assert.Equal(2, priorityList.Count);
+        }
+
+        [Fact]
+        public void DeleteRemovesEventHook()
+        {
+            PriorityList<TestItem> priorityList = new PriorityList<TestItem>();
+            TestItem toRemove;
+            priorityList.Add(toRemove = new TestItem(5));
+            priorityList.Add(new TestItem(5));
```
evildour: Adding this item doesn't seem necessary, unless you want to also test below that the item remaining in the list still has `HasEventHooks` true after the remove.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+
+            priorityList.Add(new TestItem(3));
+            priorityList.Add(itemToRemove = new TestItem(11));
+            priorityList.Add(new TestItem(2));
+            priorityList.Add(new TestItem(0));
+            priorityList.Add(new TestItem(6));
+
+            priorityList.RemoveLast();
+            AssertSorted(priorityList);
+            AssertNotExist(priorityList, itemToRemove);
```
evildour: Maybe assert for the count here too, to ensure `RemoveLast` doesn't remove multiple items.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+            priorityList.Add(new TestItem(11));
+            priorityList.Add(changingItem = new TestItem(2));
+            priorityList.Add(new TestItem(0));
+            priorityList.Add(new TestItem(6));
+            AssertSorted(priorityList);
+            changingItem.Priority = 20;
+            AssertNotSorted(priorityList);
+            changingItem.FirePriorityChanged();
+            AssertSorted(priorityList);
+        }
```
evildour: Possible additional test: it may be helpful to do another version of this which changes the priority to something equal to one of the others. Then remove one of the conflicting items after the `FirePriorityChanged` to make sure the correct one is removed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+            foreach (var testItem in testItems)
+            {
+                Assert.False(priorityList.TryFindLinear(x => ReferenceEquals(x, testItem), out _));
+            }
+        }
+
+        private void AssertSorted(PriorityList<TestItem> priorityList)
+        {
+            Assert.True(IsSorted(priorityList));
+        }
```
evildour: Nit: add newline

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+        private void AssertNotSorted(PriorityList<TestItem> priorityList)
+        {
+            Assert.False(IsSorted(priorityList));
+        }
```
evildour: Nit: add newline

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        private void AssertNotExist(PriorityList<TestItem> priorityList, params TestItem[] testItems)
+        {
+            foreach (var testItem in testItems)
+            {
+                Assert.False(priorityList.TryFindLinear(x => ReferenceEquals(x, testItem), out _));
+            }
+        }
+
+        private void AssertSorted(PriorityList<TestItem> priorityList)
```
evildour: I would argue that this and `AssertNotSorted` are not really necessary. The tests would be clearer to just perform the checks directly inline. If you want to keep them though, consider expression-bodied members perhaps.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+            int lastPriority = 0;
+            foreach (var testItem in priorityList.GetSnapshot())
+            {
+                if (lastPriority > testItem.Priority)
+                {
+                    return false;
+                }
+                lastPriority = testItem.Priority;
+            }
+            return true;
```
evildour: Consider this alternative implementation:
```csharp
var snapshot = priorityList.GetSnapshot();
return snapshot.SequenceEquals(snapshot.OrderBy(item => item.Priority));
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+                PriorityChanged?.Invoke(this, EventArgs.Empty);
+            }
+
+            /// <inheritdoc />
+            public int CompareTo(TestItem other)
+            {
+                return Comparer<int>.Default.Compare(Priority, other.Priority);
+            }
+
+            public bool HasEventHooks => PriorityChanged?.GetInvocationList().Any() ?? false;
```
evildour: I think `=> PriorityChanged != null;` should work.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/Google.Cloud.Spanner.Data.Tests.csproj
```diff
@@ -0,0 +1,28 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp1.0;net452</TargetFrameworks>
+    <IsPackable>false</IsPackable>
+    <AssemblyOriginatorKeyFile>../../GoogleApis.snk</AssemblyOriginatorKeyFile>
+    <SignAssembly>true</SignAssembly>
+    <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netcoreapp1.0|AnyCPU'">
```
jskeet: As before, I think you should be able to just remove this PropertyGroup entirely.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/Google.Cloud.Spanner.Data.Tests.csproj
```diff
@@ -0,0 +1,28 @@
...
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
+    <PackageReference Include="Moq" Version="4.7.63" />
+    <PackageReference Include="xunit" Version="2.2.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
+    <Reference Condition="'$(TargetFramework)' == 'net452'" Include="Microsoft.CSharp" />
+  </ItemGroup>
+
+  <ItemGroup>
```
jskeet: Have you run generateprojects.sh? I'd expect that to add a project reference for this already, so you don't need this ItemGroup. Ditto the one with the service.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -0,0 +1,235 @@
...
+    {
+        [Fact]
+        public void CanAddMultipleOfSamePriority()
+        {
+            PriorityList<TestItem> priorityList = new PriorityList<TestItem>();
+            TestItem item1, item2;
+            priorityList.Add(item1 = new TestItem(5));
+            priorityList.Add(item2 = new TestItem(5));
+            Assert.Equal(2, priorityList.Count);
+            AssertExist(priorityList, item1, item2);
```
jskeet: Do we want to assert anything about the order here?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
jskeet: Or an AssertInOrder(...) that basically checks that the order is exactly as specified.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
jskeet: +1

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/AssemblyInfo.cs
```diff
@@ -14,10 +14,30 @@
...
+
+[assembly: InternalsVisibleTo("Google.Cloud.Spanner.Data.Tests, PublicKey="
+                              + "0024000004800000940000000602000000240000525341310004000001000100afab79952ee222"
+                              + "15f12b4e09337e65509c943fbc22d7006bc371d581d0f0ebf0da5d8039aab2607fb68a138a5d80"
+                              + "a71bc02b7ebf586dbe1f2493c0ab20423ababfd15ce74d2264a6b37745f3658f016abaad662182"
+                              + "aaef634a60f1346fcc45343acab5b6781535a3134818e13fac895a6c106c0480e34bbb06cb123e"
+                              + "5583d8d2"
+)]
+
+[assembly: InternalsVisibleTo("Castle.Core, PublicKey=002400000480000094000000060200000024000052534131000400000"
```
jskeet: As mentioned in another PR, I hope we don't need this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/Google.Cloud.Spanner.Data.Tests.csproj
benwulfe: this will update once we get the test project PR in master and all of these PRs merged/sync'd.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
benwulfe: done/code cleaned up.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
benwulfe: Not really.  if the priority is the same, then the relative position is not deterministic.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
benwulfe: thats not a bad idea...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
benwulfe: kind of a dupe of remove + prioritychangeresortslist, but cannot hurt for sure.

benwulfe: a possibly more interesting one is a removal of an item that has changed its priority, but has yet to fire its event.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
benwulfe: I'm good with expression body.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
benwulfe: that's definitely more compact in terms of code, but it also is doing an unnecessary sort, right?
also, I assume it would work in cases of the priority matching on two instances because the prioritylist happens to use IComparer -- but it'd be nice to have less dependent on that design choice.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/AssemblyInfo.cs
benwulfe: its been removed and was only needed in the pool tests for the spannerclientfactory, where I made my own mock class.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
jskeet: I'd say it's clearer precisely *because* PriorityList uses the fact that IPriorityListItem is comparable. The "coincidence" is that TestItem implements `IComparable<TestItem>` by comparing Priority properties. So the check would be more general IMO.

(And yes, it does an unnecessary sort, but that's trivial amounts of extra work for these tests.)

I'm not massively bothered either way though - not a merge-blocker for me.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
jskeet: Right - I wasn't sure whether that was the case, or whether we were asserting insert order preservation.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
evildour: Actually, what I wrote also depends on comparing `Priority`. Change the lambda in my proposed lines to `item => item` to be more general. I too don't feel too strongly about it though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
```diff
@@ -8,10 +22,69 @@ namespace Google.Cloud.Spanner.Data.Tests
...
+        private class TestItem : IPriorityListItem<TestItem>
+        {
+            public int Priority { get; set; }
+
+            public bool HasEventHooks => PriorityChanged != null;
+
+            public TestItem(int priority) => Priority = priority;
+
+            /// <inheritdoc />
+            public int CompareTo(TestItem other) => Comparer<int>.Default.Compare(Priority, other.Priority);
```
evildour: I think `Priority.CompareTo(other.Priority)` should work here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/PriorityListTests.cs
benwulfe: I've actually been toying with the idea of having IPriorityListItem not derive from IComparable, but instead of a member "PriorityKey" which can optionally "return this".
If the priorityitem was cloneable (or a primitive), it would allow the prioritylist to more efficiently track items, because it could clone the old prioritykey.
So I consider that the item implements icomparable somewhat of a turd and not something I'd want to design a lot of tests around.
I agree the perf angle is not that of a big issue for a test -- but neither is the 3 lines of code that it saves the test.  So, on the whole I think I'll keep the current test as is, but it was a good suggestion.

benwulfe: If the priority*key* member was cloneable

PR #1176, opened 2017-07-05
ClientPool and SessionPool unit tests (including some individual stre…

benwulfe:
…ss/threading tests).

Runtime changes:
 Changes pool classes to be nonstatic so that pool unit tests can run concurrently and not affect other tests.
 All options on sessionpool are now instance based (except for marking sessions as expired, which is still static)
 Minor change in transaction mode where we account for mode not being set (encountered during testing).

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/MockClientFactory.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/MockClientFactory.cs
@@ -1,0 +1,39 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Threading.Tasks;
+using Google.Api.Gax.Grpc;
+using Google.Apis.Auth.OAuth2;
+using Google.Cloud.Spanner.V1;
+using Moq;
+
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    internal class MockClientFactory : ISpannerClientFactory
+    {
+        private SpannerClient _currentClient;
+
+        public int Invocations { get; private set; }
+
+        public MockClientFactory(SpannerClient firstClient) => _currentClient = firstClient;
+
+        public Task<SpannerClient> CreateClientAsync(ServiceEndpoint endpoint, ITokenAccess credential)
+        {
+            Invocations++;
+            var result = _currentClient;
+            _currentClient = new Mock<SpannerClient>().Object;
+            return Task.FromResult(result);
+        }
+    }
+}
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ClientPool.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ClientPool.cs
@@ -22,25 +22,32 @@ using Google.Api.Gax.Grpc;
 using Google.Apis.Auth.OAuth2;
 using Google.Cloud.Spanner.V1;
 using Google.Cloud.Spanner.V1.Logging;
-using Grpc.Auth;
-using Grpc.Core;
 
 namespace Google.Cloud.Spanner.Data
 {
 
-    internal static class ClientPool
+    internal class ClientPool
     {
-        private static readonly ConcurrentDictionary<ClientCredentialKey, CredentialClientPool> s_clientPoolByCredential =
-            new ConcurrentDictionary<ClientCredentialKey, CredentialClientPool>();
+        public static ClientPool Default { get; } = new ClientPool();
 
-        public static async Task<SpannerClient> AcquireClientAsync(
+        private readonly ConcurrentDictionary<ClientCredentialKey, CredentialClientPool> _clientPoolByCredential =
+            new ConcurrentDictionary<ClientCredentialKey, CredentialClientPool>();
+        private readonly ISpannerClientFactory _clientFactory;
+
+        // ReSharper disable once MemberCanBePrivate.Global
+        internal ClientPool(ISpannerClientFactory clientFactory = null)
+        {
+            _clientFactory = clientFactory ?? SpannerClientFactory.Default;
+        }
+
+        public async Task<SpannerClient> AcquireClientAsync(
             ITokenAccess credentials = null,
             ServiceEndpoint endpoint = null)
         {
             var key = new ClientCredentialKey(credentials, endpoint);
-            var poolEntry = s_clientPoolByCredential.GetOrAdd(key, k => new CredentialClientPool(k));
-            var result = await poolEntry.AcquireClientAsync().ConfigureAwait(false);
-            Logger.LogPerformanceCounter("SpannerClient.TotalCount", () => s_clientPoolByCredential.Count);
+            var poolEntry = _clientPoolByCredential.GetOrAdd(key, k => new CredentialClientPool(k));
+            var result = await poolEntry.AcquireClientAsync(_clientFactory).ConfigureAwait(false);
+            Logger.LogPerformanceCounter("SpannerClient.TotalCount", () => _clientPoolByCredential.Count);
             return result;
         }
 
@@ -48,13 +55,13 @@ namespace Google.Cloud.Spanner.Data
         //Returns the total of all reference counts.
         //For test purposes only.
         // poolContents will be filled with the current contents of the pool and may not be null.
-        internal static int GetPoolInfo(StringBuilder poolContents)
+        internal int GetPoolInfo(StringBuilder poolContents)
         {
             GaxPreconditions.CheckNotNull(poolContents, nameof(poolContents));
             int referenceCountTotal = 0;
             poolContents.AppendLine("ClientPool.Contents:");
             int i = 0;
-            foreach (var kvp in s_clientPoolByCredential)
+            foreach (var kvp in _clientPoolByCredential)
             {
                 poolContents.AppendLine($"s_clientPoolByCredential({i}) Key:${kvp.Key}");
                 referenceCountTotal += kvp.Value.DumpCredentialPoolContents(poolContents);
@@ -63,7 +70,7 @@ namespace Google.Cloud.Spanner.Data
             return referenceCountTotal;
         }
 
-        public static void ReleaseClient(SpannerClient spannerClient,
+        public void ReleaseClient(SpannerClient spannerClient,
             ITokenAccess credentials = null,
             ServiceEndpoint endpoint = null)
         {
@@ -71,7 +78,7 @@ namespace Google.Cloud.Spanner.Data
             {
                 var key = new ClientCredentialKey(credentials, endpoint);
                 CredentialClientPool poolEntry;
-                if (s_clientPoolByCredential.TryGetValue(key, out poolEntry))
+                if (_clientPoolByCredential.TryGetValue(key, out poolEntry))
                 {
                     poolEntry.ReleaseClient(spannerClient);
                 }
@@ -142,7 +149,7 @@ namespace Google.Cloud.Spanner.Data
                 }
             }
 
-            public Task<SpannerClient> AcquireClientAsync()
+            public Task<SpannerClient> AcquireClientAsync(ISpannerClientFactory clientFactory)
             {
                 Task<SpannerClient> result;
 
@@ -163,7 +170,7 @@ namespace Google.Cloud.Spanner.Data
                     //now grab the first item in the sorted list, increment refcnt, re-sort and return.
                     //The re-sorting will happen as a consequence of AcquireClientAsync changing its
                     //state and firing an event the prioritylist listens to.
-                    result = _clientPriorityList.GetTop().AcquireClientAsync();
+                    result = _clientPriorityList.GetTop().AcquireClientAsync(clientFactory);
                 }
                 return result;
             }
@@ -186,80 +193,27 @@ namespace Google.Cloud.Spanner.Data
             private Lazy<Task<SpannerClient>> _creationTask;
             private int _refCount = 0;
             private readonly ClientCredentialKey _parentKey;
+            // ReSharper disable once InconsistentNaming
 
             public SpannerClientCreator(ClientCredentialKey parentKey)
             {
                 _parentKey = parentKey;
-                _creationTask = new Lazy<Task<SpannerClient>>(AcquireClientImplAsync);
             }
 
             internal int RefCount => _refCount;
 
-            public bool MatchesClient(SpannerClient client) => _creationTask.IsValueCreated &&
-                !_creationTask.Value.IsFaulted &&
-                ReferenceEquals(_creationTask.Value.ResultWithUnwrappedExceptions(), client);
+            public bool MatchesClient(SpannerClient client) => _creationTask != null
+                && _creationTask.IsValueCreated
+                && !_creationTask.Value.IsFaulted
+                && ReferenceEquals(_creationTask.Value.ResultWithUnwrappedExceptions(), client);
 
-            private async Task<ChannelCredentials> CreateDefaultChannelCredentialsAsync()
+            public async Task<SpannerClient> AcquireClientAsync(ISpannerClientFactory clientFactory)
             {
-                var appDefaultCredentials = await GoogleCredential.GetApplicationDefaultAsync().ConfigureAwait(false);
-                if (appDefaultCredentials.IsCreateScopedRequired)
-                {
-                    appDefaultCredentials = appDefaultCredentials.CreateScoped(SpannerClient.DefaultScopes);
-                }
-                return appDefaultCredentials.ToChannelCredentials();
-            }
-
-            private async Task<SpannerClient> AcquireClientImplAsync()
-            {
-                var endpoint = _parentKey.Endpoint;
-                ChannelCredentials channelCredentials;
-
-                if (_parentKey.Credential == null)
-                {
-                    channelCredentials = await CreateDefaultChannelCredentialsAsync().ConfigureAwait(false);
-                }
-                else
-                {
-                    channelCredentials = _parentKey.Credential.ToChannelCredentials();
-                }
-
-                var channel = new Channel(
-                    endpoint.Host,
-                    endpoint.Port,
-                    channelCredentials);
-                Logger.LogPerformanceCounterFn("SpannerClient.RawCreateCount", x => x + 1);
-
-                //Pull the timeout from spanner options.
-                //The option must be set before OpenAsync is called.
-                var callSettings = CallSettings.FromCallTiming(
-                    CallTiming.FromRetry(
-                        new RetrySettings(
-                            SpannerSettings.GetDefaultRetryBackoff(),
-                            SpannerSettings.GetDefaultTimeoutBackoff(),
-                            Expiration.FromTimeout(SpannerOptions.Instance.Timeout),
-                            SpannerSettings.NonIdempotentRetryFilter
-                        )));
-
-                return SpannerClient.Create(
-                    channel, new SpannerSettings
-                    {
-                        CreateSessionSettings = callSettings,
-                        GetSessionSettings = callSettings,
-                        DeleteSessionSettings = callSettings,
-                        ExecuteSqlSettings = callSettings,
-                        ReadSettings = callSettings,
-                        BeginTransactionSettings = callSettings,
-                        CommitSettings = callSettings,
-                        RollbackSettings = callSettings
-                    });
-            }
-
-            public async Task<SpannerClient> AcquireClientAsync()
-            {
-                if (_creationTask.Value.IsFaulted)
+                if (_creationTask == null || _creationTask.Value.IsFaulted)
                 {
                     //retry an already failed task.
-                    _creationTask = new Lazy<Task<SpannerClient>>(AcquireClientImplAsync);
+                    _creationTask = new Lazy<Task<SpannerClient>>(
+                        () => clientFactory.CreateClientAsync(_parentKey.Endpoint, _parentKey.Credential));
                 }
 
                 var spannerClient = await _creationTask.Value.ConfigureAwait(false);
@@ -291,8 +245,15 @@ namespace Google.Cloud.Spanner.Data
                 {
                     return refCountComparison;
                 }
-                // This has a chance of returning 0 even if the object is different.
+                // This has a chance of returning 0 even if the object is not the same instance.
                 // That is fine and is handled by the PriorityHeap.
+
+                // Note: SpannerClients if created sequentially should only use a single channel.
+                // This is to faciliate good session pool hits (which is per channel).  If we always
+                // round robin'd clients, then we would get hard cache misses until we populated all
+                // of the caches per channel.
+                // We accomplish this by having a specified sort order based on hash if there is
+                // a tie based on refcnt.
                 return GetHashCode().CompareTo(other.GetHashCode());
             }
 
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientFactory.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientFactory.cs
@@ -1,0 +1,97 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Threading.Tasks;
+using Google.Api.Gax;
+using Google.Api.Gax.Grpc;
+using Google.Apis.Auth.OAuth2;
+using Google.Cloud.Spanner.V1;
+using Google.Cloud.Spanner.V1.Logging;
+using Grpc.Auth;
+using Grpc.Core;
+
+namespace Google.Cloud.Spanner.Data
+{
+    internal interface ISpannerClientFactory
+    {
+        Task<SpannerClient> CreateClientAsync(ServiceEndpoint endpoint, ITokenAccess credential);
+    }
+
+    /// <summary>
+    /// For mockability and testability of the <see cref="ClientPool"/> class.
+    /// </summary>
+    internal class SpannerClientFactory : ISpannerClientFactory
+    {
+        internal static readonly ISpannerClientFactory Default = new SpannerClientFactory();
+
+        private SpannerClientFactory()
+        {
+        }
+
+        private static async Task<ChannelCredentials> CreateDefaultChannelCredentialsAsync()
+        {
+            var appDefaultCredentials = await GoogleCredential.GetApplicationDefaultAsync().ConfigureAwait(false);
+            if (appDefaultCredentials.IsCreateScopedRequired)
+            {
+                appDefaultCredentials = appDefaultCredentials.CreateScoped(SpannerClient.DefaultScopes);
+            }
+            return appDefaultCredentials.ToChannelCredentials();
+        }
+
+        /// <inheritdoc />
+        public async Task<SpannerClient> CreateClientAsync(ServiceEndpoint endpoint, ITokenAccess credential)
+        {
+            ChannelCredentials channelCredentials;
+
+            if (credential == null)
+            {
+                channelCredentials = await CreateDefaultChannelCredentialsAsync().ConfigureAwait(false);
+            }
+            else
+            {
+                channelCredentials = credential.ToChannelCredentials();
+            }
+
+            var channel = new Channel(
+                endpoint.Host,
+                endpoint.Port,
+                channelCredentials);
+            Logger.LogPerformanceCounterFn("SpannerClient.RawCreateCount", x => x + 1);
+
+            //Pull the timeout from spanner options.
+            //The option must be set before OpenAsync is called.
+            var idempotentCallSettings = CallSettings.FromCallTiming(
+                CallTiming.FromRetry(
+                    new RetrySettings(
+                        SpannerSettings.GetDefaultRetryBackoff(),
+                        SpannerSettings.GetDefaultTimeoutBackoff(),
+                        Expiration.FromTimeout(SpannerOptions.Instance.Timeout),
+                        SpannerSettings.IdempotentRetryFilter
+                    )));
+
+            return SpannerClient.Create(
+                channel, new SpannerSettings
+                {
+                    CreateSessionSettings = idempotentCallSettings,
+                    GetSessionSettings = idempotentCallSettings,
+                    DeleteSessionSettings = idempotentCallSettings,
+                    ExecuteSqlSettings = idempotentCallSettings,
+                    ReadSettings = idempotentCallSettings,
+                    BeginTransactionSettings = idempotentCallSettings,
+                    CommitSettings = idempotentCallSettings,
+                    RollbackSettings = idempotentCallSettings
+                });
+        }
+    }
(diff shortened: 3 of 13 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ClientPool.cs
```diff
@@ -22,39 +22,46 @@
...
-using Grpc.Core;
 
 namespace Google.Cloud.Spanner.Data
 {
 
-    internal static class ClientPool
+    internal class ClientPool
     {
-        private static readonly ConcurrentDictionary<ClientCredentialKey, CredentialClientPool> s_clientPoolByCredential =
+        public static ClientPool Default { get; } = new ClientPool();
```
evildour: Since there is a shared common one, I assume we want this to be immutable? If so, add `readonly` to `_clientFactory`.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolImpl.cs
```diff
@@ -38,10 +41,13 @@ internal class SessionPoolImpl : IPriorityListItem<SessionPoolImpl>
 
         public SessionPoolKey Key { get; }
 
-        internal SessionPoolImpl(SessionPoolKey key)
+        internal SessionPoolImpl(SessionPoolKey key, SessionPoolOptions options)
         {
+            GaxPreconditions.CheckNotNull(options, nameof(options));
+            _options = options;
```
evildour: I believe this can just be `_options = GaxPreconditions.CheckNotNull(options, nameof(options));`

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Spanner.V1
+{
+    internal sealed class SessionPoolOptions {
```
evildour: Nit: move `{` to next line

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// If true, then CreateSessionFromPoolAsync will block until CurrentActiveSessions is less than MaximumActiveSessions
+        /// If false, then CreateSessionFromPoolAsync will throw an exception if CurrentActiveSessions is equal to or greater than MaximumActiveSessions
+        /// </summary>
+        public bool WaitOnResourcesExhausted { get; set; } = true;
+
+        /// <summary>
+        /// The amount of time before sessions get forcibly evicted from the session pool.
+        /// A lower value will cause the process to free sessions more aggressively when they get released
+        /// at the cost of performance due to lower reuse of sessions.
+        /// This value must be less than the expire timer on the Spanner server currently set at 60 minutes.
```
evildour: Nit: "...the Spanner server, **which is** currently set..." perhaps?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+
+        /// <summary>
+        /// If set to true, Sessions placed back into the pool will have a new transaction created in the background with
+        /// the same settings that were just used.  This will allow a later consumer of this session to skip creating a new
+        /// transaction if the options were identical.
+        /// </summary>
+        public bool UseTransactionWarming { get; set; } = true;
+
+        /// <summary>
+        ///
```
evildour: Missing summary

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
```diff
@@ -0,0 +1,171 @@
...
+{
+    public class ClientPoolTests
+    {
+        public ClientPoolTests(ITestOutputHelper outputHelper)
+        {
+            TestLogger.TestOutputHelper = outputHelper;
+            TestLogger.Install();
+            Logger.LogLevel = V1.Logging.LogLevel.Debug;
+        }
+        private static Mock<ISpannerClientFactory> SetupMockClientFactory(Mock<SpannerClient> mockClient)
```
benwulfe: renamed the argument and made it optional.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
```diff
@@ -0,0 +1,171 @@
...
+        [Fact]
+        public async Task SequentialCreatesOnlyUseOneChannel()
+        {
+            //SpannerClients if created sequentially should only use a single channel.
+            //This is to faciliate good session pool hits (which is per channel).  If we always
+            //round robin'd clients, then we would get hard cache misses until we populated all
+            //of the caches per channel.
+            //Since channels are lazily created, this also saves resources for scenarios like rich clients
+            //that may only need a single session for multiple sequential workloads.
+            //Our default setting for # channels = 4 needs to work well for all scenarios.
```
benwulfe: same client = same channel.  so by asserting that the same client is returned each time, you know that its resuing the same channel.  I'll be sure to copy this comment in the implementation, but this context is important to know if/when the test fails.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
```diff
@@ -0,0 +1,171 @@
...
+            var testPool = new ClientPool(mockClientFactory.Object);
+            var expectedClient = firstReturnedClient.Object;
+
+            Assert.Same(expectedClient, await testPool.AcquireClientAsync());
+            testPool.ReleaseClient(expectedClient);
+
+            Assert.Equal(expectedClient, await testPool.AcquireClientAsync());
+            testPool.ReleaseClient(expectedClient);
+
+            Assert.Equal(expectedClient, await testPool.AcquireClientAsync());
```
benwulfe: should be 'Same'

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
```diff
@@ -0,0 +1,171 @@
...
+            var testPool = new ClientPool(mockClientFactory.Object);
+
+            List<SpannerClient> clientList = new List<SpannerClient>();
+            for (var i = 0; i < SpannerOptions.Instance.MaximumGrpcChannels; i++)
+            {
+                var newClient = await testPool.AcquireClientAsync();
+                foreach (var existing in clientList)
+                {
+                    Assert.NotSame(existing, newClient);
+                }
```
benwulfe: wont that call Equals?  I want to ensure it does a referenceequals.

evildour: Yes it will, but if that's a problem, then line 155 below will be as well: `GroupBy` groups keys using the default equality comparer, which uses `Equals`. The clients don't appear to be overriding `Equals`, so I assume they should both be fine.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
```diff
@@ -0,0 +1,171 @@
...
+            }
+
+            //now we wrap around.
+            var firstReusedClient = await testPool.AcquireClientAsync();
+            Assert.Same(clientList[0], firstReusedClient);
+            testPool.ReleaseClient(firstReusedClient);
+
+            foreach (var client in clientList)
+            {
+                testPool.ReleaseClient(client);
```
benwulfe: yes thats intended because it got acquired twice.

evildour: Ah yes, sorry. Ignore my comment below about releasing 10 times then too.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
```diff
@@ -0,0 +1,171 @@
...
+            //now that everything got released, lets ensure the client order is preserved so
+            //that we again get the first client.
+            Assert.Same(firstReusedClient, await testPool.AcquireClientAsync());
+            testPool.ReleaseClient(firstReusedClient);
+
+            mockClientFactory.Verify(x => x.CreateClientAsync(It.IsAny<ServiceEndpoint>(), It.IsAny<ITokenAccess>()),
+                Times.Exactly(SpannerOptions.Instance.MaximumGrpcChannels));
+        }
+
+        private async Task<SpannerClient> GetSpannerClientAsync(ClientPool pool)
```
benwulfe: I've rerun code cleanup and should do the trick.,

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
```diff
@@ -0,0 +1,171 @@
...
+            Assert.Equal(SpannerOptions.Instance.MaximumGrpcChannels, grouping.Count());
+            foreach (var group in grouping)
+            {
+                Assert.Equal(multiplier, group.Count());
+            }
+
+            foreach (var client in concurrentQueries.Select(x => x.Result))
+            {
+                testPool.ReleaseClient(client);
+            }
```
benwulfe: doing each 10 times is intentional to ensure the refcnt goes to zero.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,443 @@
...
+            new ConcurrentDictionary<Mock<SpannerClient>, List<Session>>();
+
+        private ConcurrentStack<Transaction> _transactions = new ConcurrentStack<Transaction>();
+
+        public SessionPoolTests(ITestOutputHelper outputHelper)
+        {
+            //Uncomment these lines to debug a specific test.
+//            TestLogger.TestOutputHelper = outputHelper;
+//            TestLogger.Install();
+//            Logger.LogLevel = V1.Logging.LogLevel.Debug;
```
benwulfe: using a compilation symbol actually makes it harder IMO.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,443 @@
...
+        }
+
+        //Creates a client that will log created sessions in the _createdSession dictionary.
+        private Mock<SpannerClient> CreateMockClient(DatabaseName dbName = null)
+        {
+
```
benwulfe: removed in the code cleanup pass.  resharper is great...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,443 @@
...
+            var mockClient = new Mock<SpannerClient>();
+            if (dbName == null)
+            {
+                dbName = s_defaultName;
+            }
+            _createdSessions.TryAdd(mockClient, new List<Session>());
```
benwulfe: its actually meant to track all sessions across all clients and done in a central place so that it doesnt need to be set up locally in each test.
At some point I'll add some handling to add clients not found as that was the initial intent.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,443 @@
...
+        public async Task EmptyPoolCreatesNewSession()
+        {
+            using (var pool = new SessionPool())
+            {
+
+                var client = CreateMockClient();
+                var session = await pool.CreateSessionFromPoolAsync(
+                        client.Object, s_defaultName.ProjectId,
+                        s_defaultName.InstanceId, s_defaultName.DatabaseId, null, CancellationToken.None)
+                    .ConfigureAwait(false);
```
benwulfe: probably not, but it doesnt hurt.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,443 @@
...
+
+                Assert.Equal(1, _createdSessions.Count);
+                Assert.Equal(1, _createdSessions[client].Count);
```
benwulfe: yes, because the test is over at this point.  xunit creates an instance of the test class per test method.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
evildour: You can still do that. All I'm suggesting is changing this line to
```csharp
var mockClientSessions = new List<Session>();
_createdSessions.TryAdd(mockClient, mockClientSessions);
```
And then later, use `mockClientSessions` rather than looking up the list in the dictionary.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
evildour: Fair enough

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,443 @@
...
+                await pool.CreateSessionFromPoolAsync(
+                        client.Object, s_defaultName.ProjectId,
+                        s_defaultName.InstanceId, s_defaultName.DatabaseId, null, CancellationToken.None)
+                    .ConfigureAwait(false);
+                try
+                {
+                    await pool.CreateSessionFromPoolAsync(
+                        client.Object, s_defaultName.ProjectId,
+                        s_defaultName.InstanceId, s_defaultName.DatabaseId, null, CancellationToken.None)
+                        .ConfigureAwait(false);
```
benwulfe: thats true, but I'm leaning towards keeping it the way it is as its self documenting that an exception is expected versus "what is that assert(false) doing there?".  Then you'll have to add a comment to explain it, etc etc.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
benwulfe: caught!  yes thats true.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
benwulfe: hmm this is true...  sure sounds fine to me.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,443 @@
...
+                    client.Object, s_defaultName.ProjectId,
+                    s_defaultName.InstanceId, s_defaultName.DatabaseId, null, CancellationToken.None)
+                    .ConfigureAwait(false);
+                await pool.CreateSessionFromPoolAsync(
+                    client.Object, s_defaultName.ProjectId,
+                    s_defaultName.InstanceId, s_defaultName.DatabaseId, null, CancellationToken.None)
+                    .ConfigureAwait(false);
+
+                var releaseTask = Task.Delay(TimeSpan.FromMilliseconds(10)).ContinueWith(
+                    (x, y) => { pool.ReleaseToPool(client.Object, session1); }, null);
```
benwulfe: yeah I know a lot of folks dislike continuewith...  I'm good with your suggestion.

benwulfe: actually, resharper one-upped you and suggested an inline method:

                async Task ReleaseTask()
                {
                    await Task.Delay(TimeSpan.FromMilliseconds(10));
                    pool.ReleaseToPool(client.Object, session1);
                }
. . .
                await Task.WhenAll(createTask, ReleaseTask()).ConfigureAwait(false);

  I had no idea you could even do that.

benwulfe: "local function" is the C# 7 feature.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,443 @@
...
+
+                var createTask = pool.CreateSessionFromPoolAsync(
+                    client.Object, s_defaultName.ProjectId,
+                    s_defaultName.InstanceId, s_defaultName.DatabaseId, null, CancellationToken.None);
+
+                await Task.WhenAll(createTask, releaseTask).ConfigureAwait(false);
+                Assert.NotNull(createTask.ResultWithUnwrappedExceptions());
```
benwulfe: yep thats a good idea.  It shouldnt be anything else... but who knows...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,443 @@
...
+                    .ConfigureAwait(false);
+                pool.ReleaseToPool(client.Object, session);
+
+                var session2 = await pool.CreateSessionFromPoolAsync(
+                    client.Object, "newproject", "newinstance", "newdbid", null, CancellationToken.None)
+                    .ConfigureAwait(false);
+
+                Assert.NotSame(session, session2);
+                Assert.Equal(1, _createdSessions.Count);
+                Assert.Equal(1, _createdSessions[client].Count);
```
benwulfe: because I am only tracking default database sessions up in CreateMockClient:
Key line is:
"                        It.Is<DatabaseName>(y => y.Equals(dbName)),  "
Could probably at some point add more tracking for other sessions.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/TestLogger.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using Google.Cloud.Spanner.V1.Logging;
+using Xunit.Abstractions;
+
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    internal class TestLogger : DefaultLogger
+    {
+        private static Lazy<TestLogger> s_Instance = new Lazy<TestLogger>(
+            () => new TestLogger());
```
benwulfe: this is a copy of the one in integrationtests since I dont want to manage them referencing each other (yet) for just this class.  static is fine.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
benwulfe: actually this turned out to reveal an ordering issue that while doesn't result in a bug, could result in an unnecessary session alloc -- good catch!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
evildour: Yes, you can do that too. You can even move them down to the bottom of the owning function if you want (you can call them before they're declared). It can go anywhere as long as the locals that are closed over have been declared (and possibly assigned, I believe) before it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
```diff
@@ -20,86 +32,85 @@ public ClientPoolTests(ITestOutputHelper outputHelper)
...
-                        //return a unique instance each time.
-                        var result = mockClient.Object;
-                        mockClient = new Mock<SpannerClient>();
-                        return result;
-                    });
-            return mockClientFactory;
+            if (firstClient == null)
+            {
+                firstClient = new Mock<SpannerClient>();
+            }
```
evildour: If you push this into the constructor of `MockClientFactory`, you can remove `SetupMockClientFactory`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
```diff
@@ -20,86 +32,85 @@ public ClientPoolTests(ITestOutputHelper outputHelper)
...
 
-            Assert.Equal(expectedClient, await testPool.AcquireClientAsync());
-            testPool.ReleaseClient(expectedClient);
+            await Task.WhenAll(concurrentQueries);
+            Assert.Equal(SpannerOptions.Instance.MaximumGrpcChannels, mockClientFactory.Invocations);
 
-            Assert.Equal(expectedClient, await testPool.AcquireClientAsync());
-            testPool.ReleaseClient(expectedClient);
+            var grouping = concurrentQueries.GroupBy(x => x.Result).ToList();
+            Assert.Equal(SpannerOptions.Instance.MaximumGrpcChannels, grouping.Count());
```
evildour: Since you create a list, you can use `Count` instead of `Count()`, although it shouldn't matter too much either way, just an extra casting internally.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/MockClientFactory.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public Task<SpannerClient> CreateClientAsync(ServiceEndpoint endpoint, ITokenAccess credential)
+        {
+            var completionSource =
+                new TaskCompletionSource<SpannerClient>();
+
+            Invocations++;
+            completionSource.SetResult(_currentClient);
+            _currentClient = new Mock<SpannerClient>().Object;
+            return completionSource.Task;
```
evildour: You can use `Task.FromResult` instead so you don't have to create and populate a `TaskCompletionSource`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -72,67 +85,165 @@ private Mock<SpannerClient> CreateMockClient(DatabaseName dbName = null)
...
+        /// Creates <paramref name="parallelCount"/> write sessions simultaneously over
+        /// <paramref name="iterations"/> number of iterations.
+        /// </summary>
+        private async Task CreateReleaseWriteSessions(
+            SessionPool pool,
+            SpannerClient client,
+            int iterations,
+            int parallelCount)
+        {
+            await Task.Yield();
```
evildour: Add a comment to explain this and the yield below. Someone could come along in the future and remove it, thinking it is unnecessary.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/ClientPoolTests.cs
benwulfe: are you running a resharper analysis?  because resharper caught this and I doubt a human would ;)

evildour: Haha, nope. Something about it just caught my eye.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.UnitTests/SessionPoolTests.cs
```diff
@@ -0,0 +1,462 @@
...
+        /// Creates <paramref name="parallelCount"/> read sessions simultaneously over
+        /// <paramref name="iterations"/> number of iterations.
+        /// </summary>
+        private async Task CreateReleaseReadSessions(
+            SessionPool pool,
+            SpannerClient client,
+            int iterations,
+            int parallelCount)
+        {
+            await Task.Yield();
```
evildour: Add note to this one as well.

benwulfe:
Note that merging from master means that squashing the commits will be difficult as you cannot squash a merge.

evildour:
What about a rebase instead of a merge?

benwulfe:
you cannot rebase/squash a merge commit -- thats the issue.

evildour:
Sorry, I meant what about rebasing rather than merging. Perhaps it is too late now though.

jskeet:
I'm happy to try to get this rebased cleanly in a new branch if you'd like... let me know how I can help.

benwulfe:
ok managed to rebase into a single commit.  please approve if you think its ready.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/MockClientFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+    {
+        private SpannerClient _currentClient;
+
+        public int Invocations { get; private set; }
+
+        public MockClientFactory(SpannerClient firstClient) => _currentClient = firstClient;
+
+        public Task<SpannerClient> CreateClientAsync(ServiceEndpoint endpoint, ITokenAccess credential)
+        {
+
```
jskeet: Nit: remove blank line

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+            var callSettings = CallSettings.FromCallTiming(
+                CallTiming.FromRetry(
+                    new RetrySettings(
+                        SpannerSettings.GetDefaultRetryBackoff(),
+                        SpannerSettings.GetDefaultTimeoutBackoff(),
+                        Expiration.FromTimeout(SpannerOptions.Instance.Timeout),
+                        SpannerSettings.NonIdempotentRetryFilter
+                    )));
+
+            return SpannerClient.Create(
```
jskeet: As noted in the other PR, I'm nervous about these all using the NonIdempotentRetryFilter.
(Maybe GAX should provide something to make it easier to apply a timeout across all operations? Or maybe just have a CallSettings.WithTimeout() extension method or something like that?)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolImpl.cs
```diff
@@ -130,7 +143,7 @@ private bool TryPop(TransactionOptions options, out SessionPoolEntry entry)
                             {
                                 Logger.Debug(() => "found a session with matching transaction semantics.");
                                 indexToUse = i;
-                                if (options.ModeCase == TransactionOptions.ModeOneofCase.ReadOnly
+                                if (options?.ModeCase != TransactionOptions.ModeOneofCase.ReadWrite
```
jskeet: Possibly rename the local variable to `transactionOptions` to avoid having both `_options` and `options` in scope? (Or rename `_options` to `_poolOptions`?)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
```diff
@@ -0,0 +1,58 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Spanner.V1
+{
+    internal sealed class SessionPoolOptions
```
jskeet: Were all of these somewhere else before? I was looking for a corresponding delete but couldn't see it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientFactory.cs
benwulfe: I think we chatted on this briefly.  definitely seems harder than it should be to set an overall timeout.  I should not have to know to specifics about how to construct the entire retry settings.

It would be much better if the client call settings werent completely generated code in vkit but was in some way part of a base framework with overrideable/mutable properties.  I know you are big on immutability here as a way of ensuring that developers better understand when settings are frozen, but in the case of retry settings, you don't actually need to freeze them until you are inside a call.
So what I would do is make all of the retrysettings mutable on (BASE)Client and freeze them into immutable properties only when a callcontext is constructed (an rpc call is underway).

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
benwulfe: yes, these were part of SessionPool earlier.  I've extracted them so I can pass them down to child objects without needing to pass the entire root object.

https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/1176/files#diff-651f8cc55e7fee6255d710a72c806914

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientFactory.cs
jskeet: We had mutable RetrySettings etc classes before. They were a disaster. That much is not changing now. But we could add methods to make it easier to create a new RetrySettings from an existing one.

benwulfe: actually these should all be *idempotent*. (I mistakenly thought all were marked nonidempotent in SpannerClient, but two *are* marked idempotent -Get/Delete Session).

However, Commit is the only API really at risk here which has a set of arguments which are not idempotent -- you can do a single commit call without a begintransaction to commit a set of mutations in one roundtrip.  This call, since it does not have a transactionid, is not idempotent.  However, I specifically avoid this because its not idempotent and instead use CommitSettings that always include a transactionid.  This is idempotent.

So actually I will purposely deviate from the base setting for a few of these APIs.
I'm not sure why CreateSession, ExecuteSql, Read, BeginTx are marked as nonidempotent.  Maybe vkit is using some general rule based on the operation type?

benwulfe: Also - curious why it was a disaster...? maybe send me details offline?

jskeet: Well it's all configured in the YAML file - I don't know whether that was manually configured or just generated (which may well be based on the operation type).

benwulfe:
Jon -- were you also ok with this one?

jskeet:
Yes, other than the setting part, I was happy.

PR #1180, opened 2017-07-06
Initial retry logic for consumers

jskeet:
This is #1108 rebased into master and squashed to a single commit.


jskeet:
Ah - having accidentally pushed into Ian's branch after all, it looks like I can kill this.

PR #1183, opened 2017-07-06
Hi-perf pubsub integration test

chrisdunelm:


jskeet:
I'm *drowning* in PRs today, so this may end up being reviewed tomorrow.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.PubSub.V1.IntegrationTests
+{
+    [Collection(nameof(PubsubFixture))]
+    public class HighPerformancePubsub
```
jskeet: Test suffix? (Also I'd stick with PubSub with a capital S wherever possible. The lower case s in the proto-generated names is unfortunate.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
```diff
@@ -0,0 +1,163 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.PubSub.V1.IntegrationTests
+{
+    [Collection(nameof(PubsubFixture))]
```
jskeet: Possibly add a comment explaining why there's console output. (I think it's fine to do, but worth explaining.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
```diff
@@ -0,0 +1,163 @@
...
+            // Watchdog to report progress and fail test on deadlock
+            int prevSentCount = -1;
+            int prevRecvCount = -1;
+            CancellationTokenSource watchdogCts = new CancellationTokenSource();
+            Task.Run(async () =>
+            {
+                while (true)
+                {
+                    await Task.Delay(TimeSpan.FromSeconds(1), watchdogCts.Token).ConfigureAwait(false);
+                    var localSentCount = Interlocked.Add(ref sentCount, 0);
```
jskeet: This is to do a thread-safe read, yes? Another alternative is to use CompareExchange.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
```diff
@@ -0,0 +1,163 @@
...
+            int sentCount = 0;
+            long sentSum = 0L; // Sum of bytes of sent messages
+
+            // Watchdog to report progress and fail test on deadlock
+            int prevSentCount = -1;
+            int prevRecvCount = -1;
+            CancellationTokenSource watchdogCts = new CancellationTokenSource();
+            Task.Run(async () =>
+            {
+                while (true)
```
jskeet: Possibly `while (!watchdogCts.IsCancellationRequested)` just to make it clearer that this *will* end? (It doesn't much matter whether it ends due to spotting that or the Task.Delay being cancelled.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
```diff
@@ -0,0 +1,163 @@
...
+            });
+
+            // Publish
+            var rnd = new Random(1234);
+            var activePubs = new HashSet<Task>();
+            int sentCount = 0;
+            long sentSum = 0L; // Sum of bytes of sent messages
+
+            // Watchdog to report progress and fail test on deadlock
+            int prevSentCount = -1;
```
jskeet: Do these really need to be outside the scope of the watchdog task?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
```diff
@@ -0,0 +1,163 @@
...
+                    Assert.True(false, "Test cancelled by watchdog");
+                }
+                var msgSize = rnd.Next(minMessagesize, maxMessageSize + 1);
+                var msg = new byte[msgSize];
+                rnd.NextBytes(msg);
+                sentSum += msg.Sum(x => (long)x);
+                // Send message, and record Task
+                var pubTask = bulkPublisher.PublishAsync(msg);
+                Interlocked.Increment(ref sentCount);
+                lock (activePubs) activePubs.Add(pubTask);
```
jskeet: Braces or use Locked. (Or just use a concurrent collection?)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
```diff
@@ -0,0 +1,163 @@
...
+                // Remove Task from active when the message has been sent to server
+                Task unused = Task.Run(async () =>
```
jskeet: Rather than use Task.Run, maybe just attach a continuation to pubTask?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
```diff
@@ -0,0 +1,163 @@
...
+                });
+                // If too many messages are currently in flight, wait a bit
+                while (activePubs.Locked(() => activePubs.Count) >= maxMessagesInFlight)
+                {
+                    await Task.Delay(TimeSpan.FromMilliseconds(1)).ConfigureAwait(false);
+                }
+            }
+            Console.WriteLine("Publishing complete");
+            // Wait for all messages to be sent to server
+            await Task.WhenAll(activePubs.Locked(() => activePubs.ToArray())).ConfigureAwait(false);
```
jskeet: I don't think you need to worry about ConfigureAwait in these tests...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
```diff
@@ -0,0 +1,163 @@
...
+            Console.WriteLine("Publishing completed sending to server");
+
+            // Wait for subscriber to finish shutdown
+            await subTask.ConfigureAwait(false);
+            watchdogCts.Cancel();
+            Console.WriteLine("Subscriber finished shutdown");
+            Console.WriteLine($"Sent: {sentCount}; Recv: {recvCount}");
+
+            // Check that all messages are correctly received (probably)
```
jskeet: Why the probably?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
chrisdunelm: Done

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
chrisdunelm: Oops, done :)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
chrisdunelm: Fixed.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
chrisdunelm: Done

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
chrisdunelm: Yes, to make it thread-safe.
I can never quite remember how CompareExchange works! So using Add is quicker ;)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
chrisdunelm: Done.
There isn't a concurrent HashMap. There's a ConcurrentBag, but it's not the same thing.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
chrisdunelm: Done

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
chrisdunelm: I've had an instance of an xunit test where it did matter, so I no longer take the chance ;)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
chrisdunelm: Because simply adding all the bytes together isn't fool-proof :) There could be some weird error condition that causes the count and sums to match, but is actually wrong. Unlikely though.

jskeet: Ah, I see. That makes sense, but isn't terribly clearly indicated by "probably". I'd have something like:

> This isn't foolproof (we can get to the right sum with wrong values) but it's a pretty strong indicator.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubsub.cs
jskeet: You could use ConcurrentDictionary easily, using either a fixed value or the same key and value (basically use it as a set). Not really a problem, but worth considering.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubSubTest.cs
```diff
@@ -0,0 +1,163 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.PubSub.V1.IntegrationTests
+{
+    // Notes:
+    // * These tests output progress to the console, as they can take minutes to run.
+    // * These send/recv considerable amounts of network data.
+    //   Only test with a machine and cloud account that can handle this.
```
jskeet: Hmm. Given the warning, I wonder whether we need to create a new kind of test project - StressTests or similar. (Ben has the same requirement for Spanner.) I normally run all the integration tests before doing a release, and we probably don't want to do that from my home machine/network/account... it's probably okay for now, but let's think about it.

jskeet:
A couple of potential nits you could consider addressing, but LG already and if you do address the nits, I'm sure I don't need to check them :)

chrisdunelm:
One nit addressed.

Thanks for the LG, but you *haven't approved it* ;)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/HighPerformancePubSubTest.cs
chrisdunelm: Yes, some formal indication that this has stressy requirements sounds like a good idea.

PR #1191, opened 2017-07-07
Add copyright analyzer

evildour:
I also had to remove the `Resources.resx` file. Since moving to the new style projects, it no longer re-generates the associated designer code file. Strings are now used inline in the analyzers.

Easiest to review commit-by-commit.


tools/Google.Cloud.Tools.Analyzers/MissingCopyrightNoticeAnalyzer.cs
```diff
@@ -0,0 +1,132 @@
...
+                        Rule,
+                        context.Tree.GetLocation(new TextSpan(0, 1))));
+            }
+        }
+
+        private static bool RequiredCopyrightIsMissing(SyntaxNode rootNode)
+        {
+            if (rootNode.FullSpan.Length == 0)
+            {
+                // Ignore empty files. Temporaries files for the build process are sometimes empty.
```
jskeet: Nit: Temporaries => Temporary

evildour:
Hmm, the analyzer seems to be finding issues on travis. Not really sure why. Perhaps newline issues.

tools/Google.Cloud.Tools.Analyzers/MissingCopyrightNoticeAnalyzer.cs
evildour: Done

evildour:
Nope, that didn't fix it. Also, the first error (in `Source/V1/SourceContext.File.cs`) is for a file generated by the protocol buffer compiler and that check just looks at the first comment line. It doesn't look at newlines. I'm going to add some temporary additional info into the error message so we can see what it's seeing.

evildour:
Actually, I was wrong, that file has the normal copyright, so it still could be a newline issue.

evildour:
Oh, I think I just realized what it is: the 2nd `Replace` call in `CreateCopyrightRegex` is looking for `\r\n`, but it won't find it.

evildour:
That seems to have worked. Apparently, in GoogleStackdriverAppender.cs, the newlines are '\r\n' on travis, but they are '\n' in all other files. Not a clue why that would be, but now the regex is robust enough to recognize either instead of just one or the other.

evildour:
8th time's a charm?

evildour:
Finally! Will wait for AppVeyor to catch up before merging.

PR #1194, opened 2017-07-07
Add analyzers to tests

evildour:
Recommended to review one commit at a time


jskeet:
I was briefly worried about async - but of course, that's a different checker. LGTM.

evildour:
Thanks. Yes, it's only the two I added. I actually thought the default parameter checker would be an issue so we'd need a ruleset file, but forgot that it restricts to same assembly calls only. But if/when we need it, it should be easy to add a common ruleset file that all tests point to which disables certain rules.

Rebased with latest changes. Will merge when green.

evildour:
Looks like the rate limiter test again. Going to merge.

PR #1200, opened 2017-07-10
Try to fix RateLimiterTest.CanTrace_StressTest Appveyor Failures.

iantalarico:
The test keeps failing by undercounting the number of traces that should be allowed.  Try to space out the time to allow for more wiggle room.

I'm still unable to repo a failure locally with the current code or this new change.


evildour:
LGTM. Have you tried using [Process.ProcessorAffinity](https://msdn.microsoft.com/en-us/library/system.diagnostics.process.processoraffinity(v=vs.110).aspx) to pin things to a single core while trying to reproduce it? It's possible it only happens in that case.

jskeet:
I've tried setting processor affinity and still failed to reproduce this. This change LGTM though.

iantalarico:
Thanks, if this doesn't work I'll put in a some timers and see if we can figure out where the time is going.

PR #1201, opened 2017-07-11
Move API metadata int Google.Cloud.Tools.Common

jskeet:
This will allow more tools to be written using the metadata.

(This is basically just a refactoring.)


jskeet:
Thanks! Had forgotten you'd got that PR waiting...

PR #1203, opened 2017-07-11
Upgrade to docfx 2.20

jskeet:
This fixes two issues that are important to us:

- The .manifest file is now stable and formatted, so changes wil
  be much easier to understand
- The bug around dependencies being pulled in automatically has
  been fixed

Quotes are now escaped in the YAML file, which seems to be harmless.
I've already pushed the new "dependencies" branch using 2.20.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -23,7 +23,7 @@ install:
   # Download and unpack docfx
   - mkdir docfx
   - cd docfx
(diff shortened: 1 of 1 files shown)


jskeet:
Waiting for AppVeyor this time, given that the change is only in AppVeyor configuration :)

PR #1204, opened 2017-07-11
Include direct package reference to Grpc.Core in some "other" project types

### LINKED ISSUE #1187: Add dependencies to Grpc.Core from more projects
Projects which are of type "other" in apis.json don't get the automatic Grpc.Core dependency. We should either do that automatically, or make sure we do it explicitly ourselves.

jskeet:
Fixes #1187.


jskeet:
Aargh, need to be a little smarter here. Will fix.

evildour:
Is this in a state where it can be reviewed?

jskeet:
Just tweaked whitespace a bit more. I could pull that into a separate commit if it would help - I just noticed the odd indentation and thought I'd fix that while I'm here.

I'd wait until this goes green on Travis, then review.

PR #1206, opened 2017-07-11
Remove the option to log exceptions straight to the error reporting API.

iantalarico:
The error reporting API has not gone GA yet so these libraries cannot GA with this option.

We do this by making the method to create an event target to report events to the stackdriver error reporting API internal.


iantalarico:
Well we still have the event targets just only for logging.  I'm happy to just rip out the ability to create an event target for error reporting and leave everything.

I still use the error reporting API here to format the messages as it was easier then formatting it manually.

If you are okay with it I'll revert most of this and just remove the error reporting option?

jskeet:
Will have a look in the morning, but I'm concerned if we're not removing the dependency - basically I don't want to go GA with an API that's V1Beta1, and I don't want a GA library to depend on a non-GA library. We may need to be creative with this...

iantalarico:
I think with this

iantalarico:
@jskeet friendly ping!

jskeet:
I'm confused as to what the plan is, to be honest. We can't plan to GA with a beta dependency. What exactly is the formatting here? Can we replace that code instead?

iantalarico:
I think I misunderstood your initial comment.  The initial commit ripped out all of the logic to send a event to the error reporting API.  The only reason we kept the dependency on the nuget package was to format the message to send to the logging API.  So not calling the API just using a data structure.  My thought was if it got deprecated we could just manually format the message?

In the second iteration I just removed the option to use the error reporting option but left all of the code if we ever want to turn it on. (this is the part I misunderstood).

jskeet:
I'd suggest:

- Ripping all code and the dependency out, formatting the message manually
- Leaving enough options so that if we want to add in the GA API later, we can do so in a non-breaking way.

iantalarico:
Closing this in favor of a new PR for simplicity,

PR #1208, opened 2017-07-11
switch to using the CI project

benwulfe:
once this is in, we can turn on the integration tests in CI.


jskeet:
Could we make this configurable via an environment variable, as it is for all the other integration tests?

benwulfe:
appveyor false negative.  merging...

benwulfe:
oh just saw your comment jon.  sure

PR #1209, opened 2017-07-11
Spanner column

benwulfe:
Adds API sugar so that insert, update and delete commands use "SpannerColumns" instead of SpannerParameter.  spannerparameter makes less sense because the commandtext contains no parameters and you instead should be specifying which "columns" to update/insert, etc.


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerColumnCollection.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>
+        /// Creates a new <see cref="SpannerParameterCollection" />
+        /// </summary>
+        /// <param name="parameters">The initial collection of parameters</param>
+        public SpannerColumnCollection(IEnumerable<SpannerColumn> parameters)
+            : base(parameters) { }
+
+        /// <inheritdoc />
+        public new IEnumerator<SpannerColumn> GetEnumerator() => InnerList.Cast<SpannerColumn>().GetEnumerator();
```
evildour: This worries me. What if someone has some general logic for dealing with `SpannerParameterCollection` and that logic creates and adds `SpannerParameter` instances to the collection (using either `Add(object)`, `AddRange(Array)`, or `Insert(int, object)`) and then someone passes a `SpannerColumnCollection` to it? It appears those operations will succeed since they are only trying to cast the the instances to `SpannerParameter`s. So it won't be until someone goes to enumerate over the collection here that things will blow up.

There are ways to handle this, but the general approach worries me a bit. Can any of these alternatives work?
- Find some name for the collection that satisfies both use cases? `SpannerFieldCollection` perhaps?
- Derive both collections from an abstract, generic, base collection.
- Have both collections contain an internal, generic, implementation collection which does the actual work and delegate to it.

I'm not going to review the rest for now in case there are other changes based on this discussion.

evildour: Oh actually `SpannerFieldCollection` doesn't work because in select commands it is for arbitrary parameters, right? I don't have a good suggestion then at the moment.

benwulfe: right spannerfieldcollection doesn't work.

I agree with your overall comment, but I dont think the fix need be as drastic.  we can make the innerlist be provided by the derived class or simply add some (protected virtual) validation when adding to the innerlist.  either would work.

however spannercolumn needs to derive from spannerparameter and spannercolumncollection needs to derive from spannerparametercollection in order for the rest of the framework to work properly.

jskeet: I thought a SpannerColumn had *less* information than a SpannerParameter - which would suggest that the inheritance should be the other way round, shouldn't it? It makes more sense to me to be able to use a parameter when I need to specify a column, for example. But SpannerParameter deriving from DbParameter makes this problematic anyway, IMO... fundamentally, a column isn't a parameter, is it? It doesn't have a value - just a name and type.

benwulfe: unfortunately its not possible to have the inheritance the other way (even though it clearly makes more sense) because of how the ado.net base classes enforce that commands have *parameters*.
So unfortunately, I'm starting to think that while this change was well intentioned, it doesn't completely solve the problem and could make things more confusing.

Sorry Jon as this change was to address some of your concerns about using parameters in some of the commands.  But I am not seeing a path forward as it is.

evildour: Another concern I have is that for mutates, the collection is initializing one list of values out of possibly many in a [Mutation.Write](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Mutation.Write) Do you see it being possible that we'll allow multiple rows to get inserted/updated/deleted in a single command? If so, we might want to make these separate collections anyway (if possible). Similarly, it looks like [ReadRequest](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#readrequest) supports selected rows with ranges rather than single values. Would we ever want to support that with `CreateSelectCommand`? If we want to support either of these scenarios, keeping things separate may be best (once again, if possible).

benwulfe: multiple rows being affected is usually done by executing a command multiple times.
spanner.net does not use readrequest -- it only does an executesql for streaming reads.

evildour: What is spanner.net? And so does that mean it doesn't support key ranges?

benwulfe: spanner.net = this library.  and correct it does not support the read api at all.  to do ranges, you should write a sql query that has parameters for the range.

jskeet:
Okay - it's a pity, but such is life, sometimes. Are there any aspects we still want to keep? For example, the reorder of `Add` to (name, type, value)?

benwulfe:
yes -- thats a good change because it can otherwise cause some (really bad) mistakes.  I'll remove most of this and reduce it down to just that change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerColumnCollection.cs
evildour: Oh, haha. Thanks, that makes sense. Perhaps it is too general, but what about `SpannerKeyValueCollection` for a common name?

benwulfe: it would still ultimately derive from dbparameter so then you would have spannerparametercollection -> spannerkeyvaluecollection -> dbparametercollection?
and what would the name of an individual item be?  spannerkey?  then spannerkey would derive from dbparameter and you arguably have the same oddness in class heirarchy.

evildour: Oh sorry, I was just suggesting that as a rename for the existing collection, but keeping a single collection like it is now. I guess for consistency, that would mean the item would be `SpannerKeyValue` or `SpannerKeyValuePair`. Not sure I like this better than what's here already, just throwing out alternatives.

benwulfe:
closing this out -- submitting the reduced version as a different PR.

PR #1212, opened 2017-07-12
Publisher max send/recv sizes to unlimited

chrisdunelm:
And comment about data sizes in integration tests


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SimplePublisher.cs
```diff
@@ -237,9 +237,15 @@ public FlowState(long elementCount, long byteCount)
             var endpoint = clientCreationsettings?.ServiceEndpoint ?? PublisherClient.DefaultEndpoint;
             var clients = new PublisherClient[clientCount];
             var shutdowns = new Func<Task>[clientCount];
+            // Set channel send/recv message size to unlimited. It defaults to ~4Mb which causes failures.
```
jskeet: Do we need this part as well as the timing change?

chrisdunelm: Yes, this is definitely worth having here, to make sure it doesn't become a problem.

PR #1214, opened 2017-07-12
Remove InternalsVisibleTo between prod assemblies

jskeet:
We may wish to consider moving some types into an "Internals"
namespace to discourage users further.

This is very much just a first pass to get the lay of the land - I'm very open to suggestions, including that we don't do this at all if we're happy to be *really* diligent about internal versioning. (I'd rather remove the risk though, obviously.)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ExecuteHelper.cs
```diff
@@ -14,14 +14,18 @@
...
-    internal static class ExecuteHelper
+    /// <summary>
+    /// Helper class to wrap operations.
+    /// </summary>
+    public static class ExecuteHelper
     {
+        /// <summary>
+        /// Waits for <paramref name="task"/> to complete, handling session expiry by marking the session appropriately.
+        /// </summary>
         public static async Task<T> WithSessionChecking<T>(this Task<T> task, Func<Session> sessionFunc)
```
benwulfe: Would be nice if this wasnt in the public api.

I think ideally, we'd be intercepting all the rpc calls and automatically tracking expired sessions that came from the pool.  Then the developer wouldnt have to do anything to have his expired sessions tracked properly (and excluded from the pool).

does vkit have any notion of an exception filter hook?

jskeet: Not that I can remember, at the moment. But we could potentially add something in `CallSettings`. @chrisdunelm any thoughts on that?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Logging/LogLevel.cs
```diff
@@ -15,9 +15,9 @@
 namespace Google.Cloud.Spanner.V1.Logging
 {
     /// <summary>
-    /// 
+    /// Log levels for Spanner client internal logging.
     /// </summary>
-    internal enum LogLevel
```
benwulfe: ugh, I forgot about logging.  This was one of the more important ones to keep internal for when we switch to a more official logging/trace framework.
maybe we just move this to V1.Internal.Logging ns?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PriorityList.cs
```diff
@@ -22,10 +22,14 @@ namespace Google.Cloud.Spanner.V1
     /// A simple sorted list that sorts based on the result of CompareTo,
     /// but also allows duplicate entries based on CompareTo.
     /// </summary>
-    internal class PriorityList<T> where T : IPriorityListItem<T>
+    /// <remarks>This is not intended for general purpose use; it is designed for use with Spanner clients.</remarks>
+    public class PriorityList<T> where T : IPriorityListItem<T>
```
benwulfe: move to V1.Internal...?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
```diff
@@ -38,7 +38,7 @@ namespace Google.Cloud.Spanner.V1
     /// and it will incur a penalty on other Spanner processes because it will be an hour before the server frees the session if 
     /// it's not properly deleted.
     /// </summary>
-    internal static class SessionPool
+    public static class SessionPool
```
benwulfe: in an ideal scenario, you'd be able to turn on session pooling and use the same apis you would normally -- but the sessions would be pooled.  the problem with doing this is with modifying (or hooking into) generated code.

So I'm not sure I'm completely happy with the current api and how someone would use the pool directly at the V1 level.  But at the same time, there probably are going to be very few folks who need to...

It would just be nice to give a little more thought on how folks would use this functionality.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClient.Streaming.cs
```diff
@@ -28,7 +28,7 @@ public partial class SpannerClient
         /// <param name="request"></param>
         /// <param name="callSettings"></param>
         /// <returns></returns>
-        internal virtual AsyncServerStreamingCall<PartialResultSet> ExecuteSqlStream(ExecuteSqlRequest request, CallSettings callSettings = null)
+        public virtual AsyncServerStreamingCall<PartialResultSet> ExecuteSqlStream(ExecuteSqlRequest request, CallSettings callSettings = null)
```
benwulfe: this class is modifying the partial autogenerated class.
some of this is to be replaced by work chris bacon is doing on streaming calls -- but not the api that returns a reliablestreamreader.

I actually think this is where these APIs need to be -- but I'm wondering if you noticed the class this is in.
What is the objection with modifying the partial class?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/TransactionPool.cs
```diff
@@ -118,7 +118,8 @@ public static TransactionOptions GetLastUsedTransactionOptions(this Session sess
             return null;
         }
 
-        public static bool IsPreWarmedTransactionReady(this Session session)
+
```
benwulfe: nit newline.

jskeet: Fixed (locally, not pushed yet).

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClient.Streaming.cs
jskeet: Eek, hadn't spotted that this was generated.
Will backtrack to work out why I made these public.
(Not sure which partial class you're asking about in terms of modifying it - we really don't want to modify anything autogenerated though.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
jskeet: Happy to think about it - and if there's some way we could do this automatically, that would be great :) We do have the ability to hook modify requests shortly before they're sent, if that's useful. (There are partial methods generated.) When's the right time to investigate this?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PriorityList.cs
jskeet: SGTM. I'd been wondering about creating an Internal namespace. Glad we're thinking alike.
Will do that in a second commit in this PR.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClient.Streaming.cs
jskeet: Hang on - no, *this* source file isn't generated, as far as I can tell. Yes, it's modifying SpannerClient *most* of which is generated, but I'm fine with that. No problems :)

jskeet:
Just doing this, I've noticed the tests for PriorityList are in Google.Cloud.Spanner.Data.Tests, not Google.Cloud.Spanner.V1.Tests... any reason for that?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
benwulfe: Maybe instead of this:
        private readonly ApiCall<CreateSessionRequest, Session> _callCreateSession;
...
            _callCreateSession = clientHelper.BuildApiCall<CreateSessionRequest, Session>(
                GrpcClient.CreateSessionAsync, GrpcClient.CreateSession, effectiveSettings.CreateSessionSettings);

we could have this:
        private readonly IApiCall<CreateSessionRequest, Session> _callCreateSession; <=IApiCall
...
            _callCreateSession = new PassThroughProxy( clientHelper.BuildApiCall<CreateSessionRequest, Session>(
                GrpcClient.CreateSessionAsync, GrpcClient.CreateSession, effectiveSettings.CreateSessionSettings));
             Modify_CallProxy(nameof(GrpcClient.CreateSessionAsync), ref _callCreateSession);

Then, I could implement the partial method Modify_CallProxy to wrap the underlying api call to do either exception filtering, pooling, or whatever.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClient.Streaming.cs
benwulfe: ahhh -- so its completely ok to modify the client classes.  I misunderstood you then.

benwulfe:
re: ProirityList tests in Data was just an oversight.  Also, sessionpool tests are in data (which is wrong).
If you like, you can leave that for me for my next change and I will refactor the appropriate tests where they belong.

jskeet:
Yup, let's leave the test movement to another PR.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ExecuteHelper.cs
```diff
@@ -55,7 +62,10 @@ internal static bool CheckForSessionExpiredError(this RpcException rpcException,
             return false;
         }
 
-        internal static bool IsSessionExpiredError(this RpcException rpcException)
+        /// <summary>
+        /// Determines whether <paramref name="rpcException"/> is due to a session expiry.
+        /// </summary>
+        public static bool IsSessionExpiredError(this RpcException rpcException)
```
benwulfe: unfortunately, yes and its where I got it.

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/df975b0354273cc984c8cda2f31d2697b770b3b3/google-cloud-spanner/src/main/java/com/google/cloud/spanner/SessionPool.java#L689

jskeet:
Rebased from master - had to fix up a couple of things that had changed with other PRs. Should still be okay though...

Anything else required before merging? (I think we should look at the ApiCall hooks separately to this PR.)

PR #1217, opened 2017-07-12
Remove EndSpan and rely on the IDisposable from StartSpan

### LINKED ISSUE #1124: ASP.NET TraceHeaderPropagatingHandler fails
The TraceHeaderPropagatingHandler on ASP.NET fails.  It appears to be an issue with no span being available after the request has finished.

reported by @jsimonweb 

To repo:

git clone -b addTrace https://github.com/jsimonweb/csharp-docs-samples.git
cd csharp-docs-samples/trace/
Open Trace.sln (build and run)

iantalarico:
Remove EndSpan from the IManagedTracer and only rely on the IDisposable from StartSpan to end a span in a trace. The IDisposable from StartSpan will contain more information about the span and allow for better handling of Spans across threads.

Fixes #1124


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -496,16 +498,178 @@ public void MultipleSpans_Dispose()
...
+                    tcs1.SetResult(true);
+                }
+            });
+            Thread t2 = new Thread(() =>
+            {
+                try
+                {
+                    if (expectedException == null)
+                    {
+                        var temp = tcs1.Task.Result;
```
evildour: Can you just use `Wait()` here.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -496,16 +498,178 @@ public void MultipleSpans_Dispose()
...
+        /// the second.
+        /// </summary>
+        /// <param name="actionOne">The first action to run.</param>
+        /// <param name="actionTwo">The second action to run.  If expectedException is not null, this action
+        ///     is expected to throw an exception with the type of expectedException.</param>
+        /// <param name="expectedException">Optional, an exception excepted to be thrown in actionTwo.</param>
+        private async Task RunInDisjointThreads(Action actionOne, Action actionTwo, Exception expectedException = null)
+        {
+            var tcs1 = new TaskCompletionSource<bool>();
+            var tcs2 = new TaskCompletionSource<bool>();
```
evildour: I think you can do away with these and just use `Task.Run` instead of creating threads.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -496,16 +498,178 @@ public void MultipleSpans_Dispose()
...
-            tracer.StartSpan("span-name-0");
-            tracer.StartSpan("span-name-1");
-            tracer.StartSpan("span-name-2");
-            tracer.EndSpan();
-            tracer.EndSpan();
+            var one = tracer.StartSpan("span-name-0");
+            var two = tracer.StartSpan("span-name-1");
+            var three = tracer.StartSpan("span-name-2");
+            three.Dispose();
+            two.Dispose();
```
evildour: Use usings for these two?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/ISpan.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Google.Cloud.Diagnostics.Common
+{
+    /// <summary>
+    /// A trace span.
+    /// <remark>
+    /// The functions here, aside from <see cref="IDisposable.Dispose"/>, do not need to be used in most cases. 
+    /// They need to be used when updating the current span or starting a new span when the current span in a
```
evildour: "...the current span **is** in a..."

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/ISpan.cs
```diff
@@ -0,0 +1,83 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Google.Cloud.Diagnostics.Common
+{
+    /// <summary>
+    /// A trace span.
+    /// <remark>
```
evildour: This should be remark**s**

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/ISpan.cs
```diff
@@ -0,0 +1,83 @@
...
+    ///         span1.AnnotateSpan(new Dictionary&lt;string, string&gt; { { "new", "label"} });
+    ///         span1.Dispose();
+    ///     });
+    ///     
+    ///     t.Start();
+    ///     t2.Start();
+    /// }
+    /// </code>
+    /// </remark>
+    /// </summary>
```
evildour: Move the summary close above the start of the remarks.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/ISpan.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        /// <summary>
+        /// Gets span's id.
+        /// </summary>
+        ulong SpanId();
+
+        /// <summary>
+        /// Creates an <see cref="IManagedTracer"/> where the parent of all root spans is this <see cref="ISpan"/>.
+        /// </summary>
+        IManagedTracer CreateManagedTracer();
```
evildour: This is an interesting choice. I don't necessarily disagree with it, but did you also consider instead having a `StartSpan` defined on this for creating child parents explicitly parented to this one? You could then use that in the `SimpleManagedTracer`. Unless I'm missing some other benefit to having this, in which case please let me know.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/NullManagedTracer.cs
```diff
@@ -27,9 +27,39 @@ public sealed class NullManagedTracer : IManagedTracer
         /// <summary>
         /// A disposable class that does nothing when disposed.
         /// </summary>
-        private class NullDisposable : IDisposable
+        internal class Span : ISpan
         {
-            public void Dispose() { }
+            private bool _disposed = false;
```
evildour: Nit: you don't need to initialize this to false.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -30,15 +31,50 @@ namespace Google.Cloud.Diagnostics.Common
     internal sealed class SimpleManagedTracer : IManagedTracer
     {
         /// <summary>
-        /// A class that represents a running trace span.
+        /// A class that represents a trace span.
         /// </summary>
-        private class Span : IDisposable
+        public sealed class Span : ISpan
```
evildour: Does this need to be public? It's fine if it does, just curious.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -30,15 +31,50 @@ namespace Google.Cloud.Diagnostics.Common
...
+            {
+                _tracer = GaxPreconditions.CheckNotNull(tracer, nameof(tracer));
+                TraceSpan = GaxPreconditions.CheckNotNull(traceSpan, nameof(traceSpan));
+            }
 
             /// <summary> Ends the current span.</summary>
-            public void Dispose() => _tracer.EndSpan();
+            public void Dispose()
+            {
+                lock (_mutex)
```
evildour: Hmm, I don't like having to lock here. Would it be possible to have some internal representation of the `EndTime` that we can do an `Interlocked.CompareExchange` operation on to set it only if it is not set and determine if that succeeded.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -30,15 +31,50 @@ namespace Google.Cloud.Diagnostics.Common
...
+                }
+                _tracer.EndSpan(this);
+            }
+
+            /// <inheritdoc />
+            public bool Disposed() => TraceSpan.EndTime != null;
+
+            /// <inheritdoc />
+            public void AnnotateSpan(Dictionary<string, string> labels)
+                =>_tracer.AnnotateSpan(TraceSpan, labels);
```
evildour: Nit: add space after `=>`

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -210,23 +249,20 @@ public void SetStackTrace(StackTrace stackTrace)
...
+        
         /// <inheritdoc />
         public ulong? GetCurrentSpanId() => GetCurrentSpanId(TraceStack);
 
         /// <summary>
         /// Gets the current span id of the specified stack or null if none exists.
         /// </summary>
-        private ulong? GetCurrentSpanId(ImmutableStack<TraceSpan> traceStack)
-            => traceStack.IsEmpty ? _rootSpanParentId : traceStack.Peek().SpanId;
+        private ulong? GetCurrentSpanId(ImmutableStack<Span> traceStack) =>
```
evildour: This will need to pop off all disposed spans as well: when annotating spans, we don't want to annotate one ended from another thread. Similarly, then starting a span we don't want the new one to be parented to a disposed span.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -210,23 +249,20 @@ public void SetStackTrace(StackTrace stackTrace)
...
+            AnnotateSpan(currentStack.Peek().TraceSpan, TraceLabels.FromStackTrace(stackTrace));
         }
 
         /// <inheritdoc />
-        public string GetCurrentTraceId()
-        {
-            return _trace.TraceId;
-        }
-
+        public string GetCurrentTraceId() =>_trace.TraceId;
```
evildour: Nit: add space after `=>`

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -157,20 +194,22 @@ public T RunInSpan<T>(Func<T> func, string name, StartSpanOptions options = null
...
-            TraceSpan span;
-            currentStack = currentStack.Pop(out span);
-            TraceStack = currentStack;
-            span.EndTime = Timestamp.FromDateTime(DateTime.UtcNow);
+            if (!currentStack.IsEmpty)
+            {
+                Span spanOut = null;
+                while (!currentStack.IsEmpty && currentStack.Peek().Disposed())
+                {
+                    currentStack = currentStack.Pop(out spanOut);
```
evildour: No need to declare `spanOut`. This can just be this now:
```csharp
currentStack = currentStack.Pop(out _);
```

jskeet:
Assigned Mike officially as he's clearly on the case :)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
iantalarico: I think it needs to be changed here and in annotate but when we start a span the parent may already be disposed like in the case when a span starts in a parent after the parent has been disposed.  In the current thread this span will be removed from the stack already but in the child thread it will be just marked as disposed.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
iantalarico: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -225,7 +225,7 @@ public void AnnotateSpan(Dictionary<string, string> labels)
         {
             GaxPreconditions.CheckNotNull(labels, nameof(labels));
 
-            var currentStack = TraceStack;
+            var currentStack = GetUpdatedStack(TraceStack);
             CheckStackNotEmpty(currentStack);
```
evildour: Probably best to re-assign `currentStack` to `TraceStack` at some point here so you only do the work of skipped over a particular disposed span once at most.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -246,23 +246,24 @@ private void AnnotateSpan(TraceSpan span, Dictionary<string, string> labels)
...
         public ulong? GetCurrentSpanId() => GetCurrentSpanId(TraceStack);
 
         /// <summary>
         /// Gets the current span id of the specified stack or null if none exists.
         /// </summary>
-        private ulong? GetCurrentSpanId(ImmutableStack<Span> traceStack) =>
-            traceStack.IsEmpty? _rootSpanParentId : traceStack.Peek().TraceSpan.SpanId;
+        private ulong? GetCurrentSpanId(ImmutableStack<Span> traceStack)
+        {
+            traceStack = GetUpdatedStack(traceStack);
```
evildour: I think I'd actually remove the call to `GetUpdatedStack` here (sorry for the change) and make it the caller's responsibility to get the updated stack and reassign it to `TraceStack`

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -246,23 +246,24 @@ private void AnnotateSpan(TraceSpan span, Dictionary<string, string> labels)
...
-            AnnotateSpan(currentStack.Peek().TraceSpan, TraceLabels.FromStackTrace(stackTrace));
+            AnnotateSpan(TraceLabels.FromStackTrace(stackTrace));
         }
 
         /// <inheritdoc />
-        public string GetCurrentTraceId() =>_trace.TraceId;
+        public string GetCurrentTraceId() => _trace.TraceId;
         
         /// <inheritdoc />
         public ulong? GetCurrentSpanId() => GetCurrentSpanId(TraceStack);
```
evildour: As mentioned in my comment below, make it the caller's responsibility to update `TraceStack`, so this would change to
```csharp
var currentStack = GetUpdatedStack(TraceStack);
var result = GetCurrentSpanId(currentStack);
TraceStack = currentStack;
return result;
```

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet/Trace/ContextInstanceManager.cs
```diff
@@ -18,27 +18,19 @@
...
 
         /// <summary>
-        /// Gets the current <see cref="IManagedTracer"/> or null if non exists.
+        /// Gets the current value for a type.
         /// </summary>
-        public static IManagedTracer GetCurrentTracer()
-        {
-            return (IManagedTracer) HttpContext.Current.Items[TraceKey];
-        }
+        public static T Get<T>() => (T)HttpContext.Current.Items[typeof(T)];
```
evildour: Ok, so there's a potential for a `NullReferenceException` here then if someone specifies a value type that doesn't exist. Maybe it's not a huge concern though.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
evildour: Oh I see. Makes sense.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/ISpan.cs
evildour: Good point about switching. Your approach makes it easier to have common tracing logic that just deals with an `IManagerTracer`.

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet/Trace/ContextInstanceManager.cs
iantalarico: I'm not really concerned here as this is purely an internal class

evildour:
Ok, I think I'm starting to get confused a bit now. We may want to sync up offline about this. But thinking about this some more, I think in some cases you really do want to pop off all disposed items form the start when you start/annotate/end a span. Let's say thread A starts a parent span, spawns thread B while that span is open, and then thread C disposes that parent span. If thread A starts a new span, you don't want it to be a child of that disposed parent. In that case you want it to pop off disposed ones before creating the child. But if thread B starts a new span, you _do_ want it to be a child of the disposed parent since thread B was spawned while the parent span was active. I don't know if there's an easy way to compare execution contexts though to know whether or not to pop off from the stack. I'm going to look into that. Also, things seem to be getting a little complicated now, so I'll try to think of a better way.

iantalarico:
I'm getting confused too, I'll put something on your calendar.

iantalarico:
@evildour I pushed a change (just so you cant TAL before we talk tomorrow) that I think should solve the problem with async tasks.  I don't think we can ever track things across actually threads thought.

iantalarico:
@evildour PTAL

@jskeet Mike and I iterated on a solution for this.  In the end I think we figured the best solution for this is to allow a thread to end in another thread but not guarantee the state of new spans after this.  I think given that this is an uncommon case it's fairly safe for now.  If it becomes and issue in the future we will probably need to revisit and/or create a managed tracer.  Does this sound reasonable?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/ISpan.cs
```diff
@@ -0,0 +1,73 @@
...
+    ///         tcs.Task.Wait();
+    ///         span.AnnotateSpan(new Dictionary&lt;string, string&gt; { { "new", "label"} });
+    ///         span.Dispose();
+    ///     });
+    ///     
+    ///     t.Start();
+    ///     t2.Start();
+    /// }
+    /// </code>
+    /// NOTE: While it is possible to end a span in another thread any new spans after this may be in
```
evildour: Nit: I'd probably wrap the remarks above in a `<para>` tag and put this note within its own `<para>` tag.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -30,15 +31,50 @@ namespace Google.Cloud.Diagnostics.Common
...
+            {
+                _tracer = GaxPreconditions.CheckNotNull(tracer, nameof(tracer));
+                TraceSpan = GaxPreconditions.CheckNotNull(traceSpan, nameof(traceSpan));
+            }
 
             /// <summary> Ends the current span.</summary>
-            public void Dispose() => _tracer.EndSpan();
+            public void Dispose()
+            {
+                GaxPreconditions.CheckState(!Disposed(), "Span cannot be disposed twice.");
```
evildour: Nit: this seems unnecessary.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -30,15 +31,50 @@ namespace Google.Cloud.Diagnostics.Common
...
+
             private readonly SimpleManagedTracer _tracer;
-            public Span(SimpleManagedTracer tracer) { _tracer = tracer; }
+
+            private double _ticks;
+
+            internal Span(SimpleManagedTracer tracer, TraceSpan traceSpan)
+            {
+                _tracer = GaxPreconditions.CheckNotNull(tracer, nameof(tracer));
+                TraceSpan = GaxPreconditions.CheckNotNull(traceSpan, nameof(traceSpan));
```
evildour: Nit: these checks don't seem necessary since you always create it and pass in valid values.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -30,15 +31,50 @@ namespace Google.Cloud.Diagnostics.Common
     internal sealed class SimpleManagedTracer : IManagedTracer
     {
         /// <summary>
-        /// A class that represents a running trace span.
+        /// A class that represents a trace span.
         /// </summary>
-        private class Span : IDisposable
+        internal sealed class Span : ISpan
```
evildour: Can this be private again? If not that's fine.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -157,20 +194,17 @@ public T RunInSpan<T>(Func<T> func, string name, StartSpanOptions options = null
...
+        private void EndSpan(Span span)
         {
             var currentStack = TraceStack;
-            CheckStackNotEmpty(currentStack);
-
-            TraceSpan span;
-            currentStack = currentStack.Pop(out span);
-            TraceStack = currentStack;
-            span.EndTime = Timestamp.FromDateTime(DateTime.UtcNow);
+            if (!currentStack.IsEmpty && span.SpanId() == currentStack.Peek().SpanId())
```
evildour: You can probably just use `&& span == currentStack.Peek()` here.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -292,7 +320,7 @@ private ImmutableStack<TraceSpan> TraceStack
         }
 #endif
 
-        private sealed class ImmutableStack<T>
+        internal sealed class ImmutableStack<T>
```
evildour: Does this need to be internal?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
iantalarico: I prefer to always have checks like this but can remove them if you feel strongly

iantalarico:
All fixed up PTAL

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/ISpan.cs
```diff
@@ -45,8 +47,10 @@ namespace Google.Cloud.Diagnostics.Common
     ///     t2.Start();
     /// }
     /// </code>
```
evildour: Not sure if the code tags need to be in or out of the para tag. You may just want to double-check the help that gets generated.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -50,20 +51,24 @@ internal Span(SimpleManagedTracer tracer, TraceSpan traceSpan)
             /// <summary> Ends the current span.</summary>
             public void Dispose()
             {
-                lock (_mutex)
+                GaxPreconditions.CheckState(!Disposed(), "Span cannot be disposed twice.");
+                if (Interlocked.CompareExchange(ref _ticks, DateTime.UtcNow.Ticks, 0) == 0)
+                {
+                    TraceSpan.EndTime = Timestamp.FromDateTime(new DateTime((long)_ticks).ToUniversalTime());
```
evildour: I think this comment may have fallen through the cracks (or maybe you're waiting to see what Jon says).

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
evildour: No, that's fine.

evildour:
It looks like some of the relevant tests may be failing.

jskeet:
Sorry, hadn't seen the comment asking for more input. I'm afraid I don't remember enough context of what's going on here, but if you and Mike are happy it's the best solution for now, that's fine by me.

evildour:
Yeah, it's unfortunate: we either have this 'partial' fix with a warning not to close spans on other threads or we have a hacky solution of popping off all disposed spans before trying to do anything with the top of the stack, but only if you're in the same thread that started the disposed span(s).

We thought we had a good workaround in the form of capturing the current `ExecutionContext` when starting a span and then running a lambda within that EC to pop off from the stack when closing the span. However, we learned that capturing an `ExecutionContext` gives a point-in-time snapshot of the current `AsyncLocal` values in the logical call context. You can't write back to the originals or get updated values. So only logic within the lambda running in the captured EC saw the span as ended. All other threads thought it was still open.

evildour:
Assuming tests pass now, LGTM.

evildour:
There is an overload of `Interlocked.CompareExchange` that works on longs. Can you use that?

iantalarico:
Did not realizes that, done.

PR #1224, opened 2017-07-13
Use the invariant culture everywhere necessary

### LINKED ISSUE #1223: Spanner tests fail in non-US culture
Currently if I run build.sh on my UK machine, it will run the Spanner tests which have 5 tests that fail like this:

Google.Cloud.Spanner.Data.Tests.TypeTests.TestSerializeToValue(clrValue: "1/31/2017 3:15:30 AM", spannerDbType: Date, expectedJsonValue: "\"2017-01-31\"", testType: ClrToValue) [FAIL]
[xUnit.net 00:00:01.5301906]       , type:String, spannerType:Date String was not recognized as a valid DateTime.

I suspect the problem is that there are hard-coded US-formatted dates that we're parsing. Looking into it.
(Will first check that I see the same problem running just the Spanner tests from the command line. I'm sure I will.)

jskeet:
We now test in the invariant culture, French and US English,
but only on the desktop framework.

Fixes #1223.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/TypeTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/TypeTests.cs
@@ -19,6 +19,8 @@ using System.Linq;
 using Google.Protobuf;
 using Google.Protobuf.WellKnownTypes;
 using Xunit;
+using System.Globalization;
+using System.Threading;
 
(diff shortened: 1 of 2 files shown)


jskeet:
(I don't believe Convert.ToChar and Convert.ToBoolean are culture-sensitive, which explains why there are removals as well as additions. Also, CultureInfo.InvariantCulture was getting very long-winded - I've broken this up into two commits for the sake of review.)

PR #1225, opened 2017-07-13
Session create throttling.  This ensures that we never attempt to cre…

benwulfe:
…ate too many sessions concurrently.

Creating too many sessions concurrently can result in severe perf penalties server side (on the order of several seconds if several hundred sessions are created in one go).  This change throttles all session creates to ensure this cannot happen.

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolImpl.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolImpl.cs
@@ -35,6 +35,7 @@ namespace Google.Cloud.Spanner.V1
         private const int MaximumLinearSearchDepth = 50;
 
         private readonly List<SessionPoolEntry> _sessionMruStack = new List<SessionPoolEntry>();
+        private readonly SemaphoreSlim _sessionCreateThrottle;
         private int _lastAccessTime;
         private static int s_activeSessionsPooled;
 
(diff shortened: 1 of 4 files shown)


benwulfe:
Also -- this is another candidate for something that would be nice to do in an ApiCall interceptor.

evildour:
I may not get able to get to this one until tomorrow, FYI

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolImpl.cs
```diff
@@ -205,12 +207,22 @@ private void Push(SessionPoolEntry entry)
                     sw.Start();
                 }
                 //create a new session, blocking or throwing if at the limit.
-                var result = await Key.Client.CreateSessionAsync(new DatabaseName(Key.Project, Key.Instance, Key.Database), cancellationToken).ConfigureAwait(false);
-                if (sw != null)
+                await _sessionCreateThrottle.WaitAsync(cancellationToken).ConfigureAwait(false);
```
jskeet: How feasible is it that by the time the semaphore has released, there's actually a suitable pooled session anyway? I suspect it's significantly more complex to wait for "either I can create a session or there's a suitable one" but it's worth considering.

benwulfe: that's actually a great suggestion and I don't think its much more complicated.  its just another call to trypop.

benwulfe: and actually there is also a minor issue here in that the perf counter is including the throttle time -- which is 'ok', but I'd prefer to have the raw times to the server since its been known to have issues.

PR #1226, opened 2017-07-13
Spanner cleanup

benwulfe:
Rework/reduction of SpannerColumn change.  This has two changes:
1. Removes an overload of SpannerParameterCollection.Add() that took the type as the third parameter which could easily have been mistaken for the value.  Now SpannerDbType is always the second parameter.

2. Changes some xmldocs for CreateXXXCommand to say "column" to (hopefully) better communicate what needs to be specified -- even if we cannot do the full change for SpannerColumn.  The wording for these might need to be spruced up a bit.

Code cleanup --- just resharper code cleanup where ran (moves methods/vars into standard relative locations, line width enforcement, etc)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -52,148 +52,39 @@ public WriteTests(TestDatabaseFixture testFixture, ITestOutputHelper outputHelpe
...
+        private async Task InsertAndRecordBytesAsync(
+            byte[] bytes,
             IDictionary<string, byte[]> record)
         {
             Assert.Equal(
                 1, await InsertAsync(
                     new SpannerParameterCollection
                     {
-                        new SpannerParameter("BytesValue", SpannerDbType.Bytes, bytes)
+                        new SpannerParameter("BytesValue", SpannerDbType.Bytes) {Value = bytes}
```
evildour: What was wrong with the existing code?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -284,17 +182,133 @@ public WriteTests(TestDatabaseFixture testFixture, ITestOutputHelper outputHelpe
         }
 
         [Fact]
-        public async Task WriteNanValue()
+        public async Task BadColumnName()
```
evildour: Why did everything shift around here? It doesn't appear it was to alphabetize the tests. It makes it difficult to compare diffs when they shift like this. If there was a good reason to do it then that's fine though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -157,19 +127,49 @@ public override string ConnectionString
...
+        /// A SpannerConnectionStringBuilder containing a formatted connection string.  Must not be null.
+        /// </param>
+        public SpannerConnection(SpannerConnectionStringBuilder connectionStringBuilder)
+        {
+            GaxPreconditions.CheckNotNull(connectionStringBuilder, nameof(connectionStringBuilder));
+            TrySetNewConnectionInfo(connectionStringBuilder);
+        }
+
+        /// <summary>
+        /// Begins a readonly transaction using the optionally provided <see cref="CancellationToken" />
```
evildour: Nit: add '.' at end.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -812,9 +790,48 @@ public void Dispose()
...
+        private TimestampBound _timestampBound;
+        private VolatileResourceManager _volatileResourceManager;
+#endif
+
+#if NET45 || NET451
+
+        /// <summary>
+        /// Opens and establishes a session with the Spanner database with the given
+        /// <see cref="TimestampBound" /> settings that control the implicit transaction created
+        /// from the active <see cref="System.Transactions.TransactionScope" />
```
evildour: Nit: add '.' here and at end of next summary.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
```diff
@@ -137,11 +104,56 @@ public override ParameterDirection Direction
...
+        /// <param name="type">
+        /// One of the <see cref="SpannerDbType" /> values that indicates the type of the parameter.
+        /// Must not be null.
+        /// </param>
+        /// <param name="value">An object that is the value of the SpannerParameter. May be null.</param>
+        /// <param name="sourceColumn">
+        /// The name of the DataTable source column (SourceColumn) if this SpannerParameter is
+        /// used in a call to Update. May be null.
+        /// </param>
+        /// <param name="size">The length of the parameter. The value is for informational purposes only.</param>
```
evildour: Can you give some more detail here? I don't know what "for informational purposes only" means.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
benwulfe: nothing.  I missed undoing this as part of the undo left Value being set outside of the ctor.
Both ways are fine and the change can be undone.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
benwulfe: resharper code cleanup.  Apologies, but this move will make later cleanups have a smaller diff.

benwulfe: will put this in a separate review next time.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
benwulfe: made more explicit.  its not used to enforce anything, though it might be used in some datagridview scenarios to enforce some behavior in the UI.  Our library does nothing with size however.

jskeet:
(If Mike's happy with this, I'm happy - I won't review it unless there's something you specifically want me to  look at.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
evildour: So why is it included then in the library?

evildour:
Just one more question about `Size`, LGTM.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
benwulfe: Its on the base dbparameter and likely used in some datagrid scenarios to do some enforcement in the ui (eg on string length).
I suppose its a separate question of should we have it as a ctor arg when you can set it via the property and that's a legit point.  I believe at the time I was comparing it to commands from other providers who passed it in.
I think we can remove it from the ctor...

benwulfe:
mike, going to merge after the ci.  holler if you disagree.

PR #1227, opened 2017-07-14
[DO NOT MERGE] Prototype of Translation v2 (breaking change for models)

jskeet:
This treats models as simple strings. No validation is performed.
Everything defaults to null, meaning "don't specify any model".
If a model is specified as the default for a client, then a null
argument in a method call means "use the default" but an empty string
means "wipe out the default; don't send any model in the request."

All of this is a breaking change, hence the major version bump.
We don't know if we want to do this - this PR is intended to help
the conversation.


evildour:
Would it make sense to have a class with string constants for which models are currently known?

jskeet:
The Translation team has suggested they'd prefer not to do that.

evildour:
Oh, interesting. So using this library, how would I know what string to pass in? I guess in that case I probably would just not specify it.

jskeet:
Yes - basically it would be a case of looking at the docs, and if you have early access to some models, you'd already know about them.

evildour:
I see, makes sense.

jskeet:
(We're going in the alternative direction - closing this PR.)

PR #1229, opened 2017-07-14
Add AdvancedTranslationClient

jskeet:
This adds a new client (AdvancedTranslationClient) for scenarios
where a custom model is required. The model is identified by string rather than by an enum.


apis/Google.Cloud.Translation.V2/Google.Cloud.Translation.V2.IntegrationTests/AdvancedTranslationClientTest.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            var client = AdvancedTranslationClient.Create();
+            var detection = client.DetectLanguage(LargeText);
+            Assert.Equal("en", detection.Language);
+        }
+        
+        [Fact]
+        public void Translate_LargeText()
+        {
+            var client = TranslationClient.Create();
```
chrisdunelm: Is this creating the wrong client type?

jskeet: It is indeed. I don't know how that got past search and replace :( Will fix, thanks.

PR #1236, opened 2017-07-14
Remove the option to log exceptions straight to the error reporting API

iantalarico:
The error reporting API has not gone GA yet so these libraries cannot GA with this option.

This is done by removing the EventTarget for error reporting.  This is the only externally exposed item being remove.  If we want to add back this option when the API GAs we can add back the type and add a function to create an EventTarget with it.

We still have a dependency on the Error Reporting API in the test to get the events.


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/ErrorReporting/ErrorReportingContextExceptionLogger.cs
```diff
@@ -83,64 +91,69 @@ void IContextExceptionLogger.Log(Exception exception, IContextWrapper context)
...
             }
 
             StackFrame frame = frames[0];
-            return new SourceLocation
-            {
-                FilePath = frame.GetFileName() ?? "",
-                LineNumber = frame.GetFileLineNumber(),
-                FunctionName = frame.GetMethod()?.Name ?? "",
-            };
+            Struct sourceLocation = new Struct();
```
jskeet: This can be written as

```csharp
Struct sourceLocation = new Struct {
    Fields = {
        { "filePath", Value.ForString(frame.GetFileName() ?? "") },
        { "lineNumber", Value.ForNumber(frame.GetFileLineNumber()) },
        { "functionName", Value.ForString(frame.GetMethod()?.Name ?? "" }
    }
}
```

(Or you could use indexers instead.)

jskeet:
(Except that Travis and AppVeyor are red, of course :)

iantalarico:
No, the only downside is we need to undo a large chunk of this if we want to add error reporting.  The only public thing that will change is adding an enum value and a function to error target.

Looks like Travis and AppVeyor were issues with default credentials, not sure if it's related but I will see what happens with this run.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/Google.Cloud.Diagnostics.Common.IntegrationTests.csproj
```diff
@@ -11,8 +11,8 @@
     <NoWarn>1701;1702;1705;4014</NoWarn>
   </PropertyGroup>
   <ItemGroup>
-    <ProjectReference Include="..\..\Google.Cloud.ErrorReporting.V1Beta1\Google.Cloud.ErrorReporting.V1Beta1\Google.Cloud.ErrorReporting.V1Beta1.csproj" />
     <ProjectReference Include="..\Google.Cloud.Diagnostics.Common\Google.Cloud.Diagnostics.Common.csproj" />
+    <PackageReference Include="Google.Cloud.ErrorReporting.V1Beta1" Version="1.0.0-beta07" />
```
jskeet: You should be able to add it in testDependencies in apis.json, IIRC.

iantalarico:
@jskeet That does seem to work.  The only issue is it adds it to test and integration tests.

jskeet:
Yup, it's just a simplification to have the same deps for all test projects :)

iantalarico:
works for me :)

jskeet:
Looks like you'll need to fix things up when rebasing to master... but then are we in a state where we can get this in?

iantalarico:
@jskeet What needs to be fixed up? Can't I just squash and merge?

jskeet:
I haven't tried rebasing it, but github believes there are conflicts. I would suggest squashing into a single commit to start with, then rebasing to master... depending on where the conflict was, there may or may not be manual work to do.

PR #1238, opened 2017-07-17
Allow job-creation operations to specify a project ID

### LINKED ISSUE #1237: Allow project ID to be specified for operations that create jobs
This should default to the project ID of the client, but could be overridden in the options for each call (CreateCopyJobOptions, QueryOptions etc).

jskeet:
Fixes #1237.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/QueryOptions.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/QueryOptions.cs
@@ -100,6 +100,12 @@ namespace Google.Cloud.BigQuery.V2
         /// </summary>
         public bool? DryRun { get; set; }
 
+        /// <summary>
+        /// The ID of the project in which to create the job. If this is not set,
+        /// it defaults to the project ID of the client.
+        /// </summary>
+        public string ProjectId { get; set; }
+
         internal void ModifyRequest(JobConfigurationQuery query)
         {
             // Note: no validation of combinations (flatten results etc). Leave this to the server,
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadAvroOptions.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadAvroOptions.cs
@@ -41,6 +41,12 @@ namespace Google.Cloud.BigQuery.V2
         public CreateDisposition? CreateDisposition { get; set; }
 
         /// <summary>
+        /// The ID of the project in which to create the job. If this is not set,
+        /// it defaults to the project ID of the client.
+        /// </summary>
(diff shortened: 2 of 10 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/QueryOptions.cs
```diff
@@ -100,6 +100,12 @@ public sealed class QueryOptions
         /// </summary>
         public bool? DryRun { get; set; }
 
+        /// <summary>
+        /// The ID of the project in which to create the job. If this is not set,
```
chrisdunelm: Probably not "create the job"

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadAvroOptions.cs
```diff
@@ -41,6 +41,12 @@ public sealed class UploadAvroOptions
         public CreateDisposition? CreateDisposition { get; set; }
 
         /// <summary>
+        /// The ID of the project in which to create the job. If this is not set,
```
chrisdunelm: Probably not "create the job"

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadCsvOptions.cs
```diff
@@ -87,6 +87,12 @@ public sealed class UploadCsvOptions
         /// </summary>
         public string NullMarker { get; set; }
 
+        /// <summary>
+        /// The ID of the project in which to create the job. If this is not set,
```
chrisdunelm: Probably not "create the job"

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadJsonOptions.cs
```diff
@@ -52,6 +52,11 @@ public sealed class UploadJsonOptions
         /// </summary>
         public bool? Autodetect { get; set; }
 
+        /// <summary>
+        /// The ID of the project in which to create the job. If this is not set,
```
chrisdunelm: Probably not "create the job"

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/QueryOptions.cs
jskeet: Why not? All these operations create a job, and the job is owned by a project. (So I might create a job in my project, even though it's querying public data.) I suspect in theory it's possible for three projects to be involved: "Create a job in project A to copy a table from project B to project C." (So long as the credentials allow that, of course.)

Any suggested wording that would get that idea across more clearly?

PR #1242, opened 2017-07-18
Spanner Threadsafe fixes.

benwulfe:
This makes async methods static where appropriate.
This ensures that the async method runs on local state and eliminates a few race conditions that existed.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
@@ -216,32 +216,56 @@ namespace Google.Cloud.Spanner.Data
 
         /// <summary>
         /// Executes the command against the <see cref="SpannerConnection"/>.
+        /// This method is thread safe.
         /// </summary>
         /// <returns>a <see cref="SpannerDataReader"/>.</returns>
-        public async Task<DbDataReader> ExecuteReaderAsync(
+        public Task<DbDataReader> ExecuteReaderAsync(
             TimestampBound singleUseReadSettings,
             CancellationToken cancellationToken = default(CancellationToken))
+            => ExecuteReaderImplAsync(
+                SpannerConnection,
+                GetExecuteSqlRequest(),
+                CommandTimeout,
+                GetSpannerTransaction(),
+                Transaction,
+                singleUseReadSettings,
+                cancellationToken
+                );
+
+        private static async Task<DbDataReader> ExecuteReaderImplAsync(
+            SpannerConnection spannerConnection,
+            ExecuteSqlRequest request,
+            int commandTimeout,
+            ISpannerTransaction spannerTransaction,
+            DbTransaction transaction,
+            TimestampBound singleUseReadSettings,
+            CancellationToken cancellationToken)
         {
             GaxPreconditions.CheckNotNull(singleUseReadSettings, nameof(singleUseReadSettings));
             // There must be a valid and open connection.
-            if (SpannerConnection == null)
+            if (spannerConnection == null)
             {
                 throw new InvalidOperationException(
                     "You must assign a SpannerConnection to this command to execute it.");
             }
-            if (Transaction != null)
+            if (transaction != null)
             {
                 throw new InvalidOperationException(
                     "singleUseReadSettings cannot be used within"
                     + " another transaction.");
             }
             var singleUseTransaction =
-                await SpannerConnection.BeginSingleUseTransactionAsync(
+                await spannerConnection.BeginSingleUseTransactionAsync(
                     singleUseReadSettings,
                     cancellationToken).ConfigureAwait(false);
-            return await ExecuteDbDataReaderAsync(
-                CommandBehavior.Default, cancellationToken,
-                singleUseTransaction).ConfigureAwait(false);
+            return await ExecuteDbDataReaderImplAsync(
+                spannerConnection,
+                request,
+                commandTimeout,
+                spannerTransaction,
+                CommandBehavior.Default,
+                singleUseTransaction,
+                cancellationToken).ConfigureAwait(false);
         }
 
         /// <inheritdoc />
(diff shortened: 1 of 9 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -358,13 +411,7 @@ public override Task<int> ExecuteNonQueryAsync(CancellationToken cancellationTok
...
                 mutations.Add(new Mutation {Delete = w});
             }
-
-            // Make the request. This will commit immediately or not depending on whether a transaction was explicitly created.
-            await GetSpannerTransaction().ExecuteMutationsAsync(mutations, cancellationToken)
-                .WithTimeout(TimeSpan.FromSeconds(CommandTimeout), "The timeout of the SpannerCommand was exceeded.")
-                .ConfigureAwait(false);
-            // Return the number of records affected.
-            return mutations.Count;
+            return mutations;
```
jskeet: Off-PR note: if you need to build key-sets a bit like this in multiple places, let's separate that out. (Aside from anything else, we could add `ListValue.FromValues` and `KeySet.FromKeys` methods, which would reduce the nesting by two...)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -428,40 +488,75 @@ public override void Prepare()
...
-                await SpannerConnection.OpenAsync(cancellationToken).ConfigureAwait(false);
+                throw new InvalidOperationException("Unable to open the Spanner connection to the database.");
             }
 
-            if (!SpannerConnection.IsOpen)
+            var tx = singleUseTransaction ?? spannerTransaction;
+
+            // Execute the command.
+            var resultSet = await tx.ExecuteQueryAsync(request, cancellationToken)
+                .WithTimeout(TimeSpan.FromSeconds(commandTimeout), "The timeout of the SpannerCommand was exceeded.")
```
jskeet: I've seen that this is a locally enforced timeout. Do we pass the timeout to gRPC as well? I think we should - at which point I *think* it's enforced locally already. Let's chat about that.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
```diff
@@ -230,11 +230,12 @@ private void FillSpannerInternalTypes(MapField<string, V1.Type> typeDictionary)
             }
         }
 
-#if NET45 // It's odd that these are required, but it looks like it's due to a discrepancy
-// between the reference assemblies and real assemblies. See
-// https://stackoverflow.com/questions/44197176 for details.
-// Fortunately the real implementations all just return false too.
-/// <inheritdoc />
+#if NET45
+        // It's odd that these are required, but it looks like it's due to a discrepancy
```
jskeet: We should probably remove this comment now - it looks like it's *actually* a difference between .NET 4.5 and 4.5.1, and the docs being wrong. (I think the old project.json version used "whatever .NET 4.5 version you had installed" rather than the reference assemblies.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -262,46 +286,75 @@ public override Task<int> ExecuteNonQueryAsync(CancellationToken cancellationTok
...
+            if (!spannerConnection.IsOpen)
             {
                 throw new InvalidOperationException(
                     "Unable to open the Spanner connection to the database to execute the query.");
             }
 
-            // Execute the command.
+            // Make the request. This will commit immediately or not depending on whether a transaction was explicitly created.
+            await spannerTransaction.ExecuteMutationsAsync(mutations, cancellationToken)
+                .WithTimeout(TimeSpan.FromSeconds(commandTimeout), "The timeout of the SpannerCommand was exceeded.")
```
evildour: Uses `nameof(SpannerCommand)` (and similar in other places below)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -216,32 +216,56 @@ public override int ExecuteNonQuery() => ExecuteNonQueryAsync(_synchronousCancel
 
         /// <summary>
         /// Executes the command against the <see cref="SpannerConnection"/>.
+        /// This method is threadsafe.
```
evildour: "thread safe" perhaps for the doc comments?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -107,7 +117,10 @@ public sealed class SpannerTransaction : DbTransaction, ISpannerTransaction
                 {
                     var taskCompletionSource = new TaskCompletionSource<int>();
                     cancellationToken.ThrowIfCancellationRequested();
-                    _mutations.AddRange(mutations);
+                    lock (_mutations)
+                    {
+                        _mutations.AddRange(mutations);
+                    }
                     taskCompletionSource.SetResult(mutations.Count);
```
evildour: Should this count be obtained within the lock? Seems like there's a race here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
benwulfe: dont think the need is that great yet -- mostly just this one method that does it

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
benwulfe: it is, that's true.  however an overall timeout did catch the deadlock that affected transactionscopes -- so I'm sort of on the fence on yanking it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
benwulfe: fair enough

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
benwulfe: nah thats the local count -- its ok

evildour: Oh right, didn't look closely enough.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
benwulfe: actually this is a different timeout which is modifiable after the client is created (which is sort of a shame, but the design of ado.net).  So this means its possible for someone to set a timeout on the command which is different than the one on the client.
let's follow up on this after this change?  This change is altering the timeout, just ensuring it doesn't change in the middle of a call by snapping its state.

jskeet: We can definitely follow up later - but the generated code allows per-call settings; it doesn't all have to be set up at the very start.

PR #1244, opened 2017-07-18
Temporary: Update the appveyor.yml file to point to another nuget host

iantalarico:
This is to work around an ongoing appveyor outage: https://appveyor.statuspage.io/incidents/m2vdvw39kdk8

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,6 +11,10 @@ branches:
 
 init:
   - git config --global core.autocrlf true
+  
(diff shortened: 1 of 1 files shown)


jskeet:
If it works, I'm happy to get it in. I was going to wait until the outage had been sorted, with an assumption it wouldn't be too long :)

iantalarico:
@jskeet It looks like it's working, if appveyor passed I merge.  It's almost been 24 hours (which means it will probably be fixed in 20 min) but I have a few PRs I want to get in.

jskeet:
SGTM - go for it :)

PR #1259, opened 2017-07-19
Order APIs to generate alphabetically

jskeet:


jskeet:
(Will run this locally to check it's a no-op.)

PR #1260, opened 2017-07-19
Expose SpannerDataReader from ExecuteReaderAsync.

benwulfe:
By default, the base class dbdatareader is returned, which does not expose any of our suggested APIs -- such as GetFieldValue(string column).
This change simply overrides the base version to cast the result as a spannerdatareader.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
@@ -244,6 +244,30 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         }
 
         [Fact]
+        public async Task ReadAllowsNewApis()
+        {
+            // ReSharper disable once RedundantAssignment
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
```diff
@@ -244,6 +244,30 @@ private class CredentialWrapper : ITokenAccess
         }
 
         [Fact]
+        public async Task ReadAllowsNewApis()
+        {
+            // ReSharper disable once RedundantAssignment
+            int rowsRead = -1;
```
evildour: Is this set to -1 here and then 0 below to distinguish different kinds of failures by looking at the test error? Perhaps a string or enum describing the 3 possible states would be better then.

benwulfe: thats a good suggestion and would make it clearer.  a few tests use this pattern.  i'll get around to cleaning them up a little in the future.

PR #1261, opened 2017-07-19
scope.Complete for ReadOnlyTransactions should no-op.

benwulfe:
found during doc sample writing.
We should simply accept the scope.Complete because there may be another resource that is being used.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
@@ -52,6 +52,10 @@ namespace Google.Cloud.Spanner.Data
                 //timestamp bound read and doesn't have anything to commit.
                 Logger.Debug(() => "Received a COMMIT for a two phase commit but without changes. This is allowed.");
                 enlistment.Done();
+
+                // For write transactions with no mutations, this ensures the transaction
+                // gets closed and the locks get released if any were made.
+                CommitToSpanner();
                 return;
             }
             Logger.Warn(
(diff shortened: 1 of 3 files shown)


jskeet:
Sorry - I hadn't realized that I hadn't submitted my review. (Probably due to a dodgy network connection.)

Basically, I don't understand TransactionScope much either.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
```diff
@@ -115,8 +115,15 @@ public void SinglePhaseCommit(SinglePhaseEnlistment singlePhaseEnlistment)
         {
             try
             {
-                ExecuteHelper.WithErrorTranslationAndProfiling(() =>
-                    _transaction?.Commit(), "VolatileResourceManager.Commit");
+                // If we have no changes, then just tell the outer transaction that everything is good.
+                // This can happen with a readonly transaction or a write transaction where we never
+                // executed any mutations.
```
benwulfe: Readonly transactions do not take locks -- but readwrite transactions may actually have locks if a read was previously done on it.  So while we do a prewarm that would create a new transaction, thats a bit fragile to rely on.  So I'm going to narrow the logic here to only ignore readonly transactions -- so empty readwrite tx will still get committed.

jskeet: So we're certain that the Spanner team is happy never to hear about a read-only transaction? I can imagine it being "safe but a bit rude" (if that wastes server-side resources) - but I may very well be wrong.

benwulfe:
I'm double checking on that as well.

benwulfe:
"for readonly tx there is no need to cal [sic] rollback"

PR #1264, opened 2017-07-20
adds drop command as a valid ddl command

benwulfe:
fixes a related bug in spannercommandtextbuilder
adds tests

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
@@ -14,6 +14,7 @@
 
 using System;
 using System.Linq;
+using System.Text;
 using Google.Api.Gax;
 
 // ReSharper disable UnusedParameter.Local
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -34,7 +35,9 @@ public sealed class SpannerCommandTextBuilder
         private const string AlterCommand = "ALTER";
         private const string CreateCommand = "CREATE";
         private const string DropCommand = "DROP";
+        //TODO(benwu): verify if whitespace between CREATE/DROP and DB is allowed.
         private const string CreateDatabaseCommand = "CREATE DATABASE";
+        private const string DropDatabaseCommand = "DROP DATABASE";
```
jskeet: Should you have a space after this? At the moment it *looks* like a DDL of `DROP DATABASEFOO` will have the same effect as `DROP DATABASE FOO` and presumably it shouldn't.

(Ditto CREATE)

benwulfe: yep --- and I should have a bunch of negative tests as well.

benwulfe: also its probably safer to fail on these sorts of statements:

DROP DATABASE NAMEWITH ANACCIDENTALSPACE

than to drop "NAMEWITH"

benwulfe:
added space and more test coverage.

PR #1284, opened 2017-07-28
Fixes reference of Common to point to public nuget packages instead o…

benwulfe:
…f project references.

This allows us to publish Common without worrying about it possibly referencing updated internal bits.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -126,15 +126,15 @@
 
   {
     "id": "Google.Cloud.Diagnostics.Common",
-    "version": "1.0.0-beta04",
+    "version": "1.0.0-beta05",
     "type": "other",
     "targetFrameworks": "netstandard1.5;net45",
     "testTargetFrameworks": "netcoreapp1.0;net452",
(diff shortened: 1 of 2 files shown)


apis/apis.json
```diff
@@ -126,15 +126,15 @@
...
+    "version": "1.0.0-beta05",
     "type": "other",
     "targetFrameworks": "netstandard1.5;net45",
     "testTargetFrameworks": "netcoreapp1.0;net452",
     "description": "Google Stackdriver Instrumentation Libraries Common Components.",
     "tags": [ "Error", "Reporting", "Stackdriver", "ExceptionLogger", "Trace", "Diagnostics" ],
     "dependencies": {
-      "Google.Cloud.Logging.V2": "",
-      "Google.Cloud.Trace.V1": "",
+      "Google.Cloud.Logging.V2": "1.0",
```
jskeet: I need to look at all of this more carefully, but we should at least be referencing 2.0.0 rather than 1.0.
(In general I'd prefer to keep it as project references, but make sure we release updated bits when we need to. I do understand the concern though, and I'll keep thinking about it.)

PR #1296, opened 2017-07-31
CI code coverage using dotcover and codecov.

benwulfe:
This currently only runs coverage on unit tests for CI (not integration tests).

Jon -- I'm testing the appveyor integration with codecov.  I've tested runcoverage.sh with reportgenerator and it's looking good. 

Another thing we can do is run this with integrationtests, generating a combined report and publishing the results.  If codecov doesn't pan out, we can still get reportgenerator output somewhere, which would be good.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,14 +34,20 @@ build_script:
   - dotnet --info
   - echo "Regenerating projects: if this fails, run generateprojects.sh and commit changes"
   - generateprojects.sh && git diff --exit-code
-  - bash build.sh
+  - bash build.sh --notests
   - cd docs
   - bash builddocs.sh
+  - cd ..
 
-# TODO: Reinstate these...
-#  - bash runcoverage.sh
-#  - bash coveralls.sh
+# scripts to run before tests
+before_test:
+  - choco install codecov
 
+# scripts to run after tests
+after_test:
+  - bash createcoveragereport.sh
+  - codecov -f "coverage/coverage.xml"
 
-# The tests are run as part of the build.
-test: off
+# to run your custom scripts instead of automatic tests
+test_script:
+  - bash runcoverage.sh
--- a/build.sh
+++ b/build.sh
@@ -21,8 +21,20 @@ dotnet publish -c Release -f netstandard1.3 tools/Google.Cloud.Tools.Analyzers/G
 # Command line arguments are the APIs to build. Each argument
 # should be the name of a directory, either relative to the location
 # of this script, or under apis.
-apis=$@
-if [ -z "$apis" ]
+apis=()
+runtests=true
+while (( "$#" )); do
+  if [[ "$1" == "--notests" ]]
+  then 
+    echo "Not Running Tests..."
+    runtests=false
+  else 
+    apis+=($1)
+  fi
+  shift
+done
+
+if [ ${#apis[@]} -eq 0 ]
 then
   apis="tools $(echo apis/Google.* | sed 's/apis\///g')"
 fi
@@ -77,9 +89,12 @@ dotnet restore AllProjects.sln
 echo "$(date +%T) Building solution"
 dotnet build -c Release AllProjects.sln
 
-# Could use xargs, but this is more flexible
-while read testproject
-do  
-  echo "$(date +%T) Testing $testproject"
-  dotnet test -c Release $testproject
-done < AllTests.txt
+if [[ "$runtests" = true ]]
+then
+  # Could use xargs, but this is more flexible
+  while read testproject
+  do  
+    echo "$(date +%T) Testing $testproject"
+    dotnet test -c Release $testproject
+  done < AllTests.txt
+fi
--- a/createcoveragereport.sh
+++ b/createcoveragereport.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+
+# This script merges all coverage files found under the 'coverage' directory
+# and then creates a dotcover XML report and from that creates a browsable
+# ReportGenerator html report.
+
+# Use an appropriate version of nuget... preferring
+# first an existing NUGET variable, then NuGet, then
+# just falling back to the path.
+if [ -z "$NUGET" ]
+then
+  if [ -n "$NuGet" ]
+  then
+    NUGET="$NuGet"
+  else
+    NUGET="nuget"
+  fi
+fi
+set -e
+
+$NUGET install -Verbosity quiet -OutputDirectory packages -Version 2017.1.20170613.162720 JetBrains.dotCover.CommandLineTools
+$NUGET install -Verbosity quiet -OutputDirectory packages -Version 2.4.5.0 ReportGenerator
+
+DOTCOVER=$PWD/packages/JetBrains.dotCover.CommandLineTools.2017.1.20170613.162720/tools/dotCover.exe
+REPORTGENERATOR=$PWD/packages/ReportGenerator.2.4.5.0/tools/ReportGenerator.exe
+FIND=/usr/bin/find
+
+if [ ! -d coverage ]
+then
+ mkdir coverage
+fi
+
+[ -f coverage/all.dvcr ] && rm coverage/all.dvcr
+
+echo "Merging reports..."
+$DOTCOVER merge -Source=$(echo coverage/*.dvcr | sed 's/ /;/g') -Output=coverage/all.dvcr ""
+
+echo "Generating detailed xml report..."
+$DOTCOVER report -Source=coverage/all.dvcr -Output=coverage/coverage.xml -ReportType=DetailedXML ""
+
+echo "Running ReportGenerator to create an html report..."
+
+$REPORTGENERATOR \
+   -reports:coverage/coverage.xml \
+   -targetdir:coverage/report \
+   -verbosity:Error
+
+# TODO(benwu): upload coverage/report to public blob storage
--- a/runcoverage.sh
+++ b/runcoverage.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# THIS DOES NOT WORK YET!
+# We use dotcover instead of OpenCover.
 # OpenCover doesn't support portable pdbs, and we don't
 # really want to go back to full pdbs.
 # See https://github.com/OpenCover/opencover/issues/601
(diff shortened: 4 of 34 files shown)


appveyor.yml
```diff
@@ -5,6 +5,9 @@ version: 1.0.{build}-appveyor
 
 image: Visual Studio 2017
 
+before_build:
```
ILMTitan: We can probably change this from `before_build:` to `install:`. 

From [appveyor.yml documentation](https://www.appveyor.com/docs/appveyor-yml/):
```
# scripts that run after cloning repository
install:
  # by default, all script lines are interpreted as batch
  - echo This is batch
  # to run script as a PowerShell command prepend it with ps:
  - ps: Write-Host 'This is PowerShell'
  # batch commands start from cmd:
  - cmd: echo This is batch again
  - cmd: set MY_VAR=12345
```

appveyor.yml
```diff
@@ -37,11 +40,9 @@ build_script:
   - bash build.sh
   - cd docs
   - bash builddocs.sh
-
-# TODO: Reinstate these...
-#  - bash runcoverage.sh
-#  - bash coveralls.sh
-
+  - cd ..
+  - bash runcoverage.sh
```
ILMTitan: Why not make this part of `test_script:` ?

appveyor.yml
benwulfe: before_build is what codecov recommends.  I don't really have a preference as long as its installed before its used.

appveyor.yml
benwulfe: good question.  this is where docs are created and where code coverage was run previously.

benwulfe: I'm also not against moving this, but I want to make sure its passing appveyor first.

runcoverage.sh
```diff
@@ -23,40 +23,67 @@ then
...
-    -searchdirs:$testdir/bin/$CONFIG/net452 \
-    -register:user \
-    -excludebyfile:$generatedFiles
 
+  echo "<CoverageParams>" > "coverage/$api.xml"
+  echo " <TargetExecutable>/Program Files/dotnet/dotnet.exe</TargetExecutable>" >> "coverage/$api.xml"
+if [ -d "$testdir/bin/Release/net452" ] ; then
+  echo " <TargetArguments>test --no-build -f net452 -c Release</TargetArguments>" >> "coverage/$api.xml"
+else
+  echo " <TargetArguments>test --no-build -f dotnet -c Release</TargetArguments>" >> "coverage/$api.xml"
```
jskeet: Is this meant to be netcoreapp1.0?

runcoverage.sh
```diff
@@ -23,40 +23,67 @@ then
...
 
 while read csproj
 do
   testdir=$(dirname $csproj)
   api=$(basename $(realpath $testdir/..))
+
+  [ ! -d "$testdir/bin/Release/net452" ] && [ ! -d "$testdir/bin/Release/netcoreapp1.0" ] && continue
+
   echo "Running coverage for $api"
-  # OpenCover is picky about how it finds the excluded files. There may be a better approach than this,
```
jskeet: Can we do a similar exclusion for generated files? Without that, the numbers we get out aren't going to be as useful.

runcoverage.sh
```diff
@@ -23,40 +23,67 @@ then
...
-    -mergeoutput \
-    -hideskipped:File \
-    -output:coverage/coverage.xml \
-    -filter:"+[$api]*" \
-    -oldStyle \
-    -searchdirs:$testdir/bin/$CONFIG/net452 \
-    -register:user \
-    -excludebyfile:$generatedFiles
 
+  echo "<CoverageParams>" > "coverage/$api.xml"
```
jskeet: For later, would it be useful to generate this file as coverage.xml within the project generator? That might make it easier to run in an ad hoc manner from the command line. ("I've made my changes, let's run the coverage tool just for that one API.")

runcoverage.sh
```diff
@@ -23,40 +23,67 @@ then
...
+echo "<Output>coverage.xml</Output>" >> "coverage/report.xml"
+echo "<ReportType>DetailedXML</ReportType>" >> "coverage/report.xml"
+echo "</ReportParams>" >> "coverage/report.xml"
+
+$DOTCOVER report coverage/report.xml
+
+#Manual CodeCov integration
+#curl -s https://codecov.io/bash > codecov
+#chmod +x codecov
+#./codecov -f "coverage/coverage.xml" -t 8d9d6796-50f9-4b32-898e-fd6f18eb3769
```
jskeet: Is this token okay to include in source control, or should it be a secret that's decrypted based on an AppVeyor-stored key?

appveyor.yml
jskeet: We have "test: off" as tests run as part of build.sh already. We might want to change build.sh to accept a --no-test flag or similar so that we don't end up running all the tests twice... (I'm not sure whether we used to run all the tests twice or not. Back then there was less code to build...)

runcoverage.sh
jskeet: Looking at https://www.jetbrains.com/help/dotcover/dotCover__Console_Runner_Commands.html#cover, there is filtering available, but it looks like it's based on built artifacts (type, member, module) rather than source. Hmm. I'll investigate this.

runcoverage.sh
```diff
@@ -23,40 +23,67 @@ then
...
+
+$DOTCOVER merge coverage/merge.xml
+
+echo "<ReportParams>" > "coverage/report.xml"
+echo "<Source>all.dvcr</Source>" >> "coverage/report.xml"
+echo "<Output>coverage.xml</Output>" >> "coverage/report.xml"
+echo "<ReportType>DetailedXML</ReportType>" >> "coverage/report.xml"
+echo "</ReportParams>" >> "coverage/report.xml"
+
+$DOTCOVER report coverage/report.xml
```
jskeet: Just use command line parameters for this and maybe merge? Building an XML file for the cover command seems reasonable (and is useful for per-API ad hoc testing), but I think we can probably do merge/report more readably without the XML.

runcoverage.sh
jskeet: We can get quite a long way using

```xml
  <AttributeFilters>
    <AttributeFilterEntry>System.Diagnostics.DebuggerNonUserCodeAttribute</AttributeFilterEntry>
  </AttributeFilters>
```

to exclude *most* of the protobuf code, but really we want to filter by source file. I'm asking a JetBrains contact...

jskeet: There's a feature request for this: https://youtrack.jetbrains.com/issue/DCVR-2211
However, as it was filed in 2010, with the last update in 2014, it's not promising :( There may be something we could manually do with the XML output...

An alternative is to add `[CompilerGenerated]` or similar to *all* the generated code - the trouble is, we don't want to add it to the class itself, as parts of the class may *not* be generated :(

runcoverage.sh
benwulfe: i like it.
what do you think about including integration tests optionally as well?
We can make this script scan all coverage[Foo].xml files under the api and run each.

So when you run the generator, it will create just the version for tests.  But you are free to add additional coverage files to include other projects.

runcoverage.sh
benwulfe: I thought dotnet worked for core as well.  I'll check

runcoverage.sh
benwulfe: whenever I tried this I had two problems:
1. escaping the text became quite a PITA (admittedly less of a problem with 'report').
2. the command line always errorred out that the config file was not present.

runcoverage.sh
benwulfe: it can be removed as its not needed when executed in appveyor.

runcoverage.sh
jskeet: I was thinking we could generate a coverage.xml in each test project directory (i.e. as a peer of Google.Cloud.Storage.V1.Tests.csproj). So to run the tests, you just go into the right directory and run "dotcover cover coverage.xml". But we can definitely tweak the details.

runcoverage.sh
jskeet: Weird. Will have a play and see what I can discover...

runcoverage.sh
benwulfe: sgtm

runcoverage.sh
benwulfe: nope i was wrong.  switching

runcoverage.sh
benwulfe: there is actually an attribute explicitly for this:
https://msdn.microsoft.com/en-us/library/system.diagnostics.codeanalysis.excludefromcodecoverageattribute(v=vs.110).aspx

but yes -- it requires a change to our generated code.
I suggest we put both this and debuggernonusercode as filters as a first step and we can update the generators after.

jskeet: Hmm. Good to see the attribute. It feels slightly awkward to add that to the generators for a few reasons:

- There are three to add them to - protobuf, gRPC, toolkit - which means they'll be rolled out at different rates
- The toolkit code is meant to look like "the code you'd write by hand if you had to" as far as possible; this will definitely be uglier
- I'm still not sure how we handle the protobuf code like this:

        private static readonly pb::MessageParser<Policy> _parser = new pb::MessageParser<Policy>(() => new Policy());

 There's one of those per proto message, and the lambda expression doesn't get covered unless the parser is used. We can't put the attribute on the field declaration, and I don't *think* you could apply it to the lambda either. Will have to investigate further.

All that said, it seems reasonable that other users may well want to do code coverage involving proto and gRPC generated code too. It's just such a shame when filtering by source code does *exactly* what we want :(

jskeet:
Quick note: when #1301 is merged, we'll need to reapply the changes to the project generator. It shouldn't be hard to do though.

Now looking at what we can do with the XML files in terms of source filtering...

createcoveragereport.sh
```diff
@@ -0,0 +1,61 @@
...
+echo "<Source>all.dvcr</Source>" >> "coverage/report.xml"
+echo "<Output>coverage.xml</Output>" >> "coverage/report.xml"
+echo "<ReportType>DetailedXML</ReportType>" >> "coverage/report.xml"
+echo "</ReportParams>" >> "coverage/report.xml"
+
+$DOTCOVER report coverage/report.xml
+
+echo "Running ReportGenerator to create an html report..."
+
+$REPORTGENERATOR \
```
jskeet: I've just looked at version 2.5.10 of ReportGenerator, and that contains file filters - so we should be able to use our previous "find the generated files" code (no doubt with tweaks for how the filter is expressed) and pass it in there. Yay!

Alternatively, before finding that (grr) I wrote a short tool to remove generated files from the XML file. That may be cleaner, given that the complete set of generated files (after merging) would be huge.

I suggest we get this change merged and then I'll add the tool in another commit.

build.sh
```diff
@@ -21,8 +21,20 @@ dotnet publish -c Release -f netstandard1.3 tools/Google.Cloud.Tools.Analyzers/G
...
 # should be the name of a directory, either relative to the location
 # of this script, or under apis.
-apis=$@
-if [ -z "$apis" ]
+apis=()
+runtests=true
+while (( "$#" )); do
+if [[ "$1" == "--notests" ]]
+then 
+	echo "Not Running Tests..."
```
jskeet: Two spaces rather than tabs?

createcoveragereport.sh
```diff
@@ -0,0 +1,60 @@
...
+  if [ -n "$NuGet" ]
+  then
+    NUGET="$NuGet"
+  else
+    NUGET="nuget"
+  fi
+fi
+
+$NUGET install -Verbosity quiet -OutputDirectory packages -Version 2017.1.20170613.162720 JetBrains.dotCover.CommandLineTools
+$NUGET install -Verbosity quiet -OutputDirectory packages -Version 2.4.5.0 ReportGenerator
```
jskeet: Might as well use 2.5.10, regardless of whether we use file filters.

createcoveragereport.sh
```diff
@@ -0,0 +1,60 @@
...
+
+echo "Generating detailed xml report..."
+
+echo "<ReportParams>" > "coverage/report.xml"
+echo "<Source>all.dvcr</Source>" >> "coverage/report.xml"
+echo "<Output>coverage.xml</Output>" >> "coverage/report.xml"
+echo "<ReportType>DetailedXML</ReportType>" >> "coverage/report.xml"
+echo "</ReportParams>" >> "coverage/report.xml"
+
+$DOTCOVER report coverage/report.xml
```
jskeet: I've worked out how to run this without a configuration file - you just put "" at the end of the command line. Weird, but it works:

```text
dotcover report -Source=all.dvcr -Output=coverage.xml -ReportType=DetailedXML ""
```

(You need `/` from a Windows command line, but `-` from bash.)

createcoveragereport.sh
```diff
@@ -0,0 +1,60 @@
...
+echo "Merging reports..."
+
+echo "<MergeParams>" > "coverage/merge.xml"
+for filename in coverage/*.dvcr; do
+  echo "<Source>$(basename $filename)</Source>" >> "coverage/merge.xml"
+done
+echo "<Output>all.dvcr</Output>" >> "coverage/merge.xml"
+echo "</MergeParams>" >> "coverage/merge.xml"
+
+$DOTCOVER merge coverage/merge.xml
```
jskeet: Given the report part below, we could probably work out how to build the command line for merge, too... 

I think this should work:

```text
dotcover merge -Source=$"(echo coverage/*.dvcr | sed 's/ /;/g') -Output=coverage/all.dcvr ""
```

PR #1298, opened 2017-08-02
Update versions ready for release

jskeet:
Google.Cloud.Logging.Log4Net now has package references rather than
project references.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -19,7 +19,7 @@
     "id": "Google.Cloud.BigQuery.V2",
     "productName": "Google BigQuery",
     "productUrl": "https://cloud.google.com/bigquery/",
-    "version": "1.0.0-beta15",
+    "version": "1.0.0-beta16",
     "type": "rest",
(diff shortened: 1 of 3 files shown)


apis/apis.json
```diff
@@ -200,16 +200,16 @@
...
     "type": "other",
     "targetFrameworks": "netstandard1.5;net45",
     "testTargetFrameworks": "netcoreapp1.0;net452",
     "description": "Log4Net client library for the Google Stackdriver Logging API.",
     "tags": [ "Log4Net", "Logging", "Stackdriver" ],
     "dependencies": {
       "log4net": "2.0.8",
-      "Google.Cloud.Logging.V2": "",
-      "Google.Cloud.DevTools.Common": "",
+      "Google.Cloud.Logging.V2": "2.0.0",
```
chrisdunelm: Is this explicit just as a precaution, or is this now required because the versions of log4net/logging are now different?

chrisdunelm:
LGTM, apart from the appveyor failure, that is!

apis/apis.json
jskeet: It's not checked yet, but it *will* be - basically, any project references (rather than package references) should only exist when you're going to release those packages as well.

jskeet:
Doh - the appveyor failure is because I forgot to run generateprojects.sh!

Will do that now, and merge then release on green.

PR #1299, opened 2017-08-02
First draft of documentation for releasing, versioning etc

jskeet:
(Use "view" to see the rendered markdown)


PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+This repository includes multiple projects at multiple versions,
+some of which depend on each other. That makes things tricky, and
+leads to rather more complicated procedures than a simple "single
+package" repository. Please read this guide carefully to understand
+how releases happen, and why they happen that way.
+
+## The API catalog
+
+Only the source within the `apis` directory is published, and not
+all of that. (There are test projects and tools alongside the
```
chrisdunelm: "not all of that" isn't too clear. Maybe something like "...is published, although not every project under that directory."

PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+- The first `PropertyGroup` element (for general properties)
+- The first `ItemGroup` element (for dependencies)
+- The `ItemGroup` element for Source Link (with a label of "dotnet pack instructions")
+- The common import to only attempt to build desktop assemblies on Windows
+
+Any other elements are left as they are - so if you wish to add an
+`ItemGroup` such as for file grouping, just add it anywhere after
+the generated one, and it should do the right thing.
+
+Additionally, the project generator adds all projects under an API
```
chrisdunelm: "...*the* `api` directory..." ?

PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+- Updating the documentation in GitHub (in the `gh-pages` branch)
+
+### Current process
+
+- After the version number is committed, run `tagreleases.sh` from the
+  root directory to tag all updated versions.
+- Run `buildrelease.sh` manually, specifying one of the API names
+  and versions just tagged.
+- Go into the `releasebuild/nuget` directory and push
+- Make sure you have a clone of the gh-pages branch, copy the
```
chrisdunelm: Copy from where?

PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+- Pushing the package to nuget.org
+- Updating the documentation in GitHub (in the `gh-pages` branch)
+
+### Current process
+
+- After the version number is committed, run `tagreleases.sh` from the
+  root directory to tag all updated versions.
+- Run `buildrelease.sh` manually, specifying one of the API names
+  and versions just tagged.
+- Go into the `releasebuild/nuget` directory and push
```
iantalarico: Run runintegrationtests.sh?

PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+- Make sure you have a clone of the gh-pages branch, copy the
+  documentation, commit and push.
+  
+Issues with this:
+
+- Integration tests are not run automatically
```
iantalarico: running the integration tests is part of the current release though right? Just not automated?

PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+
+- The mostly recently released package version
+- The version that's *about* to be released (see below)
+- A prerelease with a suffix of 00 to indicate that the *next* release
+  should be a first prerelease, typically of a new minor version,
+  due to changes that have been merged. This is to avoid accidentally
+  including new features in a patch release. For example, having
+  released Google.Cloud.Storage.V1 version 2.0.0, new features were
+  merged into the `master` branch and the version was immediately
+  changed to 2.1.0-alpha00
```
benwulfe: this seems a bit confusing that the version number doesn't indicate which of the three cases it is.
It could easily lead to misinterpretation.  eg -- the difference between the 2nd and 3rd cases is subtle.  Wouldn't you want some protection in scenarios where you arent doing a prerelease as well?  ie, couldnt someone accidentally include new features in an old release in the second case?

How about a rule that major+minor must always indicate the *next* release (matching whats in source code).  patch # is automatically incremented by tooling based on git tags.  So when we run a tool to create a release tag, it will automatically choose the right patch number for us.

PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+
+Typically version numbers should be changed in a commit which does
+nothing else, for clarity. Include both the `apis.json` change and
+the project file changes that occur when the project generator has
+been run, in the same commit.
+
+Dependencies in the API catalog are specified as properties where
+the property name is the package name and the value is the version
+number. If the version number is left blank, then:
```
benwulfe: wdyt about changing this to "latest", "current" or "indevelopment" instead of a blank string?

benwulfe: or simply "project"

PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+- Go into the `releasebuild/nuget` directory and push
+- Make sure you have a clone of the gh-pages branch, copy the
+  documentation, commit and push.
+  
+Issues with this:
+
+- Integration tests are not run automatically
+- The build takes a long time as it rebuilds all packages, runs unit
+  tests for them all and builds all the documentation
+- There's no validation of project dependencies as described above
```
benwulfe: more specifically, there is no validation that you are "releasing all dependent packages as required"  and there is no verification step that ensures "there are no breaking changes between dependent packages".

The latter is important (as we discussed) because we are making the claim in our nuget packages that eg Common.Diagnostics depends on Trace >= 1.0.
This can break in a variety of ways:
  * When we release a new version of Trace, we should be running common.diagnostic tests on the new version of trace.
  * When we release a new version of Common.Diagnostics, shouldn't we be re-testing the claims we made (=>1.0 trace) and run against at least the min and latest versions of Trace?  Hopefully the project wouldnt have to change -- just the version of dependencies we install via nuget -- so it'd be an aspect of the test runner.

PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+
+### Future process
+
+- `tagreleases.sh` will perform validation of project references for
+  packages being tagged
+- A Google build machine (with appropriate secrets) will listen for
+  tags being created
+- When a set of tags is created (with some delay to avoid "spotting"
+  one tag when more are on the way), a build will be run of just the
```
benwulfe: the last step before publishing to nuget can be a second tag check to ensure it hadnt changed whilst running all the release steps.

PROCESSES.md
```diff
@@ -0,0 +1,147 @@
...
+some of which depend on each other. That makes things tricky, and
+leads to rather more complicated procedures than a simple "single
+package" repository. Please read this guide carefully to understand
+how releases happen, and why they happen that way.
+
+## The API catalog
+
+Only the source within the `apis` directory is published, and not
+all of that. (There are test projects and tools alongside the
+production code.) Nothing in the `tools` is published as a package.
```
evildour: "...in the \`tools\` **directory**..." perhaps?

PROCESSES.md
jskeet: Done.

PROCESSES.md
jskeet: Done.

PROCESSES.md
jskeet: Clarified in a different way.

PROCESSES.md
jskeet: I would really prefer to keep the "last released version" approach. The main advantage is that it prevents accidental release of a new package - you can build all the packages and try to push them all to nuget, and they'll fail because they don't already exist.

I don't want us to automatically increment version numbers at the moment, as there are too many options to consider between "next alpha/beta", "promote from alpha/beta to beta/GA", "create patch from GA to GA", "move from GA to next minor version pre-release" or (for entirely-auto-generated packages) "move from GA to next minor GA".

The difference between the second and third cases is handled by `tagreleases.sh` - it knows never to tag anything with a `alpha/beta00` suffix. Clarified that in the doc.

PROCESSES.md
jskeet: I'd prefer to keep it as it is, to be honest - I think it's reasonably clear, and handles both in-project references and "default dependencies" such as gRPC and GAX. (It's also how project.json used to work, IIRC.)

Happy to hear compelling reasons to change, of course,

PROCESSES.md
jskeet: Done.

PROCESSES.md
jskeet: Done.

PROCESSES.md
jskeet: Yes. Somewhat dependent on developer diligence though :)

PROCESSES.md
jskeet: I don't know how far we want to go in this testing - it blows up pretty quickly. Suppose we release a new version of Diagnostics which depends on Trace 1.0, which depends on GAX 2.0 which depends on Grpc.Core 1.4.  Now suppose there have since been versions Trace 1.1, GAX 2.1, Grpc.Core 1.5. How many permutations do we test?

- Trace 1.0, GAX 2.0, Grpc.Core 1.4
- Trace 1.0, GAX 2.0, Grpc.Core 1.5
- Trace 1.0, GAX 2.1, Grpc.Core 1.4
- Trace 1.0, GAX 2.1, Grpc.Core 1.5
- Trace 1.1, GAX 2.0, Grpc.Core 1.4
- Trace 1.1, GAX 2.0, Grpc.Core 1.5
- Trace 1.1, GAX 2.1, Grpc.Core 1.4
- Trace 1.1, GAX 2.1, Grpc.Core 1.5

It rapidly spirals out of control, IMO.

Perhaps we should run integration tests in two modes:

- Leave everything at minimally supported
- Upgrade everything to latest release

I think that would give a good balance of confidence vs reasonable speed. We *would* need to work out which "upstream" packages to run the integration tests on each time as well of course, which is interesting. Easier to do if we only consider packages within our own repo.

(What about for pre-releases? Do we upgrade everything to the latest pre-release? That may well not be valid...)

I suspect we need to talk about this separately...

PROCESSES.md
jskeet: The tag commit itself? Do you mean just in case someone had (in parallel) spotted a problem and retagged? Could do... I'm not sure how much benefit there is, but it wouldn't hurt. Added in as a possibility.

PROCESSES.md
```diff
@@ -143,5 +148,13 @@ Issues with this:
   tagged packages
 - The integration tests for the tagged packages will be run (as well
   as the regular unit tests)
+  - We may wish to run integration tests of any packages that depend
+    on the package being built, too.
+- Potentially check tht the GitHub tags are still what they were
```
evildour: "tht" => "that"

jskeet: Humbug :) Fixed, thanks.

jskeet:
@benwulfe: Are you happy with this, given the changes and more explanation around versioning?

PROCESSES.md
benwulfe: I simply mean that using an empty string in apis.json has a hidden meaning.  I had no idea what an empty string meant and its unlikely to be clear to anyone else unless they examine the tooling to see what it does, like we did.

PROCESSES.md
benwulfe: pre-release seems less important.  Testing base version and latest sounds like a reasonable compromise and we'll get some (not all) of the intermediate coverage by default as we release each version.

PROCESSES.md
benwulfe: yes on the commit used for release... I meant if our delay isn't long enough.

benwulfe:
LGTM.  I suspect that as we evolve the process, the aspirational parts might grow or lessen.

benwulfe:
oh but I still do not like the empty strings that mean project-project references.  that's super confusing unless you look at the tool source

jskeet:
The empty string doesn't always mean "project-to-project" - it just means "default, latest". Let's merge this doc as it's accurate at the moment, but I'll file a new issue to reconsider that as a separate piece of work. If nothing else, more accurate comments at the top of apis.json wouldn't hurt.

PR #1301, opened 2017-08-04
Change dependency handling in project generator

jskeet:


apis/apis.json
```diff
@@ -54,9 +59,13 @@
     "type": "grpc",
     "description": "Recommended Google client library to access the Google Cloud Datastore API, which accesses the schemaless NoSQL database to provide fully managed, robust, scalable storage for your application.",
     "tags": [ "Datastore" ],
+    "dependencies": {
+      "Grpc.Core": "1.4.0",
+      "Google.Api.Gax.Grpc": "2.0.0"
```
chrisdunelm: I don't really like this requirement to hardcode versions here. It seems to somewhat defeat the point of generating project files. As more packages go GA it's going to get more difficult to update all these correctly.

apis/apis.json
```diff
@@ -54,9 +59,13 @@
...
+    },
     "testDependencies": {
-      "Google.Cloud.ClientTesting": "",
-      "Google.Api.Gax.Grpc.Testing": ""
+      "Google.Cloud.ClientTesting": "project",
+      "Google.Api.Gax.Grpc.Testing": "2.0.0"
```
chrisdunelm: Why is this .Testing version hardcoded?

apis/apis.json
jskeet: But the point is that for stable releases we *shouldn't* be updating them.

Suppose we release GAX 2.1.0 and that becomes the default, but we then want to release Google.Cloud.Logging.V2 v2.0.1 - that *must* still use GAX 2.0.0, because Google.Cloud.Logging.V2 v2.0.0 did. When the version is bumped to 2.1.0-beta01, we'd go back to "default" for GAX and pick up GAX 2.1.0.

apis/apis.json
jskeet: To make sure we test against the version of GAX that the package depends on.

Suppose we have a bug in GAX 2.0.0 that is inadvertently fixed in 2.1.0. If we depend on Google.Api.Gax.Grpc.Testing 2.1.0, that will force GAX 2.1.0 as a dependency, so we'll never see the bug, despite it still being present in the released combination.

PR #1307, opened 2017-08-07
Improve coverage filtering

jskeet:
(Best to review commit-by-commit; the first commit is smaller than it looks!)

### DIFF
--- a/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
+++ b/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
@@ -339,6 +339,9 @@ namespace Google.Cloud.Tools.ProjectGenerator
             var targetExecutable = new XElement("TargetExecutable", "/Program Files/dotnet/dotnet.exe");
             var targetArguments = new XElement("TargetArguments",
                 $"test --no-build -f {GetPreferredCoverageFramework(api)} -c Release");
+            // TODO: Some APIs integration tests naturally contribute to others, but it's good to always have
+            // at least one filter, to avoid displaying coverage to test code etc.
+            var filters = new XElement("Filters", new XElement("IncludeFilters", new XElement("FilterEntry", new XElement("ModuleMask", api.Id))));
             var attributeFilters = new XElement("AttributeFilters",
                 new XElement("AttributeFilterEntry",
                     "System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverageAttribute"),
@@ -350,6 +353,7 @@ namespace Google.Cloud.Tools.ProjectGenerator
             var doc = new XElement("CoverageParams",
                 targetExecutable,
                 targetArguments,
+                filters,
                 attributeFilters,
                 workingDir,
                 output
--- a/tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
+++ b/tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
@@ -1,0 +1,121 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Xml.Linq;
+
+namespace TrimCoverageReport
+{
+    /// <summary>
+    /// Application to trim the detailed XML coverage report produced by dotCover to remove generated code.
+    /// </summary>
+    class Program
+    {
+        // Attributes used multiple times in the XML
+        const string TotalStatements = "TotalStatements";
+        const string CoveredStatements = "CoveredStatements";
+        const string CoveragePercent = "CoveragePercent";
+
+        static int Main(string[] args)
+        {
+            if (args.Length != 2)
+            {
+                Console.WriteLine("Arguments: <input file> <output file>");
+                return 1;
+            }
+            XDocument doc = XDocument.Load(args[0]);
+            ProcessDocument(doc);
+            using (var writer = File.CreateText(args[1]))
+            {
+                doc.Save(writer);
+            }
+            return 0;
+        }
+
+        static void ProcessDocument(XDocument document)
+        {
+            // TODO: Check how local methods are handled (issue 1320)
+
+            // Steps:
+            // - Work out all files that are generated
+            // - Remove File elements for generated files
+            // - Remove each Statement element with a FileIndex for a generated file
+            // - Recompute the coverage for each Method element
+            // - Remove Method elements with 0 statements
+            // - Recompute the coverage for each Type element (based on descendant Method elements)
+            // - Remove Type elements with 0 methods
+            // - Recompute the coverage for each Namespace element (based on descendant Type elements)
+            // - Remove Namespace elements with 0 types
+            // - Recompute the coverage for each Assembly element (based on descendant Namespace elements)
+            // - Remove Assembly elements with 0 namespaces
+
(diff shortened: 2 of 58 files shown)


jskeet:
Looking at the generated coverage report, I'm not sure why the per-file counts aren't present, despite the percentages being present and the folder totals being present. Maybe they're only indexed for merged PRs? Other than that, the new report looks okay...

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                Console.WriteLine("Arguments: <input file> <output file>");
+                return 1;
+            }
+            XDocument doc = XDocument.Load(args[0]);
+            ProcessDocument(doc);
+            // Don't use File.CreateText as that omits the byte order mark.
+            // While byte order marks are nasty, Visual Studio will add it back any time a project file is
+            // manually edited, so it's best if we follow suit.
+            using (var writer = File.CreateText(args[1]))
```
evildour: I'm confused. The comment above says not to use `File.CreateText`. Should this be something else?

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+            // TODO: Check how local methods are handled
+
+            // Steps:
+            // - Work out all files that are generated
+            // - Remove File elements for generated files
+            // - Remove each Statement element with a FileIndex for a generated file
+            // - Recompute the coverage for each Method element
+            // - Remove Method elements with 0 statements
+            // - Recompute the coverage for each Type element (based on descendant Method elements)
+            // - Remove Type elements with 0 statements
```
evildour: Nit: should "statements" be "methods" here? Similarly, seems like the next two should be "types" and "namespaces", respectively.

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+            // - Recompute the coverage for each Namespace element (based on descendant Type elements)
+            // - Remove Namespace elements with 0 statements
+            // - Recompute the coverage for each Assembly element (based on descendant Namespace elements)
+            // - Remove Assembly elements with 0 statements
+
+            var fileElements = document.Root.Element("FileIndices").Elements("File");
+            var generatedFiles = fileElements.Where(file => IsGeneratedFile(file.Attribute("Name").Value)).ToList();
+            var generatedIndices = new HashSet<int>(generatedFiles.Select(file => (int) file.Attribute("Index")));
+            generatedFiles.Remove();
```
evildour: Curious: how does this work? Doesn't `Remove` need to take semething to remove? Is there an extension method somewhere?

tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
```diff
@@ -341,6 +341,7 @@ private static void GenerateCoverageFile(ApiMetadata api, string directory)
             var targetExecutable = new XElement("TargetExecutable", "/Program Files/dotnet/dotnet.exe");
             var targetArguments = new XElement("TargetArguments",
                 $"test --no-build -f {GetPreferredCoverageFramework(api)} -c Release");
+            var filters = new XElement("Filters", new XElement("IncludeFilters", new XElement("FilterEntry", new XElement("ModuleMask", api.Id))));
```
benwulfe: so this will exclude coverage that you might have from a test in a different folder?  I think that will be true indefinitely for spanner integration tests against V1 (its also true for unit tests, but that part is fixable).  Can we only have this filter for unit tests?

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
benwulfe: also I remember reading during codecov's initial support for dotCover files that the BOM was breaking it -- and codecov did just fix support for our file.  So might want to test this if it changes bits from the original xml file.

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+            using (var writer = File.CreateText(args[1]))
+            {
+                doc.Save(writer);
+            }
+            return 0;
+        }
+
+        static void ProcessDocument(XDocument document)
+        {
+            // TODO: Check how local methods are handled
```
benwulfe: assign todo or bug it...

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
benwulfe: +1 especially after a ToList() -- this is magic sauce indeed :)

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
jskeet: Whoops - the code is right, the comment is wrong. (Code was originally copied from project generator, then rewritten...)

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
jskeet: Yup, it's an extension method.

tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
jskeet: Hmm... that sounds like definitely the right call for Spanner, but not necessarily right for other things. Never mind - will do it for now and we can see.

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
benwulfe: I suspect that without re-indexing the file IDs so they are all consecutive, something internal in codecov doesnt like it.
But I'm not sure its really worth fixing to get file coverage info.

tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
jskeet: Ah, just thought - removing the filter would bring back coverage of all the tests etc.
I'll file a bug for Spanner on this, and add it into apis.json for "extra assemblies to cover". (Or consider including all dependencies on APIs.)

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
jskeet: Filed #1320.

jskeet:
Right, all done. I'll investigate the new issues tomorrow.

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+            ((int)container.Attribute(TotalStatements), (int)container.Attribute(CoveredStatements));
+
+        /// <summary>
+        /// Detects whether a file is generated. All three code generators we use (protoc, grpc,
+        /// toolkit) include a comment starting with "// Generated" early in the file.
+        /// </summary>
+        /// <param name="path"></param>
+        /// <returns></returns>
+        static bool IsGeneratedFile(string path) =>
+            File.ReadLines(path).Take(20).Any(line => line.StartsWith("// Generated"));
```
benwulfe: this seems a bit fragile, but I dont really have any alternative to suggest.
I suppose we could make a unit test the asserts this assumption (fails if a known generated file doesnt have it).

jskeet: I think it'll be fine - it's true for all the generators we have now, and we'll soon spot if we introduce a new generator, I think.

benwulfe:
overall lgtm.  only risk I see altered is the gaps in the file indices caused by the removals. -- but hopefully that doesn't cause an issue.  Alternatively, you could leave the files there but with 0 total statements so they wouldnt affect the totals.

jskeet:
Good call on trying without removing the files. Will give that a go.

jskeet:
Interestingly, the "half way through" report from current CI has numbers:
https://codecov.io/gh/GoogleCloudPlatform/google-cloud-dotnet/tree/6b9cf235dcef13226ffeb1a917cab4d88d1c97d4/apis

jskeet:
Ah, got it - it's coverage *changes* that are null, which is unsurprising. Merging.

PR #1311, opened 2017-08-07
Reinstate coverage option for integration tests

jskeet:
Note that this and #1307 both change the coverage XML files; whichever is merged first, the other will need to be rebased and generateprojects will need to be rerun. (I'll take care of doing that.)


benwulfe:
for integration tests I was actually thinking about having reportgenerator publish its html report via a new task on jenkins.  so each jenkins test run would dump its dvcr in a shared folder on the machine and then a new task would run "createcoveragereport.sh" at the very end.

This way you could browse to jenkins if you ever wanted to see your full combined coverage.

but no objections to being able to have coverage info via runintegrationtests.

jskeet:
It's definitely useful to be able to run them locally. Is it feasible to publish the XML to a sort of "second bucket" in codecov so we could have "current unit tests" and "integration tests last time they were run" as separate badges? My guess is that this isn't something that's covered in the design of codecov, but you know more than I do about it.

benwulfe:
yeah I haven't seen anything like that in codecov, but that doesn't mean its not there.
this PR lgtm.

jskeet:
Thanks - will rebase and regenerate the projects in the other PR.

PR #1315, opened 2017-08-07
Minor tweaks to AppVeyor build

jskeet:
- Build the docs after running the tests (as tests are more likely
  to fail)
- Specify scripts in appveyor.yaml in execution order
- Make sure we don't use automatic AppVeyor test reporting (which
  can be flaky) when running tests under coverage


jskeet:
Adding Ian as an assignee as he appears to be in slightly earlier :)

PR #1316, opened 2017-08-07
Make Spanner snippets use same instance as integration tests

jskeet:
(Arguably "spanner-integration-tests" would be even clearer than
"spannerintegration" but the cost of getting all current users to
create a new instance outweighs the benefit, I think.)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SnippetFixture.cs
```diff
@@ -30,7 +30,7 @@ public class SnippetFixture : IDisposable, ICollectionFixture<SnippetFixture>
     {
         private readonly Lazy<Task> _creationTask;
 
-        public string TestInstanceName => "spannerinstance";
+        public string TestInstanceName => "spannerintegration;
```
benwulfe: looks like you forgot a ending quote.

otherwise lgtm.

jskeet:
Humbug - thanks!

jskeet:
(Will merge on green.)

PR #1319, opened 2017-08-07
Add table update/patch functionality

jskeet:
This is a copy/paste of dataset functionality, with the only
difference being that specifying a table reference requires three
pieces of information instead of two.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryTable.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryTable.cs
@@ -197,6 +197,45 @@ namespace Google.Cloud.BigQuery.V2
             _client.CreateCopyJob(Reference, destination, options);
 
         /// <summary>
+        /// Updates this table to match the specified resource.
+        /// </summary>
+        /// <remarks>
+        /// This method delegates to <see cref="BigQueryClient.UpdateTable(TableReference, Table, UpdateTableOptions)"/>.
+        /// A simple way of updating the table is to modify <see cref="Resource"/> and then call this method with no arguments.
+        /// This is convenient, but it's important to understand that modifying <see cref="Resource"/> in this way leaves this object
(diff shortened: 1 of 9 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryTable.cs
```diff
@@ -197,14 +197,53 @@ internal BigQueryTable(BigQueryClient client, Table resource)
...
+        {
+            if (matchEtag)
+            {
+                resource.ETag = Resource.ETag;
+            }
+            return _client.PatchTable(Reference, resource, options);
+        }
+        
+        /// <summary>
+                 /// Asynchronously uploads a stream of CSV data to this table.
```
chrisdunelm: Indent is wrong from this line to the end of the xmldoc.

jskeet: Gosh - I wonder how that happened. Thanks!

NPSF3000:
Hi @jskeet, it's nice to see someone of your reputation when chasing down bugs ;)  

I just read through the code and saw the "client.PatchTable(datasetId, tableId, new Table { FriendlyName = "Patched table" });" snippet, but unfortunately this is not yet in nuget.

What's the process between testing and release?  Should I just grab the repo myself and build?

jskeet:
@NPSF3000: We do a release periodically, typically after we've added a whole set of changes - often if we're not expecting more changes imminently. I'm in the middle of doing some more work, so there may not be another release for a week or so. If you really need this *right now*, you can pull the source, but typically I'd recommend waiting until there's a release.

You can always get the `BigqueryService` from the client and call `Patch` yourself, of course - this library just makes things a bit simpler.

NPSF3000:
@jskeet thanks for the info.  Doing some once off maintenance now so I'll figure it out.  Glad to see .Net getting some love!

PR #1322, opened 2017-08-07
Ensure the RpcRetryConsumer retry logic does not run if the retry type is "NoRetry".

### LINKED ISSUE #1305: Google.Cloud.Diagnostics.AspNetCore 1.0.0-beta 13 as very high CPU usage
@iantalarico as discussed on slack

```
 dotnet --info
.NET Command Line Tools (2.0.0-preview2-006497)

Product Information:
 Version:            2.0.0-preview2-006497
 Commit SHA-1 hash:  06a2093335

Runtime Environment:
 OS Name:     Mac OS X
 OS Version:  10.12
 OS Platform: Darwin
 RID:         osx.10.12-x64
 Base Path:   /usr/local/share/dotnet/sdk/2.0.0-preview2-006497/

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0-preview2-25407-01
  Build    : 40c565230930ead58a50719c0ec799df77bddee9
```

Just starting up the app with one of the following enabled in startup.cs (no need to call a controller or whatever):
```
           // var tracingQPS = 1000;
           // services.AddGoogleTrace(options =>
            // {
            //     options.ProjectId = Configuration["AppSettings:GoogleProjectID"];
            //     options.TraceFallbackPredicate = TraceDecisionPredicate.Default;
            //     options.Options = TraceOptions.Create(tracingQPS);
            // });

            {
                options.ProjectId = Configuration["AppSettings:GoogleProjectID"];
                options.ServiceName = "API";
                options.Version = "1.0";
            });
```
The `Use*()` is commented out but even if it is enabled, issue appears.

`Google.Cloud.Diagnostics.AspNetCore 1.0.0-beta 13 ` -> dotnet goes above 100% CPU.
`Google.Cloud.Diagnostics.AspNetCore 1.0.0-beta 11 ` -> dotnet has normal CPU usage.

Also happens with .net core 1.1.

Note: With such high CPU usage, App engine automatically scaled up the number of instances to the default max (20). Projected billing was around 700$ for three days for all our services...

iantalarico:
Also ensure that the RetryInterval is greater than 0 to avoid a loop without a pause.

Fixes #1305.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/RpcRetryConsumerTest.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/RpcRetryConsumerTest.cs
@@ -208,7 +208,7 @@ namespace Google.Cloud.Diagnostics.Common.Tests
         /// </summary>
         private class FakeSequentialThreadingTimer : ISequentialThreadingTimer
         {
-            internal Action Callback = null;
+            internal Action Callback { get; private set; }
 
             public void Initialize(Action callback, TimeSpan waitTime) => Callback = callback;
 
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/RpcRetryConsumerTest.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/RpcRetryConsumerTest.cs
@@ -64,8 +64,7 @@ namespace Google.Cloud.Diagnostics.Common.Tests
                 Assert.Throws<RpcException>(() => { retryConsumer.Receive(intArray); });
 
                 mockConsumer.Verify(c => c.Receive(intArray), Times.Once);
-                timer.Call();
-                mockConsumer.Verify(c => c.Receive(intArray), Times.Once);
+                Assert.Null(timer.Callback);
             }, RetryOptions.NoRetry(ExceptionHandling.Propagate));
         }
 
@@ -78,8 +77,7 @@ namespace Google.Cloud.Diagnostics.Common.Tests
                 retryConsumer.Receive(intArray);
 
                 mockConsumer.Verify(c => c.Receive(intArray), Times.Once);
-                timer.Call();
-                mockConsumer.Verify(c => c.Receive(intArray), Times.Once);
+                Assert.Null(timer.Callback);
             }, RetryOptions.NoRetry());
         }
 
@@ -210,11 +208,11 @@ namespace Google.Cloud.Diagnostics.Common.Tests
         /// </summary>
         private class FakeSequentialThreadingTimer : ISequentialThreadingTimer
         {
-            private Action _callback;
+            internal Action Callback = null;
 
-            public void Initialize(Action callback, TimeSpan waitTime) => _callback = callback;
+            public void Initialize(Action callback, TimeSpan waitTime) => Callback = callback;
 
-            public void Call() => _callback();
+            public void Call() => Callback();
 
             public void Dispose() { }
         }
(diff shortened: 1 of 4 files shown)


jskeet:
I can review this first thing tomorrow, but feel free to get Ben or Mike to review if you need it before then.

iantalarico:
@jskeet It's a pretty small change.  I'll see if Ben can do it so I can get a fix out today.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
```diff
@@ -72,7 +72,10 @@ internal class RpcRetryConsumer<T> : IConsumer<T>
             _timerFactory = GaxPreconditions.CheckNotNull(timerFactory, nameof(timerFactory));
 
             // Start the background timer that will retry failed requests.
-            _timer = _timerFactory(FlushBuffer, _options);
+            if (options.RetryType != RetryType.None)
+            {
```
benwulfe: is there value in asserting the interval here where its closer to what would cause the issue?  Or are you forced to always go thru the retryoption validation?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/RpcRetryConsumerTest.cs
```diff
@@ -210,11 +208,11 @@ private void SetupTestReceiveThrows(Action<FakeSequentialThreadingTimer, Mock<IC
         /// </summary>
         private class FakeSequentialThreadingTimer : ISequentialThreadingTimer
         {
-            private Action _callback;
+            internal Action Callback = null;
```
benwulfe: make this a readonly internal property instead of a field

iantalarico: This can't be readonly due to the api of ISequentialThreadingTimer (which I need to revisit)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/RpcRetryConsumer.cs
iantalarico: A user will always have to go through this channel.  I'd rather fail here than later if possible.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/RpcRetryConsumerTest.cs
benwulfe: you can still make it a property, right?  having a public field like that is usually not done.  I know its just a fake, but this should work just as well:

internal Action Callback { get; set; }

also, initializing to null is probably redundant.

benwulfe:
LGTM overall except I would switch the internal field to an internal property.
Also, it may be worth understanding the functional difference between net45 and netcore that surfaced this and if it that difference might occur elsewhere.

iantalarico:
Thanks Ben, I fixed up the internal property.  As for what caused this I'm pretty curious.  All in all this was due to a bug in my code but I'd like to see why this happens on linux but not windows.

benwulfe:
yeah my only guess there was that Task.Delay(0) is acting differently -- like its causing a thread context switch in windows.  If we know Delay acts differently, we might want to review it across the board in our libs.

jskeet:
The changes look fine, but I haven't validated the original cause yet. I'll try to follow your logic on that today, and repro the difference between Linux and Windows.

iantalarico:
Thanks Ben and Jon, will do a release shortly.

PR #1330, opened 2017-08-08
Sends timeout as an override on a per call basis using values from Sp…

benwulfe:
…annerCommand.CommandTimeout instead of timing out the task (which may let the task continue until completion).
This streamlines the timeouts and removes some of the duplication that existed where timeouts at the grpc level were controlled via spanneroptions and the entire task had a timeout controlled by spannercommand.commandtimeout.


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
```diff
@@ -372,10 +372,35 @@ private class CredentialWrapper : ITokenAccess
...
+                async () =>
+                {
+                    using (await _testFixture.GetTestDatabaseConnectionAsync())
+                    {
+                        using (var connection =
+                            new SpannerConnection(_testFixture.ConnectionString))
+                        {
+                            var cmd =
+                                connection.CreateSelectCommand("SELECT 1");
+                            cmd.CommandTimeout = 0;
```
jskeet: The docs for DbCommand.CommandTimeout include:

> Note to implementers, it is recommended that 0 means no timeout.

Do we definitely want to violate that? (I understand it makes it harder to test otherwise, but...)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -58,6 +58,13 @@ public sealed class SpannerTransaction : DbTransaction, ISpannerTransaction
 
         private Session Session { get; }
 
+        // Note: We use seconds here to follow the convention set by DbCommand.CommandTimeout.
```
jskeet: We could make that explicit with a suffix? Inconsistent with DbCommand in terms of naming, but at least it's clearer. On the other hand, if we're going to be inconsistent anyway, we could make it a TimeSpan. Hmm. Not sure.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/CallSettingsMutations.cs
```diff
@@ -0,0 +1,80 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using Google.Api.Gax;
+using Google.Api.Gax.Grpc;
+
+namespace Google.Cloud.Spanner.V1
+{
+    /// <summary>
```
jskeet: Possibly add a note that suggests some of these might be better in GAX?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/CallSettingsMutations.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// </summary>
+    internal static class RetrySettingMutations
+    {
+        /// <summary>
+        /// Sets the total expiration of a new <see cref="RetrySettings"/>, copying all other members from this instance.
+        /// </summary>
+        /// <param name="retrySettings"></param>
+        /// <param name="expiration"></param>
+        /// <returns></returns>
+        public static RetrySettings SetTotalExpiration(this RetrySettings retrySettings, Expiration expiration) =>
```
jskeet: Any reason this isn't WithTotalExpiration instead of SetTotalExpiration?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
benwulfe: it definitely suggests that in the msdn docs doesnt it.
also, it would make sense to change spanneroptions to also be seconds instead of timespan because you'd otherwise get strange inconsistencies when the value was carried over (either CommandTimeout would lie when it returns its value due to loss of precision or the actual timeout used would lose precision, making TimeSpan uselss).  So I think I'll have to return to my original instinct in SpannerOptions (I think it was this way at some point before).
It will surely make the timeout tests more challenging as 1 second isnt enough time to reliably trigger a timeout.  thinking...

jskeet: Definitely tricky for testing, yes. I'll have a think and see whether I can come up with something that would help...

jskeet: (Possibly an internal flag of "make 0 mean 0 for CommandTimeout"?)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
benwulfe: I'd rather keep it consistent with Command, which is locked in to the name due to it being an override.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
```diff
@@ -43,8 +44,9 @@ public EphemeralTransaction(SpannerConnection connection)
         /// </summary>
         /// <param name="mutations"></param>
```
benwulfe: sure but note this interface and method is not public

benwulfe:
Jon, I think this (AllowZeroTimeouts) is how I want options to work (for spanner and BQ ado), but there is still some work to move the rest of spanner options into this pattern.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
```diff
@@ -372,10 +372,35 @@ private class CredentialWrapper : ITokenAccess
...
+        public async Task CommandTimeout()
+        {
+            long result = 0;
+            var e = await Assert.ThrowsAsync<SpannerException>(
+                async () =>
+                {
+                    using (await _testFixture.GetTestDatabaseConnectionAsync())
+                    {
+                        using (var connection =
+                            new SpannerConnection($"{_testFixture.ConnectionString};AllowZeroTimeouts=true"))
```
jskeet: Ooh, I hadn't thought of making it part of the connection string. That's neat. Not sure about the name though, as a timeout of 0 is "allowed" anyway - it just means "infinite".

How about AllowImmediateTimeouts?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
```diff
@@ -372,10 +372,35 @@ private class CredentialWrapper : ITokenAccess
...
+                        {
+                            var cmd =
+                                connection.CreateSelectCommand("SELECT 1");
```
jskeet: Unwrap the line?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/SpannerStressTests.cs
```diff
@@ -131,6 +132,10 @@ public SpannerStressTests(TestDatabaseFixture testFixture, ITestOutputHelper out
 
         private async Task RunStress(Func<Stopwatch, Task<TimeSpan>> writeFunc)
         {
+            // Clear current session pool to eliminate the chance of a previous
```
jskeet: Is this logically part of this change, or just an issue you noticed while testing this change?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ClientPool.cs
```diff
@@ -45,9 +44,10 @@ internal ClientPool(ISpannerClientFactory clientFactory = null)
 
         public async Task<SpannerClient> AcquireClientAsync(
             ITokenAccess credentials = null,
-            ServiceEndpoint endpoint = null)
+            ServiceEndpoint endpoint = null,
+            IDictionary additionalOptions = null)
```
jskeet: Does this need to be non-generic? Is that following a convention from somewhere? I now understand it's so that we can pass the connection string builder directly, but that means there will be all the existing aspects like the service name in there as well. Can we make it genuinely *additional* options, and restrict it to a `Dictionary<string, string>`?`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ClientPool.cs
```diff
@@ -114,10 +120,10 @@ public override int GetHashCode()
                 unchecked
                 {
                     return ((Credential?.GetHashCode() ?? 0) * 397) ^
-                        (Endpoint?.GetHashCode() ?? 0);
+                        (Endpoint?.GetHashCode() ?? 0) ^
+                        (AdditionalOptions?.Count.GetHashCode() ?? 0);
```
jskeet: Possibly write a better hash code here than just the count? Not sure how often we'll have keys that differ only by options...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -441,7 +442,8 @@ public override Task OpenAsync(CancellationToken cancellationToken)
                     {
                         localClient = await ClientPool.Default.AcquireClientAsync(
                                 _connectionStringBuilder.Credential,
-                                _connectionStringBuilder.EndPoint)
+                                _connectionStringBuilder.EndPoint,
+                                _connectionStringBuilder)
```
jskeet: This *looks* distinctly weird - but I understand it now... see my earlier comment about ideally only passing genuinely "additional" options (rather than all options including data source etc).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerOptions.cs
```diff
@@ -157,11 +157,12 @@ public ResourcesExhaustedBehavior ResourcesExhaustedBehavior
         }
 
         /// <summary>
-        /// Defines the global timeout duration.
+        /// Defines the default values for <see cref="SpannerCommand.CommandTimeout"/> and
```
jskeet: Maybe add the part about this being seconds (and 0 having a variable meaning) here so that users don't need to refer to other docs?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
jskeet: Ack. Maybe add the meaning of 0 here.
(I know it's going to be annoying to copy it all over the place, but it'll definitely make users' lives easier :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TypeUtil.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            if (d1 == null || d2 == null)
+            {
+                return false;
+            }
+
+            return d1.Count == d2.Count
+                && d1.Keys.All(key => d2.ContainsKey(key) && Equals(d2[key], d1[key]));
+        }
+    }
```
jskeet: I'd add a hash code generator here. Here's one you could copy:

https://github.com/google/protobuf/blob/35db2675b418ea105ef89f88fd1bf0257eb8a3ff/csharp/src/Google.Protobuf/Collections/MapField.cs#L364

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/SpannerStressTests.cs
benwulfe: its part of the change -- very indirectly.
What's going on is that because clients are now indexed on their full set of options (which is by design), some of the tests that set AllowZeroTimeouts are creating pools that do not otherwise get used -- causing a pretty significant difference in their sizes.
this does a full reset before the stress test, which allows all other tests to use different options without messing up this test.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ClientPool.cs
benwulfe: in fact, I'm more likely to eliminate it all and just use the connectionstringbuilder itself as the key.  then, it can override equals as the full formed normalized connectionstring (plus the credentials I guess).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TypeUtil.cs
benwulfe: I'm not 100% convinced that iterating through the entire dictionary to product a hashcode is always a great idea.  In that case the running time of hashcode is on the same order as equals.
In my specific case, it would only make sense if I expected a large number of hits on the same spanner resource along with the same number of options, but with different values.
And even then, the total number of those cases would have to be a fairly large number to have any noticeable impact.  In the more typical case of having many different spanner resources, they would all have different hashcodes and the net result would be faster due to a more efficient hashcode.
Doing the hash code generator here feels like over-optimization.

benwulfe:
"anywhere I pass in a timeout of 0 seconds, that's actually affected by the SpannerSettings.AllowZeroTimeout"  yes that should be the case.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TypeUtil.cs
jskeet: Okay, let's leave it as it is for now. It does feel odd having differences we know will be seen in Equals but ignoring them in GetHashCode, but I take your point about making that quick.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ClientPool.cs
benwulfe: I'm going to with it as is and when I revisit to move all options into this pattern, I will alter the key to simply be the connectionstringbuilder itself (which is always required and built).
Its a little wasteful, but technically correct.

jskeet:
As a thought - do we clone the ConnectionStringBuilder before using it in the pool? I'm very wary of the use of a mutable key...

benwulfe:
It surely could not hurt to clone it -- there is a chance on client release
that it would not work properly which would mess up the fairness of the
round robin.  Looks like you'll get your dictionary<string, object> after
all :)

benwulfe:
Actually SpannerConnection keeps its connectionstringbuilder internal and
does not allow it to be modified (which is by design).

There *is* a hole here where when it is initially set from a user's
instance, it needs to be cloned.  But otherwise it should be good.
If you attempt to change the connection string by setting a new datasource,
the spannerconnection must first be closed -- which will release the client
with the old connectionstringbuilder.

benwulfe:
squashed for merge...

PR #1338, opened 2017-08-09
While individual rpc methods within spanner.data are all idempotent, …

benwulfe:
…the entire transaction is not.

Therefore, allowing retries at the transaction level on deadline_exceeded is especially dangerous.

Also, unavailable is already retried at the rpc level -- so allowing another set of exponential backoff
retries at the transaction level is unnecessary (but not as harmful as deadline_exceeded).

The only error that should be retried at the transaction level is aborted -- which indicates a transaction deadlock.
Verified that this behavior is now fully in sync with the java client.


benwulfe:
all individual calls are idempotent and configured that way (we had reviewed that earlier.
However, I hadn't previously properly considered the fact that a transaction retry is not idempotent and therefore should not include those error conditions.
Really at the transaction level the only error that should cause a retry is the deadlock condition which throws ABORTED.
At some point before GA we may want to change the name of the extension method "IsTransientSpannerError" to be more accurate.  Overall I'm pretty sure that the change in this PR is correct and even if its (somehow) wrong, its *safer* than what was there before.

PR #1339, opened 2017-08-09
Create the 'packages' directory if it does not exist when installing docfx

### LINKED ISSUE #1337: When running buildrelease.sh toolsversion.sh cannot find packages the packages directory.

iantalarico:
The packages directory may not exist when building a release from a clean client.

Fixes #1337

### DIFF
--- a/toolversions.sh
+++ b/toolversions.sh
@@ -42,6 +42,7 @@ install_docfx() {
   if [[ ! -f $DOCFX ]]
   then
     (echo "Fetching docfx v${DOCFX_VERSION}";
+     mkdir -p $REPO_ROOT/packages;
      cd $REPO_ROOT/packages;
      mkdir docfx.$DOCFX_VERSION;
      cd docfx.$DOCFX_VERSION;


toolversions.sh
```diff
@@ -42,6 +42,7 @@ install_docfx() {
   if [[ ! -f $DOCFX ]]
   then
     (echo "Fetching docfx v${DOCFX_VERSION}";
+     mkdir -p $REPO_ROOT/packages;
```
benwulfe: why not use nuget like the other dependencies?

https://www.nuget.org/packages/docfx/

iantalarico: I'm not really sure why this, this just pulls in the zip but AFAICT will fix the current issue.

@jskeet any reason to not use nuget?

jskeet: I've never tried using the nuget package instead of the zip file. I'll look into it - I can imagine reasons why it might not work, if it's more intended for cases where you're using the code in docfx. But if we *can* use it, we should :)

jskeet: Have just tried to use nuget install with the package, and it fails to install due to invalid dependencies. Let's stick with the zip file :)

PR #1341, opened 2017-08-10
Instructions for adding a new API

jskeet:


jskeet:
(Will wait for feedback from Ian too - this doc was mostly written as it became obvious that it was needed when he was adding Cloud Debugger.)

NEWAPI.md
```diff
@@ -0,0 +1,202 @@
...
+
+(Note: currently if your API uses long-running operations, it *won't* compile until you've added a partial class like [this](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/58a38a605df48f8ee70d36e41c948bbd1b3628f3/apis/Google.Cloud.Speech.V1/Google.Cloud.Speech.V1/CloudSpeechGrpcPartial.cs).
+We will be generating that code soon.)
+
+Step 10: Commit changes
+-----------------------
+
+Commit the change to `apis.json` and the generated files.
+
+Optional: Step 11: Create documentation
```
iantalarico: I think we should make this a required step, like you mentioned to me earlier it's simple and at least we have a minimum starting point.

NEWAPI.md
```diff
@@ -0,0 +1,202 @@
...
+{{description}}
+
+{{installation}}
+
+{{auth}}
+```
+
+- Ideally, add a "getting started" section indicating the client
+  classes that act as an entry point to the API.
+- Even more ideally, add some examples in the Snippets project and embed them in the docs.
```
iantalarico: I think we should make this it's own step and require at least one (simple) custom snippet just as a simple check.

NEWAPI.md
jskeet: There are some cases where an API is whitelisted and your request to join the whitelist may not have been processed yet, making it impossible to verify that the code definitely works.

How about "strongly encouraged"?

NEWAPI.md
jskeet: Happy to make it another step, sure. Maybe I'll reorder things so that it's "Write a snippet to test things if you possibly can" followed by "Write docs that show your snippet if you wrote one".

iantalarico: +1 to that.  I think that would be a good flow.

NEWAPI.md
iantalarico: I still think this should be required.  I highly doubt we will have someone generate an code for an API they cannot use.  If that's the case I say we make an exception or have someone else test it?

jskeet: Righto. Aside from anything else, having optional parts makes the whole thing more complex.

However, that then suggests that we should possibly autogenerate the basic docs/index.md file, at least if we have a product name/url. Then I can simplify these instructions even further. What do you think? (If we go with that, I'll modify the project generator in a separate PR, then come back to this one.)

iantalarico: I'm all for auto generating the docs.  If it's not much work I say go for it, if it is a fair amount of work I wouldn't worry about it.  I don't see this happening that much and a copy paste of an existing doc would work.

jskeet: It'll be very simple - will send you a PR in 15 minutes, I suspect :)

jskeet:
I've added a second commit with the requested changes. (I'll squash to one commit before merging.)

PTAL.

I've removed the part about LRO needing extra work, as Chris has now fixed that!

PR #1342, opened 2017-08-10
BigQuery code generator

jskeet:
The code generator code itself is reasonably ugly, but that's not *too* much of a concern.
The second commit shows an example with four methods for label modification.
If we like this tool in general, I'll probably go back and apply it to the subset of existing methods that an easily be templated.


jskeet:
(If this looks good, I'll only merge the first commit, and put the rest in another PR that has the implementation as well.)

jskeet:
Overnight I've worked out some ways I can make the code much cleaner. Should do exactly the same thing though... feel free to hold off reviewing until I've pinged this thread again.

jskeet:
Okay, refactoring complete. Now most of the code works in terms of Parameter objects, which allows a lot more uniformity.

jskeet:
The solution file changes are due to https://github.com/dotnet/project-system/issues/1821

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClient.DatasetLabels.cs
```diff
@@ -0,0 +1,416 @@
...
+using System.Threading.Tasks;
+
+namespace Google.Cloud.BigQuery.V2
+{
+    public abstract partial class BigQueryClient
+    {
+        // Methods modifying labels on datasets
+
+#region SetDatasetLabel
+        /// <summary>
```
chrisdunelm: Might be helpful to put a comment in the generated regions, to state that they were generated.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.GenerateOverloads/Program.cs
```diff
@@ -0,0 +1,119 @@
...
+                return 0;
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine($"Error: {e}");
+                return 1;
+            }
+        }
+
+        private static IEnumerable<string> ModifyCode(ApiMethod method, IEnumerable<string> existingCode)
```
chrisdunelm: I'd be happier with code-generation that completely generates and replaces the source file.
(In my opinion) this would be cleaner, and be less likely to produce problems in the future.
But I can see that'll be more complex, so this is fine. Could be changed later if desirable.

jskeet: Yup - that would also prohibit methods being grouped together logically.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClient.DatasetLabels.cs
jskeet: Sure, will do.

jskeet:
(Reduced to one commit, so no actual generated code. Will merge on green, then start working on the labels feature...)

PR #1349, opened 2017-08-10
Changes to allow trim to run on jenkins.

benwulfe:
This adds an option to modify the generated paths to be re-routed to the current repo.
This is important for jenkins where each coverage dvcr is coming from a different path.

It also ignores bad paths which may happen due to timing in when the jenkins job starts.

NOTE: There is a possibility that the paths could be out of sync with each coverage report, but its not really possible to fix it.  *Most* of the time, when  jenkins runs, coverage.xml will be based off the latest source.

### DIFF
--- a/tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
+++ b/tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
@@ -30,13 +30,18 @@ namespace TrimCoverageReport
         const string CoveredStatements = "CoveredStatements";
         const string CoveragePercent = "CoveragePercent";
 
+        private static bool s_fixFilePaths;
+        private static bool s_skipBadPaths;
+
         static int Main(string[] args)
         {
-            if (args.Length != 2)
+            if (args.Length < 2 || args.Length > 4)
             {
                 Console.WriteLine("Arguments: <input file> <output file>");
                 return 1;
             }
+            s_fixFilePaths = args.Any(x => x.Equals("--fixpaths", StringComparison.OrdinalIgnoreCase));
+            s_skipBadPaths = args.Any(x => x.Equals("--skipbadpaths", StringComparison.OrdinalIgnoreCase));
             XDocument doc = XDocument.Load(args[0]);
             ProcessDocument(doc);
             using (var writer = File.CreateText(args[1]))
@@ -63,7 +68,8 @@ namespace TrimCoverageReport
             // - Recompute the coverage for each Assembly element (based on descendant Namespace elements)
             // - Remove Assembly elements with 0 namespaces
 
-            var fileElements = document.Root.Element("FileIndices").Elements("File");
+            var fileElements = document.Root.Element("FileIndices").Elements("File").ToList();
+            FixPaths(fileElements);
             var generatedFiles = fileElements.Where(file => IsGeneratedFile(file.Attribute("Name").Value)).ToList();
             var generatedIndices = new HashSet<int>(generatedFiles.Select(file => (int) file.Attribute("Index")));
             generatedFiles.Remove();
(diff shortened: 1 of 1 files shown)


tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -30,13 +30,18 @@ class Program
         const string CoveredStatements = "CoveredStatements";
         const string CoveragePercent = "CoveragePercent";
 
+        private static bool s_fixFilePaths;
+        private static bool s_skipBadPaths;
+
         static int Main(string[] args)
         {
-            if (args.Length != 2)
+            if (args.Length < 2 || args.Length > 4)
```
iantalarico: Suggestion: use parsing package (I personally like CommandLine https://commandline.codeplex.com/)

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -109,13 +115,61 @@ private static void RecomputeAndSimplify(XDocument document, string parentElemen
         private static (int statements, int covered) GetContainerCoverage(XElement container) =>
             ((int)container.Attribute(TotalStatements), (int)container.Attribute(CoveredStatements));
 
+        private static void FixPaths(IEnumerable<XElement> fileElements)
+        {
+            if (!s_fixFilePaths) return;
+            foreach (var fileNode in fileElements)
+            {
+                fileNode.Attribute("Name").Value = AdjustedLocalPath(fileNode.Attribute("Name")?.Value);
```
iantalarico: Is this null check needed here? won't it fail elsewhere still?

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -109,13 +115,61 @@ private static void RecomputeAndSimplify(XDocument document, string parentElemen
...
+                currentFolder = currentFolder.Parent;
+            }
+            return originalPath;
+        }
+
         /// <summary>
         /// Detects whether a file is generated. All three code generators we use (protoc, grpc,
         /// toolkit) include a comment starting with "// Generated" early in the file.
         /// </summary>
         /// <param name="path"></param>
```
iantalarico: nit: empty params and returns

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -30,13 +30,18 @@ class Program
...
 
+        private static bool s_fixFilePaths;
+        private static bool s_skipBadPaths;
+
         static int Main(string[] args)
         {
-            if (args.Length != 2)
+            if (args.Length < 2 || args.Length > 4)
             {
                 Console.WriteLine("Arguments: <input file> <output file>");
```
jskeet: Add the options in the output here?

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
```diff
@@ -109,13 +115,61 @@ private static void RecomputeAndSimplify(XDocument document, string parentElemen
...
+
         /// <summary>
         /// Detects whether a file is generated. All three code generators we use (protoc, grpc,
         /// toolkit) include a comment starting with "// Generated" early in the file.
         /// </summary>
         /// <param name="path"></param>
         /// <returns></returns>
-        static bool IsGeneratedFile(string path) =>
-            File.ReadLines(path).Take(20).Any(line => line.StartsWith("// Generated"));
+        private static bool IsGeneratedFile(string path)
```
jskeet: I'd use `File.Exists` here instead:

```csharp
return (s_skipBadPaths && !File.Exists(path)) || (code as before)
```

I'd also rename the method. (I very nearly had this code in the original version...)

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
benwulfe: no all nodes here should have a name attribute as they are all file identifiers.

benwulfe:
+taralrico
Ian, did you git trouble using a single repo?  It looks like it works for some users:

https://stackoverflow.com/questions/9914664/how-can-i-get-multiple-jenkins-builds-to-work-from-one-local-git-repo

last guy to comment seemed to have an issue with it updating however.

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
benwulfe: I want to know when this error condition exists even when you pass this flag because this can also happen if there is a bug in the path fixing logic.

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
benwulfe: it could likely be removed here.  I was just making resharper happy.

tools/Google.Cloud.Tools.TrimCoverageReport/Program.cs
benwulfe: good suggestion.  if it gets more complicated, will do.

iantalarico:
I've never tried to use a central repo.  I like to have everything as separate as possible when doing tests/builds.  Each workspace has a clean repo and build and then is wiped out after.  I'm open to changing this if we see benefit but lets move this to email if we want to continue.

jskeet:
Do we still want this? It seems potentially useful, but it's extra complexity if we don't actually need it...
(Perhaps we want some aspects but not all?)

jskeet:
Ping for delete/prune?

benwulfe:
lets revisit this if/when we decide to keep multiple jenkins runs

PR #1352, opened 2017-08-11
Do not swallow exceptions thrown by an application if the diagnostics library fails. 

### LINKED ISSUE #1350: ErrorReportingExceptionLoggerMiddleware can swallow Exceptions
For a few minutes today, our ASP.NET Core Web API was returning unexpected 500 exceptions.

While we normally see these errors via StackDriver's ErrorReporting module, today we were told by our colleagues, and after we were informed we did not see any errors get added to the StackDriver ErrorReporting module. When we investigated the logs of the web servers directly, we could not glean any additional information because the ErrorReporting middleware appeared to be throwing its own exceptions on how it was failing to report failures to StackDriver, which appears to swallow the original exception it was trying to report in the first place.

Here's the stack trace I found in the logs:

```
      Request finished in 35825.8552ms 500 
  [40m[32minfo[39m[22m[49m: Microsoft.AspNetCore.Hosting.Internal.WebHost[2]
     at Microsoft.AspNetCore.Server.Kestrel.Internal.Http.Frame`1.<RequestProcessingAsync>d__2.MoveNext()
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  --- End of stack trace from previous location where exception was thrown ---
     at Microsoft.AspNetCore.Hosting.Internal.RequestServicesContainerMiddleware.<Invoke>d__3.MoveNext()
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  --- End of stack trace from previous location where exception was thrown ---
     at Google.Cloud.Diagnostics.AspNetCore.ErrorReportingExceptionLoggerMiddleware.<Invoke>d__3.MoveNext()
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  --- End of stack trace from previous location where exception was thrown ---
     at Google.Cloud.Diagnostics.Common.GrpcErrorEventConsumer.<ReceiveAsync>d__4.MoveNext()
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  --- End of stack trace from previous location where exception was thrown ---
     at Google.Api.Gax.Grpc.ApiCallRetryExtensions.<>c__DisplayClass0_0`2.<<WithRetry>b__0>d.MoveNext()
     at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
     at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
  Grpc.Core.RpcException: Status(StatusCode=DeadlineExceeded, Detail="Deadline Exceeded")
```

My first instinct from reading through the source is that [this line](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/ab44b24aa869c880ff7c834588b4f085f0850f29/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/ErrorReporting/ErrorReportingExceptionLoggerMiddleware.cs#L57) should have a try/catch of its own and, if it were to fail, the original Exception and this new Exception that was raised during logging should be combined into an [`AggregateException`](https://msdn.microsoft.com/en-us/library/system.aggregateexception(v=vs.110).aspx), and the aggregate should be thrown. If the logging succeeds, the original exception can be rethrown via a naked `throw;` as its does today.

Thoughts?

iantalarico:
Currently when an exception occurs in an application and the diagnostics library needs to report this state it will and then rethrow.  However, if the diagnostics library throws and exception the application's exception is lost. Instead rethrow both as an AggregateException so no information is lost.

Fixes #1350

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/ErrorReporting/ErrorReportingExceptionLoggerMiddlewareTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/ErrorReporting/ErrorReportingExceptionLoggerMiddlewareTest.cs
@@ -40,10 +40,24 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
         public async Task Invoke_LogsAndThrows()
         {
             var mockLogger = new Mock<IExceptionLogger>();
-            RequestDelegate requestDelegate = context => { throw new Exception(); };
+            RequestDelegate requestDelegate = context => { throw new DivideByZeroException(); };
             var middleware = new ErrorReportingExceptionLoggerMiddleware(requestDelegate, mockLogger.Object);
 
-            await Assert.ThrowsAsync<Exception>(() => middleware.Invoke(new DefaultHttpContext()));
+            await Assert.ThrowsAsync<DivideByZeroException>(() => middleware.Invoke(new DefaultHttpContext()));
+
+            mockLogger.Verify(l => l.LogAsync(It.IsAny<Exception>(), It.IsAny<DefaultHttpContext>(), CancellationToken.None), Times.Once());
+        }
+
+        [Fact]
+        public async Task Invoke_LogThrowsAndThrows()
+        {
+            var mockLogger = new Mock<IExceptionLogger>();
+            mockLogger.Setup(l => l.LogAsync(It.IsAny<Exception>(), It.IsAny<DefaultHttpContext>(), CancellationToken.None))
+                .Throws(new ArgumentOutOfRangeException());
+            RequestDelegate requestDelegate = context => { throw new DivideByZeroException(); };
+            var middleware = new ErrorReportingExceptionLoggerMiddleware(requestDelegate, mockLogger.Object);
+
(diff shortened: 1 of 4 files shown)


jskeet:
What might be catching the exception thrown here? I'm worried that if something is set up to catch FooException, it might fail to do so if it gets an AggregateException *containing* a FooException.

iantalarico:
It is possible that something could be expecting FooException, I considered just re-throwing but then we would swallow all exceptions from our libs (which may be useful if the user doesn't have auth set up).

jskeet:
It's tricky. Is it worth making this an option?

iantalarico:
It already kind of is.  We by default  swallow all grpc exceptions, so this will only happen if a user explicitly sets the option to throw exceptions from our libs.

I'll add some extra note to our docs for these two libraries about how these are handled.

PR #1353, opened 2017-08-11
Add an explicit note about when an AggregateException will be thrown in AspNetCore Diagnostics libraries

iantalarico:
Extra comments for #1352


iantalarico:
Thanks Jon!

PR #1359, opened 2017-08-15
Initial API skeleton for BQ ADO.NET

benwulfe:
finally getting on this jon.  I dont think this will be too bad.
I'm pulling into a main repo branch and will iterate there before pushing into main/master.  sound good?

My biggest concerns so far:

namespace:  Its inconsistent with the overall approach for spanner where we had a "data" namespace completely isolated from the vkit ns.  This approach means that when we move to grpc, it will be a new namespace or breaking change -- but I don't see another clean alternative here.  Maybe we should consider this approach for spanner?

transactions:  I'm not sure how bigquery implicit transactions (which are not idempotent) are going to work here with transactionscope especially.  There might be places where we need to push on BQ folks to give us a better ACID-like mechanism (like at least idempotency) so we don't create false assumptions from consumers.

type conversions:  We might want to do a shared library depending on how similar BQ is.

all tests continue to pass...

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/CodeHealthTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/CodeHealthTest.cs
@@ -22,7 +22,8 @@ namespace Google.Cloud.BigQuery.V2.Tests
         [Fact]
         public void PrivateFields()
         {
-            CodeHealthTester.AssertAllFieldsPrivate(typeof(BigQueryClient));
+            // BigQueryProviderFactory must have a public Instance field to work properly as its loaded and instantiated via reflection.
+            CodeHealthTester.AssertAllFieldsPrivate(typeof(BigQueryClient), new []{ typeof(BigQueryProviderFactory)});
         }
 
         [Fact]
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
@@ -1,0 +1,76 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Data;
+using System.Data.Common;
+
+namespace Google.Cloud.BigQuery.V2
+{
+    public sealed partial class BigQueryCommand : DbCommand
+    {
+        /// <inheritdoc />
+        public override void Cancel()
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        public override int ExecuteNonQuery()
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        public override object ExecuteScalar()
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        public override void Prepare()
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        public override string CommandText { get; set; }
+        /// <inheritdoc />
+        public override int CommandTimeout { get; set; }
+        /// <inheritdoc />
+        public override CommandType CommandType { get; set; }
+        /// <inheritdoc />
+        public override UpdateRowSource UpdatedRowSource { get; set; }
+        /// <inheritdoc />
+        protected override DbConnection DbConnection { get; set; }
+        /// <inheritdoc />
+        protected override DbParameterCollection DbParameterCollection { get; }
+        /// <inheritdoc />
+        protected override DbTransaction DbTransaction { get; set; }
+        /// <inheritdoc />
+        public override bool DesignTimeVisible { get; set; }
+
+        /// <inheritdoc />
+        protected override DbParameter CreateDbParameter()
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <inheritdoc />
+        protected override DbDataReader ExecuteDbDataReader(CommandBehavior behavior)
+        {
+            throw new NotImplementedException();
+        }
+    }
+}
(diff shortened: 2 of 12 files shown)


jskeet:
Branch: Yes, that's absolutely fine.

Namespace: I'm okay with the difference here. The two libraries *are* structured somewhat differently, and there are far more examples of using the BigQuery API directly than the Spanner API, so it makes sense for the API/ADO split to be more balanced here than with Spanner. We've always said that even for Spanner.Data, adopting Spanner.V2 would mean a major version bump (so *potential* breaking changes) - but they're far less likely than they are in this case.

Transactions: I suspect this will me more a matter of documentation, to be honest. BigQuery is much more a data warehouse than OLTP service, and we just need to make that clear. Worth investigating though.

Type conversions: Maybe. Dependencies make me nervous - if it's only going to be in two places, I'm happy for the *moment* with copy/paste as a code reuse mechanism...

Will review later today.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/CodeHealthTest.cs
```diff
@@ -22,7 +22,8 @@ public class CodeHealthTest
         [Fact]
         public void PrivateFields()
         {
-            CodeHealthTester.AssertAllFieldsPrivate(typeof(BigQueryClient));
+            // BigQueryProviderFactory must have a public Instance field to work properly as its loaded and instantiated via reflection.
```
jskeet: Eek. Is that a requirement in ADO.NET in general? What's doing that?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
```diff
@@ -0,0 +1,76 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Data;
+using System.Data.Common;
+
+namespace Google.Cloud.BigQuery.V2
+{
+    public sealed partial class BigQueryCommand : DbCommand
```
jskeet: I like this way of splitting the ADO.NET aspect from the rest, within the same class.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -0,0 +1,210 @@
...
+using System.Data;
+using System.Data.Common;
+
+namespace Google.Cloud.BigQuery.V2
+{
+    /// <summary>
+    /// </summary>
+    public sealed class BigQueryDataReader : DbDataReader
+    {
+#if NET45 || NET451
```
jskeet: Do we need to build both net45 and net451? If we're targeting net45, let's just use that in the `#if`.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.csproj
```diff
@@ -27,6 +27,9 @@
     <PackageReference Include="SourceLink.Create.CommandLine" Version="2.1.2" PrivateAssets="All" />
     <Analyzer Condition="Exists('..\..\..\tools\Google.Cloud.Tools.Analyzers\bin\$(Configuration)\netstandard1.3\publish\Google.Cloud.Tools.Analyzers.dll')" Include="..\..\..\tools\Google.Cloud.Tools.Analyzers\bin\$(Configuration)\netstandard1.3\publish\Google.Cloud.Tools.Analyzers.dll" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'net45'">
```
jskeet: I *think* this should already work, but I think it would be clearer to move this after the dotnet pack instructions, so we have all the autogenerated stuff first.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/CodeHealthTest.cs
benwulfe: https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/obtaining-a-dbproviderfactory

psuedo magic mechanisms for obtaining the provider factory where you only specify its type information in your config.  Why they didnt opt to simply specify the public member (which could be a property) I have no idea -- but its designed to automatically look for a public static field when building the providers from config.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
benwulfe: there will be some overlap of members due to name conflicts where original members become overrides.
If you don't mind, I'll move those over as well when I'm going thru it.  someone looking at this after its done wouldn't care about the distinction between methods that existed previously.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
benwulfe: yes just net45 should work

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
jskeet: Sure, no problem.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/CodeHealthTest.cs
jskeet: Ah well. At least it's read-only :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameterCollection.cs
```diff
@@ -75,16 +75,16 @@ public BigQueryParameter Add(BigQueryParameter parameter)
...
         /// <inheritdoc />
-        public BigQueryParameter this[int index] => _parameters[index];
+        public new BigQueryParameter this[int index] => _parameters[index];
 
         /// <inheritdoc />
-        public IEnumerator<BigQueryParameter> GetEnumerator() => _parameters.GetEnumerator();
+        IEnumerator<BigQueryParameter> IEnumerable<BigQueryParameter>.GetEnumerator() => _parameters.GetEnumerator();
 
         /// <inheritdoc />
         IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
```
evildour: This seems like it can be removed now.

benwulfe: I'll verify when i do the implementation.

PR #1361, opened 2017-08-15
Change how we run integration tests

jskeet:
(I'm still testing this, but thought I'd show you while I'm doing so.)

- Instead of --continue, have --retry, which retries failures
  (We store failures in a file, rather than successes.)
- Continue on failure, having emitted the failure

This means that from Jenkins we can just run:

```sh
./runintegrationtests.sh --coverage
./runintegrationtests.sh --retry --coverage
./runintegrationtests.sh --retry --coverage
```

If there are no failures, the retries will be no-ops basically...
and otherwise, we'll have a list of test failures.

While we could retry a given number of times within the script,
doing it this way improves the chances of success in the face of
transient errors and makes quota failures less likely.

### DIFF
--- a/runintegrationtests.sh
+++ b/runintegrationtests.sh
@@ -6,53 +6,58 @@
 set -e
 source toolversions.sh
 
-CONTINUE_ARG=
+RETRY_ARG=
 COVERAGE_ARG=
 
 for arg in "$@"
 do
   case $arg in
-  --continue)
-    CONTINUE_ARG=yes
+  --retry)
+    RETRY_ARG=yes
     ;;
   --coverage)
     install_dotcover
     COVERAGE_ARG=yes
     ;;
   *)
-    echo "Unknown argument: $arg. Supported arguments: --coverage --continue"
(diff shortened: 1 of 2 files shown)


runintegrationtests.sh
```diff
@@ -6,53 +6,58 @@
 set -e
```
benwulfe: take out set -e?

benwulfe:
lgtm.  or we could opt to retry at different times using jenkins (ie space them 20 minutes apart) with a report run at the end.

runintegrationtests.sh
jskeet: Nope, I think we still want to fail if the script itself has bugs.

jskeet:
Agreed on the spacing out - if it's only run once a day, it doesn't really matter much how long it takes.

iantalarico:
LGTM!

jskeet:
Okay, tested with a deliberate failure, and the retry worked fine. Merging.

PR #1371, opened 2017-08-17
Analyzer fixes after upgrading to VS 15.3

### LINKED ISSUE #1370: Analyzer throwing errors
> Analyzer 'Google.Cloud.Tools.Analyzers.InternalOptionalParametersRequiredAnalyzer' threw an exception of type 'System.MissingMethodException' with message 'Method not found: 'System.Collections.Immutable.ImmutableArray`1<Microsoft.CodeAnalysis.Semantics.IArgument> Microsoft.CodeAnalysis.Semantics.IHasArgumentsExpression.get_ArgumentsInParameterOrder()'.'.	

Will investigate, until @evildour wants to...

evildour:
Fixed #1370

### DIFF
--- a/tools/Google.Cloud.Tools.Analyzers/Google.Cloud.Tools.Analyzers.csproj
+++ b/tools/Google.Cloud.Tools.Analyzers/Google.Cloud.Tools.Analyzers.csproj
@@ -5,8 +5,9 @@
     <TargetFrameworks>netstandard1.3</TargetFrameworks>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="2.0.0" />
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.0.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis" Version="2.3.0" />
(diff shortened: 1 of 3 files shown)


evildour:
I think it should be ok: for the analyzer it may just change the order in which the diagnostics are reported and for the code fix, the specified args only help us determine if a bool is true or false so order won't matter, and for the default args, order _might_ change (not sure) but their evaluation order won't matter since they can only be constants.

evildour:
Hmm, for some reason on travis the Microsoft.CodeAnalysis package being restored is 2.3.1, but it fails looking for 2.3.0 when building. Going to try adding an explicit package reference.

evildour:
Nope, that didn't work. Upon close inspection it appears the problem occurs when building a project that uses the analyzers. Maybe 2.3.0 is the latest available for use during analysis. Bumping versions down to see if that fixes it.

evildour:
Still not working. This seems to be the same issue: https://github.com/dotnet/roslyn-analyzers/issues/1215. Looks like we need to use a newer version of CSC. Is that something you specify to travis?

evildour:
Perhaps setting the dotnet version in the .travis.yml to 1.1.0? What do you think?

evildour:
Superseded by #1376

PR #1372, opened 2017-08-17
Update Moq to 4.7.99

### LINKED ISSUE #1369: Fix warnings under VS15.3
I'm receiving a lot of warnings for NU1603:

> Castle.Core 4.0.0 depends on System.ComponentModel.TypeConverter (>= 4.0.1) but System.ComponentModel.TypeConverter 4.0.1 was not found. An approximate best match of System.ComponentModel.TypeConverter 4.1.0 was resolved.

Haven't delved into it yet. (It's only in tests. Maybe we need to update xUnit...)

evildour:
Fixes #1369


evildour:
I think the build failure is due to a regression in Moq. Mock.Setup is [documented](https://github.com/moq/moq4/blob/master/Source/Mock.Generic.xdoc#L126) that it overwrites previous calls, which it did previously, but doesn't seem to be doing now. We can easily work around this though.

evildour:
Reported Moq issue: https://github.com/moq/moq4/issues/433

evildour:
AppVeyor build broken due to the VS 2017 image being updated to 2.0. Going to merge this then see about fixing this in #1371.

jskeet:
The AppVeyor build looks like it's at least *partly* broken due to not running shell scripts by default - I'm seeing whether running it explicitly with bash will help. But as we know it now does contain the .NET Core 2 .0 tools, let's see if we can go all the way...

PR #1375, opened 2017-08-17
Enable running Google.Cloud.Diagnostics.AspNetCore integration tests in parallel.

iantalarico:
This was a restriction put in place when the tests were not as robust but is no longer needed.


jskeet:
LGTM if the tests go green after rebasing to fix the AppVeyor issues.

PR #1378, opened 2017-08-18
Explicitly run bash in AppVeyor scripts

jskeet:
(Hoping this fixes AppVeyor without us moving to .NET Core 2.0, as a first step.)


jskeet:
This doesn't fully fix AppVeyor, but it gets a lot further. Doc building fails now.

PR #1379, opened 2017-08-18
Upgrade docfx to version 2.22.3

jskeet:
This change uses Microsoft.Net.Compilers 2.3.1, which may fix the
AppVeyor doc build (which has been broken due to AppVeyor updating
to .NET Core 2.0, for reasons not fully understood).


jskeet:
We're going to run docfx from a separate Jenkins instance we own, rather than getting it to work on AppVeyor. We'll want to update docfx in another PR, but only when it handles default literals. Closing this PR.

PR #1382, opened 2017-08-18
Set MSBuildSDKs path

jskeet:


jskeet:
We're going to run docfx from a separate Jenkins instance we own, rather than getting it to work on AppVeyor. Closing this PR.

PR #1383, opened 2017-08-18
Temporarily disable building docs on AppVeyor

jskeet:
This will get us back to green after the Core 2.0 change, but
obviously we want to go back to building the docs in CI.


jskeet:
(Merging without review, as Chris had given the OK for the principle, and it's only comments...)

PR #1386, opened 2017-08-18
Move to C# 7.1

evildour:


tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
```diff
@@ -292,6 +292,7 @@ private static void GenerateMainProject(ApiMetadata api, string directory, HashS
                 // Build-related properties
                 new XElement("Version", api.Version), // TODO: Version, or VersionPrefix/VersionSuffix?
                 new XElement("TargetFrameworks", targetFrameworks),
+                new XElement("LangVersion", "latest"),
```
jskeet: Do we want to go for latest, or explicitly 7.1? The latter means we don't get broken implicitly when 7.2 comes out, if there are any breaking changes... but it means we need to go through this again. I'm in two minds.

evildour: Well I guess technically we had that situation already, right? We were implicitly opted into the latest major version, which arguably has more of a chance for breaking changes, right? I assume minor versions would be safer in that regard. But if you feel otherwise, I can make it explicit.

jskeet: True, although I believe the language team is *more* likely to make "breaking in corner cases" changes within minor versions if they're changing a new feature that hasn't been used as much yet. (It's not really semver, basically.)

Let's stick with latest :)

LGTM.

evildour: Makes sense. At least there's an easy fix if there's ever an issue. Thanks.

PR #1387, opened 2017-08-18
Move tests to C# 7.1

evildour:


evildour:
AppVeyor seems to be stuck. I think this should be safe though since it's very likely to build and travis passed. Going to merge.

jskeet:
SGTM.

PR #1388, opened 2017-08-18
Cleanup default expressions

evildour:


jskeet:
I'm doing a build locally to check that docfx is happy with this too.

jskeet:
This is just for the manually-written code at the moment, right?

@chrisdunelm presumably it won't be hard to change the codegen for DatastoreClient etc?

evildour:
Correctly, I went through all changed files and manually reverted the auto-generated ones. If you think I missed any, let me know. Most of them were ...Grpc.cs files from the protobuf compiler.

jskeet:
SGTM. I suggest we then change the generator, regenerate, then chase stragglers. Still building...

jskeet:
Docs build failing at the moment:

> Error:Error extracting metadata for C:/Users/jon/Test/Projects/google-cloud-dotnet/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.csproj: Value cannot be null.

I haven't actually tried *without* this change yet - will do that now. Upgrading docfx could sort it anyway.

jskeet:
Nope - works in `master`, fails in this - even with docfx 2.22.3.

Will file a bug with docfx on Monday (with a short but complete example).

evildour:
Thanks Jon. I'll leave this open until that gets resolved.

jskeet:
Hmm - I can't reproduce the docfx problem in a standalone sample, so this will take a bit more investigation.

jskeet:
For anyone interested, I've [blogged](https://codeblog.jonskeet.uk/2017/08/19/diagnosing-a-corner-case-tooling-bug/) about the process of diagnosing the bug, and [filed an issue](https://github.com/dotnet/docfx/issues/1982) with docfx.

jskeet:
I'd hoped this was fixed in docfx 2.23.1, but it generates invalid (and partial) API metadata.

evildour:
Based on your analysis, it seems we could work around this by changing some expression bodied methods to normal methods. But maybe we should wait a bit longer to see what happens with the docfx issue.

jskeet:
Would definitely rather wait. I suspect it may well be sorted soon.

jskeet:
This can be merged now after a rebase.

PR #1397, opened 2017-08-23
Fix Moq test the right way

evildour:
This fixes an incorrect "workaround" in #1372 for something I thought was a Moq regression, but it was just my misunderstanding how Moq works. Lesson learned: don't modify captured variables in argument match conditions for Moq setups.


ILMTitan:
Right, because `Mock.Setup` does not take a `Function<...>`, it takes an `Expression<Function<...>>`

evildour:
That was not what I got wrong. I mistakenly assumed the setup with match predicates on the arguments was _validating_ the arguments, instead of testing them to see if the setup applies. If you're interested, [here](https://github.com/moq/moq4/issues/433#issuecomment-323907504) is where I realized my mistake and then had it explained to me.

PR #1401, opened 2017-08-25
Bump Google.Cloud.Diagnostics.* to GA

iantalarico:
@jskeet mind doing a quick double check?

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -130,18 +130,18 @@
     "tags": [ "Error", "Reporting", "Stackdriver", "ExceptionLogger", "Trace", "Diagnostics" ],
     "dependencies": {
       "Google.Cloud.Diagnostics.Common": "project",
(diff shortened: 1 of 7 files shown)


jskeet:
Will do when I get a chance, but I don't know when I'll be back on a laptop

iantalarico:
@jskeet No rush, I just wanted a second set of eyes.  I keep forgetting your on vacation.  I'll add Mike.

apis/apis.json
```diff
@@ -111,7 +111,7 @@
       "Microsoft.AspNet.WebApi.Core": "5.2.3",
```
jskeet: Is it worth checking these dependency versions are appropriate? Are there more recent versions that are worth picking up, for example?

iantalarico: Good catch, bumped all to the latest stable release, thanks Jon!

iantalarico:
Thanks Jon :)

PR #1412, opened 2017-08-29
Update test dependencies

### LINKED ISSUE #1409: Update dependencies for tests
We're using older versions of the test SDK etc than we probably should be.

Having updated one project to xUnit 2.3.0-beta4, I've found that it complains about some tests we're making, such as

```csharp
Assert.Equal(5, collection.Count);
```

... which seems like a reasonable test to me. We can disable these in the warnings section - in this case it's `xUnit2013` than needs ignoring.

jskeet:
This fixes #1409, and keeps us on our toes with xUnit.

Currently on my machine I'm getting errors in Visual Studio about long paths when generating assembly binding redirects in some projects, but:

- That may be due to where my code sits on disk
- It doesn't happen when building from the command line
- It doesn't seem to stop the rest of VS from working

### DIFF
--- a/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
+++ b/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
@@ -60,9 +60,10 @@ namespace Google.Cloud.Tools.ProjectGenerator
         // the packages in DefaultPackageVersions should be specified precisely in stable packages.
         private static readonly Dictionary<string, string> CommonTestDependencies = new Dictionary<string, string>
         {
-            { "Microsoft.NET.Test.Sdk", "15.0.0" },
-            { "xunit", "2.3.0-beta1-build3642" },
-            { "xunit.runner.visualstudio", "2.3.0-beta1-build1309" },
+            { "Microsoft.NET.Test.Sdk", "15.3.0" },
+            { "xunit", "2.3.0-beta4-build3742" },
+            { "xunit.runner.visualstudio", "2.3.0-beta4-build3742" },
+            { "xunit.analyzers", "0.6.1" }, // Temporary measure to fix some broken analyzers in 0.6.0 that throw exceptions
             { "Moq", "4.7.99" }
         };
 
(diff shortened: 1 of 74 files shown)


jskeet:
Hmm. Moving the project elsewhere on disk didn't help with the binding redirects. As an example, it's failing when I open `Google.Cloud.Diagnostics.AspNetCore.sln` - the failure points at `Google.Cloud.Diagnostics.Common.IntegrationTests`. I haven't worked out which binding redirect is causing a problem yet.

tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
```diff
@@ -359,9 +360,12 @@ private static void GenerateTestProject(ApiMetadata api, string directory, HashS
                     new XElement("PublicSign", new XAttribute("Condition", " '$(OS)' != 'Windows_NT' "), true),
                     new XElement("TreatWarningsAsErrors", true),
                     // 1701, 1702 and 1705 are disabled by default.
-                    // 4014 is required as snippets for streaming samples call Task.Run and don't await the result.
-                    // See https://github.com/googleapis/toolkit/issues/1271 - when that's fixed, we can remove this.
-                    new XElement("NoWarn", "1701;1702;1705;4014")
+                    // xUnit2004 prevents Assert.Equal(true, value) etc, preferring Assert.True and Assert.False, but
+                    //   Assert.Equal is clearer (IMO) for comparing values rather than conditions.
+                    // xUnit2013 prevents simple checks for the number of items in a collection
+                    // AD001 is the error for an analyzer throwing an exception. We can remove this when
```
chrisdunelm: Super-nitty nit: missing a "0" in AD001 in this comment

jskeet: Thanks, will fix.

jskeet:
Will leave this just now until I've had a chance to investigate the VS issue a bit more. That will probably take a few hours of diagnosis. If we can't get anywhere, we should probably merge anyway.

jskeet:
We've decided to merge this despite the VS issues, which we'll investigate separately. Otherwise this will get increasingly out of date with other changes (in particular, dependency changes we're planning to make very soon.)

jskeet:
Blogged about the VS issue here: https://codeblog.jonskeet.uk/2017/08/30/diagnosing-a-vs-only-build-problem/

PR #1413, opened 2017-08-29
Pull default timeout from commitsettings in case of gapic change.

benwulfe:
Caught when reviewing a default timeout change from the spanner team.
We need to be pulling the default timeout directly from generated code instead of having a hard coded value.
Also added a test that will fail if the generated settings change in an incompatible way.

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using System;
+using System.Diagnostics;
 
 namespace Google.Cloud.Spanner.V1
 {
@@ -21,6 +22,16 @@ namespace Google.Cloud.Spanner.V1
     /// </summary>
     public sealed class SessionPoolOptions
     {
+        /// <summary>
+        /// Constructs a new <see cref="SessionPoolOptions"/>.
+        /// </summary>
+        public SessionPoolOptions()
+        {
+            var retry = SpannerSettings.GetDefault().CommitSettings.Timing.Retry;
+            // ReSharper disable once PossibleInvalidOperationException
+            Timeout = (int)(retry?.TotalExpiration.Timeout).Value.TotalSeconds;
+        }
+
         /// <summary>
         /// Maximum number of active sessions that can be in use by the application at any one time.
         /// </summary>
@@ -56,7 +67,7 @@ namespace Google.Cloud.Spanner.V1
         /// <summary>
         /// The total time in seconds allowed for a network call to the Cloud Spanner server, including retries.
         /// </summary>
-        public int Timeout { get; set; } = 60;
+        public int Timeout { get; set; }
 
         /// <summary>
         /// The maximum number of session create operations allowed to occur simultaneously.
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SessionPoolTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SessionPoolTests.cs
@@ -484,6 +484,16 @@ namespace Google.Cloud.Spanner.Data.Tests
             }
         }
 
+        [Fact]
+        public void SessionPoolDefaultTimeout()
+        {
+            SessionPoolOptions options = new SessionPoolOptions();
+            Assert.Equal(options.Timeout,
+                // ReSharper disable once PossibleInvalidOperationException
+                (int)SpannerSettings.GetDefault().CommitSettings.Timing.Retry
+                    .TotalExpiration.Timeout.Value.TotalSeconds);
+        }
+
         private class ParallelSpannerClient : SpannerClient
         {
             private int _concurrentRequests = 0;


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
```diff
@@ -22,6 +23,24 @@ namespace Google.Cloud.Spanner.V1
...
+            var retry = SpannerSettings.GetDefault().CommitSettings.Timing.Retry;
+            var timeout = retry?.TotalExpiration.Timeout;
+            if (timeout.HasValue)
+            {
+                Timeout = (int)timeout.Value.TotalSeconds;
+            }
+            else
+            {
+                Debug.Assert(false, "Config change has invalidated default timeout calc");
+                Timeout = (int)TimeSpan.FromMilliseconds(36000000).TotalSeconds;
```
vkedia: 10 hours?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
```diff
@@ -22,6 +23,24 @@ namespace Google.Cloud.Spanner.V1
...
         /// <summary>
+        /// Constructs a new <see cref="SessionPoolOptions"/>.
+        /// </summary>
+        public SessionPoolOptions()
+        {
+            var retry = SpannerSettings.GetDefault().CommitSettings.Timing.Retry;
+            var timeout = retry?.TotalExpiration.Timeout;
+            if (timeout.HasValue)
+            {
+                Timeout = (int)timeout.Value.TotalSeconds;
```
vkedia: which rpcs is this timeout used for?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
benwulfe: that's whats in gapic:

https://github.com/googleapis/googleapis/pull/424/files#diff-4501db3e3507bcce8496c9aea2017d96R54

bug there?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
benwulfe: in ado.net there is no way to distinguish since the timeout is at the command level.
So this timeout will be used for commit, but also for session create, rollback and begin transaction.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
vkedia: Oops. That is bug in there. Let me fix it first.

vkedia: https://github.com/googleapis/googleapis/pull/438 sent

benwulfe: thanks.  I'll update this PR with a code regen after that's merged.

jskeet: It's not clear to me why we need this "else" part. If it's guarding against the settings having been changed in the YAML file incorrectly, won't the unit test catch that? We wouldn't release something without running the unit tests, so shouldn't that be okay?

I'd prefer for us to simply assume everything's okay in the production code, at which point you could get rid of the constructor entirely and just use a property initializer:

```csharp
public int Timeout { get; set; } = (int) SpannerSettings.GetDefault().CommitSettings.Timing.Retry.TotalExpiration.Timeout.TotalSeconds;
```

benwulfe: it's actually a bit difficult to debug when this changes underneath you.  I initially used a property initializer, got the initialization wrong and proceeded to get a whole bunch of typeloadexceptions from the static initializer of sessionpooloptions which indirectly creates once of these in its cctor.

If you *really* feel strong about this I'll consider changing, but after that experience, I'd rather it fail in a very clear and diagnosable way.

jskeet: Happy to have it in the constructor for easier debugging. (It will still fail in the cctor of SessionPoolOptions if it's in the constructor here, of course... but it may be easier to break on.)

I think I'd still rather not try to "recover" from it being broken though, with the else block you've currently got. Does that make sense?

benwulfe: It should never happen in a released build, so I'm fine with that.  I've been waiting for gapic config to be updated so I can regen the libs -- I'll update this change with that (it will have 2 separate commits).

jskeet: Regen is now merged, so assuming it was the one you were waiting for, it should be simple after a rebase.

PR #1420, opened 2017-08-31
Fix Google.Cloud.VideoIntelligence.V1Beta2 snippets.

iantalarico:
Looks like the API changed from beta1 to beta2.  This caused our build docs Jenkins job to fail.


jskeet:
Gah - sorry for missing that before. LGTM, assuming the docs now build :)

iantalarico:
No worries :)  local build just finished and all is good.

PR #1422, opened 2017-08-31
Initial checking for ado.net support for BQ.

benwulfe:
Adds the implementation for Connection, Command, DataReader, Parameters
Missing:
* Proper docs
* Tests that cover non-generated code
* Test to see if the code generate should be rerun
* Other methods that should not be generated, but used as Commands (eg ListRows)
* Did not yet do some method renames to keep existing code compatible.
* nit:There are some cases of generated comments mentioning a deleted parameter.

This checkin has the overall structure and framework for how we can flip methods between exposing them directly on Connection versus as a Command.
So if we wanted to expose ListRows, we'd
a) add it to the excluded method list (so no codegen)
b) create a options class for it that also implements ExecuteDbReader (see BigQueryConnection.Queries.cs for the sql version)
c) add a new overload such as CreateListRowsCommand(...) (again see BigQueryConnection.Queries.cs)


jskeet:
Hey Ben, I'll try to get to this ASAP - but I need to focus on perf review today.

Is there anything I could review quickly that would be useful to unblock you from further work before I'm done on perf?

benwulfe:
I'm not blocked, but I don't want to create even more for you to review
later.  If that doesn't bother you, feel free to punt to next week.

I'm currently working on tests and comments and the few items left.

jskeet:
It bothers me in that I dislike huge code reviews, but I think in this case that it's a necessary evil - go ahead :) (Adding the comments will make the code review easier rather than harder, mind you...)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -14,197 +14,207 @@
...
-            throw new NotImplementedException();
-        }
+        public override bool NextResult() => false;
 
         /// <inheritdoc />
         public override bool Read()
-        {
-            throw new NotImplementedException();
-        }
+            => ReadAsync(CancellationToken.None).ResultWithUnwrappedExceptions();
```
jskeet: I see the point of using the async API here - basically we can't mix and match between Read and ReadAsync easily in terms of the use of BigQueryResults.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -14,197 +14,207 @@
...
 
         /// <inheritdoc />
-        public override int Depth { get; }
+        public override async Task<bool> ReadAsync(CancellationToken cancellationToken)
+            => _isEnumerating = await MoveNextAsync().ConfigureAwait(false);
```
jskeet: Use the cancellation token

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -14,197 +14,207 @@
...
+
         /// <inheritdoc />
-        public override bool IsClosed { get; }
+        public override int Depth => 0;
+
         /// <inheritdoc />
-        public override int RecordsAffected { get; }
+        public override bool HasRows => _isEnumerating && _results?.Current != null;
```
jskeet: So if we haven't called Read, this will return false always, because Current will return null - is that appropriate?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -14,197 +14,207 @@
...
-        }
+        public override Type GetFieldType(int i) => Current?.Schema.Fields[i].ToClrType();
 
         /// <inheritdoc />
-        public override int GetValues(object[] values)
-        {
-            throw new NotImplementedException();
-        }
+        public override T GetFieldValue<T>(int ordinal)
+            => (T) Convert.ChangeType(Current?[ordinal], typeof(T));
```
jskeet: Pass in CultureInfo.InvariantCulture?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -14,197 +14,207 @@
...
-            throw new NotImplementedException();
-        }
+        public override decimal GetDecimal(int i) => GetFieldValue<decimal>(i);
 
         /// <inheritdoc />
-        public override double GetDouble(int ordinal)
-        {
-            throw new NotImplementedException();
-        }
+        public override double GetDouble(int i) => GetFieldValue<double>(i);
```
jskeet: Do we want this to perform conversions from string values that look like doubles to double values, etc? Maybe that's just the ADO.NET way of things. Urgh if so.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -14,197 +14,207 @@
...
+        private BigQueryRow Current => _isEnumerating ? _results?.Current : null;
+
+        private BigQueryResults BigQueryResults { get; }
+
+        private Task<bool> MoveNextAsync() => GetResultsAsync().MoveNext();
+
+        private IAsyncEnumerator<BigQueryRow> GetResultsAsync()
         {
-            throw new NotImplementedException();
+            if (_results != null)
```
jskeet: As far as I remember, GetRowsAsync doesn't immediately make any calls - so perhaps it would be simpler to do this eagerly on construction? I believe we only start fetching data when MoveNext() is falled. Probably worth checking though.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -14,197 +14,207 @@
...
-            throw new NotImplementedException();
-        }
+        public override double GetDouble(int i) => GetFieldValue<double>(i);
 
         /// <inheritdoc />
-        public override float GetFloat(int ordinal)
+        public override IEnumerator GetEnumerator()
         {
-            throw new NotImplementedException();
+#if NET45 || NET451
```
jskeet: Only NET45 required, I'd expect

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -14,197 +14,207 @@
...
+            return new DbEnumerator(this);
```
jskeet: Do I need to look into what DbEnumerator does? Does it ever start iterating from the start again?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDbType.cs
```diff
@@ -100,5 +101,38 @@ internal static class BigQueryDbTypeExtensions
         };
 
         internal static string ToParameterApiType(this BigQueryDbType type) => s_typeToNameMapping[type];
+
+        internal static Type ToClrType(this BigQueryDbType type)
```
jskeet: Maybe use a `Dictionary<BigQueryDbType, type>`?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameter.Data.cs
```diff
@@ -15,33 +15,124 @@
...
 {
-    public sealed partial class BigQueryParameter : DbParameter 
+    public sealed partial class BigQueryParameter : DbParameter
+#if NET45 || NET451
+        , ICloneable
+#endif
     {
         /// <inheritdoc />
-        public override void ResetDbType()
+        public override DbType DbType
```
jskeet: Again, maybe dictionaries could simplify this? (Ideally, only specify the mapping in one place and reverse it...) Maybe an internal attribute on BigQueryDbType?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameterCollection.Data.cs
```diff
@@ -26,105 +27,109 @@ public sealed partial class BigQueryParameterCollection : DbParameterCollection
...
-            throw new NotImplementedException();
-        }
+            => _parameters.Insert(index, (BigQueryParameter)value);
 
         /// <inheritdoc />
         public override void Remove(object value)
-        {
-            throw new NotImplementedException();
-        }
+            => _parameters.Remove((BigQueryParameter)value);
```
jskeet: Do we want this to throw an exception if the value isn't a BigQueryParameter? I'd expect it to basically ignore any value not in the collection.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameterCollection.Data.cs
```diff
@@ -26,105 +27,109 @@ public sealed partial class BigQueryParameterCollection : DbParameterCollection
...
-        {
-            throw new NotImplementedException();
-        }
+        public override void RemoveAt(int index) => _parameters.RemoveAt(index);
 
         /// <inheritdoc />
         public override void CopyTo(Array array, int index)
         {
-            throw new NotImplementedException();
+            if (array == null)
```
jskeet: GaxPreconditions for both of these.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameterCollection.Data.cs
```diff
@@ -26,105 +27,109 @@ public sealed partial class BigQueryParameterCollection : DbParameterCollection
...
+            {
+                throw new ArgumentNullException(nameof(array));
+            }
+            if (array.Length < _parameters.Count + index)
+            {
+                throw new ArgumentOutOfRangeException(
+                    nameof(array), "There is not enough space in the array to copy values.");
+            }
+            foreach (var item in _parameters)
```
jskeet: Can't we use _parameters.CopyTo?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetQueryResultsOptions.cs
```diff
@@ -61,6 +61,13 @@ internal void ModifyRequest(GetQueryResultsRequest request)
             // Nothing to do? We may have something later...
         }
 
+        /// <summary>
```
jskeet: Trying to figure out where this is used, and indeed where options in general are stored - I suspect you know that off the top of your head. Care to help me out? :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/SqlCommandTest.cs
```diff
@@ -0,0 +1,640 @@
...
+
+                using (var sqlCmd = connection.CreateSqlCommand(sql))
+                using (var reader = await sqlCmd.ExecuteReaderAsync())
+                {
+                    var rows = 0;
+                    while (await reader.ReadAsync())
+                    {
+                        if (rows == 0)
+                        {
+                            Assert.Equal("hamlet", reader.GetFieldValue<string>("title"));
```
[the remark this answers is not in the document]
benwulfe: Don't believe SqlDataReader has that overload but could be wrong:
https://github.com/Microsoft/referencesource/blob/master/System.Data/System/Data/SqlClient/SqlDataReader.cs

GetFieldValue can be used for any type and GetString is something of a turd from pre generic days of 1.1.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/SqlCommandTest.cs
```diff
@@ -0,0 +1,640 @@
...
+        }
+
+        [Fact]
+        public async Task AsynchronousTemporaryQueryAsync()
+        {
+            // You must always place the datasource from which reads will execute in the connection string.
+            // In this case, since we only read from temp tables, we can exclude the dataset.
+            using (var connection = new BigQueryConnection($"Data Source=projects/{_fixture.ProjectId}"))
+            {
+                await connection.OpenAsync();
```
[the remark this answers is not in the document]
benwulfe: Unique to BQ due to how it runs queries by storing them in temp tables.
In Spanner (and every provider I can think of), you open the connection against what you are reading and updating and there is no notion of executing the query "elsewhere" like BQ.
However, introducing a separate option concept of an "ExecutingProject" or "LocalProject" created a lot of complications where it was unclear what context some methods would operate under (eg GetTable/GetDataSet without a project id).
For situations where you read data in the same project its stored, the connection string is as expected, where you can place the project + dataset in the string.
However, if you want to read from a public dataset using your own project's context for temp tables, you must fully qualify the table you are reading from in the SQL and specify your own project id in the connection string.

So TLDR is you always put *your* project in the connection string.  If you are reading someone else's data, fully qualify the SQL.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/SqlCommandTest.cs
```diff
@@ -0,0 +1,640 @@
...
+                            [1234567890] AS Int64,
+                            [1.234567] AS Float64,
+                            [DATE '2016-11-29'] AS Date,
+                            [TIME '12:34:56.123456'] AS Time,
+                            [DATETIME '2016-11-29 12:34:56.123456'] AS DateTime,
+                            [TIMESTAMP '2016-11-29 12:34:56.123456'] AS Timestamp"))
+                using (var reader = await sqlCmd.ExecuteReaderAsync())
+                {
+                    var rowCount = 0;
+                    while (await reader.ReadAsync())
```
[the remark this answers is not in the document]
benwulfe: yeah this was just a copy/paste thing.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/SqlCommandTest.cs
```diff
@@ -0,0 +1,640 @@
...
+            {
+                await connection.OpenAsync();
+
+                using (var sqlCmd = connection.CreateSqlCommand("SELECT STRUCT<x string>(NULL) AS s"))
+                using (var reader = await sqlCmd.ExecuteReaderAsync())
+                {
+                    var rowCount = 0;
+                    while (await reader.ReadAsync())
+                    {
+                        var s = (Dictionary<string, object>)reader["s"];
```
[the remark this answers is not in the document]
benwulfe: It's consistent with how json is used and its what spanner does so I am fine with it.
Maybe ExpandoObject?  Seems to get into the realm of EF though.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
```diff
@@ -15,62 +15,240 @@
...
+            set => CommandOptions.CommandText = value;
         }
 
         /// <inheritdoc />
-        public override int ExecuteNonQuery()
+        public override int CommandTimeout
         {
-            throw new NotImplementedException();
+            get => CommandOptions.CommandTimeout.GetValueOrDefault((int) GetQueryResultsOptions.DefaultTimeout
+                .TotalSeconds);
```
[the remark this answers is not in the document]
benwulfe: 120 I think.  it just needs to decide a better break point -- like the parameters of GetValueOrDefault.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
```diff
@@ -15,62 +15,240 @@
...
 
         /// <inheritdoc />
-        public override object ExecuteScalar()
+        public override CommandType CommandType
         {
-            throw new NotImplementedException();
+            get => CommandType.Text;
+            set
+            {
+                if (value != CommandType.Text)
```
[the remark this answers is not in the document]
benwulfe: yes codacy caught these as well.  I'll run another cleanup pass with R#

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
```diff
@@ -15,62 +15,240 @@
...
-        public override void Prepare()
+        public override UpdateRowSource UpdatedRowSource
+        {
+            get => UpdateRowSource.None;
+            set
+            {
+                if (value != UpdateRowSource.None)
+                    throw new NotSupportedException(
+                        "Cloud BigQuery does not support updating datasets on update/insert queries."
+                        + " Please use UUIDs instead of auto increment columns, which can be created on the client.");
```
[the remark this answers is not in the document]
benwulfe: this is most often used in the scenario where you have an autoincrement column in the database and you need to set the updaterowsource to "both" or "output" to have the cell in the local dataset display the correct data after the insert.  So the comment addresses the underlying use case advising the user to use UUIDs for key columns instead of autoincrement fields.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
```diff
@@ -15,62 +15,240 @@
...
         /// <inheritdoc />
-        protected override DbConnection DbConnection { get; set; }
+        public override void Cancel()
+        {
+            SynchronousCancellationTokenSource.Cancel();
+        }
+
         /// <inheritdoc />
-        protected override DbParameterCollection DbParameterCollection { get; }
+        public override object ExecuteScalar() => ExecuteScalarAsync(SynchronousCancellationTokenSource.Token)
```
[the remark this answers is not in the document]
benwulfe: its too bad you cannot tell Wait() or Result that the current thread can be used to process tasks while waiting.  that would be nice.
but yes I agree, its wasting a thread on synchronous cases.  Developer should be doing async, but we don't need to punish them too severely by doubling their thread usage if they choose poorly.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
```diff
@@ -15,62 +15,240 @@
...
+        /// <summary>
+        /// </summary>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        public async Task<T> ExecuteScalarAsync<T>(CancellationToken cancellationToken)
+        {
+            using (var reader = await ExecuteReaderAsync(cancellationToken)
+                .ConfigureAwait(false))
+            {
+                if (await reader.ReadAsync(cancellationToken).ConfigureAwait(false) && reader.HasRows &&
```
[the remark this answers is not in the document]
benwulfe: yeah no need to mention more than once on these kinds of issues.  its just an auto-code cleanup thing.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
```diff
@@ -15,62 +15,240 @@
...
+        /// <summary>
+        /// Sends the command to Cloud BigQuery and builds a <see cref="BigQueryDataReader"/>.
+        /// </summary>
+        public new Task<BigQueryDataReader> ExecuteReaderAsync()
+            => ExecuteReaderAsync(CommandBehavior.Default, CancellationToken.None);
+
+        /// <summary>
+        /// Sends the command to Cloud BigQuery and builds a <see cref="BigQueryDataReader"/>.
+        /// </summary>
+        public new async Task<BigQueryDataReader> ExecuteReaderAsync(CommandBehavior behavior,
```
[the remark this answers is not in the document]
benwulfe: yes poor line break choice.  when I fix these manually, they wont get auto-changed.  This happens if I let something past the 120 guideline and the autoformat decides whats best.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommandType.cs
```diff
@@ -0,0 +1,33 @@
...
+using System.Text;
+
+namespace Google.Cloud.BigQuery.V2
+{
+    /// <summary>
+    /// The type of BigQuery command to execute.
+    /// </summary>
+    public enum BigQueryCommandType
+    {
+        /// <summary>
```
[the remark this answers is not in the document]
benwulfe: thanks

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
benwulfe: hmm we have a choice of returning true or false before we actually know for sure what the value should be (when ReadAsync is called).  prefilling the state adds unnecessary complication and possible perf side effects -- so I don't want to do that.
I agree, returning "true" here is probably better than "false" -- but developers will need to handle the case of us returning True for HasRows and then immediately returning false for ReadAsync.  Not entirely clear, but I suppose better than the alternative.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
benwulfe: yes...?  culture problems again.  I was hoping to get away with just using the framework to do the conversion -- whose behavior is easily documented (we call "ChangeType").

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
benwulfe: I dont *think* so.  It seems to be rarely used externally, but I remember seeing places where its used internally.  Is there a concern with re-enumerating if it does that?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
benwulfe: But HasRows uses the existence of Current for its value.  So it might make code a little cleaner, but wouldnt help with predetermining HasRows.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDbType.cs
benwulfe: Considered that, but ultimately, it felt redundant (the declaration would be on order size top the switch) and not really adding anything perf-wise.

jskeet:
(Will go through and respond to comments tomorrow.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameter.Data.cs
benwulfe: it's not 1-1 though.  timestamp maps to datetime, but datetime maps back to  (BQType.)datetime.
for the same reason as above and I felt the switch was more readable by displaying these elided cases better.
let me noodle on it though.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameterCollection.Data.cs
benwulfe: it should throw -- that's a code error on the part of the developer.  He should not attempt to add or remove any type that isnt a BQParameter.  Ignoring the issue on Remove might create false expectations on Add -- where we cannot ignore the error.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetQueryResultsOptions.cs
benwulfe: this is used to snap the options on the current thread before doing any awaits which could cause a context switch.
Checkout BigQueryConnection.Queries.cs which has a SqlCommandOptions that uses this.
It's eventually to ensure that if a developer calls something like a WaitAll on a linq Select which returns tasks created by ExecuteReader/NonQuery that there arent threading issues.  Code like this:
  Task.WaitAll(  parameters.Select(x => {  
                    bqCommand.Parameter[name]=x; 
                    return bqCommand.ExecuteNonQuery();
                    });
  the options on the command need to be snapped before the first context switch (await) or else you get threading issues with the next cmd invocation.  shouldn't happen often, but I saw it in one of the spanner samples, so I'm just making sure there is consistency here.
Of course, I messed up cmd parameter usage (as in, it isnt working) -- so I need to update the PR again.  fix inc...

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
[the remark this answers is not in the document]
benwulfe: i thought we agreed on 120 chars?  I've tried to make it more readable with a better break, but there is a balance here so that it can fit properly in some scenarios.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
benwulfe: I've fixed this up in the latest commit.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameterCollection.Data.cs
benwulfe: no because the method takes an Array type.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryCommand.Data.cs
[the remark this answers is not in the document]
jskeet: I use 120 as a guideline, but where things look significantly better by breaking that a bit, I'll generally do so.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/SqlCommandTest.cs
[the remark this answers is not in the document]
jskeet: I tend to think of GetString etc as *preferable* to the generic methods, given that they're not *really* fully generic (they only accept a certain set of types, which isn't specified via a constraint in the type system).

But the lack of the overloads makes me very confused - you're right, they're not there, and that's puzzling in terms of my memory. Oh well.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/SqlCommandTest.cs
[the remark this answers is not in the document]
jskeet: Sorry, it was more about whether we really need to call "open" at all than about the connection string. There's nothing to open or close.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/SqlCommandTest.cs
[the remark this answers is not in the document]
jskeet: Would rather not go into ExpandoObject if possible - if you're reasonably happy with the dictionary approach, that's fine by me.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDbType.cs
jskeet: Sure. No big deal, and it wouldn't be an API change if we decided to do it.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetQueryResultsOptions.cs
jskeet: Ah, I haven't reviewed the generated code yet - that explains it. Will come back to looking at options.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryConnection.Queries.cs
```diff
@@ -0,0 +1,247 @@
...
+            => GetBigQueryClient().ListRowsAsync(tableReference, schema, options);
+        #endregion
+    }
+
+    /// <summary>
+    /// Command options for configuring a sql query or DML statement.
+    /// You may create a new <see cref="BigQueryCommand" /> and set <see cref="BigQueryCommand.CommandOptions" />
+    /// to an instance of this class to perform a SQL query or DML statement.
+    /// </summary>
+    public sealed class SqlCommandOptions : BigQueryCommandOptions
```
jskeet: Could this be moved to a separate file? This was why I missed it before :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/SqlCommandTest.cs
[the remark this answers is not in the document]
benwulfe: True, I was thinking about it in terms of allocating resources for a BQClient and if that had any underlying persistent connection.  If it doesn't maintain a connection, then open can be a noop and we can create client up front.  then open becomes completely optional.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryConnection.Queries.cs
benwulfe: yes thats reasonable.  my initial intention was that each file would contain the related functionality whether autogenerated or not -- so you wouldnt have to know up front where to look -- but the class can certainly be moved.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryDataReader.cs
```diff
@@ -14,197 +14,209 @@
...
+            => await MoveNextAsync(cancellationToken).ConfigureAwait(false);
+
         /// <inheritdoc />
-        public override bool IsClosed { get; }
+        public override int Depth => 0;
+
+        //HasRows is true before any MoveNext and thereafter as long as Current is not null.
         /// <inheritdoc />
-        public override int RecordsAffected { get; }
+        public override bool HasRows => !_isEnumerating || _results.Current != null;
```
jskeet: I've been thinking about this more, and it feels like we should return the right answer, even if that means making an RPC. That's ugly in terms of maybe needing to buffer the results of a MoveNext call, but I don't think returning the wrong answer is a good idea.

PR #1424, opened 2017-09-01
Added the ability to configure credentials using a JSON file

iamatree:
Added a property that allows configuration of credentials using a JSON file instead of using the default credentials. This is necessary to support GCP instances where the configured service account cannot be granted the logging.logWriter role.


jskeet:
Travis is generally flaky at the moment - I'm happy to restart builds etc.

That said, there's already the `GOOGLE_APPLICATION_CREDENTIALS` environment variable that can be used - I'm not sure whether we really want multiple want. Let's see what @chrisdunelm thinks.

chrisdunelm:
@iamatree Thanks for the PR. Yes, this functionality is a good idea. I'll review the code shortly.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -194,6 +201,38 @@ public override void ActivateOptions()
...
+            GoogleCredential credential;
+            if(!TryGetCredentialFromFile(out credential))
+            {
+                return LoggingServiceV2Client.Create();
+            }
+            
+            return LoggingServiceV2Client.Create(new Grpc.Core.Channel(LoggingServiceV2Client.DefaultEndpoint.Host, LoggingServiceV2Client.DefaultEndpoint.Port, Grpc.Auth.GoogleGrpcCredentials.ToChannelCredentials(credential)));
+        }
+
+        private bool TryGetCredentialFromFile(out GoogleCredential credential)
```
chrisdunelm: Please just make this method return `GoogleCredential`, or `null`. We don't use the `Try...(out)` pattern particularly.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -194,6 +201,38 @@ public override void ActivateOptions()
...
 
+        private LoggingServiceV2Client BuildLoggingServiceClient()
+        {
+            GoogleCredential credential;
+            if(!TryGetCredentialFromFile(out credential))
+            {
+                return LoggingServiceV2Client.Create();
+            }
+            
+            return LoggingServiceV2Client.Create(new Grpc.Core.Channel(LoggingServiceV2Client.DefaultEndpoint.Host, LoggingServiceV2Client.DefaultEndpoint.Port, Grpc.Auth.GoogleGrpcCredentials.ToChannelCredentials(credential)));
```
chrisdunelm: Please make the `.ToChannelCredentials()` call an extension method cal on `credential`. This will also remove the fully qualified `ToChannelCredentials` name here which will be better.

Please don't use the fully qualified `Grpc.Core.Channel` name; please have a suitable `using` statement.

And split this line if it's too long; ~120 characters, we don't have a hard limit, but it looks longer than all the lines around it.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -194,6 +201,38 @@ public override void ActivateOptions()
...
+        {
+            credential = null;
+            if(string.IsNullOrWhiteSpace(CredentialFile))
+            {
+                return false;
+            }
+            
+            using (FileStream credStream = File.OpenRead(CredentialFile))
+            {
+                credential = GoogleCredential.FromStream(credStream);
```
chrisdunelm: Please use the new `GoogleCredential.FromFile()` method; available in Google.Apis.Auth from v1.29

jskeet: It may be worth holding off for a week until we've rolled out the use of 1.29.1 - at the moment this still depends on 1.28.

iamatree: I've made the other changes. I'll watch for the 1.29.1 roll out so I can switch to the FromFile method and push an update.

chrisdunelm: Great, thanks. Apologies for not realising that this code hasn't yet upgraded to v1.29!

iamatree: An update on this. I noticed that 1.29.1 is out, however it looks like I have to wait until the Google.Api.Gax.Grpc package is updated to use it since I believe that's where the Google.Api.Auth reference is coming from and even the 2.1.0-beta02 version of that package still uses 1.28. It's not impossible that this could be worked around by applying the Google.Api.Auth package directly but I'd prefer not to get into NuGet package fiddling if I can avoid it. It's far to easy to inadvertently break something.

jskeet:
While we *could* add a direct dependency on 1.29.1, I'd rather wait until we've updated GAX, if you're okay to hold on with this PR for just a bit longer. There's a long chain of releases that will happen soon... Shouldn't be much longer.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
jskeet: The dependencies are now updated - if you rebase to master you should be able to use it without any change in dependencies.

iamatree: If I open something like the Google.Cloud.Storage.V1 I can use it from there but Google.Cloud.Logging.Log4Net is still using the old version of Google.Apis.Auth.  I think that comes down to Google.Cloud.Logging.V2 still referencing Google.Api.Gax.Grpc 2.0.0 instead of the new Google.Api.Gax.Grpc 2.1.0.

jskeet:
Ah, yes. Will update this to be 2.1.0-alpha00 and update the dependency. Will be in the UK morning though.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
iamatree: Okay, all the changes should be in now. Let me know if you spot anything else that should change.

Issue #1415, opened 2017-08-30
Lots of stalled Travis builds


jskeet:
Several times in the last couple of days, we've had Travis builds which have died after receiving no input for 10 minutes. I'm not sure why yet, but we should look into it - reliable CI is a must-have.

jskeet:
It's getting silly. Upping to P0, and looking into this immediately.

jskeet:
(Looks like the changes in #1432 have worked.)

PR #1432, opened 2017-09-06
Update build procedure to help Travis

jskeet:
Instead of building everything in one giant solution, build one API
at a time. This is a bit less efficient, but it means we don't have
one huge build that stalls on Travis for over 10 minutes.

The Speech solution file change is to remove an integration test
project which doesn't actually exist.


jskeet:
Hopefully fixes #1415. We'll see what it does to the build times...

jskeet:
Humbug - broken on Travis already! Will ditch the building of tools automatically...

jskeet:
This approach needs some work due to projects which only target desktop .NET. Hmm.

jskeet:
Looks like this does have a noticeable effect on build time, but if it means we don't have to do as many of them, it's probably still a win. Merging, and we can revisit later.

PR #1434, opened 2017-09-06
Improve ListTables support

### LINKED ISSUE #1428: Can't access the table resource
I need to update the metadata on a lot of tables, so I thought I would write a small F# script to do it. To figure out if a table need to be updated I need to look at the table resource data, but it seems like the resource is not properly popluated. Short example of what I'm doing:

```
#load ".paket/load/net45/Google.Cloud.BigQuery.V2.fsx"

open Google.Cloud.BigQuery.V2

let client = BigQueryClient.Create("projectId")

let datasetsList = ["datasets"; "that"; "contains"; "tables"]

let datasets = 
    datasetsList 
    |> List.map (fun datasetName -> client.GetDataset(datasetName))

datasets 
|> List.map (fun (ds:BigQueryDataset) -> ds.ListTables() |> Seq.toList)
|> List.concat
|> List.map (fun table -> table.FullyQualifiedId, table.Resource.CreationTime, table.Resource.FriendlyName, table.Resource.CreationTime, table.Resource.Description, table.Resource.ETag, table.Resource.Id, table.Resource.Kind)
|> List.iter (printfn "%A")
```
I would expect this to print the all the data specified in the last `map`, but instead only the `FullyQualifiedId` is printed, all the fields on the `Resource` property is printed as `null`.

To run the code above do the following:

* Follow instructions here: https://fsprojects.github.io/Paket/getting-started.html
* Add nuget: `mono .paket/paket.exe add Google.Cloud.BigQuery.V2 -V 1.0.0-beta17`
* Generate load script: `mono .paket/paket.exe generate-load-scripts --framework net45`
* Make sure the code above is in a file in the same folder as the `.paket` folder, so the `#load` directive can find the generated load script.

### Workaround

To get the actual resource data you need to use `dataset.GetTable`. I think this is a design that most likely will result in some issues in the long run.

jskeet:
- Add more documentation emphasizing that some properties will be
  missing
- Add more properties
- Move the conversion to the TablePageManager, which is the only
  thing that needs to know/care.

Fixes #1428 as far as we intend to.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/TablesTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/TablesTest.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using Google.Apis.Bigquery.v2.Data;
+using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
 using Xunit;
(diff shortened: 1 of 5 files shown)


jskeet:
I wondered about a test - it will be pretty trivial, but I guess it would be a good idea. Will add that now, and a brief integration test.

jskeet:
Now that we have tests, I'll merge on green.

PR #1438, opened 2017-09-07
Support for netstandard 2.0 (and thereby netcore 2.0)

benwulfe:
.Net Core 2.0 adds *all* of the missing data features.  So now net45 == netstandard20.  hoorah!

Also cleaned up some preprocessor directives that were redundant (eg don't need NET451 if its not declared as a target fx).

NOTE: Tests pass only if you erase Grpc.Core.dll and run dotnet test with --no-build due to https://github.com/xunit/xunit/issues/1363


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/Google.Cloud.Spanner.Data.IntegrationTests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.0;net452</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.0;net452;netcoreapp2.0</TargetFrameworks>
```
jskeet: This should be done in apis.json - I'm surprised AppVeyor hasn't complained that the project has changed. (Ditto for the main project.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DataAdapterTests.cs
```diff
@@ -35,7 +35,7 @@ public DataAdapterTests(TestDatabaseFixture testFixture, ITestOutputHelper outpu
             TestLogger.TestOutputHelper = outputHelper;
         }
 
-#if NET45 || NET451 || NET452
+#if NET452 || NETCOREAPP2_0
```
jskeet: Maybe just #if !NETCOREAPP1_0 if we're expecting future TFMs to support it?

(This applies to a whole bunch of things below.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/MySqlStressTests.cs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#if NET45 || NET451 || NET452
+#if NET452
```
jskeet: What about NETCOREAPP2_0 here?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.sln
```diff
@@ -20,7 +20,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Google.Cloud.Spanner.Data.T
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Google.Cloud.Spanner.Data.Snippets", "Google.Cloud.Spanner.Data.Snippets\Google.Cloud.Spanner.Data.Snippets.csproj", "{EBF9F4E0-C115-4DF0-BD06-2951A88A65AB}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Google.Cloud.ClientTesting", "..\..\tools\Google.Cloud.ClientTesting\Google.Cloud.ClientTesting.csproj", "{9FABF89B-6D3F-43BE-A9B5-C5D75ADBBB4D}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Google.Cloud.ClientTesting", "..\..\tools\Google.Cloud.ClientTesting\Google.Cloud.ClientTesting.csproj", "{9FABF89B-6D3F-43BE-A9B5-C5D75ADBBB4D}"
```
jskeet: Just in case you were wondering why this is happening, it's discussed in https://github.com/dotnet/project-system/issues/1821

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/MySqlStressTests.cs
benwulfe: This is mysql which isnt working for netcore 2.0

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.sln
benwulfe: cool thanks

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DataAdapterTests.cs
benwulfe: *seems* safe enough and more future proof.

jskeet:
TargetFrameworks issue might be due to the behaviour of StripDesktopOnNonWindows.xml - I only tried that with two TFMs...

benwulfe:
I was thinking that (which is why I tried reversing the frameworks), but it
looks like something is converting the resulting value directly to
"TargetFramework" -- but the strip target doesnt seem to be doing that.

jskeet:
No - it's definitely confusing that it's reporting it that way. I can try to have a look on Monday if you get too frustrated :)

benwulfe:
yes confirmed, it is strip.  I made strip run on windows and now the build
fails locally.
I'll see what I can do.

jskeet:
Does that conditional approach work with VS? I thought I'd tried something similar initially, before doing the stripping option. I may have done something slightly different though - or VS may have improved. Either way, if VS is happy with it, that's great :)

benwulfe:
Everything looks good so far in VS...  Of course, this is VS 2017 so its very likely fixed if you saw it in VS 2015.

jskeet:
It wouldn't have been in 2015 as this project format didn't exist then. But we're on our third major update of VS2017 anyway, so...

Let's go for it.

PR #1441, opened 2017-09-11
Implement Datastore transaction options in DatastoreDb.

jskeet:
These have been available in the generated code (so in
BeginTransactionRequest) for a while, but not exposed in DatastoreDb.

Unfortunately we can't just overload BeginTransaction, as that would
change the meaning of db.BeginTransaction(null).

(I haven't implemented this in terms of delegation, but could do. We'd need to be very careful about what delegated to what, given the existing release. I think it's simpler not to.)


jskeet:
Hold on with this at the moment - there may be better options. The TransactionOptions message is a bit awkward to work with.

jskeet:
Definitely want to revisit this. Closing for now.

PR #1445, opened 2017-09-12
Adds ability to specify a json credential file in the connection string of spanner.

benwulfe:
via CredentialFile=<json_file>
Can be a full path or relative to the application folder retrieved via AppDomain.CurrentDomain.BaseDirectory (AppContext for netcore < 2).

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/ConnectionStringBuilderTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/ConnectionStringBuilderTests.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using System;
+using System.IO;
 using Google.Cloud.Spanner.V1;
 using Xunit;
 
@@ -116,6 +117,51 @@ namespace Google.Cloud.Spanner.Data.Tests
             Assert.Equal("project1", connectionStringBuilder.Project);
             Assert.Equal("instance1", connectionStringBuilder.SpannerInstance);
             Assert.Null(connectionStringBuilder.SpannerDatabase);
+        }
+
+        [Fact]
+        public void CredentialFile()
+        {
+            // ReSharper disable once JoinDeclarationAndInitializer
+            string appFolder;
+#if NETCOREAPP1_0
+            appFolder = AppContext.BaseDirectory;
+#else
+            appFolder = AppDomain.CurrentDomain.BaseDirectory;
+#endif
+            string jsonFile = $"{appFolder}{Path.DirectorySeparatorChar}SpannerEF-8dfc036f6000.json";
+            Assert.True(File.Exists(jsonFile));
+            using (var connection = new SpannerConnection($"CredentialFile={jsonFile}"))
+            {
+                Assert.NotNull(connection.GetCredential());
+            }
+        }
+
+        [Fact]
+        public void CredentialFileRelative()
+        {
+            using (var connection = new SpannerConnection("CredentialFile=SpannerEF-8dfc036f6000.json"))
+            {
+                Assert.NotNull(connection.GetCredential());
+            }
+        }
+
+        [Fact]
+        public void CredentialFileP12Error()
+        {
+            using (var connection = new SpannerConnection("CredentialFile=SpannerEF-8dfc036f6000.p12"))
+            {
+                Assert.Throws<InvalidOperationException>(() => connection.GetCredential());
+            }
+        }
+
+        [Fact]
+        public void CredentialFileNotFound()
+        {
+            using (var connection = new SpannerConnection("CredentialFile=..\\BadFilePath.json"))
+            {
+                Assert.Throws<FileNotFoundException>(() => connection.GetCredential());
+            }
         }
     }
 }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
@@ -14,6 +14,7 @@
 
 using System;
 using System.Data.Common;
+using System.IO;
 using Google.Api.Gax.Grpc;
 using Google.Apis.Auth.OAuth2;
 using Google.Apis.Util;
@@ -31,6 +32,7 @@ namespace Google.Cloud.Spanner.Data
     /// </summary>
     public sealed class SpannerConnectionStringBuilder : DbConnectionStringBuilder
     {
+        private const string CredentialFileKeyword = "CredentialFile";
         private const string DataSourceKeyword = "Data Source";
         private InstanceName _instanceName;
         private DatabaseName _databaseName;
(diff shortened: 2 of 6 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/ConnectionStringBuilderTests.cs
```diff
@@ -117,5 +118,51 @@ public void InstanceOnlyDataSource()
...
+                Assert.Throws<ArgumentException>(() => connection.Credential);
+            }
+        }
+
+        [Fact]
+        public void CredentialFileNotFound()
+        {
+            using (var connection = new SpannerConnection("CredentialFile=..\\BadFilePath.json"))
+            {
+                Assert.Throws<FileNotFoundException>(() => connection.Credential);
```
evildour: The .NET framework [property design guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/property) suggest not doing this: "AVOID throwing exceptions from property getters.". Can the constructor throw or can this property be turned into a method?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/ConnectionStringBuilderTests.cs
```diff
@@ -117,5 +118,51 @@ public void InstanceOnlyDataSource()
...
+
+        [Fact]
+        public void CredentialFile()
+        {
+            // ReSharper disable once JoinDeclarationAndInitializer
+            string appFolder;
+#if NETCOREAPP1_0
+            appFolder = AppContext.BaseDirectory;
+#endif
+#if !NETCOREAPP1_0
```
evildour: What about #else instead of these two lines?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -197,6 +245,16 @@ private string GetValueOrDefault(string key, string defaultValue = "")
             return defaultValue;
         }
 
+        private string GetApplicationFolder()
+        {
+#if NETSTANDARD1_5
+            return AppContext.BaseDirectory;
+#endif
+#if !NETSTANDARD1_5
```
evildour: #else?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/ConnectionStringBuilderTests.cs
benwulfe: good point, I'll switch to a method.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -41,7 +43,57 @@ public sealed class SpannerConnectionStringBuilder : DbConnectionStringBuilder
...
+            {
+                return CredentialOverride;
+            }
+
+            //Calculate it from the CredentialFile argument in the connection string.
+            GoogleCredential credential = null;
+            string jsonFile = CredentialFile;
+            if (!string.IsNullOrEmpty(jsonFile))
+            {
+                if (string.Compare(Path.GetExtension(jsonFile), ".json", StringComparison.OrdinalIgnoreCase) != 0)
```
jskeet: string.Equals(..., ..., StringComparison.OrdinalIgnoreCase)?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -41,7 +43,57 @@ public sealed class SpannerConnectionStringBuilder : DbConnectionStringBuilder
...
+                {
+                    throw new ArgumentException($"{nameof(CredentialFile)} should only be set to a JSON file.");
```
jskeet: This shouldn't be an ArgumentException, as it's a parameterless method. InvalidOperationException?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -41,7 +43,57 @@ public sealed class SpannerConnectionStringBuilder : DbConnectionStringBuilder
...
+                }
+
+                if (!File.Exists(jsonFile))
```
jskeet: Is there precedence for this in other ADO.NET providers? For example, SqlCeConnection specifies a file as a data source - does that do this sort of thing? It's one of those things like lenient conversions - useful in some cases, but hard to predict in others. I'm not definitely against it - just slightly wary.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -41,7 +43,57 @@ public sealed class SpannerConnectionStringBuilder : DbConnectionStringBuilder
...
+                {
+                    //throw a meaningful error that tells the developer where we looked.
+                    throw new FileNotFoundException($"Could not find {nameof(CredentialFile)}. Also looked in {jsonFile}.");
+                }
+
+                //TODO(benwu) replace with GoogleCredential.FromFile
+                using (var credStream = File.OpenRead(jsonFile))
+                {
+                    credential = GoogleCredential.FromStream(credStream);
+                    if (credential.IsCreateScopedRequired)
```
jskeet: It's safe to unconditionally call CreateScoped - it'll just return the original credential if it doesn't need scoping. We almost certainly want to move to that pattern elsewhere.

cc @chrisdunelm to fact-check me on that.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -41,7 +43,57 @@ public sealed class SpannerConnectionStringBuilder : DbConnectionStringBuilder
...
+            // Check the ctor override.
+            if (CredentialOverride != null)
+            {
+                return CredentialOverride;
+            }
+
+            //Calculate it from the CredentialFile argument in the connection string.
+            GoogleCredential credential = null;
+            string jsonFile = CredentialFile;
+            if (!string.IsNullOrEmpty(jsonFile))
```
jskeet: Is there any reason we want to completely prohibit p12 files? I know json files are preferred these days, but if a user already has a p12 file, I'd expect it to work - is there something Spanner-specific here?

benwulfe: My read of GoogleCredential.FromStream was that it only supported json.  As p12 has been discouraged on pantheon, I didn't consider supporting it a high priority item.  Is there an equivalent easy way to support p12?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
benwulfe: sure

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
benwulfe: I did find an example of another provider supporting a file through a path -- it might have been MySql.CertificateFile which is not a credential, but an ssl pem file IIRC.  I would have liked to support a resource as I think EF creates its own "res:" uri syntax, but that's too much work and duplication of what they do.  I'm hoping once we move to EF that we can support that better.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
jskeet: I think you're right - I really thought we supported p12 files, but I can't find any sign of it now.

@chrisdunelm care to chime in?

jskeet: Just checked with Chris, and you're right, while we do support p12 it's via a different call. Let's not do it for now.

PR #1447, opened 2017-09-14
Implement Datastore transaction options in DatastoreDb.

jskeet:
These have been available in the generated code (so in
BeginTransactionRequest) for a while, but not exposed in DatastoreDb.

I don't propose to merge this until we can test it directly in whitelisted projects. (I've tested using a test environment.)

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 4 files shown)


jskeet:
I'll at least wait until we know what the timing is going to be like, yes.

jskeet:
In other news, Codacy's "'partial' is gratuitous in this context." claim is completely bogus. It doesn't compile without the partial modifier.

chrisdunelm:
I've already learned to ignore codacy.
Clearly you know good quality C# better than any machine :)

jskeet:
I've now renamed BeginTransactionWithOptions back to BeginTransaction having thought about it more - everything means the same as it did before :)

jskeet:
Hmm. Not sure why the codecov change doesn't seem to have had an impact here. Will have a look when I'm back from travels.

jskeet:
This is now available in production, so I'm merging on green (having rebased).

PR #1455, opened 2017-09-15
Remove project dependency on LongRunning and IAM to allow independent releases

benwulfe:
This removes the risk that these dependencies could be bumped with a breaking change internally without us releasing an update along with spanner.

### DIFF
--- a/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1.Snippets/coverage.xml
+++ b/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1.Snippets/coverage.xml
@@ -7,12 +7,6 @@
       <FilterEntry>
         <ModuleMask>Google.Cloud.Spanner.Admin.Database.V1</ModuleMask>
       </FilterEntry>
-      <FilterEntry>
-        <ModuleMask>Google.LongRunning</ModuleMask>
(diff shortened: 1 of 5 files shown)


jskeet:
In theory at least, there should be no risk now - you'd have been forced to do this by the release generator:

https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/tools/Google.Cloud.Tools.TagReleases/Program.cs#L143

evildour:
Were the package references added in manually? When I re-run the project generator after this change, they get removed. I thought the build process was supposed to pick up on that though and error out when that happens.

benwulfe:
Hmm could be my bad.  I'm checking.

benwulfe:
Hmm yes, this is my mistake.  I must have modified the projects after
generation and missed the travis issue.
I will fix it.

evildour:
Still seems odd though. The build should have done the diff with the current and autogenerated projects and errored out on these. @jskeet was this removed when the builds were changed to not use a single solution anymore?

benwulfe:
Looks like travis never ran and it's very possible since I was in the
middle of a release that was blocked on this that I didn't wait for the
travis queue to flush.  My bad.  I have a fix queued.

jskeet:
@evildour: It *should* still be part of the AppVeyor build. (It's not part of Travis.) I'll check that today, as I've a sneaking suspicion I've seen another build that passed when I wouldn't expect it to.

PR #1463, opened 2017-09-18
Fix build

jskeet:
(The problem was due to two PRs, one of which added a project and
one of which changed project generation, being inflight
simultaneously.)


jskeet:
The good news is that AppVeyor failing proves that the project generation check is working :)

jskeet:
Self-merging as the build does seem to be fixed by it...

evildour:
Oops. Sorry I should has rebased and regenerated after merging the analyzer.

PR #1465, opened 2017-09-18
Add references to existing snippets from spannerconnection members.  …

benwulfe:
…This ensures most of the non-ADO.NET methods have snippets.

Moved the overviews to SpannerConnection snippets so they could more easily be referenced via 'Additional'


benwulfe:
oddly on my commit before sending to origin, this had a much cleaner diff as it saw it as a rename.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -36,5 +58,251 @@ public void CreateConnection()
...
+                var createDbCmd = connection.CreateDdlCommand($"CREATE DATABASE {_databaseName}");
+                await createDbCmd.ExecuteNonQueryAsync();
+
+                var createTableCmd = connection.CreateDdlCommand(
+                    @"CREATE TABLE TestTable (
+                                            Key                STRING(MAX) NOT NULL,
+                                            StringValue        STRING(MAX),
+                                            Int64Value         INT64,
+                                          ) PRIMARY KEY (Key)");
+                await createTableCmd.ExecuteNonQueryAsync();
```
jskeet: What's the result of this for CREATE TABLE? Worth printing to the console? (Ditto for CREATE DATABASE.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -36,5 +58,251 @@ public void CreateConnection()
...
+        public async Task CreateDatabaseAsync()
+        {
+            // CreateDatabaseAsync will create a new database and dispose of it right after.
+            // But for consistency sake, we want to show the same variables in the snippets.
+            _databaseName = "t_" + Guid.NewGuid().ToString("N").Substring(0, 28);
+
+            // Sample: CreateDatabaseAsync
+            // Additional: CreateDdlCommand
+            using (var connection =
+                new SpannerConnection($"Data Source=projects/{_projectId}/instances/{_instanceName}/databases/{_databaseName}"))
```
jskeet: Possibly worth having local variables declared outside the snippet so we can use names without underscores?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -36,5 +58,251 @@ public void CreateConnection()
...
+                        {"Int64Value", SpannerDbType.Int64}
+                    });
+
+                // This executes 5 distinct transactions with one row written per transaction.
+                for (var i = 0; i < 5; i++)
+                {
+                    cmd.Parameters["Key"].Value = Guid.NewGuid().ToString("N");
+                    cmd.Parameters["StringValue"].Value = $"StringValue{i}";
+                    cmd.Parameters["Int64Value"].Value = i;
+                    await cmd.ExecuteNonQueryAsync();
```
jskeet: Print out this result, and maybe comment that it's 1?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -36,5 +58,251 @@ public void CreateConnection()
...
+                $"Data Source=projects/{_projectId}/instances/{_instanceName}/databases/{_databaseName}"))
+            {
+                await connection.OpenAsync();
+
+                // Read the first two keys in the database.
+                var keys = new List<string>();
+                var selectCmd = connection.CreateSelectCommand("SELECT * FROM TestTable");
+                using (var reader = await selectCmd.ExecuteReaderAsync())
+                {
+                    while (await reader.ReadAsync())
```
jskeet: How about

```csharp
while (keys.Count < 3 && reader.ReadAsync())
{
    keys.Add(reader.GetFieldValue<string>("Key"));
}
```
?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -36,5 +58,251 @@ public void CreateConnection()
...
+                            cmd.Transaction = transaction;
+
+                            // This executes a single transactions with alls row written at once during CommitAsync().
+                            // If a transient fault occurs, this entire method is re-run.
+                            for (var i = 0; i < 5; i++)
+                            {
+                                cmd.Parameters["Key"].Value = Guid.NewGuid().ToString("N");
+                                cmd.Parameters["StringValue"].Value = $"StringValue{i}";
+                                cmd.Parameters["Int64Value"].Value = i;
+                                await cmd.ExecuteNonQueryAsync();
```
jskeet: What does this return when it's not committing at the same time? Always 1, on the grounds that it assumes it *will* insert later?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -36,5 +58,251 @@ public void CreateConnection()
...
+                            }
+
+                            scope.Complete();
+                            await Task.Yield();
+                        }
+                    }
+                });
+            // End sample
+            //TODO(benwu): Remove Task.Yield once grpc# fixes threadpool issue.
```
jskeet: We're using Grpc.Core 1.6.1 now, so if we're thinking about the same thread pool issue, you may be able to remove it now.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
benwulfe: it always should return number of rows affected even if it doesn't commit at that time.
Seeing how its built into ADO.NET, I don't feel its super important to highlight -- as its also very reasonable to ignore the result and rely on an exception not being thrown.

jskeet: Sure, not a problem.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
benwulfe: You mean the return value of the execute?  Its ignorable.  you could query information_schema to get the table info afterwards, but this is meant to be a super simple example.  Maybe I'll add that later as another snippet.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
benwulfe: seems reasonable here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
jskeet: I was more genuinely intrigued as to what the result would be, but I don't think it's important for a snippet :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
benwulfe: the names are meant to be unresolved when you paste in the snippet.  I don't see the underscore as necessarily bad as we cannot know whether the user will want fields or locals.  It's really just a placeholder for the developer to know he needs to fix (so it generates a compile error on purpose).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
benwulfe: oh I think we return 0 here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
jskeet: Yup, I wasn't suggesting trying to make the snippet compile - I just wonder whether the underscore might be a distraction. Not a biggie though.

PR #1481, opened 2017-09-21
Downgrade Microsoft.AspNetCore.* dependencies to 1.0

### LINKED ISSUE #1449: Can't add reference to Google.Cloud.Diagnostics.AspNetCore in VS2017
- Create an ASP.NET Core **1.0** application
- In "Manage NuGet packages" try to install Google.Cloud.Diagnostics.AspNetCore

Error:

```text
Restoring packages for C:\Users\skeet\Test\Projects\Issues\Issue1448\Issue1448\Issue1448.csproj...
Detected package downgrade: Microsoft.Extensions.Logging.Debug from 1.1.2 to 1.0.2. Reference the package directly from the project to select a different version. 
 Issue1448 (>= 1.0.0) -> Google.Cloud.Diagnostics.AspNetCore (>= 1.0.0) -> Microsoft.Extensions.Logging.Debug (>= 1.1.2) 
 Issue1448 (>= 1.0.0) -> Microsoft.Extensions.Logging.Debug (>= 1.0.2)
Package restore failed. Rolling back package changes for 'Issue1448'.
Time Elapsed: 00:00:01.9913824
```

VS2017 version 15.3.4

The problem is that the default csproj file contains:

```
<PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="1.0.2" />
```

... whereas we depend on 1.1.2.

Options for user:

- Update the dependency to 1.1.2 or higher
- Remove the explicit dependency
- Use ASP.NET Core 1.1 from the start

Given that ASP.NET Core 2.0 is now the default, maybe this isn't an issue... but I thought it worth recording, as we may want to document it.

iantalarico:
This will ensure netcore 1.0 apps can also use the libraries.

This fixes #1449.


jskeet:
1.0.2 rather than 1.0.3, I think. (When it's green, I'm happy.)

PR #1482, opened 2017-09-22
Set Monitoring API version to 1.0.0

jskeet:


jskeet:
(Will self-merge on green.)

Issue #1485, opened 2017-09-22
API docs for REST clients are incomplete


jskeet:
Since upgrading to docfx 2.24 _and_ updating the dependencies, we're not seeing all the relevant docs for the REST-based APIs, e.g. the base class (BaseClientService) is missing in

https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.BigQuery.V2/api/Google.Apis.Bigquery.v2.BigqueryService.html

I'll investigate. (I'm pushing the newly generated docs anyway, as I suspect the difference is REST-specific and there are lots of new docs that should be out.)

jskeet:
Rebuilding with the previous commit for the "dependencies" branch shows the right thing. Still not sure whether the problem is with docfx 2.24 or the way I've rebuilt the dependencies though. Investigating further.

jskeet:
Fixed by #1488. Rebuilding docs now.

PR #1488, opened 2017-09-25
Fix docs dependencies

jskeet:
Looks like we still need to restore before running docfx even on
.NET Core 2.0.


jskeet:
(This fixes #1485. I've pushed the rebuilt dependencies branch already; I'm now rebuilding all the docs to repush those.)

docs/fetchdependencies.sh
```diff
@@ -39,6 +40,11 @@ cp dependencies-docfx/docfx-google-api-dotnet-client.json dependencies/google-ap
 
 (cd dependencies/google-api-dotnet-client;
  rm NuGet.config;
+ dotnet restore Src/Support/GoogleApisClient.sln;
```
chrisdunelm: (Nothing to do now, but this will break once GoogleApisClient supports UWP)

jskeet: Righto. Will work out what we need to do at that point.

PR #1490, opened 2017-09-25
Ensure Google.Cloud.Diagnostics.AspNetCore works with net461

### LINKED ISSUE #1448: MissingMethodException when using ASP.NET Core in .NET framework
- Create new ASP.NET Core 2.0 project targeting .NET 4.6.1
- Add NuGet reference to Google.Cloud.Diagnostics.AspNetCore 1.0.0 (see #1449 for details...)
- In Startup.cs:
  - Add a using directive: `using Google.Cloud.Diagnostics.AspNetCore;`
  - In Configure, add `app.UseGoogleTrace();`
  - In ConfigureServices, add `services.AddGoogleTrace(options => options.ProjectId = "your-project-id-here");`
- In HomeController.cs
  - Add a constructor: `public HomeController(TraceHeaderContext context) {}` and add the corresponding using directive (`using Google.Cloud.Diagnostics.Common;`)
- Run...

Exception:

```text
MissingMethodException: Method not found: 'System.Nullable`1<Boolean>
Google.Cloud.Diagnostics.Common.TraceDecisionPredicate.ShouldTrace(Microsoft.AspNetCore.Http.HttpRequest)'.
```

The problem is that the netstandard build of Google.Cloud.Diagnostics.Common.TraceDecisionPredicate uses Microsoft.AspNetCore.Http.HttpRequest where the net45 build uses System.Web.HttpRequest. When an ASP.NET Core application runs on .NET 4.6.1, those two collide.

I'm not at all sure what we're going to do about this...

iantalarico:
Ensure that Google.Cloud.Diagnostics.AspNetCore will work with net461.

- Remove mixed deps on HttpRequest objects
- Run tests on net461

This fixes #1448


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.0;net461</TargetFrameworks>
```
benwulfe: Is targeting .net standard 2.0 not a possibility?

iantalarico: It is but this library will work with both in the 1.0 state.  For now this will simplify things.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.csproj
```diff
@@ -2,7 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <Version>1.0.0</Version>
-    <TargetFrameworks>netstandard1.5</TargetFrameworks>
+    <TargetFrameworks>netstandard1.5;net461</TargetFrameworks>
```
jskeet: Any reason for using 461 here rather than net45 as we do for everything else?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -171,7 +171,7 @@ public T RunInSpan<T>(Func<T> func, string name, StartSpanOptions options = null
                 catch (Exception e) when (SetStackTraceAndReturnFalse(e))
                 {
                     // This will never return as the condition above will always be false.
-                    return default;
+                    return default(T);
```
jskeet: Any reason for this change? Just `return default` should be fine, assuming you're using an up-to-date version of VS2017 (update 3).

apis/apis.json
```diff
@@ -138,7 +138,7 @@
     "id": "Google.Cloud.Diagnostics.AspNetCore",
     "version": "1.0.0",
     "type": "other",
-    "targetFrameworks": "netstandard1.5",
+    "targetFrameworks": "netstandard1.5;net461",
```
jskeet: As before - net45?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.csproj
iantalarico: The minimum library we can target is 451 based on the aspnetcore libs we depend on and AsyncLocalis not available in 451.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
iantalarico: upgraded and fixed :)

iantalarico:
The tests did fail before I fixed this.

The integration tests will run netcore and desktop.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.csproj
jskeet: Okay. It looks like net46 should be okay - maybe that would be better? Might as well go as low as we can, to support as many users as possible.

iantalarico:
Changed to net46.  I'll merge when green.

Thanks Jon!

PR #1493, opened 2017-09-27
Move to a simpler query approach

jskeet:
Overall intention: become less ADO-like, to avoid confusion with a
later dedicated ADO package. Currently we're in a bit of an uncanny
valley: it's deliberately ADO-like, but without actually supporting
ADO. Even without the prospect of an ADO package that has its
disadvantages, and it's more clearly "wrong" when there's an actual
ADO package.

BigQueryCommand and BigQueryParameterCollection become obsolete, as
a collection of BigQueryParameter values are provided directly on
calls to BigQueryClient. The old calls (taking either just a string
or a BigQueryCommand) are now obsolete. I've left them here for now
so that it's not a completely breaking change, and they have some
time to migrate.

The new tests are copies of the legacy tests, just using the new
overloads. All snippets have been migrated.

The "parameters" parameter is deliberately required, as I expect it
to be used in most cases, and we want to encourage query
parameterization. Additionally, if it were optional, a call to
ExecuteQuery("sql here") would be ambiguous between the old and new
overloads.

A call to ExecuteQuery("sql here", null) will now call the new
overload instead of the old one, but the effect will be the same -
no parameters, default options.


jskeet:
(There's relatively little genuinely new code here, btw.)

cc @jsimonweb as samples will need updating after the next release (I'm happy to do this)
cc @benwulfe for the ADO side
cc @juliuscesarnl @ssgtcookie as interested parties

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClient.Queries.cs
```diff
@@ -28,6 +29,23 @@ public abstract partial class BigQueryClient
         /// </summary>
         /// <remarks>
         /// <para>
+        /// This method will only return when the query has completed. It simply delegates to <see cref="CreateQueryJobAsync(string, IEnumerable{BigQueryParameter}, QueryOptions, CancellationToken)"/>
```
evildour: "CreateQueryJobAsync" = "CreateQueryJob"

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameter.cs
```diff
@@ -212,12 +212,20 @@ public BigQueryParameter(string name, BigQueryDbType? type, object value)
             Value = value;
         }
 
+        /// <summary>
```
chrisdunelm: I vaguely recall that this class was going to be renamed to "QueryParameter"?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/QueryOptions.cs
```diff
@@ -100,6 +100,15 @@ public sealed class QueryOptions : JobCreationOptions
         /// </summary>
         public bool? DryRun { get; set; }
 
+        /// <summary>
+        /// If set, determines whether parameters are named or positional. If not set,
+        /// this will default to named parameters. When named parameters are used, all parameters
+        /// must have a name. When positional parameters are used, parameters may still have names,
+        /// but don't have to.
+        /// </summary>
+        /// <remarks>This property is not used by the legacy query overloads using <see cref="BigQueryCommand"/>.</remarks>
```
chrisdunelm: Put in a TODO to remove this remark?

jskeet: No need - I've deliberately made it refer to BigQueryCommand so that we'll get an error when we delete that, at which point I'll remove the remark.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameter.cs
jskeet: I was going to, yes - but there's a class called QueryParameter in the REST API as well, so it would clash with that. I figured it was cleaner to leave it as-is.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClient.Queries.cs
jskeet: Fixed, thanks.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryParameter.cs
Jonathan34: `BigQueryParameter` is cleaner imho

PR #1494, opened 2017-09-27
NOT FOR SUBMISSION: analyzer testing

jskeet:
Deliberately try to provoke the analyzer


evildour:
It's odd that it complained about the copyright notice multiple times. I'm not sure what's going on there, but I can look into it. It is expected that the missing default arg is not flagged because there is no local or parameter available at the call site of type string and with the name 'x'.

jskeet:
Yup, worked that bit out with more experimentation. Don't worry about the multiple errors - it may well be one per target framework. Closing this now as we've worked out the issue.

PR #1497, opened 2017-09-27
Move all snippets to one project that targets netcore 2.0

iantalarico:


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/Google.Cloud.Diagnostics.AspNetCore.Snippets.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp1.0</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp1.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp2.0</TargetFrameworks>
```
jskeet: You can't just change the project file - you need to change apis.json, or possibly the project generator. I suspect you just want to set TestTargetFrameworks, although we also want to run the integration tests on net461 don't we?

iantalarico: Blah, okay I now remember why I split the snippets.

We want the main library to target 1.0 so it can be consumed by multiple versions.

The problem with the api.json is it sets all deps to a given version.  We need the version in the majority of the solution to be 1.0 while the version for one snippet to target 2.0.

jskeet: So set "testTargetFrameworks" to "netcoreapp1.0;netcoreapp2.0;net461" and use `#if` to comment out the core2.0-only parts. I'd expect that to work.

If you need the snippets to be different to the rest of the test projects, we'll need to modify the project generator a bit.

jskeet: Or if you mean the dependencies, you can set "testDependencies" separately to the main dependencies.

iantalarico:
@jskeet Thanks for the help here. PTAL

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests.csproj
```diff
@@ -33,4 +30,19 @@
     <Reference Condition="'$(TargetFramework)' == 'net452'" Include="Microsoft.CSharp" />
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp1.0'">
```
jskeet: This could be in one condition with the next one if we wanted, but it's fine to do it like this for more flexibility.

iantalarico:
Thanks Jon!

PR #1498, opened 2017-09-28
Rename BigQueryParameter.ArrayType to ArrayElementType

### LINKED ISSUE #1403: BigQueryParameter.Type + BigQueryParameter.ArrayType names
I'm not changing right now, but logging this for tracking.

These are potentially confusing.
Better names for each would be:

BigQueryParameter.Type -> BigQueryParameter.BigQueryDbType  (propertyname == typename as done in ADO.NET)

BigQueryParameter.ArrayType -> BigQueryParameter.ArrayElementType  (as done in reflection).

jskeet:
This is consistent with reflection and clearer.

I haven't renamed Type to BigQueryDbType as suggested in #1403, as
we're no longer trying to make this library feel like ADO.NET - and
Type is a considerably shorter name than BigQueryDbType with no loss
of clarity IMO.

Fixes #1403.


benwulfe:
LGTM

Issue #1499, opened 2017-09-28
Add netcoreapp2.0 testing to Google.Cloud.Diagnostics.AspNetCore


iantalarico:
Currently testing against netcoreapp2.0 causes a failure.

We need to wait for xunit 2.3.0-rc1-build3780 to be released.

See https://github.com/xunit/xunit/issues/1363

jskeet:
We now use xUnit 2.3.0 (GA) by default, so it's worth trying this again now...

PR #1500, opened 2017-09-28
Remove netcoreapp2.0 tests until Xunit 2.3.0-rc1-build3780 is released.

iantalarico:
This is causing integration tests to fail.
 
#1499 for tracking re-adding the testing.


iantalarico:
Thanks Jon!

PR #1501, opened 2017-09-28
Switch from ITokenAccess to ChannelCredentials

benwulfe:
Preferred method will be for the user to use a json file specified in the connectionstring (if not using default app creds).  But we'll keep this mechanism open in case the user has any custom requirement for specifying the creds.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
@@ -46,7 +46,7 @@ namespace Google.Cloud.Spanner.Data.Snippets
         [Fact]
         public void CreateConnection()
         {
-            // Snippet: #ctor(string, ITokenAccess)
+            // Snippet: #ctor(string, ChannelCredentials)
             string connectionString = "Data Source=projects/my-project/instances/my-instance/databases/my-db";
             SpannerConnection connection = new SpannerConnection(connectionString);
             Console.WriteLine(connection.Project);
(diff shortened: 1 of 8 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -46,7 +46,7 @@ public SpannerConnectionSnippets(SnippetFixture fixture)
         [Fact]
         public void CreateConnection()
         {
-            // Snippet: #ctor(string, ITokenAccess)
+            // Snippet: #ctor(string, ChannelCredential)
```
jskeet: ChannelCredentials

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -215,7 +212,7 @@ public string SpannerInstance
         /// Credentials can be retrieved from a file or obtained interactively.
         /// See Google Cloud documentation for more information. May be null.
         /// </param>
-        public SpannerConnectionStringBuilder(string connectionString, ITokenAccess credential = null)
+        public SpannerConnectionStringBuilder(string connectionString, ChannelCredentials credential = null)
```
jskeet: Everywhere that the parameter is called credential we could *consider* calling it credentials instead, just to match the noun. I'm not overly bothered, but I thought I'd mention it as an option. (Likewise GetCredential would be GetCredentials.)

benwulfe: seems reasonable.  I'm with it.

PR #1506, opened 2017-10-02
NOT FOR SUBMISSION Bigtable helpers for the Data API

evildour:
Skeleton code, (most) triple slash docs, and snippets for the proposed Bigtable library additions for the Data API. I changed a couple of the names from the internal doc, but most of the style is still the same. Also, ModifyRows is now included as well as that will be required for the initial release.

This is not for submission, but is just to get a more in-depth review of the approach for the P0s.

cc @igorbernstein2, @sduskis, @garye, @Orbz


benwulfe:
was there some design doc or proposal for bigtable support that i missed Mike? Or is this to support your current work that requires bigtable?

benwulfe:
more detail: whenever I previously had discussed bigtable, we postponed it because it supports hbase for which there are existing libraries.  If this is custom veneer work, then I worry we might be creating more maintenance down the line.

jskeet:
Let's get together in a VC for this. My recollection was that we found that there HBase didn't have traction in .NET, so while it was an attractive proposition for the Java crowd, it didn't make as much sense for C#.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Mutations.cs
```diff
@@ -0,0 +1,138 @@
...
+
+namespace Google.Cloud.Bigtable.V2
+{
+    /// <summary>
+    /// Static class with convenience methods for creating various kinds of <see cref="Mutation"/> instances.
+    /// </summary>
+    public static class Mutations
+    {
+        /// <summary>
+        /// Creates a <see cref="MutateRowsRequest.Types.Entry"/> to use with MutateRows requests.
```
jskeet: So what is an Entry here? It may be worth repeating some of the autogenerated docs just for clarity.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,407 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Protobuf;
+using System;
+
+namespace Google.Cloud.Bigtable.V2
+{
+    /// <summary>
+    /// Static class with convenience methods for creating various kinds of <see cref="RowFilter"/> instances.
```
jskeet: Given the factory methods here, I'd move the ones from ColumnRange to here too.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,407 @@
...
+        /// escape sequence must be used if a true wildcard is desired. The `.`
+        /// character will not match the new line character `\n`, which may be
+        /// present in a binary qualifier.
+        /// </para>
+        /// </remarks>
+        /// <param name="regex">
+        /// The RE2 regex used to match column qualifiers.
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter ColumnQualifierRegex(ColumnQualifier regex)
```
jskeet: Overload to accept a string for the regex?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,407 @@
...
+
+        /// <summary>
+        /// Creates a new <see cref="RowFilter"/> instance which matches only cells from
+        /// columns within the given range.
+        /// </summary>
+        /// <param name="range">
+        /// The range of columns from which cells should be matched.
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter ColumnRange(ColumnRange range)
```
jskeet: Remove this, given the OpenClosed methods etc?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,407 @@
...
+        /// sequence must be used if a true wildcard is desired. The `.` character
+        /// will not match the new line character `\n`, which may be present in a
+        /// binary key.
+        /// </para>
+        /// </remarks>
+        /// <param name="regex">
+        /// The RE2 regex used to match row keys.
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter RowKeyRegex(RowKey regex)
```
jskeet: Again, string overload?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowKey.cs
```diff
@@ -0,0 +1,147 @@
...
+
+using Google.Protobuf;
+using System;
+
+namespace Google.Cloud.Bigtable.V2
+{
+    /// <summary>
+    /// The key of a row.
+    /// </summary>
+    public sealed class RowKey : IEquatable<RowKey>
```
jskeet: Are row keys ever compared directly? If so, it would be useful for this to implement `IComparable<RowKey>`

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowKey.cs
```diff
@@ -0,0 +1,147 @@
...
+        /// <param name="x"></param>
+        /// <param name="y"></param>
+        /// <returns></returns>
+        public static bool operator !=(RowKey x, RowKey y)
+        {
+            throw new NotImplementedException();
+        }
+
+
+        /// <summary>
```
jskeet: Possibly explicit operators going the other way?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,339 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Bigtable.V2.Snippets
+{
+    public class BigtableClientSnippets
```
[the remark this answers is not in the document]
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,339 @@
...
+            // Create client
+            BigtableClient bigtableClient = BigtableClient.Create();
+            // Initialize request argument(s)
+            TableName tableName = new TableName("[PROJECT]", "[INSTANCE]", "[TABLE]");
+            RowKey rowKey = "r1";
+            RowFilter predicateFilter = RowFilters.Chain(
+                RowFilters.FamilyNameRegex(@"cf\d"),
+                RowFilters.ColumnQualifierRegex(@"[a-z]\d"),
+                RowFilters.ValueRegex(@"test-value"));
+            IEnumerable<Mutation> trueMutations = new Mutation[]
```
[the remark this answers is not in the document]
evildour: Do you just mean for the snippet code, or do you mean the parameter names for the method should change as well? They are documented [here](https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#google.bigtable.v2.CheckAndMutateRowRequest) as true_mutations and false_mutations. Also, these are the parameter names for the auto-generated CheckAndMutateRow method. So while I think your names are a bit clearer, for consistency it might make sense to stick with these.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,339 @@
...
+                RowFilters.CellsPerRowLimit(1),
+                RowFilters.VersionRange(new BigtableVersion(1), new BigtableVersion(25)));
+            Mutation trueMutation1 = Mutations.DeleteFromFamily("cf1");
+            Mutation trueMutation2 = Mutations.SetCell("cf1", "c1", "test-value2", new BigtableVersion(2));
+            // Make the request
+            CheckAndMutateRowResponse response = bigtableClient.CheckAndMutateRow(
+                tableName,
+                rowKey,
+                predicateFilter,
+                trueMutation1, trueMutation2);
```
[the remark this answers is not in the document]
evildour: As discussed in the internal doc, this is the overload that omits the CallSettings parameter and instead lets you specify the mutations as a params array. I wasn't sure what level of verbosity was needed in the snippets to show the types being used, but perhaps the previous snippet should show the type names and this once should show the more concise usage with the arguments created in the call itself. WDYT?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,339 @@
...
+        public async Task MutateRows()
+        {
+            // Snippet: MutateRows(TableName,IEnumerable<MutateRowsRequest.Types.Entry>,CallSettings)
+            // Create client
+            BigtableClient bigtableClient = BigtableClient.Create();
+            // Initialize request argument(s)
+            TableName tableName = new TableName("[PROJECT]", "[INSTANCE]", "[TABLE]");
+            IEnumerable<MutateRowsRequest.Types.Entry> entries = new MutateRowsRequest.Types.Entry[]
+            {
+                Mutations.CreateEntry("r1",
```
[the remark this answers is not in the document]
evildour: Each `Entry` is an atomic, ordered set of mutations for a particular row. `Mutations.CreateEntry` takes the key of the row to act upon and then the mutations to use. Perhaps this should have a different name to make it clearer. `Mutations.RowMutationsEntry` perhaps?

Also, this doesn't need to live on `Mutations`. I just didn't want to add static methods to the nested `MutateRowsRequest.Types.Entry` class and `Entries` sounded too generic to follow the naming convention I've been using here. I'm open to suggestions for where else this could live.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,339 @@
...
+                    Mutations.DeleteFromFamily("cf1"),
+                    Mutations.DeleteFromColumn("cf2", "c1", new BigtableVersionRange(null, new BigtableVersion(DateTime.UtcNow)))),
```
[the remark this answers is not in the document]
evildour: Done, and added one for longs as well. Also added a TODO to get clarification on use cases.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,339 @@
...
+            // End snippet
+
+            // TODO: Verifications
+        }
+
+        public async Task ReadRows()
+        {
+            // Snippet: ReadRows(TableName,RowSet,RowFilter,long,CallSettings)
+            // Create client
+            BigtableClient bigtableClient = BigtableClient.Create();
```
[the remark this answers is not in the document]
evildour: Good point. I'll make this one simpler and add some additional comments.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,339 @@
...
+            long rowsLimit = 10;
+            // Make the request
+            BigtableClient.ReadRowsStream streamingResponse = bigtableClient.ReadRows(
+                tableName,
+                rows,
+                filter,
+                rowsLimit);
+            // Read streaming responses from server until complete
+            IAsyncEnumerator<Row> responseStream = streamingResponse.GetRowsAsync().GetEnumerator();
+            while (await responseStream.MoveNext())
```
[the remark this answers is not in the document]
evildour: Ok, I'll add that here. Also, now that I think of it, rather than adding a `Cancel()` to `ReadRowsStream`, it probably makes more sense for `GetRowsAsync()` to take an optional CancellationToken which it monitors for cancellations and then disposes the underlying gRPC call when is detects cancellation. I'm going to change things to do that, but let me know if you feel otherwise.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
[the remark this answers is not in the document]
sduskis: The simplest scenario is a single range ("a" is closed, "b" is open).  I'ad also add two filters in a chain: `RowFilters.FamilyNameRegex("A")` and a latest version filter `RowFilters.CellsPerColumnLimitFilter(1)`

That translates to get me the latest cells for:
- rows starting with the letter "a"
- the latest cells that have a Column Family of "A".

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersion.cs
```diff
@@ -0,0 +1,169 @@
...
+        }
+
+        /// <summary>
+        /// Compares the versions to another.
+        /// </summary>
+        /// <param name="other">The other version to compare this to.</param>
+        /// <returns>
+        /// A negative value if this is less than the other version, a positive value if this is greater, and zero if they are equal.
+        /// </returns>
+        public int CompareTo(BigtableVersion other) => Compare(this, other);
```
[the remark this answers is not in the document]
evildour: This is the `IComparable<BigtableVersion>` implementation. Should this implement `IComparer<BigtableVerision>`? I was under the impression that values don't usually do that (just checked with TimeSpan, DateTime, and Int32, and they do not). I will add `IComparable`.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersion.cs
```diff
@@ -0,0 +1,169 @@
...
+    /// <summary>
+    /// A version which uniquely identifies a cell within a column.
+    /// </summary>
+    public struct BigtableVersion : IComparable<BigtableVersion>, IEquatable<BigtableVersion>
+    {
+        /// <summary>
+        /// Creates a new <see cref="BigtableVersion"/> version from a value.
+        /// </summary>
+        /// <param name="value">
+        /// The version value non-negative version value, or -1 to initialize from the microseconds of DateTime.UtcNow.
```
sduskis: -1 should get a method like `ServerTime()`

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Cell.Helpers.cs
```diff
@@ -0,0 +1,26 @@
...
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Bigtable.V2
+{
+    public partial class Cell
+    {
+        /// <summary>
+        /// Gets the version of the
```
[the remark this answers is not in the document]
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Cell.Helpers.cs
```diff
@@ -0,0 +1,26 @@
...
+using System;
+
+namespace Google.Cloud.Bigtable.V2
+{
+    public partial class Cell
+    {
+        /// <summary>
+        /// Gets the version of the 
+        /// </summary>
+        public BigtableVersion Version { get => new BigtableVersion(TimestampMicros); }
```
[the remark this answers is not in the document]
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,407 @@
...
+        /// an Interleave to contain multiple ApplyLabelTransformer filters, as they
+        /// will be applied to separate copies of the input. This may be relaxed in
+        /// the future.
+        /// </para>
+        /// </remarks>
+        /// <param name="label">
+        /// The label to apply to all cells in the output row.
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter ApplyLabelTransformer(string label)
```
sduskis: There are some Filters that shouldn't be used by most users, specifically, Label and Sink.  They were used for an obscure HBase filter. I think that we should remove those filters from this helper class.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
sduskis: +1

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,407 @@
...
+        /// "A:B" and timestamp 2, because one copy passed through the all filter while
+        /// the other was passed through the label and sink. Note that one copy has label
+        /// "foo", while the other does not.
+        /// <para>
+        /// Cannot be used within the predicateFilter, trueFilter, or falseFilter of a
+        /// <see cref="Condition"/>.
+        /// </para>
+        /// </remarks>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter Sink()
```
sduskis: As per the previous comment, I'd leave out Sink() from the helper class.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ColumnQualifier.cs
```diff
@@ -0,0 +1,147 @@
...
+    /// <summary>
+    /// A column qualifier.
+    /// </summary>
+    public sealed class ColumnQualifier : IEquatable<ColumnQualifier>
+    {
+        /// <summary>
+        /// Creates a new <see cref="ColumnQualifier"/> instance from the UTF8 encoding of a string.
+        /// </summary>
+        /// <param name="qualifierAsUtf8">The qualifier whose UTF8 representation should be used to initialize the qualifier.</param>
+        public ColumnQualifier(string qualifierAsUtf8)
```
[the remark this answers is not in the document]
evildour: I figure since we have the implicit operators, those will be used more than either constructors or factory methods, so I'm not sure it matters too much which way we go. I don't feel too strongly about it. I guess one benefit to factory methods is we could then have methods to take a string as UTF8 and another as Base64. But then again, I don't want to duplicate the whole `ByteString` API. This is just meant as a handy wrapper with implicit operators so you can specify column qualifiers as strings if necessary. I'm leaning towards leaving as is.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,407 @@
...
+        {
+            throw new InvalidOperationException();
+        }
+
+        /// <summary>
+        /// Creates a new <see cref="RowFilter"/> instance which replaces each cell's
+        /// value with the empty string.
```
sduskis: I think that it would be useful to add that this filter is useful when the objective is to perform counts rather than get the values.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowRange.Helpers.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// Creates a <see cref="RowRange"/> instance with an inclusive lower bound and an exclusive upper bound.
+        /// </summary>
+        /// <param name="startKeyClosed">
+        /// The inclusive lower bound for the range. If null, it is interpreted as the empty string.
+        /// </param>
+        /// <param name="endKeyOpen">
+        /// The exclusive upper bound for the range. If null, it is interpreted as the infinite row key.
+        /// </param>
+        /// <returns>The created range.</returns>
+        public static RowRange ClosedOpen(RowKey startKeyClosed, RowKey endKeyOpen)
```
sduskis: ClosedOpen is the most common use of Row Range.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ColumnRange.Helpers.cs
```diff
@@ -0,0 +1,103 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Google.Cloud.Bigtable.V2
+{
+    public partial class ColumnRange
+    {
+
+        /// <summary>
```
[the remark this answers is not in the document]
evildour: Yes, see [here](https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#columnrange). That said, I don't think using Inclusive/Exclusive terms would be all that unclear, so I don't have a problem with it. Still think it makes sense to change?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowRange.Helpers.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// Creates a <see cref="RowRange"/> instance with exclusive bounds.
+        /// </summary>
+        /// <param name="startKeyOpen">
+        /// The exclusive lower bound for the range. If null, it is interpreted as the empty string.
+        /// </param>
+        /// <param name="endKeyOpen">
+        /// The exclusive upper bound for the range. If null, it is interpreted as the infinite row key.
+        /// </param>
+        /// <returns>The created range.</returns>
+        public static RowRange Open(RowKey startKeyOpen, RowKey endKeyOpen)
```
sduskis: Open is useful for "smart retries" of RowReads.  In a case of partial responses of a long read, a request that started with ClosedOpen would start after the previously returned row via a Open() RowRange.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ColumnRange.Helpers.cs
```diff
@@ -0,0 +1,103 @@
...
+        /// The name of the column family within which this range falls.
+        /// </param>
+        /// <param name="startQualifierClosed">
+        /// The inclusive upper bound for the range. If null, it is interpreted as the infinite row key.
+        /// </param>
+        /// <param name="endQualifierClosed">
+        /// The inclusive lower bound for the range. If null, it is interpreted as the empty string.
+        /// </param>
+        /// <returns>The created range.</returns>
+        public static RowFilter Closed(
```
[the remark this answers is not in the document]
evildour: Oops. Copy/paste error. Fixed.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Mutations.cs
evildour: Unfortunately, the `Entry` type itself has no docs. The param docs are copied though from its auto-generated properties. I'll add some more to the summary and returns to clarify.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersion.cs
garye: In another doc we discussed removing the concept of "server time" and always setting the timestamp, so `ServerTime()` is not the right name but I'm not sure what is. Maybe something with `Default` or `Now`.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: Just to clarify, are you saying there should be 4 additional overloads to `RowFilters.ColumnRange` that take the start/end values, and named appropriately like `ColumnRangeOpen` or `ColumnRangeExclusive` (depending on which naming we go with)? Also, when you say 'move' them, do you mean they should no longer appear as factory methods on `ColumnRange` or that they should exist in both places?

Also, I assume the same deicsions would apply to `ValueRange`, right?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: `string` is implicitly convertible to `ColumnQualifier`. I specifically added `RowKey` and `ColumnQualifier` to reduce the number of overloads required, especially in cases where multiple keys, qualifiers were needed. The only case where I haven't done this is for the cell values, and it results in 8 overloads of `Cell.ValueRange` instead of 4 (it really should be 16 to cover all cases, but I think what I have should be sufficient).

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: As with `ColumnQualifier`, string can be used here.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
jskeet: Ah, that makes sense. It's probably worth mentioning that in remarks just for clarity - eventually. (I'm very aware this is just an early stage skeleton.)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
jskeet: I'm saying that if I know that there are factory methods in `RowFilters` to create row filters, I'm not sure I'd think to look elsewhere as well. But this is very much from the perspective of a review without reading any docs etc. I can see that there's a lot of benefit in using the ColumnRange and ValueRange types to give extra information in the type name - I hadn't appreciated that there'd be both, and they both create row filters...

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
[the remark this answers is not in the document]
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: Ok, I'm going to leave as is for now, but add TODOs for these two filter types to see if we should have methods taking start/end values for convenience.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: Done (and also anywhere else `ColumnQualifier` is used).

evildour: ...and `RowKey` as well.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: I'm assuming this no longer applies, but correct me if I'm wrong.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowKey.cs
evildour: I assume so. Added `IComparable` and `IComparable<RowKey>` and added a TODO to verify this assumption.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowKey.cs
evildour: Done. And the same in `ColumnQualifier` as well.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersion.cs
evildour: I think `Default` would make more sense than `Now` here since it sounds like we want to avoid naming which implies the versions are always timestamps. I could make this a static method which always returns the version with the timestamp micros for UTC now. (I want to avoid making it a property so users don't assume it would return the same value each time, although `DateTime.Now` is a property...)

So it would look like `var version = BigtableVersion.Default();` That seem ok?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowRange.Helpers.cs
evildour: Ok, good to know about this and about `Open` below. Do you think it makes sense then to remove some of the options and/or name this a bit differently to make it the default choice?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersion.cs
```diff
@@ -61,11 +61,20 @@ public DateTime ToDateTime()
         /// <summary>
         /// Compares the versions to another.
         /// </summary>
+        /// <param name="obj">The other version to compare this to.</param>
+        /// <returns>
+        /// A negative value if this is less than the other version, a positive value if this is greater, and zero if they are equal.
+        /// </returns>
+        public int CompareTo(object obj) => Compare(this, obj as BigtableVersion?);
```
jskeet: I think this should throw ArgumentException if compared with a non-BigtableVersion.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersionRange.cs
```diff
@@ -33,6 +33,41 @@ public BigtableVersionRange(BigtableVersion? start, BigtableVersion? end)
...
+            throw new NotImplementedException();
+        }
+
+        // TODO: Get clarification on use cases to see how useful this would be.
+        /// <summary>
+        /// Creates a new <see cref="BigtableVersionRange"/>.
+        /// </summary>
+        /// <param name="startTimestamp">
+        /// Inclusive lower bound timestamp whose microseconds should be used as the version value.
+        /// If null, interpreted as 0. It is recommended that timestamps be specified in UTC so there
```
jskeet: I'd *require* the Kind to be UTC to be honest. Better to alert callers to problems early, IMO.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ColumnQualifier.cs
```diff
@@ -130,6 +131,15 @@ public override string ToString()
         /// 
         /// </summary>
         /// <param name="qualifier"></param>
+        public static implicit operator string(ColumnQualifier qualifier)
```
jskeet: Not sure about the conversions this way being implicit. Let's discuss it when we get further along.
(Particularly for this one, which will presumably throw if the qualifier key isn't actually UTF-8-encoded text.)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ColumnRange.Helpers.cs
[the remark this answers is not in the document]
jskeet: Let's see what other languages are doing.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ColumnRange.Helpers.cs
```diff
@@ -34,7 +39,7 @@ public partial class ColumnRange
         /// The inclusive lower bound for the range. If null, it is interpreted as the empty string.
         /// </param>
         /// <returns>The created range.</returns>
-        public static RowFilter Closed(
+        public static ColumnRange Closed(
```
jskeet: Okay, with these returning ColumnRange, it makes a lot more sense :)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowKey.cs
```diff
@@ -56,6 +58,38 @@ public RowKey(ByteString key)
 
         // TODO: Docs
 
+        // TODO: Make sure making RowKey comparable is useful.
```
jskeet: One potential use: if the server always returns the rows in RowKey order, it will make it easier to write tests if you can take a bunch of rows and order them (locally) by row key, then expect them to be returned like that by the server.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowKey.cs
```diff
@@ -100,22 +134,47 @@ public override string ToString()
...
         /// <param name="x"></param>
         /// <param name="y"></param>
         /// <returns></returns>
-        public static bool operator !=(RowKey x, RowKey y)
-        {
-            throw new NotImplementedException();
-        }
+        public static bool operator <=(RowKey x, RowKey y) => Compare(x, y) <= 0;
 
+        /// <summary>
```
jskeet: We'll probably want to optimize this - in particular, if the keys are of different lengths, there's no point in comparing them as far as they're the same.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersion.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersionRange.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ColumnQualifier.cs
evildour: Oops. More copy/paste errors. These should have been explicit. Fixed the `RowKey` ones as well.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowKey.cs
evildour: Not sure about any ordering guarantee. I'll ask internally.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowKey.cs
evildour: Done (for ColumnQualifier too)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ColumnRange.Helpers.cs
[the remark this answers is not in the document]
evildour: I just did a quick pass: Java uses the builder pattern, which uses names similar to the parameter names, python has a constructor with start/end params and then inclusive_start/end optional params for controlling inclusive/exclusive state, and Go just takes start/end and forces a closed start and open end. So it's really a mixed bag.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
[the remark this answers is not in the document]
evildour: I changed these to be more concise but let me know if you disagree.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ColumnQualifier.cs
```diff
@@ -115,6 +119,10 @@ public override string ToString()
         /// <returns></returns>
         public static bool operator !=(ColumnQualifier x, ColumnQualifier y)
         {
+            if (x?.ByteString.Length != y?.ByteString.Length)
```
jskeet: Consider just delegating: `!=(ColumnQualifier x, ColumnQualifier y) => !(x == y)`

Sorry for not suggesting that before.

evildour: Not a problem. Thanks for pointing that out. I usually do that but just forgot in this case.

evildour:
Good question and I guess it's up to you, but here's what I was thinking: I didn't want to merge anything which wasn't implemented. So I was thinking I would create a separate PRs for each set of methods on the `BigtableClient` implemented along with all supporting helper types/snippets/docs/tests (mostly copied from here). Once all PRs are created, I would close this one. Does that sound ok or do you think that would create unnecessary re-reviews of the stuff I already did here?

jskeet:
That sounds like a good option to me - I'm doing something similar with Firestore, although for that I'll probably go with multiple commits in the same PR. Feel free to take either path.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowKey.cs
evildour: So it looks the cells will sorted by row key, then family (by some internal id though, so this needs to be re-sorted lexicographically by this library), then column qualifier, then timestamp/version descending. So this could be comparable for testing and so could `ColumnQualifier`. The more I think of it, the more I think it makes sense to merge these types into one and then use the new type for cell values as well. They are pretty much identical. `BigtableBytestring` perhaps? Also, I think it makes sense to make it a struct to reduce any additional GC pressure when dealing with a large numbers of cells. WDYT?

jskeet: +1 for using a struct to wrap the ByteString. (I'm creating a Blob struct in Firestore, but I'm not making that comparable yet as I don't need to.)

I don't understand the domain well enough to judge whether one type is better than three here or not... I'd go with your gut, and we can always change our minds. (Also, check what Java does.)

evildour: Java seems to always uses byte[] or ByteString. The samples show all row keys being sent to `Bytes.toBytes` before making requests. Go seems to take string always. And python looks like it can take bytes or strings for the same parameter and internally converts if necessary.

I feel like it make sense to merge these. The whole point is so we can introduce implicit operators so you don't have to see the type name of `ByteString` or whatever these are called. The only possible thing which might be useful for one and not the others is a method/operator to convert the byte string to/from a long for cases where a cell value can be treated like a 8-byte, big-endian encoded signed integer. It would make getting/setting the current value in those cells much easier. But I don't think having those unnecessary members would hurt the row key or column qualifiers in any way.

jskeet:
With all the code that's been submitted already and the way things have moved around as you've implemented, is this PR still relevant? (Fine if so, let's close otherwise.)

evildour:
Yes, I wanted to leave this so it contains the skeleton code I could pull into each PR as I implement various pieces. Each rebase will make it smaller and smaller until it's no longer needed.

jskeet:
Aha, I hadn't spotted that you were rebasing. Carry on :)

evildour:
Closing. All proposed changes have been updated/implemented in other PRs.

jskeet:
Congrats - this is a great milestone :)

evildour:
Thanks! And just a day over my promise to get it done by Q4 (a few minor TODOs aside).

PR #1509, opened 2017-10-02
Remove obsolete members of BigQuery package

jskeet:
Two new tests that should probably have been there before: check
that we can't use unnamed parameters when the parameter mode is named.

Note: I don't intend to submit this PR for a while; I want the obsolete members to be available probably until the end of October (giving a month for migration) and it's unclear whether there may be more releases before then.


jskeet:
It's been a couple of weeks now, and we have no specific plan for an imminent next release. Merging now to avoid this getting stale.

PR #1514, opened 2017-10-03
For now, pass None into gRPC MoveNext and document the limitation.

benwulfe:
In the future, we may alter the design to support better cancellation of MoveNext.

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs
@@ -92,6 +92,19 @@ namespace Google.Cloud.Spanner.V1 {
         }
 
         /// <summary>
+        /// This helper ensures we invalidate the _currentCall state if we want to throw canceled.
+        /// </summary>
+        private void ThrowIfCancellationRequested(CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                _currentCall?.Dispose();
+                _currentCall = null;
+                cancellationToken.ThrowIfCancellationRequested();
+            }
+        }
+
+        /// <summary>
         ///     Connects or reconnects to Spanner, fast forwarding to where we left off based on
         ///     our _resumeToken and _resumeSkipCount.
         /// </summary>
@@ -99,11 +112,11 @@ namespace Google.Cloud.Spanner.V1 {
         private async Task<bool> ReliableConnectAsync(CancellationToken cancellationToken) {
             if (_currentCall == null) {
                 await ConnectAsync().ConfigureAwait(false);
-
                 Debug.Assert(_currentCall != null, "_currentCall != null");
-                cancellationToken.ThrowIfCancellationRequested();
 
                 for (int i = 0; i <= _resumeSkipCount; i++) {
+                    ThrowIfCancellationRequested(cancellationToken);
+
                     //This calls a simple movenext on purpose.  If we get an error here, we'll fail out.
                     //TODO(benwu): Fix cancel on MoveNext in a subsequent change targeting spanner GA
                     _isReading = await _currentCall.ResponseStream.MoveNext(CancellationToken.None).WithSessionChecking(() => _session).ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs
```diff
@@ -101,11 +101,23 @@ public sealed class ReliableStreamReader : IDisposable
...
+                {
+                    _currentCall?.Dispose();
+                    _currentCall = null;
+                    cancellationToken.ThrowIfCancellationRequested();
+                }
 
                 for (int i = 0; i <= _resumeSkipCount; i++) {
                     //This calls a simple movenext on purpose.  If we get an error here, we'll fail out.
-                    _isReading = await _currentCall.ResponseStream.MoveNext(cancellationToken).WithSessionChecking(() => _session).ConfigureAwait(false);
+                    //TODO(benwu): pass in cancellationtoken and capture cancel to clean up _currentCall if needed.
```
iantalarico: Is there an issue to track this?  If this is needed for GA I'd consider filing an issue.

benwulfe: I don't believe we need to fix this for GA as its a limitation in (already GA'd) gRPC.  I'll let Jon comment on it however.

jskeet: I think we should at least note it as a limitation in the documentation somewhere - it's pretty surprising. I think I'd like to hold of on GA'ing in the current state until we at least know what we're going to do - I don't think we'll want it to be like this long-term.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs
```diff
@@ -101,11 +101,23 @@ public sealed class ReliableStreamReader : IDisposable
...
+                    _isReading = await _currentCall.ResponseStream.MoveNext(CancellationToken.None).WithSessionChecking(() => _session).ConfigureAwait(false);
```
jskeet: An alternative approach would be to use `Task.WhenAny` with the `MoveNext` task and a "just monitor cancellation token" task (I think there's a method to produce one of those). That way we wouldn't have an unbounded delay between token cancellation and response.

This PR is better than the previous code, so I'm happy for it to be merged, but the unbounded nature worries me at the moment.

benwulfe:
I actually need to rework this a bit.  I believe I can rework this in such
a way as to use the stream recovery code to handle rewinding on a cancel.
That is, if the ReadAsync is canceled (anywhere), I invalidate the current
response stream and ensure the iterator counter + state was not advanced.
A subsequent call to ReadAsync should recover and move to the correct place.

I could do the WhenAny trick or simply pass in None. I'd actually prefer to
hold off on this part of the logic until you and Jan get it sorted.
However, the logic around it to ensure iteration at the DbDataReader stays
valid should not have to change as much once I fix it.  Inner logic will
eventually do one of three things:
1. send in None and not support cancel within the MoveNext call.  it might
still hit cancel during a reconnect and invalidate the response stream
there.
2. do WhenAny and invalidate the response stream if cancelled is hit.
3. do your workaround of cancelling the response stream token -- again
using reconnect recovery to advance back to the correct place if the
dbdatareader is continued to be used.

#1 is the simplest and least risky tactically until we get the correct
solution sorted out.  I prefer your #3 and would have added it, but it
requires quite a bit more testing.  #2 is 'ok' but isn't something we want
long term anyway due to the fact it lets the operation run out.

jskeet:
I'd definitely at least pause for a little while - https://github.com/grpc/grpc/pull/12822 is attempting to fix this on the gRPC side.

benwulfe:
If that's in progress I'll just hold off for a bit.  Right now, sending in
a token into ReadAsync will throw with the same message, but its not
entirely blocking.
also.. looks like a missing test case.

benwulfe:
I need to get this in to unblock a user.  I'll switch from passing None to passing the actual token once grpc releases and we update our side.  At that point I will also need to catch the RpcException and detect it as a proper cancellation so I don't attempt a recover.

PR #1521, opened 2017-10-10
Firestore option types

jskeet:


apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/SetOptions.cs
```diff
@@ -0,0 +1,83 @@
...
+        /// Whether or not to merge values. If <see cref="FieldMask"/> is non-empty,
+        /// this is always <c>true</c>.
+        /// </summary>
+        public bool Merge { get; }
+
+        /// <summary>
+        /// The fields to merge. This property is never null; if the mask is empty,
+        /// all fields will be merged or overwritten.
+        /// </summary>
+        public ReadOnlyCollection<FieldPath> FieldMask { get; }
```
chrisdunelm: Would `IReadOnlyCollection` be better?

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/Precondition.cs
```diff
@@ -0,0 +1,68 @@
...
+        /// Creates a precondition that the document has the specified last update time.
+        /// </summary>
+        public static Precondition LastUpdated(Timestamp timestamp) =>
+            new Precondition(timestamp, null, new V1Beta1.Precondition { UpdateTime = timestamp.ToProto()});
+
+        /// <summary>
+        /// Condition that the document was last updated at the specified timestamp, if specified.
+        /// If this is non-null, <see cref="Exists"/> will be null.
+        /// </summary>
+        public Timestamp? LastUpdateTime { get; }
```
chrisdunelm: Why not read these properties from the proto property, rather than duplicate the data?

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/SetOptions.cs
jskeet: Hmm. Or maybe `IReadOnlyList`... will check that works with netstandard1.5 and go to that.

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/Precondition.cs
jskeet: Could do, yes. Will do that. (I originally produced the proto on demand in a method, but now I've got it as a property anyway, I might as well use it.)

PR #1522, opened 2017-10-10
Automatically add a resource prefix header to gRPC calls

jskeet:
This does not affect Listen/Write as they are client-streaming calls.

We reduce the resource name down to just the project ID and database
portion.

### DIFF
--- a/apis/Google.Cloud.Firestore.V1Beta1/Google.Cloud.Firestore.V1Beta1/FirestoreClientPartial.cs
+++ b/apis/Google.Cloud.Firestore.V1Beta1/Google.Cloud.Firestore.V1Beta1/FirestoreClientPartial.cs
@@ -1,0 +1,115 @@
+﻿// Copyright 2017, Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax.Grpc;
+using System;
+
+namespace Google.Cloud.Firestore.V1Beta1
+{
+    public partial class FirestoreClientImpl
+    {
+        private const string ResourcePrefixHeader = "google-cloud-resource-prefix";
+
+        partial void Modify_BatchGetDocumentsRequest(ref BatchGetDocumentsRequest request, ref CallSettings settings) =>
+            ApplyResourcePrefixHeader(ref settings, request.Database);
+
+        partial void Modify_BeginTransactionRequest(ref BeginTransactionRequest request, ref CallSettings settings) =>
+            ApplyResourcePrefixHeader(ref settings, request.Database);
+
+        partial void Modify_CommitRequest(ref CommitRequest request, ref CallSettings settings) =>
+            ApplyResourcePrefixHeader(ref settings, request.Database);
+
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.Firestore.V1Beta1/Google.Cloud.Firestore.V1Beta1/FirestoreClientPartial.cs
```diff
@@ -0,0 +1,115 @@
...
+            }
+            string database = GetDatabaseResourceName(resource);
+            settings = settings.WithHeader(ResourcePrefixHeader, database);
+        }
+        
+        // Visible for testing
+
+        /// <summary>
+        /// Retrieves the database resource name from a full resource name.
+        /// Validation is performed as far as the database ID but no further; the database ID is deemed to end at the first slash.
```
chrisdunelm: first? slash

apis/Google.Cloud.Firestore.V1Beta1/Google.Cloud.Firestore.V1Beta1/FirestoreClientPartial.cs
```diff
@@ -0,0 +1,115 @@
...
+        /// </summary>
+        /// <param name="resource">The resource name, which must start with projects/{project_id}/databases/{database_id}</param>
+        /// <returns>The database resource name</returns>
+        internal static string GetDatabaseResourceName(string resource)
+        {
+            const string projectsPrefix = "projects/";
```
chrisdunelm: This could probably all be done in a single regex.
Is this undesirable because it would be slower?

apis/Google.Cloud.Firestore.V1Beta1/Google.Cloud.Firestore.V1Beta1/FirestoreClientPartial.cs
jskeet: First slash from the start of the ID, yes. Suggestions for clearer wording?

apis/Google.Cloud.Firestore.V1Beta1/Google.Cloud.Firestore.V1Beta1/FirestoreClientPartial.cs
jskeet: That's a good question. Will benchmark, but yes, I suspect that would be slower.

jskeet: Now benchmarked: 50ns for manual code vs 300ns for regex.

apis/Google.Cloud.Firestore.V1Beta1/Google.Cloud.Firestore.V1Beta1/FirestoreClientPartial.cs
chrisdunelm: I think I'm misunderstanding something. Doesn't the database ID end at the fourth slash?

jskeet: The fourth slash of the full resource name, yes. But if you think of it as "I'm starting looking for the end of the database ID, from the start of the database ID" then you stop as soon as you hit a slash.

chrisdunelm: OK, thanks.

apis/Google.Cloud.Firestore.V1Beta1/Google.Cloud.Firestore.V1Beta1/FirestoreClientPartial.cs
chrisdunelm: :)

jskeet: (Almost certainly insignificant in the grand scheme of things, but pretty easy to do, and it gets applied to all calls...)

PR #1530, opened 2017-10-11
Get snapshots - third attempt

jskeet:


jskeet:
(This was created when github was having problems. It seems to have recovered now.)

PR #1531, opened 2017-10-11
Introduce FirestoreDb.GetDocumentSnapshotsAsync

jskeet:
Currently just an internal method; this will be exposed via
other public methods.


jskeet:
(This was created when github was having problems. It seems to have recovered now.)

PR #1533, opened 2017-10-11
Introduce collection-listing methods

jskeet:
These aren't unit tested as it's relatively hard to fake out
IPagedAsyncEnumerable - something we should address. For the moment,
I'm relying on integration tests.

### DIFF
--- a/apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data.IntegrationTests/CollectionListingTest.cs
+++ b/apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data.IntegrationTests/CollectionListingTest.cs
@@ -1,0 +1,61 @@
+﻿// Copyright 2017, Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Grpc.Core;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Firestore.Data.IntegrationTests
+{
+    [Collection(nameof(FirestoreFixture))]
+    public class CollectionListingTest
+    {
+        private readonly FirestoreFixture _fixture;
+
+        public CollectionListingTest(FirestoreFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public async Task ListRootCollectionsAsync()
+        {
+            var db = _fixture.FirestoreDb;
+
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/FirestoreDb.cs
```diff
@@ -201,5 +201,21 @@ string ExtractPath(DocumentReference documentReference)
...
+
+        /// <summary>
+        /// Asynchronously retrieves the root collection IDs from the server.
+        /// </summary>
+        /// <param name="cancellationToken">A cancellation token for the operation.</param>
+        /// <returns>All the root collection IDs.</returns>
+        public Task<IList<CollectionReference>> ListRootCollectionsAsync(CancellationToken cancellationToken = default) =>
+            ListCollectionsAsync(null, cancellationToken);
+
+        internal async Task<IList<CollectionReference>> ListCollectionsAsync(DocumentReference parent, CancellationToken cancellationToken) =>
```
chrisdunelm: Is exposing this as a `Task<IList...>` the right thing to do? Rather than as an async collection of some sort?

Is the size of the returned collection always/usually small enough that fetching the whole thing is reasonable?

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data.IntegrationTests/CollectionListingTest.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            var doc = _fixture.NonQueryCollection.Document(null);
+            var col1 = doc.Collection("col1");
+            var col2 = doc.Collection("col2");
+            // Put documents in the collections to force them to exist.
+            await col1.AddAsync(new { Name = "doc1" });
+            await col2.AddAsync(new { Name = "doc2" });
+
+            var collections = await doc.ListCollectionsAsync();
+            Assert.Equal(new[] { col1, col2 }, collections.OrderBy(c => c.Path));
```
chrisdunelm: Not really important, but why the inconsistent `Assert`ing between this test and `ListRootCollectionsAsync` above?

jskeet: We've just created a new document, so we know that there only these collections will be present in this case. In the root collections case, there'll be collections from all previous test runs, and whichever other tests have already run in this case - so Contains is the best we can do.

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/FirestoreDb.cs
jskeet: The return type is as per the specification I've got - but looking at the Java code, that only seems to create a single request. I'll clarify this with the team.

We could potentially use "Get" as the prefix for now with this return type, and add "List" later on that returns a paged enumerable.

chrisdunelm: Yes, I think using a `Get` prefix for this is a good idea.

jskeet:
Moved to IAsyncEnumerable, with TODOs to consider adding Get methods. PTAL.

PR #1536, opened 2017-10-12
Resource prefix header for Bigtable gRPC calls

evildour:


apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClientPartial.cs
```diff
@@ -0,0 +1,75 @@
...
+            ApplyResourcePrefixHeaderForInstance(ref settings, request.Parent);
+
+        // TODO: Determine why this generated as Cluster instead of UpdateClusterRequest
+        partial void Modify_Cluster(ref Cluster request, ref CallSettings settings) =>
+            ApplyResourcePrefixHeader(ref settings, request.ClusterName);
+
+        partial void Modify_DeleteClusterRequest(ref DeleteClusterRequest request, ref CallSettings settings) =>
+            ApplyResourcePrefixHeader(ref settings, request.ClusterName);
+
+        private static void ApplyResourcePrefixHeader(ref CallSettings settings, InstanceName instanceName)
```
jskeet: Rather than have all of these methods accepting the resource types, I think it would be simpler (and more efficient) just to use the string in the proto - `request.Cluster` etc. Is there any downside to that that I haven't considered?

evildour: I was largely basing this off #1522. I noticed you had validation logic in there and was going to do something similar but the resource name parsing logic validates the string anyway, so I thought it would be good to reuse that.

jskeet: I have *some* validation logic, but only as far as I need to in order to get the right value. (I don't check that anything after the slash is "documents" for example.) I'd suggest setting it blindly, and let routing/servers validate.

evildour: Ok, makes sense.

evildour:
PTAL

apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClientPartial.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            ApplyResourcePrefixHeader(ref settings, instanceName.ToString());
+        }
+
+        private static void ApplyResourcePrefixHeader(ref CallSettings settings, ClusterName clusterName)
```
jskeet: Not sure why these methods are still needed - if you just use `request.Cluster` etc, don't you already get strings that you can specify directly?

So for example, instead of:

```csharp
partial void Modify_DeleteClusterRequest(ref DeleteClusterRequest request, ref CallSettings settings) =>
    ApplyResourcePrefixHeader(ref settings, request.ClusterName);
```

have


```csharp
partial void Modify_DeleteClusterRequest(ref DeleteClusterRequest request, ref CallSettings settings) =>
    ApplyResourcePrefixHeader(ref settings, request.Cluster);
```

... and just have the one ApplyResourcePrefixHeader method.

jskeet: (It's possible I've missed something important, of course...)

evildour: Oh, I had no idea the string version was always present. It is called `Name`, so I didn't see it in the intellisense list next to `ClusterName`. Thanks. Will fix.

evildour:
I believe so. I'll loop you into a discussion I had about this internally.

PR #1539, opened 2017-10-12
Add extra time to the error reporting integration tests.

iantalarico:
The tests started failing a few days ago, it looks like they are taking a little longer to get processed.


iantalarico:
Thanks Jon!

PR #1543, opened 2017-10-13
BigtableByteString wrapper type

evildour:
This has a few changes from #1506:
- Added a Length property
- Added an indexer
- Made the conversion from BigtableByteString to ByteString implicit (so now both directions are implicit).

@jskeet I copied your EqualityTester from Firebase to make the ComparisonTester, but on line 47 there's a test for `default(T) == null`. I believe that should be `default(T) != null`, right? I left it as you had it for now just incase I'm missing something. But if it should be changed, I'll update both places.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableByteStringTest.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableByteStringTest.cs
@@ -39,7 +39,9 @@ namespace Google.Cloud.Bigtable.V2.Tests
             var unequal = new[]
             {
                 new BigtableByteString(0),
-                new BigtableByteString(0L)
+                new BigtableByteString(0L),
+                new BigtableByteString("a"),
+                new BigtableByteString(ByteString.CopyFrom(1, 2, 3))
             };
             ComparisonTester.AssertEqual(control, equal, unequal);
             ComparisonTester.AssertEqualityOperators(control, equal, unequal);
@@ -144,13 +146,13 @@ namespace Google.Cloud.Bigtable.V2.Tests
         }
 
         [Theory]
-        [InlineData("BigtableByteString: Length=0; UTF-8 String=")]
+        [InlineData("BigtableByteString: Length=0; Hex=")]
         [InlineData("BigtableByteString: Length=6; Hex=00 01 02 03 0A 10", 0, 1, 2, 3, 10, 16)]
         [InlineData("BigtableByteString: Length=32; Hex=00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F 10 11 12 13 14 15 16 17 18 19 1A 1B 1C 1D 1E 1F",
             0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31)]
         [InlineData("BigtableByteString: Length=33; Hex (first 32 bytes only)=00 01 02 03 04 05 06 07 08 09 0A 0B 0C 0D 0E 0F 10 11 12 13 14 15 16 17 18 19 1A 1B 1C 1D 1E 1F",
             0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28, 29, 30, 31, 32)]
-        [InlineData("BigtableByteString: Length=8; UTF-8 String=Bigtable", 66, 105, 103, 116, 97, 98, 108, 101)]
+        [InlineData("BigtableByteString: Length=8; Hex=42 69 67 74 61 62 6C 65", 66, 105, 103, 116, 97, 98, 108, 101)]
         public void Formatting(string expectedText, params int[] data)
         {
             // xUnit isn't happy if the parameter type is actually byte[].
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableByteStringTest.cs
```diff
@@ -0,0 +1,161 @@
...
+                new BigtableByteString((byte[])null),
+                new BigtableByteString((ByteString)null),
+                new BigtableByteString(""),
+                new BigtableByteString(new byte[0]),
+                new BigtableByteString(ByteString.Empty)
+            };
+            var unequal = new[]
+            {
+                new BigtableByteString(0),
+                new BigtableByteString(0L)
```
jskeet: Maybe add non-empty strings and ByteStrings here?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/ComparisonTester.cs
```diff
@@ -0,0 +1,176 @@
...
+                Assert.False(candidate.Equals(control));
+                Assert.False(control.Equals((object) candidate));
+                Assert.False(candidate.Equals((object) control));
+                // While this isn't actually required, if it fails it would usually be due to a bug.
+                Assert.NotEqual(control.GetHashCode(), candidate.GetHashCode());
+            }
+
+            Assert.False(control.Equals(null));
+
+            if (default(T) == null)
```
jskeet: This is deliberately == - it's basically saying "If T is a reference type or nullable value type, check that control.Equals(null) returns false."

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableByteString.cs
```diff
@@ -0,0 +1,301 @@
...
+
+using Google.Api.Gax;
+using Google.Protobuf;
+using System;
+using System.Text;
+
+namespace Google.Cloud.Bigtable.V2
+{
+    /// <summary>
+    /// The bytes string which can be used for a row key, column qualifier, or cell value.
```
jskeet: "The bytes string" doesn't sound quite right. How about "A string of bytes"? Or "A sequence of bytes"?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableByteString.cs
```diff
@@ -0,0 +1,301 @@
...
+        public override int GetHashCode() => Value.GetHashCode();
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            var value = Value;
+            var builder = new StringBuilder($"{nameof(BigtableByteString)}: Length={value.Length}; ");
+            int initialLength = builder.Length;
+
+            // For short byte strings which can be interpreted as simple strings, display them that way.
```
jskeet: Hmm. I'm not sure whether I'd do this. It would be weird to have a BigtableByteString based on a counter (for example) which kept changing representation. I'd just go for hex all the time, I think.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableByteString.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// </param>
+        public static implicit operator BigtableByteString(ByteString value) => new BigtableByteString(value);
+
+        /// <summary>
+        /// Converts the <see cref="BigtableByteString"/> value to its underlying byte string.
+        /// </summary>
+        /// <param name="value">
+        /// The BigtableByteString value to convert.
+        /// </param>
+        public static implicit operator ByteString(BigtableByteString value) => value.Value;
```
jskeet: Is this one deliberately implicit because it's just unwrapping? That's *probably* okay...

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableByteStringTest.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/ComparisonTester.cs
evildour: Oh I see, the first null check was checking the object overload. This one is checking the typed overload, but only if it's a reference type. Make sense now.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableByteString.cs
evildour: Oops. That was supposed to be "byte string", but I think "sequence of bytes" sounds better anyway.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableByteString.cs
evildour: Yeah, that makes sense. I was trying to make the debugging experience nicer for anyone using readable strings for row keys, column names, or cell values, but I guess it kind of hurts the experience in other ways.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableByteString.cs
evildour: Yes, I figured since it's just a thin wrapper, it should be ok to convert between them automatically. However, the only oddity is that converting a null `ByteString` to `BigtableByteString` and then back would change it to `ByteString.Empty`. I guess I don't feel too strongly about it. If you think it should go back to being explicit, I will change it.

jskeet: Nope, I think that's okay.

PR #1554, opened 2017-10-16
Spanner Logging now instance based (vs static) and allows bridging to…

benwulfe:
… EF logging which is also instance based.

New tests show leveraging logs to double check back end SQL Query construction.

### DIFF
--- a/apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Diagnostics/SpannerEventId.cs
+++ b/apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Diagnostics/SpannerEventId.cs
@@ -1,0 +1,43 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Diagnostics;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.Extensions.Logging;
+
+namespace Google.Cloud.EntityFrameworkCore.Spanner.Diagnostics
+{
+    /// <summary>
+    /// Event IDs for Cloud Spanner log messages
+    /// </summary>
+    public static class SpannerEventId
+    {
+        //For now, we are aggregating all the logging messages done within the ADO.NET
+        //layer under the "SpannerDiagnosticLog" Id
+
+        private enum Id
+        {
+            SpannerDiagnosticLog = CoreEventId.ProviderBaseId
+        }
+
+        private static readonly string s_logPrefix = DbLoggerCategory.Database.Name + ".";
+        private static EventId MakeEventId(Id id) => new EventId((int)id, s_logPrefix + id);
+
+        /// <summary>
+        /// </summary>
+        public static EventId SpannerDiagnosticLog { get; } = MakeEventId(Id.SpannerDiagnosticLog);
+
+    }
+}
--- a/apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Diagnostics/SpannerLogBridge.cs
+++ b/apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Diagnostics/SpannerLogBridge.cs
@@ -1,0 +1,94 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+
+namespace Google.Cloud.EntityFrameworkCore.Spanner.Diagnostics
+{
+    /// <summary>
+    /// This is internal functionality and not intended for public use.
+    /// </summary>
+    public class SpannerLogBridge<TLoggerCategory> : Cloud.Spanner.V1.Internal.Logging.DefaultLogger
+        where TLoggerCategory : LoggerCategory<TLoggerCategory>, new()
+    {
+        private readonly IDiagnosticsLogger<TLoggerCategory> _efLogger;
+
+        /// <summary>
+        /// This is internal functionality and not intended for public use.
+        /// </summary>
+        public SpannerLogBridge(IDiagnosticsLogger<TLoggerCategory> efLogger)
+        {
+            _efLogger = efLogger;
+        }
+
+        /// <summary>
+        /// This is internal functionality and not intended for public use.
+        /// </summary>
+        public override bool EnableSensitiveDataLogging
+        {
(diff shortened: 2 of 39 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
```diff
@@ -13,172 +13,304 @@
...
 {
     /// <summary>
-    /// Provides diagnostic information on Cloud Spanner.
-    /// By default, diagnostic logs will be sent to Trace.WriteLine for .Net desktop
-    /// and stderr for .Net Core.
-    /// The default LogLevel is LogLevel.Error which will only log error information.
-    /// Performance logging can be turned on by setting LogPerformanceTraces = true.
-    /// TODO: examine https://github.com/grpc/grpc/tree/master/src/csharp/Grpc.Core/Logging
-    /// to determine if it should be used instead.
+    /// This is an internal property and is not intended to be used by external code.
```
jskeet: Internal class?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -119,6 +119,11 @@ public override string ConnectionString
         [Category("Data")]
         public string SpannerInstance => _connectionStringBuilder?.SpannerInstance;
 
+        /// <summary>
+        /// This property is intended for internal use only.
+        /// </summary>
+        public Logger Logger { get; set; } = Logger.DefaultLogger;
```
jskeet: Just to check the intended semantics here - is it okay that changing the default logger after creating an instance of SpannerConnection doesn't change where that connection logs to?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
```diff
@@ -13,172 +13,304 @@
 // limitations under the License.
 
 using System;
+using System.Collections;
```
jskeet: I'm going to assume that all the implementation here was just moved from DefaultLogger - I haven't reviewed it in any detail.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/TransactionPool.cs
```diff
@@ -39,6 +39,8 @@ public static class TransactionPool
         // Holds a list of all active transactions with back pointers to the owning session.
         static readonly ConcurrentDictionary<ByteString, Session> s_activeTransactionTable = new ConcurrentDictionary<ByteString, Session>();
 
+        private static Logger Logger { get; } = Logger.DefaultLogger;
```
jskeet: Again, this won't see any changes to `Logger.DefaultLogger` - is that intentional? (More of a concern than the SpannerConnection one, as it's part of type initialization.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
```diff
@@ -13,172 +13,304 @@
...
-    /// to determine if it should be used instead.
+    /// This is an internal property and is not intended to be used by external code.
     /// </summary>
     public abstract class Logger
     {
         private static Logger s_instance;
+        private int _perfLoggingTaskEnabled;
+        private readonly ConcurrentDictionary<string, PerformanceTimeEntry> _perfCounterDictionary
+            = new ConcurrentDictionary<string, PerformanceTimeEntry>();
+        private readonly object _perfLogSyncObject = new object();
```
evildour: Two questions here:
1. Is `_perfLogSyncObject` needed? It seems like you could lock on `_perfCounterDictionary` instead in all cases.
2. Why does `_perfCounterDictionary` need to be a `ConcurrentDictionary<...>`? It seems to always be used inside a lock. Shouldn't `Dictionary<...>` suffice?

apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Diagnostics/SpannerEventId.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            SpannerDiagnosticLog = CoreEventId.ProviderBaseId
+        }
+
+        private static readonly string s_logPrefix = DbLoggerCategory.Database.Name + ".";
+        private static EventId MakeEventId(Id id) => new EventId((int)id, s_logPrefix + id);
+
+        /// <summary>
+        /// </summary>
+        public static readonly EventId SpannerDiagnosticLog = MakeEventId(Id.SpannerDiagnosticLog);
```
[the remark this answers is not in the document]
benwulfe: i do not *believe* so. I'm following patterns I see in existing ef core 2.0 code -- but I do not think anything is done here with reflection, but I could be wrong.

apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Diagnostics/SpannerLogBridge.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        /// <summary>
+        /// This is internal functionality and not intended for public use.
+        /// </summary>
+        public override bool EnableSensitiveDataLogging
+        {
+            get => _efLogger.ShouldLogSensitiveData();
+            set
+            {
+                //Not enabled here.  Use DbContextOptionsBuilder.EnableSensitiveDataLogging.
```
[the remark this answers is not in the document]
benwulfe: I think its slightly safer to ignore it.  There is an off chance they have code running against ADO.NET spanner and are enabling the feature in a code path that isn't always executed. Then they migrate to use EntityFramework. In that case it shouldn't be treated as a fatal error.

I will however log an error to complain about the mistake in the setter...

apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Update/Internal/SpannerModificationCommandBatch.cs
```diff
@@ -18,8 +18,10 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Google.Api.Gax;
+using Google.Cloud.EntityFrameworkCore.Spanner.Diagnostics;
```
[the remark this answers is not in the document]
benwulfe: Yes, that's actually called out in my EF spec.  This is the recommended approach for naming entity framework packages and is unfortunately a bit inconsistent with the rest of our package names.  I felt it was more important to follow EF conventions here just in case it pops up later in (for example) how you can reference things in powershell, etc.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
benwulfe: Yes, I didn't want the ctor to require a Logger (or even hint at one) since it will not normally be set by the user.  Its expected that if its necessary, this will be overwritten in the object initializer.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
benwulfe: that's correct

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/TransactionPool.cs
benwulfe: There were some code paths where it did not make sense to log messages via an instance -- shared pools were one case.  So right now these will not be seen from EF -- I might revisit this, but it could be odd to see some messages from a pool show up in one EF logger and other messages in another logger.

apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Update/Internal/SpannerModificationCommandBatch.cs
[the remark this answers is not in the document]
jskeet: Completely agree.

jskeet:
(Still LGTM :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
benwulfe: I agree.  Got rid of sync and switched to a normal Dictionary.

PR #1555, opened 2017-10-16
Adds a public way to clear out Cloud Spanner pooled resources

benwulfe:
Adds a public way to clear out resources as a fallback for possible race conditions with grpc Environment shutdown event callbacks.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
@@ -83,6 +83,11 @@ namespace Google.Cloud.Spanner.Data
         /// </summary>
         public static SpannerOptions SpannerOptions => SpannerOptions.Instance;
 
+        /// <summary>
+        /// Releases all pooled Cloud Spanner sessions.
+        /// </summary>
+        public static Task ClearPooledResourcesAsync() => SessionPool.Default.ReleaseAllAsync();
+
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -83,6 +83,12 @@ public sealed class SpannerConnection : DbConnection
         /// </summary>
         public static SpannerOptions SpannerOptions => SpannerOptions.Instance;
 
+        /// <summary>
+        /// Releases all pooled Cloud Spanner sessions.
+        /// This should only be called before the current AppDomain has begun to shut down.
```
jskeet: Do you mean "it shouldn't be called after the AppDomain has begun to shut down" or "it should only be called because the AppDomain is about to shut down"? I suspect the former, but it may be worth clarifying. Also, what's the impact if it's called *after* the current AppDomain has begun to shut down?

benwulfe: You are correct its the former.  I'll wordsmith.  There actually isn't much of an impact if it is called at shutdow.  It at worst will cause all the current sessions to be simultaneously evicted via multiple tasks -- but the actual act of evicting a session is synchronized and no-ops if its already removed, so it should be ok, just wasteful.  
On further thought, I might simply remove the comment altogether.

PR #1557, opened 2017-10-17
First pass at transaction support

jskeet:
No integration tests yet, as I'm trying to find out how to reliably force a retriable failure.
The rules about when to roll back and which commits can be retried are slightly up in the air, but
we can change them later. Right now:

- We roll back if we haven't even tried to commit
- We roll back if we tried to commit and it failed with abort/cancellation/deadline
- We retry the whole transaction if commit failed with abort

### DIFF
--- a/apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data.Tests/FirestoreDbTest.RunTransactionAsync.cs
+++ b/apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data.Tests/FirestoreDbTest.RunTransactionAsync.cs
@@ -1,0 +1,227 @@
+﻿// Copyright 2017, Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Firestore.V1Beta1;
+using Google.Protobuf;
+using Grpc.Core;
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Xunit;
+using static Google.Cloud.Firestore.Data.Tests.ProtoHelpers;
+
+namespace Google.Cloud.Firestore.Data.Tests
+{
+    public partial class FirestoreDbTest
+    {
+        // Just check the basics of each overload...
+        [Fact]
+        public async Task RunTransactionAsync_CommitSuccess_Sync_NoResult()
+        {
+            var client = new TransactionTestingClient();
+            var db = FirestoreDb.Create("proj", "db", client);
+            await db.RunTransactionAsync(transaction =>
+            {
+                transaction.Create(db.Document("col/doc1"), new { Name = "Test" });
+                transaction.Delete(db.Document("col/doc2"));
+            });
+            Assert.Equal(new[] { CreateCommitRequest("transaction 1") }, client.CommitRequests);
+            Assert.Empty(client.RollbackRequests);
+        }
+
+        [Fact]
+        public async Task RunTransactionAsync_CommitSuccess_Async_NoResult()
+        {
+            var client = new TransactionTestingClient();
+            var db = FirestoreDb.Create("proj", "db", client);
+            await db.RunTransactionAsync(async transaction =>
+            {
+                await transaction.GetDocumentSnapshotAsync(db.Document("col/x"));
+                transaction.Create(db.Document("col/doc1"), new { Name = "Test" });
+                transaction.Delete(db.Document("col/doc2"));
+            });
+            Assert.Equal(new[] { CreateCommitRequest("transaction 1") }, client.CommitRequests);
+            Assert.Empty(client.RollbackRequests);
+        }
+
+        [Fact]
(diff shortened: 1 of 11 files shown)


apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/FirestoreDb.cs
```diff
@@ -215,5 +217,129 @@ string ExtractPath(DocumentReference documentReference)
...
+        /// </summary>
+        /// <typeparam name="T">The result type of the callback.</typeparam>
+        /// <param name="callback">The callback to execute. Must not be null.</param>
+        /// <param name="options">The options for the transaction. May be null, in which case default options will be used.</param>
+        /// <param name="cancellationToken">A cancellation token for the operation. This is exposed to the callback through <see cref="Transaction.CancellationToken"/>
+        /// and applied to all RPCs to begin, rollback or commit the transaction.</param>
+        /// <returns>A task which completes when the transaction has committed. The result of the task then contains the result of the callback.</returns>
+        public async Task<T> RunTransactionAsync<T>(Func<Transaction, Task<T>> callback, TransactionOptions options = null, CancellationToken cancellationToken = default)
+        {
+            ByteString previousTransactionId = null;
```
chrisdunelm: Check callback isn't null

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/FirestoreDb.cs
```diff
@@ -215,5 +217,129 @@ string ExtractPath(DocumentReference documentReference)
...
+        /// <typeparam name="T">The result type of the callback.</typeparam>
+        /// <param name="callback">The callback to execute. Must not be null.</param>
+        /// <param name="options">The options for the transaction. May be null, in which case default options will be used.</param>
+        /// <param name="cancellationToken">A cancellation token for the operation. This is exposed to the callback through <see cref="Transaction.CancellationToken"/>
+        /// and applied to all RPCs to begin, rollback or commit the transaction.</param>
+        /// <returns>A task which completes when the transaction has committed. The result of the task then contains the result of the callback.</returns>
+        public Task<T> RunTransactionAsync<T>(Func<Transaction, T> callback, TransactionOptions options = null, CancellationToken cancellationToken = default)
+        {
+            GaxPreconditions.CheckNotNull(callback, nameof(callback));
+            return RunTransactionAsync(async transaction => callback(transaction), options, cancellationToken);
```
chrisdunelm: Is there any advantage of `async transaction => ...` to ` transaction => Task.FromResult(...)` ?
The later seems to me to be easier to understand! And won't give CS1998 warning.

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/FirestoreDb.cs
```diff
@@ -215,5 +217,129 @@ string ExtractPath(DocumentReference documentReference)
         internal IAsyncEnumerable<CollectionReference> ListCollectionsAsync(DocumentReference parent) =>
             Client.ListCollectionIdsAsync(parent?.Path ?? RootPath)
                 .Select(id => new CollectionReference(this, parent, id));
+
+        // TODO: Is it appropriate to use ConfigureAwait(false) here? Should the caller be able to specify the context for executing their callback?
+        // TODO: Collect up all the exceptions we've retried?
+        // TODO: Is it nice to have all these overloads?
```
chrisdunelm: I think just keeping the two overloads with async callbacks is fine. As the `RunTransactionAsync` method is itself async, I think requiring an async callback is reasonable.

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data.Tests/FirestoreDbTest.RunTransactionAsync.cs
```diff
@@ -0,0 +1,229 @@
...
+                    await transaction.GetDocumentSnapshotAsync(db.Document("col/x"));
+                    throw new IOException("Bang!");
+                }));
+            Assert.Equal("Bang!", exception.Message);
+
+            Assert.Empty(client.CommitRequests);
+            Assert.Equal(new[] { CreateRollbackRequest("transaction 1") }, client.RollbackRequests);
+        }
+
+        /*[Theory]
```
chrisdunelm: Remove?

jskeet: Whoops - reinstated instead. Changed it to a fact to make debugging work.

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/FirestoreDb.cs
jskeet: I thought so until I started writing tests - but as the modification methods are synchronous, I think it won't be uncommon to have a synchronous callback, in which case you get a warning due to an async lambda with no awaits. (Or you have to return Task.FromResult directly.)

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/FirestoreDb.cs
jskeet: Done.

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/FirestoreDb.cs
jskeet: It was for the sake of propagating exceptions in the same way - but as we're awaiting the result immediately, it should be fine actually. Will change it.

PR #1558, opened 2017-10-18
Implement sentinel values for Firestore

jskeet:
The internal representation of sentinel values is ghastly: we use
the fact that NullValue is a single-value enum, but that protos can
store undefined enums. So the temporary Value representations of
sentinel values are highly unusual null values... but fortunately
they'll never be visible to the user, or transmitted on the wire.

Everything else about this is *reasonably* straightforward. There's
an odd situation where a field has a map value, but all the
sub-values are sentinel values; we end up transmitting an empty map,
which *may* not be what's desired. Will check that with the
Firestore team, but I don't think it should block this PR.

### DIFF
--- a/apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/WriteBatch.cs
+++ b/apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/WriteBatch.cs
@@ -22,11 +22,11 @@ using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using static Google.Cloud.Firestore.V1Beta1.DocumentTransform.Types;
+using static Google.Cloud.Firestore.V1Beta1.DocumentTransform.Types.FieldTransform.Types;
 
 namespace Google.Cloud.Firestore.Data
 {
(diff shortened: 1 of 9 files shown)


jskeet:
There's a TODO in this PR about FieldPath.Append. I'm going to do that in a separate PR.

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/WriteBatch.cs
```diff
@@ -170,6 +168,93 @@ public WriteBatch Set(DocumentReference documentReference, object documentData,
...
+                    {
+                        ValidateNoSentinelValues(value.ArrayValue.Values);
+                    }
+                }
+                removals?.ForEach(key => fields.Remove(key));
+            }
+
+            // Adds an item to an existing list, or creates a new list containing the item if there's currently no
+            // list. This allows us to avoid creating lots of lists redundantly.
+            List<string> AddOrCreate(List<string> list, string item)
```
chrisdunelm: Make this a nested function within `ProcessSentinelValues`, and you don't need to pass `list` as a parameter ;)

PR #1561, opened 2017-10-18
Allow unauthenticated access for StorageClient

jskeet:
This makes it simple to work with public datasets.

(I won't be merging this immediately; I want it to get wider review first.)

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UnauthenticatedAccessTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UnauthenticatedAccessTest.cs
@@ -1,0 +1,67 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UnauthenticatedAccessTest.cs
```diff
@@ -0,0 +1,67 @@
...
+        private const string LandsatObject = LandsatPrefix + "LC80440342016259LGN00_MTL.txt";
+
+        private readonly StorageFixture _fixture;
+
+        public UnauthenticatedAccessTest(StorageFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public void DownloadPublicData()
+        {
+            var client = StorageClient.CreateUnauthenticated();
```
frankyn: @jskeet by creating the client library this way will users be able to perform other operations if the appropriate IAM roles are given to "allUsers"? For example, upload a file if the bucket is writable for "allUsers" -- I hope no one uses a bucket this way, but wondering if it's possible.

jskeet: Users can *try* to do anything with such a client - whether they succeed will depend on the IAM configuration. I haven't tried creating a bucket that anyone can write to, but I *suspect* it would work fine. (That would certainly be consistent.)

jskeet:
Thanks Frank - merging.

PR #1562, opened 2017-10-18
Sample of query cursors using DocumentReference values

jskeet:
(This isn't to be merged yet; it's to make it easier to discuss the correct behaviour.)


jskeet:
Have decided I'd like to merge this if possible, partly to stop things getting out of sync, and partly as it documents current behaviour.

PR #1563, opened 2017-10-18
Prohibit array-in-array values

jskeet:


jskeet:
Yup, it's in the [proto](https://github.com/googleapis/googleapis/blob/ffe0aaf5fb0229072ffc3a47e075a8ab7980fc26/google/firestore/v1beta1/document.proto#L123):

```text
// An array value.
//
// Cannot contain another array value.
ArrayValue array_value = 9;
```

PR #1565, opened 2017-10-19
Mutations factory methods and MutateRow API

evildour:
TODO: Snippets

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
@@ -1,0 +1,140 @@
+﻿// Copyright 2017, Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax.Grpc;
+using Google.Apis.Auth.OAuth2;
+using Google.Cloud.Bigtable.Admin.V2;
+using Grpc.Core;
+using System;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Bigtable.V2.IntegrationTests
+{
+    [CollectionDefinition(nameof(BigtableFixture))]
+    public class BigtableFixture : IDisposable, ICollectionFixture<BigtableFixture>
+    {
+        private const string EmulatorEnvironmentVariable = "BIGTABLE_EMULATOR_HOST";
+        private const string TestProjectEnvironmentVariable = "TEST_PROJECT";
+
+        public const string ColumnFamily1 = "cf1";
+        public const string OtherColumnFamily = "test_data";
+
+        public ProjectName ProjectName { get; private set; }
+        public InstanceName DefaultInstanceName { get; private set; }
+        public TableName DefaultTableName { get; private set; }
+
+        public BigtableInstanceAdminClient DefaultInstanceAdminClient { get; private set; }
+        public BigtableTableAdminClient DefaultTableAdminClient { get; private set; }
+        public BigtableClient DefaultTableClient { get; private set; }
+
+        public Channel EmulatorChannel { get; }
+
+        public BigtableFixture()
+        {
+            string emulatorHost = Environment.GetEnvironmentVariable(EmulatorEnvironmentVariable);
+
+            string projectId;
+            if (!string.IsNullOrEmpty(emulatorHost))
+            {
+                projectId = "emulator-test-project";
+                EmulatorChannel = new Channel(emulatorHost, ChannelCredentials.Insecure);
+            }
+            else
+            {
+                projectId = Environment.GetEnvironmentVariable(TestProjectEnvironmentVariable);
+                if (string.IsNullOrEmpty(projectId))
+                {
+                    throw new InvalidOperationException(
+                        $"Please set either the {EmulatorEnvironmentVariable} or {TestProjectEnvironmentVariable} environment variable before running tests");
+                }
+            }
+
+            ProjectName = new ProjectName(projectId);
+
+            Task.Run(InitBigtableInstanceAndTable).Wait();
+        }
+
+        private async Task InitBigtableInstanceAndTable()
+        {
+            DefaultInstanceName = new InstanceName(ProjectName.ProjectId, "i" + Guid.NewGuid().ToString("N"));
+
+            if (EmulatorChannel == null)
+            {
+                DefaultInstanceAdminClient = BigtableInstanceAdminClient.Create();
+                DefaultTableAdminClient = BigtableTableAdminClient.Create();
+                DefaultTableClient = BigtableClient.Create();
+
+                // Cluster ID must be max 30 characters
+                string clusterId = DefaultInstanceName.InstanceId.Substring(0, 22) + "-cluster";
+
+                var instanceOperation = await DefaultInstanceAdminClient.CreateInstanceAsync(new CreateInstanceRequest
+                {
+                    Parent = ProjectName.ToString(),
+                    InstanceId = DefaultInstanceName.InstanceId,
+                    Instance = new Instance
+                    {
+                        DisplayName = "Integration tests instance",
+                        Type = Instance.Types.Type.Development
+                    },
+                    Clusters =
+                    {
+                        {
+                            clusterId,
+                            new Cluster
+                            {
+                                ClusterName = new ClusterName(ProjectName.ProjectId, DefaultInstanceName.InstanceId, clusterId),
+                                DefaultStorageType = StorageType.Hdd,
+                                LocationAsLocationName = new LocationName(ProjectName.ProjectId, "us-central1-c")
+                            }
+                        }
+                    }
+                });
+                await instanceOperation.PollUntilCompletedAsync();
+            }
+            else
+            {
+                DefaultTableAdminClient = BigtableTableAdminClient.Create(EmulatorChannel);
+                DefaultTableClient = BigtableClient.Create(EmulatorChannel);
+            }
+            
+            DefaultTableName = new TableName(ProjectName.ProjectId, DefaultInstanceName.InstanceId, "default-table");
+            await DefaultTableAdminClient.CreateTableAsync(
+                DefaultInstanceName,
+                DefaultTableName.TableId,
+                new Table
+                {
+                    Granularity = Table.Types.TimestampGranularity.Millis,
+                    ColumnFamilies =
+                    {
+                        { ColumnFamily1, new ColumnFamily { GcRule = new GcRule { MaxNumVersions = 3 } } },
+                        { OtherColumnFamily, new ColumnFamily { GcRule = new GcRule { MaxNumVersions = 3 } } }
+                    }
+                });
+        }
+
+        public void Dispose()
+        {
+            if (DefaultTableAdminClient != null && DefaultTableName != null)
+            {
+                DefaultTableAdminClient.DeleteTable(new DeleteTableRequest { Name = DefaultTableName.ToString() });
+            }
+
+            if (DefaultInstanceAdminClient != null && DefaultInstanceName != null)
+            {
+                DefaultInstanceAdminClient.DeleteInstance(DefaultInstanceName);
+            }
+        }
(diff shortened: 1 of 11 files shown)


jskeet:
I suspect that starting a kloc review late on a Friday afternoon would be a bad idea - are you happy for me to wait until Monday?

evildour:
Sure, no problem. I just realized now I forgot to assign this, so it's my fault.

jskeet:
I wouldn't have had time earlier today anyway :)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            ProjectName = new ProjectName(projectId);
+
+            Task.Run(InitBigtableInstanceAndTable).Wait();
+        }
+
+        private async Task InitBigtableInstanceAndTable()
+        {
+            DefaultInstanceName = new InstanceName(ProjectName.ProjectId, "i" + Guid.NewGuid().ToString("N"));
```
jskeet: If we're not using the emulator, we won't be creating the instance as far as I can see. How does that work?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowTest.cs
```diff
@@ -0,0 +1,359 @@
...
+            await VerifySingleRowAsync(
+                client,
+                tableName,
+                rowKey,
+                BigtableFixture.ColumnFamily1,
+                "column_name",
+                "test12345",
+                new BigtableVersion(1000));
+
+            // Cleanup
```
jskeet: Do we rely on this? I generally do all the clean-up when the fixture is disposed. If there's a good reason to clean up here for other tests, maybe add a comment explaining why?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowTest.cs
```diff
@@ -0,0 +1,359 @@
...
+
+        public MutateRowTest(BigtableFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public async Task AddNewRow()
+        {
+            var tableName = _fixture.DefaultTableName;
+            var client = _fixture.DefaultTableClient;
+            BigtableByteString rowKey = Guid.NewGuid().ToString();
+            client.MutateRow(
```
jskeet: This seems to be testing SetCell rather than anything explicitly adding a new row. Is it really "setting a cell in a non-existent row creates the row"? Would it be worth having a test for "setting a cell in an existing row"?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutationsTest.cs
```diff
@@ -0,0 +1,106 @@
...
+        public void DeleteFromFamilyInvalidFamilyName(string familyName)
+        {
+            Assert.Throws<ArgumentException>(() => Mutations.DeleteFromFamily(familyName));
+        }
+
+        [Fact]
+        public void DeleteFromRow()
+        {
+            var mutation = Mutations.DeleteFromRow();
+            Assert.NotNull(mutation.DeleteFromRow);
```
jskeet: Just to check, is this actually deleting the whole row? Why is it DeleteFromRow rather than DeleteRow?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutationsTest.cs
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        [Theory]
+        [MemberData(nameof(ValidFamilyNames))]
+        public void SetCell(string familyName)
+        {
+            var mutation =
+                Mutations.SetCell(familyName, "cq1", "value", new BigtableVersion(3));
```
jskeet: Are there other types you can set the cell value to?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -13,9 +13,184 @@
...
+        /// <remarks>
+        /// <para>
+        /// Note that string is implicitly convertible to <see cref="BigtableByteString"/>, so <paramref name="rowKey"/> can
+        /// be specified using a string as well and its UTF-8 representations will be used.
+        /// </para>
+        /// <para>
+        /// This method simply delegates to <see cref="MutateRow(MutateRowRequest, CallSettings)"/>.
+        /// </para>
+        /// </remarks>
+        /// <param name="tableName">
```
jskeet: Specify which values can't be null, and validate them?

jskeet: (And for simplicity, maybe extract the MutateRowRequest construction to a method?)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutationsTest.cs
```diff
@@ -0,0 +1,106 @@
...
+        [Theory]
+        [MemberData(nameof(ValidFamilyNames))]
+        public void DeleteFromColumn(string familyName)
+        {
+            var mutation =
+                Mutations.DeleteFromColumn(familyName, "cq1", new BigtableVersionRange(1, 2));
+            Assert.NotNull(mutation.DeleteFromColumn);
+            Assert.Equal(familyName, mutation.DeleteFromColumn.FamilyName);
+            Assert.Equal("cq1", mutation.DeleteFromColumn.ColumnQualifier.ToStringUtf8());
+            Assert.Equal(1, mutation.DeleteFromColumn.TimeRange.StartTimestampMicros);
```
jskeet: I thought micros always had to be a multiple of 1000, as per another email thread? Maybe I'm confused.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
evildour: With the emulator, you can just interact with any instance name. Actually, I'm not sure if the instance name is simply ignored or if it keeps separate tables per instance name used. I'll test this out.

evildour: Oh sorry, misread your question. If we're not using the emulator (there is no channel for it), we DO create the instance. If we're using the emulator, we don't.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutationsTest.cs
evildour: I just realized that as I was writing these integration tests and then started that thread (was there another one I forgot about)? Yes, I need to fix these tests and the `BigtableVersion` class as well so the multiple-of-1000s requirement is hidden for people wanting to use a custom versioning scheme. I was thinking of having the `BigtableVersion(long)` constructor interpret the param as millis, not micros, and having `BigtableVersion.Value` return millis, but still store micros internally. I'm a little worried about new granularities getting added in the future though and that this locks us into that granularity. But perhaps in the future if that were to ever happen these could be the defaults and we could add a constructor with an extra parameter for the granularity and a `GetValue(...)` method taking a granularity. I was going to do this all in another PR though. Thoughts?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutationsTest.cs
evildour: Yes, that's correct. Not sure about the _why_ of the naming. I was trying to keep the [API naming](https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#google.bigtable.v2.Mutation), but if you think it would be clearer as `DeleteRow`, I guess that's not too far from the "normal" naming, so it will still be easily discoverable. Still think it should change?

jskeet: No, I think it's probably best to keep it as it is.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
jskeet: Ah, with you - yes, I'd misread the condition.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutationsTest.cs
jskeet: Eek - yes, it's awkward. Best to be as granularity-neutral as possible, I think. I suggest we tackle this in a separate PR, where the effects will be clearer.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowTest.cs
evildour: Yes, it is testing to see if it creates a row. And yes, you're right that I'm missing the existing row case. Added.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowTest.cs
evildour: Not needed. You're right. removed.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutationsTest.cs
evildour: Yes, an arbitrary byte string and an 8-byte big-endian int64. Added tests for these as well.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Done and done.

evildour:
Thanks. PTAL.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
```diff
@@ -0,0 +1,85 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Bigtable.V2.Tests
+{
+    public class BigtableClientTest
+    {
+        [Fact]
```
jskeet: *Could* refactor this to a `ValidateArguments<TException>` method that makes the four calls. Not bothered though.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -182,13 +173,30 @@ public partial class BigtableClient
...
-                new MutateRowRequest
-                {
-                    TableName = tableName.ToString(),
-                    RowKey = rowKey,
-                    Mutations = { mutations }
-                },
+                CreateMutateRowRequest(tableName, rowKey, mutations),
                 CallSettings.FromCancellationToken(cancellationToken));
+
+        private static MutateRowRequest CreateMutateRowRequest(
```
jskeet: Is it okay for mutations to be empty?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
evildour: Good idea. Done.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Nope. Adding check and test.

PR #1573, opened 2017-10-23
Update API generation script (and corollaries)

jskeet:
Definitely review one commit at a time :)


jskeet:
Will look into the build failure. Maybe the protoc call isn't working as I expected.

jskeet:
Fixed previous problem, and added another few commits. I've re-enabled generation of all APIs other than the one responsible for Google.Cloud.Language.V1.Experimental (as fiddling with that manually is a pain).

jskeet:
Ooh - breakage in Video Intelligence. Looking at it...

jskeet:
Hooray for green-ness - PTAL.

apis/Google.Cloud.Vision.V1/Google.Cloud.Vision.V1.SmokeTests/ImageAnnotatorSmokeTest.g.cs
```diff
@@ -0,0 +1,68 @@
...
+            // Create client
+            ImageAnnotatorClient client = ImageAnnotatorClient.Create();
+
+            // Initialize request argument(s)
+            IEnumerable<AnnotateImageRequest> requests = new[]
+            {
+                new AnnotateImageRequest
+                {
+                    Image = new Image
+                            {
```
chrisdunelm: I don't like this formatting. At some point I'll see if I can persuade the codegen to do this better.

jskeet: Ooh, yes - I hadn't noticed that. At least it's a smoke test rather than a snippet...

Issue #1569, opened 2017-10-20
Google.Cloud.Diagnostics.* Tests are failing with the upgrade to xUnit 2.3.0


iantalarico:
apis\Google.Cloud.Diagnostics.Common\Google.Cloud.Diagnostics.Common.Tests> dotnet test

These tests will fail for netcoreapp 1.0.  netcoreapp 2.0 and netdesktop 4.6 pass.
Google.Cloud.Diagnostics.Common.Tests.ErrorReportingContextExceptionLoggerTest.Log*

Downgrade xUnit to "2.3.0-beta5-build3769" the above tests will all pass

benwulfe:
I had to upgrade to the full 2.3 release to get .net core 2.0 tests working.  Can we do something other than downgrade back to a beta?

iantalarico:
We need the 2.3.0 for 2.0.  I don't want to downgrade that.

It's failing on jenkins for the some of the diagnostics tests.  

I need to dig into this more.  It seems like an odd failure

benwulfe:
FWIW, I can still run .netcore 1.0 tests on my box.  Maybe its as simple as updating the Jenkins machines?

iantalarico:
@benwulfe I just emailed you the failure.  The Jenkins machines are able to run .netcore 1.0 and 2.0.

iantalarico:
Test run for <Google.Cloud.Diagnostics.AspNetCore.IntegrationTests.dll**(.NETCoreApp,Version=v1.0)**>
Microsoft (R) Test Execution Command Line Tool Version 15.3.0-preview-20170628-02
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
[xUnit.net 00:00:01.1054007]   Discovering: Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
[xUnit.net 00:00:01.2939945]   Discovered:  Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
[xUnit.net 00:00:01.3010773]   Starting:    Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
...
Failed   Google.Cloud.Diagnostics.AspNetCore.IntegrationTests.ErrorReportingTest.LogsException
...
Failed   Google.Cloud.Diagnostics.AspNetCore.IntegrationTests.ErrorReportingTest.LogsMultipleExceptions
...
[...]

PR #1575, opened 2017-10-23
Update the DebugType for the ASP.NET Core integration tests.

iantalarico:
I wasn't able to find good documentation on the differences between the types on each runtime but this configuration works.

This may also be an issue with xunit (#1569) still but cannot tell as 2.0.0 support is still new.


benwulfe:
maybe this is related to:
https://github.com/dotnet/sdk/issues/1238

I wonder if this warrants something in our documentation about when the stacktrace info will be there and what they need to do if its not there and they want it...?

iantalarico:
Yea I need to figure out if this is a xUnit issue or something with netcoreapp2.0.

I opened #1577 to track this.

PR #1578, opened 2017-10-23
Make BigtableVersion use milliseconds

evildour:


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableVersionTest.cs
```diff
@@ -39,28 +39,30 @@ public void Comparisons()
...
         [Fact]
-        public void MicrosScaleFromDateTime()
+        public void ScaleFromDateTime()
         {
             var utcNow = DateTime.UtcNow;
             var v1 = new BigtableVersion(utcNow);
             var v2 = new BigtableVersion(utcNow + TimeSpan.FromSeconds(1));
             var diff = v2.Value - v1.Value;
+            Assert.Equal(1_000, diff);
+            diff = v2.Micros - v1.Micros;
```
jskeet: I think this chunk of code would be clearer as

```csharp
Assert.Equal(1_000, v2.Value - v1.Value);
Assert.Equal(1_000_000, v2.Micros - v1.Micros);
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersion.cs
```diff
@@ -23,74 +23,93 @@ namespace Google.Cloud.Bigtable.V2
     /// <remarks>
     /// <para>
     /// Note: when using ReadModifyWriteRow, modified columns automatically use a server version, which
```
jskeet: I think it's probably worth explaining why both millis and micros are referred to here at all. I'm certainly confused about it, having only paid half attention to what's going on.

Is micros the raw underlying format, but the server only ever sets values in millis??

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableVersionTest.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableVersion.cs
evildour: That's correct. Ok, I'll update all these comments.

evildour: Actually, I think I'll make another note preceding this one to explain the whole thing.

evildour:
Thanks Jon! PTAL

PR #1582, opened 2017-10-25
Prevent optional parameter analyzer suggesting unassigned locals

evildour:


tools/Google.Cloud.Tools.Analyzers/InternalOptionalParametersRequiredAnalyzer.cs
```diff
@@ -81,12 +84,23 @@ private static void AnalyzeInvocation(OperationAnalysisContext context)
...
+
+            bool ShouldAnalyzeArgument(IArgument arg) =>
+                arg.ArgumentKind == ArgumentKind.DefaultValue && arg.Parameter != null;
         }
 
-        internal static IEnumerable<ISymbol> GetVariablesInScope(SyntaxNode syntaxNode, SemanticModel semanticModel) =>
-            from symbol in semanticModel.LookupSymbols(syntaxNode.SpanStart)
-            where symbol.Kind == SymbolKind.Local || symbol.Kind == SymbolKind.Parameter
-            select symbol;
+        internal static IEnumerable<ISymbol> GetVariablesInScope(SyntaxNode syntaxNode, SemanticModel semanticModel)
```
jskeet: Possibly rename to `GetUsableVariables` or similar? A variable is in scope from the start of the block that it's declared in, even before its declaration.

This is possibly the nit-pickiest code review comment I've ever made ;)

evildour: Nope. Makes sense. It _was_ getting variables in scope previously, and that was the problem :)

PR #1585, opened 2017-10-25
RowFilter factory methods and CheckAndMutateRow support

evildour:
Probably easiest to review one commit at a time.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
@@ -1,0 +1,171 @@
+﻿// Copyright 2017 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Bigtable.V2.IntegrationTests
+{
+    internal static class BigtableAssert
+    {
+        public static async Task HasNoValuesAsync(
+            BigtableClient client,
+            TableName tableName,
+            BigtableByteString rowKey)
+        {
+            // TODO: Use cleaner API when available
+            var response = client.ReadRows(new ReadRowsRequest
+            {
+                TableName = tableName.ToString(),
+                Rows = new RowSet { RowKeys = { rowKey } }
+            });
+
+            while (await response.ResponseStream.MoveNext(default))
+            {
+                var current = response.ResponseStream.Current;
+                foreach (var chunk in current.Chunks)
+                {
+                    Assert.True(false, $"No chunks expected for the row at {rowKey.Value.ToStringUtf8()}");
+                }
+            }
+        }
+
+        public static async Task HasSingleValueAsync(
+            BigtableClient client,
+            TableName tableName,
+            BigtableByteString rowKey,
+            string familyName,
+            BigtableByteString columnQualifier,
+            BigtableByteString value,
+            BigtableVersion? version = null)
+        {
+            // TODO: Use cleaner API when available
+            var response = client.ReadRows(new ReadRowsRequest
+            {
+                TableName = tableName.ToString(),
+                Rows = new RowSet { RowKeys = { rowKey } }
+            });
+
+            bool had_value = false;
+            while (await response.ResponseStream.MoveNext(default))
+            {
+                var current = response.ResponseStream.Current;
+                foreach (var chunk in current.Chunks)
+                {
+                    Assert.Equal(rowKey.Value, chunk.RowKey);
+                    Assert.Equal(familyName, chunk.FamilyName);
+                    Assert.Equal(columnQualifier.Value, chunk.Qualifier);
+                    Assert.Equal(value.Value, chunk.Value);
+                    if (version != null)
+                    {
+                        Assert.Equal(version.Value.Micros, chunk.TimestampMicros);
+                    }
+                    had_value = true;
+                }
+            }
+            Assert.True(had_value, "The value was not present.");
+        }
+
+        public static async Task DoesNotHaveValueAsync(
+            BigtableClient client,
+            TableName tableName,
+            BigtableByteString rowKey,
+            string familyName,
+            BigtableByteString columnQualifier,
+            BigtableByteString value,
+            BigtableVersion? version = null)
+        {
+            bool has_value = await HasValueHelperAsync(
+                client,
+                tableName,
+                rowKey,
+                familyName,
+                columnQualifier,
+                value,
+                version);
+            Assert.False(has_value);
+        }
+
+        public static async Task HasValueAsync(
+            BigtableClient client,
+            TableName tableName,
+            BigtableByteString rowKey,
+            string familyName,
+            BigtableByteString columnQualifier,
+            BigtableByteString value,
+            BigtableVersion? version = null)
+        {
+            bool has_value = await HasValueHelperAsync(
+                client,
+                tableName,
+                rowKey,
+                familyName,
+                columnQualifier,
+                value,
+                version);
+            Assert.True(has_value);
+        }
+
+        private static async Task<bool> HasValueHelperAsync(
+            BigtableClient client,
+            TableName tableName,
+            BigtableByteString rowKey,
+            string familyName,
+            BigtableByteString columnQualifier,
+            BigtableByteString value,
+            BigtableVersion? version)
+        {
+            // TODO: Use cleaner API when available
+            var response = client.ReadRows(new ReadRowsRequest
+            {
+                TableName = tableName.ToString(),
+                Rows = new RowSet { RowKeys = { rowKey } }
+            });
+
+            bool had_value = false;
+            while (await response.ResponseStream.MoveNext(default))
+            {
+                var current = response.ResponseStream.Current;
+                BigtableByteString currentRowKey;
+                string currentColumnFamily = null;
+                foreach (var chunk in current.Chunks)
+                {
+                    if (!chunk.RowKey.IsEmpty)
+                    {
+                        currentRowKey = chunk.RowKey;
+                    }
+                    if (chunk.FamilyName != null)
+                    {
+                        currentColumnFamily = chunk.FamilyName;
+                    }
+                    Assert.Equal(rowKey, currentRowKey);
+                    if (familyName == currentColumnFamily && columnQualifier.Value == chunk.Qualifier)
+                    {
+                        Assert.Equal(value.Value, chunk.Value);
+                        if (version != null)
+                        {
+                            Assert.Equal(version.Value.Micros, chunk.TimestampMicros);
+                        }
+                        had_value = true;
+                    }
+                }
+            }
+
+            return had_value;
+        }
+    }
+}
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
@@ -37,10 +37,7 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
             while (await response.ResponseStream.MoveNext(default))
             {
                 var current = response.ResponseStream.Current;
-                foreach (var chunk in current.Chunks)
-                {
-                    Assert.True(false, $"No chunks expected for the row at {rowKey.Value.ToStringUtf8()}");
-                }
+                Assert.Empty(current.Chunks);
             }
         }
 
@@ -60,7 +57,7 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
                 Rows = new RowSet { RowKeys = { rowKey } }
             });
 
-            bool had_value = false;
+            bool hadValue = false;
             while (await response.ResponseStream.MoveNext(default))
             {
                 var current = response.ResponseStream.Current;
@@ -74,10 +71,11 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
                     {
                         Assert.Equal(version.Value.Micros, chunk.TimestampMicros);
                     }
-                    had_value = true;
+                    Assert.False(hadValue);
+                    hadValue = true;
                 }
             }
-            Assert.True(had_value, "The value was not present.");
+            Assert.True(hadValue, "The value was not present.");
         }
 
         public static async Task DoesNotHaveValueAsync(
@@ -89,7 +87,7 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
             BigtableByteString value,
             BigtableVersion? version = null)
         {
-            bool has_value = await HasValueHelperAsync(
+            bool hasValue = await HasValueHelperAsync(
                 client,
                 tableName,
                 rowKey,
@@ -97,7 +95,7 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
                 columnQualifier,
                 value,
                 version);
-            Assert.False(has_value);
+            Assert.False(hasValue);
         }
 
         public static async Task HasValueAsync(
@@ -109,7 +107,7 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
             BigtableByteString value,
             BigtableVersion? version = null)
         {
-            bool has_value = await HasValueHelperAsync(
+            bool hasValue = await HasValueHelperAsync(
                 client,
                 tableName,
                 rowKey,
@@ -117,7 +115,7 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
                 columnQualifier,
                 value,
                 version);
-            Assert.True(has_value);
+            Assert.True(hasValue);
         }
 
         private static async Task<bool> HasValueHelperAsync(
@@ -136,7 +134,7 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
                 Rows = new RowSet { RowKeys = { rowKey } }
             });
 
-            bool had_value = false;
+            bool hadValue = false;
             while (await response.ResponseStream.MoveNext(default))
             {
                 var current = response.ResponseStream.Current;
@@ -160,12 +158,12 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
                         {
                             Assert.Equal(version.Value.Micros, chunk.TimestampMicros);
                         }
-                        had_value = true;
+                        hadValue = true;
                     }
                 }
             }
 
-            return had_value;
+            return hadValue;
         }
     }
 }
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
@@ -28,6 +28,113 @@ namespace Google.Cloud.Bigtable.V2.Tests
         private class TestBigtableClient : BigtableClient { }
 
         [Fact]
+        public async Task CheckAndMutateRow_Validate_TableName()
+        {
+            var client = new TestBigtableClient();
+            await CheckAndMutateRow_ValidateArguments<ArgumentNullException>(
+                null,
+                "abc",
+                RowFilters.PassAllFilter(),
+                new[] { Mutations.DeleteFromRow() },
+                new[] { Mutations.DeleteFromRow() });
+        }
+
+        [Fact]
+        public async Task CheckAndMutateRow_Validate_RowKey()
+        {
+            var client = new TestBigtableClient();
+            var tableName = new TableName("project", "instance", "table");
+            await CheckAndMutateRow_ValidateArguments<ArgumentException>(
+                tableName,
+                "",
+                RowFilters.PassAllFilter(),
+                new[] { Mutations.DeleteFromRow() },
+                new[] { Mutations.DeleteFromRow() });
+            await CheckAndMutateRow_ValidateArguments<ArgumentException>(
+                tableName,
+                new byte[0],
+                RowFilters.PassAllFilter(),
+                new[] { Mutations.DeleteFromRow() },
+                new[] { Mutations.DeleteFromRow() });
+            await CheckAndMutateRow_ValidateArguments<ArgumentException>(
+                tableName,
+                ByteString.Empty,
+                RowFilters.PassAllFilter(),
+                new[] { Mutations.DeleteFromRow() },
+                new[] { Mutations.DeleteFromRow() });
+            await CheckAndMutateRow_ValidateArguments<ArgumentException>(
+                tableName,
+                default,
+                RowFilters.PassAllFilter(),
+                new[] { Mutations.DeleteFromRow() },
+                new[] { Mutations.DeleteFromRow() });
+        }
+
+        [Fact]
+        public async Task CheckAndMutateRow_Validate_Mutations()
+        {
+            var client = new TestBigtableClient();
+            var tableName = new TableName("project", "instance", "table");
+            await CheckAndMutateRow_ValidateArguments<ArgumentException>(
+                tableName,
+                "abc",
+                RowFilters.PassAllFilter(),
+                null,
+                null);
+            await CheckAndMutateRow_ValidateArguments<ArgumentException>(
+                tableName,
+                "abc",
+                RowFilters.PassAllFilter(),
+                new Mutation[0],
+                new Mutation[0]);
+            await CheckAndMutateRow_ValidateArguments<ArgumentException>(
+                tableName,
+                "abc",
+                RowFilters.PassAllFilter(),
+                new Mutation[] { null },
+                new[] { Mutations.DeleteFromRow() },
+                callTrueMutationsOnlyOverload: false);
+            await CheckAndMutateRow_ValidateArguments<ArgumentException>(
+                tableName,
+                "abc",
+                RowFilters.PassAllFilter(),
+                new[] { Mutations.DeleteFromRow() },
+                new Mutation[] { null },
+                callTrueMutationsOnlyOverload: false);
+        }
+
+        private async Task CheckAndMutateRow_ValidateArguments<TException>(
+            TableName tableName,
+            BigtableByteString rowKey,
+            RowFilter predicateFilter,
+            IEnumerable<Mutation> trueMutations,
+            IEnumerable<Mutation> falseMutations,
+            bool callTrueMutationsOnlyOverload = true)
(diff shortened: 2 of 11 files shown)


jskeet:
I'm on vacation tomorrow - will review it first thing Friday.

evildour:
No problem. Enjoy your vacation

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,380 @@
...
+        /// Note that string is implicitly convertible to <see cref="BigtableByteString"/>,
+        /// so <paramref name="regex"/> can be specified using a string as well and its UTF-8
+        /// representations will be used.
+        /// </para>
+        /// </remarks>
+        /// <param name="regex">
+        /// The RE2 regex used to match column qualifiers.
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter ColumnQualifierRegex(BigtableByteString regex) =>
```
jskeet: Out of interest, what's the benefit in this accepting `BigtableByteString` rather than just string? When do we expect anyone to pass anything *other* than a `string` in? My guess is that using a value from one call as a filter input in another, but I thought I'd check..

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,380 @@
...
+        /// returns any results. If null, no results will be returned in the true case.
+        /// </param>
+        /// <param name="falseFilter">
+        /// The filter to apply to the input row if <paramref name="predicateFilter"/>
+        /// does not return any results. If null, no results will be returned in the
+        /// false case.
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter Condition(
+            RowFilter predicateFilter, RowFilter trueFilter, RowFilter falseFilter) =>
```
jskeet: Validate that at least one of trueFilter or falseFilter is non-null?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,380 @@
...
+                    // TODO: Verify this on server. Might be allowed for (row contains any cells).
+                    PredicateFilter = GaxPreconditions.CheckNotNull(predicateFilter, nameof(predicateFilter)),
+                    TrueFilter = trueFilter,
+                    FalseFilter = falseFilter
+                }
+            };
+
+        /// <summary>
+        /// Creates a new <see cref="RowFilter"/> instance which matches only cells
+        /// from columns whose families satisfy the given RE2 regex.
```
jskeet: Is "RE2" a well-understood term within Bigtable? Presumably it's as per https://github.com/google/re2

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,380 @@
...
+        /// if it is not being used as a literal. Note that, since column families
+        /// cannot contain the new line character `\n`, it is sufficient to use `.`
+        /// as a full wildcard when matching column family names.
+        /// </para>
+        /// </remarks>
+        /// <param name="regex">
+        /// The RE2 regex used to match column families. Must not be null.
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter FamilyNameRegex(string regex) =>
```
jskeet: It's odd that this accepts a string but `ColumnQualifierRegex` accepts a `BigtableByteString`. Presumably that's just the way the API is specified, but we should *consider* removing the inconsistency at this level. I'm slightly torn.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,380 @@
...
+            new RowFilter { FamilyNameRegexFilter = GaxPreconditions.CheckNotNull(regex, nameof(regex)) };
+
+        // TODO: Copy docs from Interleave.Filters, but only once the formatting issues are resolved.
```
jskeet: Definitely - I've no idea what this means at the moment compared with Chain :)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,380 @@
...
+        /// </para>
+        /// </remarks>
+        /// <param name="regex">
+        /// The RE2 regex used to match row keys.
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter RowKeyRegex(BigtableByteString regex) =>
+            new RowFilter { RowKeyRegexFilter = regex };
+
+        // TODO: Seems like this is 0-1. Verify and document it.
```
jskeet: And validate input?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,380 @@
...
+        /// <summary>
+        /// Creates a new <see cref="RowFilter"/> instance which matches only cells
+        /// with versions within the given range.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Note: version values are stored on the server as if they are microseconds since the Unix epoch.
+        /// However, the server only supports millisecond granularity, so the server only allows microseconds
+        /// in multiples of 1,000. <see cref="BigtableVersion"/> attempts to hide this complexity by exposing
+        /// it's underlying <see cref="BigtableVersion.Value"/> in terms of milliseconds, so if desired, a
```
jskeet: Nit: it's => its

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -0,0 +1,380 @@
...
+        /// with versions within the given range.
+        /// </summary>
+        /// <param name="range">
+        /// The range of versions from which cells should be matched. Must not be null.
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter VersionRange(BigtableVersionRange range) =>
+            new RowFilter { TimestampRangeFilter = GaxPreconditions.CheckNotNull(range, nameof(range)).ToTimestampRange() };
+
+        /// <summary>
```
jskeet: Possibly remove this overload for now and replace it with a TODO, to match the others ranges?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RowFiltersTest.cs
```diff
@@ -0,0 +1,245 @@
...
+
+namespace Google.Cloud.Bigtable.V2.Tests
+{
+    public class RowFiltersTest
+    {
+        [Fact]
+        public void BlockAllFilter()
+        {
+            var filter = RowFilters.BlockAllFilter();
+            Assert.True(filter.BlockAllFilter);
```
jskeet: I'm in two minds about this - it's not testing as much as `Assert.Equal(new RowFilter { BlockAllFilter = true }` but then that really would be a copy of the production code. Ditto a lot of the rest. This is the problem with testing mostly-trivial (but still useful, of course) code.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -24,6 +23,232 @@ namespace Google.Cloud.Bigtable.V2
     public partial class BigtableClient
     {
         /// <summary>
+        /// Mutates a row atomically based on the output of a predicate Reader filter.
```
jskeet: If the row doesn't exist at all before and falseMutations contains some mutations, is the row created with those mutations?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -24,6 +23,232 @@ namespace Google.Cloud.Bigtable.V2
...
+        /// <param name="cancellationToken">A cancellation token to monitor for the asynchronous operation.</param>
+        /// <returns>
+        /// The response from trying to mutate the row.
+        /// </returns>
+        public virtual Task<CheckAndMutateRowResponse> CheckAndMutateRowAsync(
+            TableName tableName,
+            BigtableByteString rowKey,
+            RowFilter predicateFilter,
+            IEnumerable<Mutation> trueMutations,
+            CancellationToken cancellationToken = default) =>
```
jskeet: It feels slightly odd that whether or not you can specify a cancellation token or a call settings depends on whether you're providing `falseMutations`.

I can see the following potential overloads (and there could be more that I haven't thought about):

```csharo
// No settings/cancellation token due to params
CAMRA(tableName, rowKey, predicateFilter, params trueMutations)
CAMRA(tableName, rowKey, predicateFilter, trueMutations, token)
CAMRA(tableName, rowKey, predicateFilter, trueMutations, callSettings = null)
CAMRA(tableName, rowKey, predicateFilter, trueMutations, falseMutations, token)
CAMRA(tableName, rowKey, predicateFilter, trueMutations, falseMutations, callSettings = null)
```

That creates an ambiguity if anyone calls `CAMRA(tableName, rowKey, predicateFilter, trueMutations, null)` as the null could be call settings or `falseMutations`.

Do we know how common "true mutations only" is? Does it deserve its own method name to avoid overloading problems?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -24,6 +23,232 @@ namespace Google.Cloud.Bigtable.V2
...
+        /// The key of the row to which the conditional mutation should be applied.
+        /// Must not be empty.
+        /// </param>
+        /// <param name="predicateFilter">
+        /// The filter to be applied to the contents of the specified row. Depending
+        /// on whether or not any results are yielded, either <paramref name="trueMutations"/> or
+        /// <paramref name="falseMutations"/> will be executed. If null, checks that the row contains
+        /// any values at all.
+        /// </param>
+        /// <param name="trueMutations">
```
jskeet: Document that true/false can't *both* be null/empty?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -24,6 +23,232 @@ namespace Google.Cloud.Bigtable.V2
...
+        /// <summary>
+        /// Asynchronously mutates a row atomically based on the output of a predicate Reader filter.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Note that string is implicitly convertible to <see cref="BigtableByteString"/>, so <paramref name="rowKey"/> can
+        /// be specified using a string as well and its UTF-8 representations will be used.
+        /// </para>
+        /// <para>
+        /// This method simply delegates to <see cref="CheckAndMutateRow(CheckAndMutateRowRequest, CallSettings)"/>.
```
jskeet: Async

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -24,6 +23,232 @@ namespace Google.Cloud.Bigtable.V2
...
```
jskeet: Async

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
```diff
@@ -28,6 +28,113 @@ public class BigtableClientTest
...
+            var client = new TestBigtableClient();
+            await CheckAndMutateRow_ValidateArguments<ArgumentNullException>(
+                null,
+                "abc",
+                RowFilters.PassAllFilter(),
+                new[] { Mutations.DeleteFromRow() },
+                new[] { Mutations.DeleteFromRow() });
+        }
+
+        [Fact]
```
jskeet: Could use a theory with a property returning the different invalid row keys?

See https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data.Tests/WriteBatchTest.SentinelValues.cs for some simple examples.

(Applies to various tests here.)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
```diff
@@ -28,6 +28,113 @@ public class BigtableClientTest
         private class TestBigtableClient : BigtableClient { }
 
         [Fact]
+        public async Task CheckAndMutateRow_Validate_TableName()
```
jskeet: Any unit tests for valid calls? Or is it sufficiently trivial to just delegate that to the integration tests?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
```diff
@@ -0,0 +1,171 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Bigtable.V2.IntegrationTests
+{
+    internal static class BigtableAssert
```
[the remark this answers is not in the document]
evildour: Oops. Somehow my 80% life start to bleed into my other life :)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
```diff
@@ -0,0 +1,171 @@
...
+    {
+        public static async Task HasNoValuesAsync(
+            BigtableClient client,
+            TableName tableName,
+            BigtableByteString rowKey)
+        {
+            // TODO: Use cleaner API when available
```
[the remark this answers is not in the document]
evildour: Perhaps, I will ignore on a case by case basis.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
```diff
@@ -0,0 +1,171 @@
...
+            var response = client.ReadRows(new ReadRowsRequest
+            {
+                TableName = tableName.ToString(),
+                Rows = new RowSet { RowKeys = { rowKey } }
+            });
+
+            while (await response.ResponseStream.MoveNext(default))
+            {
+                var current = response.ResponseStream.Current;
+                foreach (var chunk in current.Chunks)
```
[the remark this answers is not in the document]
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
```diff
@@ -0,0 +1,171 @@
...
+                {
+                    Assert.True(false, $"No chunks expected for the row at {rowKey.Value.ToStringUtf8()}");
+                }
+            }
+        }
+
+        public static async Task HasSingleValueAsync(
```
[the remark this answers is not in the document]
evildour: The class is named `...Assert`, so I think `BigtableAssert.AssertSingleValueAsync` is a bit redundant. I was going with something you could read as Bigtable**Assert** that it **Has** a single value (asynchronously), almost like `Assert.DoesNotContain` - normally you'd expect `DoesNotContain` to return a bool, but not in the context of assert.

I think I'll keep it as-is rather than going with a `Get...` alternative for 2 reasons:
* It doesn't make sense in the concept of the `...Assert` class, so it means adding a new place for this common code (not the end of the world though).
* It would be verifying everything specified (family name, column name, version) but then it's the callers responsibility to verify the value. Seems easier for the callers to just have it do everything.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
```diff
@@ -0,0 +1,171 @@
...
+            {
+                TableName = tableName.ToString(),
+                Rows = new RowSet { RowKeys = { rowKey } }
+            });
+
+            bool had_value = false;
+            while (await response.ResponseStream.MoveNext(default))
+            {
+                var current = response.ResponseStream.Current;
+                foreach (var chunk in current.Chunks)
```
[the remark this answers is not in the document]
evildour: Yes, no, yes :)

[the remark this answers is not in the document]
evildour: Plus this is only temporary until the proper read rows can be implemented, which will take those things into account.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableAssert.cs
```diff
@@ -0,0 +1,171 @@
...
+                {
+                    Assert.Equal(rowKey.Value, chunk.RowKey);
+                    Assert.Equal(familyName, chunk.FamilyName);
+                    Assert.Equal(columnQualifier.Value, chunk.Qualifier);
+                    Assert.Equal(value.Value, chunk.Value);
+                    if (version != null)
+                    {
+                        Assert.Equal(version.Value.Micros, chunk.TimestampMicros);
+                    }
+                    had_value = true;
```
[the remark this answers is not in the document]
evildour: Good call. Although this will only matter with the emulator. The production Bigtable (unlike the emulator) omits data that is the same from the previous chunk (row key or family name, for example). So if there were a 2nd value, the checks above would fail.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
evildour: Might be worthwhile just as a sanity check.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
evildour: Right now I'm not doing any validation of row keys other than checking for non-empty. The RPC docs don't mention any constraints on the row key for these requests, so I don't want to bake in any extra validation just incase the implementation changes.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RowFiltersTest.cs
evildour: Yeah, I agree. I figured rather than testing for a copy/pasted creation of what the implementation does, I'd rather test for the effect of those instances being created. Like you said, it's trivial and really just a smoke test to catch something that went terribly wrong.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Yes (see `CheckAndMutateRowTest.NullPredicate_MissingRow`). Do you think something should go in the remarks here?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: This line for the param points this out: `Must contain at least one entry if <paramref name="falseMutations"/> is null or empty"`. Similar for `falseMutations`. Do you think something else should be added to the remarks?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Not sure about usage. I'll ask internally. Good call about oddness. I guess I had these as mirrors to the synchronous params overloads, but kept the cancellation token because it is used for the async pattern, which meant they couldn't be params, so we lost the benefit there. What about just these two overloads:
```
CAMRA(tableName, rowKey, predicateFilter, params trueMutations)
CAMRA(tableName, rowKey, predicateFilter, trueMutations, falseMutations, callSettings = null)
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: This is to match the API specification (see column_qualifier_regex_filter [here](https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#google.bigtable.v2.RowFilter)). I assumed you could just specify an arbitrary byte string to find values that match that, but I guess it would always try to interpret it at RE2, so perhaps string is best.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
jskeet: That's probably a reasonable pair, yes. Let's at least start with that...

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: While it would be odd to not set either one, it isn't documented as not allowed. Again I'd like to err on the side of not imposing restrictions I assume make sense. But if you think it's best to do it anyway, I will.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: I'm assuming that's it, but I'll confirm and maybe add links in the remarks section.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: I'm asking about this internally.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
jskeet: I think I got confused between this and CreateCheckAndMutateRowRequest that checks that there's at least one mutation.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: You can think of Chain like and AND boolean operator, but the filters are ordered. Interleave can be thought of like OR, but if multiple filters match a cell, it will appear in the output once for each of those matching filters. Chain will only "output" a cell if it flows through each filter in the chain, in order. Interleave works in parallel and outputs each cell that passes each filter. There is more info about interleave here: https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#google.bigtable.v2.RowFilter.Interleave, but it just doesn't format correctly, so I need to figure out the best way to include it.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: Perhaps. Again, not sure if we should validate what isn't documented. I've updated the TODO.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: Fixed (in all 7 locations)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: Done, but I'm going to leave `TimestampRange` since that's the name used in the docs, even though it delegates to `VersionRange`.

evildour:
Thanks Jon! PTAL

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: Oh I see. Yes, that is documented that you need at least one true/false mutation.

jskeet:
I'm not going to be able to get to this again before Monday - and even then it'll be small amounts of time here and there.

I see what you mean about the Assert prefix given the class name - it was only when I saw it within the same class that it wasn't clear.

It looks like you've given good consideration to all the comments, so unless there are any specific changes you'd like me to look at carefully, I'm happy for you just to merge.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Oops, I made the comment and then forgot to actually make this change. Hold on I'll change this now.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: And yes, that's for the API is specified by the way.

evildour:
Ok thanks. I'm going to update some things when I hear back internally, so we can go over anything else that needs to be addressed at that point. I'll merge when green.

PR #1589, opened 2017-10-27
Update versions

jskeet:


jskeet:
I'm going to wait for green just in case there are weird dependency problems.

chrisdunelm:
Always a good plan :)

jskeet:
Hmm. We appear to have broken Translation somehow, but it's not immediately clear how.

Will investigate when I can.

PR #1590, opened 2017-10-27
Remove gRPC streaming workaround

jskeet:
This will be just one commit when #1589 is merged, so only bother looking at the last commit :)


jskeet:
Now down to the single commit. PTAL.

PR #1593, opened 2017-10-29
Update SimplePublisher.cs

mdietz94:
The maximum batching bytes should be 10 MB, as described in the Pub/Sub API.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SimplePublisher.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SimplePublisher.cs
@@ -206,7 +206,7 @@ namespace Google.Cloud.PubSub.V1
         /// <see cref="BatchingSettings.ElementCountThreshold"/> = 1,000;
         /// <see cref="BatchingSettings.ByteCountThreshold"/> = 9,500,000;
         /// </summary>
-        public static BatchingSettings ApiMaxBatchingSettings { get; } = new BatchingSettings(1000L, 9_500_000L, null);
+        public static BatchingSettings ApiMaxBatchingSettings { get; } = new BatchingSettings(1000L, 10_000_000L, null);
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SimplePublisher.cs
```diff
@@ -206,7 +206,7 @@ public FlowState(long elementCount, long byteCount)
         /// <see cref="BatchingSettings.ElementCountThreshold"/> = 1,000;
         /// <see cref="BatchingSettings.ByteCountThreshold"/> = 9,500,000;
         /// </summary>
-        public static BatchingSettings ApiMaxBatchingSettings { get; } = new BatchingSettings(1000L, 9_500_000L, null);
+        public static BatchingSettings ApiMaxBatchingSettings { get; } = new BatchingSettings(1000L, 10_000_000L, null);
```
chrisdunelm: This needs to be less than 10Mb, because the value here does not take into account overhead of putting all the messages into a proto.
That is, the size specified here only accounts for the byte-count of the messages themselves, not the enclosing proto(s).

The current value of 9.5Mb is probably too conservative, but it must be less than 10Mb.
Apologies that I didn't have a comment explaining this. I'll add one shortly.

mdietz94: The server enforces against a higher limit, in order for users to not have to worry about that overhead. If this limit is set to 10 MB, and a users topic name is the maximum topic name length, and there are 1000 messages, whose size summed totals 10 MB, the message will still be accepted. This was done for several reasons, one being the simplicity of using the API, and the other was to avoid from projects with long project or topic names from affecting what messages can be sent.

chrisdunelm:
@mdietz94 You are (unsurprisingly!) correct. 10,000,000 byte works fine :)
However, I've created PR #1621 to fix this, so I can add a specific test for this. Once that has been merged, I'll close this PR.

chrisdunelm:
#1621 is merged, so closing this.

PR #1596, opened 2017-10-29
Default literal analyzer

### LINKED ISSUE #1594: Check our uses of default literals with nullable types
Just found this:

https://github.com/dotnet/roslyn/issues/22578

If we're using `default` in any parameter lists for nullable value types, that could end up causing a problem.

Not sure of the best way to detect that - @evildour how simple would a Roslyn analyzer be to find that? Any alternatives you can think of? A quick text search for `? [a-zA-Z0-9]+ = default` might be a good first pass...

evildour:
Fixes #1594

Note that the issue only seems to affect value types that have a default value expressible as a constant, like 0 or false, so the analyzer only checks for these.

### DIFF
--- a/tools/Google.Cloud.Tools.Analyzers.Tests/DefaultLiteralTest.cs
+++ b/tools/Google.Cloud.Tools.Analyzers.Tests/DefaultLiteralTest.cs
@@ -102,18 +102,7 @@ public class A
         bool? x = default;
     }
 }";
-            var expected = CreateDiagnostic("bool?", 6, 19);
-            VerifyCSharpDiagnostic(test, expected);
-
-            var newSource = @"
-public class A
-{
-    public void B()
-    {
-        bool? x = default(bool?);
-    }
-}";
-            VerifyCSharpFix(test, newSource);
+            VerifyCSharpDiagnostic(test);
         }
 
         [Fact]
--- a/tools/Google.Cloud.Tools.Analyzers.Tests/DefaultLiteralTest.cs
+++ b/tools/Google.Cloud.Tools.Analyzers.Tests/DefaultLiteralTest.cs
@@ -1,0 +1,126 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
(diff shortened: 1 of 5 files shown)


tools/Google.Cloud.Tools.Analyzers.Tests/DefaultLiteralTest.cs
```diff
@@ -0,0 +1,126 @@
...
+using System.Collections.Generic;
+using System.Text;
+using TestHelper;
+using Xunit;
+
+namespace Google.Cloud.Tools.Analyzers.Tests
+{
+    public class DefaultLiteralTest : CodeFixVerifier
+    {
+        [Fact]
```
jskeet: Possibly add tests to verify it doesn't trigger for non-nullable versions?

evildour: Done

evildour:
I realized I could disable the analyzer automatically when the bug is fixed. Also, I added some tests for local declarations instead of parameter declarations. PTAL.

tools/Google.Cloud.Tools.Analyzers.Tests/DefaultLiteralTest.cs
```diff
@@ -78,6 +92,31 @@ public class A
...
+        {
+            var test = @"
+public class A
+{
+    public void B()
+    {
+        bool? x = default;
+    }
+}";
+            var expected = CreateDiagnostic("bool?", 6, 19);
```
jskeet: I don't think the bug affects local variables - at least as far as I can see. It looks like it's only parameters.

tools/Google.Cloud.Tools.Analyzers/DefaultLiteralAnalyzer.cs
```diff
@@ -56,6 +56,19 @@ private static void AnalyzeDefaultLiteral(SyntaxNodeAnalysisContext context)
                 if (underlyingType.IsPrimitive() ||
                     underlyingType.BaseType?.SpecialType == SpecialType.System_Enum)
                 {
+                    // Check to see if https://github.com/dotnet/roslyn/issues/22578 has been fixed so we don't warn unnecessarily on parameters.
```
jskeet: Nice :)

(Alternatively, maybe we should add a warning at that point that the analyzer could be removed - although perhaps the tests will then start failing anyway, because it won't have the expected diagnostic?)

tools/Google.Cloud.Tools.Analyzers.Tests/DefaultLiteralTest.cs
evildour: Yup, you're right. I just assumed it would. Fixed.

tools/Google.Cloud.Tools.Analyzers/DefaultLiteralAnalyzer.cs
evildour: Yeah, I think it's best to just rely on the tests failing to know whether to remove the analyzer.

PR #1597, opened 2017-10-30
Regenerate Bigtable.V2

evildour:


jskeet:
Just FYI, I'm about to board a plane. If this is pure regeneration, LGTM without looking :)

evildour:
Yes, it is. I was just going to merge when green because I assumed it would be fine.

PR #1603, opened 2017-10-31
Fix copyright notice analyzers to allow Google LLC

jskeet:
(Google Inc. is still permitted as well, and is the suggested fix if
there's no notice. If we decide that LLC is the preferred and only
format, we can change that later.)


jskeet:
This time, I'll wait for green before merging :)

evildour:
Don't merge yet :) Will comment in one minute

tools/Google.Cloud.Tools.Analyzers/MissingCopyrightNoticeAnalyzer.cs
```diff
@@ -127,7 +127,9 @@ private static bool RequiredCopyrightIsMissing(SyntaxNode rootNode)
 
         private static Regex CreateCopyrightRegex(string copyrightTemplate)
         {
-            var pattern = Regex.Escape(copyrightTemplate).Replace("____", @"\d{4}\,?");
+            var pattern = Regex.Escape(copyrightTemplate)
+                .Replace("____", @"\d{4}\,?")
+                .Replace(@"Google\ Inc\.", @"Google\ (Inc\.|LLC)");
```
evildour: Update the constants above to use 'LLC' instead and then update the replace here. The code fix provider uses the constants above to insert new copyright messages. You'll also need to update the unit tests as well.

jskeet: See the original PR comment - it's not 100% clear to me that LLC is the preferred licence to be used everywhere yet. It's used in the codegen, but that's all.

If you're reasonably confident that it should be the preferred approach, I'm happy to update the PR - I was trying to be minimal to start with (and to fix the build ASAP).

evildour: No, I was just assuming everything needed to change over. This is fine then. Apologies.

PR #1608, opened 2017-11-01
Add VideoIntelligence.V1 code

jskeet:
The service isn't active yet, but I want to get the code in ASAP


jskeet:
Service is now active, and snippet passes.

PR #1621, opened 2017-11-06
Increase pubsub batching to 10,000,000 bytes, and add test

chrisdunelm:


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
```diff
@@ -224,5 +224,39 @@ public SimplePubSubTest(PubsubFixture fixture)
...
+            var simplePublisher = await SimplePublisher.CreateAsync(topicName,
+                new SimplePublisher.ClientCreationSettings(clientCount: 1, publisherSettings: publisherSettings),
+                new SimplePublisher.Settings { BatchingSettings = batchingSettings }).ConfigureAwait(false);
+            var msgCount = SimplePublisher.ApiMaxBatchingSettings.ElementCountThreshold.Value;
+            var msgSize = SimplePublisher.ApiMaxBatchingSettings.ByteCountThreshold.Value / msgCount;
+            var rnd = new Random(1234);
+            var publishTasks = new Task[msgCount];
+            for (int i = 0; i < msgCount; i++)
+            {
+                var msg = new byte[msgSize];
```
jskeet: Just for my edification, really: this could be outside the loop, because `PublishAsync` will take a copy of the bytes, right?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
```diff
@@ -224,5 +224,39 @@ public SimplePubSubTest(PubsubFixture fixture)
...
+        {
+            // Test sending messages such that the maximum batch size is reached.
+            var topicId = _fixture.CreateTopicId();
+            // Create topic
+            var topicName = new TopicName(_fixture.ProjectId, topicId);
+            var publisher = await PublisherClient.CreateAsync().ConfigureAwait(false);
+            await publisher.CreateTopicAsync(topicName).ConfigureAwait(false);
+            // Create SimplePublisher, with no retry
+            var batchingSettings = new BatchingSettings(SimplePublisher.ApiMaxBatchingSettings.ElementCountThreshold,
+                SimplePublisher.ApiMaxBatchingSettings.ByteCountThreshold, TimeSpan.FromSeconds(4));
```
jskeet: Is 4 seconds long enough for 10MB?

chrisdunelm: Yes, 4 seconds is for locally filling the batch, not for sending over the network.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
chrisdunelm: I've put it within the loop to ensure the entire batch has no repetition, so cannot be compressed. I'm not sure if the server measures compressed or uncompressed size, but this way they'll always be ~equal.

jskeet: I was only suggesting (ish) moving the byte array allocation - not the call to `rand.Next()`. Basically this was me checking behaviour more than really suggesting a change. If you moved *just this line* before the loop, it would still be valid, right? (i.e., the Publish call copies the message data rather than maintaining an array reference.) If it wouldn't be valid to do that, we should look at API usability again.

chrisdunelm: Oh sorry, I misunderstood your point ;) Yes, the call to `PublishAsync` does indeed copy the bytes. So yes, the allocation could be outside the loop. I probably won't change it now though.

Issue #407, opened 2016-09-22
Review use of HttpClient


jskeet:
We have some code which uses HttpClient to fetch resources

We should review these articles to work out what makes sense here.
- http://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/
- http://odetocode.com/blogs/scott/archive/2004/06/08/http-connection-limits.aspx
- http://byterot.blogspot.co.uk/2016/07/singleton-httpclient-dns.html

evildour:
I recently saw this article which summarizes two of those links and also lists out the methods which are thread safe on HttpClient: https://www.infoq.com/news/2016/09/HttpClient

manigandham:
The advice is basically that HttpClient is thread-safe and automatically manages multiple connections and requests so it should be a singleton for the lifetime of the application. It implements `IDisposable` but only because it has resources to be disposed, not because it needs to be disposed on every request. There's no need to use more than one instance for the entire application but it's not too big of a deal if major libraries use their own.

Reusing a single instance or allowing for an instance to be passed in the construction/config would likely be best.

jskeet:
@manigandham: Except it's not quite that simple due to DNS caching as I understand it.

manigandham:
@jskeet 

There is the `ServicePointManager.DnsRefreshTimeout` property specifically for controlling the DNS resolution cache settings for new connections to a service point (address). The default is 2 mins.

https://msdn.microsoft.com/en-us/library/system.net.servicepointmanager.dnsrefreshtimeout(v=vs.110).aspx

Will do some testing but never encountered any DNS problems with `HttpClient` before.

jskeet:
@manigandham: I believe the problem is for new connections to the same name, when it's got a new IP address. See http://byterot.blogspot.co.uk/2016/07/singleton-httpclient-dns.html We can't know whether that's relevant to our users or not, unfortunately :(

manigandham:
@jskeet 

That blog post seems to be confusing DNS TTLs with HTTP/TCP connection lifetimes.

HttpClient is just a wrapper around HttpClientHandler which uses the HttpWebRequest stack, and this stack uses connection pooling with `ServicePoint` objects managed by the `ServicePointManager`. HTTP/TCP connections stay open indefinitely and don't do DNS lookups again, which is the correct behavior. New requests on a new connection will do a new DNS lookup and get the updated address.

[...]

PR #1622, opened 2017-11-06
StorageService implements IDisposable.

chrisdunelm:
Updates #407

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.cs
@@ -37,8 +37,14 @@ namespace Google.Cloud.Storage.V1
     /// All instance methods declared in this class are virtual, with an implementation which simply
     /// throws <see cref="NotImplementedException"/>. All these methods are overridden in <see cref="StorageClientImpl"/>.
     /// </para>
+    /// <para>
+    /// This class implements <see cref="IDisposable"/>; however, the <see cref="Dispose"/> method only requires calling
+    /// if many short-lived instances of <see cref="StorageClient"/> are being created.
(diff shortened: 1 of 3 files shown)


chrisdunelm:
Once we're happy with the code and xmldoc, I'll do the other REST-based services.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.cs
```diff
@@ -37,8 +37,14 @@ namespace Google.Cloud.Storage.V1
     /// All instance methods declared in this class are virtual, with an implementation which simply
     /// throws <see cref="NotImplementedException"/>. All these methods are overridden in <see cref="StorageClientImpl"/>.
     /// </para>
+    /// <para>
+    /// This class implements <see cref="IDisposable"/>; however, the <see cref="Dispose"/> method only requires calling
+    /// if many short-lived instances of <see cref="StorageClient"/> are being created.
+    /// Most code would be expected to create a single <c>StorageClient</c> instance, and use this instance throughout
+    /// the lifetime of the application. In this case, <c>Dispose</c> need not be called.
```
jskeet: Even creating multiple clients isn't a problem in many cases, of course. We'll definitely want to have a broader FAQ entry about this, but we can do that later.

chrisdunelm:
Yes, I wondered about making Dispose virtual here, with an empty method body; then overriding in the Impl to actual do the disposal.

jskeet:
Empty rather than throwing NotImplementedException? Not sure which I prefer.

chrisdunelm:
Empty means that an alternative implementation (possibly for testing) doesn't need to worry about disposal if disposal isn't required; otherwise code outside of that users control may call Dispose which will throw the exception.

jskeet:
Sure - it's easy enough to override where appropriate. Let's go for that.

chrisdunelm:
I'll merge on green

PR #1624, opened 2017-11-06
Translation and BigQuery implement IDisposable

chrisdunelm:


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClient.cs
```diff
@@ -42,8 +42,14 @@ namespace Google.Cloud.BigQuery.V2
...
     /// </para>
+    /// <para>
+    /// This class implements <see cref="IDisposable"/>; however, the <see cref="Dispose"/> method only requires calling
+    /// if many short-lived instances of <see cref="BigQueryClient"/> are being created.
+    /// Most code would be expected to create a single <c>BigQueryClient</c> instance, and use this instance throughout
+    /// the lifetime of the application. In this case, <c>Dispose</c> need not be called.
+    /// </para>
     /// </remarks>
-    public abstract partial class BigQueryClient
+    public abstract partial class BigQueryClient : IDisposable
```
benwulfe: how will this affect the version of the package?  presumably this is on the level of a breaking change since it adds new requirements for using the client.

jskeet:
@benwulfe: No, because it's not actually a requirement. As documented, it's something you should do in some situations, but it's not only allowing you to be more efficient. Any code that doesn't dispose of the client will work just as well as it did before.

chrisdunelm:
`AdvancedTranslationClient` done

benwulfe:
If you have singletons that implement IDisposable, not disposing it is a pretty common (if maybe not entirely correct) usage pattern.  I'd argue this caveat doesn't really make it special or unique.

benwulfe:
I guess all I was saying is that the very existence of having an object that implements IDisposable that is not disposed is likely to create a lint error where it was ok before.
linters will not be able to read your comments.

jskeet:
Well, I'd argue that the expected behaviour at that point is for the developer to read the linter error, read the doc comment then, and configure the linter appropriately. If the linter can't be configured, I think that's a bigger problem. (And how does it cope with HttpClient and Task for example?)

I really believe that bumping the major version number would be a mistake here, giving an unrealistic impression of breaking.

PR #1633, opened 2017-11-08
BigtableClient.ReadModifyWriteRow[Async]

evildour:


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/ReadModifyWriteRulesTest.cs
```diff
@@ -0,0 +1,67 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+using Xunit;
+
+namespace Google.Cloud.Bigtable.V2.Tests
+{
+    public class ReadModifyWriteRulesTest
+    {
+        public static IEnumerable<object[]> InvalidFamilyNames { get; } = new List<object[]>
```
jskeet: I've found it easier to do this by creating a `List<object>` and then calling `.Select(t => new[] { t }).ToList()` at the end. I may put something in the ClientTesting library for this.

Alternatively, we can use `TheoryData<T>` which I haven't tried yet, but which looks like it should be simple.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -419,6 +419,220 @@ public partial class BigtableClient
         }
 
         /// <summary>
+        /// Modifies a row atomically. The method reads the latest existing version
```
jskeet: This sounds like the client is doing the reading then writing, whereas in reality it's server-side, right?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -419,6 +419,220 @@ public partial class BigtableClient
...
+        /// and value from the specified columns and writes a new entry based on
+        /// pre-defined read/modify/write rules. The new value for the version is the
```
jskeet: "pre-defined" => "the specified"? (Otherwise it sounds like the user doesn't get to say what happens...)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -419,6 +419,220 @@ public partial class BigtableClient
...
+        /// greater of the existing version or the current server version. The method
```
jskeet: It's not clear to me what the difference is between "the existing version" and "the current server version". Can we clarify that?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/ReadModifyWriteRulesTest.cs
evildour: Thanks. Didn't know about `TheoryData<T>`. That cleans things up a bit. Also used it in the client tests to hold all invalid row keys.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Correct. Note that these are just the auto-generated docs copied here, so if someone uses the generated overloads, they're still going to see this summary. But I can add "...on the server." to the end of this sentence to clear up these overloads.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Again these are copies of auto-generated content. I think in this context it means "defined before the request gets sent to the server" so the user doesn't think they can read a value, then decide how to modify and write the new value based on that. If you still think it should be changed though, I don't feel too strongly about it.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Done. Let me know if that is still unclear.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
jskeet: Right, I wondered whether that would be the case. Possibly the best thing to do is feed back to the Bigtable team that the proto comments could be clearer, then regenerate - and copy the updated comments when they come through. But this is fine for now :)

evildour: Ok thanks. I'll reach out to them internally.

PR #1637, opened 2017-11-08
Bigtable integration tests use same instance

evildour:


jskeet:
If you could write up a README.md file somewhere suitable (in the root Bigtable directory perhaps, if the same instance can be used for snippets too) for testing requirements, that would be handy - both for me and for @iantalarico who'll need to set up our CI appropriately (and anyone else, of course).

PR #1638, opened 2017-11-08
Improve pubsub integration test

### LINKED ISSUE #1610: Google.Cloud.PubSub.V1.IntegrationTests BulkMessaging test hangs on deadlock
Jon, Chris: Not sure which of you is best to talk to on this.  This is showing up on the Windows CI.  Just using this as a place holder so we don't lose track.

I have had two tests run that hang for about 30 min.

chrisdunelm:
Handle duplicate messages and nacks properly, hopefully make it less flakey

Fix #1610

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
@@ -16,6 +16,7 @@ using Google.Api.Gax;
 using Google.Api.Gax.Grpc;
 using Grpc.Core;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
```diff
@@ -41,13 +42,18 @@ public SimplePubSubTest(PubsubFixture fixture)
 
         private readonly PubsubFixture _fixture;
 
-        private async Task RunBulkMessaging(int messageCount, int minMessagesize, int maxMessageSize, int maxMessagesInFlight, int initialNackCount, TimeSpan? timeouts = null)
+        private async Task RunBulkMessaging(
+            int messageCount, int minMessageSize, int maxMessageSize, int maxMessagesInFlight, int initialNackCount,
+            TimeSpan? timeouts = null, int? cancelAfterRecvCount = null)
         {
+            // Force messages to be at least 4 byte long, so an int ID can be used.
```
jskeet: Nit: byte => bytes

jskeet: (And don't we control the arguments? Can't we just validate that minMessageSize is at least 4?)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
```diff
@@ -73,29 +79,50 @@ public SimplePubSubTest(PubsubFixture fixture)
...
+                int id = BitConverter.ToInt32(msg.Data.ToArray(), 0);
+                lock (nackedIds)
                 {
-                    return Task.FromResult(SimpleSubscriber.Reply.Nack);
+                    if (nackedIds.Count < initialNackCount)
+                    {
+                        if (nackedIds.Add(id))
+                        {
+                            // This ID not already nacked
+                            var localRecvCount = Interlocked.Increment(ref recvCount);
```
jskeet: What if this is cancelAfterRecvCount? Or is that always greater than initialNackCount? (Is that checked?)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
```diff
@@ -73,29 +79,50 @@ public SimplePubSubTest(PubsubFixture fixture)
...
+                            return Task.FromResult(SimpleSubscriber.Reply.Nack);
+                        }
+                    }
+                }
+                bool wasAdded = recvedIds.TryAdd(id, false);
+                if (wasAdded)
+                {
+                    var localRecvCount = Interlocked.Increment(ref recvCount);
+                    Interlocked.Add(ref recvSum, msg.Data.Sum(x => (long)x));
+                    if (localRecvCount == cancelAfterRecvCount || localRecvCount >= messageCount + initialNackCount)
```
jskeet: Why would it be greater than messageCount + initialNackCount?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
chrisdunelm: cancelAfterRecvCount only applies to non-nacked (acked) messages, although I agree that's not very clear.
(Also got rid of pointless allocation)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
chrisdunelm: It shouldn't be, but this is just in case...

PR #1640, opened 2017-11-08
Ensure the run integration test script does not fail on Linux.

iantalarico:
The 'realpath' command fails on Linux when the file does not exist (but does not fail on windows).

### DIFF
--- a/runintegrationtests.sh
+++ b/runintegrationtests.sh
@@ -6,6 +6,9 @@
 set -e
 source toolversions.sh
 
+SCRIPT=$(readlink -f "$0")
+ROOT_DIR=$(dirname "$SCRIPT")
+
 RETRY_ARG=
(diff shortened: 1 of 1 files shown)


runintegrationtests.sh
```diff
@@ -6,6 +6,9 @@
 set -e
 source toolversions.sh
 
+SCRIPT=$(readlink -f "$0")
+ROOT_DIR=$(dirname "$SCRIPT")
```
jskeet: I've generally used `realpath` for this:

```sh
ROOT_DIR=$(realpath $(dirname $0))
```

... but if `dirname` works, that's fine...

iantalarico: I really don't know either, it's just what I'd used in some other scripts.

I did a quick Google Search and it seems they are more or less the same.

PR #1650, opened 2017-11-10
Fixes an issue where an unused session could still be running its pre…

benwulfe:
…warm when its released synchronously back into the pool.  To fix this, we now have a condition that all returned sessions from a pool have their prewarm completed before they can be used.

Also fixes a minor issue in the Logger where the exception was not logged.


jskeet:
Can't say I follow all the details here, but it looks like it makes sense...

PR #1651, opened 2017-11-10
[Bigtable] Use generated table names in integration tests

evildour:
I forgot to do this when converting to use a single instance in #1637

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
@@ -42,6 +42,7 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
         public BigtableTableAdminClient TableAdminClient { get; private set; }
         public BigtableClient TableClient { get; private set; }
 
+        public List<TableName> CreatedTables { get; } = new List<TableName>();
         public Channel EmulatorChannel { get; }
 
         public BigtableFixture()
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
```diff
@@ -200,9 +203,16 @@ public BigtableFixture()
...
         {
-            if (TableAdminClient != null && TableName != null)
+            if (TableAdminClient != null)
             {
-                TableAdminClient.DeleteTable(new DeleteTableRequest { Name = TableName.ToString() });
+                foreach (var tableName in CreatedTables)
+                {
+                    try
+                    {
+                        TableAdminClient.DeleteTable(new DeleteTableRequest { Name = tableName.ToString() });
```
jskeet: Can you not do `DeleteTable(TableName)`? If not, that suggests we should probably be fixing that in the GAPIC config.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/ReadRowsTest.cs
```diff
@@ -478,11 +478,7 @@ public void ReadRowMissing()
         [Fact]
         public async Task ReadRows_Validate_SingleIteration()
         {
-            var tableName = new TableName(
-                _fixture.InstanceName.ProjectId,
-                _fixture.InstanceName.InstanceId,
-                "ReadRows_Validate_SingleIteration_Table");
-            await _fixture.CreateTable(tableName);
+            var tableName = await _fixture.CreateTable();
```
jskeet: Is this the only test that needs to create a new table? That's surprising.

evildour: Most tests use the default table created by the fixture and just insert a row with a random row key. But this one creates its own table because it iterates the rows to make sure only one row is present.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
jskeet: Ah, okay. (We've seen that before, and it's slightly annoying, but there's not much we can do about it as we don't want dependencies in either direction...)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/ReadRowsTest.cs
jskeet: Right, makes sense.

PR #1652, opened 2017-11-11
Spanner ycsb

benwulfe:
This test app is the standard one being used for spanner performance and stress testing across languages.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Google.Cloud.Spanner.Data.YCSB.csproj
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Google.Cloud.Spanner.Data.YCSB.csproj
@@ -1,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <StartupObject>Google.Cloud.Spanner.Data.Ycsb.Program</StartupObject>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\Google.Cloud.Spanner.Admin.Database.V1\Google.Cloud.Spanner.Admin.Database.V1\Google.Cloud.Spanner.Admin.Database.V1.csproj" />
+    <ProjectReference Include="..\..\Google.Cloud.Spanner.Admin.Instance.V1\Google.Cloud.Spanner.Admin.Instance.V1\Google.Cloud.Spanner.Admin.Instance.V1.csproj" />
+    <ProjectReference Include="..\..\Google.Cloud.Spanner.V1\Google.Cloud.Spanner.V1\Google.Cloud.Spanner.V1.csproj" />
+    <ProjectReference Include="..\Google.Cloud.Spanner.Data\Google.Cloud.Spanner.Data.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="workload_template">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
+
+</Project>
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
@@ -1,0 +1,171 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using Google.Cloud.Spanner.V1.Internal.Logging;
+
+namespace Google.Cloud.Spanner.Data.Ycsb
+{
+    internal class Program : YcsbRunner
+    {
+        private static readonly string s_tableName = "table";
+        private static readonly string s_cloudspanner_instanceName = "cloudspanner.instance";
+        private static readonly string s_cloudspanner_databaseName = "cloudspanner.database";
+        private static readonly string s_cloudspanner_projectName = "cloudspanner.project";
+        private static readonly string s_boundedStaleness = "cloudspanner.boundedstaleness";
+        private static readonly string s_numchannels = "cloudspanner.numchannels";
+
+        private static long s_boundedStalenessSeconds;
+        private static readonly List<string> s_keys = new List<string>();
+
+        private static string s_table;
+        private static string s_instance;
+        private static string s_database;
+        private static string s_project;
+        private static string s_connectionString;
+        private static string s_selectQuery;
+
+        private static readonly string[] s_fieldNames =
+        {
+            "field0",
+            "field1",
+            "field2",
+            "field3",
+            "field4",
+            "field5",
+            "field6",
+            "field7",
+            "field8",
+            "field9"
+        };
+
+        private static SpannerConnection s_connection;
+        private static TimestampBound s_timestampbound;
+        private static Program DefaultInstance { get; } = new Program();
+
+        protected override List<Operation> Operations { get; } = new List<Operation>
+        {
+            new Operation(
+                "readproportion", async latencies =>
+                {
+                    await Task.Yield();
+                    var sw = Stopwatch.StartNew();
+                    var cmd = s_connection.CreateSelectCommand(
+                        s_selectQuery, new SpannerParameterCollection
+                        {
+                            {"p", SpannerDbType.String, s_keys[Rand.Value.Next(s_keys.Count)]}
+                        });
+                    using (var reader = await cmd.ExecuteReaderAsync(s_timestampbound).ConfigureAwait(false))
+                    {
+                        while (await reader.ReadAsync().ConfigureAwait(false)) { }
+                    }
+                    latencies.Add(sw.ElapsedMilliseconds);
+                }),
+            new Operation(
+                "updateproportion", async latencies =>
+                {
+                    await Task.Yield();
+                    var sw = Stopwatch.StartNew();
+                    var cmd = s_connection.CreateUpdateCommand(
+                        s_table,
+                        new SpannerParameterCollection
+                        {
+                            {"id", SpannerDbType.String, s_keys[Rand.Value.Next(s_keys.Count)]},
+                            {s_fieldNames[Rand.Value.Next(10)], SpannerDbType.String, GenerateRandomString()}
+                        });
+                    await cmd.ExecuteNonQueryAsync().ConfigureAwait(false);
+                    latencies.Add(sw.ElapsedMilliseconds);
+                }),
+            new Operation("scanproportion", x => throw new NotImplementedException("scan is not yet implemented")),
+            new Operation("insertproportion", x => throw new NotImplementedException("scan is not yet implemented"))
+        };
+
+        public static int Main(string[] args) => DefaultInstance.Run(args);
+
+        /// <summary>
+        /// Runs a single worker task
+        /// </summary>
+        protected override async Task PreWarmOneInstance()
+        {
+            using (var spannerConnection = new SpannerConnection(
+                $"Data Source=projects/{s_project}/instances/{s_instance}/databases/{s_database}"))
+            {
+                await spannerConnection.OpenAsync();
+            }
+        }
+
+        protected override void InitializeSettings()
+        {
+            base.InitializeSettings();
+            if (IsDebugMode)
+            {
+                Logger.DefaultLogger.LogPerformanceTraces = true;
+                if (GetOption<int>("debugmode") == 2)
+                {
+                    Logger.DefaultLogger.LogLevel = V1.Internal.Logging.LogLevel.Debug;
+                }
+            }
+
+            int targetQps = GetOptionWithDefault(TargetQps, 0);
+            if (targetQps > 0)
+            {
+                int countToPreWarm = Math.Min(targetQps / 4, 800);
+                Options[Prewarm] = countToPreWarm;
+                SpannerOptions.Instance.MaximumActiveSessions = Math.Max(countToPreWarm + 50, 400);
+                SpannerOptions.Instance.MaximumPooledSessions = Math.Max(countToPreWarm + 50, 400);
+                SpannerOptions.Instance.MaximumGrpcChannels = Math.Max(4, 8 * targetQps / 2000);
+            }
+            else
+            {
+                SpannerConnection.SpannerOptions.MaximumGrpcChannels = GetOptionWithDefault(s_numchannels, 1);
+                SpannerConnection.SpannerOptions.MaximumPooledSessions = Math.Max(
+                    GetOption<int>(NumWorker) + 1,
+                    GetOptionWithDefault(Prewarm, 0));
+            }
+            LoadKeys();
+            s_timestampbound = s_boundedStalenessSeconds > 0
+                ? TimestampBound.OfMaxStaleness(TimeSpan.FromSeconds(s_boundedStalenessSeconds))
+                : TimestampBound.Strong;
+        }
+
+        private void LoadKeys()
+        {
+            s_connectionString = $"Data Source=projects/{s_project}/instances/{s_instance}/databases/{s_database}";
+            s_selectQuery = $"SELECT u.* FROM {s_table} u WHERE u.id=@p";
+            s_connection = new SpannerConnection(s_connectionString);
+            s_connection.Open();
+
+            DebugMessage("Loading keys...");
+            var cmd = s_connection.CreateSelectCommand($"SELECT u.id FROM {s_table} u");
+            var reader = cmd.ExecuteReader();
+            while (reader.Read())
+            {
+                s_keys.Add(reader.GetString(0));
+            }
+        }
+
+        protected override void ValidateArguments()
+        {
+            base.ValidateArguments();
+            s_table = GetOption<string>(s_tableName);
+            s_instance = GetOption<string>(s_cloudspanner_instanceName);
+            s_database = GetOption<string>(s_cloudspanner_databaseName);
+            s_project = GetOption<string>(s_cloudspanner_projectName);
+            s_boundedStalenessSeconds = GetOptionWithDefault(s_boundedStaleness, 0);
+        }
+    }
+}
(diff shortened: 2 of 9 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+        private static readonly List<string> s_keys = new List<string>();
+        private static string s_templateFile;
+
+        private static readonly ConcurrentDictionary<string, object> s_options =
+            new ConcurrentDictionary<string, object>();
+
+        private static bool s_debugMode;
+        private static readonly Random s_globalRandom = new Random(Environment.TickCount);
+
+        private static readonly ThreadLocal<Random> s_rand = new ThreadLocal<Random>(() =>
```
jskeet: I used to recommend this approach to RNGs, but it doesn't produce very random results. See https://stackoverflow.com/questions/36376888/seeding-multiple-random-number-generators

If the randomness doesn't matter too much, it's fine to keep this - otherwise, I'd probably use a single instance and just lock each time. It's annoying, but such is life :(

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/workload_template
```diff
@@ -0,0 +1,203 @@
...
+# out. When a property has a finite number of settings,
+# the default is enabled and the alternates are shown in
+# comments below it.
+#
+# Use of most explained through comments in Client.java or
+# CoreWorkload.java or on the YCSB wiki page:
+# https://github.com/brianfrankcooper/YCSB/wiki/Core-Properties
+
+# The name of the workload class to use
+workload=com.yahoo.ycsb.workloads.CoreWorkload
```
jskeet: Is this irrelevant for this code, or is it invoking Java somewhere that I'm not aware of? It would be useful to restrict this file to just the options that are relevant, IMO.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Google.Api.Gax;
+using Google.Cloud.Spanner.V1.Internal.Logging;
+
+namespace Google.Cloud.Spanner.Data.YCSB
+{
+    internal class Program
```
jskeet: Do we know whether YCSB is likely to be useful for other APIs? I'm just wondering whether it's worth separating the Spanner-specific parts from the non-Spanner-specific parts. Maybe not, or at least not now - just worth considering.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+            public Operation(string proportionField, Func<ConcurrentBag<double>, Task> action)
+            {
+                ProportionField = proportionField;
+                Name = proportionField.Replace("proportion", "").ToUpperInvariant();
+                Action = action;
+            }
+
+            public ConcurrentBag<double> Latencies { get; } = new ConcurrentBag<double>();
+            public string Name { get; }
+            public double Proportion => GetOption<double>(ProportionField);
```
jskeet: What's this meant to indicate? I can see it referred to from a bunch of places, but I can't easily understand the purpose.

(After a while I think I understand it, but a comment would be really helpful.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+        private static async Task RunWorkerWithQps(double queriesPerSecond, TimeSpan testTime)
+        {
+            var sw = Stopwatch.StartNew();
+            var all = new List<Task>();
+
+            while (sw.Elapsed < testTime)
+            {
+                if (sw.Elapsed.TotalSeconds * queriesPerSecond > all.Count)
+                {
+                    var operationTarget = s_rand.Value.NextDouble() * s_total;
```
jskeet: Call ChooseOperation instead?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+            operationTarget -= s_operations[operation].Proportion;
+            while (operationTarget > 0 && operation < s_operations.Count)
+            {
+                operation++;
+                if (operation < s_operations.Count)
+                {
+                    operationTarget -= s_operations[operation].Proportion;
+                }
+            }
+            return operation;
```
jskeet: Make this return `s_operations[Math.Min(operation, s_operations.Count - 1)]`? It looks like this should make the calling code simpler to follow.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+        {
+            using (var spannerConnection = new SpannerConnection(
+                $"Data Source=projects/{s_project}/instances/{s_instance}/databases/{s_database}"))
+            {
+                await spannerConnection.OpenAsync();
+            }
+        }
+
+        /// <summary>
+        /// Runs a single worker task that targets a specific QPS without regard to workers.
```
jskeet: This looks like it starts multiple worker tasks, not a single one. Or I may just be confused.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+        /// (this is from C# spanner stress tests)
+        /// </summary>
+        private static async Task RunWorkerWithQps(double queriesPerSecond, TimeSpan testTime)
+        {
+            var sw = Stopwatch.StartNew();
+            var all = new List<Task>();
+
+            while (sw.Elapsed < testTime)
+            {
+                if (sw.Elapsed.TotalSeconds * queriesPerSecond > all.Count)
```
jskeet: I *think* I understand what this is doing, although the tight loop makes me nervous. It feels like a "start task / sleep" loop would be simpler to understand, but presumably the problem there is with sleep granularity/accuracy?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+                }
+            }
+
+            var timeout = Task.Delay(TimeSpan.FromSeconds(10));
+            await Task.WhenAny(Task.WhenAll(all), timeout);
+            s_totaloperations = all.Count;
+        }
+
+        /// <summary>
+        /// Runs a single worker task
```
jskeet: Doesn't it run `numWorkers` tasks?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+        /// </summary>
+        /// <param name="numWorkers"></param>
+        private static async Task RunWithTargetWorkers(int numWorkers)
+        {
+            //yield to let Main continue through.
+            await Task.Yield();
+            var parallelWorkers = new List<Task>();
+            var operation = ChooseOperation();
+
+            while (Interlocked.Decrement(ref s_operationsRemaining) > 0)
```
jskeet: I don't understand why this is a global variable at all, rather than just a simple for loop with a local variable up to `s_totalOperations`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+            {
+                ProportionField = proportionField;
+                Name = proportionField.Replace("proportion", "").ToUpperInvariant();
+                Action = action;
+            }
+
+            public ConcurrentBag<double> Latencies { get; } = new ConcurrentBag<double>();
+            public string Name { get; }
+            public double Proportion => GetOption<double>(ProportionField);
+            public Func<ConcurrentBag<double>, Task> Action { get; }
```
jskeet: We always seem to invoke this with `Latencies` as the argument. Might it be simpler to make the `Action` private, and have a `Start` method that calls `action(Latencies)`?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+
+        private static void PrintResults(long totalElapsedMillis)
+        {
+            Console.WriteLine($"[OVERALL], RunTime(ms), {totalElapsedMillis}");
+            Console.WriteLine(
+                $"[OVERALL], Throughput(ops/sec), {1000 * s_totaloperations / totalElapsedMillis}");
+
+            foreach (var operation in s_operations)
+            {
+                var sortedlatencies = operation.Latencies.DefaultIfEmpty(0).OrderByDescending(x => x).ToList();
```
jskeet: Move this into the `Operation` class as a `PrintResults` method?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+                Console.WriteLine($"[{operation.Name}], Operations, {operation.Latencies.Count}");
+                Console.WriteLine(
+                    $"[{operation.Name}], AverageLatency(us), {operation.Latencies.DefaultIfEmpty(0).Average() * 1000}");
```
jskeet: Can't we just use `sortedlatencies` for all these? (Indeed, won't the MinLatency be `sortedlatencies.Last()`, and MaxLatency be `sortedlatencies.First()`?)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+            AppDomain.CurrentDomain.UnhandledException += (sender, eventArgs) =>
+            {
+                Console.Error.WriteLine($"ERROR {eventArgs.ExceptionObject.ToString()}");
+            };
+
+            Trace.Listeners.Add(new TextWriterTraceListener(Console.Out));
+            if (s_debugMode)
+            {
+                Logger.DefaultLogger.LogPerformanceTraces = true;
+                if (GetOption<int>("debugmode") == 2)
```
jskeet: Any reason "debugmode" isn't extracted as a string constant like other options?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+            foreach (var line in File.ReadLines(s_templateFile))
+            {
+                if (line.StartsWith("#"))
+                {
+                    continue;
+                }
+                var kvp = line.Split('=');
+                if (kvp.Length == 2)
+                {
+                    s_options[kvp[0].ToLowerInvariant()] = kvp[1];
```
jskeet: Trim both values? (Or is that against YCSB in general?)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+        private static readonly string s_num_worker = "num_worker";
+        private static readonly string s_operationcount = "operationcount";
+        private static readonly string s_boundedStaleness = "cloudspanner.boundedstaleness";
+        private static readonly string s_numchannels = "cloudspanner.numchannels";
+        private static readonly string s_prewarm = "cloudspanner.prewarm";
+        private static readonly string s_targetQps = "cloudspanner.qpstarget";
+        private static readonly string s_qpstestTime = "cloudspanner.qpstesttime";
+
+        private static long s_operationsRemaining;
+        private static long s_totaloperations;
```
jskeet: Nit: s_totalOperations

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -0,0 +1,513 @@
...
+        {
+            if (s_debugMode)
+            {
+                Console.WriteLine(message);
+            }
+        }
+
+        private static void PrintUsage()
+            => Console.Error.WriteLine(
+                "Usage:\n ycsb_spanner_csharp run -P <relative_path_to_template_file> [-p property=value]...");
```
jskeet: Unless ycsb_spanner_csharp exists somewhere I'm not aware of, maybe just point at the readme.md file?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
```diff
@@ -193,7 +193,7 @@ public virtual void Error(Func<string> messageFunc, Exception exception = null)
         {
             if (LogLevel >= LogLevel.Error)
             {
-                WriteLine(messageFunc());
+                WriteLine($"{messageFunc()}, Exception = {exception}");
```
jskeet: Presumably the rest of the changes in Google.Cloud.Spanner.V1 will go away after #1650 is merged?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Google.Cloud.Spanner.Data.YCSB.csproj
```diff
@@ -0,0 +1,19 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+    <StartupObject>Google.Cloud.Spanner.Data.YCSB.Program</StartupObject>
```
jskeet: Nit: I'd make this Ycsb rather than YCSB (like Http etc). Not terribly important, I understand.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Google.Cloud.Spanner.Data.YCSB.csproj
```diff
@@ -0,0 +1,19 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Google.Cloud.Spanner.Data" Version="1.0.0-beta08" />
```
jskeet: Any reason this isn't a project reference?

benwulfe: i wanted to test against the public version, but a project reference is better going forward.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
benwulfe: I was actually thinking this myself and this would be a good thing.  I'll do the refactor to two classes but for now still keep it all in one assembly.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
benwulfe: The lack of real randomness is actually OK and I'd rather not introduce any concurrency locks if it doesnt relate to a real world requirement.  This test is extremely perf sensitive

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
benwulfe: the wording is poor.  it meant this is to be used as a single task that executes multiple subtasks itself.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
benwulfe: maybe a task.yield would be good as an else clause.  this isn't the primary test, but is identical to the old stress test I ran earlier.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
benwulfe: I've clarified the comment.  It's implementation changed slightly to reduce some awaits.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
benwulfe: in a previous iteration, I had several worker tasks, so this needed to be shared.  it now doesnt even need the interlocked.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
benwulfe: likely because I defined it versus it being defined across all languages for ycsb.  I'll add it as a const.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
benwulfe: It's the old name of the command before I renamed it.  I've updated it and added the proper "dotnet ..." prefix.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
benwulfe: its cleaner, but executing the operations is in the base class now, so there wasnt as much value in it.  but may as well...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/workload_template
benwulfe: its irrelevant.  this file started from a "real" config file for spanner YCSB (as in -- not-a-pretend-YCSB-package).
So I wanted to ensure it worked as is with all properties that java uses.
Most of these are not actually used.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
benwulfe: yep

PR #1656, opened 2017-11-14
Implement language-agnostic proto tests

jskeet:
Definitely review one commit at a time.

The second commit mostly changes how we handle sentinel values, including making sure that we optimize away unnecessary updates.

Having implemented these tests, I've given feedback which will probably create *more* tests and require more changes - but I wanted to get these in first.


apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data.Tests/Proto/ProtoTest.cs
```diff
@@ -0,0 +1,255 @@
...
+                default:
+                    throw new ArgumentException($"Unknown value kind: {value.KindCase}");
+            }
+        }
+
+        // Rather than using Moq to expect a commit, we just have a small recording client. This makes
+        // it easier to get good diagnostics about what went wrong.
+        private class MockCommitClient : FirestoreClient
+        {
+            private readonly CommitRequest m_expectedRequest;
```
chrisdunelm: Why the "m" prefix?

chrisdunelm:
Oops, hang on. I LGTM'ed after only looking at the first commit.

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data.Tests/Proto/ProtoTest.cs
jskeet: Whoops, had forgotten our convention. (I've used too many conventions in too many different places...)

apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/WriteBatch.cs
```diff
@@ -58,16 +57,13 @@ public WriteBatch Create(DocumentReference documentReference, object documentDat
...
-                {
-                    Fields = { ValueSerializer.SerializeMap(documentData) },
-                    Name = documentReference.Path
-                }
-            };
-            AddUpdate(write);
+            var fields = ValueSerializer.SerializeMap(documentData);
+            var serverTimestamps = new List<FieldPath>();
+            var deletes = new List<FieldPath>();
+            FindSentinels(fields, FieldPath.Empty, serverTimestamps, deletes);
```
chrisdunelm: Could make `FindSentinels` return a `(List<FieldPath>, List<FieldPath>)` tuple?
(Or even better, use `ImmutableList<FieldPath>`)

jskeet: I suspect we'll want to introduce a dependency on ValueTuple eventually, but I'd rather put that off for now. (Especially as I seem to remember there are some issues with ValueTuple dependencies due to a mistake in one version. I'd need to validate that though.)

Fortunately all of this is internal - we can do it later if we want.

jskeet:
Will fix the m_, rebase and merge on green.

PR #1657, opened 2017-11-14
Make cursor limits for queries more restrictive

### LINKED ISSUE #1629: Prevent indirect child docs in cursors
query.OrderBy(FieldPath.DocumentId).StartAt("foo/bar/baz") should fail - only StartAt("foo") should be accepted.

jskeet:
Fixes #1629

### DIFF
--- a/apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/Query.cs
+++ b/apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/Query.cs
@@ -390,11 +390,12 @@ namespace Google.Cloud.Firestore.Data
                         case string relativePath:
                             // Note: this assumes querying over a single collection at the moment.
                             // Convert to a DocumentReference for the cursor
+                            PathUtilities.ValidateId(relativePath, nameof(fieldValues));
                             value = Collection.Document(relativePath);
                             break;
                         case DocumentReference absoluteRef:
-                            // Just validate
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Firestore.Data/Google.Cloud.Firestore.Data/Query.cs
```diff
@@ -390,11 +390,12 @@ private Cursor CreateCursor(object[] fieldValues, bool before)
...
                             // Note: this assumes querying over a single collection at the moment.
                             // Convert to a DocumentReference for the cursor
+                            PathUtilities.ValidateId(relativePath, nameof(fieldValues));
                             value = Collection.Document(relativePath);
                             break;
                         case DocumentReference absoluteRef:
-                            // Just validate
-                            GaxPreconditions.CheckArgument(absoluteRef.IsDescendantOf(Collection), nameof(fieldValues),
+                            // Just validate that the given document is a direct child of the parent collection.
+                            GaxPreconditions.CheckArgument(absoluteRef.Parent.Equals(Collection), nameof(fieldValues),
```
chrisdunelm: It's guaranteed that `Parent` can never be `null`?

jskeet: Yup (and even documented :)

(The parent of a collection can be null, but never the parent of a document.)

PR #1660, opened 2017-11-14
Update Language to v1.1.0 with dependencies

jskeet:
(Grpc.Core dependency is 1.7.1 despite being 1.7.0 in GAX to allow the memory leak fix to be in.)


jskeet:
(Will self-merge on green, then do a release.)

PR #1662, opened 2017-11-14
Add Integration tests for Google.Cloud.Diagnostics.AspNet.Trace 

iantalarico:
These are sudo integration tests as we still have to fake the HttpContext as an HttpApplication cannot be started programatically.


iantalarico:
Correct, the core tests are able to run fully in memory.

Issue #1480, opened 2017-09-20
Users want faster cloud storage uploads.


SurferJeffAtGoogle:
I compared the performance of our current cloud storage library to gsutil.  Source code is here:
https://github.com/SurferJeffAtGoogle/scratch/tree/master/StorageSpeed

Uploading 1 GB file:
```
rennie@rennie:~/gitrepos/scratch/StorageSpeed$ dotnet run newfile bigone.bin -g 1
rennie@rennie:~/gitrepos/scratch/StorageSpeed$ ls -sh
total 954M
954M bigone.bin  4.0K bin  4.0K obj  4.0K Program.cs  4.0K StorageSpeed.csproj
rennie@rennie:~/gitrepos/scratch/StorageSpeed$ time dotnet run upload bigone.bin bookshelf-dotnet.appspot.com

real    0m35.922s
user    0m6.176s
sys     0m6.428s
rennie@rennie:~/gitrepos/scratch/StorageSpeed$ time gsutil cp bigone.bin gs://bookshelf-dotnet.appspot.com/bigone2.bin
Copying file://bigone.bin [Content-Type=application/octet-stream]...
==> NOTE: You are uploading one or more large file(s), which would run
significantly faster if you enable parallel composite uploads. This
feature can be enabled by editing the
"parallel_composite_upload_threshold" value in your .boto
configuration file. However, note that if you do this large files will
be uploaded as `composite objects
<https://cloud.google.com/storage/docs/composite-objects>`_,which
means that any user who downloads such objects will need to have a
compiled crcmod installed (see "gsutil help crcmod"). This is because
without a compiled crcmod, computing checksums on composite objects is
so slow that gsutil disables downloads of composite objects.

/ [1 files][953.7 MiB/953.7 MiB]   70.5 MiB/s
Operation completed over 1 objects/953.7 MiB.

real    0m13.647s
user    0m5.068s
sys     0m2.472s
rennie@rennie:~/gitrepos/scratch/StorageSpeed$ gsutil ls -lh gs://bookshelf-dotnet.appspot.com/bigone*
953.67 MiB  2017-09-20T19:07:57Z  gs://bookshelf-dotnet.appspot.com/bigone.bin
953.67 MiB  2017-09-20T19:08:48Z  gs://bookshelf-dotnet.appspot.com/bigone2.bin
TOTAL: 2 objects, 2000000000 bytes (1.86 GiB)
[...]

PR #1666, opened 2017-11-16
Simple throughput measurement tool for storage

jskeet:
We can use this to investigate #1480.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.ThroughputTool/Program.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.ThroughputTool/Program.cs
@@ -1,0 +1,90 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.IO;
+
+namespace Google.Cloud.Storage.V1.ThroughputTool
+{
+    public static class Program
+    {
+        private static void Main(string[] args)
+        {
+            Configuration configuration = null;
+            try
+            {
+                configuration = Configuration.FromCommandLineArguments(args);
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.ThroughputTool/Program.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            var random = new Random();
+            long bytesLeft = configuration.SizeBytes;
+            byte[] buffer = new byte[1024 * 1024];
+            using (var output = File.Create(configuration.LocalFile))
+            {
+                while (bytesLeft > 0)
+                {
+                    random.NextBytes(buffer);
+                    int count = Math.Min((int) bytesLeft, buffer.Length);
```
chrisdunelm: Do the Min as `long`s, then cast to int; otherwise I suspect there could be buuuuuuuugs...

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.ThroughputTool/ProgressReporter.cs
```diff
@@ -0,0 +1,73 @@
...
+                    return;
+                }
+                DateTime now = DateTime.UtcNow;
+                TimeSpan elapsed = _stopwatch.Elapsed;
+                double totalSeconds = elapsed.TotalSeconds;
+                long bytesProcessed = Interlocked.Read(ref _bytesProcessed);
+                double megabytesProcessed = bytesProcessed / (1024.0 * 1024.0);
+                double megabitsProcessed = megabytesProcessed * 8;
+                double? mbps = totalSeconds < 0.1 ? default(double?) : megabitsProcessed / totalSeconds;
+                Console.WriteLine($"{now:HH:mm:ss}   {elapsed:hh':'mm':'ss}  {megabytesProcessed,8:####0.00}   {mbps,6:##0.00}");
```
chrisdunelm: Inconsistent use of single quotes around colons.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.ThroughputTool/Program.cs
jskeet: Ooh, yes, you're right.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.ThroughputTool/ProgressReporter.cs
jskeet: Yup, that's because they're absolutely needed for TimeSpan. If a user happens to have a non-colon time separator, I don't mind that much on the date/time... but it would fail for TimeSpan.

PR #1669, opened 2017-11-16
When building a release also run the integration tests and snippets for that package.

iantalarico:
This will streamline the release process and allow for easier automation of the release.

### DIFF
--- a/runintegrationtests.sh
+++ b/runintegrationtests.sh
@@ -27,6 +27,12 @@ while (( "$#" )); do
   shift
 done
 
+if [[ ${#APIS[@]} != 0 && "$RETRY_ARG" == "yes" ]]
+then
+  echo "The --retry flag cannot be used when specifying projects to test."
+  exit 1
+fi
+
 # We only overwrite integration-test-failures.txt at the very end,
 # so that if we abort tests early, we don't assume there's nothing to retry.
 FAILURE_FILE=$ROOT_DIR/integration-test-failures.txt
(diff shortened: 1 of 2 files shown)


runintegrationtests.sh
```diff
@@ -39,7 +37,25 @@ touch $FAILURE_TEMP_FILE
 
 cd apis
 
-if [[ "$RETRY_ARG" == "yes" && (-f "$FAILURE_FILE")]]
+if [[ ${#APIS[@]} != 0 ]]
```
jskeet: I think a comment would be useful - it had me confused for a while.
Something like:

```text
# There are three separate ways of determining the tests to run:
# - If APIs have been specified on the command line, test those
# - If --retry has been specified, run the previously-failed tests
# - Otherwise, fine all potential tests
```

Should we also check that `--retry` hasn't been specified along with a set of APIs?

iantalarico: I added a check to fail if --retry and projects are set as well as the comment you suggested.

iantalarico:
Thanks Jon, PTAL

runintegrationtests.sh
```diff
@@ -37,6 +43,10 @@ touch $FAILURE_TEMP_FILE
 
 cd apis
 
+# There are three separate ways of determining the tests to run:
+# - If APIs have been specified on the command line, test those
+# - If --retry has been specified, run the previously-failed tests
+# - Otherwise, fine all potential tests
```
ILMTitan: fine => find

PR #1670, opened 2017-11-17
raw and empty option for testing spanner.

benwulfe:
NOT FOR CHECKIN.

Jon, I was hoping you would take a quick look at this.
Will send more info in email

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
@@ -13,10 +13,19 @@
 // limitations under the License.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
+using Google.Apis.Auth.OAuth2;
+using Google.Cloud.Spanner.V1;
 using Google.Cloud.Spanner.V1.Internal.Logging;
+using Google.Protobuf.WellKnownTypes;
+using Grpc.Auth;
+using Grpc.Core;
+using TypeCode = Google.Cloud.Spanner.V1.TypeCode;
 
 namespace Google.Cloud.Spanner.Data.Ycsb
 {
@@ -28,6 +37,8 @@ namespace Google.Cloud.Spanner.Data.Ycsb
         private static readonly string s_cloudspanner_projectName = "cloudspanner.project";
         private static readonly string s_boundedStaleness = "cloudspanner.boundedstaleness";
         private static readonly string s_numchannels = "cloudspanner.numchannels";
+        private static readonly string s_raw = "cloudspanner.raw"; //execute directly bypassing veneer and gapic
+        private static readonly string s_empty = "cloudspanner.empty"; //do nothing -- just pretend to run.
 
         private static long s_boundedStalenessSeconds;
         private static readonly List<string> s_keys = new List<string>();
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -80,37 +95,109 @@ internal class Program : YcsbRunner
...
+                                Strong = true
+                            }
+                        }
+                    }
+                });
+            while (await stream.ResponseStream.MoveNext().ConfigureAwait(false))
+            {
+            }
+            stream.ResponseStream.Dispose();
+            s_rawsessions.Enqueue(session);
```
benwulfe: this is just for testing but if something bad happens and we throw and never re-add the session to the queue, then we will nullref above.  This is ok for this test since it will actually make the test run much faster.... (and should show up in console err)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -64,14 +77,16 @@ internal class Program : YcsbRunner
                 {
                     await Task.Yield();
                     var sw = Stopwatch.StartNew();
-                    var cmd = s_connection.CreateSelectCommand(
-                        s_selectQuery, new SpannerParameterCollection
-                        {
-                            {"p", SpannerDbType.String, s_keys[Rand.Value.Next(s_keys.Count)]}
-                        });
-                    using (var reader = await cmd.ExecuteReaderAsync(s_timestampbound).ConfigureAwait(false))
+                    if (!IsEmpty)
```
jskeet: It feels like having separate classes here (one for raw, one for empty, one for regular) might make a lot of this code a lot simpler. I'm happy to treat all of this as slightly experimental though, and we can refactor later, when we have a better idea of what features we really need.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -80,37 +95,109 @@ internal class Program : YcsbRunner
...
+                {
+                    {"p", SpannerDbType.String, s_keys[Rand.Value.Next(s_keys.Count)]}
+                });
+            using (var reader = await cmd.ExecuteReaderAsync(s_timestampbound).ConfigureAwait(false))
+            {
+                while (await reader.ReadAsync().ConfigureAwait(false)) { }
+            }
+        }
+
+        private static V1.Type s_stringType = new V1.Type { Code = TypeCode.String};
```
jskeet: readonly?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -28,6 +37,8 @@ internal class Program : YcsbRunner
         private static readonly string s_cloudspanner_projectName = "cloudspanner.project";
         private static readonly string s_boundedStaleness = "cloudspanner.boundedstaleness";
         private static readonly string s_numchannels = "cloudspanner.numchannels";
+        private static readonly string s_raw = "cloudspanner.raw"; //execute directly bypassing veneer and gapic
```
jskeet: This isn't currently bypassing the gapic/toolkit code - you're still using Google.Cloud.Spanner.V1.SpannerClient. To get the raw gRPC client, use the `SpannerClient.GrpcClient` property to get a `Google.Cloud.Spanner.V1.Spanner+SpannerClient` object (as in, the nested type). The naming is somewhat confusing here, but the hope is that very few users ever need to go down to the gRPC level.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.YCSB/Program.cs
```diff
@@ -55,6 +66,8 @@ internal class Program : YcsbRunner
 
         private static SpannerConnection s_connection;
         private static TimestampBound s_timestampbound;
+        private static SpannerClient s_rawclient;
```
jskeet: Nits: s_rawClient and s_rawSessions. (And ideally s_timestampBound above, although that's already in.)

benwulfe:
this was never intended for checkin.  I will follow up on an isolated repro for my issue.

PR #1671, opened 2017-11-17
Storage pubsub

jskeet:
Review commit by commit as normal :)


jskeet:
Note that when this is merged, it adds an extra requirement for integration tests (in terms of pubsub IAM). Need to work out where to document that...

BrandonY:
Code in 1671 looks good. One note: Your integration test will work the vast majority of the time, but it will flake sometimes. Two reasons:
* It can take up to 30 seconds for notification configs to become active (although will usually be less)
* Technically, there's no guarantee that a notification arrive within 5 seconds, although it will the vast majority of the time.

Depending on your testing setup's tolerance of flakes, you may want to retry once or twice, wait additional time, or disable the test. If a rare flake is fine, I'd probably still recommend adding a 15 second wait after creating the notification and before uploading an object.

One other note. Almost every user will use this library the same way, and that way is a bit tricky:
* Create a topic
* Fetch the service account
* Grant publisher permission for the service account to the topic.
* Create a notification.

It'd be worth providing sample code in the documentation for how to implement this flow, so that individual customers don't need to all figure this out independently. There was some discussion in the Python pull request about this as well, and it sounds like there's currently no place for helper functions that span multiple cloud products, but we should at least paste the appropriate code somewhere for users.

jskeet:
Thanks, I'll up the timeout for the integration test and a wait after creating the notification.

Yes, I can fairly easily create a snippet that would show doing all of that. (Once I've worked out what to do in terms of the IAM call.)

PR #1675, opened 2017-11-17
First pass at general testing documentation with API-specific details

jskeet:
This deletes the Bigtable-specific README, putting the details in
TESTING.md instead. I removed the duplication of the TEST_PROJECT
variable, but kept the part about account permissions, as I wasn't
sure whether the regular service account for a project owner was
enough. (I'll try that on Monday and slim the doc down if possible.)

I *think* these three are the only APIs that require additional
instructions for now.


TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+optionally specifying a test filter and/or target framework.
+
+Common environment variables
+============================
+
+To run the integration/smoke tests or snippets, set the following
+environment variables:
+
+- `TEST_PROJECT`: The name of the Google Cloud Project to test against
+- `GOOGLE_APPLICATION_CREDENTIALS`: The absolute path to a JSON file
```
iantalarico: Should we make a note about how to do this with gcloud?

TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+Currently there are no runnable snippets, but when they are added, they will
+have the same requirements.
+
+Google.Cloud.Firestore
+----------------------
+
+Firestore cannot currently operate on a Google Cloud project that also has a
+Datastore database. You will need a second project:
+
+- Create a project on the [Firebase Console](https://console.firebase.google.com/u/0/)
```
iantalarico: strip out '/u/0/'?

TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+- Enable Firestore on that project (within the Firebase Console)
+- Set the `FIRESTORE_TEST_PROJECT` environment variable with that
+  project ID
+- Add the account identified in your service account JSON file as an
```
iantalarico: or run these tests by themselves and change the GOOGLE_APPLICATION_CREDENTIALS?

TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+  
+Google.Cloud.Storage.V1
+-----------------------
+
+- Enable the Google Cloud Pub/Sub API as well as Google Cloud Storage
+- Determine your Google Cloud Storage service account and give it
+  publisher access for PubSub:
+  - Either run the integration tests and see the error message in
+    the failure, or determine the service account as
+    `PROJECT_ID_HERE@gs-project-accounts.iam.gserviceaccount.com`
```
iantalarico: Is this project id or project number?

TESTING.md
```diff
@@ -0,0 +1,114 @@
...
+- `GOOGLE_APPLICATION_CREDENTIALS`: The absolute path to a JSON file
+  containing service credentials for the test project.
+
+Per API configuration
+=====================
+
+Some APIs require a little more setup. If you're not testing an API
+listed below, the instructions above should be all you need.
+
+Google.Cloud.Bigtable.V2
```
evildour: It may make sense to also list the Admin library here. It doesn't have integration tests/snippets yet, but when it does, they will have the same requirements.

TESTING.md
jskeet: *Possibly* - although I suspect if you're sufficiently deeply in that you want to run our integration tests, you'll be past that and it would just be noise.

TESTING.md
jskeet: Will do.

TESTING.md
jskeet: While that would work, it means you can't run `runintegrationtests.sh`. I think this is a simpler approach.

TESTING.md
jskeet: The project ID, same as `TEST_PROJECT`. Would "where `PROJECT_ID_HERE` is the value of the `TEST_PROJECT` environment variable" help?

TESTING.md
jskeet: Good plan, will do.

TESTING.md
iantalarico: I think we should just have it as an options.  This is what I do but I also run all of these tests on semi regularly.  I'm guessing most people who are not us will run one set of tests at a time.   This way users don't have to mess around with switching permissions and adding service accounts

TESTING.md
iantalarico: I don't think it matters too much.  I just know most service accounts use the project number instead of the project id.  But if this just uses the ID what you have is good.

TESTING.md
jskeet: Done.

jskeet:
Addressed comments in a second commit (which I'll squash before merging). PTAL.

TESTING.md
```diff
@@ -66,8 +66,8 @@ Per API configuration
 Some APIs require a little more setup. If you're not testing an API
 listed below, the instructions above should be all you need.
 
-Google.Cloud.Bigtable.V2
-------------------------
+Google.Cloud.Bigtable.V2 and Google.Cloud.BigTable.Admin.V2
```
evildour: "BigTable" => "Bigtable"

jskeet: Whoops - thanks, done.

PR #1678, opened 2017-11-20
Deleting Storage objects before deleting a bucket

jskeet:
This is a not-for-submission draft PR to allow deleting a buckets objects while deleting the bucket itself.

It has no tests at the moment, which are clearly needed... but I wanted to get some confirmation that this feels appropriate first.

In particular, there's no parallelism here, which potentially there could be - but then it would require configuration for exactly how parallel everything should be, etc. Maybe that could be added later? Thoughts welcome.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.DeleteBucket.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.DeleteBucket.cs
@@ -14,6 +14,7 @@
 
 using Google.Apis.Storage.v1;
 using Google.Apis.Storage.v1.Data;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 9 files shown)


jskeet:
Brokenness was due to failing to pass cancellation tokens around. I'd also failed to list ListObjectsAsync. I don't see a nice way of executing an asynchronous foreach where each operation is asynchronous too, at the moment... what I've got instead is reasonably simple though.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.DeleteBucket.cs
```diff
@@ -23,37 +24,79 @@ public sealed partial class StorageClientImpl : StorageClient
...
-            options?.ModifyRequest(request);
-            return request;
+            // Create the request at the start, so we validate the options before we
+            // start deleting objects.
+            var deleteBucketRequest = CreateDeleteBucketRequest(bucket, options);
+            string userProject = options?.UserProject;
+            if (options?.DeleteObjects ?? false)
+            {
+                // Don't start deleting objects if the bucket has the wrong metageneration to start with.
+                var preconditionOptions = options?.CreateGetBucketOptionsForPreconditions();
```
ILMTitan: Don't we know `options` is not null here?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.DeleteBucket.cs
```diff
@@ -23,37 +24,79 @@ public sealed partial class StorageClientImpl : StorageClient
...
+                    await GetBucketAsync(bucket, preconditionOptions, cancellationToken).ConfigureAwait(false);
+                }
+                var objects = ListObjectsAsync(bucket, null, new ListObjectsOptions { Versions = true, UserProject = userProject });
+                using (var iterator = objects.GetEnumerator())
+                {
+                    while (await iterator.MoveNext(cancellationToken).ConfigureAwait(false))
+                    {
+                        var obj = iterator.Current;
+                        var deleteObjectOptions = new DeleteObjectOptions { UserProject = userProject, Generation = obj.Generation };
+                        await DeleteObjectAsync(obj, deleteObjectOptions, cancellationToken).ConfigureAwait(false);
```
ILMTitan: Shouldn't we collect all of these into a list, and then to `await Task.WhenAll(taskList)`? Is there a concern about the number of open connections, or does the client self throttle?

ILMTitan: Task[] deleteTasks = await objects.Select(
        obj =>
        {
            var deleteObjectOptions =
                new DeleteObjectOptions {UserProject = userProject, Generation = obj.Generation};
            return DeleteObjectAsync(obj, deleteObjectOptions, cancellationToken);
        }).ToArray(cancellationToken).ConfigureAwait(false);
    await Task.WhenAll(deleteTasks).ConfigureAwait(false);

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.DeleteBucket.cs
jskeet: Yes, we do - doh.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.DeleteBucket.cs
jskeet: I definitely don't want to do it in an unconstrained way. There can be millions of objects in a bucket. If wanted to parallelize, I'd potentially use `AsRawResponses` and delete a page of data at a time (which is 1000 objects by default), and even then possibly set concurrency to ~8 concurrent requests.

jskeet:
@chrisdunelm: I'm tempted to delete this at the moment, as the whole parallelism part is a bit nasty. What do you think?

chrisdunelm:
I suspect this is a common requirement; i.e. if a user is deleting a bucket, then I would expect that much of time it's very convenient if it can also delete all the contents.
So I think this should be kept. I agree there are issues around many objects, and possible parallelisation optimization. But so long as this is explained in the documentation I think it's OK.

jskeet:
Okay - so does the general approach look okay and I should add tests?

jskeet:
(Discussed offline - will add tests tomorrow.)

chrisdunelm:
Discussed offline, but in summary: yes this looks fine (opt-in flag, no parallelization), yes to the tests :)

jskeet:
Now with added tests, and ready for review.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.DeleteBucket.cs
jskeet: (Now fixed.)

chrisdunelm:
Not sure if I'm just missing it, or if it isn't required...
But I was expecting a test that creates, fills a bucket; then deletes it and checks the bucket has gone. There doesn't seem to be one?

jskeet:
I'm currently relying on the delete operation succeeding - I could add a check on every successful delete call to check that the bucket is actually gone. (I'd add it in the DeleteBucket and DeleteBucketAsync methods, probably.) Is that the sort of thing you're after?

chrisdunelm:
I'd quite like to see an explicit test case that does this. It makes it very clear that this functionality is really tested.

jskeet:
Sure - will create a second commit temporarily (to make the change clear) then squash it when you're happy.

chrisdunelm:
Thanks :)

jskeet:
Well this is embarrassing. I see what you mean by the lack of tests now - I'd forgotten to add a new file. I thought you were requesting additional changes *within* the file. I've made those changes anyway, and I think the test is better for it. Added the test as a new commit.

jskeet:
Pushed the squashed version - will merge after CI goes green.

chrisdunelm:
SGTM

PR #1680, opened 2017-11-20
Do not try to run AspNet Integration tests on non windows environments

iantalarico:
They will not be able to restore and will fail


iantalarico:
Thanks Jon!

PR #1682, opened 2017-11-21
Container API initial checkin

jskeet:


apis/apis.json
```diff
@@ -67,6 +67,19 @@
     }
   },
 
+  {
+    "id": "Google.Cloud.Container.V1",
+    "productName": "Google Kubernetes Engine API",
+    "productUrl": "https://cloud.google.com/kubernetes-engine/docs/reference/rest/",
+    "version": "1.0.0-beta00",
```
chrisdunelm: Should this be beta01?

jskeet: Not yet. Will update that when I'm ready to do the first release. (Just waiting on confirmation of something.)

PR #1684, opened 2017-11-21
Add documentation for our package names

jskeet:
(This is primarily to explain why the API version appears in package
names.)

### DIFF
--- a/docs/root/guides/package-names.md
+++ b/docs/root/guides/package-names.md
@@ -1,0 +1,57 @@
+## Package names
+
+The package names for the Google Cloud Client Libraries for .NET take
+multiple forms. Here are a few examples:
+
+- Google.Cloud.Datastore.V1
+- Google.Cloud.Firestore
+- Google.Cloud.Logging.Type
+- Google.Cloud.Logging.Log4Net
+
+It's somewhat unusual to see a version number in a NuGet package
+name, as in the Datastore example. There's a reason though: this is
+a package for that specific version of the Datastore API. We
+distinguish between the *API* (exposed by Google servers, used by
+multiple languages) and the *package* (the .NET-specific NuGet
+package).
+
+The version number in the package name is the API version number.
+Separately, that package can go through multiple versions. A change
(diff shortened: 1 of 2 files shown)


docs/root/guides/package-names.md
```diff
@@ -0,0 +1,50 @@
...
+to the API will always mean a new package version, but the reverse
+isn't true: sometimes we may improve the package even without the
+API changing. (For example, adding more convenience methods.)
+
+Having the version number in the package name gives the advantage
+that if you want to migrate from one major API version to another,
+you don't have to do that for your whole codebase in a single step.
+You can take a dependency on Google.Cloud.AwesomeApi.V1 and
+Google.Cloud.AwesomeApi.V2 in the same project, migrating your code
+piecemeal.
```
chrisdunelm: Might also be worth mentioning how this brings sanity to having dependencies that themselves depend on Google APIs.
I.e. Your project can have deps on two different packages that themselves depend on different API versions.
I.e. Your project doesn't have to use the same major API version as your deps.

docs/root/guides/package-names.md
```diff
@@ -0,0 +1,50 @@
...
+the escape hatch of going down to the version-specific client.
+
+Google.Cloud.Logging.Log4Net is similar in terms of being
+version-agnostic, but is intended as an integration point with
+Log4Net. Similarly, Google.Cloud.Spanner.Data is an integration
+point with ADO.NET.
+
+Google.Cloud.Logging.Type is slightly different: it only contains a
+set of "version-agnostic" protocol buffer messages, which we'd
+expect to be a dependency of Google.Cloud.Logging.V*X* for any value
```
chrisdunelm: 'n' might be clearer than 'X' here?

docs/root/guides/package-names.md
jskeet: Good call, will do.

docs/root/guides/package-names.md
jskeet: Done.

PR #1685, opened 2017-11-21
Include missed OS when running integration tests on windows.

iantalarico:


jskeet:
LGTM, but I'm not sure why Travis is failing (and I'm on a poor internet connection, so it's hard to check). If that can be resolved, go ahead and merge :)

iantalarico:
It looks like a on-off failure
"oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"

PR #1686, opened 2017-11-21
Default to netcoreapp for coverage reports.

iantalarico:
This will allow for more easily running coverage cross platform (also as this line also influences how we run integration tests)

This will allow for all frameworks to be run.  While this will take longer it will be more complete.


jskeet:
Hmm... while it sounds fine to me, I think there was a reason for preferring the desktop framework, and I can't remember what it was. Ben, can you remember?

iantalarico:
I talked with Ben a bit yesterday he didn't seem to remember any reason for having this set this way.

iantalarico:
@jskeet I just removed the file in a new commit and will squash and merge

benwulfe:
Oh I misunderstood the question.  I think it was net452 because
a) there is extra code for net452 that will show up as uncovered in netcoreapp
b) i was running on windows anyhow.

and its a preference based on target fx.
I'm fine with changing it, but it should prefer netcoreapp 2.0 over 1.0 to capture code supported in netstandard 2.0.

tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
```diff
@@ -335,9 +335,9 @@ private static void GenerateMainProject(ApiMetadata api, string directory, HashS
         private static string GetPreferredCoverageFramework(ApiMetadata api)
         {
             var targetFrameworks = GetTestTargetFrameworks(api);
-            if (targetFrameworks.Contains(TargetFrameworkClassic))
+            if (targetFrameworks.Contains(TargetFrameworkCore))
```
benwulfe: instead of this change, can you see if the target fx includes netcoreapp 2.0 and use that over 1.0?
if you dont do that and 1.0 is used, it could miss some conditional tests and lower reported coverage.

jskeet: How about:

```
private static readonly string[] s_preferredCoverageFrameworks= ...
...
// In the method
return s_preferredCoverageFrameworks.FirstOrDefault(f => targetFrameworks.Contains(f))
    ?? targetFrameworks.Split(';').FirstOrDefault();
```

iantalarico: Should we be doing this for all supported frameworks instead? Is there a reason we only do it for one?

We also use this in runintegrationtests.sh, should we also be running all of them there?

Is the only down side it takes more time?

benwulfe: it would take quite a bit more time if target fx lists net45, core1.0 and core2.0 and at least in many cases be redundant.
But also if you want to ensure it runs on linux, you cannot include net45 anyway.

iantalarico: I *think* If we just do "test -c Release" it will run all the frameworks it has and can run ont eh current OS (@jskeet Does this sound right?).  I'd rathe be redundant and slow then miss anything especially for these automated or not often run tasks.

jskeet: Hmm. Possibly. I'm nervous about adding what could be a *very* long time to the build, especially as we get more APIs, but maybe it would be the right thing to do for now.

How easy would it be to try it once on the build machine and see how long it takes?

iantalarico: @jskeet I can just modify the script and run it on one of the machines.  It won't run it as a jenkins job but it should give us an idea.

iantalarico: Just some quick notes from some small tests.  I can run the whole thing if we want but I have the following:

- The integration tests currently take about 55min to run
- Running "test -c Release" will "do the right thing" and run all of the frameworks it can.
- Most tests have at a max run under net452, netcoreapp1.0 and nercoreapp2.0.  So given this worst case is probably just under 3 hours on windows (I guess less given not all have all 3 frameworks) and less then 2 hours on unix.

While this will take a fair amount of time this is mostly automated builds, I'd rather be complete then now.  If we ever get to a point where it is too long we can always parse the list of directories in apis/, split it up and parallelize it. 

Given the above I'd vote for running them all.

@benwulfe @jskeet: Thoughts?

jskeet: Does this increase the flakiness? Each test will run three times, and we'll fail if they don't *all* succeed. It will also potentially mean more API calls in a short space of time leading to more quota issues.

I think I'd prefer to focus on reducing flakiness first and then do this - although I agree it's a reasonable thing to do once we're there.

iantalarico: This does increase the chance for failure.  So yea this could make things more flakey and will mean more API calls. I don't think we should worry about quota issues (we can follow up offline about this)

I agree we need to keep reducing flakiness (and we have very few flaky tests ATM) but I think this is more important then some potential extra noise.  We should be testing against all of the frameworks without this we have the chance to miss possible failures.  Not even so much for the CI but for the release process.  runintegrationtests.sh only does the test on one framework.

jskeet: Okay. Let's be aggressive about it for now - we can always back off (or maybe only back off in CI) if necessary.

iantalarico:
@jskeet, @benwulfe I have updated the code to run tests on all frameworks. PTAL.

PR #1693, opened 2017-11-22
Add convenience methods for IAM

jskeet:
- AddRoleMember
- RemoveRoleMember

Both indicate whether or not they've changed anything.

We might also want to add a "get/act/set-if-necessary" with retries
for concurrency method later, but that takes a little more design
work.

(At some point we can then do a new IAM release, and use it from the Storage Pub/Sub snippets.)

### DIFF
--- a/apis/Google.Cloud.Iam.V1/Google.Cloud.Iam.V1/PolicyPartial.cs
+++ b/apis/Google.Cloud.Iam.V1/Google.Cloud.Iam.V1/PolicyPartial.cs
@@ -1,0 +1,81 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


jskeet:
The get/act/set abstraction is actually pretty hard to provide at the moment, due to the mix-in approach of including IAM in APIs. We may want to think about that a bit harder.

apis/Google.Cloud.Iam.V1/Google.Cloud.Iam.V1/PolicyPartial.cs
```diff
@@ -0,0 +1,81 @@
...
+        /// <summary>
+        /// Adds the specified member to the specified role. If the role does
+        /// not already exist, it is created.
+        /// </summary>
+        /// <param name="role">The role to add the member to. Must not be null.</param>
+        /// <param name="member">The member to add to the role. Must not be null.</param>
+        /// <returns><c>true</c> if the policy was changed; <c>false</c> if the member already existed in the role.</returns>
+        public bool AddRoleMember(string role, string member)
+        {
+            GaxPreconditions.CheckNotNull(role, nameof(role));
```
chrisdunelm: Should these also check the strings are not empty?

jskeet: Yes, probably a good plan. Will change.

jskeet: Done. Will merge on green.

PR #1697, opened 2017-11-23
(NOT FOR SUBMISSION) Prototype of IAM helper

jskeet:
This shows the rough approach.

Copying a bunch of code from GAX feels really ugly. I'd like to change that, possibly by exposing more from GAX itself.

Currently there's no default retry settings - we need to work out exactly what those would be.


jskeet:
Things to think about:

- If the caller provides a RetrySettings, how does the filter from that work?
- How do we cope with the possibility that the client class has a Modify_GetIAMPolicyRequest method?

jskeet:
Closing this for now, as I suspect the extra complexity in terms of retry etc is worse than the user just doing it themselves.

PR #1707, opened 2017-11-25
Refactorized code

cruzespinal:
Simple optimization for a cleaner code, using lambda expressions.


jskeet:
We always use braces round `if` statements, even for single statements - ditto `for` loops etc. The code in SubscriberClient is all autogenerated.

While I'm fine to use expression-bodied members, I wouldn't suggest creating PRs just to change to using them. (It's not like they're actually faster or anything.)

Thanks for your time, but I'm afraid we won't be merging this PR.

PR #1708, opened 2017-11-25
Add ThrowOnError to VideoAnnotationResults

cruzespinal:
I have created a simple test for 'VideoAnnotationResults' in 'Video Intelligence API', I have not created more because the API is still in a 'Beta' state.

Regards!

### DIFF
--- a/apis/Google.Cloud.VideoIntelligence.V1/Google.Cloud.VideoIntelligence.V1/VideoAnnotationResultsException.cs
+++ b/apis/Google.Cloud.VideoIntelligence.V1/Google.Cloud.VideoIntelligence.V1/VideoAnnotationResultsException.cs
@@ -1,0 +1,49 @@
+﻿// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using System;
+
+namespace Google.Cloud.VideoIntelligence.V1
+{
+    /// <summary>
+    /// An error occurring when annotating an video.
+    /// </summary>
+    public sealed class VideoAnnotationResultsException : Exception
+    {
+        /// <summary>
+        /// The complete response containing the error.
+        /// </summary>
+        public VideoAnnotationResults Response { get; }
+
+        /// <summary>
+        /// Constructs an exception based on the error in <paramref name="response"/>.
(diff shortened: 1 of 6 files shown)


jskeet:
There's no point in adding unit tests for `VideoAnnotationResults` - it's generated code, as is everything in this project.

I suggest that if you want to contribute, you discuss ideas of what you might want to contribute with us before you start writing code, otherwise it's very likely you'll just end up wasting your time, I'm afraid. 

I'm not trying to discourage contributions - I'm just saying that as this codebase is somewhat unusual in terms of how much of it is generated, it's worth checking first.

jskeet:
Hang on - I hadn't realized that you'd also added functionality. (That isn't part of the PR description.)

That's *potentially* useful, and we'll need to consider it. I'll edit the PR description to match what's actually in it.

jskeet:
(My advice of checking the code you're proposing before starting to write the code still applies though. It would be very easy for you to write a lot of code and then for us to reject the PR for reasons we could have explained beforehand.)

cruzespinal:
Agree.
ThrowOnError is required in the Container, Dlp, Logging, Debugger and Trace APIs? Do I add them?

Thanks,
John

jskeet:
No, it's not a pattern to be applied universally. If it were, we'd put it in codegen.

It's only useful in cases where a response can be successful at the HTTP level - and potentially still include partial success information - but also represent partial failure.

jskeet:
Okay, I've had another look at this now.

I don't think it's worth adding to just `VideoAnnotationResults`. We could potentially add `ThrowOnAnyError` to `VideoAnnotationResponse`; that would be consistent with the Vision API.

Whether it's worth adding *just* this to the API when there's no other handwritten code is debatable, however - it comes with a maintenance cost in terms of future versions (including API beta versions) where the manual code would need to be propagated. That cost would then be reflected in not being able to spend the time on other work. I believe all the other languages only have auto-generated code for the Video Intelligence API.

@chrisdunelm Thoughts on this? It's reasonably simple code, and I wouldn't hesitate if we already had other handwritten code...

chrisdunelm:
The `ThrowOnError` and potentially `ThrowOnAnyError` (or could still be called `ThrowOnError`)  looks useful, and I think should be included.
I agree about the maintenance cost, but consider it worth it in this instance.

jskeet:
Right - I'll double-check the details of this PR later, merge it, and then add ThrowOnAnyError (name matches the one in Vision) to the response.

apis/Google.Cloud.VideoIntelligence.V1/Google.Cloud.VideoIntelligence.V1/VideoAnnotationResultsException.cs
```diff
@@ -0,0 +1,49 @@
...
+
+using Google.Api.Gax;
+using System;
+
+namespace Google.Cloud.VideoIntelligence.V1
+{
+    /// <summary>
+    /// An error occurring when annotating an video.
+    /// </summary>
+    public sealed class VideoAnnotationResultsException : Exception
```
jskeet: For consistency with Vision, I'll rename this to AnnotateVideoException in a follow-up PR.
(The Vision API has the wrong filename for the class too, which I'll fix.)

PR #1718, opened 2017-11-28
Report progress after the final delay in Storage perf test

jskeet:
(It's just possible that the delay finishes just before the
completion, so we report interim numbers but then break out of the
loop, but it's simple this way and very unlikely to actually cause
issues.)


jskeet:
CI tests took just over 3 minutes each :)

PR #1720, opened 2017-11-28
Spanner GA fixes to support EF.

benwulfe:
This PR contains bug fixes that I want to get in for Spanner GA so that developers do not depend on unsupported functionality (or bugs).  More changes may come in the future to support EF, but these fixes are rather small and good to get in.  This will hopefully create a GA candidate as performance testing has passed and I just need to regenerate the docs.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
@@ -428,7 +428,8 @@ namespace Google.Cloud.Spanner.Data
                 {
                     if (string.IsNullOrEmpty(_connectionStringBuilder?.SpannerDatabase))
                     {
-                        Logger.Warn(() => "No database was defined. Therefore OpenAsync did not establish a session.");
+                        Logger.Info(() => "No database was defined. Therefore OpenAsync did not establish a session.");
+                        _state = ConnectionState.Open;
                         return;
                     }
                     if (IsOpen)
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
@@ -233,6 +233,15 @@ namespace Google.Cloud.Spanner.Data
                 DataSource = dataSource
             };
 
+        /// <summary>
+        /// Returns a new instance of a <see cref="SpannerConnectionStringBuilder"/> with the database
+        /// portion of the DataSource replaced with a new value.
+        /// </summary>
+        /// <param name="database">the new database name.</param>
+        /// <returns>a new instance of <see cref="SpannerConnectionStringBuilder"/></returns>
+        public SpannerConnectionStringBuilder WithDatabase(string database) 
+            => CloneWithNewDataSource($"projects/{Project}/instances/{SpannerInstance}/databases/{database}");
+
         private string GetValueOrDefault(string key, string defaultValue = "")
         {
             key = key.ToLower();
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
@@ -238,7 +238,7 @@ namespace Google.Cloud.Spanner.Data
         /// portion of the DataSource replaced with a new value.
         /// </summary>
         /// <param name="database">The new database name. Can be null to open a connection for Ddl commands.</param>
-        /// <returns>a new instance of <see cref="SpannerConnectionStringBuilder"/></returns>
+        /// <returns>A new instance of <see cref="SpannerConnectionStringBuilder"/></returns>
         public SpannerConnectionStringBuilder WithDatabase(string database)
             => string.IsNullOrEmpty(database)
                 ? CloneWithNewDataSource($"projects/{Project}/instances/{SpannerInstance}")
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
@@ -237,10 +237,12 @@ namespace Google.Cloud.Spanner.Data
         /// Returns a new instance of a <see cref="SpannerConnectionStringBuilder"/> with the database
         /// portion of the DataSource replaced with a new value.
         /// </summary>
-        /// <param name="database">the new database name.</param>
+        /// <param name="database">The new database name. Can be null to open a connection for Ddl commands.</param>
         /// <returns>a new instance of <see cref="SpannerConnectionStringBuilder"/></returns>
-        public SpannerConnectionStringBuilder WithDatabase(string database) 
-            => CloneWithNewDataSource($"projects/{Project}/instances/{SpannerInstance}/databases/{database}");
+        public SpannerConnectionStringBuilder WithDatabase(string database)
+            => string.IsNullOrEmpty(database)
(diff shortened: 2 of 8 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -268,7 +268,14 @@ internal V1.Type ToProtobufType()
         /// </summary>
         /// <param name="size"></param>
         /// <returns></returns>
-        public SpannerDbType WithSize(int size) => new SpannerDbType(TypeCode, size);
+        public SpannerDbType WithSize(int size)
+        {
+            if (size != 0 && TypeCode != TypeCode.Bytes && TypeCode != TypeCode.String)
```
iantalarico: Should this check be size > 0?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -428,7 +428,8 @@ public override Task OpenAsync(CancellationToken cancellationToken)
                 {
                     if (string.IsNullOrEmpty(_connectionStringBuilder?.SpannerDatabase))
                     {
-                        Logger.Warn(() => "No database was defined. Therefore OpenAsync did not establish a session.");
+                        Logger.Info(() => "No database was defined. Therefore OpenAsync did not establish a session.");
+                        _state = ConnectionState.Open;
```
jskeet: It's not clear why the state would be open if we don't have a database. What can be done with this connection afterwards? Can it be treated like other open connections? Should we throw instead?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -233,6 +233,15 @@ internal SpannerConnectionStringBuilder CloneWithNewDataSource(string dataSource
                 DataSource = dataSource
             };
 
+        /// <summary>
+        /// Returns a new instance of a <see cref="SpannerConnectionStringBuilder"/> with the database
+        /// portion of the DataSource replaced with a new value.
+        /// </summary>
+        /// <param name="database">the new database name.</param>
```
jskeet: Can this be null or empty?
(Nit: capital letter to start both this and the returns part)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -268,7 +268,14 @@ internal V1.Type ToProtobufType()
         /// </summary>
         /// <param name="size"></param>
         /// <returns></returns>
-        public SpannerDbType WithSize(int size) => new SpannerDbType(TypeCode, size);
+        public SpannerDbType WithSize(int size)
```
jskeet: Please add documentation.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
jskeet: Or possibly check that `size >= 0` first, and then check that if it *is* non-zero, that it's for a suitable type?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
benwulfe: The state needs to be considered open because some of EF will not allow Ddl statements unless the connection is opened. 
I purposely allow specifying a connection string without a database (but with project+instance) because spanner has no "master" db concept.  But it was an oversight to not set the state to opened when that happened.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
benwulfe: +1  can be null and should switch to a different string if so.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
benwulfe: this is actually pretty important because size is ambiguous for string.  for spanner, its # of characters.

jskeet: What does "characters" mean here? Unicode code points, UTF-16 code units, UTF-8 bytes? (Often the last is used for databases if UTF-8 is raw storage :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
benwulfe: I'll need to double check this one because a negative size (-1) might be used as a marker.  This check should be in the property setter.  it can be here too, but the main place would be there.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
benwulfe: 'characters' is the term used in public spanner documentation and i think it makes sense to be consistent.  While "code point" is probably the more exact term that covers formatting, it seems a bit unnecessary and overly technical.  I've never heard anyone refer to unicode 'characters' as 'code points' in vernacular. 

Specifying UTF-8 at all might be unnecessarily exposing an implementation detail since the encoding is enforced server side independent of the transport.

jskeet: Right, if it's definitely Unicode characters then the encoding is irrelevant. (I assume it's not going to perform any kind of normalization.)

Whether we need to mention that the size in characters isn't necessarily the value returned by `string.Length` is a judgement call...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
benwulfe: ok it looks like -1 is only used as a marker internally in some implementations of parameter -- so its ok to disallow on our side.
i've widened this check to disallow calls at all to "WithSize" if the typecode isnt "bytes" or "string" and have added a guard to the ctor when size is set to ensure it's not negative.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -233,6 +233,17 @@ internal SpannerConnectionStringBuilder CloneWithNewDataSource(string dataSource
                 DataSource = dataSource
             };
 
+        /// <summary>
+        /// Returns a new instance of a <see cref="SpannerConnectionStringBuilder"/> with the database
+        /// portion of the DataSource replaced with a new value.
+        /// </summary>
+        /// <param name="database">The new database name. Can be null to open a connection for Ddl commands.</param>
+        /// <returns>a new instance of <see cref="SpannerConnectionStringBuilder"/></returns>
```
iantalarico: nit: capital A

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -266,9 +268,17 @@ internal V1.Type ToProtobufType()
         /// <summary>
         /// Returns a clone of this type with the specified size constraint.
         /// </summary>
-        /// <param name="size"></param>
+        /// <param name="size">Represents the number of characters (for <see cref="TypeCode.String"/>)
+        ///  or bytes (for <see cref="TypeCode.Bytes"/>)</param>
         /// <returns></returns>
```
iantalarico: nit: empty returns

Issue #1591, opened 2017-10-27
Remove streaming gRPC cancellation workaround from SimpleSubscriber


chrisdunelm:
Now gRPC 1.7.0 is released, which supports cancelling within `MoveNext()` we can remove the workaround.

jskeet:
Am I right in saying this was done in #1683?

chrisdunelm:
Yes it was done in #1683, although unfortunately this is incorrect.
It really does need to use a `CancellationToken.None` in the `MoveNext()` call because cancelling `MoveNext()` cancels the entire streaming RPC. This is undesirable, because if it's a user-requested cancellation then there may still be further acks to send to the server over the same streaming RPC.
The next PR will fix this with a explanatory comment.

chrisdunelm:
It is required to not use a cancellation token in the call to `MoveNext()`. There is a comment in #1724 explaining why.

PR #1724, opened 2017-11-29
Further pubsub fixes, and a little more testing.

chrisdunelm:
Updates #1610
Updates #1591

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
@@ -44,7 +44,8 @@ namespace Google.Cloud.PubSub.V1.IntegrationTests
 
         private async Task RunBulkMessaging(
             int messageCount, int minMessageSize, int maxMessageSize, int maxMessagesInFlight, int initialNackCount,
-            TimeSpan? timeouts = null, int? cancelAfterRecvCount = null)
+            TimeSpan? timeouts = null, int? cancelAfterRecvCount = null, TimeSpan? interPublishDelay = null,
+            TimeSpan? debugOutputPeriod = null)
         {
             // Force messages to be at least 4 bytes long, so an int ID can be used.
             minMessageSize = Math.Max(4, minMessageSize);
@@ -52,8 +53,12 @@ namespace Google.Cloud.PubSub.V1.IntegrationTests
             var subscriptionId = _fixture.CreateSubscriptionId();
 
             Console.WriteLine("BulkMessaging test");
+            Console.WriteLine($"ProjectID: {_fixture.ProjectId}");
+            Console.WriteLine($"Topic ID: {topicId}");
+            Console.WriteLine($"Subscription ID: {subscriptionId}");
             Console.WriteLine($"{messageCount} messages; of size [{minMessageSize}, {maxMessageSize}]; " +
-                $"max in-flight: {maxMessagesInFlight}, initialNacks: {initialNackCount}, cancelAfterRecvCount: {cancelAfterRecvCount}");
+                $"max in-flight: {maxMessagesInFlight}, initialNacks: {initialNackCount}, " +
+                $"cancelAfterRecvCount: {cancelAfterRecvCount}, interPublishDelay: {interPublishDelay}");
 
             // Create topic
             var topicName = new TopicName(_fixture.ProjectId, topicId);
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
```diff
@@ -44,16 +44,21 @@ public SimplePubSubTest(PubsubFixture fixture)
 
         private async Task RunBulkMessaging(
             int messageCount, int minMessageSize, int maxMessageSize, int maxMessagesInFlight, int initialNackCount,
-            TimeSpan? timeouts = null, int? cancelAfterRecvCount = null)
+            TimeSpan? timeouts = null, int? cancelAfterRecvCount = null, TimeSpan? interPublishDelay = null,
+            TimeSpan? debugOutputPeriod = null)
```
jskeet: Our *three* main TimeSpans are timeouts, interPublishDelay, and debugOutputPeriod... (I'll come in again...)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
```diff
@@ -138,18 +143,19 @@ public SimplePubSubTest(PubsubFixture fixture)
             CancellationTokenSource watchdogCts = new CancellationTokenSource();
             Task.Run(async () =>
             {
+                var debugOutputPeriod1 = debugOutputPeriod ?? TimeSpan.FromSeconds(1);
                 int prevSentCount = -1;
                 int prevRecvCount = -1;
                 int noProgressCount = 0;
-                while (!watchdogCts.IsCancellationRequested)
+                while (true)
```
jskeet: Shouldn't there be some way to exit the loop without reporting deadlock? Or will the Task.Delay throw an exception which is then ignored, when watchdogCts.Token is cancelled by success?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SimpleSubscriberTest.cs
```diff
@@ -358,6 +366,58 @@ public void Dispose()
...
+        [Theory, CombinatorialData]
+        public void OneClientManyMsgs([CombinatorialValues(1, 2, 3, 4, 5)] int rndSeed)
+        {
+            // Test sending a large quantity of messages
+            const int maxStreamingPull = 14;
+            const int maxMoveNext = 14;
+            const int maxMsgCount = 80;
+            var rnd = new Random(rndSeed);
+            var streamingPullCount = rnd.Next(maxStreamingPull / 2, maxStreamingPull);
+            var msgs = Enumerable.Range(0, streamingPullCount).Select(rpcId =>
```
jskeet: I don't understand this bit, and I suspect I probably should...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SimpleSubscriber.cs
```diff
@@ -805,13 +803,15 @@ private void QueueLeaseExtensions(IEnumerable<string> ackIds)
...
-                    // Send acks/extensions
-                    await _taskHelper.ConfigureAwait(pull.WriteAsync(new StreamingPullRequest
+                    var req = new StreamingPullRequest
                     {
                         AckIds = { acks },
-                        ModifyDeadlineAckIds = { extends },
-                        ModifyDeadlineSeconds = { Enumerable.Repeat(_modifyDeadlineSeconds, extends.Count) },
-                    }));
+                        // Don't lease-extend any msg that is also being acked.
+                        ModifyDeadlineAckIds = { extends.Except(acks) }
```
jskeet: Might this be expensive? Is there an alternative approach we should consider?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
chrisdunelm: :)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.IntegrationTests/SimplePubSubTest.cs
chrisdunelm: Yes, watchdogCts is cancelled on test completion, and the exception thrown by Task.Delay is ignored.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SimpleSubscriberTest.cs
chrisdunelm: It's constructing a long-ish sequence of batches of messages; where all batches have varying sizes.
I'll add a comment

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SimpleSubscriber.cs
chrisdunelm: I've considered various alternatives, but I think this is the best compromise.
The size of both `acks` and `extends` are limited to 10,000 so this will never be too expensive.

PR #1726, opened 2017-11-29
More doc automation

jskeet:
I've built the Vision and Monitoring docs, and they look fine with this.

This should reduce the time taken to add a new API to less than 10 minutes if we're happy not to add any manual snippets.

### DIFF
--- a/tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
+++ b/tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
@@ -203,11 +203,45 @@ download a service account JSON file then set the `GOOGLE_APPLICATION_CREDENTIAL
 The credentials will automatically be used to authenticate. See the [Getting Started With
 Authentication](https://cloud.google.com/docs/authentication/getting-started) guide for more details.";
 
+            string clientClasses = CreateClientClassesDocumentation(api);
+
+            string clientConstruction =
+@"Create a client instance by calling the static `Create` method, optionally
+specifying an end-point or channel and settings";
+
             return text
                 .Replace("{{title}}", title)
                 .Replace("{{description}}", description)
                 .Replace("{{installation}}", installation)
-                .Replace("{{auth}}", auth);
+                .Replace("{{auth}}", auth)
+                .Replace("{{client-classes}}", clientClasses)
+                .Replace("{{client-construction}}", clientConstruction);
+        }
+
+        private static string CreateClientClassesDocumentation(ApiMetadata api)
+        {
+            if (api.Type != "grpc")
+            {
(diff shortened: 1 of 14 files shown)


tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
```diff
@@ -203,11 +203,45 @@ private static string TransformDocTemplate(ApiMetadata api, string text)
...
+        {
+            if (api.Type != "grpc")
+            {
+                return "FIXME"; // No automatic templating for this API
+            }
+            var layout = DirectoryLayout.ForApi(api.Id);
+            var packageSource = Path.Combine(layout.SourceDirectory, api.Id);
+            var sourceFiles = Directory.GetFiles(packageSource, "*Client.cs");
+            var clients = sourceFiles
+                .Where(file => File.ReadAllText(file).Contains(": ServiceSettingsBase")) // Check it contains a generated client
```
chrisdunelm: A tad crude/unreliable perhaps? Maybe a TODO to improve robustness of generated client detection?

tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
```diff
@@ -203,11 +203,45 @@ private static string TransformDocTemplate(ApiMetadata api, string text)
...
+                .Select(file => file.Substring(0, file.Length - 3)) // Trim .cs
+                .OrderBy(client => client)
+                .Select(client => $"[{client}](obj/api/{api.Id}.{client}.yml)") // Markdown link to API doc
+                .ToList();
+            switch (clients.Count)
+            {
+                case 0: return "FIXME"; // No automatic templating for this API
+                case 1: return $"All operations are performed through {clients[0]}.";
+                default:
+                    var list = string.Join("\r\n", clients.Select(client => $"- {client}"));
```
chrisdunelm: Is "\r\n" definitely required? Rather than `Environment.NewLine`?

tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
jskeet: Will add a TODO, but I strongly suspect it'll be accurate enough forever :)

tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
jskeet: I wondered about that. Ultimately, the rest of the text has line breaks of "however it was checked out from GitHub" which is kinda hard to replicate. I thought that this would at least be consistent and obvious. I suspect it won't matter anyway and git will canonicalize one way or another :)

PR #1727, opened 2017-11-29
Disable PubSub tests, these tests are hanging about 50% of the time.

iantalarico:
This causes our CI to hang.

### DIFF
--- a/runintegrationtests.sh
+++ b/runintegrationtests.sh
@@ -74,7 +74,7 @@ fi
 
 for testdir in $testdirs
 do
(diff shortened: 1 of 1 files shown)


jskeet:
@chrisdunelm submitted some test changes earlier - perhaps they'll fix this?

chrisdunelm:
@iantalarico If we could wait to see if PR #1724 fixes this, I'd appreciate it.
I've run the integration tests many times on my machine, and it never just hangs. So if this continues I may need to try running them manually on the jenkins machine - is this possible?

iantalarico:
@chrisdunelm I'll revert for now then.

Yea you can just RDP or SSH into a machine, clone the repo and run tests there.

PR #1732, opened 2017-11-30
Parameter validation tests; default settings update

chrisdunelm:
Default settings updated to make them more suitable for limited-bandwidth deployments, this will have no affect on the speed capabilities in high-bandwidth deployments


chrisdunelm:
Oops, I'll fix the CI failures momentarily...

PR #1733, opened 2017-11-30
Tweek ack/extend send size to improve delivery rate

chrisdunelm:
(This will have only _maxAckExtendCount changes in SimpleSubscriber.cs once #1732 has been merged)

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SimpleSubscriber.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SimpleSubscriber.cs
@@ -557,7 +557,7 @@ namespace Google.Cloud.PubSub.V1
                 _handlerAsync = handlerAsync;
                 _autoExtendInterval = subscriber._autoExtendInterval;
                 _modifyDeadlineSeconds = subscriber._modifyDeadlineSeconds;
-                _maxAckExtendQueue = subscriber._maxAckExtendQueue;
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SimpleSubscriber.cs
```diff
@@ -357,15 +357,15 @@ internal SimpleSubscriberImpl(SubscriptionName subscriptionName, IEnumerable<Sub
             _scheduler = settings.Scheduler ?? SystemScheduler.Instance;
             _taskHelper = GaxPreconditions.CheckNotNull(taskHelper, nameof(taskHelper));
             _flowControlSettings = settings.FlowControlSettings ?? DefaultFlowControlSettings;
-            _maxAckExtendCount = (int)Math.Min(_flowControlSettings.MaxOutstandingElementCount ?? long.MaxValue, 10_000);
+            _maxAckExtendQueue = (int)Math.Min(_flowControlSettings.MaxOutstandingElementCount ?? long.MaxValue, 20_000);
```
jskeet: Why the rename? It does sound like a count.

chrisdunelm: It used to be the count of ack/extend messages pushed within each batch to the server.
It's now the max size of the queue on the client. It should really be called `maxAckExtendQueueSize`. I'll change it then push on green.

PR #1737, opened 2017-12-01
Fix Storage build

jskeet:
The PubSub IAM reference was a package reference, which confused
things when Storage had a project reference.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -473,7 +473,7 @@
       "Google.Apis.Storage.v1": "1.30.0.1014"
     },
     "testDependencies": {
-      "Google.Cloud.Iam.V1": "project",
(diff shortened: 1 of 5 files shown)


apis/apis.json
```diff
@@ -473,7 +473,7 @@
       "Google.Apis.Storage.v1": "1.30.0.1014"
     },
     "testDependencies": {
-      "Google.Cloud.Iam.V1": "project",
+      "Google.Cloud.Iam.V1": "1.0.0-beta13",
```
chrisdunelm: beta15 is the latest

jskeet: Nope, not for IAM - PubSub is beta15, but this line is about IAM :)

PR #1739, opened 2017-12-01
De-flake Diagnostics Logging Tests

iantalarico:
Look only for specific severity levels to avoid picking up unwanted info logs which can mess up log counts

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
@@ -298,7 +298,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
         public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
         {
             SetupRoutes(app);
-            var options = BufferOptions.TimedBuffer(TimeSpan.FromSeconds(10));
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
```diff
@@ -297,7 +298,7 @@ public class TimedBufferWarningLoggerTestApplication : LoggerTestApplication
         public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
         {
             SetupRoutes(app);
-            var options = BufferOptions.TimedBuffer(TimeSpan.FromSeconds(20));
+            var options = BufferOptions.TimedBuffer(TimeSpan.FromSeconds(10));
```
benwulfe: would it be better to manually flush at test completion rather than rely on a timer?

benwulfe:
overall lgtm.  the timing of the log flushes might be better fixed with a direct call to flush, but YMMV.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
iantalarico: This was an accidental change.  I reverted it.  Yes that's probably a better way to go about this though.  I'll give that a try later.

PR #1755, opened 2017-12-06
Public doc of backlog issues

jskeet:
These are issues that aren't on our radar at the moment.

The first pass has left quite a few not covered here, as I'd really
like to see them just get done.

Will close these issues after merging this commit.

(I'm assuming the `#` approach of linking to issues will work here - if not, I'll create a follow-up PR.)

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -1,0 +1,64 @@
+# Requested enhancements backlog
+
+This file is used to keep track of feature requests that we *may*
+want to come back to at some point, but aren't on our immediate
+radar. Using this single document is simpler than keeping all the
+corresponding issues open, but we'll reopen the issue if we revisit
+the request and want to put work into it.
+
+## Static code analysis
+
+- Issue: #1709
+- Status: Further investigation required. We already have
+  some analysis in the form of custom Roslyn analyzers.
+
+## Metadata client failure
+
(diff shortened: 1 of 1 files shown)


BACKLOG.md
```diff
@@ -0,0 +1,63 @@
+# Requested enhancements backlog
+
+This file is used to keep track of feature requests that we *may*
+want to come back to at some point, but aren't on our immediate
+radar. Using this single document is simpler than keeping all the
+corresponding issues open, but we'll reopen the issue.
```
chrisdunelm: End of this sentence doesn't make sense

BACKLOG.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Static code analysis
+
+- Issue: #1709
+- Status: Further investigation required. We already have
+  some analysis in the form of custom Roslyn analyzers.
+
+## Metadata client failuer
```
chrisdunelm: failuer failure ;)

BACKLOG.md
jskeet: Indeed - I'm sure it was longer than that before. Fixed.

BACKLOG.md
jskeet: Oops. Fixed.

PR #1758, opened 2017-12-06
Do not use buffers for the ASP.NET trace integration tests

iantalarico:
The buffers are timed and will cause a delay leading to flakes.  The buffers are well tested in the Common and AspNetCore packages.


iantalarico:
Fixed, thanks Jon!

PR #1762, opened 2017-12-07
Disable parallelization on the AspNet Integration tests

iantalarico:
They are running in parallel.  The tests operate on "HttpContext.Current" which is global (we can't use an actual IIS instance), so having multiple tests touch it which is not a good idea.  Having the tests not be in parallel shouldn't be an issue as we only have 4.


iantalarico:
@jskeet The time of 0 isn't something on our end.  This is in the API, I'm not sure we can do anything (besides file a bug).

If I call:
var t0 = DateTime.UtcNow.Ticks;
var t1 = DateTime.UtcNow.Ticks;

t0 and t1 are equal. 

Thoughts?

jskeet:
I think we should check with the API team, but if we really can't report traces that are "briefer than we can detect" maybe we should just explicitly ensure that we specify at least one time unit? I think that's better than causing an error if a user *does* create a very brief trace. (Imagine something that's sometimes a cache lookup and sometimes not.)

iantalarico:
@jskeet Talked with the trace team and confirmed this.  I trimmed down this PR and will send another shortly.

PR #1763, opened 2017-12-07
Ensure each Trace Span always has at least 1 ms of time.

iantalarico:
If a span has less than 1ms of time the Trace API will not record it.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
@@ -76,6 +76,9 @@ namespace Google.Cloud.Diagnostics.Common
             public ulong SpanId() => TraceSpan.SpanId;
         }
 
+        /// <summary>The number of nanoseconds in a millisecond.</summary>
+        internal static readonly int _nanosecondsInAMillisecond = 1000000; // 1,000,000
+
         /// <summary>The trace consumer to push the trace to when completed.</summary>
         private readonly IConsumer<TraceProto> _consumer;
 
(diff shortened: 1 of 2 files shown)


iantalarico:
@jskeet  I wrote the tests first and ran them to ensure we had failures.

I'm also not super happy with how I add the milliseconds to the timespan but couldn't find a cleaner way unless I did a lot of conversions.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -76,6 +76,9 @@ public void AnnotateSpan(Dictionary<string, string> labels)
             public ulong SpanId() => TraceSpan.SpanId;
         }
 
+        /// <summary>The number of nanoseconds in a millisecond.</summary>
+        internal static readonly int _nanosecondsInAMillisecond = 1000000; // 1,000,000
```
jskeet: Rewrite as 1_000_000 and then you don't need the comment :)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
```diff
@@ -203,6 +206,14 @@ private void EndSpan(Span span)
 
             lock (_traceLock)
             {
+                // If the time between the start and end of a span is less then 1 ms
+                // the span will not be recorded.  If this is the case bump up
+                // the time by 1 ms which is the smallest amount of time for the Trace API
+                // to record a trace. 
+                if ((span.TraceSpan.EndTime.Nanos - span.TraceSpan.StartTime.Nanos) < _nanosecondsInAMillisecond)
```
jskeet: Is span.TraceSpan.EndTime a Timestamp? If so, this won't be right - the nanos value could be out of range, and there could be a whole day of difference anyway. (Additionally, it won't catch the case where the two timestamps are less than a millisecond apart, but on different days.)

If it *is* a Timestamp, it might be worth assigning a bug to me and I'll look at it tomorrow.

iantalarico:
I should have read that description more clearly.

I was worried about the precision of the nanos but realized we don't need to worry about it as all the timestamps are coming from datetimes.  Which actually makes this much cleaner IMO.

PTAL

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/SimpleManagedTracer.cs
iantalarico: I feel like you tell me this every 2-3 months and then I forget :-P

iantalarico:
@jskeet Faster is better :)  Will this require changes outside this lib?  If so I'll merge now but if we can make this faster as is I'll hold off.

jskeet:
Nope, it should all be within this library. Could you create an issue to make sure I don't forget between now and tomorrow?

iantalarico:
@jskeet Done.  See #1764

iantalarico:
closing in favor of #1765

PR #1770, opened 2017-12-10
[Bigtable] Ensure lexicographical order of families in row merger

evildour:
They are returned based on the order of some internal id and need to be reordered on the client.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
@@ -17,7 +17,6 @@ using Google.Protobuf;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowAsyncEnumerator.cs
```diff
@@ -33,6 +33,7 @@ namespace Google.Cloud.Bigtable.V2
     internal sealed class RowAsyncEnumerator : IAsyncEnumerator<Row>
     {
         private CellInfo _currentCell;
+        private SortedList<string, Family> _currentFamilies = new SortedList<string, Family>(StringComparer.Ordinal);
```
jskeet: Could this be readonly? It looks like it could.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
```diff
@@ -293,6 +292,57 @@ private class TestBigtableClient : BigtableClient { }
...
+            int rowCount = 0;
+            await stream.AsAsyncEnumerable().ForEachAsync(row =>
+            {
+                rowCount++;
+                var familyNames = row.Families.Select(r => r.Name).ToArray();
+                Assert.Equal("A", familyNames[0]);
+                Assert.Equal("a", familyNames[1]);
+                Assert.Equal("z", familyNames[2]);
+            });
+            Assert.Equal(1, rowCount);
```
jskeet: Add an assertion that each family has the right value associated with it as well?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowAsyncEnumerator.cs
evildour: Yup, done.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
evildour: Done. Also added a 'Z' family to verify we're really getting ordinal and not invariant sorting.

evildour:
Thanks Jon, will merge when green

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
jskeet: Aha. I never know what case-sensitive but invariant sorting does with things like this. Thanks for adding the test :)

PR #1771, opened 2017-12-10
BigtableClient.MutateRows

evildour:


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsStreamExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+    public static class MutateRowsStreamExtensions
+    {
+        /// <summary>
+        /// Gathers and orders the entries from all responses from the stream.
+        /// The order will match the order of the associated request entries.
+        /// </summary>
+        public static async Task<IEnumerable<MutateRowsResponse.Types.Entry>> GetResponseEntries(
+            this BigtableClient.MutateRowsStream stream)
+        {
+            var responses = new List<MutateRowsResponse>();
```
jskeet: Does `var responses = await stream.ResponseStream.ToList()` not work?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Utilities.cs
```diff
@@ -46,10 +46,10 @@ public static ByteString Concat(this ByteString left, ByteString right)
         public static IEnumerable<T> ValidateCollection<T>(
```
jskeet: Side note: It might be worth making all these methods internal - the class itself is, but in code reviews you can't always see the class declaration.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
```diff
@@ -195,6 +195,50 @@ private class TestBigtableClient : BigtableClient { }
                 () => client.MutateRowAsync(tableName, rowKey, mutations, CallSettings.FromCancellationToken(default)));
         }
 
+        [Fact]
+        public void MutateRows_Valid_Request()
+        {
+            var client = new TestBigtableClient();
+            var tableName = new TableName("project", "instance", "table");
+
+            // NotImplementedException means it got through the normal validations and tried to actually
```
jskeet: Not for this PR, but maybe it would be useful to have something that through a test-only "MadeItToServiceException" to avoid this sort of comment being worthwhile. (I think we'd want that in generic test infrastructure, not Bigtable-specific.)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
```diff
@@ -0,0 +1,230 @@
...
+                new BigtableVersion(1));
+        }
+
+        [Fact]
+        public async Task DeleteFromColumn()
+        {
+            var tableName = _fixture.TableName;
+            var client = _fixture.TableClient;
+            var rowKey = await _fixture.InsertRowAsync(tableName, version: new BigtableVersion(1));
+            client.MutateRow(
```
jskeet: So just to check, MutateRow is synchronous - whereas MutateRows is somewhat-asynchronous because it returns a stream? Not sure there's anything we can do about that, but it may be worth us highlighting in the docs.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
```diff
@@ -0,0 +1,230 @@
...
+                    "abcd",
+                    new BigtableVersion(2)));
+
+            var stream = client.MutateRows(
+                tableName,
+                Mutations.CreateEntry(
+                    rowKey,
+                    Mutations.DeleteFromColumn(
+                        BigtableFixture.ColumnFamily1,
+                        "row_exists",
```
jskeet: This is the column created in InsertRowAsync, yes? Is it worth extracting that to a constant for clarity?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
```diff
@@ -0,0 +1,230 @@
...
+                    BigtableFixture.OtherColumnFamily,
+                    "last_name",
+                    "Smith",
+                    new BigtableVersion(3)));
+
+            var stream = client.MutateRows(
+                tableName,
+                Mutations.CreateEntry(
+                    rowKey,
+                    Mutations.DeleteFromFamily(BigtableFixture.ColumnFamily1)));
```
jskeet: So DeleteFromFamily is "delete all columns within that column family"? Makes sense if so - just checking.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
```diff
@@ -0,0 +1,230 @@
...
+
+            var entries = await stream.GetResponseEntries();
+            Assert.True(entries.All(e => e.Status.Code == (int)Code.Ok));
```
jskeet: Check, but I believe there's an Assert.All method which is preferable here.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
```diff
@@ -0,0 +1,230 @@
...
+        {
+            var tableName = _fixture.TableName;
+            var client = _fixture.TableClient;
+            var rowKey = await _fixture.InsertRowAsync(tableName);
+
+            var stream = client.MutateRows(
+                tableName,
+                Mutations.CreateEntry(
+                    rowKey,
+                    Mutations.DeleteFromRow()));
```
jskeet: So is this effectively "clear (but don't delete) row" or is there some more subtlety, e.g. that you can apply a column filter over the top?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
```diff
@@ -0,0 +1,230 @@
...
+
+        [Fact]
+        public async Task MultipleRows()
+        {
+            var tableName = _fixture.TableName;
+            var client = _fixture.TableClient;
+            var rowKey1 = await _fixture.InsertRowAsync(tableName);
+            BigtableByteString rowKey2 = Guid.NewGuid().ToString();
+
+            var stream = client.MutateRows(
```
jskeet: What happens if I have a third entry that uses rowKey2?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsStreamExtensions.cs
evildour: `ResponseStream` is an `IAsyncEnumerator<T>`, not `IAsyncEnumerable<T>`. Is there a `ToList()` that works with the former?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
evildour: Yes, that's correct. I'm curious: what is the issue with that? And I guess more importantly: what do you think should be highlighted in the docs?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsStreamExtensions.cs
jskeet: Ah, missed that. No, you're right.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
jskeet: Well, one might assume that the two behave very similarly - that if `client.MutateRow(mutation)` behaves synchronously, then `client.MutateRows(new[] { mutation })` will as well. It's not clear to me whether `MutateRows(...)` makes the initial call synchronously (so will throw exceptions if there's anything invalid) or not, for example.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
evildour: Yes, that's correct.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
evildour: I thought so too. Unfortunately, `Assert.All` takes actions, not predicates, so I'd have to do something like this:
```csharp
Assert.All(entries, e => Assert.Equal(e.Status.Code, (int)Code.Ok));
```
I think the other way is slightly cleaner, but if you disagree I'll change it.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
evildour: I believe it is effectively "clear _and delete_ row". There is no way to filter it.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
evildour: Good question. It should merge the row edits. Updated the test to verify.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableClientTest.cs
evildour: Good idea. It's not too much churn, so I'll do it specific to this for now, but yes perhaps something common would be good in the future.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Utilities.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
evildour: Makes sense. I added something, but let me know if you think more is needed.

evildour:
Thanks Jon, PTAL.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
```diff
@@ -32,7 +32,9 @@ public class BigtableFixture : IDisposable, ICollectionFixture<BigtableFixture>
         private const string TestInstanceEnvironmentVariable = "BIGTABLE_TEST_INSTANCE";
         private const string TestProjectEnvironmentVariable = CloudProjectFixtureBase.TestProjectEnvironmentVariable;
 
-        public const string ColumnFamily1 = "cf1";
+        public const string DefaultColumnFamily = "cf1";
+        public const string DefaultQualifier = "row_exists";
```
jskeet: I'm still somewhat confused about what a qualifier is, but I think that's Bigtable-ness rather than anything in this API... is "row_exists" a useful constant outside tests? Is it something we should put in the public API anywhere?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -425,6 +425,10 @@ public partial class BigtableClient
         /// </summary>
         /// <remarks>
         /// <para>
+        /// Note this method executes a server streaming RPC. It is not a synchronous call. The caller must process the responses from
```
jskeet: +1 - thanks for this.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
evildour: There's no significance here. It could have just been called "foo". I figured it might be good to have some meaning for the default case (kind of like how in one of the the `CheckAndMutateRow` tests, I use a column name of "original_row_status" and then values or "row_had_cells" and "row_had_no_cells"). If you think there's too much opportunity for confusion though, I'm happy to change them to more nonsensical things.

jskeet: I think it's fine - I'll just read up on what qualifiers are in the context of Bigtable.

evildour: Basically, as I understand it, column families are what you would think of as columns in a relational DB table: they are fixed and relatively expensive to change. Within families, you have cells with arbitrary qualifier/value byte strings, like key/value pairs (they are really key/value/version tuples though because you can update the value associated with a qualifier and the history of old values is maintained). This means qualifiers can contain data as well as the values, like in [this example](https://cloud.google.com/bigtable/docs/overview#storage-model) from the Bigtable docs.

jskeet: Aha. I'd just think of that as a column name, scoped to the column family. Right - good to know what's going on, thanks.

evildour: Yes, sometimes they're called 'column qualifiers' and other times they're just called 'qualifiers'. I've kind of settled on the latter, but I'll rename this to `DefaultColumnQualifier` so its clear.

evildour:
(rebased and merged)

PR #1786, opened 2017-12-15
Update new API documentation

jskeet:


NEWAPI.md
```diff
@@ -131,22 +123,24 @@ You'll typically want JSON like this:
 Fix everything with "FIXME". There's no set number of tags, but these are used for the NuGet package,
 so consider what users will search for.
 
-The version number of "1.0.0-alpha00" is used to indicate that there
-hasn't been a release of this API yet. It will be incremented to
-`1.0.0-alpha01` as part of the first release.
+The above assumes you're happy to create an initial beta release for
+the generated code immediately. As of late 2017, that's usually
+fine. If you want to avoid creating a release, use a version of
+`1.0.0-alpha00` or `1.0.0-beta00` depending on whether or not you
```
ILMTitan: [Nuget 4.3.0 supports Semver 2.0](https://docs.microsoft.com/en-us/nuget/reference/package-versioning#semantic-versioning-200) which would let us use dotted prerelease version numbers like `1.0.0-alpha.1`.

jskeet: Mmm... Not sure how much benefit that will give compared with the cost of changing conventions tbh. Will bear it in mind.

PR #1787, opened 2017-12-18
 BigtableClient.SampleRowKeys helpers and tests

evildour:


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -829,6 +829,35 @@ public partial class BigtableClient
...
+        /// <param name="tableName">
+        /// The unique name of the table from which to sample row keys. Must not be null.
+        /// </param>
+        /// <param name="callSettings">
+        /// If not null, applies overrides to this RPC call.
+        /// </param>
+        /// <returns>
+        /// The server stream with the row key samples.
+        /// </returns>
+        public virtual SampleRowKeysStream SampleRowKeys(
```
jskeet: Is TableName a resource name class here? If so, this might be better achieved by changing the GAPIC config and getting codegen to do it. Otherwise we'll have a problem when we *do* do that. (There may be something I've missed, of course...)

evildour: No you're right. That seems like a better way, although I'm not that familiar with the GAPIC config structure. In the [current config](https://github.com/evildour/googleapis/blob/master/google/bigtable/v2/bigtable_gapic.yaml#L56), would that just mean setting request_object_method to true?

jskeet: I think it's a request_flattening, but @chrisdunelm would know better.

evildour: I'm not so sure. Some of the other methods, such as ReadModifyWriteRow, are generated with separate parameters and they doesn't have request_flattening set.

chrisdunelm: I'm fairly sure this is because `SampleRowKeys` is a server-streaming method, and we don't currently generate flattened methods for server-streaming methods.

So the config is correct, in that it specifies that a flattening *should* be generated, but the C# code-gen doesn't yet do it.
I had not really realised this is missing from the C# codegen.

I suggest that this method is *not* added here, and we (I) fix the codegen as soon as possible.

evildour: What if I just add a TODO on here to remove it once the generator is updated and we're ready to regenerate Bigtable? I'm trying to get the client cleanup finished in Q4 if possible, and we're getting close to the end here :)

chrisdunelm: Hmm, OK, this is still in alpha, so I guess it's OK. I'll create a release-blocking issue to make sure this is fixed by codegen before release.

PR #1795, opened 2017-12-27
Allow return valid document for ValueSerializer

Thaina:
I have my own JsonConverter to convert json value into google protobuf value. And I want to have it work with `FirestoreDb` api

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
@@ -114,8 +114,11 @@ namespace Google.Cloud.Firestore
         /// This is effectively the map-only part of <see cref="Serialize"/>, but without wrapping the
         /// result in a Value.
         /// </summary>
-        internal static Dictionary<string, Value> SerializeMap(object value)
+        internal static IDictionary<string, Value> SerializeMap(object value)
         {
+            if(value is IDictionary<string, Value> dict)
+                return dict;
+
             GaxPreconditions.CheckNotNull(value, nameof(value));
             switch (value)
             {


jskeet:
Thank you for your contribution. I will look at this when I return to work on January 2nd.

jskeet:
(Although just looking at the stats, I very much doubt that I'll be accepting a pull request affecting over 300 files across so many projects. If this is due to reformatting etc, please undo those changes.)

Thaina:
@jskeet I change only 4 lines of codes

jskeet:
Ah yes - it was codecov. Will look later - as it's small, I may be around to do that tonight.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
```diff
@@ -114,8 +114,11 @@ internal static Value Serialize(object value)
         /// This is effectively the map-only part of <see cref="Serialize"/>, but without wrapping the
         /// result in a Value.
         /// </summary>
-        internal static Dictionary<string, Value> SerializeMap(object value)
+        internal static IDictionary<string, Value> SerializeMap(object value)
         {
+            if(value is IDictionary<string, Value> dict)
```
jskeet: Please just add this to the switch statement - that would be much more in-keeping with the rest of the code.

There's another aspect to consider though: in some cases, I believe we modify the values returned by this method; we could make this method clone each dictionary value if it's passed an `IDictionary<string, Value>` perhaps.

It's possible that I don't need to do that modification - some things have changed over time. Please make the suggested change so it's *potentially* ready to merge. I need to look at the rest of the code more thoroughly when I'm back at work to check. (A lot of this reflection-based code is likely to change anyway for the sake of performance at some point, admittedly.)

PR #1796, opened 2017-12-28
Allow return valid document for ValueSerializer

Thaina:
I have my own JsonConverter to convert json value into google protobuf value

And so I want to have raw value work with FirestoreDb api

ps: I don't know why I can't continue edit the previous pull request so I need to make new one

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
@@ -119,8 +119,10 @@ namespace Google.Cloud.Firestore
             GaxPreconditions.CheckNotNull(value, nameof(value));
             switch (value)
             {
+                case IDictionary<string, Value> map:
+                    return map.ToDictionary(pair => pair.Key, pair => pair.Value);
                 case IDictionary<string, object> map:
-                    return map.ToDictionary(pair => pair.Key, pair => Serialize(pair.Value));
+                    return map.ToDictionary(pair => pair.Key, pair => (pair.Value as Value) ?? Serialize(pair.Value));
                 case object anon when IsAnonymousType(anon.GetType()):
                     return ConvertAnonymousType(anon);
                 case object attributed when IsFirestoreAttributedType(attributed.GetType()):


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
```diff
@@ -119,8 +119,10 @@ internal static Value Serialize(object value)
             GaxPreconditions.CheckNotNull(value, nameof(value));
             switch (value)
             {
+                case IDictionary<string, Value> map:
+                    return map.ToDictionary(pair => pair.Key, pair => pair.Value);
                 case IDictionary<string, object> map:
-                    return map.ToDictionary(pair => pair.Key, pair => Serialize(pair.Value));
+                    return map.ToDictionary(pair => pair.Key, pair => (pair.Value as Value) ?? Serialize(pair.Value));
```
evildour: It might be better to instead update `Serialize` to have a case for `Value` that just returns the `value` parameter and then leave this line alone. That way any other type/collection that already is/contains `Value` instances will work similarly. @jskeet would know better if that's a good idea or not though.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
```diff
@@ -119,8 +119,10 @@ internal static Value Serialize(object value)
             GaxPreconditions.CheckNotNull(value, nameof(value));
             switch (value)
             {
+                case IDictionary<string, Value> map:
+                    return map.ToDictionary(pair => pair.Key, pair => pair.Value);
```
evildour: Why not just return `map` here so you don't need to create a new collection?

evildour:
You may want to add a test or two in [ValueSerializerTest.cs](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ValueSerializerTest.cs) for this support.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
Thaina: @evildour It was a comment from my previous request

https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/1795#issuecomment-354103718

evildour: I see. Thanks.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
Thaina: @evildour I am not sure about this too, so I would wait for @jskeet too

PR #1797, opened 2017-12-28
Request manager

AVaksman:
RequestManager for MutateRows
RequestManager for ReadRows
Unit tests for both


evildour:
Is this ready for review or still WIP? We should probably discuss this a bit more offline, but just looking at the overall approach a bit, I don't feel this is the right way to go, for a couple of reasons:

1. The request managers are public. I had planned on making the retry logic a bit more transparent. Perhaps a method called on the response stream which returns another stream that retries transparently, such as `var stream = client.MutateRows(...).WithRetries()`, or perhaps an additional option specified to `MutateRows`, but in a way that works with the autogenerated methods.
2. It appears the onus is on the caller to drive the process of letting the manager know what the responses are and then getting and performing the subsequent requests for items which can be retried. Like I said, I'd like things to be a bit more transparent.

If instead this is meant to just be utility code for what we eventually expose to the user, then please mark these request managers internal instead of public.

AVaksman:
The idea and specs are taken from the approach in Java client. [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/MutateRowsRequestManager.java) and [here](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/ReadRowsRequestManager.java)
RetryManager meant to be a utility that will assist smart retries and meant to be used by the utility in charge of the retry logic. 

The idea was laid out so that user would just call existing `MutateRows(...)` or `ReadRows(...)` method. All the retry logic will be 'magically' happening behind the seance and the user would just get a response stream after all retries (if any) where done. So user doesn't know whether any retries even happened for any particular call. User just calls on a method and thing get done.
We should discuss it offline to tailor it the way it should be.

sduskis:
@evildour, the java implementation turned out to be quite a bit of code, so I figured there should be a separation of concerns.  The overall goal is still transparent retries from the user's perspective.

This PR represents the core business logic of maintaining the state of the request / response state  happens in one class and is well tested.  As you can see, there's quite a bit of non-trivial (and somewhat convoluted) logic just to create new requests.  FWIW, the convoluted logic is based on the Java implementations, so I take blame for the approach.

I figured that you, @evildour, would work on the grpc/retry framework integration separately.

evildour:
Ok that makes sense. I still have to read up a bit more on what's required for the support to comment on the big picture here (although if it is already working in Java, I'm sure I don't need to comment too much on it). I'll review the PR now though and then perhaps make small adjustments later when support is added publicly.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -0,0 +1,254 @@
...
+        {
+            _modifiedRequest = new MutateRowsRequest();
+            _mapToOriginalIndex = new int[indiciesToRetry.Count];
+            for (int i = 0; i < _mapToOriginalIndex.Length; i++)
+            {
+                _mapToOriginalIndex[i] = indiciesToRetry[i];
+                _modifiedRequest.Entries.Add(_originalRequest.Entries[indiciesToRetry[i]]);
+            }
+            return _modifiedRequest;
+        }
```
evildour: This can be simplified to:
```csharp
private MutateRowsRequest CreateRetryRequest(List<int> indiciesToRetry)
{
    _mapToOriginalIndex = indiciesToRetry.ToArray();
    _modifiedRequest = new MutateRowsRequest { Entries = { indiciesToRetry.Select(i => _originalRequest.Entries[i]) } };
    return _modifiedRequest;
}
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -0,0 +1,254 @@
...
+        {
+            List<MutateRowsResponse.Types.Entry> entries = new List<MutateRowsResponse.Types.Entry>();
+
+            for (int i = 0; i < Results.Length; i++)
+            {
+                Rpc.Status status = Results[i] == null ? STATUS_INTERNAL : Results[i];
+                entries.Add(CreateEntry(i, status));
+            }
+            return new MutateRowsResponse { Entries = { entries } };
+        }
```
evildour:
```csharp
public MutateRowsResponse BuildResponse() =>
    new MutateRowsResponse
    {
        Entries = { Results.Select((result, i) => CreateEntry(i, result ?? STATUS_INTERNAL)) }
    };
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+            for (int i = 0; i < size; i++)
+            {
+                int comparison = key1[i].CompareTo(key2[i]);
+                if (comparison != 0)
+                {
+                    return comparison;
+                }
+            }
+            return size1.CompareTo(size2);
+        }
```
evildour: This is not needed if you use `BigtableByteString` as suggested below.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        // Memeber variables form the constructor.
```
evildour: Nit: "Memeber" => "Member", "form" => "from"

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+        private readonly ReadRowsRequest _originalRequest;
+
+        // The number of rows read so far.
+        private long _rowCount;
+
+        /// <summary>
+        /// Constructor for ResumingStreamingResultScanner.
```
evildour: "ResumingStreamingResultScanner" => "BigtableReadRowsRequestManager". Use `<see>` tags to keep things in sync.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// </summary>
+        /// <param name="originalRequest">
+        /// Original ReadRowsRequest containg all of the parameters of the API call.
```
evildour: Nit: "containg" => "containing"

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// </param>
+        public BigtableReadRowsRequestManager(ReadRowsRequest originalRequest)
+        {
+            _originalRequest = originalRequest;
+        }
```
evildour: Use expression-bodied member.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        /// <summary>
+        /// Update last found Rowkey
+        /// </summary>
+        public ByteString LastFoundKey { get; set; }
```
evildour: Make this `BigtableByteString`

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+        // The number of rows read so far.
+        private long _rowCount;
+
+        /// <summary>
+        /// Constructor for ResumingStreamingResultScanner.
+        /// </summary>
+        /// <param name="originalRequest">
+        /// Original ReadRowsRequest containg all of the parameters of the API call.
+        /// </param>
+        public BigtableReadRowsRequestManager(ReadRowsRequest originalRequest)
```
evildour: Make this and all other public members internal.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+        /// </summary>
+        public ByteString LastFoundKey { get; set; }
+
+        /// <summary>
+        /// Update amount of rows read so far
+        /// </summary>
+        public long RowCount
+        {
+            get => _rowCount;
+            set => _rowCount += value;
```
evildour: Using the property setter as an increment instead of a normal set is will not be expected by callers and might cause some maintenance issues. I suggest making this a readonly property: `internal long RowCount { get; private set; }` and then having a separate `IncrementRowCount(long value)` method.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        /// <summary>
+        /// Builds and returns updated subrequest
+        /// </summary>
+        /// <returns></returns>
```
evildour: Remove empty returns or provide description.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+        public ReadRowsRequest BuildUpdatedRequest()
+        {
+            ReadRowsRequest newReadRowsRequest = new ReadRowsRequest()
+
```
evildour: Nit: remove newline and trailing `()`

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                TableNameAsTableName = _originalRequest.TableNameAsTableName,
+            };
+
+            #region Set up RowFilter
+
+            if (_originalRequest.Filter != null)
+            {
+                newReadRowsRequest.Filter = new RowFilter();
+                newReadRowsRequest.Filter = _originalRequest.Filter;
```
evildour: This creates a new `RowFilter` instance and then immediately discards it. I think the whole if statement should be replaced by this one line and then the region should be removed.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+            }
+
+            #endregion
+
+            #region Set up RowsLimit
```
evildour: I don't think these regions are necessary. They just add clutter here. If this was enough code where regions were necessary, I think breaking things up into helper methods would be preferable instead.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            // If the row limit is set, update it.
+            long numRowsLimit = _originalRequest.RowsLimit;
+
+            // Update <see cref="numRowsLimit"/> by removing the number of rows already read.
+            if ((numRowsLimit -= RowCount) > 0)
+            // The remaining number of rows must be greater than 0.
```
evildour: Nit: I would move the comment above the if condition so it doesn't break up the block.

However, bigger issue: I'm assuming this comment means we would never get into a situation where the number of remaining rows is zero, right? If so, I would be a bit more _assertive_ here (apologies for the attempt at wordplay humor):
```csharp
if (_originalRequest.RowsLimit != 0)
{
    long rowsRemaining = _originalRequest.RowsLimit - RowCount;
    Debug.Assert(rowsRemaining > 0, "The remaining number of rows must be greater than 0.");
    newReadRowsRequest.RowsLimit = Math.Max(1, rowsRemaining);
}
```
That way if the assumption is ever wrong for some unforeseen reason, we'll realize it when debugging and in release builds we won't start firing off followup ReadRows requests that are unlimited.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+            {
+                newReadRowsRequest.RowsLimit = numRowsLimit;
+            }
+
+            #endregion
+
+            #region Set up RowSet
+
+            newReadRowsRequest.Rows = new RowSet();
+            newReadRowsRequest.Rows = FilterRows();
```
evildour: Again, here a `RowSet` is created and then discarded. Remove the 1st line (and region).

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+
+            RowSet newRowSet = new RowSet();
+
+            foreach (ByteString key in originalRows.RowKeys)
+            {
+                if (!StartKeyIsAlreadyRead(key))
+                {
+                    newRowSet.RowKeys.Add(key);
+                }
+            }
```
evildour: You can replace loop with:
```csharp
newRowSet.RowKeys.AddRange(originalRows.RowKeys.Select(key => !StartKeyIsAlreadyRead(key)))
```

Actually, if you make a negated version (`StartKeyIsntAlreadyRead`), it can be even simpler:
```csharp
newRowSet.RowKeys.AddRange(originalRows.RowKeys.Select(StartKeyIsntAlreadyRead))
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+
+                if (startKeyOneofCase == RowRange.StartKeyOneofCase.StartKeyClosed
+                    && StartKeyIsAlreadyRead(rowRange.StartKeyClosed)
+                    || (startKeyOneofCase == RowRange.StartKeyOneofCase.StartKeyOpen
+                        && StartKeyIsAlreadyRead(rowRange.StartKeyOpen))
+                    || startKeyOneofCase == RowRange.StartKeyOneofCase.None)
+                {
+                    newRange.StartKeyOpen = LastFoundKey;
+                }
+                newRowSet.RowRanges.AddRange(new List<RowRange> { newRange });
```
evildour: Use this instead so you don't need to create a list for each range:
```csharp
newRowSet.RowRanges.Add(newRange);
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+            }
+
+            return newRowSet;
+        }
+
+        private bool StartKeyIsAlreadyRead(ByteString startKey)
```
evildour: Make parameter `BigtableByteString` and then the code below can be:
```csharp
... => LastFoundKey != null && (startKey != null || startKey <= LastFoundKey);
```
(similar for `EndKeyIsAlreadyRead` below).

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            // empty start key emplies the smallest key.
+            return LastFoundKey != null && (startKey.IsEmpty
+                                            || ByteStringCompare(startKey, LastFoundKey) <= 0);
+        }
+
+        private bool EndKeyIsAlreadyRead(ByteString endKey)
+        {
+            // empty string emplies the largest key.
```
evildour: Nit: "emplies" => "implies"

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -0,0 +1,254 @@
...
+        /// </summary>
+        public enum ProcessingStatus
+        {
+            /// <summary> All responses produced OK. </summary>
+            SUCCESS,
+
+            /// <summary> All responses produced OK or a retruable code. </summary>
+            RETRYABLE,
+
+            /// <summary> Some responses had a non-retriable code. </summary>
```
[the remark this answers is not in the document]
AVaksman: I suggest that `NOT_RETRYABLE` should be the value where _all_ responses produced ether OK or NOT_RETRYABLE
`OK` == all produced Ok
`RETRYABLE` == any mix of Ok and NotRetryable where there is at least one Retryable

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -0,0 +1,254 @@
...
+                    // An individual mutation failed with a retryable code, usually DEADLINE_EXCEEDED.
+                    toRetry.Add(i);
+                    if (processingStatus == ProcessingStatus.SUCCESS)
+                    {
+                        processingStatus = ProcessingStatus.RETRYABLE;
+                    }
+                }
+                else
+                {
+                    // Don't retry if even a single response is not retryable.
```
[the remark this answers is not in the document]
AVaksman: '_modifiedRequest' is created on constructor and initially is equal to '_originalRequest'.
If we will stay with not retrying in case of a single NotRetryable status, than I suggest to nullify or clear `_modifiedRequest` and clear `toRetry`. Otherwise `GetRetryRequest()` at this point will still return full `_modifiedRequest == _originalRequest`

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableMutateRowsRequestManagerTest.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        private static List<T> Fill<T>(T value, int size)
+        {
+            List<T> list = new List<T>();
+            while (--size >= 0)
+            {
+                list.Add(value);
+            }
+            return list;
+        }
```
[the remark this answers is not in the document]
AVaksman: Originally I looked for an optimal way and referring to [this](https://stackoverflow.com/questions/1014005/how-to-populate-instantiate-a-c-sharp-array-with-a-single-value#answer-1014017) answer loop seemed to be faster, however just for a test  it doesn't really matter.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
AVaksman: Wouldn't that be slower, since that will traverse the `List` twice?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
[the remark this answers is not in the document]
evildour: For the `RETRYABLE/NOT_RETRYABLE`, yes I agree, this seems most reasonable. I haven't read through all the retry docs, but it seems like if you have a mix of retryable and non-retryable errors, you could just retry the entries with the former kind of error.

However, why do you need `OK` at all? This class is essentially helping manage retries, so what you really want to know is whether retries should be attempted or not (or whether things went wrong somewhere along the way). 3 values should be sufficient for this. Why do we need to know about a 4th case?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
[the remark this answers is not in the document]
evildour: Or perhaps `RetryRequest` (like I said, that should be a property and `_modifiedRequest` should be removed) should be null always until the processing status is `RETRYABLE`. Then `RetryRequest`, as the name would suggest, only holds the request that should be made on a subsequent retry request. How does that sound?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
evildour: Yes, but I wouldn't try to over-optimize though just yet. I would assume the retries are the exception, not the rule, so you wouldn't have a huge number of them to deal with. But if you are really worried about it, what you have is ok, or you can do the somewhat messier version of:
```csharp
private MutateRowsRequest CreateRetryRequest(List<int> indiciesToRetry)
{
    _mapToOriginalIndex = new int[indiciesToRetry.Count];
    _modifiedRequest = new MutateRowsRequest { Entries =
    {
        indiciesToRetry.Select((indexValue, i) =>
        {
            _mapToOriginalIndex[i] = indexValue;
            return _originalRequest.Entries[indexValue];
        })
    } };
    return _modifiedRequest;
}
```
But I think if you like to go with the optimized version, what you have is better anyway.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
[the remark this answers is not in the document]
AVaksman: Based on the ProcessingStatus the "Retryer" utility will decide its next step, for example,
* `SUCCESS` => `BuildResponse()` => return full response to user
* `RETRYABLE` => `RetryRequest` and retry or (if exhausted all retry attempts) `BuildResponse()` => return full response to user.
* `NOT_RETRYABLE` =>`BuildResponse()` => return full response to user.
* `INVALID` => return message to user that something went wrong

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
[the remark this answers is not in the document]
AVaksman: Based on my conversation with Solomon and after doing some tests I think we should probably stick to not retrying even if one response is not retryable. Details in the email

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableMutateRowsRequestManagerTest.cs
```diff
@@ -0,0 +1,259 @@
...
+using System.Collections.Generic;
+using System;
+using Xunit;
+
+namespace Google.Cloud.Bigtable.V2.Tests
+{
+    /// <summary>
+    /// Test for <see cref="BigtableMutateRowsRequestManager"/>.
+    /// </summary>
+    public class BigtableMutateRowsRequestManagerTest
```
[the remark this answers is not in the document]
AVaksman: I will explain in the email.

evildour:
I think there was some issue with the latest push. Perhaps a rebase error? It shows 225 commits and over 100k lines changed. Maybe this PR should be closed and another opened?

AVaksman:
I had to fetch updates from the upstream in the middle to get the latest code and helper functions that where used in addressing the comments. That was 200+ commits, plus mine. I can close this PR and open a new one, don't want to loose our conversations. Your preference.

PR #1798, opened 2017-12-29
BigtableClient snippets

evildour:
No rush, but once this is merged, we should probably publish an updated build to nuget.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
@@ -79,7 +79,6 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
             return rowKey;
         }
 
-
         public async Task InsertRowsAsync(
             TableName tableName,
             IEnumerable<BigtableByteString> rowKeys,
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
@@ -27,108 +27,23 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
     // Note: this does not use CloudProjectFixtureBase as a base class, as an emulator can be used as an alternative.
     // If that pattern becomes more widespread, we can bring that into CloudProjectFixtureBase.
     [CollectionDefinition(nameof(BigtableFixture))]
-    public class BigtableFixture : IDisposable, ICollectionFixture<BigtableFixture>
+    public class BigtableFixture : BigtableFixtureBase, ICollectionFixture<BigtableFixture>
     {
-        private const string EmulatorEnvironmentVariable = "BIGTABLE_EMULATOR_HOST";
-        private const string TestInstanceEnvironmentVariable = "BIGTABLE_TEST_INSTANCE";
-        private const string TestProjectEnvironmentVariable = CloudProjectFixtureBase.TestProjectEnvironmentVariable;
-
         public const string DefaultColumnFamily = "cf1";
         public const string DefaultColumnQualifier = "row_exists";
         public const string DefaultValue = "true";
         public const string OtherColumnFamily = "test_data";
 
-        public ProjectName ProjectName { get; private set; }
-        public InstanceName InstanceName { get; private set; }
-        public TableName TableName { get; private set; }
-
-        public BigtableInstanceAdminClient InstanceAdminClient { get; private set; }
-        public BigtableTableAdminClient TableAdminClient { get; private set; }
-        public BigtableClient TableClient { get; private set; }
-
-        public List<TableName> CreatedTables { get; } = new List<TableName>();
-        public Channel EmulatorChannel { get; }
-
-        public BigtableFixture()
-        {
-            string emulatorHost = Environment.GetEnvironmentVariable(EmulatorEnvironmentVariable);
-
-            string projectId;
-            string instanceId;
-            if (!string.IsNullOrEmpty(emulatorHost))
+        protected override Table CreateDefaultTable() =>
+            new Table
             {
-                projectId = "emulator-test-project";
-                EmulatorChannel = new Channel(emulatorHost, ChannelCredentials.Insecure);
-
-                instanceId = "doesnt-matter";
-            }
-            else
-            {
-                projectId = Environment.GetEnvironmentVariable(TestProjectEnvironmentVariable);
-                if (string.IsNullOrEmpty(projectId))
+                Granularity = Table.Types.TimestampGranularity.Millis,
+                ColumnFamilies =
                 {
-                    throw new InvalidOperationException(
-                        $"Please set either the {EmulatorEnvironmentVariable} or {TestProjectEnvironmentVariable} environment variable before running tests");
+                    { DefaultColumnFamily, new ColumnFamily { GcRule = new GcRule { MaxNumVersions = 3 } } },
+                    { OtherColumnFamily, new ColumnFamily { GcRule = new GcRule { MaxNumVersions = 3 } } }
                 }
-
-                instanceId = Environment.GetEnvironmentVariable(TestInstanceEnvironmentVariable);
-                if (string.IsNullOrEmpty(instanceId))
-                {
-                    throw new InvalidOperationException(
-                        $"Please set the {TestInstanceEnvironmentVariable} environment variable before running non-emulator tests.");
-                }
-            }
-
-            ProjectName = new ProjectName(projectId);
-            InstanceName = new InstanceName(projectId, instanceId);
-
-            Task.Run(InitBigtableInstanceAndTable).Wait();
-        }
-
-        public async Task<TableName> CreateTable()
-        {
-            var tableName = new TableName(ProjectName.ProjectId, InstanceName.InstanceId, Guid.NewGuid().ToString());
-            CreatedTables.Add(tableName);
-            await TableAdminClient.CreateTableAsync(
-                new InstanceName(tableName.ProjectId, tableName.InstanceId),
-                tableName.TableId,
-                new Table
-                {
-                    Granularity = Table.Types.TimestampGranularity.Millis,
-                    ColumnFamilies =
-                    {
-                        { DefaultColumnFamily, new ColumnFamily { GcRule = new GcRule { MaxNumVersions = 3 } } },
-                        { OtherColumnFamily, new ColumnFamily { GcRule = new GcRule { MaxNumVersions = 3 } } }
-                    }
-                });
-            return tableName;
-        }
-
-        private async Task InitBigtableInstanceAndTable()
-        {
-            if (EmulatorChannel == null)
-            {
-                InstanceAdminClient = BigtableInstanceAdminClient.Create();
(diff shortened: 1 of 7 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,367 @@
...
+            // Snippet: CheckAndMutateRowAsync(TableName,BigtableByteString,RowFilter,*)
+            CheckAndMutateRowResponse response = await client.CheckAndMutateRowAsync(
+                new TableName(projectId, instanceId, tableId),
+                "user12345",
+                // If the latest value in the "Address:Country" column is "USA"...
+                RowFilters.Chain(
+                    RowFilters.FamilyNameRegex("Address"),
+                    RowFilters.ColumnQualifierRegex("Country"),
+                    RowFilters.CellsPerColumnLimit(1),
+                    RowFilters.ValueRegex("USA")),
```
jskeet: Is ValueRegex the most appropriate method to use even for simple literal matches?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,367 @@
...
+                Mutations.SetCell("Address", "Country", "USA"),
+                Mutations.SetCell("Score", "Level 1", 12));
+
+            // Snippet: MutateRowAsync(TableName,BigtableByteString,*)
+            await client.MutateRowAsync(
+                new TableName(projectId, instanceId, tableId),
+                // From the row with key '0xABCD1234'...
+                new byte[] { 0xAB, 0xCD, 0x12, 0x34 },
+                // delete all cells from the column family "Address",
+                Mutations.DeleteFromFamily("Address"),
```
jskeet: Neither of these methods return anything saying how many cells/rows were deleted, I assume?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,367 @@
...
+                ReadModifyWriteRules.Increment("Score", "Level 1", 55));
+
+            Cell modifiedCell = response.Row.Families[0].Columns[0].Cells[0];
+
+            // Implicitly convert the ByteString Value into a BigtableByteString,
+            // which supports additional conversions specific to Bigtable.
+            BigtableByteString value = modifiedCell.Value;
+
+            // Convert the BigtableByteString to a long to obtain the updated value.
+            long newScore = (long)value;
```
jskeet: Maybe print this out? It's odd to have a snippet which ends with a conversion.
(I'd also personally have a space: `(long) value` - but I'm not overly bothered on that front.)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,367 @@
...
+                new TableName(projectId, instanceId, tableId),
+                "user12345",
+                Mutations.DeleteFromRow(),
+                Mutations.SetCell("Address", "Country", "USA"),
+                Mutations.SetCell("Address", "Line 1", "1234 Main Street"),
+                Mutations.SetCell("Address", "Line 2", "Apt. 2"),
+                Mutations.SetCell("Score", "Level 1", 12));
+
+            // Snippet: ReadRowAsync(TableName,BigtableByteString,RowFilter,CallSettings)
+            // Only read cells from the "Line 1" and "Line 2" columns in the
```
jskeet: Just to clarify, it's really "anything starting with line and a digit" but we know that there *are* only Line 1 and Line 2, yes?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,367 @@
...
+                    Console.WriteLine($"    Column qualifier: {column.Qualifier.ToStringUtf8()}");
+                    foreach (Cell cell in column.Cells)
+                    {
+                        Console.WriteLine($"      Cell value: {cell.Value.ToStringUtf8()}");
+                        Console.WriteLine($"      Timestamp: {cell.Version.ToDateTime()}");
+                        Console.WriteLine($"      Labels: {string.Join(", ", cell.Labels)}");
+                    }
+                }
+            }
+
```
jskeet: Remove blank line at the end of the snippet? We may do that automatically, but I'm not sure.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,367 @@
...
+        {
+            var cts = new CancellationTokenSource();
+            var cancellationToken = cts.Token;
+            var client = _fixture.TableClient;
+            var projectId = _fixture.TableName.ProjectId;
+            var instanceId = _fixture.TableName.InstanceId;
+            var tableId = _fixture.TableName.TableId;
+
+            // Snippet: ReadRows(TableName,RowSet,RowFilter,long?,CallSettings)
+            // Initialize request arguments
```
jskeet: Not sure this comment line is useful.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -0,0 +1,367 @@
...
+                CallSettings.FromCancellationToken(cancellationToken));
+
+            // Get all streaming responses from server in a list so they can be easily iterated.
+            IList<SampleRowKeysResponse> rowKeySamples =
+                await streamingResponse.ToListAsync(cancellationToken);
+
+            long previousRowOffsetBytes = 0;
+            foreach (SampleRowKeysResponse rowKeySample in rowKeySamples)
+            {
+                long offsetDifference = rowKeySample.OffsetBytes - previousRowOffsetBytes;
```
jskeet: Is this guaranteed to always be positive?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixture.cs
```diff
@@ -164,6 +79,7 @@ public BigtableFixture()
             return rowKey;
         }
 
+
```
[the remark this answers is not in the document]
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
evildour: Yes. I guess we could offer a `RowFilters.Value(string)` method which delegates to `ValueRegex` with an escaped string (and similarly for the other `...Regex` methods). Does that sound like a good idea? If so I can followup with another PR.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
evildour: That's correct. MutateRowResponse contains no information (see [here](https://github.com/googleapis/googleapis/blob/546d3afd75187e65229a2670eb440e84a05f5d9d/google/bigtable/v2/bigtable.proto#L245))

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
evildour: Good point. Updated.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
jskeet: Let's create a feature request issue, and check with the Bigtable team. It sounds like something that should probably be consistent across languages.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
evildour: Yeah, this looks like it was left over from something I copy/pasted from the generated snippets.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
evildour: I would assume, but I'm not sure. Do you think it would make sense to add a `Math.Max(0, ...)` to be safe?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
evildour: Ok, will do.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
jskeet: I think we should probably find out what's guaranteed, and add a comment to that effect. If we use a Math.Max and there's no need to, that's bad... if we assume that it *will* always be positive and it's not, that's bad too.

evildour: Sounds good. I'll ask internally, and update accordingly with a follow PR if that's ok with you.

evildour: Ok, the response I got internally is that these "should be non-decreasing", so I'll add a comment.

PR #1804, opened 2018-01-02
Avoid Storage bucket create/delete rate limits

### LINKED ISSUE #1801: Integration tests fail due to Storage bucket creation/deletion rate quota
We should sleep when we create or delete a bucket to avoid this.

jskeet:
Sleep for 2 seconds after each create/delete bucket operation in tests.

Fixes #1801.

(This will conflict with #1678 a bit of course, but I'll sort them both...)


jskeet:
Fixing the build failure... not sure why I haven't seen that myself.

PR #1808, opened 2018-01-03
Permit protos to be serialized and deserialized in Firestore

jskeet:
We always clone the proto when serializing, as the WriteBatch class
can modify the result. We don't clone when deserializing, as *our*
code doesn't modify the result - if user code does, it may see those
modifications if deserialization occurs more than once, but that's
probably okay.

This is an alternative to #1796. cc @Thaina - please comment if you're concerned about the cloning cost. (I'm not sure there's a good alternative at the moment though.)

The deserialization change will allow you to deserialize to a type that has a `Value` property, or an `IDictionary<string, Value` etc.


jskeet:
Thanks Chris - @Thaina has given a thumbs up, so I'll merge.

PR #1810, opened 2018-01-03
Request Manager

AVaksman:
BigtableMutateRowsRequestManager
BigtableReadRowsRequestManager
Unit tests

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableReadRowsRequestManagerTest.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableReadRowsRequestManagerTest.cs
@@ -1,0 +1,187 @@
+﻿// Copyright 2017 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Xunit;
+
+namespace Google.Cloud.Bigtable.V2.Tests
+{
+    /// <summary>
+    /// Test for <see cref="BigtableReadRowsRequestManager"/>.
+    /// </summary>
+    public class BigtableReadRowsRequestManagerTest
+    {
+        private static ReadRowsRequest CreateRowRangeRequest(RowRange rowRange) =>
+            new ReadRowsRequest { Rows = RowSet.FromRowRanges(rowRange) };
+
+        private static ReadRowsRequest CreateRowKeysRequest(params BigtableByteString[] keys) =>
+            new ReadRowsRequest { Rows = RowSet.FromRowKeys(keys) };
+
+        private static ReadRowsRequest CreateRowFilterRequest(RowFilter rowFilter) =>
+            new ReadRowsRequest { Filter = rowFilter };
+
+        /// <summary>
+        /// Test rowfilter appears in the <see cref="BigtableReadRowsRequestManager.BuildUpdatedRequest()"/>
+        /// </summary>
+        [Fact]
+        public void TestRowFilter()
+        {
+            var rowFilter = RowFilters.Chain(
+                RowFilters.CellsPerColumnLimit(1),
+                RowFilters.CellsPerRowOffset(2),
+                RowFilters.CellsPerRowLimit(10));
+
+            ReadRowsRequest originalRequest = CreateRowFilterRequest(rowFilter);
+
+            BigtableReadRowsRequestManager underTest = new BigtableReadRowsRequestManager(originalRequest);
+
+            Assert.Equal(originalRequest, underTest.BuildUpdatedRequest());
+        }
+
+        /// <summary>
+        /// Test a single, full table scan scenario for updating <see cref="BigtableReadRowsRequestManager.LastFoundKey"/>.
+        /// </summary>
+        [Fact]
+        public void TestFilterRowsTestAllRange()
+        {
+            BigtableByteString key1 = "row1";
+
+            ReadRowsRequest originalRequest = CreateRowRangeRequest(RowRange.ClosedOpen(null, null));
+
+            BigtableReadRowsRequestManager underTest = new BigtableReadRowsRequestManager(originalRequest);
+
+            underTest.LastFoundKey = key1;
+            Assert.Equal(CreateRowRangeRequest(RowRange.Open(key1, null)), underTest.BuildUpdatedRequest());
+
+        }
+
+        /// <summary>
+        ///  Test rowKeys scenario for <see cref="BigtableReadRowsRequestManager.BuildUpdatedRequest()"/>.
+        /// </summary>
+        [Fact]
+        public void TestFilterRowsRowKeys()
+        {
+            BigtableByteString key1 = "row1";
+            BigtableByteString key2 = "row2";
+            BigtableByteString key3 = "row3";
+
+            ReadRowsRequest originalRequest = CreateRowKeysRequest(key1, key2, key3);
+
+            BigtableReadRowsRequestManager underTest = new BigtableReadRowsRequestManager(originalRequest);
+
+            Assert.Equal(originalRequest, underTest.BuildUpdatedRequest());
+            underTest.LastFoundKey = key1;
+
+            Assert.Equal(CreateRowKeysRequest(key2, key3), underTest.BuildUpdatedRequest());
+        }
+
+        /// <summary>
+        /// Test multiple rowset filter scenarios for <see cref="BigtableReadRowsRequestManager.BuildUpdatedRequest()"/>.
+        /// </summary>
+        [Fact]
+        public void TestFilterRowsMultiRowSetFilters()
+        {
+            BigtableByteString key1 = "row1";
+            BigtableByteString key2 = "row2";
+            BigtableByteString key3 = "row3";
+
+            RowSet fullRowSet = RowSet.FromRowKeys(key1, key2, key3);
+            fullRowSet.RowRanges.Add(new[]
+            {
+                RowRange.OpenClosed(null, key1), // should be filtered out
+                RowRange.Open(null, key1), // should be filtered out
+                RowRange.Open(key1, key2), // should stay
+                RowRange.ClosedOpen(key1, key2), // should be converted (key1 -> key2)
+                RowRange.Closed(key1, key2), // should be converted (key1 -> key2]
+                RowRange.Open(key2, key3), // should stay
+                RowRange.ClosedOpen(key2, key3) // should stay
+            });
+
+            RowSet filteredRowSet = RowSet.FromRowKeys(key2, key3);
+            filteredRowSet.RowRanges.Add(new[]
+            {
+                RowRange.Open(key1, key2), // should stay
+                RowRange.Open(key1, key2), // should be converted (key1 -> key2)
+                RowRange.OpenClosed(key1, key2), // should be converted (key1 -> key2]
+                RowRange.Open(key2, key3), // should stay
+                RowRange.ClosedOpen(key2, key3)// should stay
+            });
+
+            ReadRowsRequest originalRequest = new ReadRowsRequest { Rows = fullRowSet };
+            ReadRowsRequest filteredRequest = new ReadRowsRequest { Rows = filteredRowSet };
+
+            BigtableReadRowsRequestManager underTest = new BigtableReadRowsRequestManager(originalRequest);
+            Assert.Equal(originalRequest, underTest.BuildUpdatedRequest());
+            underTest.LastFoundKey = key1;
+            Assert.Equal(filteredRequest, underTest.BuildUpdatedRequest());
+        }
+
+        /// <summary>
+        /// For this test we will assume that the table contains sequentially numbered rowkeys from "row000" to "row999"
+        /// </summary>
+        [Fact]
+        public void TestRowsLimit()
+        {
+            BigtableByteString key1 = "row050";
+            BigtableByteString key2 = "row100";
+
+            ReadRowsRequest originalRequest = CreateRowRangeRequest(RowRange.OpenClosed(key1, null));
+            originalRequest.RowsLimit = 100;
+
+            ReadRowsRequest updatedRequest = CreateRowRangeRequest(RowRange.OpenClosed(key2, null));
+            updatedRequest.RowsLimit = 50;
+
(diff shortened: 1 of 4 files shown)


AVaksman:
Sorry for the previous mess. Nice and clean now.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -0,0 +1,212 @@
...
+
+        private static MutateRowsResponse.Types.Entry CreateEntry(int i, Rpc.Status status)
+        {
+            return new MutateRowsResponse.Types.Entry { Index = i, Status = status };
+        }
+
+        /// <summary>
+        /// Contains possible processing statuses
+        /// </summary>
+        internal enum ProcessingStatus
```
evildour: (repeating your comment here from #1797)
_Based on the ProcessingStatus the "Retryer" utility will decide its next step, for example,_

* _SUCCESS => BuildResponse() => return full response to user_
* _RETRYABLE => RetryRequest and retry or (if exhausted all retry attempts) BuildResponse() => return full response to user._
* _NOT_RETRYABLE =>BuildResponse() => return full response to user._
* _INVALID => return message to user that something went wrong_

So like I was saying, it seems like `SUCCESS` is unnecessary. For the purposes of this class and callers to it, `SUCCESS` and `NOT_RETRYABLE` appear to be semantically equivalent. They both mean "BuildResponse() => return full response to user". All successful mutations is just one case of something that shouldn't be retried.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/BigtableReadRowsRequestManagerTest.cs
```diff
@@ -0,0 +1,187 @@
...
+            BigtableByteString key1 = "row1";
+            BigtableByteString key2 = "row2";
+            BigtableByteString key3 = "row3";
+
+            ReadRowsRequest originalRequest = CreateRowKeysRequest(key1, key2, key3);
+
+            BigtableReadRowsRequestManager underTest = new BigtableReadRowsRequestManager(originalRequest);
+
+            Assert.Equal(originalRequest, underTest.BuildUpdatedRequest());
+            underTest.LastFoundKey = key1;
```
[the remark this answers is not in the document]
AVaksman: I don't think so, you probably mean a case with `RowRange`(from "row01" to "row10") and `LastFoundKey` would be "row07". I will add such a test. 

When we set up `Rows.RowKeys` we are passing a list of specific `rowKeys` we are looking for, in this example "row1", "row2" and "row3". There should not be a found `rowKey` "row1a". It would actually be an equivalent of `processingStatus.Invalid` from `MutateRowsRequestManager`. 

Hmm we don't even check that `LastFoundKey` is actually within the list/range of `rowKeys` we are requesting, should we?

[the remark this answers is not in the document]
evildour: Nope, I meant what I said, but yes checking for the range cut in the middle somewhere would be good too.

"There should not be a found `rowKey` "row1a" - Yes, there won't be a found key like that, but the proto field, found [here](https://github.com/evildour/googleapis/blob/master/google/bigtable/v2/bigtable.proto#L171), indicates the last key _scanned_, not found. I'm guessing if it has to skip form "row1" to "row2" internally but return a response before getting to "row2", `last_scanned_row_key` will let you know where it left off. We can check internally to confirm, but it wouldn't hurt to have the test anyway just in case the server does return a last scanned key that doesn't match one being searched for.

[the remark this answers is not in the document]
AVaksman: Added the test

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,134 @@
...
+            RowSet originalRows = _originalRequest.Rows;
+
+            if (LastFoundKey.Value.IsEmpty)
+            {
+                return originalRows;
+            }
+
+            RowSet newRowSet = new RowSet();
+
+            newRowSet.RowKeys.AddRange(originalRows.RowKeys.Where(key => !StartKeyIsAlreadyRead(key)));
```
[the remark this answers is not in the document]
AVaksman: Then I'll rename `RowSet FilterRows()` to `RepeatedField<RowRange> FilterRowRanges` and `RowSet newRowSet = new RowSet` to `RepeatedField<RowRange> newRowRanges = _originalRequest.Rows.RowRanges` accordingly.

This way object creation expression will have 
```
TableNameAsTableName = _originalRequest.TableNameAsTableName,
// Transfer RowFilter.
Filter = _originalRequest.Filter,
// Remove received rowKeys form Rows.
Rows =
{
    RowKeys = {_originalRequest.Rows.RowKeys.Where(key => !StartKeyIsAlreadyRead(key))},
    RowRanges = {FilterRowRanges()}
}
```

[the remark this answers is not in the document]
evildour: I'm not quite sure I follow, but I'll just look at the updated code. Also, you should be able to use `IEnumerable<...>` rather than `RepeatedField<...>` and use any collection or iterator block.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        /// <summary>
+        /// Constructor for retrying failed mutation entries.
+        /// </summary>
+        /// <param name="retryStatuses">
+        /// Collection of Grpc status codes to retry on.</param>
+        /// <param name="mutateRowsRequest">
+        /// <see cref="MutateRowsRequest"/> that was received from the user.</param>
+        internal BigtableMutateRowsRequestManager(IEnumerable<StatusCode> retryStatuses, MutateRowsRequest mutateRowsRequest)
```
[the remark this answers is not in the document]
AVaksman: Yes, initially as per specs there where _Always Retryable_ codes and codes that could be _specified to be retryable_ as well. I will confirm with Solomon today and change the code if `retryStatuses` parameter is not needed.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -0,0 +1,205 @@
...
+            _messageIsInvalid = _messageIsInvalid || _results.Any(result => result == null);
+            
+            // There was a problem in the data found in SetStatus(), so fail the RPC.
+            if (_messageIsInvalid)
+            {
+                return ProcessingStatus.Invalid;
+            }
+
+            // This list hold indices of entries to be included in retry request
+            List<int> toRetry = new List<int>();
```
evildour: Similarly to avoid heap allocations when not necessary, initialize this to null here and the create it if necessary before the first `Add`.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,134 @@
...
+        /// <summary>
+        /// Removes received rows from the RowSet.
+        /// </summary>
+        /// <returns>
+        /// New RowSet with not received rows.</returns>
+        private RowSet FilterRows()
+        {
+            RowSet originalRows = _originalRequest.Rows;
+
+            if (LastFoundKey.Value.IsEmpty)
```
evildour: I see this used a bunch here. I think it probably makes sense to add an IsEmpty delegating property on `BigtableByteString`. I can add this and update the code here after it's merged.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,134 @@
...
+                RowRange newRange = rowRange;
+                RowRange.StartKeyOneofCase startKeyOneofCase = rowRange.StartKeyCase;
+
+                if (startKeyOneofCase == RowRange.StartKeyOneofCase.StartKeyClosed &&
+                    StartKeyIsAlreadyRead(rowRange.StartKeyClosed)
+                    || (startKeyOneofCase == RowRange.StartKeyOneofCase.StartKeyOpen &&
+                        StartKeyIsAlreadyRead(rowRange.StartKeyOpen))
+                    || startKeyOneofCase == RowRange.StartKeyOneofCase.None)
+                {
+                    newRange.StartKeyOpen = (ByteString) LastFoundKey;
```
evildour: I just realized this is modifying the original range. You should probably make a clone instead and modify the clone if you get in here.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,134 @@
...
+                }
+                newRowSet.RowRanges.Add(newRange);
+            }
+
+            return newRowSet;
+        }
+    
+
+    private bool StartKeyIsAlreadyRead(BigtableByteString startKey) =>
```
evildour: Nit: indentation

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
AVaksman: Currently it could be done as I did,
1. Get the underlying byte string and check the value using `IsEmpty` property.
or
2. Get the underlying byte string and check the value against the `null` -> 
`LastFoundKey.Value == null`

evildour: There's also a 3rd way: `LastFoundKey.Lenght == 0`. However, that said, I'm adding `IsEmpty` in #1822, so when that's merged I'll fix up these references.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -0,0 +1,215 @@
...
+        /// <see cref="MutateRowsRequest"/> that was received from the user.</param>
+        internal BigtableMutateRowsRequestManager(IEnumerable<StatusCode> retryStatuses, MutateRowsRequest mutateRowsRequest)
+        {
+            _originalRequest = mutateRowsRequest;
+
+            _results = new Rpc.Status[_originalRequest.Entries.Count];
+
+            // This is a map between RetryRequest and _originalRequest
+            // as they could diverge if a retry occurs.
+            //_mapToOriginalIndex = Enumerable.Range(0, _originalRequest.Entries.Count).ToArray();
```
evildour: Remove this and associated comment

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -0,0 +1,131 @@
...
+                RowRange.EndKeyOneofCase endKeyOneofCase = rowRange.EndKeyCase;
+                if (endKeyOneofCase == RowRange.EndKeyOneofCase.EndKeyClosed &&
+                    EndKeyIsAlreadyRead(rowRange.EndKeyClosed)
+                    || endKeyOneofCase == RowRange.EndKeyOneofCase.EndKeyOpen &&
+                    EndKeyIsAlreadyRead(rowRange.EndKeyOpen))
+                {
+                    continue;
+                }
+
+                RowRange newRange = rowRange.Clone();
```
evildour: You don't need to do this until you get into the if statement below (but its fine to do it here if you're really trying to be safe about the ranges not being shared). Also, do any of the tests fail if this `Clone` call is removed? If not, it would be good to have one.

AVaksman: If the current `rowRange` is not skipped by the first if statement on line 100 then, modified or not by the the second if statement, it will still be added to `newRowSet.RowRanges` on line 119

evildour: I realize that, but it may be safe to assume they can be shared if others play nicely and don't modify the ranges in an existing request. So we could defer cloning until we know we're going to modify it. But my point was that if you want to be safe about things and prevent sharing altogether, that is fine. Either way, a test which fails without `Clone` being called should be added, or an assert should be added to one of the existing tests that modifies ranges to make sure the original request has the original range values.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -114,9 +113,14 @@ private RowSet FilterRows()
...
-                    newRange.StartKeyOpen = (ByteString) LastFoundKey;
+                    RowRange newRange = rowRange.Clone();
+                    newRange.StartKeyOpen = LastFoundKey.Value;
+                    newRowSet.RowRanges.Add(newRange);
+                }
+                else
+                {
+                    newRowSet.RowRanges.Add(rowRange);
                 }
-                newRowSet.RowRanges.Add(newRange);
```
evildour: I think this would be a bit cleaner:
```csharp
RowRange newRange = rowRange;
if (...)
{
    newRange = newRange.Clone();
    newRange.StartKeyOpen = LastFoundKey.Value;
}
newRowSet.RowRanges.Add(newRange);
```

AVaksman: Much cleaner. 
Is it OK that `newRange` will very likely be initialized and then re-assigned before being added to the `newRowSet`? Although it is only a reference.

evildour: I don't see any issue with that

evildour:
Merging now. Thanks again Alex!

AVaksman:
Thanks

PR #1812, opened 2018-01-04
Update Grpc.Core to 1.8.0

jskeet:
Only the first commit is manual.

### DIFF
--- a/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
+++ b/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
@@ -46,7 +46,7 @@ namespace Google.Cloud.Tools.ProjectGenerator
         private const string DefaultVersionValue = "default";
         private const string GrpcPackage = "Grpc.Core";
         private const string DefaultGaxVersion = "2.2.1";
-        private const string GrpcVersion = "1.7.1";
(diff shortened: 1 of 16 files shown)


tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
```diff
@@ -46,7 +46,7 @@ public class Program
         private const string DefaultVersionValue = "default";
         private const string GrpcPackage = "Grpc.Core";
         private const string DefaultGaxVersion = "2.2.1";
-        private const string GrpcVersion = "1.7.1";
+        private const string GrpcVersion = "1.8.0";
```
chrisdunelm: We need to make sure the next release of every package is at least a minor version release (not a patch release).
Do we have tooling that ensures this? Or do we just need to be careful?

jskeet: Any GA'd releases have explicit versions for everything anyway, which is why only a subset of projects are updated in the second commit. We need to be careful to upgrade those manually when we do a next minor update...

(We also need to look at GAX, and update the protobuf generator and dependency to 3.5.1 at the same time. We should probably chat about that.)

PR #1814, opened 2018-01-04
Add delay to UrlSignerTests to guard against minor time diff with server

evildour:


iantalarico:
Once this is in I can kick off another test run to test this on the CI

PR #1822, opened 2018-01-09
[Bigtable] Various minor TODOs

evildour:

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/ReadRowsTest.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/ReadRowsTest.cs
@@ -215,22 +215,22 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
             var client = _fixture.TableClient;
             var originalIndexes = new Dictionary<BigtableByteString, int>();
             int endRowIndex = 100;
-            var rowKeys = new BigtableByteString[endRowIndex];
+            var rowKeys = new List<BigtableByteString>();
             for (int i = 0; i < endRowIndex; i++)
             {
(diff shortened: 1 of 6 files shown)


jskeet:
I'm at CodeMash this week - would you prefer this to be reviewed by someone else, or delayed a bit? (I *may* be able to do it during the week, but no promises...)

evildour:
Either way. There's no rush to get this in, so it can wait.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/ReadRowsTest.cs
```diff
@@ -264,28 +255,27 @@ public void ReadRowMissing()
             var tableName = _fixture.TableName;
             var client = _fixture.TableClient;
             var originalIndexes = new Dictionary<BigtableByteString, int>();
-            var rowKeys = new List<BigtableByteString>();
-            for (int i = 0; i < 100; i++)
+            var rowKeys = new BigtableByteString[100];
```
ILMTitan: Why go from a `List` to an `Array` here? That feels like going backwards.

evildour: `RowSet.FromRowKeys` takes a params array, so I thought it would be unnecessary to populate a list just to call `ToArray()` on it when I knew the size it was going to be at the start. However, thinking about it again, perhaps the better thing to do is have another overload of `RowSet.FromRowKeys` which takes an `IEnumerable<BigtableByteString>` and have the current overload delegate to it.

evildour:
PTAL. I've added the additional RowSet factory overloads and changed the testing logic to use lists again.

PR #1823, opened 2018-01-09
[Bigtable] Exact match row filters to complement regex filters

evildour:
(no rush on this one either)

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
@@ -52,10 +52,10 @@ namespace Google.Cloud.Bigtable.V2.Snippets
                 "user12345",
                 // If the latest value in the "Address:Country" column is "USA"...
                 RowFilters.Chain(
-                    RowFilters.FamilyNameRegex("Address"),
-                    RowFilters.ColumnQualifierRegex("Country"),
+                    RowFilters.FamilyNameExact("Address"),
+                    RowFilters.ColumnQualifierExact("Country"),
                     RowFilters.CellsPerColumnLimit(1),
-                    RowFilters.ValueRegex("USA")),
+                    RowFilters.ValueExact("USA")),
                 // ...remove all previous Level 1 scores and set the value to 0.
                 Mutations.DeleteFromColumn("Score", "Level 1"),
                 Mutations.SetCell("Score", "Level 1", 0, new BigtableVersion(1)));
@@ -230,7 +230,7 @@ namespace Google.Cloud.Bigtable.V2.Snippets
                 new TableName(projectId, instanceId, tableId),
                 "user12345",
                 RowFilters.Chain(
-                    RowFilters.FamilyNameRegex("Address"),
+                    RowFilters.FamilyNameExact("Address"),
                     RowFilters.ColumnQualifierRegex(@"Line \d"),
                     RowFilters.CellsPerColumnLimit(1)),
                 CallSettings.FromCancellationToken(cancellationToken));
@@ -294,7 +294,7 @@ namespace Google.Cloud.Bigtable.V2.Snippets
 
             // Only read cells from the "Address" column family and only take the most recent cell from each column.
             RowFilter filter = RowFilters.Chain(
-                RowFilters.FamilyNameRegex("Address"),
+                RowFilters.FamilyNameExact("Address"),
                 RowFilters.CellsPerColumnLimit(1));
 
             // Just read the first 10 rows.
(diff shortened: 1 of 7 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Utilities.cs
```diff
@@ -43,6 +47,58 @@ internal static ByteString Concat(this ByteString left, ByteString right)
...
+                {
+                    // Note that this special handling is not strictly required for RE2,
+                    // but this quoting is required for other regexp libraries such as
+                    // PCRE.
+                    // Can't use @"\0" since the next character might be a digit.
+                    result.AddRange(EscapedNullBytes);
+                }
+                else
+                {
+                    if (ShouldEscape(b))
```
jskeet: This will escape each byte within a UTF-8 sequence - which means we end up with invalid UTF-8. I doubt that that's the intention, is it? Shouldn't we be iterating in terms of characters rather than bytes?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
```diff
@@ -177,6 +194,28 @@ public static class RowFilters
...
+        /// Note that string is implicitly convertible to <see cref="BigtableByteString"/>, so <paramref name="value"/> can
+        /// be specified using a string as well and its UTF-8 representations will be used.
+        /// </para>
+        /// </remarks>
+        /// <param name="value">
+        /// The exact value used to match column families. Must not be null
+        /// </param>
+        /// <returns>The created row filter.</returns>
+        public static RowFilter FamilyNameExact(string value) =>
+            new RowFilter
```
jskeet: I'd really like to avoid the conversion to UTF-8 and back again if possible. Let's talk about that when the big "iterate over bytes" issue is sorted - it may drop out of that.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
```diff
@@ -52,10 +52,10 @@ public class BigtableClientSnippets
                 "user12345",
                 // If the latest value in the "Address:Country" column is "USA"...
                 RowFilters.Chain(
-                    RowFilters.FamilyNameRegex("Address"),
-                    RowFilters.ColumnQualifierRegex("Country"),
+                    RowFilters.FamilyNameExact("Address"),
```
jskeet: Having changed all these uses, do we have tests for actual regex patterns?

evildour: Yes, there are 7 calls into `...Regex` helpers in Tests and 5 in IntegrationTests.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Utilities.cs
evildour: The `(b & 0x80) == 0x80` check in `ShouldEscape` should cover that. In any multi-byte pattern in UTF-8, all bytes have the high order bit set, so we'll never escape them.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
evildour: I assumed since this is unlikely to be a source of performance bottlenecks, it would be better to have a single escape helper operate over byte strings rather than a duplication of logic. But I can change it if you feel otherwise.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Utilities.cs
jskeet: Ah, I'm with you. Yup, makes sense.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowFilters.cs
jskeet: Happy to leave it for now. It's an implementation detail after all - we can fix it later if we see issues.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Snippets/BigtableClientSnippets.cs
jskeet: SGTM.

PR #1827, opened 2018-01-15
API regeneration

jskeet:
Best reviewed commit-by-commit

I need to talk to Chris about smoke test generation to see what's going on there, but that can happen after this change has been merged.


jskeet:
Looks like the copyright detector needs updating too - I suspected that might happen. Will get on it.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/ResourceNames.cs
```diff
@@ -1,10 +1,10 @@
-// Copyright 2017, Google LLC All rights reserved.
+// Copyright 2018 Google LLC
```
chrisdunelm: Did you mean to leave this file in this PR?
(It doesn't really matter one way or the other)

jskeet: Not particularly - but as you say, it doesn't really matter. I might as well leave it in :)

PR #1831, opened 2018-01-16
Permit serialization of specific dictionaries

jskeet:
For example, users can now pass a `Dictionary<string, int>`
or `Dictionary<string, Value>` to the various
DocumentReference/CollectionReference/WriteBatch methods.

Will hold off merging until we have confirmation from @Thaina that this addresses their issue, but I think it's okay to review already.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
@@ -35,6 +35,9 @@ namespace Google.Cloud.Firestore
     /// </summary>
     internal static class ValueSerializer
     {
+        private static readonly MethodInfo s_serializeDictionaryMethod =
+            typeof(ValueSerializer).GetTypeInfo().GetMethod(nameof(SerializeDictionary), BindingFlags.Static | BindingFlags.NonPublic);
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
```diff
@@ -124,8 +128,13 @@ internal static Value Serialize(object value)
             GaxPreconditions.CheckNotNull(value, nameof(value));
             switch (value)
```
evildour: Perhaps not for this PR, but the four cases here are pretty much identical to those above (except the result is then passed to `CreateMapValue` above). Do you think it would make sense to extract them into some kind of `TrySerializeMap` helper?

jskeet: Potentially, yes. At some point I'd like to redo a lot of this code to use expression trees or something similar - we can make it a lot faster, I suspect. But yes, a little refactoring before then may be a good idea.

Thaina:
By seeing he test code I think my case should be covered already

jskeet:
Great. I'll merge this and create a new NuGet package.

PR #1838, opened 2018-01-17
[Bigtable] Sync up with latest RowMerge logic on Java side

evildour:


evildour:
Thanks Jon. Yes, I believe it has. Most of it is shifting around where verification happens anyway. The only new verification is making sure split cells report the same value size in each chunk, which exposed more errors in the original test data.

PR #1841, opened 2018-01-22
Update Firestore proto test framework

jskeet:
Instead of being a delimited file, we now have a TestSuite proto.

(I can split this into two commits if it's helpful. Basically you can ignore test.proto, Test.cs and test-suite.binproto.)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/Proto/test.proto
```diff
@@ -5,10 +5,16 @@ syntax = "proto3";
 package tests;
 
 option csharp_namespace = "Google.Cloud.Firestore.Tests.Proto";
+option java_package = "com.google.cloud.firestore.conformance";
```
chrisdunelm: This is slightly unexpected in a C# repo. Is it correct? Is this also going to be used for Java tests?

jskeet: The proto is copied (by script) from google-cloud-common

PR #1843, opened 2018-01-22
Pubsub class renames

chrisdunelm:
Auto-gen clients now named: `PublisherServiceApiClient` and `SubscriberServiceApiClient`.
Hand-written clients, previously `Simple*` now: `PublisherClient` and `SubscriberClient`.

PR #1845, opened 2018-01-23
Update pubsub version to beta16

chrisdunelm:
Only to be merged after #1843


jskeet:
Won't be mergeable until generateprojects.sh is rerun, of course... but then that's fine.

chrisdunelm:
generateprojects.sh run. Will push on green.

chrisdunelm:
Or even *merge* on green.

PR #1850, opened 2018-01-23
Update pubsub docs, after pubsub rename

chrisdunelm:


jskeet:
Note on process - while it's fine to do it like this, I personally usually just push straight to the gh-pages branch. I'm not sure why you've got conflicts though...

chrisdunelm:
Conflicts are due to my basing in on an out-of-date local version of the gh-pages branch. Should be fixed in a moment
I'd rather do it this way, so you can quickly double-check I haven't done anything foolish ;)

chrisdunelm:
Conflicts resolved. Please can you quickly check it? Thanks.

PR #1878, opened 2018-01-30
[Bigtable] Default CallSettings for idempotent MutateRow requests

evildour:


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -924,14 +924,27 @@ public partial class BigtableClientImpl
...
         partial void Modify_ReadRowsRequest(ref ReadRowsRequest request, ref CallSettings settings) =>
             ApplyResourcePrefixHeader(ref settings, request.TableName);
 
         partial void Modify_SampleRowKeysRequest(ref SampleRowKeysRequest request, ref CallSettings settings) =>
             ApplyResourcePrefixHeader(ref settings, request.TableName);
 
-        partial void Modify_MutateRowRequest(ref MutateRowRequest request, ref CallSettings settings) =>
+        partial void Modify_MutateRowRequest(ref MutateRowRequest request, ref CallSettings settings)
+        {
+            if (settings == null && request.IsIdempotent())
```
jskeet: I don't think we can use settings in this way - we want to merge the call timing in with settings, if the user hasn't specified their own call timing, I think.
(The call settings can contain a cancellation token or auth for example, which shouldn't affect the retry semantics.)

Basically, the settings here aren't quite the same as the ones in ApiCall, which is what the defaults are used for.

You could potentially use

```csharp
if (request.IsIdempotent()
{
    settings = _idempotentMutateRowSettings.MergedWith(settings);
}
```

@chrisdunelm Does that sound right to you?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableSettingsPartial.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <item><description>Timeout maximum delay: 20000 milliseconds</description></item>
+        /// </list>
+        /// Retry will be attempted on the following response status codes:
+        /// <list>
+        /// <item><description><see cref="StatusCode.DeadlineExceeded"/></description></item>
+        /// <item><description><see cref="StatusCode.Unavailable"/></description></item>
+        /// </list>
+        /// Default RPC expiration is 600000 milliseconds.
+        /// </remarks>
+        public CallSettings IdempotentMutateRowSettings { get; set; } = CallSettings.FromCallTiming(
```
jskeet: I suspect this doesn't want to be a `CallSettings`, but just a `RetrySettings`, given where it's being applied.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Ok, thanks. I'll use this (unless Chris disagrees).

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableSettingsPartial.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -924,14 +924,28 @@ public partial class BigtableClientImpl
...
         partial void Modify_ReadRowsRequest(ref ReadRowsRequest request, ref CallSettings settings) =>
             ApplyResourcePrefixHeader(ref settings, request.TableName);
 
         partial void Modify_SampleRowKeysRequest(ref SampleRowKeysRequest request, ref CallSettings settings) =>
             ApplyResourcePrefixHeader(ref settings, request.TableName);
 
-        partial void Modify_MutateRowRequest(ref MutateRowRequest request, ref CallSettings settings) =>
+        partial void Modify_MutateRowRequest(ref MutateRowRequest request, ref CallSettings settings)
+        {
+            if (request.IsIdempotent())
```
jskeet: If the caller has passed in other timing settings already, I'm not sure we'd want to change them. I suspect you can keep `_idempotentMutateRowSettings` as a `CallTiming`, and then use:

```csharp
if (request.IsIdempotent() && settings?.Timing == null)
{
    settings = settings.WithCallTiming(_idempotentMutateRowSettings);
}
```

Does that sound right to you?

evildour: I was under the impression that MergedWith would only take the receiver's settings if they were not set in the argument's. The comments say "with overlaid taking priority". If the caller passed in other timing settings, won't those be given priority?

jskeet: Ah, of course, you're right. Sorry!

jskeet:
Explicitly added @chrisdunelm to review this too, as I'm getting settings-blindness with it. I suspect it's okay, but a second pair of eyes would help.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableSettingsPartial.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <item><description>Timeout maximum delay: 20000 milliseconds</description></item>
+        /// </list>
+        /// Retry will be attempted on the following response status codes:
+        /// <list>
+        /// <item><description><see cref="StatusCode.DeadlineExceeded"/></description></item>
+        /// <item><description><see cref="StatusCode.Unavailable"/></description></item>
+        /// </list>
+        /// Default RPC expiration is 600000 milliseconds.
+        /// </remarks>
+        public RetrySettings IdempotentMutateRowRetrySettings { get; set; } =
```
chrisdunelm: Why is this only a `RetrySettings`, not a complete `CallSettings`?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableSettingsPartial.cs
```diff
@@ -0,0 +1,55 @@
...
+// limitations under the License.
+
+using Google.Api.Gax;
+using Google.Api.Gax.Grpc;
+using Grpc.Core;
+using System;
+
+namespace Google.Cloud.Bigtable.V2
+{
+    public partial class BigtableSettings
```
chrisdunelm: The partial method `OnCopy(existing)` should be implemented to copy `IdempotentMutateRowRetrySettings`

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableSettingsPartial.cs
evildour: I originally had it as `CallSettings`, but Jon felt it should be `RetrySettings` instead. I'm fine either way. Given the merging capability, I guess it just depends on what the convention is for something like this (if there is one) and whether we feel that the user will want to modify other settings for these kinds of calls or if they would only want to customize the retry settings. I'm guessing it's the latter.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableSettingsPartial.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableSettingsPartial.cs
jskeet: While it's tricky to remember precise motivations, I suspect my suggestion around RetrySettings may have been linked to my misguided suggestion around the merging. So long as we can make merging work appropriately, I'm okay for it to be a CallSettings.

evildour: Ok, I changed it back. Also added a comment about how these will be merged with MutateRowSettings.

evildour:
I'm going to assume this is ok to merge now. If there's any issues, let me know and I'll fix them in a followup PR.

PR #1879, opened 2018-01-31
Retention policy locking feature.

jskeet:
This will remain in a branch until the feature is more publicly available.

(I can't run the integration test just yet as I'm not whitelisted - will do so as soon as I can.)


jskeet:
(Made minor modification to snippet and integration test, but thrust remains the same. Will merge on green - still just to the separate branch though.)

chrisdunelm:
sgtm

PR #1887, opened 2018-02-02
Multiple minor Firestore changes

jskeet:
Suggest reviewing one commit at a time

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/QuerySnapshotTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/QuerySnapshotTest.cs
@@ -1,0 +1,57 @@
+﻿// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Firestore.V1Beta1;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+using static Google.Cloud.Firestore.Tests.ProtoHelpers;
+
+namespace Google.Cloud.Firestore.Tests
+{
+    public class QuerySnapshotTest
+    {
+        [Fact]
+        public void ConvenienceMembers()
+        {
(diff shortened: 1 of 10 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/QuerySnapshotTest.cs
```diff
@@ -0,0 +1,57 @@
...
+            Assert.Equal(2, querySnapshot.Count);
+            // Indexer
+            Assert.Same(doc1, querySnapshot[0]);
+            Assert.Same(doc2, querySnapshot[1]);
+            Assert.Throws<ArgumentOutOfRangeException>(() => querySnapshot[-1]);
+            Assert.Throws<ArgumentOutOfRangeException>(() => querySnapshot[2]);
+            // Use IEnumerable<DocumentSnapshot>
+            Assert.Equal(docs, querySnapshot);
+            // Use IEnumerable
+            Assert.Equal(docs, querySnapshot.Cast<object>().Cast<DocumentSnapshot>());
```
chrisdunelm: nit: why the double cast? Why not a single `Cast()` to `DocumentSnapshot`?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Utilities/EqualityHelpers.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// <summary>
+        /// Computes a hash code for a list
+        /// </summary>
+        internal static int GetListHashCode<T>(IReadOnlyList<T> list) where T : IEquatable<T>
+        {            
+            if (list == null)
+            {
+                return 0;
+            }
+            unchecked
```
chrisdunelm: nit: unchecked is the default, so is there a need to explicitly use it?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/QuerySnapshotTest.cs
jskeet: It's an attempt to confuse any optimization LINQ might have that notices it's already the right iterable type. I'm not 100% sure it's effective, but I'm hopeful.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Utilities/EqualityHelpers.cs
jskeet: While it's the default, I suspect we may well be tempted to make checked the default at some point... but hash combinations should *never* be done in a checked context.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/QuerySnapshotTest.cs
jskeet: I think OfType will work better for this, actually:
https://codeblog.jonskeet.uk/2011/01/13/reimplementing-linq-to-objects-part-33-cast-and-oftype/

PR #1897, opened 2018-02-08
[Bigtable] NOT FOR SUBMISSION MutateRows retries

evildour:
Not for submission...yet

Best to review one commit at a time.

1st commit contains hand-written changes to the auto-generated BigtableClient. If these look good, I'll update the generator to make these kinds of changes for all client files: adding OnConstruction partial method to the settings and allowing modification of the response streams for streaming methods. The latter could be done for all methods too, but I thought it would be best to reduce churn to the minimum amount necessary. After the generator is updated and Bigtable is regenerated, then this will be ready for submission.

Two things to note here:
* This differs from the approach originally discussed offline. Original plan: add a BigtableSettings.MutateRowsStreamRetrySettings property for the default retry settings and a MutateRowsStream.WithRetry(RetrySettings) method to override the default retry settings. As I was prototyping this though, I realized there was a better approach: change the default BigtableSettings.MutateRowsSettings to have RetrySettings, strip them out when making the initial request, but then use them to manage the retry logic. Two benefits here: 1) users can use the normal MutateRowsSettings property and/or CallSettings parameter to BigtableClient.MutateRows(...) and don't have to deal with two separate properties with two separate ways of overridding default settings, and 2) when internally firing off retry requests, we don't need to use stream.WithRetry(null) to opt out of the nested retry stream, which avoids the unnecessary wrapping and then unwrapping of the retry request's stream. However, drawback here is there is a fairly complicated Modify_MutateRowsResponse signature. I think it is acceptable, but would like to hear your thoughts.
* Since the retry logic is invoked from the BigtableClientImpl.Modify_MutateRowsResponse partial method, the RetrySettings of the MutateRowsSettings are only honored when the derived client type is BigtableClientImpl. I asusme this is ok though because any user-dervied BigtableClient meant to actually make the calls will likely contain a BigtableClientImpl to delegate to, so it will still get the settings honored. Please let me know if this assumption is wrong though.


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -946,8 +952,25 @@ public partial class BigtableClientImpl
...
 
-        partial void Modify_MutateRowsRequest(ref MutateRowsRequest request, ref CallSettings settings) =>
+        partial void Modify_MutateRowsRequest(ref MutateRowsRequest request, ref CallSettings settings)
+        {
+            // Strip off the retry settings specified. We will apply them to the response stream.
+            // However, keep the overall expiration so the first individual call can use it.
+            if (settings?.Timing?.Retry != null)
+            {
+                settings = settings.WithCallTiming(CallTiming.FromExpiration(settings.Timing.Retry.TotalExpiration));
+            }
```
evildour: Actually, I think I just realized an error with this approach. The settings passed in here will probably not have the defaults from the BigtableSettings merged in, so I don't think we'd get the RetrySettings here by default. I'll verify tomorrow and fix it if it is an issue. Perhaps I can add some testing into the integration tests as well to verify the type of stream returned that it is setup correctly. I think some logic here and in the method below might change, but feel free to review the rest of the PR.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
```diff
@@ -994,6 +998,7 @@ public BigtableClientImpl(Bigtable.BigtableClient grpcClient, BigtableSettings s
         partial void Modify_SampleRowKeysRequest(ref SampleRowKeysRequest request, ref CallSettings settings);
         partial void Modify_MutateRowRequest(ref MutateRowRequest request, ref CallSettings settings);
         partial void Modify_MutateRowsRequest(ref MutateRowsRequest request, ref CallSettings settings);
+        partial void Modify_MutateRowsResponse(MutateRowsRequest request, ref MutateRowsStream response, CallSettings originalCallSettings, CallSettings settings);
```
jskeet: So would you anticipate this being generated for all streaming RPCs, but only those?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
```diff
@@ -1140,8 +1145,11 @@ internal SampleRowKeysStreamImpl(AsyncServerStreamingCall<SampleRowKeysResponse>
             MutateRowsRequest request,
             CallSettings callSettings = null)
         {
+            var originalCallSettings = callSettings;
```
jskeet: Needing both the original call settings and the modified ones is a little odd. I still need to review the rest of the PR to understand what's going on here, but I'd love to find a simpler approach if possible.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ApiCallRetryExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+                    DateTime attemptDeadline = clock.GetCurrentDateTimeUtc() + callTimeout;
+                    // Note: this handles a null total deadline due to "<" returning false if overallDeadline is null.
+                    DateTime combinedDeadline = overallDeadline < attemptDeadline ? overallDeadline.Value : attemptDeadline;
+                    CallSettings attemptCallSettings = callSettings.WithCallTiming(CallTiming.FromDeadline(combinedDeadline));
+                    TimeSpan actualDelay = retrySettings.DelayJitter.GetDelay(retryDelay);
+                    DateTime expectedRetryTime;
+                    try
+                    {
+                        var response = await fn(request, attemptCallSettings).ConfigureAwait(false);
+                        bool isResponseOk = false;
```
jskeet: Shouldn't this be true? Alternatively:

```csharp
bool isResponseOk = postResponse == null ? true : await postResponse(response).ConfigureAwait(false);
```

Or

```csharp
bool isResponseOk = await (postResponse?.Invoke(response) ?? Task.FromResult(true)).ConfigureAwait(false);
```

But probably the first.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ApiCallRetryExtensions.cs
```diff
@@ -0,0 +1,97 @@
...
+                        if (postResponse != null)
+                        {
+                            isResponseOk = await postResponse(response).ConfigureAwait(false);
+                        }
+
+                        expectedRetryTime = clock.GetCurrentDateTimeUtc() + actualDelay;
+                        if (isResponseOk || expectedRetryTime > overallDeadline)
```
jskeet: If the response is "bad" but we've exceeded the deadline, wouldn't we want to throw something?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -153,6 +153,14 @@ internal ProcessingStatus OnOk()
                 }
                 if (_retryableCodes.Contains(status.Code))
                 {
+                    // Skip non-idempotent entries. We could check this at the other continue above, but since
+                    // this check needs to walk through each mutation in the entry, best to way until we know
```
jskeet: best to *wait*?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableMutateRowsRequestManager.cs
```diff
@@ -153,6 +153,14 @@ internal ProcessingStatus OnOk()
                 }
                 if (_retryableCodes.Contains(status.Code))
                 {
+                    // Skip non-idempotent entries. We could check this at the other continue above, but since
```
jskeet: I think I need to look at *all* of this more carefully - I'm losing the plot of how it all hangs together.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableSettingsPartial.cs
```diff
@@ -16,11 +16,34 @@
...
+using System.Diagnostics;
 
 namespace Google.Cloud.Bigtable.V2
 {
     public partial class BigtableSettings
     {
+        partial void OnConstruction()
+        {
+            var originalMutateRowsSettings = MutateRowsSettings;
+            Debug.Assert(
```
jskeet: Make this a proper GaxPrecondition call? It's cheap, after all.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RetryingMutateRowsStream.cs
```diff
@@ -0,0 +1,128 @@
...
+        private BigtableClient.MutateRowsStream _currentStream;
+
+        internal RetryingMutateRowsStream(
+            BigtableClient client,
+            MutateRowsRequest initialRequest,
+            BigtableClient.MutateRowsStream initialStream,
+            CallSettings callSettings,
+            RetrySettings streamRetrySettings)
+        {
+            Debug.Assert(!(initialStream is RetryingMutateRowsStream));
```
jskeet: Likewise I'd use GAX preconditions - I dislike Debug.Assert in general.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RetryingMutateRowsStream.cs
```diff
@@ -0,0 +1,128 @@
...
+            _responseEnumerator = new RetryingResponseEnumerator(this);
+        }
+
+        /// <inheritdoc/>
+        public override AsyncServerStreamingCall<MutateRowsResponse> GrpcCall => _currentStream.GrpcCall;
+
+        /// <inheritdoc/>
+        public override IAsyncEnumerator<MutateRowsResponse> ResponseStream => _responseEnumerator;
+
+        private class RetryingResponseEnumerator : IAsyncEnumerator<MutateRowsResponse>
```
jskeet: So even though this is shown as a stream, it's really just going to return a single value? I'm wondering whether we should look at a different approach to this, leaving the "raw" stream methods as they are, but adding an extra method or set of methods to do all of the retry parts and just returning a `Task<MutateRowsResponse>`. That would be easier for the client to consume, and would probably involve less worrying about getting "inside" the infrastructure.

evildour:
@chrisdunelm Hold off on reviewing. I just discussed this offline with Jon and I'm going to try another approach in a different branch.

However, @jskeet I think I'm going to update the partial method to remove the `originalCallSettings` weirdness here just so we can see what that looks like too.

evildour:
Ok, `originalCallSettings` is removed. I'll start on the alternate approach next week.

jskeet:
Is this now obsolete?

evildour:
Yes

PR #1913, opened 2018-02-20
 Update buildrelease.sh to also build docs for gh-pages

iantalarico:
This will help with automating the release.

I think this is as far as we should take the script. Leave the pushing code out of the script and just add it on the release machine.


buildrelease.sh
```diff
@@ -48,13 +48,31 @@ done
...
+cd releasedocs
+git checkout gh-pages
+
+for project in $projects
+do
+  rm -r ./docs/$project
+  mv ../docs/output/assembled/$project ./docs
+done
+
+rm -r ./index.html
```
jskeet: Why the -r when this is a file rather than a directory?

buildrelease.sh
```diff
@@ -48,13 +48,31 @@ done
...
-echo "Release build complete for the following projects:"
+cd ..
+rm -rf releasedocs
+git clone https://github.com/GoogleCloudPlatform/google-cloud-dotnet.git releasedocs -c core.autocrlf=input
+cd releasedocs
+git checkout gh-pages
+
+for project in $projects
+do
+  rm -r ./docs/$project
```
jskeet: This will fail for new projects; use -rf instead to handle the directory not being present before.

buildrelease.sh
```diff
@@ -48,13 +48,31 @@ done
 cd docs
 ./builddocs.sh root $projects
 
-echo "Release build complete for the following projects:"
+cd ..
```
jskeet: It might be simpler to change lines 48/49 to:

```sh
(cd docs && ./builddocs.sh root $projects)
```

Then we don't need to come back to this directory.

Or better yes, maybe builddocs.sh should handle being invoked from the root directory... I can look at that later.

buildrelease.sh
```diff
@@ -48,13 +48,31 @@ done
...
+rm -rf releasedocs
+git clone https://github.com/GoogleCloudPlatform/google-cloud-dotnet.git releasedocs -c core.autocrlf=input
```
jskeet: This would be better using `-b gh-pages --depth 1` (and then remove the git checkout below). We don't need any other branches, or full history.

iantalarico: Done

buildrelease.sh
iantalarico: Done

buildrelease.sh
iantalarico: No reason, fixed.

buildrelease.sh
iantalarico: Done

iantalarico:
Thanks!

iantalarico:
@jskeet Once this is in I'll check the next release we do then automate it if everything looks goo :)

PR #1917, opened 2018-02-21
[Bigtable] Split client into API client and user client

evildour:
This is part 1 of a refactoring to eventually allow for a AppProfileId specifications, multi-channel support, and smart retries for ReadRows and MutateRows. The latter two are not included with this change.


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
```diff
@@ -0,0 +1,286 @@
...
+{
+    class Program
+    {
+        private const string AppProfileIdFieldName = "_appProfileId";
+        private const string AppProfileIdPropertyName = "AppProfileId";
+        private const string GetClientMethodName = "GetClient";
+
+        private static async Task<int> Main(string[] args)
+        {
+            // TODO: Figure out why `dotnet run` from generateapis.sh is sending 6 args instead of 3 as in: arg1 arg2 arg3 arg1 arg2 arg3
```
jskeet: You don't have a "--" in generateapis.sh to separate the program arguments from the other dotnet arguments. If you change it to 

```sh
dotnet run -f net461 -p apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient -- \
apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.csproj \
BigtableServiceApiClient \
BigtableClient
```

then that might fix it. (I don't know why it would be doubling up, but...)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
```diff
@@ -0,0 +1,286 @@
...
+                    .WithLeadingTrivia(generatedApiClientSyntax.GetLeadingTrivia());
+            var userClientImplSyntax =
+                (ClassDeclarationSyntax)generator.ClassDeclaration(
+                    userClientName + "Impl",
+                    accessibility: Accessibility.Public,
+                    modifiers: DeclarationModifiers.Sealed | DeclarationModifiers.Partial,
+                    baseType: ParseTypeName(userClientName))
+                    .WithLeadingTrivia(generatedApiClientSyntax.GetLeadingTrivia());
+
+            foreach (var methodSyntax in generatedApiClientSyntax.Members.OfType<MethodDeclarationSyntax>())
```
jskeet: Some comments on this bit could be useful. I'm surprised we're only copying the methods with request objects. Maybe more will come in later commits?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -24,6 +25,90 @@ namespace Google.Cloud.Bigtable.V2
...
+
+        /// <summary>
+        /// Creates a <see cref="BigtableClient"/> which uses the specified channel for remote operations.
+        /// </summary>
+        /// <param name="channel">The <see cref="Channel"/> for remote operations. Must not be null.</param>
+        /// <param name="settings">Optional <see cref="BigtableServiceApiSettings"/>.</param>
+        /// <param name="appProfileId">
+        /// This is a private alpha release of Cloud Bigtable replication. This feature
+        /// is not currently available to most Cloud Bigtable customers. This feature
+        /// might be changed in backward-incompatible ways and is not recommended for
```
jskeet: I'm nervous about this statement - we should talk about what sort of backward-incompatibility we mean here. Might the actual signatures change, or is it just the semantics?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.sln
```diff
@@ -105,7 +105,6 @@ Global
 		{97EC9E30-ADD3-4D80-A2D5-8D611557813E}.Debug|x86.ActiveCfg = Debug|Any CPU
 		{97EC9E30-ADD3-4D80-A2D5-8D611557813E}.Debug|x86.Build.0 = Debug|Any CPU
 		{97EC9E30-ADD3-4D80-A2D5-8D611557813E}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{97EC9E30-ADD3-4D80-A2D5-8D611557813E}.Release|Any CPU.Build.0 = Release|Any CPU
```
jskeet: I don't understand this change. I can see why we wouldn't want to *run* the generator, but why wouldn't we want to even compile it?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
evildour: I tried that too. Same result. From the docs, it looked like it was optional for this purpose anyway.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
evildour: Done. I would like to generate more, but no plans at the moment. There was a happy coincidence here where I had already copied the flattened methods so I could provide overloads with a `BigtableByteString` parameter in place of the `ByteString` row keys. Those just now delegate to these copied methods instead of the API client's methods.

evildour: (now that I have this generator, I would like to auto-generate those copies as well, but it isn't needed for this refactoring)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: This is copied from the `AppProfileId` property docs on one of the request classes. I figured since this library was also alpha, it would be ok to add this here and potentially break it, but if you'd feel more comfortable, we could add overloads or new methods entirely which allow you to specify this parameter.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
jskeet: Ah, so the partial class that used to target the "low-level" API now just targets the new API?
Presumably that means we don't have any overloads accepting ByteString, and we're okay with that?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
jskeet: Maybe leave it as it is, but add an issue so that we don't forget to address this (one way or the other) before the library goes GA?

evildour: Oh, and to answer your question, I can't say if the signature will change. I can ask internally if you'd like.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.sln
evildour: The solution wasn't building on travis, so this was a bit of a hack, but here was my thinking: when debugging, you are likely to have the solution open in VS with the Debug configuration selected. In that case, you want the project to build with the solution. However, when the solutions are building in Release config on the CI machines, there's no need to build it. When running the project directly from generateapis.sh, it will get built automatically before running.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
evildour: Correct.

Technically yes, but since there's an implicit conversion from `ByteString` to `BigtableByteString`, no one should notice the change if they were specifying `ByteString` directly.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Good idea. Will do.

evildour:
Oh and to address your initial comment: all tests are still passing.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
jskeet: Okay, that's weird. I'll have a look at it at some point, but it does seem very strange...

jskeet:
Not sure where the comment I received by email about it not building on Travis has gone, but I'd prefer to remove the hack if possible... but I'm okay with it going in as-is for now, just to unblock you.

evildour:
Well the other possibility is to modify `build.sh`. Would you prefer that instead?

jskeet:
Without knowing why it didn't build, it's hard to know for sure. I'd rather we just got it to build under Travis :) Note that over the last 24 hours we've updated the SDK version - maybe it will build on Travis now anyway? I'm happy to look into it later though.

evildour:
Here is the build log: https://travis-ci.org/GoogleCloudPlatform/google-cloud-dotnet/builds/344500517

The 4.6.1 assemblies could not be found. I just assumed that version was not supported on Travis.

jskeet:
Ah, does it rely on some part of Roslyn that doesn't work on .NET Core yet? If so, that makes sense. Will have a think about a better way to go long-term, but that's fine to start with.

evildour:
Yes, the `MSBuildWorkspace` to load the csproj file. That only exists in the full framework.

PR #1918, opened 2018-02-21
Add trace ids to log entries if they exist.

### LINKED ISSUE #1104: Add trace Ids to log entries if the id exists
We can match traces with logs if we add trace ids to log entires.

iantalarico:
We do this by allowing the IHttpContextAccessor to be passed into the logger.

The only disadvantage is an IHttpContextAccessor is not available during the initial initialization of netcoreapp 2.0 apps so the user would have to also initialize the logger in their configure method.  The alternative would require a large change to the API of this library and how we handle DI of most of our bits.

Fixes #1104

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
@@ -247,10 +247,17 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
         {
             services.TryAddSingleton<IHttpContextAccessor, HttpContextAccessor>();
             services.AddMvc();
+            services.AddGoogleTrace(options => 
+            {
+                options.ProjectId = ProjectId;
+                // Don't actually trace anything.
+                options.Options = TraceOptions.Create(qpsSampleRate: 0.00000001);
+            });
         }
 
         public void SetupRoutes(IApplicationBuilder app)
         {
+            app.UseGoogleTrace();
             app.UseMvc(routes =>
             {
                 routes.MapRoute(
@@ -262,7 +269,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
 
     /// <summary>
     /// An application that has a <see cref="GoogleLogger"/> with no buffer that will accept all logs
-    /// of level warning or above.  Adds trace information to log entries if available.
+    /// of level warning or above.
     /// </summary>
     public class NoBufferWarningLoggerTestApplication : LoggerTestApplication
     {
@@ -272,11 +279,9 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
             SetupRoutes(app);
             LoggerOptions loggerOptions = LoggerOptions.Create(
                 LogLevel.Warning, null, null, BufferOptions.NoBuffer());
-            loggerFactory.AddGoogle(ProjectId, loggerOptions, accessor: accessor);
+            loggerFactory.AddGoogle(ProjectId, loggerOptions);
         }
     }
-
-
 
     /// <summary>
     /// An application that has a <see cref="GoogleLogger"/> with no buffer that will accept all logs
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
@@ -20,9 +20,11 @@ using Google.Cloud.Logging.Type;
 using Google.Cloud.Logging.V2;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.TestHost;
 using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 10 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
```diff
@@ -239,19 +262,22 @@ public void SetupRoutes(IApplicationBuilder app)
...
         {
             SetupRoutes(app);
             LoggerOptions loggerOptions = LoggerOptions.Create(
                 LogLevel.Warning, null, null, BufferOptions.NoBuffer());
-            loggerFactory.AddGoogle(ProjectId, loggerOptions);
+            loggerFactory.AddGoogle(ProjectId, loggerOptions, accessor: accessor);
         }
     }
 
+
```
jskeet: Remove extra lines?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
```diff
@@ -146,5 +148,36 @@ public void Log()
...
             mockConsumer.VerifyAll();
         }
+
+        [Fact]
+        public void Log_Trace()
+        {
+            string traceId = "105445aa7843bc8bf206b12000100f00";
+            string fullTraceName = TraceTarget.ForProject(_projectId).GetFullTraceName(traceId);
+
+            Predicate<IEnumerable<LogEntry>> matcher = (l) =>
```
jskeet: Nit: I'd rename l to logEnties and remove the parentheses around the lambda parameter:

```csharp
Predicate<IEnumerable<LogEntry>> matcher = logEntries =>
...
```

Not a blocker though.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLoggerFactoryExtensions.cs
```diff
@@ -50,13 +51,15 @@ public static class GoogleLoggerFactoryExtensions
         ///     detected from the platform.</param>
         /// <param name="options">Optional, options for the logger.</param>
         /// <param name="client">Optional, logging client.</param>
+        /// <param name="accessor">Optional, HTTP context accessor. Allows adding trace ids to log entries.</param>
```
jskeet: Adding an extra parameter is a breaking change. Unfortunately adding an overload is tricky too - let's talk about this on a VC or similar.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLoggerProvider.cs
```diff
@@ -54,8 +61,10 @@ internal GoogleLoggerProvider(IConsumer<LogEntry> consumer, LogTarget logTarget,
         ///     detected from the platform.</param>
         /// <param name="options">Optional, options for the logger.</param>
         /// <param name="client">Optional, logging client.</param>
+        /// <param name="accessor">Optional, HTTP context accessor. Allows adding trace ids to log entries.</param>
         public static GoogleLoggerProvider Create(string projectId = null,
-            LoggerOptions options = null, LoggingServiceV2Client client = null)
+            LoggerOptions options = null, LoggingServiceV2Client client = null,
```
jskeet: Ditto previous comment.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/TraceTarget.cs
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+
+namespace Google.Cloud.Diagnostics.Common
+{
+    /// <summary>
+    /// Represents the location traces will be sent to.  The Trace API only supports projects.
```
jskeet: Presumably although this needs to be public, we don't expect users to need to use it directly?

iantalarico: Yea this needs to be public as it's in Common but I don't think the public will ever use it.

jskeet:
Will look on Monday - weekend has started :)

jskeet:
Monday has nearly entirely been taken up with other things - it seems unlikely I'll get to it today, unfortunately. Tomorrow!

iantalarico:
@jskeet No rush at all, this isn't blocking anything.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -83,6 +83,7 @@ public static void UseGoogleTrace(this IApplicationBuilder app)
         {
             GaxPreconditions.CheckNotNull(app, nameof(app));
             app.UseMiddleware<CloudTraceMiddleware>();
+            HttpContextAccessorWrapper.Accessor = app.ApplicationServices.GetService<IHttpContextAccessor>();
```
jskeet: Will GetService return null if IHttpContextAccessor hasn't been set up, or will it throw?

iantalarico: Correct, it will return null if the service cannot be found.

PR #1924, opened 2018-02-25
[Bigtable] NOT FOR SUBMISSION MutateRows retries (2nd attempt)

evildour:
[Part 3 of refactoring from #1917]

Not for submission...yet

Best to review one commit at a time.

2nd commit contains hand-written changes to the auto-generated BigtableClient. If these look good, I'll update the generator to make these kinds of changes for all client files: adding OnConstruction partial method to the settings. After the generator is updated and Bigtable is regenerated, then this will be ready for submission.

Since the retry logic waits for all responses from the original and retry requests and then packages all results into a single MutateRowsResponse, we can treat MutateRows as if that is really the return type and then have an associated MutateRowsAsync methods as well. Essentially, we are turning the streaming call into a unary call.


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
```diff
@@ -183,42 +240,52 @@ public override SyntaxNode VisitMethodDeclaration(MethodDeclarationSyntax node)
...
             {
                 node = (XmlElementSyntax)base.VisitXmlElement(node);
 
-                // Append request param tags with a note about how the AppProfileId will be added to the request object
-                // if unspecified.
-                if (node.StartTag.Name.LocalName.ToString() == "param" &&
+                var content = node.Content;
+
+                if (s_customStreamMethods.TryGetValue(_method.Name, out var customStreamMethodInfo) &&
+                    customStreamMethodInfo.CustomDocs.TryGetValue(node.StartTag.ToString(), out var comment))
```
jskeet: Could we use the start tag name rather than the ToString representation? Then replace "<summary>" with "summary" in the dictionary.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ApiCallRetryExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+using Google.Api.Gax.Grpc;
+using Grpc.Core;
+using System;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Bigtable.V2
+{
+    internal static class ApiCallRetryExtensions
+    {
+        // TODO: This is a modified/simplified version of ApiCallRetryExtensions.WithRetry from Google.Api.Gax.Grpc. Can we combine them somwhow?
```
jskeet: Nit: somwhow  => somehow :)

But yes, it would be good to remove this duplication. Suggest we look at that again after all the other refactoring is done?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -23,6 +23,32 @@
 
 namespace Google.Cloud.Bigtable.V2
 {
+    public partial class BigtableServiceApiSettings
+    {
+        partial void OnConstruction()
```
jskeet: I thought we were changing MutateRowsSettings to retry by default, now that it's seen as idempotent? Maybe I've missed something. I'm most fine with this as a temporary measure anyway - but it does mean that if a caller has modified the default settings (e.g. to use different credentials) then this would fail, right?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MockBigtableClient.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Bigtable.V2.Tests
+{
+    public class RequestMadeException : Exception { }
+
+    public class MockBigtableClient : BigtableClient
```
jskeet: I think you should be able to use Moq for this if it's easier to do so, but I don't mind a hand-rolled mock either.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingMutateRowsTest.cs
```diff
@@ -0,0 +1,185 @@
...
+                    new[]
+                    {
+                        // Original entry 1
+                        MutateRowsStreamUtilities.CreateEntry(0, Code.Ok),
+                        // Original entry 3
+                        MutateRowsStreamUtilities.CreateEntry(1, Code.DeadlineExceeded),
+                        // Original entry 4
+                        MutateRowsStreamUtilities.CreateEntry(2, Code.Ok)
+                    },
+                    // 2end retry response entries
```
jskeet: Nit: 2nd

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
```diff
@@ -33,12 +33,48 @@ class Program
     {
         private const string AppProfileIdFieldName = "_appProfileId";
         private const string AppProfileIdPropertyName = "AppProfileId";
-        private const string GetClientMethodName = "GetClient";
+        private const string GetUnderlyingClientMethodName = "GetUnderlyingClient";
```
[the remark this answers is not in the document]
evildour: It does currently return a field, but that's just a stub for the real implementation (being worked on in parallel), which will cycle through multiple underlying clients. This method, just renamed because it is now protected, could (and probably will) return a different client on each invocation.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
```diff
@@ -33,12 +33,48 @@ class Program
...
 
-        // TODO: Figure out how to replace the comments on BigtableClient.ReadRows describing cell chunks. Perhaps make this a
-        //       dictionary that maps the custom stream name to the summary comment that should be replaced on the associated
-        //       method. We can just make a custom summary then.
-        private static readonly HashSet<string> s_customStreams = new HashSet<string> { "ReadRowsStream" };
+        private class CustomStreamMethodInfo
+        {
+            public readonly Dictionary<string, string> CustomDocs = new Dictionary<string, string>();
```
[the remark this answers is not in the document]
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Yes, that's the idea. Perhaps I misunderstood this part of it though. I assumed we wanted to retry by default but still let users control the retry settings, which I would assume would mean we need to update the default call settings for MutateRows. Is there a better way to do this? Or do we not want to let users customize the retry settings?

As far as specifying other settings, I realize I was only considering per-call settings, which would merge correctly with these default retry settings, and therefore now replace them. But if they change the settings when creating the client, yes this would fail.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MockBigtableClient.cs
evildour: You might be right. Added a TODO to look into this later.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingMutateRowsTest.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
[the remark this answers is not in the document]
jskeet: Ah yes - sorry for forgetting about that. A method makes a lot of sense then.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Also, how does this work in general. If you want to use different credentials with MutateRow, which has auto-generated settings that retry by default now, wouldn't that also prevent retry?

jskeet: If the GAPIC config is changing to make MutateRows an idempotent call, the default settings should automatically change to allow retries. On the other hand, it could well be that I've missed something, particularly with it being a streaming method.

If we're only trying to change the timing, maybe we should not check anything else, and just use `originalMutateRowsSettings.WithTiming(...)`?

jskeet: Sorry, hadn't seen the second comment. I'd expect *most* users to only specify a CallSettings on a per-call basis for modifications, but it's possible that some will modify the default settings instead. It's possible that what's currently in the master branch won't work with this properly and that I failed to review properly before.

We should also consider what we want to happen if the user specifies a custom CallSettings because they want to change the timing, but they *do* still want retry. Oh, it's confusing :(

evildour: Oh, I see. Sorry, I misunderstood your "retry by default" comment originally to mean functionally, not in the auto-generated stuff. So the config is updated: https://github.com/googleapis/googleapis/blob/master/google/bigtable/v2/bigtable_gapic.yaml#L104, but the generator did not generate retry settings: https://github.com/evildour/google-cloud-dotnet/blob/master/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClient.cs#L183. I'm guessing this is because it is a streaming API so that goes through a different path. @chrisdunelm can you confirm?

Yes, using `WithCallTiming` seems much cleaner. I'll switch over to that if this `OnConstruction` is still needed.

As for my "how does this work in general" comment, I'm not even talking about what's there for MutateRow specifically. Take `FirestoreClient.GetDocument` for example: that has default idempotent retry settings. If a caller has modified the default settings to use different credentials for that method, retry for that would fail too right?

jskeet: It depends on how they modify the default settings. IIRC, we throw an exception if you try to create an ApiCall using base settings that have *no* timing - but if the client decided to create a timing that didn't include any retry then yes, that wouldn't retry.

More likely, they'd use `existingCallSettings.WithCredentials(...)` to create the new one.

evildour: Ok, so it sounds like the same thing should happen here (unless the validation for call timing doesn't happen for the streaming calls). So it sounds like it's safe to say your original concern is not an issue, but correct me if I'm wrong.

The only question left now is why we're not auto-generating these retry settings despite the GAPIC config specifying them. I'll update this code to with WithCallTiming as you suggest while we wait for clarification on that.

jskeet: Right - the difference is that we can't easily detect "the user has changed the call timing and we should obey that" vs "we should replace the call timing with a retry". But if we get to the bottom of the autogeneration diff, it may not matter.

evildour:
You think I should merge even with the manually added code to the auto-gened file (the partial On Construction)?

jskeet:
Yes, but maybe add an issue so we don't forget to either add the codegen or sort out the settings.

evildour:
Ok thanks, will do.

PR #1931, opened 2018-02-27
Change set of steps required to release docs

jskeet:
"git commit -a" doesn't include *new* files; "git add --all"
will add all modifications *and* all new files.


jskeet:
Which we would, as releasedocs is a freshly-cloned repo, so it's the root.

evildour:
Oh right, so in this case, they'd be identical I believe. `--all` is probably better since with `.` someone might fail to copy it, thinking its the end of the line.

jskeet:
I note that Travis ran out of time building this. That's a bit alarming - I'll look into it when I get a chance.

PR #1940, opened 2018-03-01
Update the correct index.html file during the release.

### LINKED ISSUE #1939: Documentation is broken
All the links from https://googlecloudplatform.github.io/google-cloud-dotnet/ are 404s.
And it looks like some CSS might be missing from the index page.

iantalarico:
Fixes #1939

### DIFF
--- a/buildrelease.sh
+++ b/buildrelease.sh
@@ -57,8 +57,8 @@ do
   mv ../docs/output/assembled/$project ./docs
 done
 
-rm ./index.html
-mv ../docs/output/assembled/index.html ./
+rm ./docs/index.html
+mv ../docs/output/assembled/index.html ./docs/
 
 echo "Release build and docs complete for the following projects:"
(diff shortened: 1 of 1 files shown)


iantalarico:
@chrisdunelm Thanks, will wait for @jskeet

buildrelease.sh
```diff
@@ -57,8 +57,8 @@ do
   mv ../docs/output/assembled/$project ./docs
 done
 
-rm ./index.html
-mv ../docs/output/assembled/index.html ./
+rm ./docs/index.html
```
jskeet: We probably don't actually need this line at all, given that we're just going to put a file in place in the next line anyway. Just remove the line?

iantalarico: I switched to ```mv -f``` instead of deleting the file.  This ensures we don't get prompted when we try and overwrite the file.

Issue #1937, opened 2018-02-28
Provide a logging hook to add custom labels to log entries


henkmollema:
It would be useful if the diagnostics library provided a hook to add custom labels to log entries. Application Insights, for example, has a concept of [telemetry initializers](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-api-filtering-sampling#add-properties-itelemetryinitializer) which allows you to add additional properties to telemetry.

Currently you can use `LoggerOptions.Labels` to provide custom labels:
https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/d0c81fd0756e69b664610b4438f4240f42d0f005/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs#L42

However these labels are initialized once when configuring the logger. It would be nice if the `GoogleLogger` accepted a set of initializers to further populate these labels.

iantalarico:
@henkmollema I'm assigning this to myself for now as I'd be the one to look into this more.  I'm a little low on bandwidth so this probably wouldn't happen for some time.  However, if you had a suggested design followed by PR I could manage that right now :)

henkmollema:
I'll see if I can put something together :smile:

henkmollema:
@iantalarico I'm thinking about a mechanism where as a user you can define one or more "log entry label providers" (e.g. `ILogEntryLabelProvider`? naming is hard) which are invoked in `GoogleLogger.Log(...)`:

https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/94161eb1a9230c71b94c6b1ab4cfc4cc0f53bc54/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs#L86

These providers could be resolved using the container in `IHttpContextAccessor` which is already available in the class, even though the use of the static `HttpContextAccessorWrapper` is a little bit tricky IMO.

The interface for a provider would look like this:

```cs
public interface ILogEntryLabelProvider
{
    void Invoke(Dictionary<string, string> labels);
}
```

I'll put up a design PR shortly. What are your thoughts about this?

iantalarico:
@henkmollema 

I think the general approach is okay (and yes naming is hard but that name looks good).

[...]

PR #1948, opened 2018-03-02
Add convenience extension method on WebHost builder

henkmollema:
An initial take on refactoring to use `IServiceProvider` rather than using the static `HttpContextAccessorWrapper`. 

This introduces breaking changes the the public APIs of `ILoggerFactory.AddGoogle()` and `GoogleLoggerProvider.Create()`, effectively making it impossible to configure logging without an `IServiceProvider` instance at hand. 

Most users will either:
- Call `AddGoogle` on an injected `ILoggerFactory` instance in `Startup` class. E.g.:
```cs
public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
{
    loggerFactory.AddGoogle();
}
```
This usage can easily be fixed by passing `app.ApplicationServices` to `AddGoogle`.

- Use `GoogleLoggerProvider.Create` in `ConfigureLogging` when configuring the web host builder. (ASP.NET Core 2.x) E.g.:
```cs
WebHost.CreateDefaultBuilder(args)
    .ConfigureLogging((hostContext, builder) =>
    {
        builder.AddProvider(GoogleLoggerProvider.Create());
    })

```
~This will impose problems as you're unable to pass the `IServiceProvider` instance at this point, as `ConfigureLogging` actually participates in composing the `IServiceCollection`.~ This could be changed to:
```cs
builder.Services.AddSingleton<ILoggerProvider>(sp =>
    GoogleLoggerProvider.Create(sp, projectId));
```

Besides that, I propose to create a new extension method on the `IWebHostBuilder` interface which configures logging. As I suggested in https://github.com/GoogleCloudPlatform/google-cloud-dotnet/issues/1937#issuecomment-369857071 it could configure: 
- Google logger provider (using a startup filter with `ILoggerFactory`)
- Google tracing and exception logging services
 - Google tracing and exception logging middleware

As a last note: the `IServiceProvider` instance could be optional (I've already implemented its usage as if it was optional) which keeps the breaking changes to a minimal. However, this probably would cause confusion as some features won't work due to the absence of an `IServiceProvider`.

I'm curious about your thoughts about this and how to go forward with it.

/cc @iantalarico @jskeet


iantalarico:
@henkmollema Hey Henk, thanks for the PR.  Just an FYI I probably won't get to this until Monday.

henkmollema:
@iantalarico no worries. I might start working on a sample with a startup filter/webhost builder.

henkmollema:
@iantalarico FYI I've worked further on this PR in [this branch](https://github.com/henkmollema/google-cloud-dotnet/compare/logging-serviceprovider...henkmollema:webhost-builder) to provide an extension method on `IWebHostBuilder` in conjunction with an `IStartupFilter`. It seems to work out nicely but there are some issues regarding configuration of options for the various diagnostics components. We could, however, provide defaults for those options and guide users to use the specific, existing extension methods if they want more control.

henkmollema:
I've been thinking about this and came to the conclusion that my [second point](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/1948#issue-172587809) isn't completely true. As of ASP.NET Core 2.0 it's possible to register `ILoggerProvider` instance which will get resolved from DI by the default `LoggerFactory` implementation. For example:

https://github.com/aspnet/Logging/blob/c85eb0fb4f080b804db5cd5f24de3cf1f20176e0/src/Microsoft.Extensions.Logging.Console/ConsoleLoggerFactoryExtensions.cs#L24

We could leverage this feature by registering a `GoogleLoggerProvider` in the service collection. This eliminates the trickery with a startup filter and logger factory and is "the right way" to add logging providers. However, this is only supported in ASP.NET Core >2.0.

Any thoughts on this @iantalarico?

iantalarico:
@henkmollema Sorry for the slow turn around.  I really appreciate the in-depth investigation! 

I would much rather "do the right thing" for 2.0 even if we can't do it the same way for 1.x.  I would still like to make sure we have a decent solution for 1.x but the presented 2.0 solution looks good to me.

henkmollema:
@iantalarico I've put something together with #if-defs. I'm not sure I like it:

https://github.com/henkmollema/google-cloud-dotnet/blob/fc2c7484ef25dce38d4d5eb828069318d04efa9f/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsWebHostBuilderExtensions.cs#L37-L50

We would still need [a startup filter](https://github.com/henkmollema/google-cloud-dotnet/blob/fc2c7484ef25dce38d4d5eb828069318d04efa9f/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsStartupFilter.cs) to configure diagnostics middleware, just without the logger factory. What do you think?

iantalarico:
@henkmollema  I'm not a big fan of tons of ifdefs but given how this project is set up this is probably the easiest solution (unless we split the library).

In general I'm good with this approach.

I'm a little confused by the below. Why do we need it w/o the logger factory?

> We would still need a startup filter to configure diagnostics middleware, just without the logger factory.

henkmollema:
@iantalarico I thought about splitting it up as well, I'm not sure about the maintenance implications of that. I guess both ways have their pros and cons.

What I was trying to say is that we still utilize a start up filter in the '.NET Core 2.0 way' to configure the middleware. See: https://github.com/henkmollema/google-cloud-dotnet/blob/fc2c7484ef25dce38d4d5eb828069318d04efa9f/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsStartupFilter.cs.

We're jumping in hoops now using ifdefs only to configure loggers using `ILoggerProvider` rather than using `ILoggerFactory` while effectively both ways result in the same.

Perhaps [the initial implementation](https://github.com/henkmollema/google-cloud-dotnet/commit/b83ad83eec73b2f91e5e4e15df66357060ad2aa7#diff-2062c005a077c4c55b393ee872db2960) using `ILoggerFactory` is cleaner as long as this package targets both .NET Core 1.x and 2.0?

iantalarico:
@henkmollema I think splitting it up would be a pain at this point.  I think enough of it is similar for now that ifdefs are the way to go for now.

I do think I like the first implementation better.  I think it's cleaner and solves the majority of what we want to solve.

henkmollema:
@iantalarico I've updated the PR with the `UseGoogleDiagnostics` extension method on `IWebHostBuilder`. Please let me know what you think 😄

iantalarico:
@henkmollema Thanks your continued work on this!  The more I look at this solution the more I like it.

@jskeet Any thoughts about this approach?

henkmollema:
@iantalarico @jskeet any update on this? 😄

jskeet:
I haven't had a chance to look at this in detail. Will try to get to it this afternoon. (I'd missed that Ian was asking for my review of it.)

iantalarico:
@henkmollema Sorry I lost track of this a little.  I'm good with this.

@jskeet No need to take a super detailed look, just hoping I could get a sanity check here.

jskeet:
Okay, so I've focused on the sample code snippet, and that looks clean to me.
Does this achieve what we wanted to in terms of allowing errors that occur early within the app lifecycle to be logged? If so, I'm happy and I'll leave Ian to check the detail.

henkmollema:
The current code will, if I'm correct, start logging as soon as the startup filter has been invoked, which is pretty late in terms of the startup sequence. If you'd like I could fiddle around a little and see what behavior that will cause and if it's possible to add the logger earlier in the sequence.

Also note that the Error Reporting component is only triggered in middleware. This means that startup errors will not be reported, only logged.

iantalarico:
@jskeet Thanks for taking a look.  Yes this does achieve what we wanted.

@henkmollema Thank you for moving this along.  If you are good with the solution just let me know when you are ready for a code review and I'll TAL.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsStartupFilter.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <param name="serviceProvider">
+        /// The <see cref="IServiceProvider"/> to pass to resolve additional services.
+        /// </param>
+        /// <param name="projectId">
+        /// The Google Cloud Platform project ID. If unspecified and running on GAE or GCE
+        /// the project ID will be detected from the platform.
+        /// </param>
+        /// <param name="loggerOptions">Optional, options for the logger.</param>
+        /// <param name="loggingServiceClient">Optional, logging client.</param>
+        public GoogleDiagnosticsStartupFilter(
```
henkmollema: I've added a bunch of optional parameters to specify options for several components.

Do you think they are necessary? We could also say that if you need more control you have to resort to calling the individual methods (the current way). Personally these options don't bother me for most apps, I just want them to have good defaults.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsWebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,73 @@
...
+                    var loggerFactory = sp.GetServiceCheckNotNull<ILoggerFactory>();
+                    return new GoogleDiagnosticsStartupFilter(loggerFactory, sp, projectId, loggerOptions, loggingServiceClient);
+                });
+
+                services.AddGoogleTrace(traceOptionsSetup ?? (options => options.ProjectId = projectId));
+                services.AddGoogleExceptionLogging(errorReportingOptionsSetp ?? (options =>
+                {
+                    options.ProjectId = projectId;
+
+                    var entryAssembly = Assembly.GetEntryAssembly();
```
henkmollema: Speaking of 'good defaults'. What do you think of these? Is there any convention about what to put in these values? Right now I've configured what I usually use in my projects.

henkmollema:
@iantalarico I've left two comments with some questions. Would you like to take a look?

Beside that, I think this PR is mostly done. It has become a combination of two things now:
- Refactor to use `IServiceProvider` rather than a static HttpContext accessor
- Convenience extension method on WebHost builder

Is it okay to keep them together in this PR, or do you want me to split them up? They're separate commits so it's no big deal to split.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsStartupFilter.cs
iantalarico: The more I think about it the more I prefer this option (but I think we should keep project id here).  This seems reasonable as most people (ideally) won't need to change the defaults.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsWebHostBuilderExtensions.cs
iantalarico: I'm not entirely sure here @jskeet any thoughts here?

We already do some of this in the default options that default to the GAE Flex values when available.  If we do go this route I think we should move this logic into the options class.

jskeet:
I can't see the original question that prompted a comment of "I'm not entirely sure here @jskeet any thoughts here?" on the call to `Assembly.GetEntryAssembly()`. I'd say the informational version number does provide information, but I wouldn't expect it to necessarily correlate with anything to do with the service itself.

henkmollema:
You can see those comments if you scroll a bit higher to my 'review'. Anyhow, the question was:
> Speaking of 'good defaults'. What do you think of these? Is there any convention about what to put in these values? Right now I've configured what I usually use in my projects.

I'm not sure if I understand what you mean Jon, the docs of the `Version` property says this:
> Represents the source code version that the developer provided.
It seems to me that the assembly info version represents the source code version?

jskeet:
Nope, I still can't see that question. I think that's GitHub being a little odd; not sure why.

Using the assembly version seems reasonable if we have nothing else. A commit hash might be even better, but we may not have that. I'm not intimately familiar with how ErrorReportingServiceOptions.Version is actually used - Ian, do you have any additional light to shed?

We might want to add fallbacks to other assembly versions though - see [this code](https://github.com/googleapis/gax-dotnet/blob/3081ae98756e64a57ff044fe612a4af2979b95bf/src/Google.Api.Gax/VersionHeaderBuilder.cs#L69) for an example.

henkmollema:
Hmm, probably because I pushed again after I've left those comments. Weird.

I think adding the commit hash would be difficult. Perhaps we can populate the values with the service name and version from GAE? And a fallback with the assembly info version (with fallbacks as well 😛) when the application is not running on GAE.

henkmollema:
By the way, it looks that I've broke the AppVeyor build after adding a conditional `<ItemGroup>`'s to the .csproj file. E.g.:
```xml
<ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
```

jskeet:
For the AppVeyor breakage, I think that's just a matter of removing all of those dependencies from `apis/apis.json` so that they don't get autogenerated.

jskeet:
I think the subtleties of what the version should be will be for @iantalarico to think about - and we should make the docs more specific. I'm not sure that having a single value which is sometimes a service version and sometimes an assembly version is a good idea. I *suspect* the service version is already populated elsewhere, but I'm not sufficiently familiar with this code to know. I'm hoping Ian does :)

henkmollema:
Yeah, I think multiple service names/versions are confusing too. Looking at an application that currently uses error reporting, it actually do seems that it sometimes log with the service name/version specified in the application and sometimes with the GAE service/version. I'm not sure what causes this. For example:
image
This is the same application, the first entry shows the assembly name/version. The second entry the GAE service name/version.

So it seems that somewhere these values are defaulted. Any idea @iantalarico?

iantalarico:
@henkmollema @jskeet Thanks for the input.

The service and version are fairly open ended but have fairly specific values for GAE. (https://cloud.google.com/error-reporting/reference/rest/v1beta1/ServiceContext).

I think given how general it is if we aren't on GAE we should require the user set these.  We should grab these by default for GAE but require the user set it otherwise.

henkmollema:
> I think given how general it is if we aren't on GAE we should require the user set these. We should grab these by default for GAE but require the user set it otherwise.

I don't think I fully understand what you mean. Do you only want to supply default values for applications running on GAE and not for those that aren't? If so, _which_ default values would you like to use?

iantalarico:
Sorry I should have been more clear.

Yes I only want to have defaults when we can ensure they are the correct thing.  We can grab these from the metadata server like we do with the project id.  You don't need to add these here.  I can do this in a follow up PR (unless you have a strong desire to do it here :) )

henkmollema:
I see, that makes sense. Though we need the platform's default values pretty early as they're required to configure exception logger services. E.g.:
```
{
   // You need the values here
};
```

iantalarico:
Yea I should have thought of this sooner.  We may need to require this along with the project Id.  Maybe:
```
UseGoogleDiagnostics(this IWebHostBuilder builder, string projectId = null, string service = null, string version = null)
```

henkmollema:
So what happens if someone doesn't specify these values? `AddGoogleExceptionLogging()` will blow up.

iantalarico:
@henkmollema Yes we will throw and exception in this case.

henkmollema:
I see, but would we still supply default values when running on GaE?

iantalarico:
@henkmollema Yes, I think we can wire it in similar to [how we do project ids](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs#L114)

We will pass it down and fail if its null and we can't grab a default. Does that sound reasonable?

henkmollema:
Yeah that sounds good! Sorry for me being slow-witted haha 😄 

Is grabbing the service name/version as trivial as retrieving an already existing key like this or does it require more work?
```monitoredResource.Labels.TryGetValue("project_id", out resourceProjectId)```

Anyhow, I will add the parameters to the method and start working on the integration test.

**Comment by iantalarico** [comment]
No no, it's a bit messy where everything finally gets resolved :) I should make things more explicit.

Yea it's the same call and would be with ```service``` and ```version```.

Awesome, thanks!

**Comment by henkmollema** [comment]
Great. Do you have something in mind and want to add them yourself or should I just add them while I'm at it. 

If so, do you think its okay to resolve the service/name in `UseGoogleDiagnostics`? We could add methods to the `Project` class, something like `GetServiceName` and `GetServiceVersion`. Or add them to some new class like `AppEngine`, which might be more appropriate. And is it okay to call `MonitoredResourceBuilder.FromPlatform()` multiple times, as it might take a while.

**Comment by henkmollema** [comment]
I've put something together, the `ref` feels rather dirty though. Another option might be a static field used for caching, but I'm not sure which implications that has. Please let me know what you think.

**Comment by iantalarico** [comment]
@henkmollema It feels a little dirty to me too.  

I think having a static field would work here as along as we always have the option of passing in a MonitoredResource.  What do you think?

**Comment by henkmollema** [comment]
You're right. I forgot about passing the resource. I will look into it tomorrow.

**Comment by henkmollema** [comment]
@iantalarico updated. Added some tests for `Project` as well.

**Comment by henkmollema** [comment]
@iantalarico are you sure the label names are `service` and `version`? Looking at the logs they seem to be `module_id` and `version_id`.

**Comment by iantalarico** [comment]
You are right for a GAE app it is ```module_id``` and ```version_id``` which correlate to ```service``` and ```version``` for error reporting (some naming changes happened awhile back). 

https://cloud.google.com/monitoring/api/resources#tag_gae_app

**Comment by henkmollema** [comment]
So should I use `module_id` and `version_id` for the monitored resource labels?

**Comment by iantalarico** [comment]
Yea, blah naming is a little messed up here.  We will get the `module_id` and `version_id` from the `MonitoredResource` and will pass them to the error reporting bits as service and version.

**Comment by henkmollema** [comment]
I see, I've changed them and it seems to work! 😄

**Comment by iantalarico** [comment]
@henkmollema Yay!  Ready for me to take another look?

**Comment by henkmollema** [comment]
Just a sec, I have a question: I was trying to create an integration test which actually calls the logging/tracing/error reporting APIs. I'm using a `MonitoredResource` instance for this, just as the other tests do. To make this work I had to add an optional `MonitoredResource` parameter to `UseGoogleDiagnostics` and pass it further down the stack. Even though this works, it might clutter the public API a bit. What do you think? Please take a look at this commit: https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/1948/commits/7714873649d1b304c2abf9aa7fbe138ebf8037c3.

**Comment by iantalarico** [comment]
@henkmollema I think to keep the API surface cleaner we can keep the public function the same and then have an internal function that accepts a MonitoredResource.  Sound good?

**Comment by henkmollema** [comment]
I think that would work, thanks!

**Comment by henkmollema** [comment]
@iantalarico I've updated the PR. Could you please take a look?

File: apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs [hunk 3]
```diff
@@ -81,35 +81,33 @@ public void ReadFile(IExceptionLogger exceptionLogger)
...
             string projectId = "[Google Cloud Platform project ID]";
-            loggerFactory.AddGoogle(projectId);
+            loggerFactory.AddGoogle(app.ApplicationServices, projectId);
         }
         // End sample
 
-#if NETCOREAPP2_0
         private class Startup { }
-        public void Configure()
+        public static void Program_Main()
```
iantalarico: Why the change here?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
```diff
@@ -81,35 +81,33 @@ public void ReadFile(IExceptionLogger exceptionLogger)
...
         {
-            // Sample: RegisterGoogleLogger2
```
iantalarico: I'd still like to keep this example and add an additional one for UseGoogleDiagnostics

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsStartupFilter.cs
```diff
@@ -0,0 +1,63 @@
...
+        private readonly MonitoredResource _monitoredResource;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="GoogleDiagnosticsStartupFilter"/> class to configure Google Diagnostics services.
+        /// </summary>
+        /// <param name="projectId">
+        /// The Google Cloud Platform project ID. If unspecified and running on GAE or GCE
+        /// the project ID will be detected from the platform.
+        /// </param>
+        /// <param name="monitoredResource"></param>
```
iantalarico: nit: empty param

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsWebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Google.Cloud.Diagnostics.AspNetCore
+{
+    /// <summary>
+    /// Extensions to configure Google Diagnostics on the <see cref="IWebHostBuilder"/>.
+    /// </summary>
+    public static class GoogleDiagnosticsWebHostBuilderExtensions
+    {
+        /// <summary>
+        /// Configures Google Diagostics services
+        /// such as Logging, Tracing and Error Reporting middleware.
```
iantalarico: nit: drop 'such as' I think it will be more clear that this is exactly what is done.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -30,7 +31,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore
     ///  and metadata will be sent to the Stackdriver Trace API.  Also allows for more
     ///  finely grained manual tracing.
     /// </summary>
-    /// 
+    ///
```
iantalarico: nit: remove added space

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ProjectTest.cs
```diff
@@ -49,8 +49,105 @@ public void GetAndCheckProjectId()
...
+
+            // The string service name is not null and the MonitoredResource does not contain a service name
+            // so the string service name is returned.
+            Assert.Equal(serviceName, Project.GetAndCheckServiceName(serviceName, new MonitoredResource()));
+
+            // The string service name is null and the MonitoredResource does contains a service name
+            // so the service name from the MonitoredResource is returned.
+            Assert.Equal(resourceServiceName, Project.GetAndCheckServiceName(null, monitoredResource));
+
+            // The stringservice name is null and the MonitoredResource does not contain a service name
```
iantalarico: nit: string service

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ProjectTest.cs
```diff
@@ -49,8 +49,105 @@ public void GetAndCheckProjectId()
...
+
+            // The string service version is not null and the MonitoredResource does not contain a service version
+            // so the string service version is returned.
+            Assert.Equal(serviceVersion, Project.GetAndCheckServiceVersion(serviceVersion, new MonitoredResource()));
+
+            // The string service version is null and the MonitoredResource does contains a service version
+            // so the service version from the MonitoredResource is returned.
+            Assert.Equal(resourceServiceVersion, Project.GetAndCheckServiceVersion(null, monitoredResource));
+
+            // The stringservice version is null and the MonitoredResource does not contain a service version
```
iantalarico: nit: string service

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
```diff
@@ -81,35 +81,33 @@ public void ReadFile(IExceptionLogger exceptionLogger)
...
+            // Sample: UseGoogleDiagnostics
```
iantalarico: Adding a new sample will also require updated our [index.md](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Diagnostics.AspNetCore/docs/index.md) file

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
henkmollema: The reason was that `Configure` implies that the code is located in the startup class. Anyhow, I could revert it.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
henkmollema: Would you like a sample where `null` for `IServiceProvider` is passed? Eg.:

```cs
var webHost = new WebHostBuilder()
    .ConfigureLogging((hostingContext, logging) =>
        logging.AddProvider(GoogleLoggerProvider.Create(null, projectId)));
```

We could change it to this though:

```cs
var webHost = new WebHostBuilder()
    .ConfigureServices(services =>
    {
        services.AddSingleton<ILoggerProvider>(sp => GoogleLoggerProvider.Create(sp, projectId));
    });
```

However that would only work in ASP.NET Core 2.0 from higher.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsStartupFilter.cs
henkmollema: Will fix

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsWebHostBuilderExtensions.cs
henkmollema: Will fix

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ProjectTest.cs
henkmollema: will fix

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ProjectTest.cs
henkmollema: will fix

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
henkmollema: How can  I do this?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
iantalarico: Fair enough.  Just more curious than anything. The users will won't see this, they will only see things between the `// Sample: ` bits.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
iantalarico: Agreed we should not be showing them passing in null :)

I'm good with the second option, this sample was just showing uses the 2.0 way of doing this so this is okay with me.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
iantalarico: Here is an [example](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/commit/b107f866fc148c036b4aaca50b0b939e0eb727a8#diff-92526e1df9484aa96d1d78717da6d068) of updating the file.  It's not the most straight froward of how this all works so feel free to ping me if you need a hand here.

henkmollema:
@iantalarico any reason why the code coverage blows up?

jskeet:
Code coverage is basically iffy for pull requests - at some point I need to remove it, so it's only computed on the full merged build.

henkmollema:
Alright, thanks. Nothing I should worry about?

jskeet:
Indeed :)

henkmollema:
@iantalarico updated! :)

iantalarico:
@henkmollema Looks good.  Only thing is it looks like a couple of the unit tests are failing now.

henkmollema:
I think I've fixed them, I messed up the rebase 😅

iantalarico:
@henkmollema No worries just ping me when it's ready!

henkmollema:
I think it's all set! I'll get started on some follow-up PRs which build on top of this after it's merged. 

Will this be released in 2.1 or in 3.0 due to breaking changes?

iantalarico:
@henkmollema This will be released in 3.0 (I have made a few other changes that will be put in).

Once #1937 is all finished up I think we will be good for a 3.0 release :)

henkmollema:
Great! Thanks for your help getting this in!

PR #1950, opened 2018-03-02
[Bigtable] Use separate RetrySettings for MutateRows

### LINKED ISSUE #1930: Determine when retry settings aren't generated for streaming call
Despite having idempotent retry settings [in the GAPIC config](https://github.com/googleapis/googleapis/blob/399ee9300491ddf376bf82dd8bc9c776f1d19840/google/bigtable/v2/bigtable_gapic.yaml#L104), the generator [does not generate the RetrySettings for BigtableClient's MutateRows](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/80a56f78d90be847cd5e286275889da43c0a6477/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClient.cs#L186), so they have been manually written to [overwrite the default settings](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/80a56f78d90be847cd5e286275889da43c0a6477/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs#L33). Determine if this is a bug or intentional.

If the OnConstruction partial method is still needed for the client settings after this, update the generator to support this (as mocked [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/1924/commits/c235acd38f49ad3c9ec50c2262b819627f13e908)) and determine if that is the right way to specify the default retry settings.

cc @chrisdunelm

evildour:
Fixes #1930 and removes the proposed changes to the auto-generated settings

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
@@ -1022,13 +1022,10 @@ namespace Google.Cloud.Bigtable.V2
             CallSettings callSettings,
             BigtableServiceApiClient.MutateRowsStream result)
         {
-            var defaultCallSettings = (UnderlyingClientSettings ?? new BigtableServiceApiSettings()).MutateRowSettings;
-            var effectiveCallSettings = defaultCallSettings.MergedWith(callSettings);
-            var retrySettings = effectiveCallSettings?.Timing?.Retry;
-            if (retrySettings != null)
-            {
-                effectiveCallSettings = effectiveCallSettings.WithCallTiming(CallTiming.FromExpiration(retrySettings.TotalExpiration));
-            }
+            var defaultSettings = UnderlyingClientSettings ?? new BigtableServiceApiSettings();
+            var effectiveCallSettings = defaultSettings.MutateRowsSettings.MergedWith(callSettings);
+            // TODO(mdour): Do we want to try to support per-call retry settings?
+            var effectiveRetrySettings = defaultSettings.MutateRowsRetrySettings;
 
             // TODO: Should the request manager be using the retry filter? We would need to fabricate RpcExceptions
             StatusCode[] retryStatuses = { StatusCode.DeadlineExceeded, StatusCode.Unavailable };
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1022,21 +1022,18 @@ private static MutateRowsRequest CreateMutateRowRequest(TableName tableName, IEn
...
+            var effectiveRetrySettings = defaultSettings.MutateRowsRetrySettings;
 
             // TODO: Should the request manager be using the retry filter? We would need to fabricate RpcExceptions
             StatusCode[] retryStatuses = { StatusCode.DeadlineExceeded, StatusCode.Unavailable };
             var requestManager = new BigtableMutateRowsRequestManager(retryStatuses, request);
 
             var currentStream = result;
             var status = await ProcessCurrentStream().ConfigureAwait(false);
-            if (retrySettings != null && status == ProcessingStatus.Retryable)
+            if (effectiveRetrySettings != null && status == ProcessingStatus.Retryable)
```
jskeet: Check that the settings are actually retrying ones? Or does RetryOperationUntilCompleted handle single deadline settings anyway?

evildour: These are `RetrySettings`, not `CallSettings`. It is possible to have a `RetrySettings` instance that doesn't retry?

evildour:
Yes, my only concern right now is about callers wanting to specify per-call retry settings. Do we add another optional parameter to all `MutateRows` overloads? I've left a TODO for now, but I think that might be the best way.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
jskeet: Whoops, hadn't spotted that - sorry.

jskeet:
I see what you mean - two different levels of retrying. Hmm. It feels like overkill to some extent, but it's hard to put in later if we don't do it now...

evildour:
Well as Chris said in #1930, "In the general case, we cannot do retry when server-streaming", so even though the per-call `CallSettings` could have a `RetrySettings`, they will be ignored, so it is really just one level of retrying.

PR #1952, opened 2018-03-05
[Bigtable] Use Moq for MutateRows retry tests

evildour:


jskeet:
Giving a talk this afternoon - will look when I can, hopefully before tomorrow.

evildour:
Ok sounds good. Thanks.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MockBigtableClient.cs
```diff
@@ -45,9 +31,9 @@ public class MockBigtableClient : BigtableClient
         public override Task<MutateRowResponse> MutateRowAsync(MutateRowRequest request, CallSettings callSettings = null) =>
             throw new RequestMadeException();
         public override MutateRowsResponse MutateRows(MutateRowsRequest request, CallSettings callSettings = null) =>
```
jskeet: If you don't override these methods at all, you'll get a NotImplementedException, won't you? Can't you just rely on that for anything you don't want to handle?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1002,22 +1002,38 @@ private static MutateRowsRequest CreateMutateRowRequest(TableName tableName, IEn
...
 
-        private BigtableServiceApiSettings UnderlyingClientSettings { get; set; }
+        private protected BigtableServiceApiSettings UnderlyingClientSettings { get; private set; }
+    }
 
-        /// <summary>
-        /// Gets an underlying API client that can be used to make requests.
-        /// </summary>
-        protected abstract BigtableServiceApiClient GetUnderlyingClient();
+    public partial class BigtableClientImpl
```
jskeet: Is this part of the change just merging two partials together?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MockBigtableClient.cs
evildour: Yes, I believe so, but I believe when I added this, I wanted to make sure we weren't seeing NotImplementedException for some other reason that wasn't due to the request being made. However, I guess that's pretty unlikely, so maybe it's safe to rip this out.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: No, nothing got merged. It's just the diff algorithm not realizing which things join up together. Also, I moved the ConvertResult methods from BigtableClient to BigtableClientImpl, so that probably threw it off.

PR #1953, opened 2018-03-05
Channel pool

AVaksman:
Added ClientCreationSettings to BigtableClient
Implemented multiple channel capability in BigtableClientPartial


evildour:
Thanks Alex, is this ready for review? I'll try to get to it today but might not be until tomorrow. Also, it looks like you need to rebase. I think it conflicts with a change I just made in #1952.

PR #1954, opened 2018-03-06
Multi channel

AVaksman:
MultiChannel Support

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutateRowsStreamUtilities.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutateRowsStreamUtilities.cs
@@ -46,7 +46,7 @@ namespace Google.Cloud.Bigtable.V2.Tests
             mock.Setup(c => c.MutateRows(initialRequest, It.IsAny<CallSettings>()))
                 .Returns(StreamFromEntries(entriesForInitialResponse));
 
-            return new BigtableClientImpl(null, mock.Object);
+            return new BigtableClientImpl(new[] { mock.Object }, appProfileId: null);
 
             MockMutateRowsStream StreamFromEntries(MutateRowsResponse.Types.Entry[] entries) =>
                 new MockMutateRowsStream(new MutateRowsResponse { Entries = { entries } });
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutateRowsStreamUtilities.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutateRowsStreamUtilities.cs
@@ -46,7 +46,7 @@ namespace Google.Cloud.Bigtable.V2.Tests
             mock.Setup(c => c.MutateRows(initialRequest, It.IsAny<CallSettings>()))
                 .Returns(StreamFromEntries(entriesForInitialResponse));
 
-            return new BigtableClientImpl(mock.Object, appProfileId: null);
+            return new BigtableClientImpl(null, mock.Object);
 
             MockMutateRowsStream StreamFromEntries(MutateRowsResponse.Types.Entry[] entries) =>
                 new MockMutateRowsStream(new MutateRowsResponse { Entries = { entries } });
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
@@ -247,6 +247,63 @@ namespace Google.Cloud.Bigtable.V2
     }
 
     /// <summary>
+    /// Settings for creating <see cref="BigtableServiceApiClient"/>s.
+    /// </summary>
+    public sealed class ClientCreationSettings
+    {
+        /// <summary>
+        /// Instantiate with the specified settings.
+        /// </summary>
+        /// <param name="clientCount">Optional.
+        /// The number of <see cref="BigtableServiceApiClient"/>s to create and use within a <see cref="BigtableClient"/> instance.</param>
+        /// <param name="bigtableServiceApiSettings">Optional. The settings to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
+        /// <param name="credentials">Optional. Credentials to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
+        /// <param name="serviceEndpoint">Optional.
+        /// The <see cref="ServiceEndpoint"/> to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
+        public ClientCreationSettings(
+            int? clientCount = null,
+            BigtableServiceApiSettings bigtableServiceApiSettings = null,
+            ChannelCredentials credentials = null,
+            ServiceEndpoint serviceEndpoint = null)
+        {
+            ClientCount = clientCount;
+            BigtableServiceApiSettings = bigtableServiceApiSettings;
+            Credentials = credentials;
+            ServiceEndpoint = serviceEndpoint;
+        }
+
+        /// <summary>
+        /// The number of <see cref="BigtableServiceApiClient"/>s to create and use within a <see cref="BigtableClient"/> instance.
+        /// If <c>null</c>, defaults to the CPU count on the machine this is being executed on.
+        /// </summary>
+        public int? ClientCount { get; }
+
+        /// <summary>
+        /// The settings to use when creating <see cref="BigtableServiceApiClient"/> instances.
+        /// If <c>null</c>, defaults to <see2 cref="BigtableServiceApiSettings.GetDefault"/>.
+        /// </summary>
+        public BigtableServiceApiSettings BigtableServiceApiSettings { get; }
+
+        /// <summary>
+        /// Credentials to use when creating <see cref="BigtableServiceApiClient"/> instances.
+        /// If <c>null</c>, defaults to using the default credentials.
+        /// </summary>
+        public ChannelCredentials Credentials { get; }
+
+        /// <summary>
+        /// The <see cref="ServiceEndpoint"/> to use when creating <see cref="BigtableServiceApiClient"/> instances.
+        /// If <c>null</c>, defaults to <see cref="BigtableServiceApiClient.DefaultEndpoint"/>.
+        /// </summary>
+        public ServiceEndpoint ServiceEndpoint { get; }
+
+        internal void Validate()
+        {
+            // Fairly arbitrary upper limit.
+            GaxPreconditions.CheckArgumentRange(ClientCount ?? 1, nameof(ClientCount), 1, 256);
+        }
+    }
+
+    /// <summary>
     /// Bigtable client wrapper, for convenient use.
     /// </summary>
     public sealed partial class BigtableClientImpl : BigtableClient
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
@@ -253,7 +253,7 @@ namespace Google.Cloud.Bigtable.V2
             /// <summary>
             /// Instantiate with the specified settings.
             /// </summary>
-            /// <param name="clientCount">Optional.
+            /// <param name="clientCount">Optional. If not set will default to number of processors.
             /// The number of <see cref="BigtableServiceApiClient"/>s to create and use within a <see cref="BigtableClient"/> instance.</param>
             /// <param name="bigtableServiceApiSettings">Optional. The settings to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
             /// <param name="credentials">Optional. Credentials to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
@@ -297,13 +297,12 @@ namespace Google.Cloud.Bigtable.V2
 
             /// <summary>
             /// Creates <see cref="ClientCreationSettings"/> to be used in creation of <see cref="BigtableClient"/> for use with Emulator.
-            /// <param name="environmentVariable"> is the environment variable that holds host endpoint values.</param>
+            /// <param name="target"> Service enpoint target in the yourhost:yourport form.</param>
             /// </summary>
-            public ClientCreationSettings FromEnvironmentVariable(string environmentVariable)
+            internal static ClientCreationSettings FromEndpointTarget(string target)
             {
-                string endpointString = Environment.GetEnvironmentVariable(environmentVariable);
-                string[] endpointValues = endpointString?.Split(':');
-                return new ClientCreationSettings(null, null, ChannelCredentials.Insecure, new ServiceEndpoint(endpointValues?[0], Convert.ToInt32(endpointValues?[1])));
+                string[] endpointValues = target.Split(':');
+                return new ClientCreationSettings(null, null, ChannelCredentials.Insecure, new ServiceEndpoint(endpointValues[0], Convert.ToInt32(endpointValues[1])));
             }
 
             internal void Validate()
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
@@ -244,62 +244,73 @@ namespace Google.Cloud.Bigtable.V2
         {
             throw new NotImplementedException();
         }
-    }
 
-    /// <summary>
-    /// Settings for creating <see cref="BigtableServiceApiClient"/>s.
-    /// </summary>
-    public sealed class ClientCreationSettings
-    {
         /// <summary>
-        /// Instantiate with the specified settings.
+        /// Settings for creating <see cref="BigtableServiceApiClient"/>s.
         /// </summary>
-        /// <param name="clientCount">Optional.
-        /// The number of <see cref="BigtableServiceApiClient"/>s to create and use within a <see cref="BigtableClient"/> instance.</param>
-        /// <param name="bigtableServiceApiSettings">Optional. The settings to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
-        /// <param name="credentials">Optional. Credentials to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
-        /// <param name="serviceEndpoint">Optional.
-        /// The <see cref="ServiceEndpoint"/> to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
-        public ClientCreationSettings(
-            int? clientCount = null,
-            BigtableServiceApiSettings bigtableServiceApiSettings = null,
-            ChannelCredentials credentials = null,
-            ServiceEndpoint serviceEndpoint = null)
+        public sealed class ClientCreationSettings
         {
-            ClientCount = clientCount;
-            BigtableServiceApiSettings = bigtableServiceApiSettings;
-            Credentials = credentials;
-            ServiceEndpoint = serviceEndpoint;
-        }
+            /// <summary>
+            /// Instantiate with the specified settings.
+            /// </summary>
+            /// <param name="clientCount">Optional.
+            /// The number of <see cref="BigtableServiceApiClient"/>s to create and use within a <see cref="BigtableClient"/> instance.</param>
+            /// <param name="bigtableServiceApiSettings">Optional. The settings to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
+            /// <param name="credentials">Optional. Credentials to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
+            /// <param name="serviceEndpoint">Optional.
+            /// The <see cref="ServiceEndpoint"/> to use when creating <see cref="BigtableServiceApiClient"/> instances.</param>
+            public ClientCreationSettings(
+                int? clientCount = null,
+                BigtableServiceApiSettings bigtableServiceApiSettings = null,
+                ChannelCredentials credentials = null,
+                ServiceEndpoint serviceEndpoint = null)
+            {
+                ClientCount = clientCount;
+                BigtableServiceApiSettings = bigtableServiceApiSettings;
+                Credentials = credentials;
+                ServiceEndpoint = serviceEndpoint;
+            }
 
-        /// <summary>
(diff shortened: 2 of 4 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
```diff
@@ -246,6 +246,63 @@ public partial class BigtableClient
         }
     }
 
+    /// <summary>
+    /// Settings for creating <see cref="BigtableServiceApiClient"/>s.
+    /// </summary>
+    public sealed class ClientCreationSettings
```
evildour: Let's make this nested within `BigtableClient` so we are consistent with the Pubsub clients.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -46,19 +48,45 @@ public abstract partial class BigtableClient
         /// "default" application profile will be used.
         /// </param>
         /// <returns>The task representing the created <see cref="BigtableClient"/>.</returns>
-        public static async Task<BigtableClient> CreateAsync(ServiceEndpoint endpoint = null, BigtableServiceApiSettings settings = null, string appProfileId = null)
+        public static async Task<BigtableClient> CreateAsync(BigtableServiceApiSettings settings = null, ClientCreationSettings clientCreationSettings = null, string appProfileId = null)
```
evildour: It doesn't look like we need the `BigtableServiceApiSettings` parameter here anymore since they can be obtained from the `ClientCreationSettings`.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -46,19 +48,45 @@ public abstract partial class BigtableClient
...
         {
-            var client = await BigtableServiceApiClient.CreateAsync(endpoint, settings).ConfigureAwait(false);
-            return new BigtableClientImpl(client, appProfileId) { UnderlyingClientSettings = settings };
+            clientCreationSettings?.Validate();
```
evildour: It might be worth trying to generalize all this with Pubsub. Not in this PR, but can you add a TODO above this so we can remember to look into it? Thanks.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -46,19 +48,45 @@ public abstract partial class BigtableClient
...
+            }
+            var clientCount = clientCreationSettings?.ClientCount ?? Environment.ProcessorCount;
+
+            var endpoint = clientCreationSettings?.ServiceEndpoint ?? BigtableServiceApiClient.DefaultEndpoint;
+            var clients = new BigtableServiceApiClient[clientCount];
+            var shutdowns = new Func<Task>[clientCount];
+            // Fill clients[] with BigtableServiceApiClient instances, each with specific channel
+            for (int i = 0; i < clientCount; i++)
+            {
+                var channel = new Channel(endpoint.Host, endpoint.Port, channelCredentials);
```
evildour: This doesn't pass the `channelOptions` like Pubsub does. Was that error with message sizes Pubsub specific?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -46,19 +48,45 @@ public abstract partial class BigtableClient
...
+                clients[i] = BigtableServiceApiClient.Create(channel, clientCreationSettings?.BigtableServiceApiSettings);
+                shutdowns[i] = channel.ShutdownAsync;
+            }
+            Func<Task> shutdown = () => Task.WhenAll(shutdowns.Select(x => x()));
+
+            return new BigtableClientImpl(appProfileId, clients) { UnderlyingClientSettings = settings };
```
evildour: `settings` here should be `clientCreationSettings?.BigtableServiceApiSettings`. Actually, I would leave this as-is, and capture `clientCreationSettings?.BigtableServiceApiSettings` in a local above the for loop, then use that local inside the loop as well.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -69,10 +97,36 @@ public abstract partial class BigtableClient
         /// "default" application profile will be used.
         /// </param>
         /// <returns>The created <see cref="BigtableClient"/>.</returns>
-        public static BigtableClient Create(ServiceEndpoint endpoint = null, BigtableServiceApiSettings settings = null, string appProfileId = null)
+        public static BigtableClient Create(BigtableServiceApiSettings settings = null, ClientCreationSettings clientCreationSettings = null, string appProfileId = null)
```
evildour: Again, `settings` not needed.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -69,10 +97,36 @@ public abstract partial class BigtableClient
...
+            // Fill clients[] with BigtableServiceApiClient instances, each with specific channel
+            for (int i = 0; i < clientCount; i++)
+            {
+                var channel = new Channel(endpoint.Host, endpoint.Port, channelCredentials);
+                clients[i] = BigtableServiceApiClient.Create(channel, clientCreationSettings?.BigtableServiceApiSettings);
+                shutdowns[i] = channel.ShutdownAsync;
+            }
+            Func<Task> shutdown = () => Task.WhenAll(shutdowns.Select(x => x()));
+
+            return new BigtableClientImpl(appProfileId, clients) { UnderlyingClientSettings = settings };
```
evildour: Rather than duplicate everything here, it would be better to make this whole method an expression-bodied member that delegates to the async version:
```csharp
=> Task.Run(() => CreateAsync(clientCreationSettings, appProfileId)).ResultWithUnwrappedExceptions();
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -93,7 +147,7 @@ public static BigtableClient Create(ServiceEndpoint endpoint = null, BigtableSer
         public static BigtableClient Create(Channel channel, BigtableServiceApiSettings settings = null, string appProfileId = null)
```
evildour: I thought we decided this overload was no longer needed.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -93,7 +147,7 @@ public static BigtableClient Create(ServiceEndpoint endpoint = null, BigtableSer
...
         {
             var client = BigtableServiceApiClient.Create(channel, settings);
-            return new BigtableClientImpl(client, appProfileId) { UnderlyingClientSettings = settings };
+            return new BigtableClientImpl(appProfileId, client) { UnderlyingClientSettings = settings };
         }
```
evildour: Pubsub has a synchronous `Create` which takes an `IEnumerable<T>` of already created clients. We should probably have something similar to be consistent. Although now that I think of it, then we wouldn't be able to set `UnderlyingClientSettings`, because they could all be different. But I think that might actually be ok. Then the usages of it would pick up the default settings.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1009,6 +1063,9 @@ public partial class BigtableClientImpl
     {
         private readonly string _appProfileId;
         private readonly BigtableServiceApiClient _client;
```
evildour: This should no longer be needed

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1009,6 +1063,9 @@ public partial class BigtableClientImpl
...
+        private readonly BigtableServiceApiClient[] _clients;
+        private int _clientNumber = 0;
+
 
         // TODO: Add a public constructor after multi-channel support?
         internal BigtableClientImpl(BigtableServiceApiClient client, string appProfileId)
```
evildour: Is this needed anymore?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1017,13 +1074,25 @@ internal BigtableClientImpl(BigtableServiceApiClient client, string appProfileId
...
         }
 
+        // TODO: Add a public constructor after multi-channel support?
+        internal BigtableClientImpl(string appProfileId, params BigtableServiceApiClient[] clients)
+        {
+            _clients = clients;
+            _appProfileId = appProfileId;
+        }
+
         // TODO: Add Multi-channel support
```
evildour: This TODO can be removed

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1017,13 +1074,25 @@ internal BigtableClientImpl(BigtableServiceApiClient client, string appProfileId
...
         /// <inheritdoc/>
-        private BigtableServiceApiClient GetUnderlyingClient() => _client;
+        private BigtableServiceApiClient GetUnderlyingClient() => _clients.Length > 1 ? _clients[GetClientIndex()] : _clients[0];
+
+        private int GetClientIndex()
+        {
+            return (Interlocked.Increment(ref _clientNumber) - 1) % _clients.Length;
```
evildour: I would initialize `_clientNumber` to `-1` and then remove the subtraction (and outer parens) here.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutateRowsStreamUtilities.cs
```diff
@@ -46,7 +46,7 @@ public static class MutateRowsStreamUtilities
             mock.Setup(c => c.MutateRows(initialRequest, It.IsAny<CallSettings>()))
                 .Returns(StreamFromEntries(entriesForInitialResponse));
 
-            return new BigtableClientImpl(mock.Object, appProfileId: null);
+            return new BigtableClientImpl(null, mock.Object);
```
evildour: If the params in the constructor was added just so we could do this, I would prefer not to have it. I would rather the self-documentation of the named argument if that means creating an array explicitly:
```csharp
return new BigtableClientImpl(new [] { mock.Object }, appProfileId: null);
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1017,13 +1074,25 @@ internal BigtableClientImpl(BigtableServiceApiClient client, string appProfileId
...
+        internal BigtableClientImpl(string appProfileId, params BigtableServiceApiClient[] clients)
+        {
+            _clients = clients;
+            _appProfileId = appProfileId;
+        }
+
         // TODO: Add Multi-channel support
         /// <inheritdoc/>
-        private BigtableServiceApiClient GetUnderlyingClient() => _client;
+        private BigtableServiceApiClient GetUnderlyingClient() => _clients.Length > 1 ? _clients[GetClientIndex()] : _clients[0];
```
evildour: I would keep things simple here and always return `_clients[GetClientIndex()]`. I don't think the atomic update is going to have a significant impact on the overall operation when this result is then used to make an API call.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
AVaksman: As of right now we need it for Integration tests to create a client to use with the emulator [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixtureBase.cs#L113)
However to make it easier to use Emulator I could add the following method to ClientCreationSettings class
```
/// <summary>
/// Creates <see cref="ClientCreationSettings"/> to be used in creation of <see cref="BigtableClient"/> for use with Emulator.
/// <param name="environmentVariable"> is the environment variable that holds host endpoint values.</param>
/// </summary>
public ClientCreationSettings FromEnvironmentVariable(string environmentVariable)
{
    string endpointString = Environment.GetEnvironmentVariable(environmentVariable);
    string[] endpointValues = endpointString?.Split(':');
    return new ClientCreationSettings(null, null, ChannelCredentials.Insecure, new ServiceEndpoint(endpointValues?[0], Convert.ToInt32(endpointValues?[1])));
}
```
And then in `BigtableFixtureBase.cs` swap
`TableClient = BigtableClient.Create(EmulatorChannel);`
with
`TableClient = BigtableClient.Create(new BigtableClient.ClientCreationSettings().FromEnvironmentVariable(EmulatorEnvironmentVariable));`

AVaksman: This will simplify use of the emulator. And if user wants to run some tests against different mock server (like we did) then user will still be able to pass a different custom environment variable to the above method or will have to manually create a `ClientCreationSettings` object with custom settings.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
AVaksman: I have only saw [this](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs#L260) documentation in PubSub. I did not pass the `channelOptions` when implemented `BigtableClientPool` in a test project when tested channel pooling effect on performance. I followed [these](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/docs/root/faq.md#how-can-i-use-non-default-credentials-for-grpc-based-apis) instructions to create a `BigtableClient` with a specified channel.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -97,58 +97,26 @@ public abstract partial class BigtableClient
...
-            return new BigtableClientImpl(appProfileId, clients) { UnderlyingClientSettings = settings };
-        }
+        public static BigtableClient Create(ClientCreationSettings clientCreationSettings = null, string appProfileId = null) =>
+            Task.Run(() => CreateAsync(clientCreationSettings, appProfileId)).ResultWithUnwrappedExceptions();
 
         /// <summary>
-        /// Creates a <see cref="BigtableClient"/> which uses the specified channel for remote operations.
+        /// Synchronously creates a <see cref="BigtableClient"/>, applying defaults for all unspecified settings,
+        /// and creating a channel connecting to the given endpoint with application default credentials where
+        /// necessary.
```
evildour: This comment does not apply here: no settings are specified and no channels get created. It needs to be updated for this overload.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -97,58 +97,26 @@ public abstract partial class BigtableClient
...
         /// </summary>
-        /// <param name="channel">The <see cref="Channel"/> for remote operations. Must not be null.</param>
-        /// <param name="settings">Optional <see cref="BigtableServiceApiSettings"/>.</param>
+        /// <param name="clients">The <see cref="BigtableServiceApiClient"/>s to use in a <see cref="BigtableClient"/>.</param>
```
evildour: This should have an extra sentence: "Must not be null."

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -97,58 +97,26 @@ public abstract partial class BigtableClient
...
         /// This value specifies routing for replication. If not specified, the
         /// "default" application profile will be used.
         /// </param>
         /// <returns>The created <see cref="BigtableClient"/>.</returns>
-        public static BigtableClient Create(Channel channel, BigtableServiceApiSettings settings = null, string appProfileId = null)
-        {
-            var client = BigtableServiceApiClient.Create(channel, settings);
-            return new BigtableClientImpl(appProfileId, client) { UnderlyingClientSettings = settings };
-        }
+        public static BigtableClient Create(BigtableServiceApiClient[] clients, string appProfileId = null) =>
```
evildour: This should take `IEnumerable<BigtableServiceApiClient>` as the first parameter and use `ToArray()` below after validating that the parameter is non-null:
```csharp
new BigtableClientImpl(GaxPreconditions.CheckNotNull(clients, nameof(clients)).ToArray(), appProfileId);
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: @chrisdunelm why was this needed in Pubsub? Is this something we need to be concerned about for Bigtable as well?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -101,11 +101,10 @@ public abstract partial class BigtableClient
             Task.Run(() => CreateAsync(clientCreationSettings, appProfileId)).ResultWithUnwrappedExceptions();
 
         /// <summary>
-        /// Synchronously creates a <see cref="BigtableClient"/>, applying defaults for all unspecified settings,
-        /// and creating a channel connecting to the given endpoint with application default credentials where
-        /// necessary.
+        /// Synchronously creates a <see cref="BigtableClient"/>, applying default settings.
```
evildour: There are still no default settings applied. All settings are used to create the clients, which are being specified by the caller. Maybe just update this to be:
```csharp
/// Synchronously creates a <see cref="BigtableClient"/> from a set of pre-existing clients.
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
```diff
@@ -244,6 +244,73 @@ public partial class BigtableClient
...
+
+        /// <summary>
+        /// Settings for creating <see cref="BigtableServiceApiClient"/>s.
+        /// </summary>
+        public sealed class ClientCreationSettings
+        {
+            /// <summary>
+            /// Instantiate with the specified settings.
+            /// </summary>
+            /// <param name="clientCount">Optional.
```
AVaksman: Perhaps should add here
If not set will defaults to count of processors

evildour: Yeah, that's probably a good idea, although we may change that heuristic in the future, but we can change the docs at the same time too if that's the case.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
```diff
@@ -253,7 +253,7 @@ public sealed class ClientCreationSettings
             /// <summary>
             /// Instantiate with the specified settings.
             /// </summary>
-            /// <param name="clientCount">Optional.
+            /// <param name="clientCount">Optional. If not set will default to number of processors.
```
evildour: This message should go at the end of the comment, after the next sentence.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
chrisdunelm: Is this referring to the code in [here](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs#L242)?

If so, then this is because it is expected that both publish and subscriber messages will exceed the default maximum individual message size allowed client-side (~4Mb). The servre-side maximum for pubsub is ~10Mb. The client is set to -1 indicating that both sent and received messages can be of any size.

Whether these particular channel options are needed here depends on the maximum individual message size you expect to send and receive on the client.

evildour: Ok thanks Chris, I'll ask internally.

evildour:
I'll merge this while we wait for internal info about the message sizes. We can add the size override later if necessary.

AVaksman:
Thanks Mike

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -46,19 +47,46 @@ public abstract partial class BigtableClient
...
+            var clients = new BigtableServiceApiClient[clientCount];
+            var shutdowns = new Func<Task>[clientCount];
+            // Fill clients[] with BigtableServiceApiClient instances, each with specific channel
+            for (int i = 0; i < clientCount; i++)
+            {
+                var channel = new Channel(endpoint.Host, endpoint.Port, channelCredentials);
+                clients[i] = BigtableServiceApiClient.Create(channel, settings);
+                shutdowns[i] = channel.ShutdownAsync;
+            }
+            Func<Task> shutdown = () => Task.WhenAll(shutdowns.Select(x => x()));
```
evildour: @AVaksman Sorry, I missed this in the initial review, but it looks like this is unused, so the channels are never shutdown. This needs to be passed along to BigtableClientImpl so they can be shutdown when the client is shutdown/disposed.

PR #1956, opened 2018-03-06
Change sentinel value handling

jskeet:
- SentinelValue is now an internal class instead of a public enum
- FieldValue replaces the SentinelValue enum as the way for sentinel values
  to be represented in non-attributed object models (e.g. dictionaries and anonymous types)
- Attributes for sentinel values are separated from FirestorePropertyAttribute

All of this will allow for parameterized sentinel values in the future

cc @mikelehen


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestorePropertyAttribute.cs
```diff
@@ -26,13 +26,7 @@ public sealed class FirestorePropertyAttribute : Attribute
...
         /// of the property is used directly.
         /// </summary>
-        public string Name { get; set; }
-
-        /// <summary>
-        /// The sentinel value for this property, defaulting to <see cref="SentinelValue.None"/>
-        /// indicating a regular value to be serialized.
-        /// </summary>
-        public SentinelValue SentinelValue { get; set; }
+        public string Name { get; set; }
```
chrisdunelm: nit: I *think* there are some trailing spaces here

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FieldValue.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <summary>
+        /// Sentinel value indicating that the field should be set to the timestamp of the
+        /// commit that creates or modifies the document.
+        /// </summary>
+        public static object ServerTimestamp { get; } = SentinelValue.ServerTimestamp;
+
+        /// <summary>
+        /// Sentinel value indicating that the field should be deleted from the document.
+        /// </summary>
+        public static object Delete{ get; } = SentinelValue.ServerTimestamp;
```
chrisdunelm: nit: Add space after `Delete`

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/DeletedFieldAttribute.cs
```diff
@@ -0,0 +1,29 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Firestore
+{
+    // TODO: Do we really want this? More present for consistency than anything else.
```
chrisdunelm: I like the consistency of including this. But if it has *absolutely no use whatsoever* then it should probably be removed, to remove the possibility of confusion.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FieldValue.cs
evildour: SentinelValue.**Delete**?

chrisdunelm: Oops, how did I miss seeing that!

evildour: It was hiding in plain sight :)

jskeet: More importantly, how do the tests pass? Will investigate and fix, obviously. (I may be using SentinelValue in tests were I should be using FieldValue.)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/DeletedFieldAttribute.cs
jskeet: We'll see what the Firestore team says.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestorePropertyAttribute.cs
jskeet: Thanks, will remove.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/SentinelValue.cs
```diff
@@ -12,33 +12,82 @@
...
     {
-        /// <summary>
-        /// Only present to allow <see cref="FirestorePropertyAttribute"/> to indicate that
-        /// a property is not a sentinel value; this value cannot be serialized.
-        /// </summary>
-        None = 0,
+        // These sentinel values are invalid "null" values - the NullValue enum only has a single entry (0), but we're
+        // using the fact that there enums are just named numbers to hide extra information there.
+        // This is deeply unpleasant, but:
+        // - It's extensible to future later sentinel values, so long as they don't require user-specified data (which I don't think they will)
```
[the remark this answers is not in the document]
jskeet: Indeed, this was a copy/paste issue... will fix.

jskeet:
Right, fixed all the comments - will squash the change and merge into master to avoid things getting stale; any comments from @mikelehen later can be addressed in a later change.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/DeletedFieldAttribute.cs
mikelehen: Yeah, we left it out of Android / Java because it doesn't make much sense.  In general we don't expect apps to use FieldValue.Delete as part of their normal operation because their objects likely match a "schema" which has a consistent set of fields, and even for "optional" fields you generally want to store `null` rather than omit the field, so you can still query/sort on it.  And if you're using object-mapping then it's even clearer that you have a schema with a well-defined set of properties.

I can kinda' conceive of one use for it in object-mapping as a form of schema migration (if you are writing your objects with SetOptions.MergeAll and at some point decide to remove some field you used to use, you could change your object to have the DeletedField attribute on that field to delete it whenever writing objects), but I'm not sure this is realistic and I think I'd rather wait and see if any user asks for it.

So I'd be inclined to remove it for now to keep our surface area minimal and consistent with the Java SDK and revisit if we ever get user feedback asking for it.  I don't feel super strongly though if you disagree.

jskeet:
Schema migration was exactly the use case I was thinking of :)

Happy to delete it for now. Will create a new issue to remind me.

Issue #1625, opened 2017-11-06
Stop using deprecated type Hashtable


evildour:
(flagged by the [API Analyzer](https://www.nuget.org/packages/Microsoft.DotNet.Analyzers.Compatibility/))

Spanner has some uses of `Hashtable`, which is deprecated. Change these to `Dictionary<string, object>`

jskeet:
This is documented to be returned by default from `SpannerDataReader.GetFieldValue(string)` for struct fields - I suspect that means we can't fix it now. Do you agree?

chrisdunelm:
Which documentation is this?

jskeet:
https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.Spanner.Data/api/Google.Cloud.Spanner.Data.SpannerDbType.html#Google_Cloud_Spanner_Data_SpannerDbType_StructOf_System_Collections_Generic_IDictionary_System_String_Google_Cloud_Spanner_Data_SpannerDbType__

chrisdunelm:
Oh right. And it's GAed, so now can't be changed presumably. That's a (small) shame.

jskeet:
I think we can close this issue now as we've done as much as we can without breaking compatibility.

PR #1957, opened 2018-03-06
Default to using Dictionary<string, object> for struct values

jskeet:
This will only be merged if we decide to bump
Google.Cloud.Spanner.Data to v2.0.

This would be the last part of fixing #1625.
Note that it doesn't prevent users from deserializing to Hashtable,
it just changes the default.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
@@ -389,9 +389,7 @@ namespace Google.Cloud.Spanner.Data
             {
                 if (targetClrType == typeof(IDictionary))
                 {
-#pragma warning disable DE0006
-                    targetClrType = typeof(Hashtable);
-#pragma warning restore DE0006
+                    targetClrType = typeof(Dictionary<string, object>);
                 }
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
```diff
@@ -389,9 +389,7 @@ private object ConvertToClrTypeImpl(Value wireValue, System.Type targetClrType,
             {
                 if (targetClrType == typeof(IDictionary))
                 {
-#pragma warning disable DE0006
-                    targetClrType = typeof(Hashtable);
-#pragma warning restore DE0006
+                    targetClrType = typeof(Dictionary<string, object>);
```
chrisdunelm: I don't know this code well...
But what if `targetClrType` is `IDictionary<string, object>`, the `Activator.Createinstance` call below will fail.
Or is `IDictionary<,>` unsupported here?

jskeet: It looks like `IDictionary<,>` has never been supported as a target type - ditto `IList<T>`. We can add support for that later if it would be desirable (it sounds useful) without it being a breaking change.

PR #1960, opened 2018-03-07
Add option to enable GetSchemaTable returning schema information

### LINKED ISSUE #1853: Return content for GetSchemaTable() of Spanner DataReader
We have a generic implementation for ADO.net that relies on reading the GetSchemaTable(). Currently it returns null but we need it to have some data for the ColumnName, Ordinal, and DataType with the addition of NumericPrecision, NumericScale and ColumnSize even if these 3 columns would just be empty.

jskeet:
Fixes #1853.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
@@ -47,13 +47,15 @@ namespace Google.Cloud.Spanner.Data
         private ResultSetMetadata _metadata;
         private readonly SingleUseTransaction _txToClose;
         private readonly SpannerConversionOptions _conversionOptions;
+        private readonly bool _provideSchemaTable;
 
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -329,44 +332,88 @@ public override void Close()
             _connectionToClose?.Close();
         }
 
-        /// <inheritdoc />
+        /// <summary>
+        /// When enabled, returns the schema of the query as a <see cref="DataTable"/>. This feature needs
+        /// to be enabled in the connection string via the <c>EnableGetSchemaTable</c> property.
```
chrisdunelm: Change the `<c>` to a `<see>` ?

jskeet: Yup, good call.

PR #1962, opened 2018-03-07
Update Google.Cloud.Spanner.Data to 2.0.0-beta00

jskeet:
We're introducing breaking changes (semantically) so let's bump the
major version.


jskeet:
Yup. beta00 and alpha00 are special version numbers which the release tagger knows to ignore.

PR #1967, opened 2018-03-07
Update project dependencies

### LINKED ISSUE #1830: Update dependencies
We should update to the latest version of Google.Protobuf and Google.Protobuf.Tools, and potentially Grpc.Core, although we might want to hold off on the latter due to https://github.com/grpc/grpc/issues/14021.

In particular, Google.Protobuf now supports:

- Retaining unknown fields
- Better comparisons of `double` values in the face of NaN (which would allow a slight test simplification)

This needs to be done in the GAX repo as well as here. This issue is basically to avoid me forgetting to do this; it's not urgent, but it would be good to get done.

jskeet:
Fixes #1830.

Review commit by commit, ignoring the "regenerate" commits which are purely generated.


apis/apis.json
```diff
@@ -436,8 +436,8 @@
     "tags": [ "Spanner" ],
     "dependencies": {
       "Google.LongRunning": "1.0.0",
-      "Google.Api.Gax.Grpc": "2.2.1",
-      "Grpc.Core": "1.7.1",
+      "Google.Api.Gax.Grpc": "2.3.0",
+      "Grpc.Core": "1.10.0",
       "Google.Cloud.Iam.V1": "1.0.0"
```
evildour: Wouldn't all these versions need to get updated too since Google.Cloud.Iam.V1 was bumped to 1.1.0-beta00? Or is this the version pulled from NuGet?

jskeet: When versions are specified explicitly, they'll be loaded from NuGet rather than the local project. If we release an IAM 1.1.0, we can update these too.

evildour: Just curious: what criteria are used to decide whether to use a NuGet reference as opposed to a local project reference?

jskeet: If the API is GA, we need to pin all its dependencies to specific versions. In theory, everything that I've updated to a new beta00 I *could* start using the defaults/project versions again - there just didn't seem to be much benefit, as we'd probably need to pin back to the current version anyway.

We have checks that say that when we tag a release, project dependencies are only allowed if those projects are being tagged at the same time.

We've had a problem in the past where we had X depending on Y with a project dependency, Y gained a feature and X used it, then we released X without releasing Y, causing the kind of problems you'd expect...

apis/apis.json
```diff
@@ -507,30 +507,30 @@
     "id": "Google.Cloud.Spanner.V1",
     "productName": "Google Cloud Spanner",
     "productUrl": "https://cloud.google.com/spanner/",
-    "version": "1.1.0-beta01",
+    "version": "1.1.0-beta02",
```
chrisdunelm: I don't think this beta is meant to be updated yet?

apis/apis.json
evildour: Interesting, makes sense. Thanks!

apis/apis.json
jskeet: Well spotted. Will revert that one - but in a separate commit as doing anything else will be too complicated :)

chrisdunelm: SGTM :)

Issue #1966, opened 2018-03-07
Memory leak when creating many ImageAnnotatorClient instances


markwalsh-liverpool:
Unsure what the root cause of the issue is but I noticed that creating a lot of **ImageAnnotatorClient** instances causes a consistent rise in unmanaged memory (first guess is - it's creating a lot of HttpClients):

![image](https://user-images.githubusercontent.com/1950856/37102862-c44026d0-2220-11e8-8d9d-bb901b032b19.png)

I was initialising a ImageAnnotatorClient per image and I am sending 50-100 images per second so I was quite quickly running out of memory.  Now, I just reuse the same ImageAnnotatorClient for each batch of images and I now see fairly consistent memory usage (sorry for the differing pictures, the bottom graph shows the unmanaged memory - and it's basically flatlined):

![image](https://user-images.githubusercontent.com/1950856/37103041-392c485c-2221-11e8-9283-5fa8dd4f0a18.png)

I was just wondering whether:

1. It's okay to reuse the client, several hundred times?
2. Whether I've exposed a bug (maybe something isn't being GC'd?)...

Is this possibly a duplicate/symptom of [this issue](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/issues/1502) or perhaps [this issue](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/issues/1060)

These are my installed packages:

```
  <package id="Google.Api.CommonProtos" version="1.2.0" targetFramework="net46" />
  <package id="Google.Api.Gax" version="2.2.1" targetFramework="net46" />
  <package id="Google.Api.Gax.Grpc" version="2.2.1" targetFramework="net46" />
  <package id="Google.Apis" version="1.32.2" targetFramework="net46" />
  <package id="Google.Apis.Auth" version="1.32.2" targetFramework="net46" />
  <package id="Google.Apis.Core" version="1.32.2" targetFramework="net46" />
  <package id="Google.Cloud.Vision.V1" version="1.1.0" targetFramework="net46" />
  <package id="Google.Protobuf" version="3.5.1" targetFramework="net46" />
  <package id="Grpc.Auth" version="1.10.0" targetFramework="net46" />
  <package id="Grpc.Core" version="1.10.0" targetFramework="net46" />
```

jskeet:
[...]

PR #1972, opened 2018-03-08
[Bigtable] Cap multi-channel support to using 2 channels during testing.

evildour:
(each integration test was taking ~5 seconds for me instead of a few milliseconds after #1954)

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixtureBase.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/BigtableFixtureBase.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Api.Gax.Grpc;
 using Google.Cloud.Bigtable.Admin.V2;
 using Google.Cloud.ClientTesting;
 using Google.Rpc;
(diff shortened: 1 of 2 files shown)


jskeet:
I'm somewhat concerned that this is necessary - does it suggest that #1954 will bite regular users too? I didn't review that PR before, would you like me to have a look now?

evildour:
Yeah that might be a good idea. I guess I didn't realize how much overhead was involved in an initial call over a channel. With many cores, each test uses a few new channels that have never been used before and it slows down those initial tests. In a normal application, I assume this is ok for application startup, but wouldn't be a problem if the same client is used throughout (so we're probably going to need good docs around that as well).

evildour:
BTW, do we make a similar recommendation anywhere with `PublisherClient`? I imagine that would have a similar issue as it too uses multiple new channels by default.

jskeet:
@chrisdunelm would know more about PublisherClient than I can remember.

There are additional issues with using explicitly-provided channels, in that we should clean them up if possible - see #1966 for example. This whole area probably needs more thought.

Friday afternoon probably isn't the best time for me to think about this, so I'll punt it to Monday and look closely then. I'm happy for us to merge this for now if you think that's a good idea though.

chrisdunelm:
This hasn't been an issue with Pubsub integration tests. However, they are all fairly long-running tests (> 1 minute each), so a few seconds extra for channel creation wouldn't be noticeable.
Next week I may do a test that does nothing, just to see the time taken for channel setup.

evildour:
As far as cleanup goes, #1973 will eventually address that, but I need to sync up with Alex about it next week.

evildour:
Closing: we're going to go with a different approach for channel pooling, so this probably won't apply after that.

PR #1973, opened 2018-03-08
shutdown use added to channelPool

AVaksman:
Adding missed from #1954 shutdown usage to channel pool.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutateRowsStreamUtilities.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/MutateRowsStreamUtilities.cs
@@ -46,7 +46,7 @@ namespace Google.Cloud.Bigtable.V2.Tests
             mock.Setup(c => c.MutateRows(initialRequest, It.IsAny<CallSettings>()))
                 .Returns(StreamFromEntries(entriesForInitialResponse));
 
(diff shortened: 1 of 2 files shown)


evildour:
Sorry Alex, I don't think I was very clear in https://github.com/GoogleCloudPlatform/google-cloud-dotnet/pull/1954#discussion_r173224292. The problem is that this `shutdown` is a delegate instance which needs to be invoked at some point later when the client is no longer in use. It necessarily needs to be passed along to and stored within the `BigtableClientImpl`, which you've done here, so that it can live as long as the client, but it still needs to be invoked at some point.

It looks like the `PublisherClient` provides two `ShutdownAsync` methods: https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs#L353. The override of one of these: https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs#L497, ultimately invokes the stored `_shutdown` field to close the channels. We will probably want to do something similar here and expose similarly named methods (although the actual implementation may differ slightly).

Let me know if you want to sync up on VC to discuss further.

AVaksman:
Sorry, Mike. I did not communicate clearly. Yes I did see those two methods in `PublisherClient` and I wasn't sure how exactly implementation should look like here.

evildour:
Closing: we're going to go with a different approach for channel pooling, so this probably won't apply after that.

PR #1974, opened 2018-03-08
adding channelOtions to channelPool

AVaksman:
Setting both sent and received messages' caps to any size


evildour:
Thanks Alex. I'm going to merge despite the travis failure. Seems like it might be an issue with travis.

AVaksman:
Thanks

PR #1978, opened 2018-03-09
[Bigtable] ReadRows retries

evildour:
Best to review one commit at a time

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
@@ -1,0 +1,450 @@
+﻿// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Protobuf;
+using System;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Bigtable.V2.Tests
+{
+    public class RetryingReadRowsTest
+    {
+        [Fact]
+        public async Task SimpleRetryWithKeys()
+        {
+            var request = new ReadRowsRequest
+            {
+                Rows = RowSet.FromRowKeys("a", "b", "c")
+            };
+            var client = Utilities.CreateReadRowsMockClient(
+                request,
+                initialStreamResponse: new[]
+                {
+                    new ReadRowsResponse
+                    {
+                        Chunks =
+                        {
+                            CreateChunk("a", "cf1", "column1", "value1", commitRow: true)
+                        }
+                    }
+                },
+                responsesForRetryStreams: new[]
+                {
+                    new []
+                    {
+                        new ReadRowsResponse
+                        {
+                            Chunks =
+                            {
+                                CreateChunk("b", "cf1", "column2", "value2", commitRow: true)
+                            }
+                        }
+                    }
+                });
+
+            var stream = client.ReadRows(request);
+
+            int rowCount = 0;
+            await stream.ForEachAsync(row =>
+            {
+                rowCount++;
+
+                var family = row.Families[0];
+                switch (row.Key.ToStringUtf8())
+                {
+                    case "a":
+                        Assert.Equal("cf1", family.Name);
+                        Assert.Equal("column1", family.Columns[0].Qualifier.ToStringUtf8());
+                        Assert.Equal("value1", family.Columns[0].Cells[0].Value.ToStringUtf8());
+                        break;
+
+                    case "b":
+                        Assert.Equal("cf1", family.Name);
+                        Assert.Equal("column2", family.Columns[0].Qualifier.ToStringUtf8());
+                        Assert.Equal("value2", family.Columns[0].Cells[0].Value.ToStringUtf8());
+                        break;
+                }
+            });
+            Assert.Equal(2, rowCount);
+        }
+
+        [Fact]
+        public async Task SimpleRetryWithRanges()
+        {
+            var request = new ReadRowsRequest
+            {
+                Rows = RowSet.FromRowRanges(RowRange.Closed("a", "z"))
+            };
+            var client = Utilities.CreateReadRowsMockClient(
+                request,
+                initialStreamResponse: new[]
+                {
+                    new ReadRowsResponse
+                    {
+                        Chunks =
+                        {
+                            CreateChunk("a", "cf1", "column1", "value1", commitRow: true)
+                        }
+                    }
+                },
+                responsesForRetryStreams: new[]
+                {
+                    new []
+                    {
+                        new ReadRowsResponse
+                        {
+                            Chunks =
+                            {
+                                CreateChunk("b", "cf1", "column2", "value2", commitRow: true)
+                            }
+                        }
+                    }
+                });
+
+            var stream = client.ReadRows(request);
+
+            int rowCount = 0;
+            await stream.ForEachAsync(row =>
+            {
+                rowCount++;
+
+                var family = row.Families[0];
+                switch (row.Key.ToStringUtf8())
+                {
+                    case "a":
+                        Assert.Equal("cf1", family.Name);
+                        Assert.Equal("column1", family.Columns[0].Qualifier.ToStringUtf8());
+                        Assert.Equal("value1", family.Columns[0].Cells[0].Value.ToStringUtf8());
+                        break;
+
+                    case "b":
+                        Assert.Equal("cf1", family.Name);
+                        Assert.Equal("column2", family.Columns[0].Qualifier.ToStringUtf8());
+                        Assert.Equal("value2", family.Columns[0].Cells[0].Value.ToStringUtf8());
+                        break;
+                }
+            });
+            Assert.Equal(2, rowCount);
+        }
+
+        [Fact]
+        public async Task RetryWithLastScannedKey()
+        {
+            var request = new ReadRowsRequest
+            {
+                Rows = RowSet.FromRowRanges(RowRange.Closed("a", "z"))
+            };
+            var client = Utilities.CreateReadRowsMockClient(
+                request,
+                initialStreamResponse: new[]
+                {
+                    new ReadRowsResponse
+                    {
+                        Chunks =
+                        {
+                            CreateChunk("a", "cf1", "column1", "value1", commitRow: true)
+                        }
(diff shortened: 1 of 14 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -55,6 +54,42 @@ public partial class BigtableServiceApiSettings
                 totalExpiration: Expiration.FromTimeout(TimeSpan.FromMilliseconds(600000)),
                 retryFilter: IdempotentRetryFilter
             );
+
+        /// <summary>
+        /// <see cref="RetrySettings"/> for calls to <c>BigtableClient.ReadRows</c> when the stream
+        /// of results is prematurely ended.
```
jskeet: "ends prematurely" rather than "is prematurely ended"?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1051,12 +1051,25 @@ internal BigtableClientImpl(BigtableServiceApiClient[] clients, string appProfil
         public override BigtableClient WithAppProfileId(string appProfileId) =>
             new BigtableClientImpl(_clients, appProfileId);
 
+        internal IClock Clock => UnderlyingClientSettings?.Clock ?? SystemClock.Instance;
```
jskeet: It might be worth copying this out of the settings on construction. I believe that's what we do for ApiCall - changes to settings are irrelevant after client construction. It would be good to be consistent about that.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowAsyncEnumerator.cs
```diff
@@ -90,18 +124,39 @@ public RowAsyncEnumerator(BigtableServiceApiClient.ReadRowsStream stream)
...
                 }
+
+                if (previouslyProcessedKey != _lastCompletedRowKey)
+                {
+                    // There was a full row found in the response.
+                    UpdateLastFoundKey(_lastCompletedRowKey);
+                }
+                else
+                {
+                    // The service indicates that it processed rows that did not match the filter, and will not
```
jskeet: I don't understand the comment - how do we know this from the fact that `_lastCompletedRowKey` is `previouslyProcessedRowKey`? I suspect this is just a bit of the protocol I'm unaware of, but it may be worth expanding the comment a little.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/Utilities.cs
```diff
@@ -28,28 +31,156 @@ public static class Utilities
...
+            TStreamItems[] entriesForInitialStream,
+            TStreamItems[][] entriesForRetryStreams,
+            Func<TStreamItems[], TMockStream> itemsToStream,
+            Action<TRequest, TMockStream> validator)
+            where TMockStream : TStream
         {
             var mock = new Mock<BigtableServiceApiClient>();
 
-            if (entriesForRetryResponses != null)
+            // Even though we want to setup the initial call last, we should call the stream conversion for it
```
jskeet: Why do we want to set up the initial call last?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
```diff
@@ -0,0 +1,450 @@
...
+                            {
+                                CreateChunk("b", "cf1", "column2", "value2", commitRow: true)
+                            }
+                        }
+                    }
+                });
+
+            var stream = client.ReadRows(request);
+
+            int rowCount = 0;
```
jskeet: Do we have any guarantee about the order in which the rows should be returned?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
```diff
@@ -0,0 +1,450 @@
...
+                                CreateChunk("b", "cf1", "column2", "value2", commitRow: true)
+                            }
+                        }
+                    }
+                });
+
+            var stream = client.ReadRows(request);
+
+            // We should be able to read one full row and then there should be an exception in the request validator
+            // of Utilities.CreateReadRowsMockClient because the response form the retry stream with the "b" row key
```
jskeet: Nit: form => from

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
```diff
@@ -0,0 +1,450 @@
...
+                },
+                responsesForRetryStreams: new[]
+                {
+                    new []
+                    {
+                        new ReadRowsResponse
+                        {
+                            Chunks =
+                            {
+                                CreateChunk("a", "cf1", "column1", "value1"),
```
jskeet: Is this guaranteed not to be (say) "value3"? What would happen in that case?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowAsyncEnumerator.cs
evildour: This comment was copied from [the Java implementation](https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/6d25f6b109302f2d84d473f76c0ef5093f15f1b3/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/scanner/RetryingReadRowsOperation.java#L191). It looks like it is slightly inaccurate because the `LastScannedRowKey` could be null, but `LastScannedRowKey` (and its analogue in Java) ignore null or empty row keys. I'll update the comment and local method name to be clearer.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
evildour: Yes, they are returned in lexicographical order. Are you worried that this test can be fooled by us incorrectly duplicating a row? I'll update it.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
evildour: Done

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
evildour: That should be fine. We would reset the state after getting the incomplete row from the first response, so the fact that it was "value1" would be forgotten. This could easily be "value3" and that's what would be seen as the result. I'll add another test for this scenario.

evildour:
Thanks Jon, PTAL. I had to rebase, so only the last commit is worth reviewing.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
jskeet: No, it was more that it feels like we could just all the results to a list and assert how they looked that way - this feels a little more roundabout to me.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
jskeet: Great, thanks - the extra test will definitely help to clarify.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1055,10 +1065,10 @@ internal BigtableClientImpl(BigtableServiceApiClient[] clients, string appProfil
 
         /// <inheritdoc/>
         public override BigtableClient WithAppProfileId(string appProfileId) =>
-            new BigtableClientImpl(_clients, appProfileId);
+            new BigtableClientImpl(_clients, appProfileId, UnderlyingClientSettings);
```
jskeet: I'm not suggesting we change this right now, but there's an interesting oddity that changes to the underlying settings are observed when we call `WithAppProfileId`. It's not quite "the same client but with different app profile ID" due to that. We could clone the settings, if we wanted to.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingReadRowsTest.cs
evildour: Yup, that's definitely better. Done.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Oh good point. I'll clone them in the constructor.

evildour:
Thanks again Jon. I'll merge on green.

igorbernstein2:
There is a really nasty edgecase that I dont think is covered here. @garye uncovered it in [the go client](https://github.com/GoogleCloudPlatform/google-cloud-go/commit/566ccf326f970b4fd1168bc0b6b474e94eeb6460). Under some circumstances, a ReadRows stream can send all of its rows, but complete in error. If I'm reading the code correctly, then I think the retry for that circumstance would trigger a full table scan because the RequestManager would return an empty request.

evildour:
Thanks for catching this @igorbernstein2. I'm not sure if that will happen or not but I'll add a test for that and make sure it doesn't re-request all rows.

igorbernstein2:
Ideally it would just close the stream w/o error

evildour:
Fixed in #1991. Thanks again.

PR #1981, opened 2018-03-09
Add more content to middleware snippets.

iantalarico:
This is in an attempt to make it more explicit that the middleware needs to be first to be able to properly work.

cc: @SurferJeffAtGoogle

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
@@ -22,7 +22,6 @@ using System.IO;
 using System.Net.Http;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.AspNetCore.Hosting;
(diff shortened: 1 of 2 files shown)


jskeet:
Will wait until the errors have been fixed before reviewing :)

iantalarico:
@jskeet Forgot to updated the csproj for all frameworks.  PTAL.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/AspNetCoreSnippets.cs
```diff
@@ -49,6 +49,15 @@ public void Configure(IApplicationBuilder app)
         {
             // Use before handling any requests to ensure all unhandled exceptions are reported.
             app.UseGoogleExceptionLogging();
+
+            app.UseStaticFiles();
```
jskeet: Is the order of these calls important? If so, let's call that out in comments.

iantalarico: Kinda but not really unless you had some overlap in routes in static files and mvc.  This is just how they order them in templates.

PR #1984, opened 2018-03-12
Remove DeletedFieldAttribute

### LINKED ISSUE #1983: Remove DeletedFieldAttribute

jskeet:
It can be restored if there's any user demand for it, but we can't
envisage its usage outside potential migration scenarios.

Fixes #1983.


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/SentinelValue.cs
```diff
@@ -77,10 +77,9 @@ internal static SentinelValue FromPropertyAttributes(PropertyInfo property)
             {
                 return ServerTimestamp;
             }
-            if (property.GetCustomAttribute<DeletedFieldAttribute>() != null)
-            {
-                return Delete;
-            }
+            // (No attribute for deleted fields)
+
+            // No attribute detected.
```
chrisdunelm: Is this comment meant to be here?

jskeet: Yes. The first comment is to make it clear that we haven't forgotten about deleted fields - there's just no attribute for them.

The second comment is to make it clear that if the property doesn't have any sentinel attributes, we just deliberately return null.

Suggested clearer alternatives would be welcome.

chrisdunelm: It's fine. My misunderstanding, that's all. I hadn't noticed that the next line was the null return.

PR #1986, opened 2018-03-13
Moving ClientCreationSettings nested class to BigtableClientPartial

AVaksman:
Moved ClientCreationSettings to BigtableClientPartial


evildour:
I think you merged with the newer changes. You should rebase instead so we don't have another 26k of changed lines.

AVaksman:
I did use `rebase`, but couldn't push the changes correctly the first time. Used `git push origin +[branchName]` as you previously advised and it looks good now.

evildour:
Thank you, but something odd still seems to be going on. The squash commit message contains messages from all the commits that you merged/rebase with and for some reason your change is showing up _before_ them in the commit list of the PR. I'm just going to close this one. It looks like we may change up the approach of using multiple underlying clients (and only use a single one), so these settings may have to change anyway. I'll [re]move if/when that happens.

AVaksman:
OK, Np.
You mean instead of creating a pool of `BigtableClient`s create a pool of `channel`s and have one `BigtableClient` (or many for that matter) round robin through channels in the `channelspool`? That would be nice. It will also eliminate an issue of inadvertently creating too many unnecessary `channel`s in case when user will be creating new `BigtableClient's for new operations.

evildour:
It's still up in the air, but yes, something like that.

PR #1987, opened 2018-03-13
Pristine copy of BreakingChangesDetector

evildour:
From https://github.com/Infragistics/BreakingChangesDetector

Note: this will not build. After this is merged, I will send multiple followup PRs to get it building, clean it up so it conforms to our style and uses newer C# features/conventions, uses newer style csproj files, etc.

This PR includes everything from the original repo except for some test projects from Mono.Cecil that contained pre-compiled assemblies (best to not add those to the repo history) and a TestDriver project which just had some basic code to experiment with the program while debugging.


evildour:
I had modified two of the files in Mono.Cecil. I'm not sure on the details, but I believe when an assembly had multiple modules, I was not able to get back to the proper assembly definition from an sub-module. Form [this line](https://github.com/jbevain/cecil/blob/master/Mono.Cecil/ModuleDefinition.cs#L1031), it looks like a dummy assembly definition is created for these. Not sure if there was another way around it though (or if there is now), but it looks like I added [this property](https://github.com/Infragistics/BreakingChangesDetector/blob/ae8bf0e1e287656200ae6950091db6fadf0ab541/Mono.Cecil/Mono.Cecil/ModuleDefinition.cs#L207) to check module owners for their assemblies if necessary.

That said, I want to try to rip out Mono.Cecil anyway and see if Roslyn will work in its place, so I'll just remove it now and we can always re-add it if necessary.

evildour:
Do you still want to take a look at this or are you ok with merging it at this point?

evildour:
Yeah, I'll do the "Squash and merge" operation.

jskeet:
Go for it - all I've done is look at the licence, and that's fine :)

evildour:
Ok thanks. I'll merge on green (not that this should affect anything from the build though).

PR #1989, opened 2018-03-13
Ensure all Diag IApplicationBuilder and IServiceCollection extensionsreturn their respective builder.

### LINKED ISSUE #1988: IServiceCollection.AddGoogleExceptionLogging(), and IServiceCollection.AddGoogleTracing() should both return a IServiceCollection object
To support syntax like:
```
var serviceProvider = new ServiceCollection()
                .AddGoogleExceptionLogging()
                .AddGoogleTrace()
                .AddLogging()
                .BuildServiceProvider();
```

Both AddGoogleExceptionLogging() and AddGoogleTrace should return an IServiceCollection object

iantalarico:
Fixes #1988


jskeet:
This is a backward-incompatible change. Are we already committed to a new major version for the other planned changed? If so, I'm fine with the change - if not, it'll need to wait until we *are* planning a new major version.

jskeet:
Just to be clear, there are two types of incompatibility here:

- Binary compatibility - I *believe* that changing the return type will break code already compiled against an earlier version
- Source compatibility - it definitely breaks compatibility in the admittedly-rare situation where a user is constructing a delegate using the method

iantalarico:
@jskeet Yes given that we are making other more important breaking changes I figured we may as well do this now.

jskeet:
SGTM.

PR #1991, opened 2018-03-15
[Bigtable] Fix ReadRows bug where a retry could try to read all rows

evildour:
See GoogleCloudPlatform/google-cloud-go@566ccf3 for more details.
cc @igorbernstein2

(also moved `ClientCreationSettings` out of the generated file while I was at it, so best to review commit by commit)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/Utilities.cs
```diff
@@ -92,8 +94,9 @@ public static class Utilities
 
             // All but the last stream should end with an RpcException which permits retrying with the
             // default RetrySettings so the higher level ReadRowsStream keeps retrying. The last stream
-            // should end normally.
-            for (int i = 0; i < streams.Count - 1; i++)
+            // should end normally, unless errorAtEndOfLastStream is true, in which case it should end
```
jskeet: Alternative change to consider here: set `ShouldErrorAtEnd` to true in the code above:

```csharp
var stream = new MockReadRowsStream(entries) { ShouldErrorAtEnd = true };
```

Then afterwards:

```csharp
streams.Last().ShouldErrorAtEnd = errorAtEndOfLastStream;
```

evildour: Yup, this is much cleaner :)

evildour: Actually, we don't need a list now, just the last stream.

PR #1993, opened 2018-03-15
Regenerate all APIs

jskeet:
This brings in new features of the codegen; I've split the PR into commits based on feature.


jskeet:
(Note that the additional test projects may well push the Travis build time beyond its limits. We'll see - I have a couple of ideas about that.)

PR #1997, opened 2018-03-15
Split Travis build into partitions

jskeet:
Currently Travis has issues anyway, but it might just work after that...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+env:
+- API_REGEX='Google\.Cloud\.[A-L].*'
+- API_REGEX='Google\.Cloud\.[M-Z].*'
+# List non-Cloud APIs here
+- API_REGEX='Google\.LongRunning'
 language: csharp
 mono: none
 dotnet: 2.1.4
@@ -13,4 +18,4 @@ addons:
     - dotnet-sharedframework-microsoft.netcore.app-1.0.5
 
 script:
-  - ./build.sh --diff
+  - ./build.sh --diff --regex $API_REGEX
--- a/build.sh
+++ b/build.sh
@@ -15,9 +15,15 @@ shopt -s nullglob
 # Command line arguments are the APIs to build. Each argument
 # should be the name of a directory, either relative to the location
 # of this script, or under apis.
+# Additional arguments:
+# --notests: Just build, don't run the tests
+# --diff: Detect which APIs to build based on a diff to the master branch
+# --regex regex: Only build APIs that match the given regex
+# --nobuild: Just list which APIs would be built; don't run the build
 apis=()
 runtests=true
 apiregex=
+nobuild=false
 while (( "$#" )); do
   if [[ "$1" == "--notests" ]]
   then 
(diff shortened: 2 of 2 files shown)


jskeet:
Added all three of you as reviewers because this is a somewhat silly-feeling approach... and because you probably all ought to know what's going on.

The build does seem to be happening though...

.travis.yml
```diff
@@ -1,3 +1,6 @@
+env:
+- API_REGEX='Google\.(Cloud\.)?[A-M].*'
```
iantalarico: Will this skip Google.LongRunning?

jskeet:
(This isn't a particularly even split, it turns out. But we could adjust that easily enough.)

iantalarico:
@jskeet Looks like half of the split is already done (yay!) but the half that isn't done has taken 2x as long.  Might need to split this up in a different way?

.travis.yml
jskeet: Nope, I thought of that - that's why Cloud is an optional segment :)

jskeet:
Yes, we can split it up any way we want, basically. There's a constant cost of building tools, and dependencies can go cross-regex (so split 1 still builds Spanner due to Google.Cloud.EntityFramework etc) but we could go to four splits easily if we wanted. This PR is more checking the basic premise.

jskeet:
The long split is still a bit long now - will try splitting that in the morning.

.travis.yml
evildour: This one looks like it will capture all of them, which is probably why one of the splits is taking 2x as long. For example, in `Google.Cloud.Vision.V1`, the 'V' fails to be matched by `[A-M]`, so the optional `(Cloud\.)` cannot be used, so the regex will fall back to capturing the leading 'C' from "Cloud" with the `[A-M]` pattern and then `.*` will capture "loud.Vision.V1"

evildour: This one looks like it will capture all of them, which is probably why one of the splits is taking 2x as long. For example, in "Google.Cloud.Vision.V1", the 'V' fails to be matched by `[A-M]`, so the optional `(Cloud\.)` cannot be used, so the regex will backtrack and capture the leading 'C' from "Cloud" with the `[A-M]` pattern and then `.*` will capture "loud.Vision.V1"

It's a shame bash regex doesn't support negative lookahead assertions, or this would be rather clean:
`Google\.(Cloud\.)?[A-M](?!loud\.).*`
meaning "don't match the `[A-M]` pattern if it is followed by "loud."

It looks like this might be the next best option:
`Google\.((Cloud\.)?[ABD-M])|(C[^l]).*`
which matches 'A'-'M' (except 'C') after an optional "Cloud." OR it matches 'C' which is followed by a character that is not 'l', and then anything else.

As Ian said, "Regexs are hard :)"

jskeet: Humbug. I thought I'd checked that all locally. Will look again - thanks for the explanation.

I suspect the simplest option may be to handle non-cloud APIs separately, explicitly listing Google.LongRunning as another job for now. Then make the Cloud part mandatory for everything else.

jskeet:
Timings with current PR:

- A-L: 23 minutes
- M-Z: 18 minutes 
- LongRunning: 2 minutes

That's a good enough start, I think. We may want to try doing the same thing in AppVeyor...

PR #1999, opened 2018-03-16
Partition AppVeyor build in same way as Travis

jskeet:


jskeet:
Actually, I'm not sure that we'll get parallel builds here anyway, and it definitely makes things more complicated in terms of code coverage. (We'd need to merge the coverage across builds.)

PR #2017, opened 2018-03-23
Regenerate all APIs

jskeet:
The expectation is that the difference here is just in terms of a
lot of partial methods on the ClientImpl classes.

(I'll check this myself.)


jskeet:
Looks fine to me apart from a couple of methods being removed from BigtableAdminClient. I'm checking that with @evildour.

[Looks like it was deliberate though](https://github.com/googleapis/googleapis/commit/69c531e5707c2f17a74fa60a38f113aa8a841d26#diff-05c73320211201124119d8c88eb6d5d2)

evildour:
Yup, you beat me to it. I was going to regenerate after my other PR was merged. Guess I don't have to now :)

jskeet:
Always happy to be of service :) Will merge on green.

PR #2023, opened 2018-03-26
Updates based on automated build process

jskeet:


PROCESSES.md
```diff
@@ -120,21 +120,12 @@ Releasing consists of five steps:
...
-- Run `buildrelease.sh` manually, specifying the commit that was
-  tagged. (`tagreleases.sh` outputs the command line to run)
-- Go into the `releasebuild/nuget` directory and push
-- Make sure you have a clone of the gh-pages branch, copy the
-  documentation from `releasebuild/docs/output/assembled` as the new
-  `docs` directory in the `gh-pages branch, commit and push.
+- Confirm that you wish to create the given tags.
+- A Jenkins build system will now notice the tags (after a short
+  delay; it checks periodically) and perform the rest of the build
+  and push process automatically.
```
chrisdunelm: I think it'd be worth keeping the full instructions to use `buildrelease.sh` and how to push the `nupkg`s somewhere in this doc. So if the automation does fail, it's still possible for people unfamiliar with the build details (e.g. me!) to complete a release.

jskeet: I've added a sub-bullet that doesn't give the instructions themselves, but says that they'll appear when running the right scripts.

PR #2024, opened 2018-03-26
Regenerate all APIs

jskeet:
These now include async request-taking methods which take
a CancellationToken.

Mike, could you check that the BigtableClient changes are okay?

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
@@ -104,6 +104,26 @@ namespace Google.Cloud.Bigtable.V2
         /// The request object containing all of the parameters for the API call.
         /// If the <see cref="MutateRowRequest.AppProfileId"/> has not been specified, it will be initialized from the value stored in the client.
         /// </param>
+        /// <param name="cancellationToken">
+        /// A <see cref="CancellationToken"/> to use for this RPC.
+        /// </param>
+        /// <returns>
+        /// A Task containing the RPC response.
+        /// </returns>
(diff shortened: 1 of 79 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClient.cs
```diff
@@ -290,6 +352,22 @@ public sealed partial class BigtableClientImpl : BigtableClient
...
+        {
+            if (request.AppProfileId == null)
+            {
+                request.AppProfileId = _appProfileId;
+            }
+
+            return GetUnderlyingClient().MutateRowAsync(request, cancellationToken);
+        }
+
+        => MutateRowAsync(
```
chrisdunelm: OK, not sure what's happened here.
Will look into it.

evildour: This is from the Bigtable custom generator, not the normal generator. I'll look into this one.

chrisdunelm: Ah right, thanks. That's a relief!

evildour:
Other than the generation error, the Bigtable changes look fine, but there's some unnecessary overrides being generated in the `BigtableClientImpl` class created by the custom generator. I can fix this in another PR, but this can be submitted before that. They are harmless and the same thing will happen with or without them.

jskeet:
We can't submit this PR as-is, as the build is currently broken... will look at your PR ASAP Mike.

jskeet:
Regenerated with the new BigTable codegen. Will merge on green.

PR #2025, opened 2018-03-26
[Bigtable] Fix error with re-writing expression-bodied methods

evildour:
Fixes the build errors in #2024


evildour:
Basically, the syntax tree node representing a method is able to hold onto a block (if you want a traditional method body) or an expression and semi-colon (if you want an expression-bodied method). It's possible to have all three of these set at the same time, which generates non-compilable code, as we have seen. So while I was replacing the block on the original method, I was not also clearing out the expression and semi-colon because I was never converting expression-bodied methods before.

chrisdunelm:
Thanks for the explanation, LGTM :)

PR #2026, opened 2018-03-26
Bigtable generator changes for CancellationToken overloads.

evildour:
This removes unnecessary overrides of the CancellationToken overloads from the BigtableClientImpl class and it also adds a CancellationToken overload to the split sync/async methods to replace the streaming MutateRows.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
@@ -33,6 +33,9 @@ namespace Google.Cloud.Bigtable.V2.GenerateClient
     {
         private const string AppProfileIdFieldName = "_appProfileId";
         private const string AppProfileIdPropertyName = "AppProfileId";
+        private const string CallSettingsTypeName = "CallSettings";
+        private const string CancellationTokenParameterName = "cancellationToken";
+        private const string CancellationTokenTypeName = "CancellationToken";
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
```diff
@@ -33,6 +33,9 @@ class Program
     {
         private const string AppProfileIdFieldName = "_appProfileId";
         private const string AppProfileIdPropertyName = "AppProfileId";
+        private const string CallSettingsTypeName = "CallSettings";
```
jskeet: Use nameof here and for CancellationTokenTypeName?

evildour: I used it for CancellationToken, but this doesn't have a reference to Gax.Grpc, so I can't use it for CallSettings. Do you think it make sense to add the reference just for this?

jskeet: I don't think the reference would do any harm - on the other hand, it's not like we're going to rename CallSettings. I'm okay either way.

evildour: Actually, nameof might be better in that it allows a future maintainer of the generator to go to definition and see what's available on that type. And since `nameof(X)` is about the same length as `XTypeName`, I think I'll remove the constants and use nameof in-place (one less go to definition required).

PR #2038, opened 2018-03-29
Add structured logs and a single default log for all AspNetCore logs.

### LINKED ISSUE #1932: Log names in ASP.NET Core logging
We have integrated the `Google.Cloud.Diagnostics.AspNetCore` package into our project and noticed that the logger passes the "category name" (`ILogger<TCategory>`) of the logger instance to `LogTarget` which creates a `logName` property:

https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/d0c81fd0756e69b664610b4438f4240f42d0f005/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs#L74

This results in a `logName` property like:
```
projects/my-project/logs/Microsoft.AspNetCore.Hosting.Internal.WebHost
``` 

Such logging names makes filtering all the applications logs in the Log Viewer pretty difficult: 
image

Is it possible consolidate the logger name to just one, such as `projects/my-project/logs/aspnetcore`, and still keep the full logger name available in another property for fine-grained filtering?

I might be missing out on something, so please point it out if my reasoning isn't correct.

### LINKED ISSUE #2019: Provide 'structure' log ability in Google.Cloud.Diagnostics.AspNetCore
Microsoft.Extensions.Logging provide structure logging out of box [MS docs](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?tabs=aspnetcore2x#log-message-template)
But this feature should be supported in **ILogger** implementation.

I see that google logs support several payloads for log entry, but for logging used simple textPayload which hold bare text. I found what **GoogleExceptionLogger** used jsonPayload to store information in more useful way.

So, I make little experiment:
rewrite **GoogleLogger.Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)** method:
```csharp
...
var jsonPayload = new Struct
{
    Fields = { { "Message", Value.ForString(message) } }
};

if (GoogleLoggerScope.Current != null)
{
    jsonPayload.Fields.Add("Scope", Value.ForString(GoogleLoggerScope.Current.ToString()));
}

if (state is IEnumerable<KeyValuePair<string, object>> structure)
{
    var properties = new Struct();
    foreach (var pair in structure)
    {
        if (pair.Key.Equals(OriginalFormatKeyValue))
        {
            continue;
        }

        properties.Fields.Add(pair.Key, Value.ForString(pair.Value.ToString()));
    }

    if (properties.Fields.Count > 0)
    {
        jsonPayload.Fields.Add("Properties", Value.ForStruct(properties));
    }
}

if (eventId.Id != 0 || eventId.Name != null)
{
    var eventStruct = new Struct();
    if (eventId.Id != 0)
    {
        eventStruct.Fields.Add("Id", Value.ForNumber(eventId.Id));
    }

    if (!string.IsNullOrEmpty(eventId.Name))
    {
        eventStruct.Fields.Add("Name", Value.ForString(eventId.Name));
    }

    jsonPayload.Fields.Add("EventId", Value.ForStruct(eventStruct));
}

LogEntry entry = new LogEntry
{
    Resource = _loggerOptions.MonitoredResource,
    LogName = _logName,
    Severity = logLevel.ToLogSeverity(),
    Timestamp = Timestamp.FromDateTime(_clock.GetCurrentDateTimeUtc()),
    JsonPayload = jsonPayload,
    Labels = { _loggerOptions.Labels },
    Trace = GetTraceName() ?? "",
};
...
```
and that send log w/o any error to GCP (also put eventId information).

Also I have configured out sink to BigQuery, and jsonPayload fit to table perfectly as record.

iantalarico:
This change does two things adds structured logs and sends all logs from this provider to a single Stackdriver log.

The structured log give the user more information at a more granular level which is easier to query.

Sending the logs to a single log makes browsing the logs in the Stackdriver Logs UI.  Due to how loggers are created users could get well over 10 logs one for each class using a logger which results in a large number of logs like "Microsoft.AspNetCore.*.*". This puts all logs into a single log 'aspnetcore'.  Users can further filter by the class name if they wish via the param 'log_name'. 

Fixes #1932 and fixes #2019.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
@@ -32,6 +32,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
     {
         private const string _logMessage = "a log message";
         private const string _logName = "log-name";
+        private const string _baseLogName = "aspnetcore";
         private const string _projectId = "pid";
         private static readonly DateTime s_dateTime = DateTime.UtcNow;
         private static readonly Exception s_exception = new Exception("some message");
@@ -46,20 +47,22 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
 
         private GoogleLogger GetLogger(
             IConsumer<LogEntry> consumer, LogLevel logLevel = LogLevel.Information,
-            Dictionary<string, string> labels = null, IHttpContextAccessor accessor = null)
+            Dictionary<string, string> labels = null, IHttpContextAccessor accessor = null,
+            string logName = null)
         {
-            LoggerOptions options = LoggerOptions.Create(logLevel, labels, MonitoredResourceBuilder.GlobalResource);
+            LoggerOptions options = LoggerOptions.Create(logName, logLevel, labels, MonitoredResourceBuilder.GlobalResource);
             return new GoogleLogger(consumer, s_logTarget, options, _logName, s_clock, accessor);
         }
 
         [Fact]
         public void BeginScope()
         {
-            var expectedMessage = $"scope => {_logMessage}";
-            Predicate<IEnumerable<LogEntry>> matcher = (l) => l.Single().TextPayload == expectedMessage;
+            Predicate<IEnumerable<LogEntry>> matcher = (l) =>
+                l.Single().JsonPayload.Fields["message"].StringValue == _logMessage &&
+                l.Single().JsonPayload.Fields["scope"].StringValue == "scope => ";
             var mockConsumer = new Mock<IConsumer<LogEntry>>();
             mockConsumer.Setup(c => c.Receive(Match.Create(matcher)));
-            var logger = GetLogger(mockConsumer.Object, LogLevel.Information);
+            var logger = GetLogger(mockConsumer.Object, logLevel:LogLevel.Information);
             using (logger.BeginScope("scope"))
             {
                 logger.Log(LogLevel.Error, 0, _logMessage, null, Formatter);
@@ -70,8 +73,9 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
         [Fact]
         public void BeginScope_Nested()
         {
-            var expectedMessage = $"parent => child => {_logMessage}";
-            Predicate<IEnumerable<LogEntry>> matcher = (l) => l.Single().TextPayload == expectedMessage;
+            Predicate<IEnumerable<LogEntry>> matcher = (l) =>
+                l.Single().JsonPayload.Fields["message"].StringValue == _logMessage &&
+                l.Single().JsonPayload.Fields["scope"].StringValue == "parent => child => ";
             var mockConsumer = new Mock<IConsumer<LogEntry>>();
             mockConsumer.Setup(c => c.Receive(Match.Create(matcher)));
             var logger = GetLogger(mockConsumer.Object, LogLevel.Information);
@@ -127,16 +131,25 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
         [Fact]
         public void Log()
         {
+            var message = "a {things} message with stuff";
+            var logStr = "log";
+
             var labels = new Dictionary<string, string> { { "some-key", "some-value" } };
             Predicate<IEnumerable<LogEntry>> matcher = logEntries =>
             {
                 LogEntry entry = logEntries.Single();
                 KeyValuePair<string, string> label = entry.Labels.Single();
-                return entry.LogName == new LogName(_projectId, _logName).ToString() &&
+                var json = entry.JsonPayload.Fields;
+                return entry.LogName == new LogName(_projectId, _baseLogName).ToString() &&
                     entry.Severity == LogLevel.Error.ToLogSeverity() &&
                     string.IsNullOrWhiteSpace(entry.Trace) &&
                     entry.Timestamp.Equals(Timestamp.FromDateTime(s_dateTime)) &&
-                    entry.TextPayload == Formatter(_logMessage, s_exception) &&
+                    json["message"].StringValue == "a log message with stuff" &&
+                    json["log_name"].StringValue == _logName &&
+                    json["event_id"].StructValue.Fields["id"].NumberValue == 28 &&
+                    json["format_parameters"].StructValue.Fields.Count == 2 &&
+                    json["format_parameters"].StructValue.Fields["things"].StringValue == logStr &&
+                    !json.ContainsKey("scope") &&
                     entry.Resource.Equals(MonitoredResourceBuilder.GlobalResource) &&
                     label.Key == "some-key" &&
                     label.Value == "some-value";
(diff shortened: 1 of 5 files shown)


jskeet:
Does this need to be reviewed before Tuesday? I can do it if it's blocking anything, but otherwise I'd like to wait.

henkmollema:
Nice work Ian, this will help a ton! 👍

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
```diff
@@ -127,25 +131,74 @@ public void Log_EmptyFormattedMessage()
...
                 LogEntry entry = logEntries.Single();
                 KeyValuePair<string, string> label = entry.Labels.Single();
-                return entry.LogName == new LogName(_projectId, _logName).ToString() &&
+                var json = entry.JsonPayload.Fields;
+                return entry.LogName == new LogName(_projectId, _baseLogName).ToString() &&
                     entry.Severity == LogLevel.Error.ToLogSeverity() &&
                     string.IsNullOrWhiteSpace(entry.Trace) &&
                     entry.Timestamp.Equals(Timestamp.FromDateTime(s_dateTime)) &&
-                    entry.TextPayload == Formatter(_logMessage, s_exception) &&
+                    json["message"].StringValue == "a log message with stuff" &&
```
henkmollema: Is the raw, non-formatted log message (e.g. `"a {things} message with stuff"`) also logged somewhere? It would be nice if we can filter on the message format regardless of the substituted values.

iantalarico: Yes it is, added an explicit check for this.

henkmollema: Awesome!

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
```diff
@@ -98,13 +103,52 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+                }
+                if (!string.IsNullOrWhiteSpace(eventId.Name))
+                {
+                    eventStruct.Fields.Add("name", Value.ForString(eventId.Name));
+                }
+                jsonStruct.Fields.Add("event_id", Value.ForStruct(eventStruct));
+            }
+
+            // If we have format params and its more than just the original message add them.
+            if (state is IEnumerable<KeyValuePair<string, object>> formatParams &&
```
jskeet: This will iterate over `formatParams` multiple times. Is that okay?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
```diff
@@ -56,21 +67,27 @@ public sealed class LoggerOptions
         /// Create a new instance of <see cref="LoggerOptions"/>.
         /// </summary>
         /// <param name="logLevel">Optional, the minimum log level.  Defaults to <see cref="LogLevel.Information"/></param>
+        /// <param name="logName">Optional, the name of the log.  Defaults to 'aspnetcore'.</param>
```
jskeet: If we're keeping this as the first parameter, let's move the docs for it above `logLevel`. But perhaps it should be the second parameter anyway? What are users more likely to want to change - the name or the level?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
```diff
@@ -127,25 +131,74 @@ public void Log_EmptyFormattedMessage()
         [Fact]
         public void Log()
         {
+            var message = "a {things} message with stuff";
```
jskeet: So what's doing the format-filling here? Is it:

- Our code?
- ASP.NET Core?
- Stackdriver?

(Basically I'm nervous if it's our code, as this kind of thing is easy to mess up in terms of security...)

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
```diff
@@ -98,13 +103,52 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
...
+            }
+
+            // If we have format params and its more than just the original message add them.
+            if (state is IEnumerable<KeyValuePair<string, object>> formatParams && 
+                !(formatParams.Count() == 1 && formatParams.Single().Key.Equals("{OriginalFormat}")))
+            {
+                var paramStruct = new Struct();
+                foreach (var pair in formatParams)
+                {
+                    paramStruct.Fields.Add(pair.Key, Value.ForString(pair.Value?.ToString() ?? ""));
```
jskeet: I wonder whether we should detect whether the value is formattable, and ask it to format itself in the invariant culture if so?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
iantalarico: This is done by ASP.NET Core in the [Logger Extensions](https://github.com/aspnet/Logging/blob/release/2.0/src/Microsoft.Extensions.Logging.Abstractions/LoggerExtensions.cs)

We don't do any formatting at all.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
iantalarico: It's okay but ideally we shouldn't iterate over it.

Can you elaborate on how this is happening?  Do you have suggestions to avoid this?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
iantalarico: I don't fully understand.  Do you mean check if the string is the format string? 

The fully formatted string is added in the 'message' feild

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
iantalarico: Log level is more likely to change.  Updated.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
iantalarico: Sounds good, I don't expect this to be a large list in any case.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
iantalarico: Yea I think this is overkill.  The formatting of the full message is all handled above I don't think we need to worry about this here.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
iantalarico: Sounds good.  This will be put in the next major release which is 3.0

jskeet: Sorry, had forgotten we were already at 2.x. Glad you caught my meaning :)

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
jskeet: Okay. It's probably worth adding a comment (not quite a TODO) as a prompt in case we come back here. I can see how it could be confusing in some cases. (Classic case: if you're running in a culture of ar-SA and you log a date, it'll come out in an entirely different calendar system...)

iantalarico: Done.

iantalarico:
Thanks Jon!

PR #2040, opened 2018-03-30
[Bigtable] Fix issue where SetCell mutations created without versions…

evildour:
… were using a 0 version.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
@@ -282,5 +282,33 @@ namespace Google.Cloud.Bigtable.V2.IntegrationTests
                         }
                     }));
         }
+
+        [Fact]
+        public async Task OmittedVersion()
+        {
+            var beforeInsertTimestamp = DateTime.UtcNow.AddMilliseconds(-1);
+
+            var tableName = _fixture.TableName;
+            var client = _fixture.TableClient;
+            BigtableByteString rowKey = Guid.NewGuid().ToString();
+            var response = await client.MutateRowsAsync(
+                tableName,
+                Mutations.CreateEntry(
+                    rowKey,
+                    Mutations.SetCell(
+                        BigtableFixture.DefaultColumnFamily,
+                        "column_name",
+                        "test12345")));
+
+            var entries = response.Entries.OrderBy(e => e.Index);
+            Assert.True(entries.All(e => e.Status.Code == (int)Code.Ok));
+
+            var row = await client.ReadRowAsync(tableName, rowKey);
+            var cell = row.Families[0].Columns[0].Cells[0];
+            var insertTimestamp = cell.Version.ToDateTime();
+            Assert.True(
+                beforeInsertTimestamp <= insertTimestamp,
+                $"Expected insert time to be on or after {beforeInsertTimestamp:s}, but was {insertTimestamp:s}");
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
```diff
@@ -282,5 +282,33 @@ public class MutateRowsTest
                         }
                     }));
         }
+
+        [Fact]
+        public async Task OmittedVersion()
+        {
+            var beforeInsertTimestamp = DateTime.UtcNow.AddMilliseconds(-1);
```
jskeet: Why the AddMilliseconds(-1) out of interest? The `<=` should cope if time is moving monotonically... and if we run the test across an NTP sync it could change by more than 1ms.

Or is this to avoid a rounding problem in converting between micros and DateTime?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
```diff
@@ -282,5 +282,33 @@ public class MutateRowsTest
...
+                tableName,
+                Mutations.CreateEntry(
+                    rowKey,
+                    Mutations.SetCell(
+                        BigtableFixture.DefaultColumnFamily,
+                        "column_name",
+                        "test12345")));
+
+            var entries = response.Entries.OrderBy(e => e.Index);
+            Assert.True(entries.All(e => e.Status.Code == (int)Code.Ok));
```
jskeet: I think there may be an `Assert.All` available - I'm not sure. (Worth checking.)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
evildour: Yup, it's to avoid rounding issues. We round down to the millisecond because that's the granularity supported by Bigtable, so without backing up a millisecond, this was failing occasionally.

jskeet: Righto - makes perfect sense, thanks.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.IntegrationTests/MutateRowsTest.cs
evildour: I think I looked into this at one point, but decided against it. Assert.All takes an action delegate, not a func used as a predicate, so you'd have to do something like this:
```csharp
Assert.All(entries, e => Assert.Equal((int)Code.Ok, e.Status.Code));
```
Assert.True(...) seemed just slightly less ugly. But if you feel strongly about it I can change it.

jskeet: I don't feel strongly about it - but if it fails, I think Assert.All would give more information about the actual value. No biggie either way though.

evildour: Oh that's a good point. I've used this in a few places. I'll fix them all up in a followup PR.

PR #2043, opened 2018-04-03
Adding ListTables tests

AVaksman:


evildour:
Thanks Alex. These edits were made to a generated file. Please add a new file with the same name, but ending in ".cs" instead of ".g.cs" and add the new tests there and then I'll review.

AVaksman:
Sure, I am out today, will do it in the evening as soon as I am back.

evildour:
Thanks Alex, I appreciate the effort. Looking at them now though, I'm wondering why it was necessary. It looks like these should have been auto-generated with the rest. @chrisdunelm, is there something about [ListTables](https://github.com/googleapis/googleapis/blob/9abcd48fbb901913271ddab966297d10b599712e/google/bigtable/admin/v2/bigtableadmin_gapic.yaml#L492) that prevents its tests from being generated?

AVaksman:
We where asked to verify that all Admin functions should have tests. I saw that `ListTables` didn't have.

evildour:
Sorry, that comment wasn't directed at you. I realize why we want the tests. I'm saying I would have thought your effort wasn't necessary because these should have been auto-generated.

evildour:
Looks like Chris is out until next week. Is this blocking you Alex or can it wait?

AVaksman:
It is not blacking me.

chrisdunelm:
@evildour, @AVaksman These are tests for paged response methods, which are currently unsupported in the unit test auto-gen; so I think this PR can be merged.
At some point the unit test auto-gen will be extended to support paged response methods, but probably not very soon. At that point these can be deleted.

evildour:
Ok thanks Chris. So @AVaksman, please add a TODO above the tests saying to remove them when they are auto-generated so we don't forget to clean them up. Also, it looks like there are still some changes to the generated file. Please revert these. Once those changes are made then I'll merge. Thanks again.

AVaksman:
Done

evildour:
Thanks Alex! I'll merge when green.

AVaksman:
Thanks

PR #2050, opened 2018-04-04
Allow set with merge and no data

jskeet:
This should create an operation with an empty (but present) update
path.

@jba: Could you have a look at this? It's implementing your conformance test change.
(There may be a simpler way to do this, but I can't immediately spot it.)


jskeet:
Chris: Sebastian and Jonathan have confirmed they don't have any cunning plan for making this massively simpler, so it's fine to review now for general C#-ness.

PR #2057, opened 2018-04-05
Fix failing integration tests.

iantalarico:
We do two things here:
- Add a rate limiter reset to the diag web host tests.
- Run tests in serial.  We need to do this as the tests depend on the rater limiter being reset

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/AssemblyInfo.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/AssemblyInfo.cs
@@ -14,4 +14,6 @@
 
 using Xunit;
 
+// This is needed to ensure trace tests that rely on the RateLimiter
+// do not affect each other.
 [assembly: CollectionBehavior(DisableTestParallelization = true)]
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/AssemblyInfo.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/AssemblyInfo.cs
@@ -1,0 +1,17 @@
+﻿// Copyright 2018 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


henkmollema:
I also call this method in `TestTrace` in the same file. Is that a problem/still necessary?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/AssemblyInfo.cs
```diff
@@ -0,0 +1,17 @@
...
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
evildour: This file (with the same content) was added on #1117 and then removed in #1375. It might be good to add a comment in here explaining why it is needed (because it isn't obvious when just looking at this one file) so we don't remove it again.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsWebHostTests.cs
```diff
@@ -33,6 +33,14 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
 {
     public class DiagnosticsWebHostTests
     {
+        public DiagnosticsWebHostTests()
+        {
+            // The rate limiter instance is static and only set once.  If we do not reset it at the
+            // beginning of each tests the qps will not change.  This is dependent on the tests not
+            // running in parallel.
+            RateLimiter.Reset();
```
evildour: Can the `RateLimiter.Reset()` call down on line 110 now be removed?

jskeet:
(I'll leave Mike to review this as I'm away for a couple of days. :)

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/AssemblyInfo.cs
iantalarico: Comment added.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsWebHostTests.cs
iantalarico: Removed.

iantalarico:
Thanks Mike!

PR #2058, opened 2018-04-08
Create BigQuery library version 1.2.0

jskeet:
New feature: geo-regionalization, which allows datasets and jobs to
be created in specific location.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -35,11 +35,11 @@
     "id": "Google.Cloud.BigQuery.V2",
     "productName": "Google BigQuery",
     "productUrl": "https://cloud.google.com/bigquery/",
-    "version": "1.1.0",
+    "version": "1.2.0",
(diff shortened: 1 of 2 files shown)


jskeet:
(No need to review this on Sunday, but I'd like to create a release on Monday morning.)

apis/apis.json
```diff
@@ -35,11 +35,11 @@
     "id": "Google.Cloud.BigQuery.V2",
     "productName": "Google BigQuery",
     "productUrl": "https://cloud.google.com/bigquery/",
-    "version": "1.1.0",
```
chrisdunelm: Is this not going to beta for a bit, before GA?

jskeet: No, I think we're happy with the change here (regionalization is the only change) so we're just going straight to GA.

PR #2059, opened 2018-04-08
GoogleStackdriverTarget for NLog

snakefoot:
Pretty much copy-pasted from the log4net appender, but without the in-memory LogQueue.

Planning to create a generic LogQueue-TargetWrapper for NLog (with retry and timeout) that can be reused by any NLog target.

### DIFF
--- a/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleCloudLoggingTarget.cs
+++ b/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleCloudLoggingTarget.cs
@@ -29,6 +29,8 @@ using NLog.Config;
 using NLog.Targets;
 using NLog.Common;
 using NLog.Layouts;
+using System.Threading.Tasks;
+using System.Threading;
 
 namespace Google.Cloud.Logging.NLog
 {
@@ -40,7 +42,7 @@ namespace Google.Cloud.Logging.NLog
     {
         /// <summary>
         /// The file path of a service account JSON file to use for authenitication.
-        /// Not necessary if running on GCE or GAE or if the GOOGLE_APPLICATION_CREDENTIALS environment variable has been set.
+        /// Not necessary if running on GCE, GAE or GKE, or if the GOOGLE_APPLICATION_CREDENTIALS environment variable has been set.
         /// </summary>
         public Layout CredentialFile { get; set; }
 
@@ -77,7 +79,7 @@ namespace Google.Cloud.Logging.NLog
 
         /// <summary>
         /// The project ID for all log entries.
-        /// Must be configured if not executing on Google Compute Engine, Google App Engine, or Google Kubernetes Engine.
+        /// Must be configured if not executing on Google Compute Engine, Google App Engine or Google Kubernetes Engine.
         /// If running on GCE, GAE or GKE, the ProjectId will be automatically detected if not set.
         /// </summary>
         public Layout ProjectId { get; set; }
@@ -106,8 +108,8 @@ namespace Google.Cloud.Logging.NLog
         private MonitoredResource _resource;
         private string _logName;
         private LogNameOneof _logNameToWrite;
-        private System.Threading.Tasks.Task _prevTask;
-        private System.Threading.CancellationTokenSource _cancelTokenSource;
+        private Task _prevTask;
+        private CancellationTokenSource _cancelTokenSource;
 
         private static readonly string[] s_oAuthScopes = new string[] { "https://www.googleapis.com/auth/logging.write" };
         private static readonly Dictionary<string, string> s_emptyLabels = new Dictionary<string, string>();
@@ -123,23 +125,20 @@ namespace Google.Cloud.Logging.NLog
         };
 
         /// <summary>
-        /// Constructor
+        /// Construct a Google Cloud loggin target.
         /// </summary>
-        public GoogleCloudLoggingTarget()
-            :this(null)
+        public GoogleCloudLoggingTarget() : this(null)
         {
         }
 
-        /// <summary>
-        /// Constructor
-        /// </summary>
-        public GoogleCloudLoggingTarget(LoggingServiceV2Client client)
+        // For testing only.
+        internal GoogleCloudLoggingTarget(LoggingServiceV2Client client)
         {
             ResourceLabels = new List<TargetPropertyWithContext>();
             _contextProperties = new List<TargetPropertyWithContext>();
             _client = client;
             LogId = "Default";
-			OptimizeBufferReuse = true;
+            OptimizeBufferReuse = true;
         }
 
         /// <summary>
@@ -147,9 +146,9 @@ namespace Google.Cloud.Logging.NLog
         /// </summary>
         protected override void InitializeTarget()
         {
-            _cancelTokenSource = new System.Threading.CancellationTokenSource();
+            _cancelTokenSource = new CancellationTokenSource();
 
-            _platform = _platform ?? Platform.Instance();
+            _platform = Platform.Instance();
             string logId = LogId?.Render(LogEventInfo.CreateNullEvent());
             if (string.IsNullOrWhiteSpace(logId))
             {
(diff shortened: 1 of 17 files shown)


snakefoot:
I signed it!

chrisdunelm:
Thanks for this PR :) Someone should be able to start taking a look at it in the next few days.

jskeet:
Assigned Chris to it as he has experience with the Log4Net appender :)

snakefoot:
@chrisdunelm Updated the NLog-Test-Project to also include `Xunit.SkippableFact`. Is something missing for review to complete?

chrisdunelm:
@snakefoot Sorry for the delay; no, nothing is missing; I'll do my best to get to this soon.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleCloudLoggingTarget.cs
```diff
@@ -0,0 +1,389 @@
...
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        protected override void Write(AsyncLogEventInfo logEvent)
+        {
+            var continuation = logEvent.Continuation;
+            var logEntry = BuildLogEntry(logEvent.LogEvent);
+            _prevTask = _client.WriteLogEntriesAsync(_logNameToWrite, _resource, s_emptyLabels, new[] { logEntry }, _cancelTokenSource.Token)
```
chrisdunelm: @snakefoot What happens if this write fails? Are the log entries permanently lost, or does NLog contain retry logic? If not, what would happen if we put a retry with backoff here?

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleCloudLoggingTarget.cs
```diff
@@ -0,0 +1,389 @@
...
+using NLog.Layouts;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace Google.Cloud.Logging.NLog
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    [Target("GoogleCloud")]
```
chrisdunelm: We'd really only what this ever used with the/an async wrapper; you've probably seen the Log4Net code essentially implements async itself.
Is there any way to enforce using the async wrapper?
What's the NLog idiom for targets that should only be used with the async wrapper?

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleCloudLoggingTarget.cs
snakefoot: One can use https://github.com/NLog/NLog/wiki/RetryingWrapper-target to enable retry. There is also https://github.com/nlog/NLog/wiki/FallbackGroup-target to use alternative target if this fails.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleCloudLoggingTarget.cs
snakefoot: Would think that `_client.WriteLogEntriesAsync(...)' was a non-blocking operation, so should not matter whether  async-wrapper is used. But there is no way to force one to use async-wrapper, besides recommendation in Wiki.

snakefoot:
> @chrisdunelm Consider the target name. We may well want to include "Stackdriver" in the name somewhere.

Have no problem with changing the type-name from ´GoogleCloud` to `GoogleStackDriver`. Maybe also rename the class from `GoogleCloudLoggingTarget` to `GoogleStackDriverTarget` ?

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleCloudLoggingTarget.cs
chrisdunelm: It's the buffering behaviour that I'm more interested in.
In a application with a lot of logging, we do not want every log statement in the application to lead to an API call to stackdriver.
Having read the async-wrapper docs, it looks like the async-wrapper will provide this buffering functionality. Please can you confirm if this is correct? Thanks.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleCloudLoggingTarget.cs
snakefoot: Yes the retry-wrapper just need to be placed before the async-wrapper. So it becomes retry-wrapper -> async-wrapper -> google-stack-driver-target.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleCloudLoggingTarget.cs
snakefoot: Yes the async-wrapper will enable batching of logevents, so they arrive in chunks instead of individual logevents.

snakefoot:
@chrisdunelm Should I make a commit where I update the NLog-target-name to GoogleStackDriver and renames the class to GoogleStackDriverTarget ? (And update smoketest etc.)

chrisdunelm:
@snakefoot Sorry, I didn't notice your replies a couple of days ago.
Yes, please do a commit to rename as in your comment: `GoogleStackdriverTarget` and `GoogleStackdriver` (note lower-case 'd').

Once this is done, I think the code is ready to merge. I'll prepare the docs in a future PR, then we should be able to do an initial beta release.
Thanks very much for persisting with this :)

snakefoot:
@chrisdunelm Renamed target, and also introduced a new Task-throttle parameter (TaskPendingLimit)

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
```diff
@@ -0,0 +1,474 @@
...
+                }
+            }
+
+            if (logEntries.Count > 0)
+            {
+                WriteLogEntries(logEntries, continuationList);
+            }
+        }
+
+        private void WriteLogEntries(IList<LogEntry> logEntries, object continuationList)
```
chrisdunelm: I don't quite understand the logic behind this throttling.
It appears that it'll wait one second, but if there are still too many tasks pending, it'll continue anyway; is that correct?
Could it wait until `_pendingTaskCount` is less than `TaskPendingLimit`? (with some kind of timeout, which will cause failure?)
To do this properly, I suspect a list/queue of Tasks will need to be kept, rather than just a single_prevTask.
What do you think?

Please can you also add a new boolean, or refactor the code some other way, so that `belowTaskLimit` doesn't change meaning.

And please comment the final code, so readers, ie me ;) , can understand it.

This task-limit code will also need testing.

I'm happy to do any/all of this if you'd like. Please let me know.

snakefoot: Yes if the `_prevTask` has gone bad, then it just starts a new task-chain ignoring the one that doesn't complete by itself.

I have now made the timeout a little more aggressive, so if throttle is required, then it doesn't wait for all pending tasks to complete, but just waits until below the pending task limit.

Not sure I understand the case with `belowTaskLimit` changing meaning. Please explain.

snakefoot: @chrisdunelm Should I make the timeout of 1.5 sec configurable, so users can configure longer delays for slower internet ?

chrisdunelm: Sorry, my mistake about `belowTaskLimit`. I had misread the code slightly.
Yes, I agree it would be a good idea to make the timeout configurable; but I think it's time to get this PR merged. I'll create an issue with the remaining work to do before release, including this timeout configuration.

bddckr:
Hey thanks for the nice work! I can't find `Google.Cloud.Logging.NLog` on NuGet, anyone know what's up with that?

snakefoot:
This is now available on nuget: https://www.nuget.org/packages/Google.Cloud.Logging.NLog/



Sent from my Samsung device


-------- Original message --------

PR #2063, opened 2018-04-10
Fixups required for generation changes

jskeet:
Will rerun API generation with these in place, then run a build.

cc @evildour as there are BigtableClient generator changes

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
@@ -28,6 +28,7 @@ using System.Threading.Tasks;
 
 using static Google.Cloud.Bigtable.V2.GenerateClient.RoslynHelpers;
 using static Microsoft.CodeAnalysis.CSharp.SyntaxFactory;
+using gax = Google.Api.Gax;
 
 namespace Google.Cloud.Bigtable.V2.GenerateClient
 {
@@ -189,7 +190,7 @@ namespace Google.Cloud.Bigtable.V2.GenerateClient
                     var clientMethod = (MethodDeclarationSyntax)requestMethodRewriter.Visit(methodSyntax);
                     userClientSyntax = userClientSyntax.AddMembers(clientMethod);
 
-                    if (method.Parameters[1].Type.Name.EndsWith("CallSettings"))
+                    if (method.Parameters[1].Type.Name.EndsWith(nameof(CallSettings)))
                     {
                         var clientImplMethod = (MethodDeclarationSyntax)requestMethodToImplRewriter.Visit(clientMethod);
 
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
```diff
@@ -275,7 +278,7 @@ public override SyntaxNode VisitMethodDeclaration(MethodDeclarationSyntax node)
                 node = node.WithBodySafe(
                     node.Invoke(
                         IdentifierName(requestParameterSyntax.Identifier),
-                        IdentifierName(nameof(CallSettings)).Member(nameof(CallSettings.FromCancellationToken)).Invoke(IdentifierName(CancellationTokenParameterName))));
+                        IdentifierName($"gaxgrpc::{nameof(CallSettings)}").Member(Identifier(nameof(CallSettings.FromCancellationToken))).Invoke(IdentifierName(CancellationTokenParameterName))));
```
evildour: `IdentifierName(...)` should just be used to create simple names. With something like this, you should really create an AliasQualifiedNameSyntax. Use `ParseName(...)` here instead.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
```diff
@@ -221,8 +222,10 @@ private class CustomStreamMethodInfo
 
             // Create a CompilationUnitSyntax from the Usings node of the original file, which will also contain the
             // copyright notice and generated code warnings in its leading trivia.
-            var compilationUnit =
-                CompilationUnit().WithUsings(syntaxTree.GetCompilationUnitRoot().Usings);
+            // We also need a using directive for GAX, so that we can use ResultWithUnwrappedExceptions.
+            var usings = syntaxTree.GetCompilationUnitRoot().Usings
+                .Add(UsingDirective(IdentifierName(typeof(gax::TaskExtensions).Namespace)));
```
evildour: IdentifierName => ParseName

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/RoslynHelpers.cs
```diff
@@ -96,7 +96,7 @@ internal static class RoslynHelpers
         internal static XmlEmptyElementSyntax SeeTag(CrefSyntax cref) =>
             XmlEmptyElement("see").AddAttributes(XmlCrefAttribute(cref));
 
-        internal static SyntaxToken Task() => Identifier("Task");
+        internal static SyntaxToken Task() => Identifier("stt::Task");
```
evildour: This is no longer just a simple identifier, so this should really use `ParseName(...)` and then the type should be changed to `NameSyntax`, but that causes other problems in some calling code. This should be fine for now, but I'll update in a follow up PR to be correct just in case we reuse these RoslynHelpers in the future in other places.

jskeet: SGTM - I'll add a TODO.

jskeet:
Okay, it's looking good - merging.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.GenerateClient/Program.cs
evildour: That still should use ParseName as it is a QualifiedNameSyntax (uses '.' to qualify as opposed to '::' for AliasQualifiedNameSyntax). IdentifierName is just a simple identifier that could be used as a local name, for example.

PR #2064, opened 2018-04-10
Regenerate all APIs

jskeet:
This includes #2063 - I'm mostly creating it as a pull request to check that it doesn't break CI.


jskeet:
Okay, rebased to just a single commit. Will obviously wait for green, but PTAL just in case you have concerns :)

PR #2065, opened 2018-04-11
Allow RoslynHelpers to create generics from complex name syntax nodes

evildour:


evildour:
Yes, that's correct. I'd like to move all the Roslyn utility code into a common tools project at some point, so I'd like this to cover all cases. That said, only the overload for `QualifiedNameSyntax` is unused at this point.

evildour:
Arg, as soon as I merged this, I realized I forgot one last cleanup step. I'll send another PR for this shortly.

PR #2069, opened 2018-04-11
Make RunInSpanAsync in DelegatingTracer really async

kontab:
Fixes incorrect time span reported in trace.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/DelegatingTracerTest.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/DelegatingTracerTest.cs
@@ -1,0 +1,63 @@
+﻿// Copyright 2018 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+using TraceProto = Google.Cloud.Trace.V1.Trace;
+
+namespace Google.Cloud.Diagnostics.Common.Tests
+{
+    public class DelegatingTracerTest
+    {
+        [Fact]
+        public async Task RunInSpanAsync_CorrectSpan()
+        {
+            const string _traceId = "trace-id";
+            const string _projectId = "project-id";
+            const string traceName = "trace-name";
+            const int delayInSeconds = 1;
+
+            TraceProto actualTraceProto = null;
+
+            Mock<IConsumer<TraceProto>> mockConsumer = new Mock<IConsumer<TraceProto>>();
+            mockConsumer.Setup(x => x.Receive(It.IsAny<IEnumerable<TraceProto>>()))
+                .Callback<IEnumerable<TraceProto>>(arg => actualTraceProto = arg?.FirstOrDefault());
+
+            IManagedTracer _simpleTracer = SimpleManagedTracer.Create(mockConsumer.Object, _projectId, _traceId);
+            IManagedTracer _delegatingTracer = new DelegatingTracer(() => _simpleTracer);
+
+            var stopwatch = Stopwatch.StartNew();
+            await _delegatingTracer.RunInSpanAsync(async () =>
+            {
+                await Task.Delay(TimeSpan.FromSeconds(delayInSeconds));
+                return 0;
+            }, traceName);
+            stopwatch.Stop();
+
+            Assert.NotNull(actualTraceProto);
+            Assert.Equal(delayInSeconds, SpanInSeconds(actualTraceProto));
+        }
+
+        private static long SpanInSeconds(TraceProto actualTraceProto)
+        {
+            return actualTraceProto.Spans.First().EndTime.Seconds - actualTraceProto.Spans.First().StartTime.Seconds;
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


kontab:
.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/DelegatingTracerTest.cs
```diff
@@ -0,0 +1,63 @@
...
+            var stopwatch = Stopwatch.StartNew();
+            await _delegatingTracer.RunInSpanAsync(async () =>
+            {
+                await Task.Delay(TimeSpan.FromSeconds(delayInSeconds));
+                return 0;
+            }, traceName);
+            stopwatch.Stop();
+
+            Assert.NotNull(actualTraceProto);
+            Assert.Equal(delayInSeconds, SpanInSeconds(actualTraceProto));
```
iantalarico: Looks like this failed on one test run and I think has potential to be flaky.  I think we should be checking a range instead.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/DelegatingTracerTest.cs
```diff
@@ -0,0 +1,63 @@
...
+            IManagedTracer _simpleTracer = SimpleManagedTracer.Create(mockConsumer.Object, _projectId, _traceId);
+            IManagedTracer _delegatingTracer = new DelegatingTracer(() => _simpleTracer);
+
+            var stopwatch = Stopwatch.StartNew();
+            await _delegatingTracer.RunInSpanAsync(async () =>
+            {
+                await Task.Delay(TimeSpan.FromSeconds(delayInSeconds));
+                return 0;
+            }, traceName);
+            stopwatch.Stop();
```
ILMTitan: What is this stopwatch used for?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/DelegatingTracerTest.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            Assert.NotNull(actualTraceProto);
+            Assert.Equal(delayInSeconds, SpanInSeconds(actualTraceProto));
+        }
+
+        private static long SpanInSeconds(TraceProto actualTraceProto)
+        {
+            return actualTraceProto.Spans.First().EndTime.Seconds - actualTraceProto.Spans.First().StartTime.Seconds;
```
ILMTitan: Unless the difference is exactly one second, this could give different values for the same total duration.
A better way to do this calculation is:
```
Duration duration = actualTraceProto.Spans.First().EndTime - actualTraceProto.Spans.First().StartTime;
return duration.Seconds;
```

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/DelegatingTracerTest.cs
```diff
@@ -0,0 +1,63 @@
...
+using TraceProto = Google.Cloud.Trace.V1.Trace;
+
+namespace Google.Cloud.Diagnostics.Common.Tests
+{
+    public class DelegatingTracerTest
+    {
+        [Fact]
+        public async Task RunInSpanAsync_CorrectSpan()
+        {
+            const string _traceId = "trace-id";
```
ILMTitan: Only fields should begin with `_`. Local variables should simply be *lowerCamelCase*.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/DelegatingTracer.cs
```diff
@@ -46,8 +46,8 @@ internal DelegatingTracer(Func<IManagedTracer> managedTracerGetter)
             _managedTracerGetter().RunInSpan(func, name, options);
 
         /// <inheritdoc />
-        public Task<T> RunInSpanAsync<T>(Func<Task<T>> func, string name, StartSpanOptions options = null) =>
-            _managedTracerGetter().RunInSpan(func, name, options);
+        public async Task<T> RunInSpanAsync<T>(Func<Task<T>> func, string name, StartSpanOptions options = null) =>
+            await _managedTracerGetter().RunInSpanAsync(func, name, options).ConfigureAwait(false);
```
evildour: Just return the task from `RunInSpanAsync`: remove `async`, `await`, and `.ConfigureAwait(false)`. No need for the compiler generated async machinery here.

henkmollema: I think the `ConfigureAwait(false)` should stay since this code might run in environments with a synchronization context.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/DelegatingTracerTest.cs
kontab: I've changed test to assert InRange between assumed span and stopwatch indication, hopefully it will be stable-passing now. But I must say that I don't understand why "actual" was  so wrong at first place.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/DelegatingTracerTest.cs
kontab: Now it's used to set the upper limit of Assert.InRange

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/DelegatingTracer.cs
kontab: I've removed async/await from method body. I guess it will be just passing task and without await it won't capture context. However, if I'm mistaken, please correct me.

evildour: You shouldn't need `ConfigureAwait(false)` because you're not awaiting anything anymore. You're passing the Task back to the caller to decided when to wait for it, so the caller that ultimately awaits on it decides whether to call ConfigureAwait or not.

Issue #1926, opened 2018-02-26
Logged with log4net but log lines never appear in stackdriver.


SurferJeffAtGoogle:
Like this user report on stack overflow,
https://stackoverflow.com/questions/48962850/cant-write-log-entry-with-google-clouds-log4net-integration
I followed the instructions here:
https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.Logging.Log4Net/index.html
Wrote this code:
https://github.com/SurferJeffAtGoogle/csharp-docs-samples/tree/log4net/logging/api/Log4NetSample
And ran it:
```
~/gitrepos/dotnet-docs-samples/logging/api/Log4NetSample$ dotnet run
log4net: log4net assembly [log4net, Version=2.0.8.0, Culture=neutral, PublicKeyToken=669e0ddf0bb1aa2a]. (.NET Framework [.NET Core 4.6.0.0] on Linux 4.9.0-5-amd64 #1 SMP Debian 4.9.65-3+deb9u2 (2018-01-04))
log4net: defaultRepositoryType [log4net.Repository.Hierarchy.Hierarchy]
log4net: Creating repository for assembly [Log4NetSample, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]
log4net: Assembly [Log4NetSample, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null] Loaded From [Not supported on .NETCore]
log4net: Assembly [Log4NetSample, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null] does not have a RepositoryAttribute specified.
log4net: Assembly [Log4NetSample, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null] using repository [log4net-default-repository] and repository type [log4net.Repository.Hierarchy.Hierarchy]
log4net: Creating repository [log4net-default-repository] using type [log4net.Repository.Hierarchy.Hierarchy]
log4net: configuring repository [log4net-default-repository] using file [log4net.xml]
log4net: configuring repository [log4net-default-repository] using stream
log4net: loading XML configuration
log4net: Configuring Repository [log4net-default-repository]
log4net: Configuration update mode [Merge].
log4net: Logger [root] Level string is [ALL].
log4net: Logger [root] level set to [name="ALL",value=-2147483648].
[...]

PR #2071, opened 2018-04-12
Regenerate all APIs

jskeet:
I'm doing a docs build locally to check that moving away from a single-character namespace alias has the desired effect.

Will also skim the diff to make sure the changes are what we expect.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ResourceNames.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ResourceNames.cs
@@ -15,15 +15,15 @@
 // Generated code. DO NOT EDIT!
 
 using gax = Google.Api.Gax;
-using s = System;
-using System.Linq;
+using sys = System;
+using linq = System.Linq;
 
 namespace Google.Cloud.Bigtable.V2
 {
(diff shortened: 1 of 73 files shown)


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/ResourceNames.cs
```diff
@@ -127,7 +127,7 @@ public partial class CheckAndMutateRowRequest
         /// </summary>
         public TableName TableNameAsTableName
         {
-            get { return string.IsNullOrEmpty(TableName) ? null : Google.Cloud.Bigtable.V2.TableName.Parse(TableName); }
+            get { return string.IsNullOrEmpty(TableName) ? null : TableName.Parse(TableName); }
```
jskeet: This doesn't work because it will resolve TableName to the string property, not the type. It needs to be qualified somehow.

jskeet:
Assigning this to Chris as the easiest way of showing the error.

chrisdunelm:
Fixed in toolkit [#1926](https://github.com/googleapis/toolkit/pull/1926)

jskeet:
Amended the PR after regeneration - fingers crossed...

PR #2072, opened 2018-04-12
Generate Dialogflow v2

jskeet:


jskeet:
Thanks. I won't do a release until I've got a way of checking that it works.

jskeet:
Whoops - had forgotten to commit the project file changes etc. That's accidentally now in the same commit, but I think that's okay.

PR #2075, opened 2018-04-13
(Not for submission) Remove all optional parameters from Google.Cloud.Firestore

jskeet:
(Except for FirestoreDb.Create and FirestoreDb.CreateAsync.)

This is for the purpose of being able to add more overloads later on
without breaking compatibility (or at least, reducing the impact).

Currently two proto conformance tests fail, because we "shouldn't"
be able to specify a precondition of MustExist explicitly, although
that was the default behaviour. I've left the tests failing rather
than skipping them, as we probably want to think carefully about this.


jskeet:
@mikelehen to review.

jskeet:
Note: I've almost certainly missed plenty of precondition checks against null at the moment.

jskeet:
Closing - we're sticking with Options classes.

PR #2088, opened 2018-04-17
Fix root snippets to use common resource names

jskeet:
(We can decide later whether or not this is appropriate; we may well
want more documentation about common vs API-specific resource names,
but let's fix the build first.)


jskeet:
Will wait for green before merging this time.

PR #2089, opened 2018-04-17
Make the AppVeyor build differ more between PR and merge

jskeet:
On a PR, we just run the build without coverage.
After a merge, run the build with code coverage as well.

It's easier to do this conditional work in a shell script, and we
don't really care much about the different build phases, so do it
all in a script.

### DIFF
--- a/appveyor.sh
+++ b/appveyor.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+dotnet --info
+echo "Regenerating projects: if this fails, run generateprojects.sh and commit changes"
+bash generateprojects.sh && git diff --exit-code
+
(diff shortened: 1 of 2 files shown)


appveyor.sh
```diff
@@ -0,0 +1,25 @@
...
+  if [[ -f "coverage/coverage-filtered.xml" ]]
+  then
+    codecov -f "coverage/coverage-filtered.xml"
+  fi
+else
+  # We're in a pull request. Don't do any coverage; just
+  # build and run tests.
+  bash build.sh --diff
+fi
+
```
chrisdunelm: nit: soooo many blank lines

jskeet: Ooh, apparently so. Interesting. Fixing.

PR #2091, opened 2018-04-18
Prune the root directory of unnecessary files

jskeet:
- .gitmodules is now empty, and has been for a while
- We haven't used myget for a long time
- Directory.Build.targets is now obsolete as the issue has been
  fixed in the .NET Core SDK


jskeet:
Will just wait until the current non-PR AppVeyor build has completed before merging this.

PR #2092, opened 2018-04-18
Implement ValueComparer for Firestore

jskeet:
This will be used within the Watch implementation.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ValueComparerTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ValueComparerTest.cs
@@ -1,0 +1,193 @@
+﻿// Copyright 2018, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Firestore.V1Beta1;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Google.Cloud.Firestore.Tests
+{
+    using System.Linq;
+    using static ProtoHelpers;
+
+    public class ValueComparerTest
+    {
+        // Taken from OrderTest.java, with a few extra tests.
+        // These are compared pairwise.
+        private static readonly dynamic[] s_valuesInOrder =
+        {
+            // Null
+            CreateNullValue(),
+
+            // Booleans
+            false,
+            true,
+
+            // Numbers
+            double.NaN,
+            double.NegativeInfinity,
+            double.MinValue,
+            long.MinValue,
+            int.MinValue,
+            -1.1,
+            -1L,
+            0L,
+            double.Epsilon,
+            0.1,
+            1L,
+            1.1,
+            int.MaxValue,
+            long.MaxValue,
+            double.MaxValue,
+            double.PositiveInfinity,
+
+            // Timestamps
+            new Timestamp(123, 0),
+            new Timestamp(123, 123),
+            new Timestamp(345, 0),
+
+            // Strings
+            "",
+            "\u0000\ud7ff\ue000\uffff",
+            "(╯°□°）╯︵ ┻━┻",
+            "AB",
+            // Longer strings later
+            "ABC",
+            // Case-sensitive, by ordinal - lower-case ASCII comes after upper-case ASCII
+            "a",
+            "abc def",
+            // Latin small letter e + combining acute accent + latin small letter b
+            "e\u0301b",
+            "æ",
+            // Latin small letter e with acute accent + latin small letter a
+            "\u00e9a",
+            // Surrogate pair (verdigris)
+            "\U0001F728",
+            // Note: positioning of this after the surrogate pair is due to using UTF-16
+            // rather than UTF-8 as per the Firestore spec. This mirrors the Java behavior,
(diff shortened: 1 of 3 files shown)


jskeet:
Java implementation and test, for reference:

https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-firestore/src/main/java/com/google/cloud/firestore/Order.java
https://github.com/GoogleCloudPlatform/google-cloud-java/blob/master/google-cloud-firestore/src/test/java/com/google/cloud/firestore/OrderTest.java

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueComparer.cs
```diff
@@ -0,0 +1,251 @@
...
+                    return CompareObjects(left, right);
+                default:
+                    throw new InvalidOperationException($"Unknown type {leftType}");
+            }
+        }
+
+        // Note: this follows the Java ordinal comparison, but the spec says to compare by UTF-8.
+        private int CompareStrings(Value left, Value right) => StringComparer.Ordinal.Compare(left.StringValue, right.StringValue);
+
+        private int CompareBlobs(Value left, Value right)
```
chrisdunelm: Add a note about omitting truncation, as noted in the spec?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueComparer.cs
```diff
@@ -0,0 +1,251 @@
...
+                    return CompareArrays(left, right);
+                case TypeOrder.Object:
+                    return CompareObjects(left, right);
+                default:
+                    throw new InvalidOperationException($"Unknown type {leftType}");
+            }
+        }
+
+        // Note: this follows the Java ordinal comparison, but the spec says to compare by UTF-8.
+        private int CompareStrings(Value left, Value right) => StringComparer.Ordinal.Compare(left.StringValue, right.StringValue);
```
chrisdunelm: Add a note about omitting truncation, as noted in the spec?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueComparer.cs
```diff
@@ -0,0 +1,251 @@
...
+            return result != 0 ? result : leftPoint.Longitude.CompareTo(rightPoint.Longitude);
+        }
+
+        private int CompareResourcePaths(Value left, Value right)
+        {
+            // Note: this does no validation, but it's simple...
+            // We can't just do a string comparison, because of cases such as:
+            // foo/bar/baz
+            // foo/bar-/baz
+            // Here "bar-" shoudl come after "bar" because it's longer, but '-' is earlier than '/'.
```
chrisdunelm: typo: "should"

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueComparer.cs
```diff
@@ -0,0 +1,251 @@
...
+            }
+            return leftArray.Count.CompareTo(rightArray.Count);
+        }
+
+        private int CompareObjects(Value left, Value right)
+        {
+            // This requires iterating over the keys in the object in order and doing a
+            // deep comparison.
+            var leftMap = new SortedDictionary<string, Value>(left.MapValue.Fields, StringComparer.Ordinal);
+            var rightMap = new SortedDictionary<string, Value>(right.MapValue.Fields, StringComparer.Ordinal);
```
chrisdunelm: I don't think `SortedDictionary`'s are required.
Could you use something like `left.MapValue.Fields.OrderBy(x => x.Key)`

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueComparer.cs
```diff
@@ -0,0 +1,251 @@
...
+        {
+            if (left.ValueTypeCase == ValueTypeOneofCase.DoubleValue)
+            {
+                if (right.ValueTypeCase == ValueTypeOneofCase.DoubleValue)
+                {
+                    return CompareDoubles(left.DoubleValue, right.DoubleValue);
+                }
+                else
+                {
+                    return CompareDoubles(left.DoubleValue, right.IntegerValue);
```
chrisdunelm: `IntegerValue` is a `long`, so this could lead to precision loss and incorrect ordering I think.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ValueComparerTest.cs
```diff
@@ -0,0 +1,183 @@
...
+            long.MinValue,
+            int.MinValue,
+            -1.1,
+            -1L,
+            0L,
+            double.Epsilon,
+            0.1,
+            1L,
+            1.1,
+            int.MaxValue,
```
chrisdunelm: A test value of something like `(double)(long.MaxValue - 1)` should be added here to check for integer/double comparison when the integer will lose precision when converted to a double.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueComparer.cs
chrisdunelm: I was thinking of a situation where:

```C#
double d = (double)(1L << 53);
left.DoubleValue = d;
right.IntegerValue = (long)d + 1;
```

Here the comparison as-is will return 0, but the IntegerValue really does contain a larger value than the DoubleValue, so it should not return 0.
What do you think?

jskeet: I think we can't tell whether the double value *actually* came from `1 << 53`, or whether it came from `1 << 53 + 1.5` (unless I'm missing something). I don't think we can give any value which will be correct more often than it's not correct, so we should stick with simplicity. (This is what the Java code does, btw - and quite possibly what the server code does too.)

chrisdunelm: OK, sounds reasonable.

PR #2093, opened 2018-04-18
Update Google Logging dependency to 2.1.0

### LINKED ISSUE #2086: Error reporting not working from GKE
Hello,

I've setup error reporting on an asp .net core project that I've pushed to GKE.
Using the same docker image on GCE, I do have error reported but on GKE I don't.

Looking at the log from the container does show the stack trace of the exception i'm provoking, but no error about error reporting.
I did not provided the projectId as the library can retrieve those information by itself.
I've tried with the version 2.0 and 3.0-beta of the library.
I also tried to provide a serviceaccount using the env variable, the serviceaccount was owner of the project, but it did not help.

How can I troubleshot this issue ?

Thanks,

Yann

iantalarico:
Fixes #2086


iantalarico:
@jskeet Yes, will do right after this.

PR #2096, opened 2018-04-19
Remove dependency on IHttpContextAccessor from ContextTracerManager

nico-francois-vex:
Remove dependency on `IHttpContextAccessor` from `ContextTracerManager` using `AsyncLocal<>`, and make it public so it can be used to trace code not in the HTTP request pipeline (e.g. a pub/sub pull handler).


nico-francois-vex:
I also just signed the CLA.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -153,13 +153,11 @@ internal static TraceHeaderContext CreateTraceHeaderContext(IServiceProvider pro
 
         /// <summary>
         /// Creates a singleton <see cref="IManagedTracer"/> that is a <see cref="DelegatingTracer"/>.
-        /// The <see cref="DelegatingTracer"/> will use an <see cref="IHttpContextAccessor"/> under the hood
-        /// to get the current tracer which is set by the <see cref="ContextTracerManager"/>.
+        /// The <see cref="DelegatingTracer"/> will get the current tracer which is set by the <see cref="ContextTracerManager"/>.
         /// </summary>
         internal static IManagedTracer CreateManagedTracer(IServiceProvider provider)
```
iantalarico: No need for the provider anymore.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -153,13 +153,11 @@ internal static TraceHeaderContext CreateTraceHeaderContext(IServiceProvider pro
...
         {
-            var accessor = provider.GetServiceCheckNotNull<IHttpContextAccessor>();
-            return ManagedTracer.CreateDelegatingTracer(() => ContextTracerManager.GetCurrentTracer(accessor));
+            return ManagedTracer.CreateDelegatingTracer(() => ContextTracerManager.GetCurrentTracer());
```
iantalarico: suggestion:

```
internal static IManagedTracer CreateManagedTracer() =>
    ManagedTracer.CreateDelegatingTracer(() => ContextTracerManager.GetCurrentTracer());
```

Or just remove this function all together now that the logic is so simple.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextTracerManager.cs
```diff
@@ -13,36 +13,32 @@
...
-        ///<summary>The key to save the tracer under in the <see cref="HttpContext"/>.</summary> 
-        internal const string TraceKey = "Google.Cloud.Diagnostics.AspNetCore.Trace";
+        private static readonly AsyncLocal<IManagedTracer> _currentTracer = new AsyncLocal<IManagedTracer>();
 
         /// <summary>
         /// Sets the current <see cref="IManagedTracer"/>.  This is called in <see cref="CloudTraceMiddleware"/>
         /// for each new request.
         /// </summary>
-        public static void SetCurrentTracer(IHttpContextAccessor accessor, IManagedTracer tracer)
+        public static void SetCurrentTracer(IManagedTracer tracer)
```
iantalarico: Can we leave this as internal unless you see a good reason to make it public? If we do leave it as public let's add some comments here about what the side effects will be.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
nico-francois-vex: Correct :)

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
nico-francois-vex: Yes, I considered doing that too, but didn't want to touch the original code too much. Removing it altogether definitely makes sense.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/ContextTracerManager.cs
nico-francois-vex: To use the tracer from other code, so code that's not in the HTTP request pipeline (e.g. a pub/sub pull handler) this will need to be public.

At Vente-Exclusive we use the class below to create a custom trace context, and as you can see it needs access to `ContextTracerManager.SetCurrentTracer`.  Perhaps it makes sense to move the `ContextTracerManager` to `Google.Cloud.Diagnostics.Common`, together with the class below -- and then make the `CloudTraceMiddleware` use them. That way the same system would be used for any trace context, in HTTP context or not.  And in that case `ContextTracerManager` could be internal again.

```csharp
public class CloudTraceContext : IDisposable
{
    private readonly ISpan _span;
    private readonly IManagedTracer _tracer;
    private readonly Action<ISpan> _annotateSpan;
    private readonly IMicroserviceContext _microserviceContext;
 
    public CloudTraceContext(string name, Action<ISpan> annotateSpan, bool? shouldTrace, IMicroserviceContext microserviceContext, Func<TraceHeaderContext, IManagedTracer> tracerFactory)
    {
        _annotateSpan = annotateSpan;
        _microserviceContext = microserviceContext;
 
        _tracer = tracerFactory(TraceHeaderContext.Create(null, 0, shouldTrace));
        ContextTracerManager.SetCurrentTracer(_tracer);
 
        if (_tracer.GetCurrentTraceId() != null)
        {
            _span = _tracer.StartSpan(name);
        }
    }
 
    public void SetStackTrace(Exception ex)
    {
        if (_tracer.GetCurrentTraceId() != null)
        {
            try
            {
                _tracer.SetStackTrace(new StackTrace(ex, true));
            }
            catch (Exception innerException)
            {
                throw new AggregateException(innerException, ex);
            }
        }
    }
 
    public void Dispose()
    {
        if (_span != null)
        {
            _annotateSpan?.Invoke(_span);

            var labels = TraceLabels.GetAgentLabel();
            labels.Add("/quantum/module", _microserviceContext.Module);
            labels.Add("/quantum/component", _microserviceContext.Component);
            labels.Add("/quantum/version", _microserviceContext.Version);
            _span.AnnotateSpan(labels);
 
            _span.Dispose();
        }
        ContextTracerManager.SetCurrentTracer(null);
    }
}
```

nico-francois-vex: For completeness, this is an example of how to use the context above:

```csharp
public async Task RunInContext(string name, Func<Task> action, Action<ISpan> annotateSpan = null, bool? shouldTrace = null)
{
    using (var context = CreateContext(name, annotateSpan, shouldTrace))
    {
        try
        {
            await action();
        }
        catch (Exception ex)
        {
            context.SetStackTrace(ex);
            throw;
        }
    }
}
```

nico-francois-vex: So what I'm trying to say is that I could submit a PR with all or parts of this together, as you think makes most sense. I'm flexible 😁

iantalarico: Thanks for the extra context.  I think moving `ContextTracerManager` to common sounds like a good move.

I think I see what you mean here.  Basically a generalize version of CloudTraceMiddleware that is disposable (which ends the span)?  If so I like that.  It would open up the usage of this library.

nico-francois-vex: OK great 😃 I'll update the pull request then taking along the feedback and discussion here.

Any naming suggestions? Does `CloudTraceContext` as shown above sound good to you, or would you like to propose a different name?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Trace/CloudTraceExtensionTest.cs
```diff
@@ -76,10 +76,14 @@ public void CreateTraceHeaderContext_UseBackUpFunc()
...
-            mockProvider.Setup(p => p.GetService(typeof(IHttpContextAccessor))).Returns(new HttpContextAccessor());
+            var mockTracer = new Mock<IManagedTracer>();
+            string traceId = Guid.NewGuid().ToString("N");
+            mockTracer.Setup(p => p.GetCurrentTraceId()).Returns(traceId);
+            ContextTracerManager.SetCurrentTracer(mockTracer.Object);
             var tracer = CloudTraceExtension.CreateManagedTracer(mockProvider.Object);
             Assert.IsType<DelegatingTracer>(tracer);
-            mockProvider.VerifyAll();
+            Assert.Equal(traceId, tracer.GetCurrentTraceId());
+            mockTracer.Verify();
```
ILMTitan: This `Verify()` does nothing, as no setups were marked with `Verifiable()`. I would drop it completely, as the `Assert.Equal(traceId, ...)` call sufficiently tests the behavior.

nico-francois-vex:
@iantalarico I just resubmitted the PR with the changes as discussed here.

Code was moved to .Common and I added two new classes: `CloudTraceContext` and `CloudTracer` to enable creating trace contexts outside of the ASP.NET Core pipeline.  Maybe these are not the best names, so feel free to suggest better ones :)  Or other improvements that could still be made...

nico-francois-vex:
Let me show you an example on how we use `CloudTracer` in a Google Pub/sub pull handler, which is obviously not part of the HTTP request pipeline:

```csharp
        public async Task<SubscriberClient.Reply> Handle(PubsubMessage message, CancellationToken cancellationToken)
        {
            try
            {
                return (SubscriberClient.Reply)await _cloudTracer.RunInContext(_traceName, () => OnHandle(new PubsubMessage<T>(message), cancellationToken),
                    span => span.AnnotateSpan(new Dictionary<string, string> {
                    { "/message/id", message.MessageId },
                    { "/message/size", message.Data.Length.ToString() },
                    { "/message/time", message.PublishTime.ToString() }
                    }));
            }
            catch (Exception ex)
            {
                _logger.LogError(ex, "Error handling Pub/sub message");
                return SubscriberClient.Reply.Nack;
            }
        }
```

jskeet:
Hmm. At the very least, a `CloudTracer` that isn't actually a tracer but provides *factories* for tracers feels confusing. I'm generally in favour of keeping the public API as small as possible.

I think we should decide two things separately:

- Do we want to expose the functionality of `CloudTracer` via the public API?
- If we do, is that genuinely common between ASP.NET and ASP.NET Core (and would expect to be forever) or would it be better to have separate functionality to allow them to diverge?

iantalarico:
@jskeet  @nico-francois-vex  Thanks for the adding context and input.

I think for now we will just expose enough that users have more flexibility.  I'm not 100% sure exposing these extra here is the best for the public surface of the API.

@nico-francois-vex  With just changing `ContextTracerManager` to be common and public users (and you) could still accomplish this just by writing the custom code themselves right?

nico-francois-vex:
@iantalarico I can resubmit without those two classes indeed, and they (or something similar) can be put in user code as long as `ContextTracerManager` stays public and relies on `AsyncLocal<>` now.

nico-francois-vex:
I must add I do feel you're missing somewhat of an opportunity here, since without especially the `CloudTraceContext` class it's not very clear how one should trace non-HTTP code ;)

iantalarico:
@nico-francois-vex Yes that sounds good.  Let's leave `ContextTracerManager` as you have it here.

I also do agree we are missing a way for people to use this library on non-HTTP code.  This has been on my mind for some time and will get done, I just need to make sure we expose it in the right place and in a way we are okay committing too (this may end up me asking you to resubmit part of this PR at some point ;) ).

Thanks for your contribution!

nico-francois-vex:
@iantalarico pull request resubmitted without `CloudTraceContext` and `CloudTracer`

Indeed, feel free to come back to this ;)

nico-francois-vex:
If you guys are happy, all good for me! 👍

iantalarico:
👍

iantalarico:
@nico-francois-vex This is now released in 3.0.0-beta04

nico-francois-vex:
@iantalarico Great, thx!

PR #2101, opened 2018-04-20
Implement comparisons for DocumentReference and CollectionReference

jskeet:
Note that we don't *really* need `PathComparer` to implement `IComparer<string>` - we could just have a static method. This feels a little tidier though.


jskeet:
cc @mikelehen - while we don't *have* to make the types implement `IComparable<T>`, it feels like the polite thing to do. We *could* just have an `IComparer<DocumentReference>` that we kept internal, but it feels like being able to order these is generally useful.

PR #2102, opened 2018-04-23
Spanner struct parameters

jskeet:
Definitely review one commit at a time.

Lots of room for changes if required. I'm most interested in the API of SpannerStruct.
(We may want to consider an overload to SpannerParameterCollection to make it easier to add structs. Not sure.)

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/StructParameterTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/StructParameterTests.cs
@@ -1,0 +1,304 @@
+﻿// Copyright 2018 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.IntegrationTests
+{
+    [Collection(nameof(TestDatabaseFixture))]
+    public class StructParameterTests
+    {
+        private readonly TestDatabaseFixture _fixture;
+
+        public StructParameterTests(TestDatabaseFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public async Task SimpleFieldAccess()
+        {
+            var structParam = new SpannerStruct
+            {
+                { "threadf", SpannerDbType.Int64, 1 },
+                { "userf", SpannerDbType.String, "bob" }
+            };
+            using (var connection = await _fixture.GetTestDatabaseConnectionAsync())
+            {
+                using (var cmd = connection.CreateSelectCommand("SELECT @struct_param.userf, @p4"))
+                {
+                    cmd.Parameters.Add("struct_param", structParam.GetSpannerDbType(), structParam);
+                    cmd.Parameters.Add("p4", SpannerDbType.Int64, 10);
+
+                    using (var reader = await cmd.ExecuteReaderAsync())
+                    {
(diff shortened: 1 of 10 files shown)


jskeet:
Will have a look at the CI failures. The plan will be to merge this into a new branch rather than master for now.

jskeet:
(Removing you from assignment until I've fixed the bug, Chris. Feel free to review if you'd otherwise be twiddling your thumbs.)

jskeet:
Okay, I now understand the issue, and part of it is easily fixed - I'll push an extra commit in a minute.
The other part is *not* easily fixed. SpannerStruct doesn't implement IEquatable, and it's hard to make it do so in a way which is useful for tests - because the value in a field can itself be a list or an array. That means our equality assertion fails in the deserialization tests.

I think the way to go is to effectively extract the deserialization tests involving structs, and hand-write the equivalents.

jskeet:
Okay, now actually ready for review - still one commit at a time. Do ping me if anything doesn't make sense.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -333,12 +307,35 @@ public override int GetHashCode()
...
+            internal string Name { get; }
+            internal SpannerDbType Type { get; }
+
+            public StructField(string name, SpannerDbType type)
+            {
+                Name = GaxPreconditions.CheckNotNull(name, nameof(name));
+                Type = GaxPreconditions.CheckNotNull(type, nameof(type));
+            }
+
+            public override bool Equals(object obj) => obj is StructField sf ? Equals(sf) : false;
```
evildour: What about `obj is StructField sf && Equals(sf);`?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/StructParameterTests.cs
```diff
@@ -0,0 +1,304 @@
...
+                        Assert.True(await reader.ReadAsync());
+                        Assert.True(reader.IsDBNull(0));
+                        Assert.False(await reader.ReadAsync());
+                    }
+                }
+            }
+        }
+
+        [Fact]
+        public async Task ServerStructEquality()
```
evildour: Maybe add an associated equality check using the wrong values to verify the values are actually being compared?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
jskeet: Done. (Not sure why I didn't think of that...)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerStruct.cs
```diff
@@ -0,0 +1,108 @@
...
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+
+        /// <summary>
+        /// Adds a new field to the struct with the given name, type and value.
+        /// </summary>
+        /// <param name="name">The name of the field. May be null, and does not need to be unique within the struct.
+        /// If the value is null, the <see cref="Field"/> value added will have an empty string for the name.</param>
+        /// <param name="type">The type of the field. Must not be null.</param>
+        /// <param name="value">The value of the field. May be null.</param>
+        public void Add(string name, SpannerDbType type, object value) =>
```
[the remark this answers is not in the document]
jskeet: Done. I'll make Add(Field) return the field as well, just for consistency.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/StructParameterTests.cs
jskeet: Sure. (This was basically a list of suggested tests - happy to add another one though.)

jskeet:
@evildour: When you looked through this, did you do so sufficiently thoroughly that it's not worth @chrisdunelm doing so as well, or was it a skim?

evildour:
I gave it a thorough review. That said, it always helps to have another set of eyes on it, but it isn't necessary.

jskeet:
Fab - given that it's only going into a new branch anyway (not straight to master) I'll merge it, and Chris can look later with no risk of problems. Thanks!

PR #2103, opened 2018-04-23
Remove the Delegating Tracer Test

iantalarico:
This only tests the call through of the delegating tracer to its child tracer.   It doesn't test any logic and is flaky


evildour:
After looking at some of the failures, it seems very strange. It looks like the trace is taking 0 time despite the delay of 1 second. Have you been able to reproduce this flakiness locally?

evildour:
Ok, I'm able to reproduce it. Part of the problem is that the test is using `stopwatch.Elapsed.Seconds` instead of `stopwatch.Elapsed.TotalSeconds` and occasionally, the stopwatch only uses 999 milliseconds, so its Seconds component is actually 0. TotalSeconds returns 0.999 in these cases. However, the Start/EndTime of the spans have the same problem because they could happen within the same second, making their difference 0. So the code that determines the number of seconds in the span also needs to account for the Nanos portion of the Duration:
```csharp
private static double SpanInSeconds(TraceProto actualTraceProto)
{
    var duration = actualTraceProto.Spans.First().EndTime - actualTraceProto.Spans.First().StartTime;
    return duration.Seconds + TimeSpan.FromMilliseconds(duration.Nanos / 1000000).TotalSeconds;
}
```
And then because that `Assert.InRange` call is assuming a minimum of 1 second, I think the `Task.Delay` call should be padded a bit, maybe by doing
```csharp
await Task.Delay(TimeSpan.FromSeconds(delayInSeconds * 1.001));
```
I'm guessing due to the granularity of the of the timer `Task.Delay` is using, it might be off by a millisecond or two.

When I do these 3 things, I can no longer reproduce the issue when doing the test in a loop 1000 times.

However, that said, if this test is really useless, then I have no problem with it being removed. I just wanted to make sure the failures weren't indicative of a real issue.

iantalarico:
@evildour Thanks for digging into this more (sorry was a little jetlagged yesterday).  

This test really doesn't add anything test wise.  If there was an issue it would be in `SimpleManagedTracer`

PR #2104, opened 2018-04-24
Add an immutable DocumentSet

jskeet:
This is an internal-only data structure used to incrementally gather
results of a query, while retaining the query ordering. (The
creation of a comparer from a query will be in another PR.)

It uses the .NET immutable collections to avoid reimplementing an
immutable sorted map and set.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/DocumentSet.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/DocumentSet.cs
@@ -1,0 +1,88 @@
+﻿// Copyright 2018, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/DocumentSet.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Google.Cloud.Firestore
+{
+    /// <summary>
+    /// Immutable set of DocumentSnapshots, ordered by a custom comparer.
+    /// </summary>
+    internal sealed class DocumentSet : IEnumerable<DocumentSnapshot>
+    {
+        // TODO: Investigate whether the immutable collections are okay here. ImmutableSortedSet uses an AVL tree,
+        // not a red-black tree.
```
chrisdunelm: Both are balanced binary trees, so it's fine.

jskeet: (Will remove the comment at a later date :)

PR #2106, opened 2018-04-24
Add a Bigtable common code project

jskeet:
Currently this only contains TableName; we might want to think about
anything else that's likely to become common. (Moving things into
the common project now is okay; we can't do it post-GA.)

Next step will be to update the configuration to avoid generating
TableName or ProjectName, but to refer to TableName in this project
and ProjectName in GAX.


jskeet:
Updated to Google.Cloud.Bigtable.Common.V2

PR #2108, opened 2018-04-25
Support Parqet file format

jskeet:
- UploadParquet operation with corresponding options
- Allow Parquet file format when creating a load job from GCS

(This isn't as big a PR as it looks - most of it is either generated or very simple.)

The docs build fine, and I've run the *new* tests specifically, but I'll rerun integration and snippet tests.


amanda-tarafa:
Thanks! Will work on it soon.

jskeet:
(Sample updated; this will be ready to merge when reviewed.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
```diff
@@ -219,5 +221,23 @@ public void InsertRow_RecordRepeatedField()
                 { "score", score },
                 { "gameStarted", gameStarted }
             };
+
+        [Fact]
+        public void CreateLoadJob_Parquet()
+        {
+            string sourceUri = "gs://cloud-samples-data/bigquery/us-states/us-states.parquet";
```
amanda-tarafa: nit-question: I think you are following different guidelines to the ones we follow on the extension? regarding the use of `var`. We use `var` only when the type is evident, so this would be a `var` and some of the following `var`s would be replaced by the actual types.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Snippets/BigQueryClientSnippets.cs
```diff
@@ -528,6 +528,40 @@ public void UploadAvro()
...
+        {
+            string projectId = _fixture.ProjectId;
+            string datasetId = _fixture.GameDatasetId;
+            string tableId = _fixture.GenerateTableId();
+
+            var typeInfo = GetType().GetTypeInfo();
+            string resourceName = typeInfo.Namespace + ".us-states.parquet";
+            var stream = typeInfo.Assembly.GetManifestResourceStream(resourceName);
+            // Snippet: UploadParquet(string, string, Stream, *)
+            BigQueryClient client = BigQueryClient.Create(projectId);
```
amanda-tarafa: nit: And then here you are using the type because of this being part of the snippet I suppose?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Snippets/BigQueryClientSnippets.cs
```diff
@@ -528,6 +528,40 @@ public void UploadAvro()
...
+            AssertJobCompletesSuccessfully(job);
+        }
+
+        // See-also: UploadParquet(string, string, Stream, *)
+        // Member: UploadParquet(TableReference, *, *)
+        // Member: UploadParquet(string, string, string, *, *)
+        // See [UploadParquet](ref) for an example using an alternative overload.
+        // End see-also
+
+        // See-also: UploadParquet(string, string, *, *)
```
amanda-tarafa: question to understand: Is this adding a "See also ref" to each of the members listed below linking to `UploadParquet(string, string, Stream, *)` or the other way around? I think yes because it says "for a synchronous example" but just wanted to confirm.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadParquetOptions.cs
```diff
@@ -0,0 +1,87 @@
...
+        public TimePartitioning TimePartitioning { get; set; }
+
+        /// <summary>
+        /// The encryption configuration to apply to the destination table, if any.
+        /// </summary>
+        public EncryptionConfiguration DestinationEncryptionConfiguration { get; set; }
+
+        internal void ModifyConfiguration(JobConfigurationLoad loadRequest)
+        {
+            if (AllowUnknownFields != null)
```
amanda-tarafa: OK, I think this answers my previous question about defaults being set server side.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
jskeet: I use var much more liberally in test code than production code... but without any dogma around when it's valid or not. We don't use it in snippets though, as there we're really focused on users who may not know any of the types.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Snippets/BigQueryClientSnippets.cs
jskeet: Yup.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Snippets/BigQueryClientSnippets.cs
jskeet: If you have:

```csharp
// See-also: A
// Member: B
// Member: C
// Member: D
// See [A](ref) for an example.
```

then the docs for B, C and D will contain a link to A. The aim is to remove duplication (and to make it far more feasible to actually provide snippets for at least one method per operation).

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/UploadParquetOptionsTest.cs
```diff
@@ -0,0 +1,46 @@
...
+                WriteDisposition = WriteDisposition.WriteAppend,
+                DestinationEncryptionConfiguration = new EncryptionConfiguration { KmsKeyName = "projects/1/locations/us/keyRings/1/cryptoKeys/1" },
+            };
+
+            JobConfigurationLoad config = new JobConfigurationLoad();
+            options.ModifyConfiguration(config);
+            Assert.Equal(true, config.IgnoreUnknownValues);
+            Assert.Equal("CREATE_IF_NEEDED", config.CreateDisposition);
+            Assert.Equal(10, config.MaxBadRecords);
+            Assert.Equal("WRITE_APPEND", config.WriteDisposition);
```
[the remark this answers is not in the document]
jskeet: Done.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadParquetOptions.cs
```diff
@@ -0,0 +1,87 @@
...
+    /// Options for <c>UploadParquet</c> operations.
+    /// </summary>
+    public sealed class UploadParquetOptions : JobCreationOptions
+    {
+        /// <summary>
+        /// Whether or not to accept rows with fields that are not specified in the schema,
+        /// ignoring the extra fields.
+        /// If not set, this is effectively false.
+        /// </summary>
+        public bool? AllowUnknownFields { get; set; }
```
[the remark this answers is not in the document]
jskeet: We make everything nullable. Early on, the idea was that you should always be able to merge options together, with null meaning "not set". We haven't actually used that anywhere in the end (in these libraries; we do it a lot with `CallSettings` in gRPC though) but we're maintaining the convention.

PR #2116, opened 2018-04-26
Regenerate DLP v2

jskeet:


jskeet:
(All generated apart from the version change; will self-merge on green and then tag the release.)

PR #2138, opened 2018-05-10
(Not for submission) Removing Firestore Options classes

jskeet:


jskeet:
Note: the TransactionOptions commit should have some argument validation for maxAttempts.

jskeet:
Closing now - we're not removing the options classes after all.

PR #2141, opened 2018-05-11
Update Storage version for beta release

jskeet:
This is still pending seeing integration tests pass on internal CI.
The release date in the history is set to Monday to avoid releasing on a
Friday/weekend.

Currently only a beta release as customer-managed encryption keys
are in beta.


jskeet:
(Will merge on Monday, then do the release.)

PR #2143, opened 2018-05-14
Regenerate all APIs

jskeet:
This takes in four changes:

- More documentation for client creation with credentials
- Some Redis resource changes
- Updating the default GAX to 2.4.0-beta02
- Using common resources where possible


chrisdunelm:
It looks like there's a manual change that's required in DataTransferServiceClientSnippets.cs
And it's odd that DataTransferServiceClientSnippets.g.cs is still failing. I thought this was fixed. I'm investigating.

jskeet:
Thanks, I'll fix up the manual snippets.

PR #2147, opened 2018-05-16
Add back the HttpContextAccessor to CloudTrace

iantalarico:
This was removed in PR #2096 as it was no longer used in most of the code but is still used to get trace headers.


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -116,6 +117,9 @@ public static IApplicationBuilder UseGoogleTrace(this IApplicationBuilder app)
 
             services.AddSingleton(tracerFactory);
 
+            // Only add the HttpContextAccessor if it's not already added.
```
evildour: Maybe add a comment about why it needs to be added (so it isn't accidentally removed again).

iantalarico: Good call. Done

PR #2157, opened 2018-05-18
Don't assume ordering of maps in test results.

iantalarico:
This is causing flaky tests.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
@@ -269,14 +269,11 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
                 var results = _polling.GetEntries(startTime, testId, 1, LogSeverity.Warning);
                 var entry = results.Single();
                 Assert.Equal(3, entry.Labels.Count);
-                var defaultLabel = entry.Labels.First();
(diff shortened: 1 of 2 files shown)


iantalarico:
@jskeet I updated the tests to use an "expected dictionary" where possible but the integration tests only assert the trace id is not null.  I think leaving that as is is easier then setting up a matcher.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
```diff
@@ -269,14 +269,11 @@ public class LoggingTest
                 var results = _polling.GetEntries(startTime, testId, 1, LogSeverity.Warning);
                 var entry = results.Single();
                 Assert.Equal(3, entry.Labels.Count);
-                var defaultLabel = entry.Labels.First();
-                Assert.Equal("some-key", defaultLabel.Key);
+                var defaultLabel = entry.Labels.Single(l => l.Key.Equals("some-key"));
```
jskeet: Alternative option:

```csharp
Assert.Equal("some-value", entry.Labels["some-key"]);
Assert.Equal("Hello, World!", entry.Labels["Foo"]);
Assert.NotEmpty(entry.Labels["Foo"]);
```

If that fails with a KeyNotFoundException, it's easy enough to see where from the stack trace.

iantalarico: I like this much better, thanks for the suggestion.

PR #2178, opened 2018-05-24
 Release of Google.Cloud.Diagnostics.* 1.0.1

iantalarico:
Catch all Exceptions in the TimedBuffer timer to avoid ever bringing down an application.

I ran all the integration and unit tests and all is good.  I'm going to have to release manually as I don't think the current deployment can handle the older scripts.


iantalarico:
Will do

PR #2183, opened 2018-05-24
Catch all Exceptions in the TimedBuffer timer to avoid ever bringing down an application.

iantalarico:


jskeet:
Not in a position to review tonight - @evildour maybe?

PR #2187, opened 2018-05-25
Refactoring EnumMap into two classes.

amanda-tarafa:
Refactoring EnumMap into two classes, one for flags enums and another one for non flags enum.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
@@ -29,6 +29,19 @@ namespace Google.Cloud.BigQuery.V2
         {
             Value = value;
         }
+
+        public static IEnumerable<KeyValuePair<T, string>> GetApiValueNamesIn<T>() where T : struct => 
+            from field in typeof(T).GetTypeInfo().DeclaredFields
+            where field.IsStatic
+            select CreateKeyValuePair<T>(field);
+
+        private static KeyValuePair<T, string> CreateKeyValuePair<T>(FieldInfo field) where T : struct
+        {
+            T value = (T)field.GetValue(null);
+            ApiValueAttribute nameAttribute = (ApiValueAttribute)field.GetCustomAttribute(typeof(ApiValueAttribute));
+            string name = nameAttribute?.Value ?? value.ToString().ToUpperInvariant();
+            return new KeyValuePair<T, string>(value, name);
+        }
     }
 
     /// <summary>
@@ -40,7 +53,7 @@ namespace Google.Cloud.BigQuery.V2
             EnumMap<T>.ToApiValue(value, paramName);
 
         internal static ISet<string> ToApiValues<T>(T value) where T : struct =>
-            EnumMap<T>.ToApiValues(value);
+            FlagsEnumMap<T>.ToApiValues(value);
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
```diff
@@ -49,32 +60,28 @@ internal static class EnumMap
...
             typeof(T).GetTypeInfo().IsDefined(typeof(FlagsAttribute))
-            ? new Dictionary<int, string>() : null;
+            ? null : new Dictionary<string, T>();
+        private static readonly Dictionary<T, string> s_valueToString = 
+            s_stringToValue == null ? null : new Dictionary<T, string>();
 
         static EnumMap()
         {
-            foreach (var field in typeof(T).GetTypeInfo().DeclaredFields.Where(f => f.IsStatic))
+            if(s_stringToValue == null)
```
jskeet: Rather than put the conditional in the initializer and then switch here, I'd check the flags attribute in this "if" condition, throw if necessary, and then assign to s_valueToString after the if statement. I think that'll make it simpler to read - basically everything necessary is in the initializer. (Ditto for the FlagsEnumMap)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
```diff
@@ -29,6 +29,17 @@ public ApiValueAttribute(string value)
         {
             Value = value;
         }
+
+        internal static void ForEachStaticField<T>(Action<T, string> execute)
```
jskeet: Rather than a ForEach method, why not write a method returning `IEnumerable<KeyValuePair<T, string>>`? (That could just use LINQ, by the looks of it.)

Then the methods below that use ForEach can just use ToDictionary instead.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
amanda-tarafa: Doing

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
amanda-tarafa: Oh yes, that's nice. Doing.

amanda-tarafa:
@jskeet Please take a look.
Addressed your comments, only thing instead of using `KeyValuePair` I created a specific `struct`, I think it might be a little cleaner that way.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
```diff
@@ -30,14 +30,26 @@ public ApiValueAttribute(string value)
             Value = value;
         }
 
-        internal static void ForEachStaticField<T>(Action<T, string> execute)
+        public static IEnumerable<ApiValueName<T>> GetApiValueNamesIn<T>() where T : struct =>
```
jskeet: I'd make this and the struct (if we keep it) both internal. Even though the class is internal anyway, it's easier to observe that it's internal if you don't need to look at the containing class declaration.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
```diff
@@ -30,14 +30,26 @@ public ApiValueAttribute(string value)
             Value = value;
         }
 
-        internal static void ForEachStaticField<T>(Action<T, string> execute)
+        public static IEnumerable<ApiValueName<T>> GetApiValueNamesIn<T>() where T : struct => 
+            from field in typeof(T).GetTypeInfo().DeclaredFields
+            where field.IsStatic
+            select ApiValueName<T>.Create(field);
+
+        public struct ApiValueName<T> where T : struct
```
jskeet: I'm not sure this adds much, to be honest - the code you had before looked pretty simple, and we can just use `KeyValuePair` for name and value. `IntValue` is only used in one place, so that cast can be there.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
amanda-tarafa: OK

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
amanda-tarafa: Rationale behind the struct:
I didn't like a lot the idea of using KeyValuePair given that
a) there's no specific ToDictionary() method for IEnumerable<KeyValuePair> that doesn't require the use of selectors (unless I'm missing an extension import)
b) our three dictionariaes have different key value pairs so the returned IEnumerable<KeyValuePair> can only match one of the dictionaries, which is confusing, i.e. receiving one KeyValuePair to turn it into another. (Unless we passed a function to build the KeyValuePair but then, because of a) we have to use selectors anyway, which seems redundant).
So I though of using a Tuple instead of a KeyValuePair. That sort of lead to the struct so as to encapsulate Creation and IntValue.
But, my heart is not set on the Tuple vs. struct question, I see that this is simple enough and just locally used so as to not require a specific struct. But what I really don't like is using KeyValuePair, I think is confusing and/or redundant (again unless I'm missing a ToDictionary() extension method that takes no selectors).
Happy to chat later.

amanda-tarafa:
@jskeet PTAL, thanks!

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/EnumMap.cs
```diff
@@ -49,32 +62,24 @@ internal static class EnumMap
...
-        private static readonly Dictionary<int, string> s_intValueToString =
-            typeof(T).GetTypeInfo().IsDefined(typeof(FlagsAttribute))
-            ? new Dictionary<int, string>() : null;
+        private static readonly Dictionary<string, T> s_stringToValue;
+        private static readonly Dictionary<T, string> s_valueToString;
 
         static EnumMap()
         {
-            foreach (var field in typeof(T).GetTypeInfo().DeclaredFields.Where(f => f.IsStatic))
+            if(typeof(T).GetTypeInfo().IsDefined(typeof(FlagsAttribute)))
```
jskeet: Tiny nit: space after "if".

amanda-tarafa: Done

PR #2190, opened 2018-05-29
[Bigtable] Expose AppProfileId from the BigtableClient.

### LINKED ISSUE #1921: [Bigtable] Decide how to expose AppProfileIds
The AppProfileIds are currently an alpha feature for Bigtable. Make sure we're not locked into anything that may change before we go GA. This applies to both `BigtableClient.Create[Async]` and `BigtableClient.WithAppProfileId`.

evildour:
Resolves #1921


jskeet:
Not working today - will look tomorrow.

evildour:
That's fine. No rush.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -120,6 +120,11 @@ public abstract partial class BigtableClient
                 appProfileId,
                 underlyingClientSettings: null);
 
+        /// <summary>
+        /// Gets the value which specifies routing for replication. If null, the "default" application profile will be used.
```
jskeet: Possibly remove the quotes around default? Otherwise it might be confused for a string which is returned if the profile ID isn't set.

evildour: Actually, I think I'll just end this sentence (and the others) with "...by the server." That should make it pretty clear.

PR #2202, opened 2018-05-31
(Not for submission) Interceptor demo

jskeet:
This is based on #2201 - only the last commit is new.

### DIFF
--- a/apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1/LanguageServiceClient.cs
+++ b/apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1/LanguageServiceClient.cs
@@ -58,6 +58,11 @@ namespace Google.Cloud.Language.V1
             OnCopy(existing);
         }
 
+        /// <summary>
+        /// This will be in GAX...
+        /// </summary>
+        public grpccore::Interceptors.Interceptor Interceptor { get; set; }
+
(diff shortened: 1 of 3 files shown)


jskeet:
(Ignore the solution file change, too :)

apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1/LanguageServiceClient.cs
```diff
@@ -58,6 +58,11 @@ private LanguageServiceSettings(LanguageServiceSettings existing) : base(existin
             OnCopy(existing);
         }
 
+        /// <summary>
+        /// This will be in GAX...
+        /// </summary>
+        public grpccore::Interceptors.Interceptor Interceptor { get; set; }
```
AnashOommen: If the setter is public here, does it mean that a user can change the interceptor after using Create(channel, settings);

jskeet: Yes, but it won't have any impact. That's true of most (all?) service-wide settings.

chrisdunelm:
LGTM, this is exactly what I was expecting :)

jskeet:
Closing now this has been implemented (but not yet merged).

PR #2203, opened 2018-05-31
[Bigtable] Remove unnecessary async/await

evildour:


evildour:
It might. How would that cause a difference though? It was being awaited here and then the caller was awaiting `EstablishStream`. How would it be different if just the latter were taking place?

jskeet:
(On a phone now, so can't check the code) It's the difference between the method throwing directly or returning a faulted task. It may not matter here - can't tell right now.

evildour:
I don't think it will matter since both callers await it immediately anyway, so either the throw would occur here or at the call site.

jskeet:
Makes sense. May affect the stack trace but that's all.

evildour:
Right, thanks

PR #2204, opened 2018-05-31
[Bigtable] Tests for smart retries exceeding overall expiration

evildour:


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingMutateRowsTest.cs
```diff
@@ -183,6 +186,88 @@ public class RetryingMutateRowsTest
...
+                },
+                entriesForRetryStreams: new[]
+                {
+                    // 1st retry response entries
+                    new[] { Utilities.CreateMutateRowsResponseEntry(0, Code.Ok) }
+                },
+                settings: settings);
+
+            bool exceptionOccurred = false;
+            try
```
jskeet: You could use

```csharp
var exception = await Assert.ThrowsAsync<RpcException>(() => client.MutateRowsAsync(request));
Assert.Equal(StatusCode.DeadlineExceeded, ex.StatusCode);
```

I *think* that would be a bit simpler to read, but I don't have much of a preference.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingMutateRowsTest.cs
```diff
@@ -183,6 +186,88 @@ public class RetryingMutateRowsTest
...
+            {
+                await client.MutateRowsAsync(request);
+            }
+            catch (RpcException ex) when (ex.StatusCode == StatusCode.DeadlineExceeded)
+            {
+                exceptionOccurred = true;
+            }
+            Assert.False(exceptionOccurred);
```
jskeet: Can we easily check that the clock is where we'd expect it to be?

jskeet: Or is there not actually a clock involved? I don't see one in the changes - is that already taken care of elsewhere?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingMutateRowsTest.cs
```diff
@@ -183,6 +186,88 @@ public class RetryingMutateRowsTest
...
         }
 
-        // TODO: Test with a mock IClock to make sure we don't go over the total timeout when retrying
+        [Fact]
+        public async Task RetryAfterTotalExpiration()
+        {
+            var settings = new BigtableServiceApiSettings();
+            // Don't allow for any time to retry.
+            settings.MutateRowsRetrySettings =
+                settings.MutateRowsRetrySettings.WithTotalExpiration(
```
jskeet: With a total expiration of 0, I'm not sure I'd expect it to make any requests at all. Would it be better to have a total expiration that is "over" by the time the first response comes back?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingMutateRowsTest.cs
evildour: Thanks, I didn't realize you could get back the thrown exception from Assert.ThrowsAsync. This is much cleaner.

evildour:
Let me know if you need more info than what I've written in the line comments.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.Tests/RetryingMutateRowsTest.cs
evildour: Yeah, I think we could do more in the future with mock clocks, and this is slightly hacky, but I didn't want to complicate the `Utilities.Create...MockClient` logic any more than necessary for now.

jskeet:
All good now - merge when you're ready :)

evildour:
Thanks!

PR #2216, opened 2018-06-04
Pubsub unary acks/extends

chrisdunelm:
No longer using streaming push to ack/nack and extend messages.

Also a fairly major rewrite of the per-channel code.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/ReQueueTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/ReQueueTest.cs
@@ -1,7 +1,18 @@
-﻿using System;
-using System.Collections.Generic;
+﻿// Copyright 2017, Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 using System.Linq;
-using System.Text;
 using Xunit;
 
 namespace Google.Cloud.PubSub.V1.Tests
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/ReQueueTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/ReQueueTest.cs
@@ -1,0 +1,68 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Xunit;
+
+namespace Google.Cloud.PubSub.V1.Tests
+{
+    public class ReQueueTest
+    {
+        [Theory, CombinatorialData]
+        public void EnqueueDequeue([CombinatorialValues(0, 1, 2, 3, 4, 5)] int dequeueCount)
+        {
+            var q = new SubscriberClientImpl.ReQueue<string>();
+            q.Enqueue("a");
+            q.Enqueue(new[] { "b", "c" });
+            var dequeued = q.Dequeue(dequeueCount, null);
+            var expected = new[] { "a", "b", "c" }.Take(dequeueCount);
+            Assert.Equal(expected, dequeued);
+        }
+
+        [Theory, CombinatorialData]
+        public void RequeueDequeue([CombinatorialValues(0, 1, 2, 3, 4, 5)] int dequeueCount)
+        {
+            var q = new SubscriberClientImpl.ReQueue<string>();
+            q.Requeue(new[] { "a" });
+            q.Requeue(new[] { "b", "c" });
+            var dequeued = q.Dequeue(dequeueCount, null);
+            var expected = new[] { "a", "b", "c" }.Take(dequeueCount);
+            Assert.Equal(expected, dequeued);
+        }
+
+        [Theory, CombinatorialData]
+        public void EnqueueRequeueDequeue([CombinatorialValues(0, 1, 2, 3, 4, 5, 6, 7)] int dequeueCount)
+        {
+            var q = new SubscriberClientImpl.ReQueue<string>();
+            q.Enqueue("d");
+            q.Requeue(new[] { "a", "b" });
+            q.Enqueue(new[] { "e", "f" });
+            q.Requeue(new[] { "c" });
+            var dequeued = q.Dequeue(dequeueCount, null);
+            var expected = new[] { "a", "b", "c", "d", "e", "f" }.Take(dequeueCount);
+            Assert.Equal(expected, dequeued);
+        }
+
+        [Fact]
+        public void Count()
+        {
+            var q = new SubscriberClientImpl.ReQueue<string>();
+            Assert.Equal(0, q.Count);
+            q.Enqueue("a");
+            Assert.Equal(1, q.Count);
+            q.Requeue(new[] { "b" });
+            Assert.Equal(2, q.Count);
+            q.Requeue(new[] { "c", "d" });
+            Assert.Equal(4, q.Count);
+            q.Dequeue(1, null);
+            Assert.Equal(3, q.Count);
+            q.Dequeue(2, null);
+            Assert.Equal(1, q.Count);
+            q.Enqueue("e");
+            q.Requeue(new[] { "f" });
+            Assert.Equal(3, q.Count);
+            q.Dequeue(10, null);
+            Assert.Equal(0, q.Count);
+        }
+    }
+}
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
@@ -174,7 +174,7 @@ namespace Google.Cloud.PubSub.V1.Tests
                 {
                     if (message.ModifyDeadlineAckIds.Count != 0 || message.AckIds.Count != 0)
                     {
-                        throw new InvalidOperationException("WriteAsync must not modify deadlines.");
+                        throw new InvalidOperationException("WriteAsync must not modify deadlines or send acks/nacks.");
                     }
                     // TODO: Record write time
                     return Task.FromResult(0);
@@ -255,14 +255,8 @@ namespace Google.Cloud.PubSub.V1.Tests
                 }
                 lock (_lock)
                 {
-                    if (ackDeadlineSeconds == 0)
-                    {
-                        _nacks.AddRange(ackIds.Select(id => new TimedId(_clock.GetCurrentDateTimeUtc(), id)));
-                    }
-                    else
-                    {
-                        _extends.AddRange(ackIds.Select(id => new TimedId(_clock.GetCurrentDateTimeUtc(), id)));
-                    }
+                    var ids = ackDeadlineSeconds == 0 ? _nacks : _extends;
+                    ids.AddRange(ackIds.Select(id => new TimedId(_clock.GetCurrentDateTimeUtc(), id)));
                 }
             }
         }
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
@@ -149,7 +149,7 @@ namespace Google.Cloud.PubSub.V1.Tests
             private class PullStream : StreamingPullStream
             {
                 public PullStream(TimeSpan writeAsyncPreDelay,
-                    IEnumerable<ServerAction> msgs, List<TimedId> extends, List<TimedId> acks, List<DateTime> writeCompletes,
+                    IEnumerable<ServerAction> msgs, List<DateTime> writeCompletes,
                     IScheduler scheduler, IClock clock, TaskHelper taskHelper, bool useMsgAsId, CancellationToken? ct)
                 {
                     _taskHelper = taskHelper;
@@ -157,8 +157,6 @@ namespace Google.Cloud.PubSub.V1.Tests
                     _writeAsyncPreDelay = writeAsyncPreDelay; // delay within the WriteAsync() method. Simulating network or server slowness.
                     _en = new En(msgs, scheduler, taskHelper, clock, useMsgAsId, ct);
                     _clock = clock;
-                    _extends = extends;
-                    _acks = acks;
                     _writeCompletes = writeCompletes;
                 }
 
@@ -168,21 +166,18 @@ namespace Google.Cloud.PubSub.V1.Tests
                 private readonly TimeSpan _writeAsyncPreDelay;
                 private readonly IAsyncEnumerator<StreamingPullResponse> _en;
                 private readonly IClock _clock;
-                private readonly List<TimedId> _extends;
-                private readonly List<TimedId> _acks;
                 private readonly List<DateTime> _writeCompletes;
 
                 public override IAsyncEnumerator<StreamingPullResponse> ResponseStream => _en;
 
-                public override async Task WriteAsync(StreamingPullRequest message)
+                public override Task WriteAsync(StreamingPullRequest message)
                 {
-                    await _taskHelper.ConfigureAwait(_scheduler.Delay(_writeAsyncPreDelay, CancellationToken.None));
-                    var now = _clock.GetCurrentDateTimeUtc();
-                    lock (_lock)
+                    if (message.ModifyDeadlineAckIds.Count != 0 || message.AckIds.Count != 0)
                     {
-                        _extends.AddRange(message.ModifyDeadlineAckIds.Select(id => new TimedId(now, id)).ToList());
-                        _acks.AddRange(message.AckIds.Select(id => new TimedId(now, id)).ToList());
+                        throw new InvalidOperationException("WriteAsync must not modify deadlines.");
                     }
+                    // TODO: Record write time
+                    return Task.FromResult(0);
                 }
 
                 public override Task WriteCompleteAsync()
@@ -190,8 +185,8 @@ namespace Google.Cloud.PubSub.V1.Tests
                     lock (_lock)
                     {
                         _writeCompletes.Add(_clock.GetCurrentDateTimeUtc());
-                        return Task.FromResult(0);
                     }
+                    return Task.FromResult(0);
                 }
             }
 
@@ -207,19 +202,22 @@ namespace Google.Cloud.PubSub.V1.Tests
                 _useMsgAsId = useMsgAsId;
             }
 
+            private readonly object _lock = new object();
             private readonly IEnumerator<IEnumerable<ServerAction>> _msgsEn;
             private readonly IScheduler _scheduler;
             private readonly IClock _clock;
             private readonly TaskHelper _taskHelper;
-            private readonly TimeSpan _writeAsyncPreDelay;
+            private readonly TimeSpan _writeAsyncPreDelay; // Simulates slow network or server
             private readonly bool _useMsgAsId;
 
             private readonly List<TimedId> _extends = new List<TimedId>();
             private readonly List<TimedId> _acks = new List<TimedId>();
+            private readonly List<TimedId> _nacks = new List<TimedId>();
             private readonly List<DateTime> _writeCompletes = new List<DateTime>();
 
             public IReadOnlyList<TimedId> Extends => _extends;
             public IReadOnlyList<TimedId> Acks => _acks;
+            public IReadOnlyList<TimedId> Nacks => _nacks;
             public IReadOnlyList<DateTime> WriteCompletes => _writeCompletes;
 
             public override StreamingPullStream StreamingPull(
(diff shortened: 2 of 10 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/ReQueueTest.cs
```diff
@@ -0,0 +1,79 @@
...
+            q.Enqueue(new[] { "b", "c" });
+            var dequeued = q.Dequeue(dequeueCount, null);
+            var expected = new[] { "a", "b", "c" }.Take(dequeueCount);
+            Assert.Equal(expected, dequeued);
+        }
+
+        [Theory, CombinatorialData]
+        public void RequeueDequeue([CombinatorialValues(0, 1, 2, 3, 4, 5)] int dequeueCount)
+        {
+            var q = new SubscriberClientImpl.ReQueue<string>();
```
jskeet: Is this a generated class? Really disliking the "class ReQueue, method Requeue" bit here...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -168,30 +166,27 @@ private class PullStream : StreamingPullStream
...
+                public override Task WriteAsync(StreamingPullRequest message)
                 {
-                    await _taskHelper.ConfigureAwait(_scheduler.Delay(_writeAsyncPreDelay, CancellationToken.None));
-                    var now = _clock.GetCurrentDateTimeUtc();
-                    lock (_lock)
+                    if (message.ModifyDeadlineAckIds.Count != 0 || message.AckIds.Count != 0)
                     {
-                        _extends.AddRange(message.ModifyDeadlineAckIds.Select(id => new TimedId(now, id)).ToList());
-                        _acks.AddRange(message.AckIds.Select(id => new TimedId(now, id)).ToList());
+                        throw new InvalidOperationException("WriteAsync must not modify deadlines.");
```
jskeet: Or send acks?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -149,16 +149,14 @@ public void Dispose()
             private class PullStream : StreamingPullStream
             {
                 public PullStream(TimeSpan writeAsyncPreDelay,
-                    IEnumerable<ServerAction> msgs, List<TimedId> extends, List<TimedId> acks, List<DateTime> writeCompletes,
+                    IEnumerable<ServerAction> msgs, List<DateTime> writeCompletes,
```
jskeet: So is this a list that we're passing in to collect values? Do we pass in a non-empty list? If not, could we create the list ourselves an expose it? (I suspect the order in which I'm reviewing the files isn't ideal.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -232,10 +230,41 @@ public override Task WriteCompleteAsync()
...
+
+            public override async Task ModifyAckDeadlineAsync(SubscriptionName subscription, IEnumerable<string> ackIds, int ackDeadlineSeconds, CancellationToken cancellationToken)
+            {
+                if (_writeAsyncPreDelay != TimeSpan.Zero)
+                {
+                    await _taskHelper.ConfigureAwait(_scheduler.Delay(_writeAsyncPreDelay, CancellationToken.None));
+                }
+                lock (_lock)
+                {
+                    if (ackDeadlineSeconds == 0)
```
jskeet: Maybe

```csharp
List<TimeId> ids = ackDeadlineSeconds == 0 ? _nacks : _extends;
ids.Add(ackIds.Select(id => new TimedId(_clock.GetCurrentDateTimeUtc(), id));
```

?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/Tasks/TestScheduler.cs
```diff
@@ -148,169 +148,213 @@ void ActionWrapper()
 
         private class TestTaskScheduler : TaskScheduler, IDisposable
         {
+            [ThreadStatic]
```
jskeet: I'm skipping reviewing this for now, in order to look at the prod code. Happy to come back to this later, but may need a chat first.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +539,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
                         _event.Set();
                     }
-                })));
+                }));
+                _registerTaskFn(task);
+            }
+        }
+
+        // internal for testing.
+        internal class ReQueue<T>
```
jskeet: Ah, this is the class from earlier. Let's think up a different name that doesn't involve the type and method name differing only be the case of a single letter.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -348,7 +348,14 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
             GaxPreconditions.CheckNotNull(settings, nameof(settings));
             settings.Validate();
             _modifyDeadlineSeconds = (int)((settings.StreamAckDeadline ?? DefaultStreamAckDeadline).TotalSeconds);
+            // Enforce server-side constraint on _modifyDeadlineSeconds.
+            GaxPreconditions.CheckArgument(_modifyDeadlineSeconds >= 10 && _modifyDeadlineSeconds <= 600, nameof(settings.StreamAckDeadline), "Must be between 10 and 600 seconds");
             _autoExtendInterval = TimeSpan.FromSeconds(_modifyDeadlineSeconds) - (settings.AckExtensionWindow ?? DefaultAckExtensionWindow);
+            if (_autoExtendInterval < TimeSpan.FromSeconds(5))
```
jskeet: I'm wondering whether some static methods in GAX would be handy at some point - MaxTimeSpan and MinTimeSpan.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -348,7 +348,14 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
             GaxPreconditions.CheckNotNull(settings, nameof(settings));
             settings.Validate();
             _modifyDeadlineSeconds = (int)((settings.StreamAckDeadline ?? DefaultStreamAckDeadline).TotalSeconds);
+            // Enforce server-side constraint on _modifyDeadlineSeconds.
+            GaxPreconditions.CheckArgument(_modifyDeadlineSeconds >= 10 && _modifyDeadlineSeconds <= 600, nameof(settings.StreamAckDeadline), "Must be between 10 and 600 seconds");
```
jskeet: How confident are we that this won't change?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -387,16 +394,16 @@ public override Task StartAsync(Func<PubsubMessage, CancellationToken, Task<Repl
             Action<Task> registerTask = task =>
             {
                 registeredTasks.Locked(() => registeredTasks.Add(task));
-                Action unregisterTask = () => registeredTasks.Locked(() => registeredTasks.Remove(task));
-                _taskHelper.Run(async () => await _taskHelper.ConfigureAwaitWithFinally(() => task, unregisterTask));
+                Action<Task> unregisterTask = t => registeredTasks.Locked(() => registeredTasks.Remove(t));
+                task.ContinueWith(unregisterTask, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, _taskHelper.TaskScheduler);
```
jskeet: Does it make sense to use ExecuteSynchronously and specify a task scheduler? Isn't the point of ExecuteSynchronously that it just keeps going on the same thread, with no scheduling needed? Maybe I've misunderstood.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
chrisdunelm: OK, renamed it "RequeueableQueue" for now. It's internal only so isn't desperately important.
Also, I like the word "Requeueable" because it has five vowels in a row!

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/ReQueueTest.cs
chrisdunelm: This was written and named somewhat incrementally, and I never noticed this :)
Now renamed, see later comment.

chrisdunelm:
PTAL

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -515,7 +516,9 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
                 }
                 // Execute fn, and schedule the following code to execute once it has completed.
                 // Register the function, so we can be sure it's completed during shutdown.
-                _registerTaskFn(_taskHelper.Run(async () => await _taskHelper.ConfigureAwaitWithFinally(fn, () =>
+                //_registerTaskFn(_taskHelper.Run(async () => await _taskHelper.ConfigureAwaitWithFinally(fn, () =>
+                Task task = null;
+                task = _taskHelper.Run(async () => await _taskHelper.ConfigureAwaitWithFinally(fn, () =>
```
jskeet: You're not using `task` within the lambda, so these two lines can be combined, I think.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -515,7 +516,9 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
                 }
                 // Execute fn, and schedule the following code to execute once it has completed.
                 // Register the function, so we can be sure it's completed during shutdown.
-                _registerTaskFn(_taskHelper.Run(async () => await _taskHelper.ConfigureAwaitWithFinally(fn, () =>
+                //_registerTaskFn(_taskHelper.Run(async () => await _taskHelper.ConfigureAwaitWithFinally(fn, () =>
```
jskeet: Remove the commented out line?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+
+        // internal for testing.
+        internal class RequeueableQueue<T>
+        {
+            private readonly Queue<T> _q = new Queue<T>();
+            private readonly LinkedList<Queue<T>> _qs = new LinkedList<Queue<T>>();
+            private int _requeueCount = 0;
+
+            public void Enqueue(T item) => _q.Enqueue(item);
+            public void Enqueue(IEnumerable<T> items)
```
jskeet: Nit: add blank line before and after this method.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
                         _event.Set();
                     }
-                })));
+                }));
+                _registerTaskFn(task);
+            }
+        }
+
+        // internal for testing.
+        internal class RequeueableQueue<T>
```
jskeet: A doc comment would be helpful :) So it looks like there's the queued data, and the requeued data. If you requeue more data, it comes after all previously-requeued data, but before queued data. Is that right?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+                Queue<T> q = _qs.First?.Value;
+                if (q != null)
+                {
+                    while (result.Count < maxCount)
+                    {
+                        if (q.Count > 0)
+                        {
+                            var item = q.Dequeue();
+                            _requeueCount -= 1;
+                            if (includeFn == null || includeFn(item))
```
jskeet: So any not-included data is discarded rather than preserved, right?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+            public void Requeue(IEnumerable<T> items)
+            {
+                var q = new Queue<T>(items);
+                _qs.AddLast(q);
+                _requeueCount += q.Count;
+            }
+
+            public int Count => _q.Count + _requeueCount;
+
+            public List<T> Dequeue(int maxCount, Predicate<T> includeFn)
```
jskeet: It feels like this could probably be refactored into a private "DequeueOne" method and then a while loop that calls it. I'm okay with it as-is for the moment, but I think it could be simpler. Maybe try that later on?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+                }
+                if (tcs != null)
+                {
+                    // Don't run in lock, as it may execute continuations synchonously.
+                    tcs.SetResult(0);
+                }
+            }
+
+            // Thread-safe, but only one dequeue is allowed at a time.
+            public async Task<T> DequeueAsync()
```
jskeet: I don't think this is safe. If two threads call DequeueAsync at the same time, they could each create and assign a separate TCS, one of which would get lost. (And an Enqueue call at the same time could set the result on the first one, so potentially neither thread sees the new item!). Happy to VC to explain what I mean. Perhaps ConcurrentQueue is useful here?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+                while (!IsComplete())
+                {
+                    // Wait for, then process next continuation.
+                    TaskNextAction nextContinuation = await _taskHelper.ConfigureAwait(_continuationQueue.DequeueAsync());
+                    // On hardstop just immediately stop this event loop.
+                    // The registered-task code ensures that all currently-active tasks finish before
+                    // return to user code.
+                    if (_hardStopCts.IsCancellationRequested)
+                    {
+                        StopStreamingPull();
```
jskeet: Should this be in a finally block?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+
+            private bool HandleRpcFailure(Exception e)
+            {
+                if (e != null)
+                {
+                    if (e.As<RpcException>()?.IsRecoverable() ?? false)
+                    {
+                        // Recoverable RPC error, stop and restart pull.
+                        StopStreamingPull();
+                        // Increase backoff internal and start stream again.
```
jskeet: You could consider using the GAX `BackoffSettings`  and the new `NextDelay` method, starting off with a delay of 0 and a minimum of 0.5s.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+                // Check if the init write failed.
+                if (initTask != null && HandleRpcFailure(initTask.Exception))
+                {
+                    return;
+                }
+                // Check if pulls need throttling due to push queues being too full, or too slow to push.
+                bool throttle = _extendQueue.TryPeek(out var qItem) && _extendThrottleLow >= qItem.Time;
+                if (!throttle)
+                {
+                    int totalQueueCount = _pushInFlight + _extendQueue.Count;
```
jskeet: Why do these not need to be accessed in the lock?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+                // Running async. Common data needs locking
+                for (int msgIndex = 0; msgIndex < msgs.Count; msgIndex++)
                 {
-                    _extendQueue.EnqueueAll(ackIds);
+                    _softStopCts.Token.ThrowIfCancellationRequested();
+                    var msg = msgs[msgIndex];
+                    msgs[msgIndex] = null;
+                    await _taskHelper.ConfigureAwait(_flow.Process(msg.CalculateSize(), async () =>
+                    {
+                            // Running async. Common data needs locking
```
jskeet: Indentation within this block has gone a bit screwy.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+                            // Lock msgsIds, this is accessed concurrently here and in HandleExtendLease().
+                            lock (msgIds)
+                        {
+                            msgIds.Remove(msg.AckId);
+                        }
+                            // Lock ack/nack-queues, this is accessed concurrently here and in "master" thread.
+                            lock (_lock)
+                        {
+                            _userHandlerInFlight -= 1;
+                            (reply == Reply.Ack ? _ackQueue : _nackQueue).Enqueue(msg.AckId);
```
jskeet: Extract the conditional expression to a local variable?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+                    anyMsgIds = msgIds.Count > 0;
+                    if (anyMsgIds)
                     {
-                        bool isCancelled = await _taskHelper.ConfigureAwaitHideCancellation(
-                            () => _scheduler.Delay(_autoExtendInterval, allMsgsHandled.Token));
-                        if (isCancelled)
+                        lock (_lock)
                         {
-                            return;
+                            // Only enqueue ack IDs that are not already being acked.
```
jskeet: I don't understand what the comment has to do with the code.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -530,297 +533,605 @@ public Flow(long maxByteCount, long maxElementCount, Action<Task> registerTaskFn
...
+                    // Too many existing concurrent pushes; do nothing.
+                    return;
+                }
+                List<string> acks;
+                List<TimedId> extends;
+                List<string> nacks;
+                lock (_lock)
+                {
+                    // Priority of sending: Acks, Extends, Nacks.
+                    // Allow 2 over _maxConcurrentPush, otherwise extend/nack queues could be unfairly penalized.
```
jskeet: It's not clear what this comment line has to do with the code after it.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
chrisdunelm: Because they are not used concurrently at all. They are only used within the continuation-based state-machine.
_ack/_nack queues are used concurrently, so are locked.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
chrisdunelm: Oops, done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
chrisdunelm: OK, done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
chrisdunelm: Erroneous comment, from previous code iteration. Removed.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
chrisdunelm: Improved comment.

chrisdunelm:
PTAL

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
jskeet: Maybe it's worth adding comments for the fields, to make correct usage clear?

chrisdunelm: There is a comment on _lock stating what needs locking. Is that sufficient?

jskeet: Sure :)

PR #2221, opened 2018-06-06
Added LogBuilderExtension for ASP.NET Core Diagnostic project

imperugo:
ASP.NET Core 2.x allows you to configure the logging, but also DI, directly into the [WebHostBuilder](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.hosting.webhostbuilder?view=aspnetcore-2.1) besides that into the "classic" `Startup` class.

This is very helpful because you can have the logging system up & running before the application startup. Unfortunately you can't do that in the same way you do into the startup class (using `ILoggerFactory`) but you have to use the `ILoggingBuilder` interface ([here](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.iloggingbuilder?view=aspnetcore-2.1) more info) and register the `ILoggerProvider`.

I've added an extension method that makes this easier, below an example:

```csharp
new WebHostBuilder()
.UseKestrel()
.ConfigureServices(services =>
{
	{
		options.ProjectId = "[Google Cloud Platform project ID]";
		options.ServiceName = "My Super ServiceName";
	});

	// Add trace service.
	services.AddGoogleTrace(options =>
	{
		options.ProjectId = "my-project-id";
		options.Options = TraceOptions.Create(bufferOptions: BufferOptions.NoBuffer());
	});
})
.ConfigureLogging((hostingContext, logging) =>
{
	string projectId = "[Google Cloud Platform project ID]";
	var serviceProvider = logging.Services.BuildServiceProvider();
	logging.AddGoogle(serviceProvider);
})
.UseIISIntegration()
.UseContentRoot(Directory.GetCurrentDirectory())
.UseStartup<Startup>()
.Build()
.Run();
```

Note:
Unfortunately the method `GoogleLoggerProvider.Create` requires an instance of `IServiceProvider` (the nuget packaged doesn't need it, probably the master is a different version) so the developer must call the extension method `BuildServiceProvider` before to use `logging.AddGoogle()`.

I see two possible ways to resolve this:

* Remove the mandatory `IServiceProvider` from `GoogleLoggerProvider.Create` 
* Add a dependency into the project (`Microsoft.Extensions.DependencyInjection`) and call `BuildServiceProvider` for the user (in case of null of course)

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogBuilderExtensions.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogBuilderExtensions.cs
@@ -1,0 +1,62 @@
+﻿#if NETSTANDARD2_0
+using System;
+using Google.Api.Gax;
+using Google.Cloud.Diagnostics.Common;
+using Google.Cloud.Logging.V2;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Google.Cloud.Diagnostics.AspNetCore.Logging
+{
+	/// <summary>
+	/// Extensions to add <see cref="GoogleLoggerProvider"/>s to an <see cref="ILoggingBuilder"/>.
+	/// </summary>
+	/// 
+	/// <example>
+	/// <code>
+	/// .ConfigureLogging((hostingContext, logging) =>
+	/// {
+	///     string projectId = "[Google Cloud Platform project ID]";
+	///     var serviceProvider = logging.Services.BuildServiceProvider();
+	///     logging.AddGoogle(serviceProvider);
+	///     ...
+	/// }
+	/// </code>
+	/// </example>
+	/// 
+	/// <remarks>
+	/// Logs to Google Stackdriver Cloud Logging.
+	/// Docs: https://cloud.google.com/logging/docs/
+	/// </remarks>
+	/// <seealso cref="GoogleLogger"/>
+	public static class GoogleLogBuilderExtensions
+    {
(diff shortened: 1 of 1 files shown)


jskeet:
Amanda and Ian, you both know more about this in terms of initialization and ASP.NET Core diagnostics... I'll welcome your comments :)

amanda-tarafa:
Hi @imperugo, two things here.

There's a way to configure Logging in Asp.Net Core 2.x directly into the WebHostBuilder that doesn't require the creation of a ServiceProvider from the client code:
```
string projectId = "[Google Cloud Platform project ID]";
var webHost = new WebHostBuilder()
    .ConfigureServices(services =>
    {
        services.AddSingleton<ILoggerProvider>(sp => GoogleLoggerProvider.Create(sp, projectId));
    })
    .UseStartup<Startup>()
    .Build();
```
But also, in your example above you are activating all the services that our Diagnostics library offers: ErrorReporting, Trace and Logging. There's a compact way to do that by simply activating the whole of the Diagnostics services:
```
string projectId = "[Google Cloud Platform project ID]";
var webHost = new WebHostBuilder()
    .UseGoogleDiagnostics(projectId)
    .UseStartup<Startup>()
    .Build();
```
You can see this examples and more in these [Google.Cloud.Diagnostics.AspNetCore samples](https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.Diagnostics.AspNetCore/index.html)

Let me know if this works for you and solves your issues.

imperugo:
Hi @amanda-tarafa 
thanks for the quick response.
Probably there is something that I don't understand. Part of the things you are mentioning, `UseGoogleDiagnostics` for example, are not available on the current nuget package.

I was using the stable version that doesn't include some methods. Just switched to the beta06 that includes all I need.

I think we can close this PR.

PR #2223, opened 2018-06-06
Update codegen to use gRPC 1.12

jskeet:
Second commit is just generated based on the changes in the first.


jskeet:
(I'll wait for green before merging :)

PR #2230, opened 2018-06-06
Remove public constructor from SpannerCommandTextBuilder

jskeet:
Now this can only be constructed via factory methods, which is
pleasantly consistent.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
@@ -24,7 +24,7 @@ namespace Google.Cloud.Spanner.Data
 {
     /// <summary>
     /// Builds the <see cref="SpannerCommand.CommandText"/> string for executing a query or operation
-    /// on a Spanner database.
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -245,13 +191,59 @@ internal string DatabaseToDrop
...
+                        throw new InvalidOperationException(
+                            $"Spanner {commandName} commands are specified as '{commandName} <table>' with " +
+                            "parameters added to customize the command with filtering or updated values.");
+                    }
+                    targetTable = ValidateTableName(commandSections[1], nameof(commandText));
+                    // This just normalizes the command text.
+                    commandText = $"{commandName} {targetTable}";
+                    break;
+            }
+            return new SpannerCommandTextBuilder(commandText: commandText, type: commandType, targetTable: targetTable, extraStatements: null);
```
chrisdunelm: nit: not sure it's worth giving the param-names for all args here.

jskeet: I might reduce it to just the null then.

PR #2232, opened 2018-06-06
Including the client lib version as part of the generated documentation.

amanda-tarafa:
Adding the client lib version on the generated docs.
It looks like this for `Diagnostics.AspNetCore`:
image
One thing, this is relaying on `apis/apis.json` for the version and I don't know if we always maintain that file up to date, it seems so.

### DIFF
--- a/tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
+++ b/tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
@@ -179,6 +179,10 @@ namespace Google.Cloud.Tools.GenerateDocfxSources
         {
             string title = $"# {api.Id}";
             string description = $"`{api.Id}` is a.NET client library for the [{api.ProductName} API]({api.ProductUrl}).";
+            string version =
+$@"Note:
+This documentation is for version `{ api.Version}` of the library.
+Some samples may not work with other versions.";
             string installation =
 $@"# Installation
 
@@ -187,10 +191,11 @@ your project in the normal way (for example by right-clicking on the
 project in Visual Studio and choosing ""Manage NuGet Packages..."").";
             if (!api.IsReleaseVersion)
             {
-                installation += @"
+                installation += $@"
 Please ensure you enable pre-release packages (for example, in the
 Visual Studio NuGet user interface, check the ""Include prerelease""
-box).";
+box). Some of the following samples might only work with the latest 
+pre-release version (`{api.Version}`) of `{api.Id}`.";
             }
 
             string auth =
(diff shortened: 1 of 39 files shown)


amanda-tarafa:
@jskeet PTAL, thanks!

tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
```diff
@@ -177,7 +177,13 @@ private static void CopyAndGenerateArticles(ApiMetadata api, string inputDirecto
         /// </summary>
         private static string TransformDocTemplate(ApiMetadata api, string text)
         {
-            string title = $"# {api.Id}";
+            string title =
```
jskeet: I think it would be better to have this either as a new "tag" or as part of the description. (I'd put it below the main description rather than before it.) It doesn't make as much sense to have it as part of the "title" tag.

tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
```diff
@@ -177,7 +177,13 @@ private static void CopyAndGenerateArticles(ApiMetadata api, string inputDirecto
         /// </summary>
         private static string TransformDocTemplate(ApiMetadata api, string text)
         {
-            string title = $"# {api.Id}";
+            string title = 
+$@"# {api.Id}
+
+About the version of these documents:
```
jskeet: I think just "Note:" would be simpler here. I'd make the rest a little more concise too:

```text
Note: This documentation is for version {api.Version} of the library.
Some samples may not work with other versions.
```

amanda-tarafa: OK

tools/Google.Cloud.Tools.GenerateDocfxSources/Program.cs
amanda-tarafa: But then we would have to add the new tag on all the index.mds manually right? Well I can do that it's not that much anyways.

jskeet: Not necessarily manually. I'd suggest:

- Updating the template in the project generator
- Deleting all the index.md files
- Regenerating the projects
- Checking each change carefully. Only the ones with manually-written doc files would need changing at that point.

amanda-tarafa: Oh I see, let me try that.

amanda-tarafa:
@jskeet Just a question, there are several files like this:

```
{{title}}

{{description}}

{{version}}

{{installation}}

{{auth}}

# Getting started

{{client-classes}}

{{client-construction}}

```
They don't have code samples but I decided to leave the `{{version}}` section in anyways because on the `{{client-classes}}` and `{{client-construction}}` sections there is mention of code that can change with versions. Is that OK?

jskeet:
Yes, it's absolutely worth including `{{ version }}` in there, because it makes it clear which version of the documentation the user is reading - which is relevant if they look at the API reference.

PR #2242, opened 2018-06-07
More Spanner investigation

jskeet:
This enables the three parts of the tool (database creation, table
creation, row insertion) to be independently configured to be
synchronous or asynchronous.


jskeet:
Thinking about this more, I could probably make a single method in each case which optionally runs asynchronously (but is always an async method). Preferences?

chrisdunelm:
Leave it as it is for now. It's simple to change later if it becomes easier to do so.

PR #2245, opened 2018-06-07
Adding labels to all job creation operation options.

amanda-tarafa:
A couple of notes for this PR:

- Labels have not been added as filters for the list jobs operation.
- The list jobs operation only populates job#configuration (and hence job#configuration.labels) if projection is set to full on the request parameters.
- We haven't yet surfaced projection as a parameter for list jobs operation, so we cannot even use the labels to filter the jobs client side.
- I didn't write snippets for labels because as long as we cannot filter there's no much use we can show, snippets would be only of setting labels.
- Is it OK if I surface projection, seems simple enough, anything I might have missed. (There's a TODO to do so on the list jobs options class).
- I'm adding an inline note on a piece of code I changed that has nothing to do with labels.
- Marking this PR as not for submission until I surface projection and can write a snippet.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Cloud.Storage.V1;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using Xunit;
 using static Google.Apis.Bigquery.v2.JobsResource.ListRequest;
@@ -22,6 +24,18 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
     [Collection(nameof(BigQueryFixture))]
     public class JobsTest
     {
+        internal static IDictionary<string, string> JobLabels =>
+            new Dictionary<string, string>()
+            {
+                // This one is a little absurd for jobs because labels can only be set
+                // on job creation. This means not to set the label because this is copying
+                // the behaviour of labels for other resources like DataSets and Table.
+                // Testing for it anyways because the api allows is.
+                { "one_label", null },
+                { "another-label-2", "label_value_2" },
+                { "yet_another_label", "" }
+            };
+
         private readonly BigQueryFixture _fixture;
 
         public JobsTest(BigQueryFixture fixture)
@@ -34,18 +48,25 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
         private long ToMillisecondsSinceEpoch(DateTime dateTime) =>
             (long) (dateTime - new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc)).TotalMilliseconds;
 
+        /// <summary>
+        /// There's an equivalent snippet for this integration test but, where the snippet
+        /// tests for ListJobs listing some jobs, this test tests for ListJobs listing an
+        /// specific job that was just created.
+        /// </summary>
         [Fact]
         public void ListJobs()
         {
             var client = BigQueryClient.Create(_fixture.ProjectId);
 
             var table = client.GetTable(_fixture.ProjectId, _fixture.DatasetId, _fixture.HighScoreTableId);
-            var jobToFind = client.CreateQueryJob("SELECT * FROM {table}", parameters: null);
+            var jobToFind = client.CreateQueryJob($"SELECT * FROM {table}", parameters: null);
 
             var oneMinuteAgo = ToMillisecondsSinceEpoch(DateTime.UtcNow.AddMinutes(-1));
 
             // Find all jobs started in the last minute.
-            var recentJobs = client.ListJobs().TakeWhile(job => job.Statistics.CreationTime >= jobToFind.Statistics.CreationTime).ToList();
+            // Using TakeWhile because, as documented:
+            // The job list is sorted in reverse chronological order, by job creation time.
+            var recentJobs = client.ListJobs().TakeWhile(job => job.Statistics.CreationTime >= oneMinuteAgo).ToList();
 
             // Note: can't find the job reference itself, as that would check equality by reference :(
             var jobIds = recentJobs.Select(job => job.Reference.JobId).ToList();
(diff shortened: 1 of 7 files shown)


jskeet:
(Away at conferences at the moment, but will look when I can. I don't know whether filtering by label is even available, to be honest.)

amanda-tarafa:
No rush at all. And filtering by labels is not available, but we can at least make it possible on client side by surfacing projection.

amanda-tarafa:
@jskeet PTAL (no rush at all) thanks!
Added integration test and snippet for filtering listed jobs by labels.
Not sure about the place where I included the snippet at all but couldn't get to have either two snippets or even to see alsos per member.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
```diff
@@ -44,11 +46,200 @@ public void ListJobs()
             var oneMinuteAgo = ToMillisecondsSinceEpoch(DateTime.UtcNow.AddMinutes(-1));
 
             // Find all jobs started in the last minute.
-            var recentJobs = client.ListJobs().TakeWhile(job => job.Statistics.CreationTime >= jobToFind.Statistics.CreationTime).ToList();
+            var recentJobs = client.ListJobs().TakeWhile(job => job.Statistics.CreationTime >= oneMinuteAgo).ToList();
```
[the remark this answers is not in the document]
jskeet: Hmm. I'm surprised the compiler didn't point out that I wasn't using the variable. Thanks for that!

And yes, we are relying on the ordering, but it's documented:

> The job list is sorted in reverse chronological order, by job creation time. 

It wouldn't hurt to add that to a comment though :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
```diff
@@ -67,5 +69,210 @@ public void ListJobs_FullProjection()
...
+        [Fact]
+        public void ListJobs_JobLabels()
+        {
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+
+            var table = client.GetTable(_fixture.ProjectId, _fixture.DatasetId, _fixture.HighScoreTableId);
+            var jobToFind = client.CreateQueryJob($"SELECT * FROM {table}", null, new QueryOptions() { Labels = JobLabels});
+
+            // Find the job after listing with full projection.
+            var jobFound = (from job in client.ListJobs(new ListJobsOptions() { Projection = ProjectionEnum.Full })
```
jskeet: I'd write this without the query expression:

```csharp
var options = new ListJobsOptions { Projection = ProjectionEnum.Full };
var jobFound = client.ListJobs(options).Single(job => job.Reference.JobId == jobToFind.Reference.JobId);
```

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
```diff
@@ -67,5 +69,210 @@ public void ListJobs_FullProjection()
...
+
+            Assert.Equal(jobToFind.Reference.JobId, jobFound.Reference.JobId);
+        }
+
+        [Fact]
+        public void GetJob_JobLabels()
+        {
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+            var table = client.GetTable(_fixture.ProjectId, _fixture.DatasetId, _fixture.HighScoreTableId);
+            var options = new QueryOptions() { Labels = JobLabels };
```
jskeet: Just as a matter of style, I'd suggest omitting the `()` when calling a parameterless constructor with an object initializer. It doesn't make any difference here, but I think it's good to have a consistent style in the snippets, and the brevity is appealing :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
```diff
@@ -67,5 +69,210 @@ public void ListJobs_FullProjection()
...
             Assert.NotNull(jobFound.Resource.Configuration);
         }
+
+        [Fact]
+        public void ListJobs_JobLabels()
+        {
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+
+            var table = client.GetTable(_fixture.ProjectId, _fixture.DatasetId, _fixture.HighScoreTableId);
+            var jobToFind = client.CreateQueryJob($"SELECT * FROM {table}", null, new QueryOptions() { Labels = JobLabels});
```
jskeet: See later comment about parentheses for empty constructor argument lists.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
```diff
@@ -67,5 +69,210 @@ public void ListJobs_FullProjection()
...
+            var table = client.GetTable(_fixture.ProjectId, _fixture.DatasetId, _fixture.HighScoreTableId);
+            var options = new QueryOptions() { Labels = JobLabels };
+            var jobToFind = client.CreateQueryJob($"SELECT * FROM {table}", null, options);
+            
+            var jobFound = client.GetJob(jobToFind.Reference);
+
+            VerifyJobLabels(jobFound?.Resource?.Configuration.Labels);
+        }
+
+        [Fact]
```
jskeet: I've generally omitted integration tests if I have a snippet that tests the same thing. (Basically I regard snippets as integration tests in their own right.) I can't easily check right now, but do these tests check anything new? If so, they're very welcome. If they're effectively duplicating snippets, I think we're better off without them.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
```diff
@@ -67,5 +69,210 @@ public void ListJobs_FullProjection()
...
+                Labels = new Dictionary<string, string>()
+                {
+                    {"this_is_valid-2", "This is Invalid & throws" }
+                }
+            };
+
+            Assert.ThrowsAny<GoogleApiException>(() => client.CreateQueryJob($"SELECT * FROM {table}", null, options));
+        }
+
+        internal static IDictionary<string, string> JobLabels =>
```
jskeet: Could you move this to the top of the file? (We tend to write fields/properties, constructors, then methods. Sometimes the properties are after the constructor, and we're not terribly consistent on that.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
```diff
@@ -67,5 +69,210 @@ public void ListJobs_FullProjection()
...
+            new Dictionary<string, string>()
+            {
+                // This one is a little absurd for jobs because labels can only be set
+                // on job creation. This means not to set the label because this is copying
+                // the bahaviour of labels for other resources like DataSets and Table.
```
jskeet: Nit: bahaviour => behaviour (or behavior if we're going for US spelling; I don't mind either way)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
```diff
@@ -67,5 +69,210 @@ public void ListJobs_FullProjection()
...
+                // Testing for it anyways because the api allows is.
+                { "one_label", null },
+                { "another-label-2", "label_value_2" },
+                { "yet_another_label", string.Empty }
```
jskeet: My personal preference is for "" over string.Empty, but I don't much mind either way (and I don't think it's worth worrying about :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Snippets/JobCreationOptionsSnippets.cs
```diff
@@ -0,0 +1,88 @@
...
+            {
+                {"label-key", "label-value" }
+            };
+
+            BigQueryClient client = BigQueryClient.Create(projectId);
+            BigQueryTable table = client.GetTable(projectId, datasetId, tableId);
+            string destinationUri = $"gs://{bucketName}/{objectName}";
+
+            // Just a couple examples of jobs marked with labels:
+            // (These jobs will most certainly be created somewhere else.)
```
jskeet: Not sure what you mean by this parenthesized line.
(We haven't generally added snippets for options, but there's certainly no harm in doing so.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
```diff
@@ -44,5 +46,19 @@ public abstract class JobCreationOptions
...
         /// <seealso cref="BigQueryClient.WithDefaultLocation(string)"/>
         public string JobLocation { get; set; }
+
+        /// <summary>
+        /// The labels associated with the job to be created.
+        /// Labels are key-value pairs.
+        /// You can use these to organize and group your jobs.
+        /// Label keys and values can be no longer than 63 characters,
+        /// can only contain lowercase letters, numeric characters,
+        /// underscores and dashes.International characters are allowed.
```
jskeet: Nit: space after the period.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
```diff
@@ -44,5 +46,19 @@ public abstract class JobCreationOptions
...
+        /// If a label value is <code>null</code>, it won't be added to the job's labels.
+        /// A label value can be <see cref="string.Empty"/>.
+        /// Label keys must start with a letter
+        /// and each label in the list must have a different key.
```
jskeet: Not sure we need to mention about each label having to have a different key - given that it's a dictionary, that would only not be the case if someone provided a very odd comparer or their own dictionary implementation. (I think the fact that it *is* a dictionary is enough.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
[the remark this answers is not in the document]
amanda-tarafa: Adding comment on the use of TakeWhile.

I think the compiler is not set to catch unused variables, for instance, in this same test the variable table wasn't being used either because the $ was missing from the front of the query string.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
amanda-tarafa: Done, definitely cleaner that way.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
amanda-tarafa: Done, here and elsewhere.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
amanda-tarafa: Removed integration tests that were duplicating snippets. Left 3 of them since they are going a little further in testing than the corresponding snippets are. I'm adding comments on this PR for the three tests that remained to signal what they are adding.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
amanda-tarafa: Done.
At some point I though of adding that to the fixture (is used in the load test and the query test also) but wasn't sure about adding the VerifyJobLabels to the fixture as well

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
amanda-tarafa: Done

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
amanda-tarafa: Changed to "" to try and keep everything the same, I don't have a conscious preference, I'm just used to string.Empty I think.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Snippets/JobCreationOptionsSnippets.cs
amanda-tarafa: With the parenthesized line I meant that you wouldn't create jobs "just for the sake of listing them", that is, that those jobs would have been created elsewhere as part of normal execution of the application.

My original intention was not to add the snippet to the options, I wanted to add it as a second sample for ListJobs(*) (list jobs and filter by labels) but couldn't find a way to do so. Don't know yet if it's possible to have two snippets per method signature, don't know if the second snippet is not generated at all or if it is just not included as part of the class docs, and if this is by design or not, wanted to ask you about it after looking at the code a little.

amanda-tarafa: My original intention was not to add the snippet to the options, I wanted to add it as a second sample for ListJobs(*) (list jobs and filter by labels) but couldn't find a way to do so. Don't know yet if it's possible to have two snippets per method signature, don't know if the second snippet is not generated at all or if it is just not included as part of the class docs, and if this is by design or not, wanted to ask you about it after looking at the code a little.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
amanda-tarafa: Done

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
amanda-tarafa: Agreed, done.

amanda-tarafa:
@jskeet  Addressed all comments. Just a question regarding the snippet:

My first thought was to add it as a second snippet to `ListJobs(*)` but couldn't. If it's not really possible to have more than one snippet per method signature (it might be that I am missing something), is this by design or just for simplicity of the snippet handling code? Does it make sense for me to look into allowing more than one snippet?
(If you are OK with the snippet on the options, any answer to the above questions can be handled on a separate PR I think).

Thanks!

jskeet:
Will have to check when I'm on a laptop instead of a phone, but I'm sure it's fine :)

And yes, at the moment we only have support for one snippet per method. I think multiple snippets could end up being a bit overwhelming anyway, to be honest.

PR #2248, opened 2018-06-08
Surfacing Projection for list jobs operation.

amanda-tarafa:
This will make work on #2245 more meaningful. Given that something like this is not supported by the api:

`bqClient.ListJobs(new ListJobsOptions() { WithLabels = labels} );`

then, by surfacing projection something like this is possible:

```
bqClient.ListJobs(new ListJobsOptions() { Projection = ProjectionEnum.Full} )
              .Where(job => job.Resource.Configuration.Labels.Contains(label));
```

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/ListJobsOptions.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/ListJobsOptions.cs
@@ -39,6 +39,11 @@ namespace Google.Cloud.BigQuery.V2
         /// </summary>
         public bool? AllUsers { get; set; }
 
+        /// <summary>
+        /// The information to retrieve for listed jobs.
+        /// </summary>
+        public ProjectionEnum? Projection { get; set; }
+
         internal void ModifyRequest(ListRequest request)
         {
             if (StateFilter != null)
@@ -56,7 +61,10 @@ namespace Google.Cloud.BigQuery.V2
                 request.AllUsers = AllUsers;
             }
 
-            // TODO: Projection?
+            if (Projection != null)
+            {
+                request.Projection = Projection;
+            }
         }
     }
 }
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
@@ -15,6 +15,7 @@
 using System;
 using System.Linq;
 using Xunit;
+using static Google.Apis.Bigquery.v2.JobsResource.ListRequest;
 
 namespace Google.Cloud.BigQuery.V2.IntegrationTests
 {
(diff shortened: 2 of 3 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/ListJobsOptions.cs
```diff
@@ -39,6 +39,11 @@ public sealed class ListJobsOptions
         /// </summary>
         public bool? AllUsers { get; set; }
 
+        /// <summary>
+        /// Whether to return all job information or just part of it.
```
jskeet: Possibly reword to something that doesn't imply it's boolean (even though it is at the moment).
Something like:

"The information to retrieve about each job"

? (The REST API wording isn't terribly helpful here.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/ListJobsOptions.cs
```diff
@@ -56,7 +61,7 @@ internal void ModifyRequest(ListRequest request)
                 request.AllUsers = AllUsers;
             }
 
-            // TODO: Projection?
+            request.Projection = Projection;
```
jskeet: Please change this to only set it if the value is non-null. While we don't compose options anywhere at the moment, I can imagine doing so at some point.

amanda-tarafa: Happy to do so for clarity, but request.Projection is nullable, that's why I didn't check before.

jskeet:
It's more than just clarity - it allows two options to be applied where only one specifies a projection. (It doesn't allow the absence to be specified, admittedly - we would need to handle that another way.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/ListJobsOptions.cs
amanda-tarafa: Done, I think I get what you are saying now, in case in the future we were to surface request.Projection as a combination of different properties in our options class, right? Or is there something I'm still missing?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/ListJobsOptions.cs
amanda-tarafa: Done

amanda-tarafa:
Thanks, rebased and merging when all green.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/ListJobsOptions.cs
jskeet: It's more if we get into a situation where we want to merge two options objects - having null mean "don't change anything" makes that merge easy. That's what we use in CallSettings for gRPC-based projects a lot.

amanda-tarafa: OK, I get it now. Thanks!

PR #2259, opened 2018-06-13
Fixing flakiness in Diagnostics.AspNetCore Integration and Snippets tests.

amanda-tarafa:
Fixing or improving several issues:

- Lack of .pdb info to report error location in Snippets Error Reporting tests for `net4.6` and `netcoreapp1.0`. Only affected Snippets because in one of the tests an exception is thrown from within a nested method call, no such scenario in integration tests. To fix this I added an extra build option.
- Reaching read quota both for Logging and Error Reporting. Started happening after adding the Snippets tests and happens in both Snippets and Integration tests, which makes sense because Snippets are also reading logs now. To fix this, I increased the waiting time between reads (`LogEntryPolling` and `ErrorEventEntryPolling`). It's possible that this can still happen, specially after adding new tests (we do have two test environments running at the same time), if it happens very often, we'll need to increase the waiting time further, but then tests will take longer to execute.
- Log entries not available for listing after written. This apparently started happening after adding the Snippets' tests, but I don't have a good explanation as to why. It affects random Logging Integration and Snippets' tests. After the tests fail I've manually checked several of the Log entries that should have been written on the logs and I've found them there and available for reading. Only docs I've found about availability of entries [are these](https://cloud.google.com/logging/docs/reference/v2/rpc/google.logging.v2#google.logging.v2.WriteLogEntriesRequest) and what's described there as unavailable entries (older than 30 days or timestamped more than 24 hours in the future) doesn't match our case. I've reduced these errors from happening 4 or 5 times in a test run, to happening once in 3 test runs by increasing the amount of time we spend trying to read the entries (`LogEntryPolling`). We can increase this time further but then tests will take longer to execute.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsWebHostTests.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsWebHostTests.cs
@@ -99,7 +99,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
 
         private static async Task TestLogging(string testId, DateTime startTime, HttpClient client)
         {
-            var polling = new LogEntryPolling(TimeSpan.FromSeconds(60));
+            var polling = new LogEntryPolling();
             await client.GetAsync($"/Main/Warning/{testId}");
             var results = polling.GetEntries(startTime, testId, 1, LogSeverity.Warning);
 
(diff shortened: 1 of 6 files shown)


amanda-tarafa:
@jskeet @iantalarico FYA:
Travis failed because of this test: 
https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/9aae7349d0e107fff22cbf2f875ed5170582bd0f/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/RateLimiterTest.cs#L69
with  (Expected 3, Actual 2)

There's this comment on the test code:
> // We should have exactly 3 traces: one at t~=0, one at t~=2, one at t~=4.
   // (The test machine would have to be very highly loaded to take half a second between
   // the check in the while loop and the increment leading to an over-count, and it would have
   // to take half a second to start the threads to lead to an under-count.)

I suppose the test machine was highly loaded at the time of the test execution :). I'm rerunning Travis.

amanda-tarafa:
Thanks!

PR #2266, opened 2018-06-18
Prepare for BigQuery release

jskeet:
Please check I haven't missed anything in the version history :)

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/docs/history.md
+++ b/apis/Google.Cloud.BigQuery.V2/docs/history.md
@@ -1,14 +1,15 @@
 # Version history
 
-# 1.3.0 (unreleased)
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.BigQuery.V2/docs/history.md
```diff
@@ -9,6 +9,7 @@ New features:
 - Support for time-based partitioning by a specific field
```
amanda-tarafa: And also support for time-based partition for load and query jobs, it was all done on the same PR if I remember correctly but they are listed as different features, rightly so I think.

jskeet: Ah yes - I'd got "in query options" but not load. Fixing.

PR #2267, opened 2018-06-18
Make the AppVeyor shell script stop on error

jskeet:
(Currently it *should* be failing on generateprojects.sh, but it's not)

This PR will include a follow-up PR to make it *stop* failing, but I want to see it fail first.


jskeet:
(This failed as expected after just the first commit.)

PR #2281, opened 2018-06-25
Regenerate all APIs

jskeet:
The first three commits are autogenerated. The final one is manually written - please look carefully and we can decide whether to merge this PR as-is, or cut it back to the first two commits and have another look at Tasks.


chrisdunelm:
The first two commits look good.
I think we should only commit those; leaving Tasks unchanged for the moment. What do you think?

PR #2288, opened 2018-06-27
Regenerate Bigtable

jskeet:
This generates the streaming server methods.

It doesn't appear to collide with any hand-written code, but it may mean we can either simplify the manual code, or generate the overload in BigtableClient from the already-generated BigtableServiceApiClient.


evildour:
Yeah I may be able to update the Bigtable generator now to take advantage of this. I'll try to look into that tomorrow.

PR #2256, opened 2018-06-11
GoogleStackdriverTarget for NLog - Added SendJsonPayload property

snakefoot:

PR #2300, opened 2018-06-29
GoogleStackdriverTarget for NLog - Added SendJsonPayload Object Properties

snakefoot:
Extended #2256 with support for DTO / POCO-objects (And collections of POCO-objects).

Still restricted from nested collections and nested objects (Avoid object recursion).

### DIFF
--- a/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Tests/GoogleStackdriverTargetTest.cs
+++ b/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Tests/GoogleStackdriverTargetTest.cs
@@ -91,7 +91,7 @@ namespace Google.Cloud.Logging.NLog.Tests
 
         private async Task<List<LogEntry>> RunTestWorkingServer(
             Func<GoogleStackdriverTarget, Task> testFn,
-            IEnumerable<KeyValuePair<string,string>> withMetadata = null,
+            IEnumerable<KeyValuePair<string, string>> withMetadata = null,
             Platform platform = null,
             bool enableResourceTypeDetection = false,
             bool includeCallSiteStackTrace = false,
@@ -192,7 +192,7 @@ namespace Google.Cloud.Logging.NLog.Tests
                 googleTarget =>
                 {
                     googleTarget.SendJsonPayload = true;
-                    LogManager.GetLogger("testlogger").Info("Favorite {Colors} and {Devices}", new string[] { "Red", "Green", "Blue" }, new Dictionary<string,int>{ ["NTSC"] = 1953, ["PAL"] = 1962, ["SECAM"] = 1956 });
+                    LogManager.GetLogger("testlogger").Info("Favorite {Colors} and {Devices}", new string[] { "Red", "Green", "Blue" }, new Dictionary<string, int> { ["NTSC"] = 1953, ["PAL"] = 1962, ["SECAM"] = 1956 });
                     return Task.FromResult(0);
                 }, includeEventProperties: true);
             Assert.Single(uploadedEntries);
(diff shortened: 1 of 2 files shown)


snakefoot:
@chrisdunelm Any chance this can be reviewed / merged ?

chrisdunelm:
@snakefoot Sorry for the delay on this. We've decided that this library probably isn't the right place for a full serializer to the `Value`  proto; and are currently looking at ways to sort this out.

snakefoot:
@chrisdunelm Possible to have this very simple/primitive object-serialization merged now? Then it can be replaced by the new serializer when it arrives ?

Can only agree that having a protobuf-serialization library would be a great idea.

chrisdunelm:
@snakefoot No, we don't really want to go any further along with the code as it is here, so this PR will not be merged as it is. We are actively looking at resolving this now; although we don't have an exact timeframe.

snakefoot:
@chrisdunelm Well if the time frame is a few weeks, then I guess it is okay. But if it just has been put on the backlog for some future unplanned sprint, then it doesn't make sense to stall this PR.

jskeet:
> But if it just has been put on the backlog for some future unplanned sprint, then it doesn't make sense to stall this PR.

Yes it does, because if we merge this PR now, we end up having to support it forever. API doesn't shouldn't be rushed through just because a more appropriate design might take longer. Pretty much every time I've thought "Oh, we'll accept X because Y isn't ready yet" I've regretted it.

Not doing anything is *absolutely* a viable option here.

If you need something faster than we're able to implement it in the way we want, I'd suggest forking the repo for the moment.

None of this is to say we don't want to improve things - but "you must accept this PR if an alternative isn't coming really soon" is not a good argument.

snakefoot:
@jskeet Not sure I understand. Any serializer that can take an object and turn it into a protobuf-object will be a better implementation than this primitive PR. This is just a broken walking stick until the real guns arrive. But it allows people to experiment with structured logging, and discover pitfalls.

jskeet:
Anything we'd do in future would have to be compatible with this. I can imagine various possible solutions where that would limit us significantly.

snakefoot:
@jskeet Just do like everyone else. Bump the major version, throw away the broken code, put in the new better code and tell people to be happy. Remember this is just a BETA version.

Alternative you can show me how to register and acquire a protobuf-serializer-interface (Converts an object into protobuf). Then I will rip out the old code, and just lookup and exercise any available protobuf-serializer in the NLog-target.

Then people can register their own serializers at runtime, that and copy-paste the code from this PR into their own individual application.

jskeet:
Bumping the major version and introducing incompatibility adds pain for everyone. I appreciate that this is only a beta version, but that doesn't mean I'm willing to put in code that we know we don't want.

The "register and acquire a serializer" approach is exactly what I was thinking of, and what I hope we'll get to soon - but I really, really don't want to merge this in the meantime.

chrisdunelm:
@snakefoot We've discussed this, and intend to do the following:

* Remove the JsonPayload code, effectively undoing PR #2256 
* Add a new configuration option that allows a custom object-to-proto-Struct conversion function to be specified.
* Create a default object-to-proto-Struct conversion; which uses json.net to perform a C#-object-to-JObject conversion, with new code to convert a JObject to a proto Struct object.
* Do a new beta release.

This will provide a reliable and maintainable default conversion, but keep the capability for users to override the default C# object to proto Struct if required.

One potential issue is that this requires a dependency on the Newtonsoft.Json package.
(This package already has that dependency transitively via our Google.Apis.Gax package, but we intend to remove that dependency at some point. Newtonsoft.Json will then need to be a direct dependency of this package)
So long as you don't see this as a problem, then this seems to us to be the simplest and neatest way of providing this structured logging.

snakefoot:
Sounds fine with me. Though it will probably require some logic to configure the serialiser  (Enum conversion, Dictionary-key-cleanup, Skip Exception TargetSite, etc). Newtonsoft is nice for controlled types but can go crazy when random objects like exceptions


Sent from my Samsung device


-------- Original message --------

chrisdunelm:
Closing. Final work now in issue #2355

PR #2305, opened 2018-07-02
Switching generated artifact to GAPIC_CODE

garrettjonesgoogle:
Details:

- This change means that the package_yaml2 file can be dropped
- One small casualty of this approach is that the smoke test csproj file won't be generated any more; it wouldn't have been logically coherent to retain it, though. (I think you guys can cope?)

cc @jskeet 

(As before, I haven't personally run the updated version of this script)


jskeet:
Thanks Garrett - I'll run this tomorrow and hopefully merge it then. I'm sure we'll be able to handle generating the smoke test projects ourselves.

jskeet:
Regenerated, and all looks good - thanks!

PR #2306, opened 2018-07-03
Update logging

jskeet:
For a change, please review even the first commit (generated code). I think it's fine to just review the Client code... I'm expecting that the changes are:

- Comments
- Timing changes
- A new overload for UpdateSink to accept a FieldMask
- Change of signature for ListLogEntries

The last is a breaking change, which is then addressed in the second commit, which was manually written, and which we're hoping to get rid of later via generator changes.


jskeet:
Much appreciated :)

PR #2315, opened 2018-07-06
Replacing 'Cannot be null' by 'Must not be null' in docstrings in Diagnostics.

amanda-tarafa:
For both Asp.Net Diagnostics and Asp.Net Core Diagnostics


amanda-tarafa:
Thanks, merging without waiting for all green since these are changes in comments.

PR #2317, opened 2018-07-09
Improved diagnostics for Spanner

jskeet:
This doesn't fix the resource issues I'm looking into, but it improves the diagnostics. I can then look at fixing it more easily...

I'd suggest reviewing a commit at a time.


jskeet:
Hold up - I've included way more in the first commit than expected. Will reassign when it's ready...

chrisdunelm:
OK, waiting...

jskeet:
Okay, it's looking better now. Thanks!

PR #2321, opened 2018-07-10
Release Logging 2.2.0 and KMS 1.0.0-beta02

jskeet:


chrisdunelm:
Should log4net and nlog be updated to depend on this new version of logging, and also released?

jskeet:
Unless they need the new features, I don't think they really need updating/releasing. Next time we want to do a release of them for some other reason, it might be worth updating their dependencies.

Issue #2316, opened 2018-07-09
Investigate UrlSigner support for ComputeCredential


jskeet:
Branched off #1056, specifically for ComputeCredential, or possibly ServiceCredential.

> [A] limitation makes it nearly impossible to use `UrlSigner` in conjunction with implicit service account that is associated with GCE instance. When running an app on GCE instance and calling to `GoogleCredential.GetApplicationDefault()` the `ComputeCredential` is returned. There is no way to get an instance of `ServiceAccountCredential` from an instance of `ComputeCredential`.
I would expect that since both are based on `ServiceCredential` type, they both can be used for signing URLs. But the library is very strict on the matter.

jskeet:
Initial investigation: `ComputeCredential` doesn't hold any secrets. The only information it has is a URL to talk to, when it needs to request an access token. Unless the metadata server is capable of performing URL signing, this could still be impossible with only a ComputeCredential, leaving the following fallback options:

- Providing a service account credential file even when running on Compute Engine
- Creating an access key and signing with HMAC-SHA1

(I'm not giving up yet though.)

minherz:
The first option requires local management of the service account credentials including rotation of private key and re-deployment of the renewal credentials to running instances. Therefore, it isn't good for production.
If it is possible to implement a signing algorithm using `ServiceCredential` token, it will be a simplest and fastest solution.
An alternative is to implement something in GCS as a new API :(

jskeet:
Okay, I've been looking into this a bit more, and I've worked out a way that I believe will work, but I want to see whether it'll be useful to you or not before I implement it.

The IAM API has a [SignBlob method](https://cloud.google.com/iam/reference/rest/v1/projects.serviceAccounts/signBlob) which I *believe* will create the appropriate signature that `UrlSigner` uses.

`UrlSigner` needs two things from the `ServiceAccountCredentials` at the moment:

- A way of signing an arbitrary blob
- An ID

You could either create a service account *specifically* for URL signing, and grant access to the default Compute Engine service account, or you could use the metadata API to discover the default Compute Engine service account and use the directly. With the first option you'd have to specify the account to use as part of your application configuration.

I'm prototyping this now just to prove that it works, and I'll update this issue with a link to a "not for submission" PR if it does.

jskeet:
Okay, I've prototyped it in #2325.

[...]

PR #2325, opened 2018-07-11
(Not for submission) Demo of URL signing via the IAM SignBlob API

jskeet:
Example of how we might solve #2316.


jskeet:
Superceded by #2338.

PR #2332, opened 2018-07-12
Implement BigQueryNumeric

jskeet:
All of this code has been reviewed already, but we're interested in:

- The integration into the rest of the BigQuery API (in the first commit)
- The API for the BigQueryNumeric type

(I'm going to revisit both myself, as well as asking Chris and Amanda to do so. I'm out tomorrow, but will look carefully on Monday.)


jskeet:
An alternative option would be to merge, then create a new issue saying "We can't release BigQuery until we've reviewed thoroughly." That might be simpler in terms of looking at code in VS rather than in GitHub. Shout what you think :)

jskeet:
Merging, and we'll then review individually. (All code has already been reviewed.)

amanda-tarafa:
This looks good! I only had a couple of comments which I created #2336 to address.

PR #2322, opened 2018-07-10
Support JSON credentials in log4net

chrisdunelm:
Fixes #2314

PR #2334, opened 2018-07-13
Support JSON credentials in NLog

### LINKED ISSUE #2314: GoogleStackdriverAppender doesn't allow for custom credentials
There is a `CredentialFile` property but please consider something like `CredentialJson` or a more generic `Credential`. That would be useful for loading the credential data for example from encrypted storage, DPAPI, or some vault rather than from a plain file on disk.

Thanks!

https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs#L204

snakefoot:
Fixes #2314 for NLog.

### DIFF
--- a/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.IntegrationTests/CredentialTest.cs
+++ b/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.IntegrationTests/CredentialTest.cs
@@ -1,0 +1,107 @@
+﻿// Copyright 2018 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


jskeet:
I'm not quite sure why this PR exists when #2314 is already fixed by #2322. Please could you give more context?

snakefoot:
@jskeet NLog and log4net are both logging frameworks. They should have the same options for configuration.

jskeet:
Whoops, I hadn't spotted the different framework involved. The comment including "Fixes #2314" confused me - I think it would be clearer to create a new issue which was NLog-specific (just as #2314 is Log4Net-specific) and then change the commit to refer to that new issue instead.

PR #2336, opened 2018-07-16
[BigQueryNumeric] Modifying tests to:

amanda-tarafa:
- Test against (not accepting) culture formats.
- CompareTo_Equal comparing two different instances.

These changes basically are my comments after reviewing #2332


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryNumericTest.cs
```diff
@@ -219,6 +219,9 @@ public void ToDecimal_Overflow(LossOfPrecisionHandling handling, bool negate)
         [InlineData("+.5")]
         [InlineData("1e6")]
         [InlineData("1e-6")]
+        // Other culture formatting
```
jskeet: It might be worth doing this in a separate test which sets the thread culture to a culture where this *would* be valid first.

amanda-tarafa: Agreed, I will merge this one since it's LGTMed already and submit a new one with the new test you are describing.

PR #2338, opened 2018-07-17
More flexible URI signing

jskeet:
This addresses #2316 by creating a seam for the signing part of
UrlSigner.

In the future we may want to release a package specifically to make
the IAM integration simple, but for the moment we've just got a
sample in the docs.

(Another possibility would be to add a dependency on the IAM
package and include the IAM signer here, but I'd prefer not to do
that.)

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
@@ -12,13 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Api.Gax;
 using Google.Apis.Auth.OAuth2;
+using Google.Apis.Iam.v1;
+using Google.Apis.Iam.v1.Data;
+using Google.Apis.Services;
 using Google.Cloud.ClientTesting;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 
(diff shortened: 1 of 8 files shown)


minherz:
> Another possibility would be to add a dependency on the IAM package

The PR already introduces a dependency on `Google.Apis.Iam.v1` package, isn't it?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
```diff
@@ -119,5 +124,88 @@ public UrlSignerSnippets(StorageSnippetFixture fixture)
...
+                _iamService = service;
+                Id = id;
+            }
+
+            public string CreateSignature(byte[] data) =>
+                CreateRequest(data).Execute().Signature;
+
+            public async Task<string> CreateSignatureAsync(byte[] data, CancellationToken cancellationToken)
+            {
+                var request = CreateRequest(data);
```
evildour: Should the snippets be using `var`?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
```diff
@@ -93,7 +92,18 @@ public static UrlSigner FromServiceAccountData(Stream credentialData)
...
             GaxPreconditions.CheckNotNull(credential, nameof(credential));
-            return new UrlSigner(credential);
+            return new UrlSigner(new ServiceAccountCredentialBlobSigner(credential));
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="signer"></param>
+        /// <returns></returns>
```
evildour: Missing docs

jskeet: Doh! Thanks, will fix.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
jskeet: Ooh, probably not. I'm not *as* bothered for this snippet as others (as it's an example implementation that can be copy/pasted) but I'll fix it anyway.

PR #2345, opened 2018-07-17
Minor improvements for Spanner resource management

jskeet:
These don't make significant changes, but at least make it better than it was.

I strongly suspect I can't improve things much further without a comprehensive rewrite of the resource management.


jskeet:
Ooh - have merged this but just seen a bunch of integration test failures. Will need to rerun...

jskeet:
(Was then fine; will run a few more times over the next couple of days.)

PR #2346, opened 2018-07-17
Update non-production dependency versions

jskeet:
(First commit is manual, second is autogenerated.)


jskeet:
Trying again without ConfigureAwait updated - I suspect that may need a more recent version of the .NET Core SDK.

jskeet:
The `Google.Cloud.Logging.NLog.Tests.GoogleStackdriverTargetTest.SingleLogEntryWithLocation` test is failing, just on net452. It isn't getting a line number - I suspect this is some change to debug information.

The relevant dependency is Microsoft.NET.Test.Sdk - 15.5.0 works, 15.6.0+ doesn't. While we *could* just revert that part of the change, I'd really rather not pin ourselves to 15.5.0 forever. Still investigating.

PR #2348, opened 2018-07-17
Add Kokoro directory with helper scripts to enable the CI/CD

iantalarico:
These will allow us to keep the logic for the CI/CD in source control and make running the CI/CD easier.

### DIFF
--- a/.kokoro/builddocs.sh
+++ b/.kokoro/builddocs.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+cd ..
+
+# Build the libraries w/o running unit tests.
+./build.sh --notests
+
+# Build the docs.
+cd docs
+./builddocs.sh
--- a/.kokoro/release.sh
+++ b/.kokoro/release.sh
@@ -15,6 +15,6 @@ if ! git diff --quiet --cached; then git commit -m 'Regenerate docs'; git push; 
 
 cd ../..
 
-#Push the changes to nuget.
+# Push the changes to nuget.
 cd ./releasebuild/nuget
 for pkg in *.nupkg; do dotnet nuget push -s https://api.nuget.org/v3/index.json -k $NUGET_API_KEY $pkg; done
--- a/.kokoro/release.sh
+++ b/.kokoro/release.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+cd ..
+
+# Build the release and run the tests.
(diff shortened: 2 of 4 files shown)


.kokoro/builddocs.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
+
+cd ..
+
+# Build the libraries w/o running unit tests.
```
jskeet: Nit: "without" would be clearer than "w/o", and it's not like we're short of space.

.kokoro/release.sh
```diff
@@ -0,0 +1,20 @@
...
+
+
+# Add any docs changes if they exist.
+cd ./releasebuild/releasedocs
+git add --all
+if ! git diff --quiet --cached; then git commit -m 'Regenerate docs'; git push; fi
+
+cd ../..
+
+#Push the changes to nuget.
```
jskeet: Nit: space after `#`

.kokoro/runintegrationtests.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+cd ..
+
+# Build the libraries w/o running unit tests.
```
jskeet: Maybe we should run the unit tests as well? Make this more of a "full build".

.kokoro/builddocs.sh
iantalarico: Done

.kokoro/release.sh
iantalarico: Done

.kokoro/runintegrationtests.sh
iantalarico: Can't hurt, updated.

.kokoro/runintegrationtests.sh
```diff
@@ -0,0 +1,13 @@
...
+set -e
+
+cd ..
+
+# Build the libraries and run unit tests.
+./build.sh
+
+# Allow each integration test 3 chances to pass.
+./runintegrationtests.sh || true
+./runintegrationtests.sh --retry || true
```
chrisdunelm: Does it matter that this is running `runintegrationtests.sh` three times, even if the first run passes?

iantalarico: No, it will run tests for all entries in the failure file.  Which will be empty if everything passes.  We could probably make this a little more efficient (not having to run some checks in a shell script) but at the cost of complexity.

PR #2357, opened 2018-07-18
Add environment variables that rely on secrets for CI/CD builds and releases

iantalarico:
I'm not a huge fan of how this worked out but this is the only way AFAICT


iantalarico:
@jskeet Yea and it also looks like others are doing the same as we are.  I was really hoping we could do the env variable swap in the configs but it doesn't seem possible.

PR #2359, opened 2018-07-19
Update default Grpc.Core version to 1.13.1

chrisdunelm:
This contains a shutdown fix, especially relevant when running on Kubernetes. See https://github.com/grpc/grpc/issues/14856


chrisdunelm:
Hmm, I hadn't considered that this update would require all project files to be regenerated. @jskeet is this something that is OK to be done now, or shall I close this PR and it can be done later?

chrisdunelm:
Will merge on green

PR #2363, opened 2018-07-19
ID generation

jskeet:
Next step to addressing #2198 

This PR does *not* change the Diagnostics solutions, but prepares the ground.

Most changes here are pure refactorings; in some places I've changed from a GUID-based ID to a date/time-based ID as that's proved useful when diagnosing issues.

The main thing to check in the review is that this *only* affects test code.

Issue #2198, opened 2018-05-31
Move and rename Google.Cloud.Diagnostics.Common.Tests.Utils to Google.Cloud.ClientTesting


amanda-tarafa:
... or at least all the parts of it that are of use for testing across all client libs and not just diagnostics.
See related comment on #2192

iantalarico:
I think the only two we should move are `GetProjectIdFromEnvironment` and `IsWindows`, I don't think the others would be used by other libs.

jskeet:
Makes sense - and even GetProjectIdFromEnvironment may not be necessary, as if you have a fixture deriving from CloudFixtureBase, that already exposes a ProjectId property.

amanda-tarafa:
`GetTestId` is just returning a `Guid.NewGuid().ToString()` and I've seen that in several places, not as test id but as unique string, so it might be nice to have it as a helper somewhere, although is so small and simple...

True for the `CloudFixtureBase`, I think I will be using that one for `ProjectId` instead of the method.

jskeet:
I'm going to have a look at this, including a common test method for Guid.NewGuid().ToString() that allows for truncation and maybe a prefix.

jskeet:
Now complete.

amanda-tarafa:
Closing via #2361, #2363, #2364

PR #2364, opened 2018-07-19
Changes so that Google.Cloud.Diagnostics uses the new ID generator.

amanda-tarafa:
Next and last step to addressing #2198 

This PR changes the diagnostics solutions so that it used the new ID Generator from #2363.

I've changed from a Guid base ID to a datetime base ID, which will help with diagnosing issues.


amanda-tarafa:
Thanks!

PR #2366, opened 2018-07-19
Change how to 'cd' into the proper directory for scripts

iantalarico:
On the CI/CD the script is not run from the same location it exists. We instead need to grab the script location and 'cd' into the right location to run scripts.

### DIFF
--- a/.kokoro/runintegrationtests.sh
+++ b/.kokoro/runintegrationtests.sh
@@ -3,9 +3,9 @@
 set -e
 
 SCRIPT=$(readlink -f "$0")
-ROOT_DIR=$(dirname "$SCRIPT")
+SCRIPT_DIR=$(dirname "$SCRIPT")
 
-cd $ROOT_DIR
+cd $SCRIPT_DIR
 cd ..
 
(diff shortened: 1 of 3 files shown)


.kokoro/runintegrationtests.sh
```diff
@@ -2,11 +2,15 @@
 
 set -e
 
-export GOOGLE_APPLICATION_CREDENTIALS="$KOKORO_KEYSTORE_DIR/73609_cloud-sharp-jenkins-compute-service-account"
-export REQUESTER_PAYS_CREDENTIALS="$KOKORO_KEYSTORE_DIR/73609_gcloud-devel-service-account"
+SCRIPT=$(readlink -f "$0")
+ROOT_DIR=$(dirname "$SCRIPT")
```
jskeet: I think "ROOT_DIR" is a misleading name here. It sounds like it's the root of the repository, when it's actually just the directory containing the script. How about SCRIPT_DIR?

iantalarico: Done

PR #2378, opened 2018-07-25
Add Windows bat files to Kokoro to ensure we can see output when running shell scripts.

iantalarico:
Currently the shell scripts run but we don't get any output.  This will allow us to see the output and diagnose errors.

I'm not the biggest fan of this solution but seems like the easiest.  It looks like other have had the same issue and just wrote all scripts in PowerShell

### DIFF
--- a/.kokoro/build.bat
+++ b/.kokoro/build.bat
@@ -1,0 +1,3 @@
+:: See documentation in type-shell-output.bat
+
+%~dp0type-shell-output.bat build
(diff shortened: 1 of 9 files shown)


iantalarico:
@jskeet @chrisdunelm Yea I 10)% agree we can do something better.  I need to do more investigation to try and figure out the best solution (ideally I would like the output to stream so we can look at it during build). I'm going to submit this for now as it will solve the immediate issue.

PR #2382, opened 2018-07-26
Update how we run shell scripts from bat files

iantalarico:
This was recommended from the kokoro team


.kokoro/release.bat
```diff
@@ -1,3 +1,2 @@
-:: See documentation in type-shell-output.bat
-
-%~dp0type-shell-output.bat release
\ No newline at end of file
+cd %~dp0
+bash -c ./build.sh
```
chrisdunelm: Is this meant to be `release.sh`

iantalarico:
This will end up running cygwin.  Do we need it to be git bash?

iantalarico:
I just tried it with cygwin.  This will not work.

PR #2381, opened 2018-07-25
Adding UpdateSchemaOptions to CreateLoadJobOptions.

amanda-tarafa:
This is something that we missed on #2176

PR #2383, opened 2018-07-27
Fixing the flakiness in the integration test from #2381

amanda-tarafa:
I assumed a table would not be inserted into by other tests so I was expecting a fixed amount of rows.


amanda-tarafa:
Thanks!

PR #2386, opened 2018-07-30
Regenerate clients with Create overload accepting CallInvoker

jskeet:
This PR *only* contains those changes. A future PR will contain commits for other API changes.
A later PR (currently scheduled for August 2nd) will contain another change within the new overload, to use the Interceptor property from GAX gRPC 2.5.0, when that's released.

### DIFF
--- a/apis/Google.Cloud.Spanner.Admin.Instance.V1/Google.Cloud.Spanner.Admin.Instance.V1/InstanceAdminClient.cs
+++ b/apis/Google.Cloud.Spanner.Admin.Instance.V1/Google.Cloud.Spanner.Admin.Instance.V1/InstanceAdminClient.cs
@@ -595,7 +595,19 @@ namespace Google.Cloud.Spanner.Admin.Instance.V1
         public static InstanceAdminClient Create(grpccore::Channel channel, InstanceAdminSettings settings = null)
         {
             gax::GaxPreconditions.CheckNotNull(channel, nameof(channel));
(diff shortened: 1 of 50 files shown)


apis/Google.Cloud.Spanner.Admin.Instance.V1/Google.Cloud.Spanner.Admin.Instance.V1/InstanceAdminClient.cs
```diff
@@ -595,7 +595,19 @@ public static InstanceAdminClient Create(gaxgrpc::ServiceEndpoint endpoint = nul
         public static InstanceAdminClient Create(grpccore::Channel channel, InstanceAdminSettings settings = null)
         {
             gax::GaxPreconditions.CheckNotNull(channel, nameof(channel));
-            InstanceAdmin.InstanceAdminClient grpcClient = new InstanceAdmin.InstanceAdminClient(channel);
+            return Create(new grpccore::DefaultCallInvoker(channel), settings);
```
amanda-tarafa: Not sure this is a problem, I certainly don't see anything wrong, but for some reason github is not "coloring" this code according to syntax, and the same happens with a lot of the changes below. Changes above, and original code is colored.

amanda-tarafa: Scratch that, I just had to reload the page, apparently the CSS hadn't applied entirely.

PR #2387, opened 2018-07-30
Remaining API regeneration

jskeet:
This is all autogenerated code. It just clears the decks for Thursday's regeneration.

cc @evildour @chrisdunelm as a heads-up for the streaming retry settings that aren't used in Bigtable.


amanda-tarafa:
Travis is not building.

jskeet:
Thanks - I suspect we're missing an update to the common protos (or possibly I just need to bump the dependency).

jskeet:
Got it - added an extra commit and pushed; will hopefully build correctly this time :)

Issue #1894, opened 2018-02-08
Integrate OpenCensus with Google implementation


iantalarico:
This will involve work for both HTTP and gRPC libraries.

For HTTP libraries we need to ensure they can easily accept a HttpMessageHandler (This already works but is complicated). 

For gRPC we need to work with the gRPC team to see how we can hook into the native lib.

cc: @jskeet @chrisdunelm

jskeet:
For gRPC, we are likely to expose an interceptor property into client settings. This can be used for logging as well as other use cases. Do we think that would handle everything we'd need?

jskeet:
(Actually, presumably this isn't something the user should set up. Ian and Amanda, let's talk about this some time soon.)

iantalarico:
Agreed the only set up that the user should be doing is to add the propagating handler to the http/grpc client and then to the library.

amanda-tarafa:
Still blocked on OpenCensus .NET integration. Renamed issue.

jskeet:
I'm going to close this as an issue for now. OpenCensus work is well underway, but won't be ready for us to integrate with for quite a while, and I don't think we'll miss the fact that we should integrate with it when we can.

PR #2389, opened 2018-07-31
[Not for submission] Propagating a gRPC trace header.

amanda-tarafa:
Work towards #1894

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/Trace/GrpcTraceHeaderFactoryTest.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/Trace/GrpcTraceHeaderFactoryTest.cs
@@ -1,0 +1,81 @@
+﻿// Copyright 2018 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.ClientTesting;
+using Google.Cloud.Diagnostics.Common.Trace;
+using Google.Cloud.Spanner.Admin.Instance.V1;
+using Google.Cloud.Trace.V1;
+using Google.Protobuf.WellKnownTypes;
+using System;
+using Xunit;
+using TraceProto = Google.Cloud.Trace.V1.Trace;
+
+namespace Google.Cloud.Diagnostics.Common.IntegrationTests.Trace
+{
+    public class GrpcTraceHeaderFactoryTest : IDisposable
+    {
+        private static readonly TraceIdFactory _traceIdFactory = TraceIdFactory.Create();
+        private static readonly TraceEntryPolling _polling = new TraceEntryPolling();
+
+        private readonly string _testId;
+
+        private readonly IConsumer<TraceProto> _consumer;
+        private readonly IManagedTracer _tracer;
+
+        private readonly Timestamp _startTime;
+
+        public GrpcTraceHeaderFactoryTest()
+        {
+            _testId = IdGenerator.FromDateTime();
+
+            _consumer = new GrpcTraceConsumer(TraceServiceClient.Create());
+            _tracer = SimpleManagedTracer.Create(_consumer, TestEnvironment.GetTestProjectId(), _traceIdFactory.NextId(), null);
+
+            _startTime = Timestamp.FromDateTime(DateTime.UtcNow);
+        }
+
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/GrpcTraceHeaderFactory.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// that trace is propagated to the gRPC request.
+        /// </summary>
+        public Action<Metadata> GetHeaderMutationForPropagation()
+        {
+            return metadata => metadata.Add(GetGrpcHeader());
+        }
+
+        /// <summary>
+        /// Builds a <see cref="Metadata.Entry"/>, that is a gRPC header, for trace propagation.
+        /// The header is built based on the tracer returned by the manageTracerFactory with
```
jskeet: Nit: given that ManagedTracerFactory isn't a type, I'd just expand it to "managed tracer factory".

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/GrpcTraceHeaderFactory.cs
```diff
@@ -0,0 +1,170 @@
...
+            var traceIdBytes = traceIdGuid.ToByteArray();
+            if (traceIdBytes.Length != TraceIdSize)
+            {
+                throw new InvalidOperationException($"Unexpected size for byte representation of trace ID. Expected: {TraceIdSize}, Actual: {traceIdBytes.Length}");
+            }
+
+            Array.Copy(traceIdBytes, 0, output, offset + 1, TraceIdSize);
+        }
+
+        private void SetSpanId(byte[] output, ulong traceId, int offset)
```
jskeet: traceId => spanId?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/Trace/GrpcTraceHeaderFactoryTest.cs
```diff
@@ -0,0 +1,81 @@
...
+            _consumer = new GrpcTraceConsumer(TraceServiceClient.Create());
+            _tracer = SimpleManagedTracer.Create(_consumer, TestEnvironment.GetTestProjectId(), _traceIdFactory.NextId(), null);
+
+            _startTime = Timestamp.FromDateTime(DateTime.UtcNow);
+        }
+
+        [Fact]
+        public void TraceOutgoing()
+        {
+            var spanName = EntryData.GetMessage(nameof(TraceOutgoing), _testId);
```
jskeet: Eek, I hadn't realized we'd need a separate CallSettings for each method. This is icky.
I wonder whether the new gRPC interceptor support will help... let's leave that until we've got it working though.

amanda-tarafa: Well, CallSettings are not needed per method, but per client instance, what we need per method is to create the proper span, whose name needs to be (according to instructions)  package.service/method. I don't think we have that info anywhere in the client libs for each of the methods, so it will fall on clients to get that info correctly. Agree, super icky.

jskeet: Ah yes, I see. I think [this from gRPC](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core/Method.cs) would really help - it's available in the ClientInterceptorContext, so an interceptor may well be the way forward here.

amanda-tarafa: Oh, interesting, I'll take a look at the interceptor then.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/GrpcTraceHeaderFactory.cs
amanda-tarafa: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/GrpcTraceHeaderFactory.cs
amanda-tarafa: Done

iantalarico:
@amanda-tarafa Looks like Jon has made some comments, is this ready for review?

amanda-tarafa:
@iantalarico Take a look if you want and leave your comments, but this is not for submission until we figure out if just sending the header is enough, and if it's being sent properly.

iantalarico:
@amanda-tarafa I took a quick look over everything and the logic seems reasonable although I don't have much understanding on grpc trace headers.

jskeet:
Would it be worth deleting this PR, but keeping the branch in your fork? (Just trying to work out how we can clear the decks a little.)

amanda-tarafa:
Yes I'll do that.

PR #2399, opened 2018-08-02
Releases for gRPC based projects (second batch)

amanda-tarafa:


apis/apis.json
```diff
@@ -398,7 +398,7 @@
 
   {
     "id": "Google.Cloud.Logging.Type",
-    "version": "2.1.0-beta00",
+    "version": "2.1.0-beta01",
```
jskeet: This doesn't need to be updated.

apis/apis.json
```diff
@@ -468,7 +468,7 @@
     "id": "Google.Cloud.OsLogin.V1",
     "productName": "Google Cloud OS Login API",
     "productUrl": "https://cloud.google.com/compute/docs/instances/managing-instance-access",
-    "version": "1.0.0-beta01",
+    "version": "1.0.0-beta02",
```
jskeet: Because we're updating this and it has a "project" dependency on Google.Cloud.OsLogin.Common, we need to update the Common one above too. (The tagreleases code would have warned about this, don't worry :) Will explain in detail in a VC

apis/apis.json
```diff
@@ -576,7 +576,7 @@
 
   {
     "id": "Google.Cloud.EntityFrameworkCore.Spanner",
-    "version": "1.0.0-alpha04",
+    "version": "1.0.0-alpha05",
```
jskeet: I'm very tempted to remove this as a project. (No change required now, just thinking...)

apis/apis.json
```diff
@@ -663,7 +663,7 @@
     "id": "Google.Cloud.Storage.V1",
     "productName": "Google Cloud Storage",
     "productUrl": "https://cloud.google.com/storage/",
-    "version": "2.2.0-beta02",
+    "version": "2.2.0-beta03",
```
jskeet: No need to update this - it's a REST API.

apis/apis.json
amanda-tarafa: Done

apis/apis.json
amanda-tarafa: Done.

PR #2401, opened 2018-08-02
Allow retries of integration tests during a release.

iantalarico:
Integration tests can be flakey and we should allow retries.

### DIFF
--- a/buildrelease.sh
+++ b/buildrelease.sh
@@ -69,9 +69,7 @@ fi
 # be a little flakey.
 if [[ "$run_tests" = true ]]
 then
-  ./runintegrationtests.sh $projects || true
-  ./runintegrationtests.sh --retry || true
-  ./runintegrationtests.sh --retry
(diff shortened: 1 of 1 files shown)


buildrelease.sh
```diff
@@ -65,9 +65,13 @@ fi
 
 ./build.sh $projects
 
+# Retry integration tests up to 3 times as they can sometimes
+# be a little flakey.
 if [[ "$run_tests" = true ]]
 then
-  ./runintegrationtests.sh $projects
+  ./runintegrationtests.sh $projects || true
```
jskeet: Alternatively...

```sh
./runintegrationtests.sh $projects || ./runintegrationtests.sh --retry || ./runintegrationtests.sh --retry
```

? That would only run it as many times as necessary, right?

iantalarico: Yea good call, we skip unneeded calls to the script.

iantalarico:
PTAL

Issue #836, opened 2017-03-02
Write a Google.Cloud.Diagnostics.AspNetCore Trace test that traces multiple requests and propagates trace headers


iantalarico:
This will catch issues with the scoping of services used in the Google.Cloud.Diagnostics.AspNetCore Trace bits.

amanda-tarafa:
Closed via #2402

PR #2402, opened 2018-08-02
Test for propagating an HTTP header across multiple requests.

amanda-tarafa:
Addresses #836 

And also, once we have been able to propagate gRPC OpenCensus headers (see #2389) we should be able to write a similar test for that.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
@@ -24,6 +24,7 @@ using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -148,6 +149,44 @@ public async Task Trace_Header()
...
+            using (var client = server.CreateClient())
+            {
+                var response = await client.GetAsync(createFirstSpanUri);
+            }
+
+            var trace = _polling.GetTrace(createFirstSpanUri, _startTime);
+
+            Assert.NotNull(trace);
+
+            var parentSpan = trace.Spans.Single(s => s.Name == createFirstSpanUri);
```
jskeet: After a few minutes, I'm getting lost here. It might be simpler to have a separate variable for each span - and definitely a comment.

amanda-tarafa: My bad, I needed a diagram to write the test so no wonder you are having a hard time. Added comments and different variables.

iantalarico:
I need to think about this a little more, I think this tests the right thing but is a little hard to follow.  We have a server start another server and call that server.

I think it might be easier to follow and be a bit more explicit if you start two test servers independently of each other and have one call the other.

I'm not opposed to the current approach, just an idea.  Thoughts?

amanda-tarafa:
@iantalarico What I was trying to mimick here was several services calling each other. The servers are started on the controller constructor, so they are independent from one another, and one calls the other. 
Let me try and alternative to spawning test servers in the controller, using some static variables, and see if it's clearer.

amanda-tarafa:
I've added a second commit with an alternative implementation where test servers are created outside the services to add clarity, hopefully. @iantalarico @jskeet PTAL

amanda-tarafa:
Yes, same here, tbh, I like the first commit better as well.

iantalarico:
I just think it's more explicit about what is happening here when as we are calling from two servers.  However, I'm really okay with either approach and will leave it up to you two if you would prefer to stick with commit 1.

amanda-tarafa:
I'll be merging the second commit, I like the first one better, but it's not clearer anyways tbh. On the second commit there are way more comments than on the first one, so it might be easier to understand.

amanda-tarafa:
Waiting for all green before merging. Thanks!

PR #2403, opened 2018-08-02
[Diagnostics ASP.Net] Fixing flakiness in a snippet.

amanda-tarafa:
Checking for the "first level method" and not for the nested method that actually throws the exception because for some reason (that I haven't found) the StackTrace when running in Jenkins doesn't go "deep".

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.Snippets/ErrorReportingSnippets.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.Snippets/ErrorReportingSnippets.cs
@@ -84,7 +84,7 @@ namespace Google.Cloud.Diagnostics.AspNet.Snippets
 
             var errorEvent = s_polling.GetEvents(_startTime, _testId, 1).Single();
             ErrorEventEntryVerifiers.VerifyFullErrorEventLogged(
-                errorEvent, _testId, "DoSomething", s_expectedStatusCode);
+                errorEvent, _testId, nameof(HomeController.CatchesAndLogsExceptions), s_expectedStatusCode);
         }
 
         public void Dispose()


jskeet:
I think there's only one framework involved here - it's Windows-only.

My guess is that this is actually a matter of whether the JIT compiler inlines the call or not. Another option we might want to try is to use `MethodImpl(MethodImplOptions.NoInlining)` (or whatever it is; my memory is rubbish) to stop that happening.

amanda-tarafa:
@iantalarico pdb's are the first thing I checked and I think they are alright. We have the `pdbonly` setting, we don't have the `Optimize code` checkbox checked. And we do get location for where the error happens, only one level up. It's probably, as @jskeet suggested, that the method call is being inlined, I didn't think of that tbh, I thought it was due to some optimization config on the compiler in Jenkins. Let me try that attribute and see what happens, I'll run it in Jenkins.

jskeet:
I've just thought: when we run the integration tests, we run with `-c Release` implicitly - which *will* optimize. You might want to try running with that locally as a way of reproducing the problem.

jskeet:
I've managed to reproduce it using `-c Release`. Just testing more thoroughly, then will try to fix and create a PR.

amanda-tarafa:
I can confirm that the test does not pass with `-c Release` unless the method is marked with `[MethodImpl(MethodImplOptions.NoInlining)]`. That is, the release configuration is inlining the method call by default.
My general question now is, should we tweak the code in this "fairly unusual" manner (marking the method with the attribute) to test for some, I think, edge case? We are already forcing things a little bit by selecting `pdbonly` in the project file which is not the default.
And also this is only happening in ASP.Net and not ASP.Net Core, where we have a [similar snippet](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/88705f5973b1fec9670b8694d72c1ba5e65dfd7a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/ErrorReportingSnippets.cs#L77) that have been passing without being marked with the attribute. 
Anyway, this code is part of a Snippet, and I really think it might be confusing for people if we add the attribute in the Snippet, although, on the other hand, something like this should be documented for users: We log whatever information is available on the StackTrace, ways of getting more information on the StackTrace include configuring your project to generate `pdbonly`, to not `Optimize code`, marking some methods with `[MethodImpl(MethodImplOptions.NoInlining)]`. And even so, the StackTrace might contain different info in ASP.Net or ASP.Net Core, or if your application is executing on a Windows environment or not.
Thoughts on the best ways to do this?

amanda-tarafa:
Closing, solution submitted in #2404

PR #2404, opened 2018-08-03
Fix flaky snippet

jskeet:
The comment wording could definitely be tweaked. It's not clear
whether we should try to give guidance about whether to put the
attribute on or not - the "right choice" is very much customer-based.

(This is an alternative to #2403.)

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.Snippets/ErrorReportingSnippets.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.Snippets/ErrorReportingSnippets.cs
@@ -20,6 +20,7 @@ using System;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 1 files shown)


jskeet:
I'd say that would work well as more documentation on https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.Diagnostics.AspNet/index.html

As Ian's out today, I suggest we merge this (your approval is required) and we can then tweak it and add more docs over time.

amanda-tarafa:
Agreed!

amanda-tarafa:
I was looking to add a little bit of documentation around this, but it is already there, bot for [Asp.Net](https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.Diagnostics.AspNet/index.html) and Asp.Net Core (https://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.Diagnostics.AspNetCore/index.html). Under the Note section.

I think that is enough, we shouldn't go into detail I think.

PR #2406, opened 2018-08-06
Fix flaky test: Prevent GCs while setting up precondition

evildour:


evildour:
With some (possibly complicated) refactoring, it might be possible to get the tree parsed by the test logic and extract the ISymbol which is the key in the table we care about here to keep it alive, but this seemed a lot simpler (if it works...I've never seen this API either). I'll merge now, but if it's still flaky or you'd prefer I fix it the other way, I'll undo this and fix it the other way.

PR #2424, opened 2018-08-16
Fix Bigtable generation

### LINKED ISSUE #2422: Update Bigtable code generator
Since updating to Visual Studio 15.8, the VS code generator isn't working for me. I suspect we need a later set of Roslyn dependencies. Will investigate.

jskeet:
Fixes #2422.


amanda-tarafa:
The last commit makes the build fail.

jskeet:
Urgh - thanks, will look when I can. (Currently drowning in smallish tasks...)

jskeet:
Looks like updating the SDK version may fix this. Will need to check that doesn't break Kokoro as well though.

jskeet:
(Will fix Kokoro if it breaks :)

evildour:
Sorry, just saw this now.

PR #2438, opened 2018-08-21
Fix xmldoc in pubsub

chrisdunelm:


chrisdunelm:
Thanks. I've just double-checked the changes are on the correct overloads, and they are :)

PR #2454, opened 2018-08-27
[Not for submission] [Bigtable] Multi-channel support using Grpc.Gcp

evildour:
Initial work for proper multi-channel support by using [Grpc.Gcp](https://www.nuget.org/packages/Grpc.Gcp/).

Some notes:
* I made a temporary copy of the `Google.Api.Gax.Grpc.ChannelPool` class and made some proposed modifications to it (called out in regions named `NewStuff`). If approved, I will create a separate PR to move these changes, with appropriate comments, into that repo. This also means that Google.Api.Gax.Grpc will need to take a dependency on Grpc.Gcp by the way.
* I added some manual changes to the auto-generated BigtableServiceApiClient.cs file (aside from the namespace change on ChannelPool, which will be reverted when the ChannelPool changes are ported as mentioned above). @chrisdunelm, I'd like to know what you think about making these kinds of changes to the auto-generation logic. I'm thinking we'll have to do something where we mark the methods as being custom implemented and then have them call off to some associated private `...Impl` method that will have to be defined in the manually written partial class. What do you think?
* I've left out most doc comments for now, but will add them before the PR is ready for its final review.


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -30,99 +30,72 @@ namespace Google.Cloud.Bigtable.V2
...
 
         /// <summary>
         /// Asynchronously creates a <see cref="BigtableClient"/>, applying defaults for all unspecified settings,
-        /// and creating a channel connecting to the given endpoint with application default credentials where
+        /// and creating channels connecting to the given endpoint with application default credentials where
         /// necessary.
         /// </summary>
-        /// <param name="clientCreationSettings">Optional. <see cref="ClientCreationSettings"/> specifying how to create
-        /// <see cref="BigtableServiceApiClient"/>s.</param>
+        /// <param name="endpoint"></param>
```
jskeet: These parameters will need documenting.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -30,99 +30,72 @@ namespace Google.Cloud.Bigtable.V2
...
 
         /// <summary>
         /// Synchronously creates a <see cref="BigtableClient"/>, applying defaults for all unspecified settings,
-        /// and creating a channel connecting to the given endpoint with application default credentials where
+        /// and creating channels connecting to the given endpoint with application default credentials where
         /// necessary.
         /// </summary>
-        /// <param name="clientCreationSettings">Optional. <see cref="ClientCreationSettings"/> specifying how to create
-        /// <see cref="BigtableServiceApiClient"/>s.</param>
+        /// <param name="endpoint"></param>
```
jskeet: Documentation again. (And the method below. I won't add any more comments like this.)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -1000,116 +973,39 @@ private static MutateRowsRequest CreateMutateRowRequest(TableName tableName, IEn
...
     }
 
     public partial class BigtableClientImpl
     {
         private readonly string _appProfileId;
-        private readonly BigtableServiceApiClient[] _clients;
-        private int _clientNumber = -1;
+        private readonly BigtableServiceApiClient _client;
 
         // TODO: Add a public constructor after multi-channel support?
```
jskeet: Remove TODO?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -133,4 +187,383 @@ public partial class BigtableServiceApiClientImpl
                 request.IsIdempotent(),
                 "Non-idempotent MutateRows requests are not allowed. Specify a version with all SetCell mutations.");
     }
+
+    // TODO: Move changes into Google.Api.Gax.Grpc.ChannelPool if approved.
```
jskeet: Is this backward-compatible with the current API?

One thing we need to think about is making GAX depend on Grpc.Gcp. I need to look at more details of that package (and possibly discuss with other languages) about that. (It's not well-defined whether GAX is only intended for Google-hosted APIs, and I assume that "GCP" is really "Google-hosted" rather than "Part of Google Cloud Platform" here. (For example, it should be okay for a YouTube API with a gRPC endpoint, right?)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -85,10 +89,60 @@ public partial class BigtableServiceApiSettings
...
+        /// <param name="settings"></param>
+        /// <returns></returns>
+        public static IEnumerable<ChannelOption> CreateChannelOptions(this BigtableServiceApiSettings settings)
+        {
+            var effectiveSettings = settings ?? BigtableServiceApiSettings.GetDefault();
+            var apiConfig = new ApiConfig
+            {
+                ChannelPool = new ChannelPoolConfig
+                {
+                    MaxSize = (uint)effectiveSettings.MaxChannels,
```
jskeet: Would it be worth making the properties uint to avoid the cast?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -133,4 +187,383 @@ public partial class BigtableServiceApiClientImpl
...
+        {
+            public readonly ServiceEndpoint Endpoint;
+            public readonly List<ChannelOption> Options;
+            public readonly int _hash;
+
+            public GcpCallInvokerKey(ServiceEndpoint endpoint, List<ChannelOption> options)
+            {
+                Endpoint = endpoint;
+                Options = options;
+                _hash = EqualityHelpers.CombineHashCodes(
```
jskeet: Any reason to do this? I'd expect GetHashCode to only be called once by Dictionary anyway.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -133,4 +187,383 @@ public partial class BigtableServiceApiClientImpl
...
+                Endpoint.Equals(other.Endpoint) && Options.SequenceEqual(other.Options, ChannelOptionComparer.Instance);
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="endpoint"></param>
+        /// <param name="options"></param>
+        /// <returns></returns>
+        public GcpCallInvoker GetGcpCallInvoker(ServiceEndpoint endpoint, IEnumerable<ChannelOption> options)
```
jskeet: So presumably if we wanted to be able to generate code for an API not on GCP, it could call the existing methods, right?

jskeet:
Oh, and as a reminder to myself: I'm overhauling Spanner resource management at the moment, so I'd like to use something like this... but we'd need channel affinity on a per session basis. I think @chrisdunelm needs something similar for PubSub. We should talk about this :)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Yes, as mentioned in the notes, I left out most doc comments for now (mostly to avoid unnecessary work if things needed to be changed). I'll add them before cleaning up the PR for final review.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: I was going to review all existing "multi-channel" TODOs after this support was already in place since this PR was big enough as-is.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Yes, I made no changes to the existing ChannelPool API.

I'm not sure about whether things can work with any Google-hosted API. @WeiranFang, can Grpc.Gcp be used with non-GCP endpoints?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: I was under the impression that public APIs should not use unsigned integers because they are not CLS-compliant. Is that no longer a concern?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Yeah you're probably right. Just a case of premature optimization.  Removed.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Yes, that's the idea. I guess if this makes the ChannelPool too closely tied to GCP and that worries you, we can even put this stuff in a totally new class. The only downside there is it would mean more changes would be necessary for the auto-generation, since the s_channelPool member would need to have a different type.

evildour:
Asking internally about the timeline for Grpc.Gcp coming out of experimental mode.

As for Spanner, yes I believe the new call invoker should support that. When you create the ApiConfig that gets serialized to the "grpc_gcp.api_config" channel option used by the GcpCallInvoker, in addition to the ChannelPool property, there is also  a Method property, which is a collection of MethodConfig instance that allow you to specify affinity, so you could do something like this:
```csharp
var apiConfig = new ApiConfig
{
    ChannelPool = new ChannelPoolConfig
    {
        MaxSize = ...,
        MaxConcurrentStreamsLowWatermark = ...
    },
    Method =
    {
        new MethodConfig
        {
            Name = { "/google.spanner.v1.Spanner/CreateSession" },
            Affinity = new AffinityConfig
            {
                Command = AffinityConfig.Types.Command.Bind,
                AffinityKey = "name"
            }
        },
        new MethodConfig
        {
            Name = { "/google.spanner.v1.Spanner/GetSession" },
            Affinity = new AffinityConfig
            {
                Command = AffinityConfig.Types.Command.Bound,
                AffinityKey = "name"
            }
        },
        new MethodConfig
        {
            Name = { "/google.spanner.v1.Spanner/DeleteSession" },
            Affinity = new AffinityConfig
            {
                Command = AffinityConfig.Types.Command.Unbind,
                AffinityKey = "name"
            }
        },
        ...
    }
};
```

jskeet:
Fantastic - that's *really* good to know, thanks :)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
WeiranFang: Hi @jskeet and @evildour , the Grpc.Gcp library is only for fixing GCP specific issues. We try to make the library as generic as possible, but the functionality is mainly designed for cloudpath (supported by CFE, a modified version of GFE). So the answer is Grpc.Gcp should be used only for GCP endpoints. Does that answer your question?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
jskeet: It sounds like we may need code-gen to be aware of whether an API is a Cloud API or not. I think that's reasonable, to be honest.

cc @chrisdunelm as he'll be the one doing the work :)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Yes, thank you.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Ok sounds good

jskeet:
I've now created Google.Api.Gax.Grpc.Gcp, so this PR can be modified to refer to that.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -85,10 +89,93 @@ public partial class BigtableServiceApiSettings
...
+        /// <para>
+        /// Note that channels are managed and shared amongst <see cref="BigtableClient"/> instances using the same
+        /// endpoint and channel options and which use the default credentials. So this property will have no bearing
+        /// across clients using different options. To create a custom grouping of channels to be managed, create a
+        /// <see cref="GcpCallInvoker"/> manually and use the
+        /// <see cref="BigtableClient.Create(CallInvoker, BigtableServiceApiSettings, string)">BigtableClient.Create</see>
+        /// overload taking a <see cref="CallInvoker"/> to create clients from it.
+        /// </para>
+        /// </remarks>
+        public int MaxChannels { get; set; } = 20;
```
evildour: @jskeet The recommendation from the Bigtable team is to cap this at 4 channels per CPU. It seems a bit odd though to initialize this to something like `Environment.ProcessorCount * 4` though. I would think we'd probably want to pick some constant default but create some docs around this. Thoughts?

jskeet: I *think* I'd be okay using the processor count - the thread pool used to be set up that way by default, for example. It's obviously not as simple as just using a hard-coded value though. (I'd probably default to 16, so assuming 4 processors.)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
jskeet: Any code generated from protobufs with uint values will expose that anyway. The CLS isn't nearly as important now as it used to be - and even VB have unsigned integer support now, I believe.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: I'm confused. So are you saying you prefer using the processor count * 4 or using hard-coded 16? I'll use 16 for now but let me know if I'm reading this wrong.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -30,99 +30,77 @@ namespace Google.Cloud.Bigtable.V2
 {
     public abstract partial class BigtableClient
     {
-        private static string UserAgent { get; } =
+        internal static string UserAgent { get; } =
             $"cbt-csharp/v{typeof(BigtableClient).GetTypeInfo().Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>().InformationalVersion}";
 
         // TODO: Auto-generate these if possible/easy after multi-channel support is added.
+        // TODO: Other overloads for providing a Channel or CallInvoker?
```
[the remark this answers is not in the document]
evildour: I think this was a TODO from earlier before I added the CallInvoker overload. Removed. I'm not sure about adding the Channel overload though. I don't know if we want to encourage that for Bigtable, and it is simple enough if someone really does want to use a Channel: simply wrap it in a DefaultCallInvoker.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClient.cs
```diff
@@ -321,7 +322,7 @@ public abstract partial class BigtableServiceApiClient
             "https://www.googleapis.com/auth/cloud-platform.read-only",
         });
 
-        private static readonly gaxgrpc::ChannelPool s_channelPool = new gaxgrpc::ChannelPool(DefaultScopes);
+        private static readonly gaxgrpcgcp::GcpCallInvokerPool s_callInvokerPool = new gaxgrpcgcp::GcpCallInvokerPool(DefaultScopes);
```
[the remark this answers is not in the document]
evildour: Correct

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -85,10 +89,93 @@ public partial class BigtableServiceApiSettings
...
+        /// long as fewer than <see cref="MaxChannels"/> are currently open.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Note that channels are managed and shared amongst <see cref="BigtableClient"/> instances using the same
+        /// endpoint and channel options and which use the default credentials. So this property will have no bearing
+        /// across clients using different options. To create a custom grouping of channels to be managed, create a
+        /// <see cref="GcpCallInvoker"/> manually and use the
+        /// <see cref="BigtableClient.Create(CallInvoker, BigtableServiceApiSettings, string)">BigtableClient.Create</see>
+        /// overload taking a <see cref="CallInvoker"/> to create clients from it.
```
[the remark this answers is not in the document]
evildour: Yes, that's right

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -85,10 +89,93 @@ public partial class BigtableServiceApiSettings
...
             ReadRowsRetrySettings = existing.ReadRowsRetrySettings;
+            MaxChannels = existing.MaxChannels;
+            PreferredMaxStreamsPerChannel = existing.PreferredMaxStreamsPerChannel;
+        }
+    }
+
+    /// <summary>
+    /// Extensions for use on <see cref="BigtableServiceApiSettings"/>.
+    /// </summary>
+    public static class BigtableServiceApiSettingsExtensions
```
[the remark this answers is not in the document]
evildour: I intend on making a snippet showing how to create a GcpCallInvoker manually. This extension will help so users don't have to create an ApiConfig and populate the special channel option themselves.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Ok, updated

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClient.cs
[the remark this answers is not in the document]
jskeet: Right. That's going to be fun then. @chrisdunelm - this is something to be aware of, as we'll need to put something in the generator.

For the moment I'm tempted to write a quick shell script to apply these changes after generation. Leave it with me.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -85,10 +89,91 @@ public partial class BigtableServiceApiSettings
...
+        /// <para>
+        /// Note that channels are managed and shared amongst <see cref="BigtableClient"/> instances using the same
+        /// endpoint and channel options and which use the default credentials. So this property will have no bearing
+        /// across clients using different options. To create a custom grouping of channels to be managed, create a
+        /// <see cref="GcpCallInvoker"/> manually and use the
+        /// <see cref="BigtableClient.Create(CallInvoker, BigtableServiceApiSettings, string)">BigtableClient.Create</see>
+        /// overload taking a <see cref="CallInvoker"/> to create clients from it.
+        /// </para>
+        /// </remarks>
+        public uint MaxChannels { get; set; } = 16;
```
jskeet: The comment is elsewhere, but I'm fine with this being 16 by default.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -85,10 +89,91 @@ public partial class BigtableServiceApiSettings
                 retryFilter: IdempotentRetryFilter
             );
 
+        // TODO: Link to snippet for creating GcpCallInvoker manually from here.
+        /// <summary>
+        /// The maximum number of channels than will be open concurrently to the Bigtable endpoint.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// Note that channels are managed and shared amongst <see cref="BigtableClient"/> instances using the same
```
jskeet: Just to check, that means that the first set of options used to create a client with a given endpoint "wins"? It's not ideal, but I can't see any other way of working that isn't more complex. (I have the same kind of issue in Spanner.)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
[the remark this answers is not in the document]
jskeet: Excellent.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -85,10 +89,91 @@ public partial class BigtableServiceApiSettings
...
+        /// and <see cref="BigtableServiceApiSettings.PreferredMaxStreamsPerChannel"/> settings values will be ignored.
+        /// </para>
+        /// </remarks>
+        /// <param name="settings">
+        /// The settings with which to create channel options. May be null, in which case the default settings will be used.
+        /// </param>
+        /// <returns>A collection of <see cref="ChannelOption"/> instances.</returns>
+        public static IEnumerable<ChannelOption> CreateChannelOptions(this BigtableServiceApiSettings settings)
+        {
+            var effectiveSettings = settings ?? BigtableServiceApiSettings.GetDefault();
```
jskeet: It's slightly annoying that there's so much work involved on every client creation. It might be worth having a default channel option array which is returned immediately if `settings` is null, given that's the most likely case. That can definitely come in a follow-up PR though.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: No, the way I have things working in the GcpCallInvokerPool is it will cache based on the pair of endpoint and channel options collection (which is why I manually implemented equality testing for channel options, btw). So if you use a different set of options, which currently just means different values for either of these two properties, you'll get a different GcpCallInvoker.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Good idea. I'll add a TODO.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClient.cs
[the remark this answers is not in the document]
evildour: So are you saying this should me merged as-is? I was thinking we'd first get the code gen changes in place and then I'd update the PR to make use of that, but a temp shell script approach would make it so we could get this in sooner.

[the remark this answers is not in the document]
jskeet: Yup, we can work out a quick fix for now - changes in code generator will be delayed about as we're rewriting it, so let's not block on that.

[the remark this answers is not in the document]
evildour: Ok sounds good. I'll probably wait until Monday to merge.

jskeet:
Thanks Mike - I'll work on the hacky codegen-fixing script.

evildour:
Thanks! And I'll start picking off the other TODOs around this at some point this week.

PR #2457, opened 2018-08-28
New GA release for Storage; new beta for Firestore

jskeet:


jskeet:
I've just noticed that I failed to update the version number for Google.Cloud.Firestore.V1Beta1. I don't think there have been any updates since its previous release, but it means the dependency looks a little odd. I think I'll just leave it, but we can skip a beta number next time we release.

jskeet:
Ah no - the problem was that I *did* update Google.Cloud.Firestore.V1Beta1, but not Google.Cloud.Firestore. Will create a new PR updating both. Sorry about that.

PR #2459, opened 2018-08-28
Correct previous Firestore release mistake

jskeet:
We should update both Google.Cloud.Firestore and
Google.Cloud.Firestore.V1Beta1 at the same time. I only updated the
latter.

The tagging would have failed if we'd updated Google.Cloud.Firestore
without Google.Cloud.Firestore.V1Beta1 due to the project-based
dependency, but not vice versa.


jskeet:
(Note that I can't update *just* Firestore, due to the project reference. I could do it manually, but it's simpler just to skip a version.)

PR #2472, opened 2018-09-04
Adding unit tests for EnsureVisibleDuration

amanda-tarafa:
And fixing and edge case.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 using Google.Cloud.Trace.V1;
-using Xunit;
+using Google.Protobuf.WellKnownTypes;
 using Moq;
 using System;
 using System.Collections.Generic;
@@ -21,7 +21,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-
+using Xunit;
 using TraceProto = Google.Cloud.Trace.V1.Trace;
 
 namespace Google.Cloud.Diagnostics.Common.Tests
@@ -30,6 +30,10 @@ namespace Google.Cloud.Diagnostics.Common.Tests
     {
         private const string TraceId = "trace-id";
         private const string ProjectId = "project-id";
+        // Constanst used to check if a Timestamp is normalized.
+        private const long BclSecondsAtUnixEpoch = 62135596800;
+        internal const long UnixSecondsAtBclMaxValue = 253402300799;
+        internal const long UnixSecondsAtBclMinValue = -BclSecondsAtUnixEpoch;
 
         private static readonly IConsumer<TraceProto> UnusedConsumer = (new Mock<IConsumer<TraceProto>>()).Object;
         private static readonly Dictionary<string, string> EmptyDictionary = new Dictionary<string, string>();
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -79,6 +84,49 @@ private static bool Is1MsSpan(TraceSpan span)
...
         }
 
+        /// <summary>
+        /// Used to ensure that <see cref="Timestamp"/>s manipulated by
+        /// <see cref="SimpleManagedTracer"/> are normalized after
+        /// manipulation.
+        /// </summary>
+        private static bool IsNormalized(Timestamp timestamp) =>
+            timestamp.Nanos >= 0 &&
+            timestamp.Nanos <= MaxNanos &&
```
jskeet: I think I'd just use Duration.NanosecondsPerSecond and make the range exclusive.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -79,6 +84,49 @@ private static bool Is1MsSpan(TraceSpan span)
...
+            timestamp.Seconds <= UnixSecondsAtBclMaxValue;
+
+        [Theory]
+        [InlineData(10, 10, 10, 15)]
+        [InlineData(10, 999000000, 10, 999000015)]
+        [InlineData(10, 999999999, 11, 15)]
+        [InlineData(10, 15, 10, 10)]
+        public void Span_EnsureVisibleDuration(int startSeconds, int startNanos, int endSeconds, int endNanos)
+        {
+            var mockConsumer = new Mock<IConsumer<TraceProto>>();
```
jskeet: What's this mock actually doing? If we don't care about how it responds, I think I'd normally create a "no-op" fake instead of mocking.

jskeet:
I think I'd probably change the existing calls to check that every call's span is *at least* 1ms instead. Deterministic, but still cautious.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
amanda-tarafa: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
amanda-tarafa: Done

amanda-tarafa:
Adressed PR comments, added test from when span is longer than 1ms and changed verification of span length to at least 1ms. PTAL.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/FakeTracer.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// Does nothing
+        /// </summary>
+        /// <param name="items"></param>
+        public void Receive(IEnumerable<TraceProto> items) { }
+
+        /// <summary>
+        /// Returns a completed task.
+        /// </summary>
+        public Task ReceiveAsync(IEnumerable<TraceProto> items, CancellationToken cancellationToken = default) =>
+#if NET452
```
jskeet: I'd just use `Task.FromResult(0)` in both cases, rather than make it conditional.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/FakeTracer.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using TraceProto = Google.Cloud.Trace.V1.Trace;
+
+namespace Google.Cloud.Diagnostics.Common.Tests
+{
+    /// <summary>
+    /// An <see cref="IConsumer{T}"> for testing that does nothing.
+    /// </summary>
+    public class FakeConsumer : IConsumer<TraceProto>
```
jskeet: *Possibly* rename to NoOpConsumer as later we may want a FakeConsumer that remembers all the traces it receives? Although it wouldn't hurt to add that functionality to this one instead...

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
```diff
@@ -73,10 +77,86 @@ private static bool IsValidSpan(TraceSpan span, string name, ulong parentId, Spa
         /// a difference of 1ms between start and end. If we do not the Trace
         /// API will not record the span.
         /// </summary>
-        private static bool Is1MsSpan(TraceSpan span)
+        private static bool IsAtLeast1MsSpan(TraceSpan span)
         {
             var duration = span.EndTime - span.StartTime;
-            return duration.Seconds == 0 && duration.Nanos == 1_000_000;
+            return duration.Seconds >= 1 || duration.Nanos >= 1_000_000;
```
jskeet: To be absolutely sure, probably

```csharp
duration.Seconds >= 1 || (duration.Seconds == 0 && duration.Nanos >= 1_000_000);
```

Otherwise if we had our start and end time the wrong way round we'd allow a negative value to pass.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/FakeTracer.cs
amanda-tarafa: Done, only I did Task.Delay(0) in both cases.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/FakeTracer.cs
amanda-tarafa: I'm renaming to NoOpConsumer. In real code there's no way for clients of IConsumer to access the traces it receives or any information about them, so I'm not sure why we would need a Fake to remember them? Unless we'd want to either verify something or expect some very specific behavior from the fake in which case we would use a Mock or write a specific Fake? Am I missing something?
If I'm wrong we can always add the remembering Fake when it's needed.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/SimpleManagedTracerTest.cs
amanda-tarafa: Yep, will change and add a comment or else that second check on seconds looks redundant.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/FakeTracer.cs
jskeet: We'd want a fake to remember them so that we could check what was passed to the consumer. I typically prefer fakes over mocks as they're less brittle in the face of irrelevant changes, but it would depend on the specific situation. I'm fine with it being NoOpConsumer for now, certainly :)

amanda-tarafa:
Thanks! Waiting on all green before merging.

PR #2492, opened 2018-09-12
[Not for submission] [Bigtable] ListAppProfiles - FailedLocationBehavior enum

evildour:
Docs and integration tests intentionally omitted for now to avoid churn. Will add when we settle on the approach.

### DIFF
--- a/apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2.Snippets/BigtableInstanceAdminClientSnippets.cs
+++ b/apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2.Snippets/BigtableInstanceAdminClientSnippets.cs
@@ -44,7 +44,7 @@ namespace Google.Cloud.Bigtable.Admin.V2.Snippets
             // Make the request and throw if AppProfile information could not be retrieved from any locations
             PagedAsyncEnumerable<ListAppProfilesResponse, AppProfile> response =
                 bigtableInstanceAdminClient.ListAppProfilesAsync(
-                    parent, failedLocationBehavior: FailedLocationBehavior.Throw);
+                    parent, FailedLocationBehavior.Throw);
 
             try
             {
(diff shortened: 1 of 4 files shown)


evildour:
Oh, and I'll add a .sed file for updating the generated file once we agree on the approach as well.

jskeet:
I've been thinking about this a bit more, and it just isn't sitting right with me at the moment.

If the developer is sufficiently aware of the potential issues to say they want to throw exceptions, shouldn't they be aware enough to check FailedLocations themselves, and probably handle it a different way? We're not protecting the unaware developers at all, because of the default behavior.

It's unclear what we really want the user experience to be. I've no idea how other languages are handling this, either.

I suspect we need a VC with both of us and the API team to make progress :(

jskeet:
> Iterating the collection normally uses AsRawResponses under the hood

I can't see how. This is the `GetEnumerator` method:

```csharp
public override IEnumerator<AppProfile> GetEnumerator() => _inner.GetEnumerator();
```

That's going to call the inner `PagedEnumerable.GetEnumerator` which doesn't check anything, isn't it? Or have I misunderstood something?

I *think* it would be good to make the extra parameter required, but I'm still on the fence at the moment.

evildour:
Oops, you're right. I saw that `GrpcPagedEnumerable ` had an implementation of `AsRawResponses().SelectMany(page => page).GetEnumerator()` and I guess I was a little too tired last night...incorrectly thought it would call back into my `AsRawResponses` override. Ok, well the intention was that it would throw with both iteration styles. I'll fix that if we do decide to go with this approach.

evildour:
Ok, I made the parameter non-optional and also made it so normal iteration honors the setting. However, I realized that GetPage should really also be honoring it if any of the responses making up the page have failed locations, and this will take some additional work in gax-dotnet, so I won't do that now unless we decide to go with this approach.

I can't really think of a better way at the moment other than removing the generated paging support, but I'm still leaning towards this way.

evildour:
CC @lukesneeringer, @igorbernstein2, @dmmcerlean

evildour:
(rebased)
@chrisdunelm have you had a chance to think about this one?

jskeet:
Ping @chrisdunelm as we should get this decided pretty soon.

jskeet:
Closing this for now as the design is still under discussion, and unlikely to be resolved imminently.

PR #2495, opened 2018-09-12
Slightly simplify API

chrisdunelm:
As requested by pubsub team.

Rationale for removing public Create(...) method is to not provide a way for users to create under-performing clients

Removing MaxBatchingSettings as it adds complexity with almost no beneifit.


chrisdunelm:
Thanks. No need to talk yet, I'll have a discussion pubsub people about the queuing first.

PR #2508, opened 2018-09-14
[Bigtable] Expose BigtableClientImpl constructor

evildour:


evildour:
Oh, I hadn't considered that. It will probably be cleaner. I'll make the change in a new commit here so we can see how it looks.

evildour:
Actually, it will require a few other changes, but I think it will end up cleaner in the end.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -178,6 +185,11 @@ public static IEnumerable<ChannelOption> CreateChannelOptions(this BigtableServi
 
     public partial class BigtableServiceApiClient
     {
+        /// <summary>
+        /// Gets the default settings used with requests.
+        /// </summary>
+        public virtual BigtableServiceApiSettings DefaultSettings => null;
```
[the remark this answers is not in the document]
evildour: They hold the defaults for the CallSettings if unspecified for a particular call, so I thought that name was appropriate, but I'm open to changing it if you feel otherwise. I'm not sure what `GetDefaultSettings` is though. Are you referring to `BigtableServiceApiSettings.GetDefault()`? I could see that causing confusion, so I could see the argument for renaming it. Also, it might also make sense to make it internal to avoid confusion altogether. I think I'd just need to adjust how the mocks are setup to return it.

jskeet:
Yes, I definitely think it should be internal if possible. IIRC, you can ask Moq to treat properties like a stub, so you can just set the property on the mock, and the code fetching it will see that value.

But could the code that applies the AppProfileId be part of BigtableServiceApiClient instead of BigtableClient, when modifying requests? That would be nice in terms of it being part of BigtableServiceApiSettings. We definitely *can* modify the requests there - is there any reason we'd particularly not want to? We could expose the AppProfileId as a property for inspection in BigtableServiceApiClient too (and delegate to it in BigtableClient). I think it would be worth copying it from the settings on construction rather than letting it change over time.

As an aside, it's unclear to me whether it's worth getting the first commit of this PR merged and then deciding the AppProfileId part separately. If you want to do that, I'm fine with it - but they may be sufficiently interlinked that it's not worth it.

evildour:
Oh yeah, that's much cleaner. Thanks!

I think I'd prefer to do it all now because they are quite interlinked. And I'd rather not change and expose the ...Impl constructor just to then change it again.

evildour:
PTAL

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
```diff
@@ -81,7 +81,8 @@ public static async Task<BigtableClient> CreateAsync(ServiceEndpoint endpoint =
             Create(BigtableServiceApiClient.Create(GaxPreconditions.CheckNotNull(callInvoker, nameof(callInvoker)), settings));
 
         /// <summary>
-        /// Gets the value which specifies routing for replication. If null, the "default" application profile will be used by the server.
+        /// Gets the value which specifies routing for replication.
+        /// If null or emtpy, the "default" application profile will be used by the server.
```
jskeet: emtpy => empty

jskeet: (I'm not sure whether returning null is appropriate here, or throwing an exception like our other virtual members do. Any thoughts either way?)

jskeet: (Same comments in BigtableApiServiceClient)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -216,28 +219,71 @@ public partial class SampleRowKeysStream
...
+
     public partial class BigtableServiceApiClientImpl
     {
         private BigtableServiceApiSettings _defaultSettings;
 
         /// <inheritdoc/>
-        public override BigtableServiceApiSettings DefaultSettings => _defaultSettings;
+        public override string AppProfileId => _defaultSettings.AppProfileId;
+
+        internal override BigtableServiceApiSettings DefaultSettings => _defaultSettings;
```
jskeet: I suspect this can just be

```csharp
internal override BigtableServiceApiSettings DefaultSettings { get; }
```

and assign to it from the constructor, with no expicitly-declared backing field.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -216,28 +219,71 @@ public partial class SampleRowKeysStream
...
+            Func<TRequest, string> appProfileIdGetter,
+            Action<TRequest, string> appProfileIdSetter)
+        {
+            if (AppProfileId != null && appProfileIdGetter(request).Length == 0)
+            {
+                appProfileIdSetter(request, AppProfileId);
+            }
+        }
+
+        partial void Modify_CheckAndMutateRowRequest(ref CheckAndMutateRowRequest request, ref CallSettings settings)
```
jskeet: Possibly use expression-bodied members for these where possible? https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/apis/Google.Cloud.Firestore.V1Beta1/Google.Cloud.Firestore.V1Beta1/FirestoreClientPartial.cs is very compact, for example? Definitely not a big deal though :)

tools/Google.Cloud.ClientTesting/NotImplementedCallInvoker.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <summary>
+    /// CallInvoker which throws NotImplementedException for all calls.
+    /// </summary>
+    public sealed class NotImplementedCallInvoker : CallInvoker
+    {
+        public static readonly NotImplementedCallInvoker Instance = new NotImplementedCallInvoker();
+
+        private NotImplementedCallInvoker() { }
+
+        public override AsyncClientStreamingCall<TRequest, TResponse> AsyncClientStreamingCall<TRequest, TResponse>(Method<TRequest, TResponse> method, string host, CallOptions options)
```
jskeet: Expression-bodied methods?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableClientPartial.cs
evildour: Fixed. Yeah perhaps throwing is better.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: No, we don't have the constructor here. Only an OnConstruction partial. Same reason why the backing field can't be readonly.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
evildour: Right, meant to do this and forgot.

tools/Google.Cloud.ClientTesting/NotImplementedCallInvoker.cs
evildour: Done

evildour:
Thanks Jon! Yes, I like the way this turned out. I'll merge on green.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
jskeet: Oops - apologies, I'm a muppet.

jskeet:
Thanks for all the back and forth - I think we've come to a better solution than either of us would have reached on our own :)

PR #2509, opened 2018-09-14
[Bigtable] Use a cached collection of channel options for null settings

evildour:


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -134,6 +134,10 @@ public partial class BigtableServiceApiSettings
     /// </summary>
     public static class BigtableServiceApiSettingsExtensions
     {
+        private static readonly Lazy<IEnumerable<ChannelOption>> s_defaultChannelOptions =
+            new Lazy<IEnumerable<ChannelOption>>(
+                () => BigtableServiceApiSettings.GetDefault().CreateChannelOptions());
```
jskeet: This is returning an array - it could be cast back to an array and modified. Would be weird, but let's be careful :) Creating a `ReadOnlyCollection` wrapping it should be enough to make it immutable.

evildour: Done

PR #2512, opened 2018-09-16
Collapse whitespace in header values included in signed URLs' signature.

evildour:


evildour:
Yes, I sent an email about it internally. Not sure what changed exactly unfortunately.

PR #2515, opened 2018-09-17
(Not for submission yet) First pass at Datastore emulator support

### LINKED ISSUE #2499: Add support for Datastore emulator
There's no imminent plan for a non-environment-variable-based emulator, so we should make it easier.

It will still be explicitly opt-in, and we probably only need it on `DatastoreDb.Create`.

jskeet:
Currently no tests - they will be relatively hard to write in the
current form. We'll work out whether to refactor for testability or
not when we've agreed on the design.

Will fix #2499 when we've eventually settled on the design. Will ask for Datastore team input too.

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
@@ -14,6 +14,7 @@
 
 using Google.Api.Gax;
 using Google.Api.Gax.Grpc;
+using Grpc.Core;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -82,6 +83,61 @@ namespace Google.Cloud.Datastore.V1
         /// <returns>A <see cref="DatastoreDb"/> operating on the specified partition.</returns>
         public static DatastoreDb Create(string projectId, string namespaceId = "", DatastoreClient client = null) =>
             new DatastoreDbImpl(projectId, namespaceId, client ?? DatastoreClient.Create());
+
+        /// <summary>
+        /// Creates a <see cref="DatastoreDb"/> by either connecting to the production servers using the
+        /// default application credentials, or by connecting to a local emulator. The behavior depends on
+        /// the DATASTORE_EMULATOR_HOST and DATASTORE_PROJECT_ID environment variables, with further control
+        /// provided by the <paramref name="environmentBehavior"/> parameter.
+        /// </summary>
+        /// <param name="environmentBehavior">The way to behave in different environments. See the enum
+        /// definition for more details.</param>
+        /// <param name="projectId">The project ID to use when connecting to production servers. Ignored
+        /// when connecting to the emulator. Must not be null for production scenarios.</param>
+        /// <param name="namespaceId">The namespace ID to use in operations requiring a partition. Must not be null.</param>
+        /// <exception cref="InvalidOperationException">The environment was not configured appropriately
+        /// for the requested behavior.</exception>
+        /// <returns>A <see cref="DatastoreDb"/> operating on the specified partition, either connected
+        /// to production servers or the Datastore emulator.</returns>
+        public static DatastoreDb CreateWithEnvironment(EnvironmentBehavior environmentBehavior, string projectId, string namespaceId = "")
+        {
+            GaxPreconditions.CheckEnumValue(environmentBehavior, nameof(environmentBehavior));
+            string emulatorHost = Environment.GetEnvironmentVariable("DATASTORE_EMULATOR_HOST");
+            string emulatorProjectId = Environment.GetEnvironmentVariable("DATASTORE_PROJECT_ID");
+            bool hasEmulatorHost = !string.IsNullOrEmpty(emulatorHost);
+            bool hasEmulatorProjectId = !string.IsNullOrEmpty(emulatorProjectId);
+
+            bool useEmulator;
+            switch (environmentBehavior)
+            {
+                case EnvironmentBehavior.ProductionOnlyThrowOnEmulatorEnvironment:
+                    GaxPreconditions.CheckState(!hasEmulatorHost && !hasEmulatorProjectId,
+                        "DATASTORE_EMULATOR_HOST and/or DATASTORE_PROJECT_ID environment configured unexpectedly.");
+                    useEmulator = false;
+                    break;
+                case EnvironmentBehavior.ProductionOnlyIgnoreEnvironment:
+                    useEmulator = false;
+                    break;
+                case EnvironmentBehavior.EmulatorOnly:
+                    GaxPreconditions.CheckState(hasEmulatorHost && hasEmulatorProjectId,
+                        "DATASTORE_EMULATOR_HOST and DATASTORE_PROJECT_ID environment must be configured.");
+                    useEmulator = true;
+                    break;
+                case EnvironmentBehavior.ProductionOrEmulator:
+                    GaxPreconditions.CheckState(!(hasEmulatorHost ^ hasEmulatorProjectId),
+                        "Exactly one of DATASTORE_EMULATOR_HOST and DATASTORE_PROJECT_ID environment configured. Either both or neither should be configured.");
+                    useEmulator = hasEmulatorHost;
+                    break;
+                default:
+                    // We shouldn't get here, due to the first line of the method.
+                    throw new InvalidOperationException($"EnvironmentBehaviour value {environmentBehavior} encountered unexpectedly.");
+            }
+            DatastoreClient client = useEmulator
+                ? DatastoreClient.Create(new Channel(emulatorHost, ChannelCredentials.Insecure))
+                : DatastoreClient.Create();
+            // This will validate projectId and namespaceId
+            return Create(useEmulator ? emulatorProjectId : projectId, namespaceId, client);
+        }
 
         /// <summary>
         /// Creates a key factory for root entities in this objects's partition.
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/EnvironmentBehavior.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/EnvironmentBehavior.cs
@@ -1,0 +1,56 @@
+﻿// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Google.Cloud.Datastore.V1
+{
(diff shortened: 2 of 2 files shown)


apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
```diff
@@ -83,6 +84,61 @@ public abstract class DatastoreDb
...
+            string emulatorHost = Environment.GetEnvironmentVariable("DATASTORE_EMULATOR_HOST");
+            string emulatorProjectId = Environment.GetEnvironmentVariable("DATASTORE_PROJECT_ID");
+            bool hasEmulatorHost = !string.IsNullOrEmpty(emulatorHost);
+            bool hasEmulatorProjectId = !string.IsNullOrEmpty(emulatorProjectId);
+
+            bool useEmulator;
+            switch (environmentBehavior)
+            {
+                case EnvironmentBehavior.ProductionOnlyThrowOnEmulatorEnvironment:
+                    GaxPreconditions.CheckState(!hasEmulatorHost && !hasEmulatorProjectId,
```
amanda-tarafa: Shoudn't this be `hasEmulatorHost || hasEmulatorProjectId`

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
```diff
@@ -83,6 +84,61 @@ public abstract class DatastoreDb
...
+                        "DATASTORE_EMULATOR_HOST and/or DATASTORE_PROJECT_ID environment configured unexpectedly.");
+                    useEmulator = false;
+                    break;
+                case EnvironmentBehavior.ProductionOnlyIgnoreEnvironment:
+                    useEmulator = false;
+                    break;
+                case EnvironmentBehavior.EmulatorOnly:
+                    GaxPreconditions.CheckState(!hasEmulatorHost && !hasEmulatorProjectId,
```
chrisdunelm: Should be `hasEmulatorHost && hasEmulatorProjectId`

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/EnvironmentBehavior.cs
```diff
@@ -0,0 +1,52 @@
...
+    /// behaves in different scenarios.
+    /// </summary>
+    public enum EnvironmentBehavior
+    {
+        /// <summary>
+        /// Always connect to the production servers, regardless of environment
+        /// variables. This is equivalent to the <see cref="DatastoreDb.Create"/>
+        /// call.
+        /// </summary>
+        ProductionOnlyIgnoreEnvironment = 0,
```
amanda-tarafa: `ProductionOnlyIgnoreEmulatorEnviroment`?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
```diff
@@ -83,6 +84,61 @@ public abstract class DatastoreDb
...
+        /// <param name="environmentBehavior">The way to behave in different environments. See the enum
+        /// definition for more details.</param>
+        /// <param name="projectId">The project ID to use when connecting to production servers. Ignored
+        /// when connecting to the emulator. Must not be null for production scenarios.</param>
+        /// <param name="namespaceId">The namespace ID to use in operations requiring a partition. Must not be null.</param>
+        /// <exception cref="InvalidOperationException">The environment was not configured appropriately
+        /// for the requested behavior.</exception>
+        /// <returns>A <see cref="DatastoreDb"/> operating on the specified partition, either connected
+        /// to production servers or the Datastore emulator.</returns>
+        public static DatastoreDb CreateWithEnvironment(EnvironmentBehavior environmentBehavior, string projectId, string namespaceId = "")
```
amanda-tarafa: I think I'd made `projectId` optional and then also check its value depending on the behaviour. I think this might make things more symetrical, since we would need `EmulatorOnlyThrowIfProduction` and `EmulatorOnlyIgnoreProduction`.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/EnvironmentBehavior.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        /// <summary>
+        /// Always connect to the emulator, throwing an exception if the environment
+        /// variables are not configured.
+        /// </summary>
+        EmulatorOnly = 2,
+
+        /// <summary>
+        /// Connect to the emulator if suitable environment variables are present,
+        /// or production otherwise.
```
chrisdunelm: Add a warning here stating that this could result in accidental data damage if emulator env variables are accidentally absent during, for example, testing?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
jskeet: We already only check its value based on whether we want to use it or not, but yes, we could make it optional.

EmulatorOnlyIgnoreProduction doesn't make sense to me as an option, because "production environment" means "no emulator environment variables set" which means we don't know how to talk to the emulator.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
jskeet: No, the condition argument is what we expect to be true. If the condition is false, we throw the exception.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/EnvironmentBehavior.cs
jskeet: Possibly. I wanted to keep it terse but explicit. We can definitely iterate on the names when we've worked out the desired behavior.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
jskeet: Whoops, yes! Will fix now.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/EnvironmentBehavior.cs
jskeet: Yes, that's probably a good idea.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
amanda-tarafa: I think "production environment" means "no emulator environment variables set" and "projectId set" (which we are checking _if_ we use it to create the `DatastoreDb`). So there's a chance that someone passes a value for projectId but ends up using the Emulator because the env variables are set, without it ever failing. I think I'd like it to fail in that case (unless EmulatorOnlyIgnoreProduction) or else I find it confussing specially because for using the production environment we do have the ignore and throw options. But this might just be me :)

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
jskeet: Yup, fixed that now.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDb.cs
jskeet: Hmm... it makes it harder if the production project ID is a constant though... it means there are then two things to change when we want to use the emulator, rather than only one. It may be worth chatting more about this over a VC. We certainly *could* fail if the `projectId` is specified but we only want to use the emulator.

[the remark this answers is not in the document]
jskeet: I suspect that'll be tricky if you want your tests and your production code to go through the same path though, with only data changing. Imagine you have a config file which has the enum value in, for example... or a config class with the enum as another property. How would you use that in the EnvironmentSettings approach?

I suspect there are possibilities here though... I'll create another PR with this to see how it looks.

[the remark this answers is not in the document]
jskeet: But then they've got to have different actual code, rather than just data. At that point, we could just have four different methods. The intention here was to allow the *data* to drive what's going on - so all you'd need is different configuration files saying which mode to use, for example. No conditional code required at all.

What kind of error are you trying to avoid using the builder pattern here?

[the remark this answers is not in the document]
simonz130: In the current state developer needs to understand how the correlation of _environmentBehavior_ enum with the environment variables (or their omission) will change the behavior of the app. So it's like analyzing the state machine to decide whether given combination does what developer expects. This is not intuitive DevX. 
Having two separate methods will simplify things (at the expense of having two separate code paths on two environments).
Having the same method with **EnvironmentSettingsBuilder** will simplify things as developer can declaratively decide what to use as opposed to relying on a state machine that they need to understand and analyze. The downside here is that there may be two different code paths to create the **EnvironmentSettings**, but the actual library call from developer's standpoint is the same.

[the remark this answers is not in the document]
jskeet: I take issue with "it's like analyzing the state machine" - it's reading a set of descriptions which can be reasonably clearly described and named. I think that's likely to be just as intuitive as - or more than - the environment builder, to be honest.

You say "The downside here is that there may be two different code paths to create the EnvironmentSettings" - but that's exactly the problem with having two different methods to start with. You're not eliminating that, you're just moving it slightly, and I doubt that it'll be moving it enough. I'd really like there to be a solution that can use the same code path for all desired behaviors, driven only by data.

Of course we could provide *both* options potentially, but I don't *think* we want to do that.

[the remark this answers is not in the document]
simonz130: I agree - having both options will only confuse the developer. It should be one way that is simple to use.

As to which one - with the builder or this PR implementation - both will work, but I think builder pattern is more intuitive for developers who are used to certain development style (.NET Framework BCL), where you have connection string builders, string builders, configuration builders, etc. 
With builders the power and responsibility is in user's hands. With state machine, the power is in library's hands but the responsibility is still in user's hands.
I think that constructing configuration in two ways for two environments is acceptable as long as the resulting API call is the same.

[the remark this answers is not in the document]
jskeet: I'm still not convinced, but I'll create another PR to see what it looks like in a more concrete form. (I also still think referring to a state machine is over the top when discussing this. It's just method parameters, one of which is an enum. No state to keep track of. Arguably the builder approach is *more* of a state machine.)

jskeet:
Closing this as I'm now working on a more generalized builder approach - not just for Datastore but for everything. It'll be a while before it's ready, and will require quite a bit of internal review as well.

PR #2522, opened 2018-09-19
Regenerate APIs (excluding Spanner)

jskeet:
These are more significant changes than normal, so need more review.

- Monitoring and DLP: It's not clear where the "NoRetryRetryFilter" is coming from. It's probably worth checking that we actually want this, as removing it later would be a breaking change.

- PubSub: This looks like renamed filters. Given that we're so close to GA, again I'd like this checked carefully.

- Bigtable: It looks like some backoffs have been removed. This *would* have been a breaking change if we'd gone to GA. Mike, can you check this?

- Redis.V1: Trivial change in comments

### DIFF
--- a/apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClient.cs
+++ b/apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClient.cs
@@ -325,7 +325,6 @@ namespace Google.Cloud.Bigtable.Admin.V2
         /// </list>
         /// Retry will be attempted on the following response status codes:
         /// <list>
-        /// <item><description><see cref="grpccore::StatusCode.DeadlineExceeded"/></description></item>
         /// <item><description><see cref="grpccore::StatusCode.Unavailable"/></description></item>
         /// </list>
         /// Default RPC expiration is 600000 milliseconds.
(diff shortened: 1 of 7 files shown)


chrisdunelm:
Bigtable changes caused by https://github.com/googleapis/googleapis/commit/59036b030f4a02b1eef2425d76064992b50c4d9e#diff-05c73320211201124119d8c88eb6d5d2
Original change description:

Updating timeouts for Cloud Bigtable admin to larger numbers.
CreateTable was timing out in a flaky way for a test that creates 2 clusters, and then does a create table.  See https://github.com/GoogleCloudPlatform/google-cloud-python/issues/5928 for more context.
While I was at it, I noticed a bunch of non-idempotent methods that were incorrectly labeled, including create table.  I made a bunch of changes which ought to be considered.

chrisdunelm:
The DLP `NoRetryRetryFilter` and `HttpGetRetryFilter` is odd. I don't see where these are coming from, as there's no reference to them in: https://github.com/googleapis/googleapis/blob/master/google/privacy/dlp/v2/dlp_gapic.yaml

chrisdunelm:
Likewise for monitoring, there's no reference to `NoRetryRetryFilter` in: https://github.com/googleapis/googleapis/blob/master/google/monitoring/v3/monitoring_gapic.yaml

jskeet:
Yup, the lack of changes in those APIs concerned me too.
I suspect [this change](https://github.com/googleapis/gapic-generator/commit/6b7484e8a6a694dec54f30e1b06abb2753c26893) is responsible - we should probably consider it carefully before merging these.

chrisdunelm:
Yes that is responsible. That change is due to the move to proto-annotations, away from gapic-config.yaml. This fully explains the DLP, monitoring and pubsub changes.

The way retry is specifying is changing, and HTTP GET RPC methods now need to default to an implicit idempotent retry. Java is re-using the existing gapic codegen for their proto-annotation work, hence changes like this appearing.
It looks like the extra retry code is simply being added, without regard to whether it's needed or not.

I suggest these changes are not merged until this is resolved. The BigTable and Redis changes are different, so perhaps should be split out into a different PR.

jskeet:
Righto, will split this into three:

- Redis, which can be merged trivially
- Bigtable, which is probably fine but which I'd like @evildour to look at (and decide whether we should do another beta release for)
- DLP/Monitoring/PubSub

chrisdunelm:
sgtm

jskeet:
Split done - closing this PR.

PR #2523, opened 2018-09-19
Spanner DML

jskeet:
This is not ready for submission quite yet, as the public endpoint doesn't have it enabled.

It's all been reviewed already, but another once-over wouldn't hurt...

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
@@ -1,0 +1,454 @@
+﻿// Copyright 2018 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Spanner.Data.CommonTesting;
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.IntegrationTests
+{
+    /// <summary>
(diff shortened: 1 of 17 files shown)


jskeet:
I've just thought of an option to make this review simpler: I could create a separate PR just for the codegen, then rebase off that. Then this PR would *only* contain the manual code changes. Would that be better for you?

amanda-tarafa:
Yest, that might be better I think. Thanks!

jskeet:
Done - this is a slightly more sensible PR now :)

amanda-tarafa:
Thanks, looking now!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -0,0 +1,484 @@
...
+            Assert.Equal(expected, actual);
+        }
+
+        [Fact]
+        public void InvalidQuery()
+        {
+            string key = CreateTestRows();
+            using (var connection = _fixture.GetConnection())
+            {
+                // Slightly odd choice here: "OriginalValue" has to be unique, so we end up with the inserted Value being the real original value.
```
amanda-tarafa: I think this comment is out of place here.

jskeet: Nope, that's deliberate - basically "original value" doesn't mean what one might expect it to.
Happy to reword the comment though.

amanda-tarafa: I just don't understand it in this test. It's in the test above and there it makes sense, but in this one?

jskeet: In other tests, OriginalValue is the value the row was created with, and Value is a modified value.

In this case we're *inserting* new rows based on existing rows - so they can't share the same OriginalValue.

amanda-tarafa: I think you are referring to the above test, the one in line 189 called InsertWithQuery, and there I get the comment.
But it being here in this test called InvalidQuery, in which you are just using an invalid query to test that it throws, I think makes no sense. I think this was just a copy paste thing?

jskeet: Aha! Yes, sorry. Yup, this was indeed just a C&P fail. I'll remove the comment.

PR #2526, opened 2018-09-19
Regenerate Bigtable API

jskeet:
Background:

The changes are from https://github.com/googleapis/googleapis@59036b0#diff-05c73320211201124119d8c88eb6d5d2
Original change description:

Updating timeouts for Cloud Bigtable admin to larger numbers.
CreateTable was timing out in a flaky way for a test that creates 2 clusters, and then does a create table. See GoogleCloudPlatform/google-cloud-python#5928 for more context.
While I was at it, I noticed a bunch of non-idempotent methods that were incorrectly labeled, including create table. I made a bunch of changes which ought to be considered.


jskeet:
Assuming we're happy with this, it's probably worth performing another Bigtable release to get the breaking change out of the way.

jskeet:
Righto - I'll merge then create a new release. Thanks for the speedy review!

PR #2527, opened 2018-09-19
Regenerate APIs with retry changes

jskeet:
The changes are due to [this GAPIC generator change](https://github.com/googleapis/gapic-generator/commit/6b7484e8a6a694dec54f30e1b06abb2753c26893), which is part of the proto-configuration-based migration.


chrisdunelm:
So these changes are completely unnecessary (ie they add no functionality); and for C# add to the public surface (other languages don't publicly expose these retry definitions).
If we merge this, then when we switch to the new C# code-gen we will be forced to manually write the code to define these (unused, pointless) properties. We'll already be needed to hand-write partials for all existing code anyway, so this doesn't massively increase the work we will need to do, but it will add yet more unused code in the new code-gen that will be there purely for backwards compatibility.

jskeet:
Right, sounds like we don't want to merge. How feasible is it to make the C# generator *not* include these?

chrisdunelm:
At the moment I don't know. And I really don't want to have to find out :(

chrisdunelm:
But I expect it's possible.

jskeet:
Closing as we definitely don't want these changes.

PR #2540, opened 2018-09-27
Kokoro codecov integration.

amanda-tarafa:
(Google internal, this PR is paired with CL 214746599)

### DIFF
--- a/.kokoro/runintegrationtests_codecov.sh
+++ b/.kokoro/runintegrationtests_codecov.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+set -e
+
+SCRIPT=$(readlink -f "$0")
+SCRIPT_DIR=$(dirname "$SCRIPT")
+
+cd $SCRIPT_DIR
+cd ..
+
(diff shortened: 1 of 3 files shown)


.kokoro/runintegrationtests_codecov.sh
```diff
@@ -0,0 +1,38 @@
...
+# Build the libraries, run unit tests, and upload unit tests coverage report.
+./build.sh --notests
+./runcoverage.sh
+./createcoveragereport.sh
+if [[ -f "coverage/coverage-filtered.xml" ]]
+then
+  codecov -f "coverage/coverage-filtered.xml" --flag unittests --c $KOKORO_GITHUB_COMMIT --b KOKORO_BUILD_NUMBER
+fi
+
+# Removing all code coverage files generated for unit tests before generating the ones for integration tests.
```
jskeet: My intention was that the integration test coverage would include the unit test coverage, but I could be persuaded otherwise.

(I can see benefits in both ways. It's easier to have a threshold of testing that includes both, but it could be useful to see what's only covered by unit tests...)

PR #2543, opened 2018-09-28
Adding unittest flag to coverage report sent from AppVeyor.

amanda-tarafa:
Uploading unit tests and integration tests exclusive reports from Kokoro.

### DIFF
--- a/appveyor.sh
+++ b/appveyor.sh
@@ -15,7 +15,7 @@ then
   ./createcoveragereport.sh
   if [[ -f "coverage/coverage-filtered.xml" ]]
   then
-    codecov -f "coverage/coverage-filtered.xml"
+    codecov -f "coverage/coverage-filtered.xml" --flag unittests
   fi
 else
   # We're in a pull request. Don't do any coverage; just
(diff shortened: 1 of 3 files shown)


runintegrationtests.sh
```diff
@@ -21,6 +21,7 @@ while (( "$#" )); do
   elif [[ "$1" == "--coverage" ]]
   then
     install_dotcover
+	mkdir -p coverage
```
jskeet: Outdent

appveyor.sh
```diff
@@ -15,7 +15,7 @@ then
   ./createcoveragereport.sh
   if [[ -f "coverage/coverage-filtered.xml" ]]
   then
-    codecov -f "coverage/coverage-filtered.xml"
+    codecov -f "coverage/coverage-filtered.xml" --flag unittests
```
jskeet: As a thought, I wonder whether we should put this logic into createcoveragereport - it would certainly simplify other scripts. Not something for this PR, but let me know if you think it's a good idea.

amanda-tarafa: Definitely, one always comes after the other, and actually on createcoveragereport there's a TODO that says something like "upload report".
I'll send you another PR with that change.

runintegrationtests.sh
amanda-tarafa: Done

PR #2545, opened 2018-09-28
Moving upload coverage report logic to createcoveragereport.sh

amanda-tarafa:


createcoveragereport.sh
```diff
@@ -37,4 +37,8 @@ $REPORTGENERATOR \
    -targetdir:coverage/report \
    -verbosity:Error
 
-# TODO(benwu): upload coverage/report to public blob storage
+choco install codecov
+
+# Assume we've created the coverage file by this point. If we haven't, there should already have been an error.
+# Pass whatever parameters we recieved.
+codecov -f "coverage/coverage-filtered.xml" "$@"
```
jskeet: I don't think we want to do this unconditionally.

I'd suggest accepting an `--upload` parameter that specifies the flag name. (It would be nice to limit the codecov uninformative name of "flag" for the report name.)

.kokoro/runintegrationtests.sh
```diff
@@ -10,9 +10,7 @@ maybe_upload_coverage() {
     return 0
   fi
   
-  ./createcoveragereport.sh
-  # Assume we've created the coverage file by this point. If we haven't, there should already have been an error.
-  codecov -f "coverage/coverage-filtered.xml" --flag $1 --c $KOKORO_GITHUB_COMMIT --b $KOKORO_BUILD_NUMBER
+  ./createcoveragereport.sh --flag $1 --c $KOKORO_GITHUB_COMMIT --b $KOKORO_BUILD_NUMBER
```
jskeet: If we make createcoveragereport.sh just succeed with a message of "no coverage detected" or similar when there's no coverage directory, we could remove this whole function, simplifying things quite a bit.

amanda-tarafa: Oh but what we are checking for here is the codecov upload token not the coverage directory. The token won't be present when running from AppVeyor so we can't move the check for the token to createcoveragereport.sh unless we pass another parameter indicating that it's running from Kokoro?
Actually, if we pass a param indicating that we are running from Kokoro, we can both move the check for the token and the passing of the $KOKORO... env vars to createcoveragereport.sh.
Sounds good?

jskeet: If we're explicitly passing a --upload or similar flag to createcoveragereport.sh, I think we should assume that we know what we're doing and the environment is already set up correctly. I would rather we kept anything Kokoro-specific in the Kokoro scripts.

I don't know whether I'm being clear in my description - I'm happy to create a PR for what I mean if it would be simpler :)

amanda-tarafa: Yes, I see, we are only running coverage if the token is detected anyways.

amanda-tarafa:
@jskeet PTAL, I've addressed your comments.
Also, I exported the CODECOV_TOKEN, something that we missed yesterday.
Thanks!

createcoveragereport.sh
```diff
@@ -37,4 +58,8 @@ $REPORTGENERATOR \
    -targetdir:coverage/report \
    -verbosity:Error
 
-# TODO(benwu): upload coverage/report to public blob storage
+choco install codecov
```
jskeet: This should be conditional - see earlier comment.
Basically, we still want to be able to run createcoverage.sh locally without uploading, but we want it to have an optional final step.

.kokoro/runintegrationtests.sh
```diff
@@ -26,31 +13,32 @@ export GOOGLE_APPLICATION_CREDENTIALS="$KOKORO_KEYSTORE_DIR/73609_cloud-sharp-je
...
+  script_flags=--coverage
+  report_flags=--commit $KOKORO_GITHUB_COMMIT --build $KOKORO_BUILD_NUMBER
 fi
 
 # Build the libraries and run unit tests, optionally with coverage.
-./build.sh $SCRIPT_FLAGS
+./build.sh $script_flags
 
-maybe_upload_coverage unittests
+./createcoveragereport.sh --upload unittests $report_flags
```
jskeet: We only want to do the upload when codecov-token exists.

In order to simplify this, we probably want *4* flags to createcoveragereport.sh:

--upload: not followed by anything. (It's just a boolean toggle) This is the "master switch"; if this isn't flipped, we don't do anything
--upload_commit - the commit
--upload_build - the build number
--upload_reportname - the report name (--flags)

So:
- Set report_flags to --upload_commit $KOKORO_GITHUB_COMMIT --upload_build $KOKORO_BUILD_NUMBER unconditionally
- In the if block a few lines above, set report_flags to "$report_flags --upload"
- The createcoveragereport.sh calls pass in `$report_flags --upload_reportname unittests` (or integrationtests)

amanda-tarafa: Ah yes, in case we want at some point to generate the report without uploading, maybe locally for instance. Yep, get it.

jskeet: And on Unix, where dotCover doesn't work :)

amanda-tarafa: True, neither does codecov.exe, totally forgot about that. Pushing now.

amanda-tarafa:
@jskeet PTAL (again :)) Thanks!

.kokoro/runintegrationtests.sh
```diff
@@ -26,31 +13,32 @@ export GOOGLE_APPLICATION_CREDENTIALS="$KOKORO_KEYSTORE_DIR/73609_cloud-sharp-je
...
 # Allow each integration test 3 chances to pass.
-./runintegrationtests.sh $SCRIPT_FLAGS || true
-./runintegrationtests.sh $SCRIPT_FLAGS --retry || true
-./runintegrationtests.sh $SCRIPT_FLAGS --retry
-
-maybe_upload_coverage integrationtests
+./runintegrationtests.sh $script_flags || true
+./runintegrationtests.sh $script_flags --retry || true
+./runintegrationtests.sh $script_flags --retry
+# Even if se set up here some upload report flags, if --upload is not present we won't upload the report.
```
jskeet: Nit: se => we

amanda-tarafa: Done

amanda-tarafa:
Thanks!

PR #2547, opened 2018-10-01
Fixup `SubscriberClient.StopAsync` when Token is Already Cancelled

iwillspeak:
If the token passed in as the 'hard stop' token is already cancelled
then the registration will never fire and the client will peform a
soft stop instead. This can take quite some time.

The fix is to check _after_ we have registered on the token if the
token has been cancelled, and if so cancel the hard stop token source
too.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
@@ -452,6 +452,10 @@ namespace Google.Cloud.PubSub.V1
             }
             _globalSoftStopCts.Cancel();
             var registration = hardStopToken.Register(() => _globalHardStopCts.Cancel());
+            if (hardStopToken.IsCancellationRequested)
+            {
+                _globalHardStopCts.Cancel();
+            }
             // Do not register this Task to be awaited on at shutdown.
             // It completes *after* _mainTcs, and all registered tasks must complete before _mainTcs
             _taskHelper.Run(async () =>


chrisdunelm:
This should already be tested in the [ImmediateStop](https://github.com/googleapis/google-cloud-dotnet/blob/fd37079499cef93e96cfce35a98fbf4cc4058e8e/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs#L341) test; so I'm slightly surprised that this might be a problem. I thought the `Register` method called the provided function immediately if the token is already cancelled.
I'll do a little more checking.

iwillspeak:
Hmm. Looking at the docs for `Register` it does say it should call it. I've seen the client take a long time to stop if the token was cancelled before StopAsync was called though and fixed it with something similar to this code outside of `StopAsync`.

chrisdunelm:
I just tried this code on .NET Core and Full Framework, both on Windows, and in both cases the registered function was called:
```C#
var token = new CancellationToken(true);
token.Register(() => Console.WriteLine("Called!"));
Thread.Sleep(500);
Console.WriteLine("Done");
```

Please can you provide some more information about the environment in which you're seeing this problem?

chrisdunelm:
@iwillspeak you say that you've added some code outside of `StopAsync` which prevents the client taking a long time to stop. Please can you share what this code looks like, as it may help point towards an explanation.

And what code are you executing within the subscription callback? If this is long-running, and doesn't properly honour cancellation, this can slow down stopping.

iwillspeak:
The subscribe callback can take a few seconds, and only checks the cancellation token at the start. That might be the source of the issue. A for the stop it looked something like:

```csharp
var stopTask = subscriber.StopAsync(stopCts.Token);
using (var registration = token.Register(() => stopCts.Cancel()))
{
    if (token.IsCancellationRequested)
    {
        stopCts.Cancel();
    }
    await stopTask;
}
```

Guessing that it was the processing block that was causing the slow stop then. The change to the cancellation must have  been a red herring.

chrisdunelm:
Thanks for closing. I think you're right that the callback is probably the problem.

PR #2560, opened 2018-10-04
Regenerate APIs

jskeet:


chrisdunelm:
The two pubsub files changed are not changes we want. Next week I'll fix the codegen so these aren't produced.
The asset changes are correct,

jskeet:
I've removed the PubSub commit.

PR #2562, opened 2018-10-04
Increasing the read timeout for error entries to tackle flakiness in Diagnostics Error Reporting integration tests.

amanda-tarafa:
Also removing a parameter from the ErrorEntryPolling helper that was not being used and was misleading by being there.


amanda-tarafa:
Thanks, waiting on green and pushing!

PR #2566, opened 2018-10-08
Logging changes for Spanner

jskeet:
(The bulk of the files here are due to not using ITestOutputHelper any more - it's not as scary as it looks.)

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/DefaultLogger.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/DefaultLogger.cs
@@ -24,13 +24,16 @@ namespace Google.Cloud.Spanner.V1.Internal.Logging
         /// <summary>
         /// This class is for internal use and not meant to be consumed directly.
         /// </summary>
-        protected override void WriteLine(string message)
-        {
+        protected override void WriteLine(LogLevel level, string message) => WriteLine($"{level}: {message}");
+
+        /// <inheritdoc />
+        public override void LogPerformanceMessage(string message) => WriteLine($"PERF: {message}");
+
+        private void WriteLine(string line) =>
 #if !NETSTANDARD1_5
-            System.Diagnostics.Trace.TraceInformation(message);
+            System.Diagnostics.Trace.TraceInformation(line);
 #else
-            Console.Error.WriteLine(message);
+            Console.Error.WriteLine(line);
 #endif
-        }
     }
 }
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/DefaultLogger.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/DefaultLogger.cs
@@ -19,11 +19,9 @@ namespace Google.Cloud.Spanner.V1.Internal.Logging
     /// <summary>
     /// This class is for internal use and not meant to be consumed directly.
     /// </summary>
-    public class DefaultLogger : Logger
+    internal class DefaultLogger : Logger
     {
-        /// <summary>
-        /// This class is for internal use and not meant to be consumed directly.
-        /// </summary>
+        /// <inheritdoc />
         protected override void WriteLine(LogLevel level, string message) => WriteLine($"{level}: {message}");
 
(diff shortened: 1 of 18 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/DefaultLogger.cs
```diff
@@ -24,13 +24,16 @@ public class DefaultLogger : Logger
         /// <summary>
         /// This class is for internal use and not meant to be consumed directly.
```
amanda-tarafa: nit: This method?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
```diff
@@ -26,32 +26,30 @@ namespace Google.Cloud.Spanner.V1.Internal.Logging
     /// </summary>
     public abstract class Logger
```
amanda-tarafa: This class is ours as well as `DefaultLogger`, and it doesn't implements any interfaces or other classes that are not "internal".
Why can't we have this class marked as `internal` and the methods as `protected internal` or even just `internal` and then make internals visible across projects.
It's weird with all these comments.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
```diff
@@ -26,32 +26,30 @@ namespace Google.Cloud.Spanner.V1.Internal.Logging
...
-            DefaultLogger = instance;
+            GaxPreconditions.CheckNotNull(instance, nameof(instance));
+            Interlocked.Exchange(ref s_defaultLogger, instance);
         }
 
         /// <summary>
         /// This is an internal property and is not intended to be used by external code.
         /// </summary>
-        public LogLevel LogLevel { get; set; } = LogLevel.Error;
+        public LogLevel LogLevel { get; set; } = LogLevel.None;
```
amanda-tarafa: Can't this change break client processes for clients who are logging errors without having to set this explicitly?
And, if this property is not supposed to be used by external code, who sets the logging level to something higher that `LogLevel.None`. I feel there's something I'm missing, but I don't know what.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
jskeet: Good call - we can make this one internal as nothing else creates a DefaultLogger.

We can't use InternalsVisibleTo between production projects because they can be versioned independently. (We did that before, and it was a mistake.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
jskeet: It's a breaking change effectively, but that's okay as we're at 2.0-betaXX. We'll include it in release notes. Fundamentally, I don't think we should ever be writing to the console by default.

I agree that the "is this meant to be used by external code or not" part is ambiguous at the moment - let's bear that in mind for future review.

amanda-tarafa: Ok

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/Logging/Logger.cs
amanda-tarafa: OK, I see how making InternalsVisibleTo between production projects can be a problem.

amanda-tarafa:
Travis is still failing.

jskeet:
Yup, will look into the Travis failure. (We'll get this quite a bit, I suspect - committing bits at a time is slightly hard to predict.)

amanda-tarafa:
Yep, no worries, just waiting on Travis passing to review all changes.

jskeet:
(Still failing, another attempt is in progress...)

PR #2568, opened 2018-10-08
Changes to get the EF provider to build

jskeet:
- Add the Common project to the solution (otherwise VS gets confused)
- Override methods in Logger
- Add TODO in Logger to separate the implementation from the API more cleanly

### DIFF
--- a/apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Diagnostics/SpannerLogBridge.cs
+++ b/apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Diagnostics/SpannerLogBridge.cs
@@ -20,7 +20,7 @@ namespace Microsoft.EntityFrameworkCore.Diagnostics
     /// <summary>
     /// This is internal functionality and not intended for public use.
     /// </summary>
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.EntityFrameworkCore.Spanner/Google.Cloud.EntityFrameworkCore.Spanner/Diagnostics/SpannerLogBridge.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.EntityFrameworkCore.Diagnostics
     /// <summary>
     /// This is internal functionality and not intended for public use.
     /// </summary>
-    public class SpannerLogBridge<TLoggerCategory> : DefaultLogger
+    public class SpannerLogBridge<TLoggerCategory> : Logger
```
amanda-tarafa: I think this one can be `internal` as well, it's only being used on the EF provider project.

jskeet: Possibly, but it's harder to tell - it may be that users are expected to instantiate it themselves. (It's not in an "Internal" folder, for example.)

I'm not overly worried about the EF provider at the moment - it's very alpha, and may want a larger overhaul before going any further anyway :)

PR #2570, opened 2018-10-09
Fix stress tests which fail with debug logging enabled

jskeet:
(The logging mechanism in tests just writes to a file - debug
logging currently means we write 140K+ lines to the file, which
affects the latency we're testing.)


jskeet:
(The test failed before this change, and passes with it.)

PR #2573, opened 2018-10-09
Coverage tool changes

jskeet:
This PR isn't as scary as it looks like - only the first commit needs reviewing.

I *believe* these are all the files that need changing...

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -46,7 +46,6 @@ while (( "$#" )); do
   elif [[ "$1" == "--coverage" ]]
   then
     runcoverage=true
-    install_dotcover
     mkdir -p coverage
   else
     apis+=($1)
--- a/build.sh
+++ b/build.sh
@@ -134,9 +134,9 @@ then
     if [[ "$runcoverage" = true && -f "$testdir/coverage.xml" ]]
     then
       echo "(Running with coverage)"
-      (cd "$testdir"; $DOTCOVER cover "coverage.xml" /ReturnTargetExitCode)    
+      (cd "$testdir"; dotnet dotcover test -c Release --no-build --dcXML=coverage.xml)    
     else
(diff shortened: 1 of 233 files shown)


build.sh
```diff
@@ -134,9 +134,9 @@ then
     if [[ "$runcoverage" = true && -f "$testdir/coverage.xml" ]]
     then
       echo "(Running with coverage)"
-      (cd "$testdir"; $DOTCOVER cover "coverage.xml" /ReturnTargetExitCode)    
+      (cd "$testdir"; dotnet dotcover test -c Release --no-build --dcXML=coverage.xml)    
     else
-      dotnet test -c Release $testproject
+      dotnet test -c Release --no-build $testproject
```
jskeet: I changed this at the same time before I *believe* it will make the build faster. (We've just built everything, so we don't need to check for it.)

runintegrationtests.sh
```diff
@@ -102,13 +102,9 @@ do
     dotnet run -p $testdir -- $TEST_PROJECT || echo "$testdir" >> $FAILURE_TEMP_FILE
   elif [[ "$COVERAGE_ARG" == "yes" && -f "$testdir/coverage.xml" ]]
   then
-    (cd $testdir; $DOTCOVER cover "coverage.xml" /ReturnTargetExitCode || echo "$testdir" >> $FAILURE_TEMP_FILE)
+    (cd $testdir; dotnet dotcover test -c Release --no-build --dcXML=coverage.xml || echo "$testdir" >> $FAILURE_TEMP_FILE)
```
amanda-tarafa: This comment applies to line 23 on this script, install_dotcover is not need anymore I think, and might generate conflicts because of different versions?

And, maybe we can now remove the instal_dotcover function from toolsversion.sh

PR #2577, opened 2018-10-10
Forcing choco install to confirm all prompts.

amanda-tarafa:
It might have been working in AppVeyor because apart from setting the `-y` option when executing `choco install`, there's a [global feature setting](https://github.com/chocolatey/choco/issues/178#issuecomment-83757651) that has the same effect for all `choco install` executions and that is probably set in AppVeyor.


amanda-tarafa:
Thanks, not waiting on AppVeyor green since I'll be running the tests anyway.

PR #2582, opened 2018-10-11
Use readlink -f instead of realpath in toolversions.sh

### LINKED ISSUE #2572: Travis error on realpath
We're seeing this in Travis:

> toolversions.sh: line 6: realpath: command not found

I'm not sure why, but we should look into it.

jskeet:
We use realpath in a couple of other places, but those are expected
to be run on developer machines, which is okay. realpath appears not
to be available on our Travis environment.

Fixes #2572.

### DIFF
--- a/toolversions.sh
+++ b/toolversions.sh
@@ -3,7 +3,7 @@
 
 # TODO: Make all of this work under Linux too, if it's useful
 
-declare -r REPO_ROOT=$(realpath $(dirname ${BASH_SOURCE}))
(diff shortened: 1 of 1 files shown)


jskeet:
Checking that it works: the travis build shouldn't show the error, basically. Previous logs included this:

```text
$ ./build.sh --diff --regex $API_REGEX
toolversions.sh: line 6: realpath: command not found
10:40:50 Building analyzers
```

PR #2584, opened 2018-10-11
Adding context info on tests that might be failing because of read timeouts to confirm if that's the case.

amanda-tarafa:
This will help confirm if and when the entries that we are not reading in a timely fashion are actually written on Stackdriver Error Reporting.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.IntegrationTests/ErrorReporting/ErrorReportingTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.IntegrationTests/ErrorReporting/ErrorReportingTest.cs
@@ -64,8 +64,7 @@ namespace Google.Cloud.Diagnostics.AspNet.IntegrationTests
 
             Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
 
-            var errorEvent = s_polling.GetEvents(_testId, 1).Single();
-
+            var errorEvent = ErrorEventEntryVerifiers.VerifySingle(s_polling, _testId);
             ErrorEventEntryVerifiers.VerifyFullErrorEventLogged(errorEvent, _testId, nameof(ErrorReportingController.ThrowsException));
 
             var content = await contentTask;
@@ -90,8 +89,7 @@ namespace Google.Cloud.Diagnostics.AspNet.IntegrationTests
             response = await requestTask4;
             Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
 
-            var errorEvents = s_polling.GetEvents(_testId, 4);
-            Assert.Equal(4, errorEvents.Count());
+            var errorEvents = ErrorEventEntryVerifiers.VerifyMany(s_polling, _testId, 4);
 
             var exceptionEvents = errorEvents
                 .Where(e => e.Message.Contains(nameof(ErrorReportingController.ThrowsException)));
@@ -114,8 +112,7 @@ namespace Google.Cloud.Diagnostics.AspNet.IntegrationTests
 
             Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
 
-            var errorEvent = s_polling.GetEvents(_testId, 1).Single();
-
+            var errorEvent = ErrorEventEntryVerifiers.VerifySingle(s_polling, _testId);
             ErrorEventEntryVerifiers.VerifyFullErrorEventLogged(errorEvent, _testId, "SendAsync");
         }
 
@@ -126,7 +123,7 @@ namespace Google.Cloud.Diagnostics.AspNet.IntegrationTests
 
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
 
-            var errorEvent = s_polling.GetEvents(_testId, 1).Single();
+            var errorEvent = ErrorEventEntryVerifiers.VerifySingle(s_polling, _testId);
             ErrorEventEntryVerifiers.VerifyErrorEventLogged(errorEvent, _testId, nameof(ErrorReportingController.ThrowCatchWithGoogleLogger));
         }
 
@@ -137,7 +134,7 @@ namespace Google.Cloud.Diagnostics.AspNet.IntegrationTests
 
             Assert.Equal(HttpStatusCode.OK, response.StatusCode);
 
-            var errorEvent = s_polling.GetEvents(_testId, 1).Single();
+            var errorEvent = ErrorEventEntryVerifiers.VerifySingle(s_polling, _testId);
             ErrorEventEntryVerifiers.VerifyFullErrorEventLogged(errorEvent, _testId, nameof(ErrorReportingController.ThrowCatchWithGoogleWebApiLogger));
         }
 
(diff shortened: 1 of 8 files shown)


amanda-tarafa:
Done cleaning up. @jskeet PTAL.

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.IntegrationTests/ErrorReporting/ErrorReportingTest.cs
```diff
@@ -64,8 +64,7 @@ public async Task LogsException()
 
             Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
 
-            var errorEvent = s_polling.GetEvents(_testId, 1).Single();
-
+            var errorEvent = ErrorEventEntryVerifiers.VerifySingle(s_polling.GetEvents(_testId, 1), _testId);
```
jskeet: Could this be simplified to `ErrorEventEntryVerifiers.VerifySingle(s_polling, _testId)`? We always call `GetEvents(testId, 1)` for `VerifySingle`, and we always pass in the same testId to VerifySingle as we do to GetEvents.

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.IntegrationTests/ErrorReporting/ErrorReportingTest.cs
```diff
@@ -90,8 +89,7 @@ public async Task LogsMultipleExceptions()
             response = await requestTask4;
             Assert.Equal(HttpStatusCode.InternalServerError, response.StatusCode);
 
-            var errorEvents = s_polling.GetEvents(_testId, 4);
-            Assert.Equal(4, errorEvents.Count());
+            var errorEvents = ErrorEventEntryVerifiers.VerifyMany(s_polling.GetEvents(_testId, 4), 4, _testId);
```
jskeet: Likewise this could be `ErrorEventEntryVerifiers.VerifyMany(s_polling, _testId, 4)`

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/ErrorReporting/ErrorEventEntryVerifiers.cs
```diff
@@ -14,13 +14,36 @@
...
+            var errorEvent = events.SingleOrDefault();
+            Assert.True(errorEvent != null, $"No entries found for context {contextInfo}");
+            return errorEvent;
+        }
+
+        /// <summary>
+        /// Checks that there are <paramref name="expected"/> events present.
+        /// Includes a message with context info to help track the possible issue.
+        /// </summary>
+        public static IEnumerable<ErrorEvent> VerifyMany(IEnumerable<ErrorEvent> events, int expected, string contextInfo)
```
jskeet: It might be worth converting the `IEnumerable<ErrorEvent>` to ` List<ErrorEvent>` using `ToList()`, so that it's only evaluated once.

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.IntegrationTests/ErrorReporting/ErrorReportingTest.cs
amanda-tarafa: I left it as is in case we could reuse VerifySingle and VerifyMany for collections obtained in different ways, but it's true that we don't need that right now, we can always make it more general if we need to. Will submit a commit shortly.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/ErrorReporting/ErrorEventEntryVerifiers.cs
amanda-tarafa: OK

amanda-tarafa:
@jskeet PTAL
Addressed review comments.
And I also increased the read timeout of or ErrorEntries to 10 minutes, which is not that alarming because the polling occurs every 30 seconds and returns as soon as it finds what it is looking for. I've been running it on local and the entries do always get written, I can see them on the UI in less that 5 minutes after they are written. When being read on the tests, right now they are being read after 4 minutes aprox (what each test takes to complete). I just think that this is not a problem on our part but just approximately the time it takes for entries to be "readable" from the API.

amanda-tarafa:
No rush!

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/ErrorReporting/ErrorEventEntryVerifiers.cs
```diff
@@ -14,13 +14,38 @@
...
+            var errorEvent = events.SingleOrDefault();
+            Assert.True(errorEvent != null, $"No entries found for context {contextInfo}");
+            return errorEvent;
+        }
+
+        /// <summary>
+        /// Checks that there are <paramref name="expected"/> events present.
+        /// Includes a message with context info to help track the possible issue.
+        /// </summary>
+        public static IEnumerable<ErrorEvent> VerifyMany(ErrorEventEntryPolling polling, int expected, string contextInfo)
```
jskeet: Definitely not blocking a merge, but if you feel like reordering the parameters to put `expected` at the end, that would feel slightly better in terms of making `VerifySingle` equivalent to `VerifyMany(polling, contextInfo, 1)` except the return type. The thing that varies is just the number, so it's nice to put that last.

If this were production code I'd suggest actually doing it - as it is, it's mostly a drive-by stylistic comment :)

amanda-tarafa: Yes will change.

amanda-tarafa:
Each test now might actually take up to 10 minutes, that's not usual though since I've since the entries being readable before hitting the 5 minutes mark, and it was working before with the timeout set at 2 minutes and then at 3 minutes. I'm not that worried about tests sometimes reaching the 10 minute mark and then even failing because the entries at not there, I think that this might happen very sporadically if at all.
What worries me if that we have tests (one for AspNetCore and one for AspNet) that check that nothing is logged when no Exception is thrown, and those need to wait for the whole timeout to actually say "nothing has been logged". So those are 30 minutes for AspNetCore, whose test run for three different platforms and 10 minutes for AspNet, for a total of 40 minutes just for that.
We can leave the timeout at 10 minutes, and we won't probably have flakiness (there's nowhere in the docs where I have found a time at which the entry is certain to be available) but we'll have those 40 minutes in our back, or we can reduce the timeout to say 5 minutes and live with the flakiness that might happen?
Unfortunately we cannot do something similar to what you described in Storage because we have this [read limitations](https://cloud.google.com/error-reporting/quotas#api_activity_limits). We have sometimes reached this and then we have to tweak the sleep between reads setting which stands now at 30 seconds.
@jskeet What do you think?

jskeet:
I'm fine with this PR as it is, but to clarify my concern: 10 minutes per *platform* is fine, but there are several tests here, and if we have 10 minutes per *test*, that could add up to a lot of elapsed time.

In general, I think it would be a good idea to make it easier to keep track of how long each test suite is taking somehow. If we could see a single block of text in the log with *all* the times, that would be great. But it'll take a bit of work to do - I'll have a think about it.

amanda-tarafa:
Changed the parameter order, waiting on green to merge.

PR #2603, opened 2018-10-15
Regenerate all APIs

jskeet:


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherServiceApiClient.cs
```diff
@@ -1909,6 +1940,71 @@ public virtual Publisher.PublisherClient GrpcClient
             throw new sys::NotImplementedException();
         }
 
+        /// <summary>
+        /// Lists the names of the snapshots on this topic.&lt;br&gt;&lt;br&gt;
+        /// &lt;b&gt;ALPHA:&lt;/b&gt; This feature is part of an alpha release. This API might be
```
chrisdunelm: This is in alpha server-side.
So I assume it shouldn't be in a GA client library?
Why is this not in a pubsub-alpha service?

jskeet: Yup, I find that alarming too. For now I'll remove the whole pubsub commit, and we can investigate separately.

jskeet:
Hmm. GetSnapshot has been present in the gRPC API for a long time. We haven't generated GAPIC methods for it before, because there's no GAPIC config. But presumably now we generate methods *by default*.

jskeet:
Chris, the failures here look like they're related to your recent changes - failure when using NoRetryRetryFilter etc.

However, that may be related to a bigger change: unconfigured methods are now generated by default. I think this is a dangerous change to apply unconditionally, so I'm not going to merge this. We should discuss it further.

jskeet:
Closing this PR; we don't want it due to the unexpected methods.

PR #2604, opened 2018-10-15
Release 1.0.0-beta09 of DLP V2 client libraries

jskeet:
Only to be merged after #2603.


jskeet:
We definitely don't want to do this yet. Too many other things going on.

jskeet:
Okay, we now have an appropriate codebase for a release :)

PR #2605, opened 2018-10-16
Regenerate APIs

jskeet:
This PR was generated by reverting an ordering commit in GAPIC. Once merged, I'll follow up with a PR that applies just the reordering.


jskeet:
Thanks - will merge on green.

PR #2606, opened 2018-10-16
Fix Travis

jskeet:
My previous attempt to make the file exist resulted in a Travis
parse failure.


jskeet:
The fact that the Travis build has started means it got further than before :)

PR #2607, opened 2018-10-16
Regenerate all APIs with GAPIC HEAD

jskeet:
This change is *just* reordering elements within ResourceNames. This should be a one-time change (at least until we change generator); resource name ordering will now be stable.


jskeet:
I haven't checked every file, but the ones I *have* checked have been just ordering.

PR #2611, opened 2018-10-17
Update MVC dependencies

jskeet:
Addresses https://nvd.nist.gov/vuln/detail/CVE-2017-0248

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -224,8 +224,8 @@
       "Google.Api.Gax.Testing": "2.5.0"
     },
     "additionalAnalyzerTestDependencies": {
-      "Microsoft.AspNetCore.Mvc.Core": "1.0.2",
-      "Microsoft.AspNetCore.Routing": "1.0.2"
+      "Microsoft.AspNetCore.Mvc.Core": "1.0.4",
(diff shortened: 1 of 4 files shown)


apis/apis.json
```diff
@@ -238,8 +238,8 @@
     "description": "Guidelines for using Google Stackdriver Instrumentation Libraries for ASP.NET Core.",
     "tags": [ "CodeAnalysis", "Analyzers", "CSharp" ],
     "testDependencies": {
-      "Microsoft.AspNetCore.Mvc.Core": "1.0.2",
-      "Microsoft.AspNetCore.Routing": "1.0.2"
+      "Microsoft.AspNetCore.Mvc.Core": "1.0.4",
+      "Microsoft.AspNetCore.Routing": "1.0.4"
     }
   },
```
chrisdunelm: Also update `Microsoft.AspNetCore.Http` in `Google.Cloud.Diagnostics.Common` below?

jskeet: I don't think we need to - certainly GitHub hasn't told me that there's a vulnerability there. (That's the only way I knew about this at all.)

I've only updated Routing in this case as it felt sensible to keep the two direct dependencies at the same version.

PR #2615, opened 2018-10-17
Deactivating coverage in AppVeyor so it won't clash with coverage from Kokoro.

amanda-tarafa:
Since we now have flagged coverage from Kokoro, both for unittests and integrationtests, if we run coverage from AppVeyor, which only includes unittests, we'll be bouncing back and forth from unittests only coverage (around 50%) and unittests+integrationtests coverage (around 70%).

### DIFF
--- a/appveyor.sh
+++ b/appveyor.sh
@@ -11,12 +11,11 @@ bash generateprojects.sh && git diff --exit-code
 if [[ -z "$APPVEYOR_PULL_REQUEST_NUMBER" ]]
 then
   # Not in a pull request.
-  # Run a full build (no --diff) with tests and coverage, then create and upload coverage report.
-  ./build.sh --coverage
-  ./createcoveragereport.sh --upload --upload_reportname unittests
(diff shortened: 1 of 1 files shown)


appveyor.sh
```diff
@@ -11,9 +11,9 @@ bash generateprojects.sh && git diff --exit-code
...
   # Not in a pull request.
-  # Run a full build (no --diff) with tests and coverage, then create and upload coverage report.
-  ./build.sh --coverage
-  ./createcoveragereport.sh --upload --upload_reportname unittests
+  # Run a full build (no --diff) with tests. Don't do any coverage so it
+  # won't clash with coverage generated through Kokoro.
+  ./build.sh
 else
   # We're in a pull request. Don't do any coverage; just
   # build and run tests.
```
jskeet: It's probably worth amending this comment as well

amanda-tarafa: Yep, will create a commit for that. And merge on AppVeyor green.

PR #2616, opened 2018-10-17
Statistics for session pools

jskeet:
This should all be trivial.
The SessionPoolTest.cs is just an empty partial class for the moment, to allow grouping. It wil be populated eventually :)


amanda-tarafa:
As you said this one was trivial :), specially compared with the others!

PR #2626, opened 2018-10-22
Fix handling of active session count for Spanner session pool

jskeet:


jskeet:
Accidental duplicate.

PR #2632, opened 2018-10-24
Revisit active session counts yet again

jskeet:
This separates out "sessions the pool has asked to create" from "sessions the user has asked for": the latter can go higher than MaxActiveSessionCount (depending on configuration), but the former shouldn't.

This feels simpler than the previous attempt, which didn't actually work properly anyway.
There's *still* a potential race condition as the number of new requests is computed before we start making them (at which point we increment the counter) - but it's just as unimportant as the previous race condition, in terms of being very unlikely *and* having minimal impact.

(Note: this is separate to the FakeScheduler fixes - they'll come in another PR.)

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
@@ -17,6 +17,7 @@ using Google.Cloud.Spanner.Common.V1;
 using Grpc.Core;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PoolRewrite/SessionPool.TargetedSessionPool.cs
```diff
@@ -29,6 +29,9 @@ namespace Google.Cloud.Spanner.V1.PoolRewrite
     public partial class SessionPool
     {
         // Note: Internal for test purposes.
+        // TODO:
+        // - Revisit naming for KnownSessionCount/ActiveSessionCount
```
amanda-tarafa: nit: Yes, a couple of things now.

- ActiveSessionCount is purely informative now, which is OK. I think the name is good.
- I would change KnownSessionCount for PooledSessionCount or ManagedSessionCount (even if they are active they are in the Pool, or managed by the Pool, at least that's how I better understand it)
- I would really change Options.MaximumActiveSessions to Options.MaximumPooledSessions or Options.MaximumManagedSessions.
- And I would really change Options.MinimumPooledSessions for Options.MinimumIdleSessions.

I've realized with this PR that those two Options' names have been bugging me since the start.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
```diff
@@ -213,6 +214,38 @@ public async Task AcquireAsync_ZeroPoolSizeAcquireUpToMaximum()
...
+            {
+                var pool = CreatePool(true);
+                var client = (SessionTestingSpannerClient)pool.Client;
+                await client.Scheduler.RunAsync(async () =>
+                {
+                    for (int i = 0; i < pool.Options.MaximumActiveSessions; i++)
+                    {
+                        await pool.AcquireSessionAsync(new TransactionOptions(), default);
+                    }
+                });
```
amanda-tarafa: nit: Maybe assert agains some statistics? And same with the test below.

jskeet: For the sake of git simplicity, I'll add those in a commit in PR #2633. Will do that now though while rebasing :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PoolRewrite/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Oh OK, I don´t know entirely what was there before and what´s new, definetely agree with minimum changes even if those options' names are not super straightforward.

And yes, totally ok with leaving this till the end.

PR #2635, opened 2018-10-25
[Not for submission, checking if tests are deflaked] Changes to FakeScheduller that might help with deflaking Spanner tests.

amanda-tarafa:
I think that if test are flaky after this change then we would need to evaluate the scenarios in wich no delay is schedulled in 100ms? Which I think are:

- because of the test itself doing "a lot" inside the `client.Scheduler.RunAsync` but not calling anything that actually schedulles a delay (I haven't seen a test like that).
- or because none of the threads get to the schedulling a delay point, which, by looking at the `TargettedSessionPool` code, seems improbable, even though all the (20 or 40) threads might be competing for the pool lock to get to the "fake spanner client" at least one of them should get through in 100ms I think.


amanda-tarafa:
Closing as this was never meant to be for submission.

PR #2636, opened 2018-10-25
[Bigtable] Smart retries logic for reads to account for request without RowSet

AVaksman:
Currently in the scenario of an exception during full table read (or with `RowsLimit` only) smart retry logic will throw an exception `Object reference not set to an instance of an object.` at this [point](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs#L99) since `originalRows` will be `null`.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
@@ -94,44 +94,41 @@ namespace Google.Cloud.Bigtable.V2
                 return originalRows;
             }
 
-            RowSet newRowSet;
-
-            if (originalRows != null)
+            // If the original request didn't specify a row set at all, create an open range from the last key
+            // to specify "all rows after this key".
+            if (originalRows == null)
             {
-                newRowSet = new RowSet
-                {
-                    RowKeys = {originalRows.RowKeys.Where(key => !StartKeyIsAlreadyRead(key))}
-                };
-
-                foreach (RowRange rowRange in originalRows.RowRanges)
-                {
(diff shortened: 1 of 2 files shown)


jskeet:
Thanks for this - I'm at a conference right now, but will review it as soon as I get the chance.

AVaksman:
Thank you

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableReadRowsRequestManager.cs
```diff
@@ -94,34 +94,44 @@ private RowSet FilterRows()
...
+                    {
+                        newRange = newRange.Clone();
+                        newRange.StartKeyOpen = LastFoundKey.Value;
+                    }
+
+                    newRowSet.RowRanges.Add(newRange);
                 }
-                newRowSet.RowRanges.Add(newRange);
+            }
+            else
```
jskeet: Could we rewrite this as:

```csharp
// If the original request didn't specify a row set at all, create an open range from the last key
// to specify "all rows after this key".
if (originalRows == null)
{
    return RowSet.FromRowRanges(RowRange.Open(LastFoundKey, null));
}

// Rest of code as before
```

That "early out" reduces the level of nesting for the rest of the method - and probably reduces this PR to only a tiny change.

AVaksman: Done. Thank you.

jskeet:
This change is now available in Google.Cloud.Bigtable.V2 version 1.0.0-beta05, which has been released.

AVaksman:
That's awesome news! Thank you

PR #2637, opened 2018-10-26
Connect up the SessionPool with TargetedSessionPool

jskeet:
Actual unit tests to come in another commit.

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PoolRewrite/SessionPool.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PoolRewrite/SessionPool.cs
@@ -48,6 +48,8 @@ namespace Google.Cloud.Spanner.V1.PoolRewrite
         /// </summary>
         public SessionPoolOptions Options { get; }
 
+        private readonly ConcurrentDictionary<DatabaseName, TargetedSessionPool> _targetedPools =
+            new ConcurrentDictionary<DatabaseName, TargetedSessionPool>();
 
         /// <summary>
         /// Creates a session pool for the given client.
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PoolRewrite/SessionPool.cs
```diff
@@ -64,6 +66,85 @@ public SessionPool(SpannerClient client, SessionPoolOptions options, Logger logg
             _logger = logger ?? Logger.DefaultLogger;
             _detachedSessionPool = new DetachedSessionPool(this);
             _sessionAcquisitionSemaphore = new SemaphoreSlim(Options.MaximumConcurrentSessionCreates);
+            Task.Run(() => PoolMaintenanceLoop(this));
+        }
+
+        /// <summary>
+        /// A long-running loop performing pool maintenance. Each iteration runs <see cref="MaintainPool"/>.
+        /// This is a static method to allow the target pool to be garbage collected, at which point the
```
amanda-tarafa: nit: Incomplete comment "at which point the"...

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PoolRewrite/SessionPool.cs
```diff
@@ -64,6 +66,85 @@ public SessionPool(SpannerClient client, SessionPoolOptions options, Logger logg
...
+        /// This is a static method to allow the target pool to be garbage collected, at which point the 
+        /// </summary>
+        /// <param name="pool">The pool to maintain.</param>
+        /// <returns></returns>
+        private static async Task PoolMaintenanceLoop(SessionPool pool)
+        {
+            // Keep a weak reference so that the pool can be garbage collected and we can stop the
+            // maintenance task.
+            var weakRef = new WeakReference(pool);
+            // Make sure that even if the pool variable is captured due to compiler implementation details,
```
amanda-tarafa: nit: Can't you just pass the week reference as the parameter to this method? Althouhg it's cleaner for calling code as it is now.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PoolRewrite/SessionPool.cs
jskeet: Whoops, thanks - will fix.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PoolRewrite/SessionPool.cs
jskeet: It would then have to be a static method within TargetedSessionPool, which feels a little odd. Basically I wanted TargetedSessionPool to feel as natural as possible, so the WeakReference part is isolated to just this method.

amanda-tarafa: Yep, actually I think I prefer leaving all the nulling and garbage collection related stuff isolated here.

jskeet:
(I'm looking into why we *still* have flaky tests here... I'm confused.)

amanda-tarafa:
I'm looking at it as well and think I found what it is, writing it down here because it's a change in the FakeScheduller, basically RunAndPause also needs to set `_stopped=true`. That change stopped the tests from timing out in VisualStudio, now they fail because the range is not good enough.
```
        public async Task RunAndPause(TimeSpan time)
        {
            var simulatedTimeTask = StartLoopAsync(Clock.GetCurrentDateTimeUtc() + time, false);
            var delayTask = Task.Delay(RealTimeTimeout);
            var completedTask = await Task.WhenAny(simulatedTimeTask, delayTask).ConfigureAwait(false);
            lock (_monitor)
            {
                _stopped = true;
                Monitor.PulseAll(_monitor);
            }
            if (completedTask == delayTask)
            {
                throw new SchedulerTimeoutException("Real time time-out; deadlock in user code?");
            }
            // Let any finally-released tasks a bit of time to execute before returning.
            await Task.Delay(PostLoopSettleTime).ConfigureAwait(false);
        }
```

amanda-tarafa:
About my last comment, although I think it's correct that RunAndPause set's `_stopped=true`, then:

- Both RunAndPause and RunAsync need to set `_stopped=false` at the very beginning.
- It doesn't fix the real timeout.

jskeet:
The intention is that RunAsync should only be called once per test (at the end), but RunAndPause can be called multiple times, so "stopped" is never set in that case. I need to make that clearer though :)

I suspect the simplest approach is to increase the delay for the maintenance loop in these tests. Will look into that. (Have been trying to fix the errors before this PR...)

jskeet:
Hooray for greenness :)

PR #2638, opened 2018-10-26
Add timeouts to TargetedSessionPool tests

jskeet:
Travis appeared to hang, which is likely to be in these tests.

(This is so that if we do still have flaky tests, we can tell which one is flaky.)


jskeet:
Merging despite the error as this change only makes failures better.

PR #2648, opened 2018-10-31
Changes to (and using) RetryHelpers

jskeet:

PR #2650, opened 2018-10-31
Read-only stress test with better diagnostics

jskeet:
(Builds on #2648)

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs
@@ -64,6 +64,8 @@ namespace Google.Cloud.Spanner.V1
             _request.SessionAsSessionName = _session.SessionName;
         }
 
+        // TODO: Take this in the constructor instead? Or maybe put it in SpannerClientSettings?
+
         /// <summary>
         /// This property is intended for internal use only.
         /// </summary>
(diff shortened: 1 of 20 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs
```diff
@@ -64,6 +64,8 @@ public sealed class ReliableStreamReader : IDisposable
             _request.SessionAsSessionName = _session.SessionName;
         }
 
+        // TODO: Take this in the constructor instead? Or maybe put it in SpannerClientSettings?
```
amanda-tarafa: Wouldn't the one in SpannerClientSettings be the one that _spannerClient is using? I would use the same as _spannerClient (if as I remember it has one). But OK with the TODO.

jskeet: Nope, there isn't one in SpannerClientSettings at the moment.

amanda-tarafa: Oh, I meant if `SpannerClientSettings` were to have one, then `SpannerClient` would use that one I suppose, so if here we use the one that `SpannerClient` currently has, we don't care if it came from settings or somewhere else, and it makes sense for everything to use the one in `SpannerClient` anyways, I think? But happy with deferring this.

jskeet: Yes, if we created a Logger property in SpannerClientSettings, we could use that in multiple places.

PR #2655, opened 2018-11-02
Guarantee a commit timestamp

jskeet:
This is a breaking API change, but one which will make customer code simpler.

Note: I've assigned this to both of you to help consider whether this breaking change is a good idea or not. It's more likely to impact customers than our other breaking changes, but if we're going to do it, now's the time *to* do it.


jskeet:
Not sure why it built for me locally but then failed in CI - will investigate!

jskeet:
Should now be fixed. I suspect I hadn't done an explicit build, relying on VS's live errors...

jskeet:
For some reason Travis hasn't picked up the modified commit. Not sure how to force it to do so...

amanda-tarafa:
It's passing now. Reviewing...

jskeet:
Righto, done. Will make sure I cover it in the "1.0 to 2.0 migration guide".

PR #2659, opened 2018-11-02
Redesign the Logger API

jskeet:
I'd suggest reviewing a commit at a time


amanda-tarafa:
Oh, and BTW, Travis failed because it couldn't connect to microsoft.packages, but AppVeyor passed so it's OK.

jskeet:
We should talk about internal vs not. If Microsoft.Extensions.Logging.Abstractions supported netstandard1.x we could use that, but it doesn't. (While we *could* consider dropping support for netstandard1.x for Spanner 2.0, that's a big step.)

jskeet:
I'll merge this PR when it's green (I've just force-pushed for the nit), and we can continue discussions around how we expose this.

Issue #2629, opened 2018-10-23
Use of TraceHeaderPropagatingHandler in .NET Core 2.1


gabrielmus:
I am using Google.Cloud.Diagnostics.AspNetCore version 3.0.0-beta06.
Due to the [change of paradigm for creating HttpClient](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-2.1), the [following code](http://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.Diagnostics.AspNetCore/index.html) for tracing outgoing http requests via Stackdriver Trace does not align well with .NET Core 2.1 any longer:

```csharp
/// <summary>
/// The <see cref="TraceHeaderPropagatingHandler"/> is populated by dependency injection.
/// </summary>
public async Task<HttpResponseMessage> TraceOutgoing(TraceHeaderPropagatingHandler traceHeaderHandler)
{
    // Add a handler to trace outgoing requests and to propagate the trace header.
    using (var httpClient = new HttpClient(traceHeaderHandler))
    {
        return await httpClient.GetAsync("https://weather.com/");
    }
}
```
Instead, the HttpClient instance must be created via an injected `IHttpClientFactory` (and "using" statement becomes unnecessary).  
More importantly, the registration and "chaining" of `HttpMessageHandler` into a handler pipeline is supposed to be configured at Startup.cs - ConfigureServices(), for example:
```csharp
services.AddHttpClient("weather").AddHttpMessageHandler<TraceHeaderPropagatingHandler>();
```
The HttpClient infrastructure takes care of chaining the `TraceHeaderPropagatingHandler` into the handler pipeline, but in return is asking that all handlers have a null `InnerHandler`, otherwise an exception is thrown.  
Unfortunately this is exactly what happens with `TraceHeaderPropagatingHandler`

```text
System.InvalidOperationException: The 'InnerHandler' property must be null. 'DelegatingHandler' instances provided to 'HttpMessageHandlerBuilder' must not be reused or cached.
Handler: 'Google.Cloud.Diagnostics.Common.TraceHeaderPropagatingHandler'
   at Microsoft.Extensions.Http.HttpMessageHandlerBuilder.CreateHandlerPipeline(HttpMessageHandler primaryHandler, IEnumerable`1 additionalHandlers)
   at Microsoft.Extensions.Http.DefaultHttpClientFactory.CreateHandlerEntry(String name)
[...]

PR #2661, opened 2018-11-05
[Diagnostics.Tracing] Supporting the System.Net.Http.IHttpClientFactory way of creating HttpClient instances for tracing outgoing requests.

amanda-tarafa:
On a closer look this way of [initiating HTTP requests](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests) is not only for Asp.Net Core 2.1, but is supported for .Net Standard 2.0 upwards.

Addresses #2629.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/TraceSnippets.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/TraceSnippets.cs
@@ -128,6 +128,28 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Snippets
             Assert.False(response.Headers.Contains(TraceHeaderContext.TraceHeader));
         }
 
+#if NETCOREAPP2_0
+        [Fact]
+        public async Task Traces_OutgoingClientFactory()
+        {
+            var uri = $"/TraceSamples/{nameof(TraceSamplesController.TraceOutgoingClientFactory)}/{_testId}";
+
+            var builder = new WebHostBuilder().UseStartup<TraceClientFactoryTestApplication>();
+            using(var server = new TestServer(builder))
+            using(var client = server.CreateClient())
+            {
+                var response = await client.GetAsync(uri);
+
+                var trace = s_polling.GetTrace(uri, _startTime);
+
+                TraceEntryVerifiers.AssertParentChildSpan(trace, uri, "https://weather.com/");
+                TraceEntryVerifiers.AssertSpanLabelsContains(
+                    trace.Spans.First(s => s.Name == uri), TraceEntryData.HttpGetSuccessLabels);
+                Assert.False(response.Headers.Contains(TraceHeaderContext.TraceHeader));
+            }
+        }
+#endif
+
         public void Dispose()
         {
             _client.Dispose();
@@ -174,6 +196,50 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Snippets
         }
         // End sample
     }
+
+#if NETCOREAPP2_0
+    internal class TraceClientFactoryTestApplication
+    {
+        private static readonly string ProjectId = TestEnvironment.GetTestProjectId();
+
+        // Sample: ConfigureHttpClient
+        public void ConfigureServices(IServiceCollection services)
+        {
+            // The line below is needed for trace ids to be added to logs.
+            services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
+
+            // Replace ProjectId with your Google Cloud Project ID.
+            services.AddGoogleTrace(options =>
+            {
+                options.ProjectId = ProjectId;
+            });
+
+            // Register an HttpClient for outgoing requests.
+            services.AddHttpClient("tracesOutgoing")
+            // Be sure to add the UnchainedTraceHeaderPropagatingHandler which will guarantee
+            // that the outgoing requests will be traced.
+            .AddHttpMessageHandler<UnchainedTraceHeaderPropagatingHandler>();
+
+            services.AddMvc();
+        }
+
+        public void Configure(IApplicationBuilder app)
+        {
+            // Use at the start of the request pipeline to ensure the entire request is traced.
+            app.UseGoogleTrace();
+
+            // ...any other configuration your application requires.
+
+            app.UseMvc(routes =>
+            {
+                routes.MapRoute(
+                    name: "default",
+                    template: "{controller=Home}/{action=Index}/{id?}");
+            });
+        }
+        // End sample
+    }
+#endif
 
     /// <summary>
     /// A controller for the <see cref="TraceTestApplication"/> used to trace operations.
@@ -232,6 +298,23 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Snippets
             }
         }
         // End sample
+
+#if NETCOREAPP2_0
+        // Sample: TraceOutgoingClientFactory
+        /// <summary>
+        /// Use the <see cref="IHttpClientFactory"/> to create an HttpClient that will guarantee
+        /// the tracing of outgoing requests.
+        /// The <see cref="IHttpClientFactory"/> is populated by dependency injection
+        /// thanks to the use of the <see cref="FromServicesAttribute"/> attribute.
+        /// </summary>
+        public async Task<HttpResponseMessage> TraceOutgoingClientFactory([FromServices] IHttpClientFactory clientFactory)
+        {
+            var httpClient = clientFactory.CreateClient("tracesOutgoing");
+            // Any code that makes outgoing requests.
+            return await httpClient.GetAsync("https://weather.com/");
+        }
+        // End sample
+#endif
     }
 
     // Sample: TraceMVCConstructor
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
@@ -121,8 +121,17 @@ namespace Google.Cloud.Diagnostics.AspNetCore
             // This is needed to get the `TraceHeaderContext`. See `CreateTraceHeaderContext`.
             services.TryAddSingleton<IHttpContextAccessor, HttpContextAccessor>();
 
-            services.AddSingleton(ManagedTracer.CreateDelegatingTracer(() => ContextTracerManager.GetCurrentTracer()));
-            services.AddSingleton(new TraceHeaderPropagatingHandler(() => ContextTracerManager.GetCurrentTracer()));
+            services.AddSingleton(ManagedTracer.CreateDelegatingTracer(ContextTracerManager.GetCurrentTracer));
+
+#if NETSTANDARD2_0
+            // On .Net Standard 2.0 or higher, we can use the System.Net.Http.IHttpClientFactory defined in Microsoft.Extensions.Http,
+            // for which we need a DelagatingHandler with no InnerHandler set. This is the recommended way.
+            // It should be registered as follows.
+            services.AddTransient(sp => new UnchainedTraceHeaderPropagatingHandler(ContextTracerManager.GetCurrentTracer));
+#endif
+            // This is to be used for explicitly creating an HttpClient instance. Valid for all platforms.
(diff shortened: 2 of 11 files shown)


amanda-tarafa:
@jskeet snippets and possible some more tests are coming in a commit, but if you can take a quick look and see if this looks alright that'd be great.

jskeet:
Will do so first thing in the morning - will be in meetings until then :(

amanda-tarafa:
No worries, no rush!

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
```diff
@@ -122,7 +122,16 @@ public static IApplicationBuilder UseGoogleTrace(this IApplicationBuilder app)
             services.TryAddSingleton<IHttpContextAccessor, HttpContextAccessor>();
 
             services.AddSingleton(ManagedTracer.CreateDelegatingTracer(() => ContextTracerManager.GetCurrentTracer()));
+
+#if NETSTANDARD2_0
+            // On .Net Standard 2.0 or higher, we can use the System.Net.Http.IHttpClientFactory defined in Microsoft.Extensions.Http,
+            // for which we need a DelagatingHandler with no InnerHandler set. This is the recommended way.
+            // It should be registered as follows.
+            services.AddTransient(sp => new UnchainedTraceHeaderPropagatingHandler(() => ContextTracerManager.GetCurrentTracer()));
```
jskeet: Does `new UnchainedTraceHeaderPropagatingHandler(ContextTracerManager.GetCurrentTracer)` work? (Just replacing the lambda expression with a method group conversion.)

Possibly likewise for TraceHeaderPropagatingHandler.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/UnchainedTraceHeaderPropagatingHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </summary>
+    /// <remarks>
+    /// Ensures the trace header is propagated in the headers for outgoing HTTP requests and 
+    /// traces the total time of the outgoing HTTP request. This is only done if tracing is initialized
+    /// and tracing is enabled for the request current request.
+    /// 
+    /// Explicitly leaves unset the InnerHandler property so that this Handler can be used with the
+    /// System.Net.Http.IHttpClientFactory defined in Microsoft.Extensions.Http.
+    /// </remarks>
+    public class UnchainedTraceHeaderPropagatingHandler : DelegatingHandler
```
jskeet: I'm not sure about "unchained" in the name here. Let's mull over some better names.

amanda-tarafa:
Yes, exactly, the client factory sets the InnerHandler, they want to have control over the lifetime of both the client and the message handlers.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/CloudTraceExtension.cs
amanda-tarafa: Oh yes, it should work, I just copy pasted from the existing one. Will change.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/UnchainedTraceHeaderPropagatingHandler.cs
amanda-tarafa: Yep, I don't like unchained very much either.

amanda-tarafa:
@jskeet This is ready for review.
Haven't changed the "Unchained...", maybe "NoInnerHandler...."?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/UnchainedTraceHeaderPropagatingHandler.cs
jskeet: Thinking about this further, do we actually need two classes? Could we just have two constructors, one with a handler and one without? Then just document the situation for each of them?

It feels a bit odd to have a subclass just for the sake of setting a property. Or is it needed for dependency injection?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/UnchainedTraceHeaderPropagatingHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Diagnostics.Common
+{
+    /// <summary>
+    /// Class for tracing outgoing HTTP requests and propagating the trace header.
+    /// </summary>
+    /// <remarks>
+    /// Ensures the trace header is propagated in the headers for outgoing HTTP requests and
```
jskeet: Nit: put the two paragraphs in `<para>` tags - although the second one may go away given the suggestion below.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/TraceSnippets.cs
```diff
@@ -175,6 +197,53 @@ public void Configure(IApplicationBuilder app)
...
+            services.AddSingleton<IHttpContextAccessor, HttpContextAccessor>();
+
+            // Replace ProjectId with your Google Cloud Project ID.
+            services.AddGoogleTrace(options =>
+            {
+                options.ProjectId = ProjectId;
+            });
+
+            // Register an HttpClient for outgoing requests.
+            services.AddHttpClient("weahter", c =>
```
jskeet: weahter => weather

But I wonder whether it would be worth making the sample not configure the client at all apart from the AddHttpMessageHandler call?

apis/Google.Cloud.Diagnostics.AspNetCore/docs/index.md
```diff
@@ -85,6 +85,21 @@ can in inject the `IManagedTracer` into the action method using the `[FromServic
...
 {{sample:Trace.UseTracerRunIn}}
 
-## Trace Outgoing HTTP Requests
+## Trace Outgoing HTTP Requests (all platforms)
 
 {{sample:Trace.TraceOutgoing}}
+
+## Trace Outgoing HTTP Requests (recommend for Asp.Net Core 2.0 upwards)
+
+The [recommend way of creating HTTPClient](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-2.1) in Asp.Net Core 2.0 and upwards is to use the
```
jskeet: recommend => recommended (here and in the heading)
HTTPClient => HttpClient
Asp.Net => ASP.NET

apis/Google.Cloud.Diagnostics.AspNetCore/docs/index.md
```diff
@@ -85,6 +85,21 @@ can in inject the `IManagedTracer` into the action method using the `[FromServic
...
-## Trace Outgoing HTTP Requests
+## Trace Outgoing HTTP Requests (all platforms)
 
 {{sample:Trace.TraceOutgoing}}
+
+## Trace Outgoing HTTP Requests (recommend for Asp.Net Core 2.0 upwards)
+
+The [recommend way of creating HTTPClient](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-2.1) in Asp.Net Core 2.0 and upwards is to use the 
+`System.Net.Http.IHttpClientFactory` defined in the Microsoft.Extensions.Http package.
+See the following example on how to register and use an HttpClient using Google Trace so that it traces
```
jskeet: "The following example demonstrates how to"?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/UnchainedTraceHeaderPropagatingHandler.cs
amanda-tarafa: The problem is that the `MessageHandler` parameter in the existing constructor is optional, a lot of our own code uses that, expecting the `InnerHandler` to be set to `new MessageHandler()` if the parameter is not, and potentially some user code as well. So we cannot add a new constructor that doesn't take a MessegaHandler, unless we want to go through a breaking change that will be breaking for us as well. And if we go with that we'd be forcing client code to always pass a `MessageHandler` even if they are happy with the "default" `new MessageHandler()`, which by the way, is always necessary except on this new client factory use case.
We can't even keep the existing constructor and add a static creation method `WithoutInnerHandler` because the existing constructor always sets InnerHandler, which cannot be set back to null later (I tried all of that before deciding on the two classes).

amanda-tarafa: Hmmm, yes, a private constructor with a dummy will work.
Name wise, for the static creation method, I like better `WithoutInnerHandler` or `NoInnerHandler` because it describes the instance instead of the usage, which is now only for HttpClientFactory, but might be for other things as well in the future? But my heart is not set on that so let me know what you prefer.

I'll try the private constructor approach in a separate commit, let's see how that looks.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/TraceSnippets.cs
amanda-tarafa: Yep, the snippet it's clearer if we only configure the `AddHttpMessageHandler` call, focus would be on our particular bit of configuration.
Changing.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/UnchainedTraceHeaderPropagatingHandler.cs
amanda-tarafa: Done, for now, I'm adding a commit just for the nits.

apis/Google.Cloud.Diagnostics.AspNetCore/docs/index.md
amanda-tarafa: Done

apis/Google.Cloud.Diagnostics.AspNetCore/docs/index.md
amanda-tarafa: Done

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/TraceSnippets.cs
amanda-tarafa: Done with a but:
`services.AddHttpClient()` returns `IServiceCollection`, all the other overloads for `AddHttpClient()` return an `IHttpClientBuilder` which is what we need to be able to add our custom message handler.
So I used the minimum extra configuration, which is just naming the HttpClient. 
```
services.AddHttpClient("tracesOutgoing")
.AddHttpMessageHandler<UnchainedTraceHeaderPropagatingHandler>();
```
This makes sense in a way, if we want to have HttpClients with extra configuration we at least have to name them.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/UnchainedTraceHeaderPropagatingHandler.cs
amanda-tarafa: > Presumably we could have two different calls and only register one... but that's probably overly brittle.

I don't get what you mean.

> am I right in saying that user code doesn't need to see it, because it just gets registered by our trace setup methods?

No, unfortunately users do get to see it, because they have to configure the HttpClient to use that particular message handler, we configure it for injection so they can use it when configuring the HttpClient.

amanda-tarafa:
Merging for now. Might still not be the definitive solution.

Issue #2534, opened 2018-09-25
Firestore: more flexible serialization


pauloevpr:
I am having some issues trying to get Firestore API to convert between `StringValue` and DDD value objects which are fully convertible from/to string. Let me give an example:

``` c#
[FirestoreData]
public class UserAccount
{
     [FirestoreProperty]
     public Email EmailAddress { get; set; }
}
public class Email 
{ 
     // Email has a TypeConverter implementation that allow full conversion from/to string
     // Email also has implicit operators to allow direct conversion from/to string
}
```
When calling `snapshot.ConvertTo<UserAccount>()`, `ValueDeserializer` fails with `System.ArgumentException: 'Unable to convert value type StringValue to Email'`.

It seems `ValueDeserializer` ignores type converters and implicit operators completely.

pauloevpr:
The issue is the same for `ValueSerializer`.

jskeet:
I *definitely* don't want to support `TypeConverter` - that would have to be for the `net45` target only (as it's not in netstandard1.5), which could cause some really subtle issues. Adding a dependency on an extra package could change which version of Google.Cloud.Firestore is used, suddenly breaking code.

While we *could* support the implicit conversion, it does introduce extra complexity which I'd really rather not - as per our previous discussions. (If there are multiple implicit conversions, which one is used? How does inheritance play into this? What about explicit conversions?) Now that private properties are supported, it's easy to work around this though with no library change:

```csharp
[FirestoreData]
public class UserAccount
{
     public Email EmailAddress { get; set; }

     [FirestoreProperty("email")]    
     public string FirestoreEmail { get => EmailAddress; set => EmailAddress = value };
}
```

pauloevpr:
Thank you for your response.

I wasn't aware of `net45` dependency of `TypeConverter`. I will agree with you on this. Keeping a low `netstandard` dependency is definitely something I support.

I will also agree that implicit conversation is not a good solution either.

Since this is clearly a feature request, I would like to propose a more generic solution to handle this and other scenarios where some sort of custom serialization is required:

Provide a generic interface like `IFirestoreValueSerializer`:

``` c#
public interface IFirestoreValueSerializer 
{
     Value SerializeValue(object value);
     object DeserializeValue(Value value);
}
```
Allow `FirestorePropertyAttribute` to receive a Type which is an implementation of `IFirestoreValueSerializer`. So consumers can set a custom serialization mechanism for a specific property:

``` c#
[...]

PR #2663, opened 2018-11-06
Rough and ready prototype of custom serialization

jskeet:
This is to demonstrate *one* possible approach to #2534.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/SerializationTestData.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/SerializationTestData.cs
@@ -131,6 +131,7 @@ namespace Google.Cloud.Firestore.Tests
             // Document references
             { Database.Document("a/b"),
                 new Value { ReferenceValue = "projects/proj/databases/db/documents/a/b" } },
+
         };
 
         public static TheoryData<IMessage, Func<Value, IMessage>> ProtoValues { get; } = new TheoryData<IMessage, Func<Value, IMessage>>
@@ -199,6 +200,42 @@ namespace Google.Cloud.Firestore.Tests
                 }
                 return Integers.All(pair => other.Integers.TryGetValue(pair.Key, out var otherValue) && pair.Value == otherValue);
             }
+        }
+
+        [FirestoreData]
+        public sealed class CustomUser
+        {
+            [FirestoreProperty]
+            public int HighScore { get; set; }
+
+            [FirestoreProperty]
+            public string Name { get; set; }
+
+            [FirestoreProperty]
+            public Email Email { get; set; }
+        }
+
+        [FirestoreData(Converter = typeof(EmailConverter))]
+        public sealed class Email
+        {
+            public string Address { get; }
+
+            public Email(string address) => Address = address;
+        }
+
+        public class EmailConverter : IFirestoreConverter<Email>
+        {
+            public Email FromFirestore(object value)
+            {
+                switch (value)
+                {
+                    case null: return null;
+                    case string address: return new Email(address);
+                    default: throw new ArgumentException($"Unexpected data: {value.GetType()}");
+                }
+            }
+
+            public object ToFirestore(Email value) => value?.Address;
         }
     }
 }
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ValueDeserializerTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ValueDeserializerTest.cs
@@ -241,7 +241,28 @@ namespace Google.Cloud.Firestore.Tests
             Assert.Equal("test", deserialized.Name);
             Assert.Equal(15, deserialized.Value);
         }
-        
+
+        [Fact]
+        public void DeserializeCustomConversion()
+        {
+            var value = new Value
+            {
+                MapValue = new MapValue
+                {
+                    Fields =
+                    {
+                        { "Name", new Value { StringValue = "test" } },
+                        { "HighScore", new Value { IntegerValue = 10L } },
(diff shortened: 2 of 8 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/SerializationTestData.cs
```diff
@@ -200,5 +201,41 @@ public bool Equals(DictionaryInterfaceContainer other)
...
+
+            [FirestoreProperty]
+            public string Name { get; set; }
+
+            [FirestoreProperty]
+            public Email Email { get; set; }
+        }
+
+        [FirestoreData(Converter = typeof(EmailConverter))]
+        public sealed class Email
```
mikelehen: What's the advantage of :
```
[FirestoreData(Converter = typeof(EmailConverter))]
public sealed class Email { ...
```
over:
```
public sealed class Email implements IFirestoreConverter<Email> { ...
```
?

The latter seems more direct and I'm not immediately thinking of what the downsides would be.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/SerializationTestData.cs
```diff
@@ -200,5 +201,41 @@ public bool Equals(DictionaryInterfaceContainer other)
...
+        public sealed class CustomUser
+        {
+            [FirestoreProperty]
+            public int HighScore { get; set; }
+
+            [FirestoreProperty]
+            public string Name { get; set; }
+
+            [FirestoreProperty]
+            public Email Email { get; set; }
```
mikelehen: What if instead of your own Email class (which you can add the Converter annotation to) you were using System.Net.Mail.MailAddress here?  Is there a way to serialize that to/from a String using this API approach?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/SerializationTestData.cs
jskeet: It's primarily a matter of separation of concerns. But more concretely:

- How would we create an instance of the converter if it's also the data model? It's very reasonable to have an immutable data model type with a parameterized constructor, as per the `Email` example. We never create an instance of `Email` ourselves, which we wouldn't be able to due to the parameterized constructor...
- This approach allows flexible serialization of types the developer doesn't own, e.g. `MailAddress` :)
- Even if the developer owns the type, they may want to have the serializer in a different project. It would be very reasonable to have a model project shared by both the server and client code, with no references to Firestore... which is necessary if the client and server refer to different Firestore libraries. If we can separate the converter from the model class, the client and server could each implement the serializer themselves (annoying redundancy, but at least it would be feasible). However, with this prototype they wouldn't be able to "attach" the converter to the model type, or even the `FirestoreData` annotation.

Possible approaches to solve the "shared model" and `MailAddress` issues:

- Permit a global (or per FirestoreDb?) registry of type converters.
- (Shared model only) Try to publish a package that both the client and server SDKs depend on, just with very core things like the attributes and interfaces. That makes me very nervous though.

mikelehen: FWIW- If we can actually solve the "flexible serialization of types the developer doesn't own" with this approach, then I'm much more on board. But without that it seems like boilerplate and indirection w/o benefit.

All models used with Firestore have to be mutable today, so I don't think it'd be problematic to require a public default constructor in order to implement IFirestoreConverter (which would probably be renamed IFirestoreConvertible or something).

I'm not overly won over by the "shared model" use case. It seems like an advanced use case and worst case they can just use the poor-mans-converter workaround of excluding Email from serialization and instead having a String getter/setter that does the Firestore conversion.

For the MailAddress case, I thought you had a previous iteration of this proposal that attached the converter to the property instead of the class, e.g.:

```
        [FirestoreData]
        public sealed class CustomUser
        {
           ...
            [FirestoreProperty(Converter = typeof(EmailConverter)]
            public Email Email { get; set; }
```
or something like that.  Would that solve this?

jskeet: > All models used with Firestore have to be mutable today, so I don't think it'd be problematic to require a public default constructor in order to implement IFirestoreConverter (which would probably be renamed IFirestoreConvertible or something).

From experience of consuming such a requirement, I'd really, really hate it. We certainly could have an `IFirestoreConvertible` interface as well, but if we're only going to support one I'd far rather keep them separate. This is introducing new functionality in general, so we should *expect* it to support use cases that we don't currently support - and being able to use immutable types would be a big plus IMO. (It naturally lends itself to various options for the "converting non-user types" as well, of course.)

Specifying the converter on the property would work, but you'd have to do it on *every* property of that type, which could be pretty annoying. (If it's only used in a few places, that's absolutely fine of course.) We could have that as an option *as well*, potentially - that may well be preferable to having a global registry.

One example which I think people may well want, and which we *could* even support out of the box is a type converter for `System.Guid`.

mikelehen: Specifying it on every property of that type seems fine to me.  That seems preferable to a global registry (e.g. since that would appear very magical to a reader of the code that doesn't know about the global registry or where it's being set up, plus all the other reasons globals are usually bad)

Maybe we can just allow both options (Specifying the converter on the class or on the usage of the class)?

mikelehen: Yup, I'm fine with the converter approach if it's actually buying us increased usefulness.

jskeet:
Closing this now as the real implementation is on the way.

PR #2665, opened 2018-11-06
Updates to Spanner documentation

jskeet:
- A guide to migrating to 2.0 with all the breaking changes I'm aware of
- Updates to the connection string documentation
- A guide to configuration, including details of session pool options

The way you use a custom SessionPoolManager is really awkward, but I
don't know what would be better right now. We should definitely
revisit it.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/docs/configuration.md
+++ b/apis/Google.Cloud.Spanner.Data/docs/configuration.md
@@ -1,0 +1,187 @@
+# Configuration
+
+Many applications will only ever need to configure Spanner in terms
+of the database to connect via the connection string. (The database
+name consists of a project ID, instance ID and database ID.)
+
+Beyond that, configuration is largely for the sake of diagnostics
+and performance tuning. There are two main sources for configuration:
+
+- The connection string provided to `SpannerConnection`. Options
+  here are also exposed as properties on
+  `SpannerConnectionStringBuilder`
+- The session pool options, as exposed via
+  `Google.Cloud.Spanner.V1.SessionPoolOptions`.
+
+The connection string options are [specified in this
+document](connection_string.md). The remaining options relate to the
+session pool, which handles acquiring, refreshing and evicting
+sessions. Spanner sessions are relatively expensive (in terms of
+time) to create, so it's important that they are handled
+appropriately.
+
+## Session pool options explained
+
+`SessionPoolOptions` exposes the following properties:
+
+**MaximumActiveSessions**
+
+Default value: 100
+
+The maximum number of sessions that can be active per database. An
+active session is one that has been acquired but not yet released
+back to the pool. Broadly speaking, this is equivalent to the number
+of concurrent Spanner-based operations your application will be able
+to perform.
+
+Applications may wish to change this setting if they require high
+levels of concurrency.
+
+**MinimumPooledSessions**
+
+Default value: 10
+
+The minimum number of sessions to maintain in the pool of available
+sessions, on a per-database basis. If the number of pooled sessions
+falls below this number, more sessions are added automatically,
+unless that would require more than the configured maximum active
+session count. (Setting the minimum pool size to a larger value than
+the maximum active session count has no purpose; the size will still
+be bounded by the maximum active session count.)
+
+Applications may wish to change this setting if they are likely to
+encounter sudden bursts of requests, and don't wish to have to wait
+for sessions to be created.
+
+**WriteSessionsFraction**
+
+Default value: 0.2
+
+The session pool maintains sessions in two states:
+
+- Without a transaction
+- With a read/write transaction
+
+The value of this setting determines how many sessions to keep in
+each state. For example, with the default value of 0.2, roughly 20%
+of sessions will be pooled with a read/write transaction.
+
+When a session is acquired, if there is not one already available
+with the requested transaction options, a transaction can be
+requested. Using a session which was previously associated with a
+read/write transaction for operations which don't use that
+transaction does not affect the behavior; it just wastes the small
+amount of time spent creating the transaction to start with.
+
+Increasing this value will increase the average time taken to release a
+session back to the pool, as more read/write transactions will be
+acquired as part of that operation. However, acquiring sessions with
+read/write transactions from the pool will then be faster.
+
+Applications that are aware of their usage patterns ahead of time
+may wish to change this setting to reduce latency. This is
+particularly true for applications which only perform read-only
+operations (in which case this value should be set to 0) or only
+perform operations using read/write transactions (in which case this
+value should be set to 1.0).
+
+**IdleSessionRefreshDelay**
+
+Default value: 15 minutes
+
+After this amount of time, a session which has been idle is
+refreshed automatically.
+
+Sessions that are unused for long periods of time can be invalidated
+by the server. The session pool periodically checks whether pooled
+sessions need refreshing. When a session needs refreshing, the pool
+just executes a simple query using that session. Sessions are
+additionally checked for refresh when they are released back to the
+pool. Jitter is applied to the session refresh to avoid large
+numbers of sessions all being refreshed at exactly the same time.
+
+Applications rarely need to change this setting.
+
+**PoolEvictionDelay**
+
+Default value: 7 days
+
+Processes that run for extended periods may refresh sessions many
+times, and the server will generally permit this indefinitely.
+However, that puts a strain on server-side resource management. This
+setting controls the time after which sessions are evicted. Like
+refresh settings, jitter is applied so that sessions are not evicted
+all at once.
+
+Applications rarely need to change this setting.
+
+**WaitOnResourcesExhausted**
+
+Default value: Block
+
+This enum value controls the behavior when an application requests a
+new session despite having already acquired the maximum number of
+active sessions.
+
+The default value (`Block`) will block the pending
+session acquisition operation until a session is available. This
+allows an application to continue to function correctly, but with
+reduced responsiveness.
+
+The value of `Fail` will cause the pending session acquisition to
+throw an exception. This is useful for spotting resource leaks, and
+some applications may prefer a swift failure over a delayed response
+when the application is heavily loaded.
+
+**Timeout**
+
+Default value: 60
+
+The total time in seconds allowed for a network call to the Cloud
+Spanner server, including retries. This setting is applied to calls
+to create, refresh and delete sessions, as well as beginning
+transactions.
+
+Applications rarely need to change this setting
+
+**MaximumConcurrentSessionCreates**
+
+Default value: 10
+
+Spanner has limits on the number of sessions that can be created
+concurrently without affecting performance. The session pool uses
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.Spanner.Data/docs/connection_string.md
```diff
@@ -64,4 +64,49 @@ v1.0.
...
+`SpannerTransaction.CommitTimeout`, and other Spanner network
+operations. Defaults to 60 seconds.
+
+Example:
+
+- Timeout=30
+
+# AllowImmediateTimeouts
+
+When set to False, a timeout of 0 means an indefinite timeout.
```
amanda-tarafa: nit: false, true without the capital letter?

apis/Google.Cloud.Spanner.Data/docs/connection_string.md
```diff
@@ -64,4 +64,49 @@ v1.0.
...
+
+Example:
+
+- MaximumGrpcChannels=8
+
+# MaxConcurrentStreamsLowWatermark
+
+The maximum number of concurrent streams per gRPC channel before
+using a new channel. Defaults to 20. This setting rarely needs to be
+modified.
```
amanda-tarafa: nit: no example here, fine with it, just broke the pattern slightly :)

apis/Google.Cloud.Spanner.Data/docs/configuration.md
```diff
@@ -0,0 +1,180 @@
+# Configuration
+
+Many applications will only ever need to configure Spanner in terms
+of the database to connect via the connection string. (The database
+name consists of a project ID, instance ID and database ID.)
+
+Beyond that, configuration is largely for the sake of diagnostics
+and performance tuning. There are two main source for configuration:
```
amanda-tarafa: nit: source_s_

apis/Google.Cloud.Spanner.Data/docs/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+
+**MinimumPooledSessions**
+
+Default value: 10
+
+The minimum number of sessions to maintain in the pool of available
+sessions, on a per-database basis. If the number of pooled sessions
+falls below this number, more sessions are added automatically,
+unless that would require more than the configured maximum active
+session count.
```
amanda-tarafa: Maybe be more explicit in saying that in practice, MinimumPooledSessions is upper bounded by MaximumActiveSessions.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+with the requested transaction options, a transaction can be
+requested. Using a session which was previously associated with a
+read/write transaction for operations which don't use that
+transaction does not affect the behavior; it just wastes the small
+amount of time spent creating the transaction to start with.
+
+Increasing this value will increase the average time taken to release a
+session back to the pool, as more read/write transactions will be
+acquired as part of that operation. However, acquiring sessions with
+read/write transactions from the pool will then be fast.
```
amanda-tarafa: nit: will then be fast_er_.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+perform operations using read/write transactions (in which case this
+value should be set to 1.0).
+
+**IdleSessionRefreshDelay**
+
+Default value: 15 minutes
+
+Sessions that are unused for long periods of time can be invalidated
+by the server. The session pool periodically checks whether pooled
+sessions need refreshing, doing so by issuing a simple query.
```
amanda-tarafa: instead of "doing so by issuing..." maybe "and if so, issues a simple query to refresh them"  (or similar), my point is that, as it is, it reads as if the query is issue to check if the session needs refreshing and not to actually refresh it.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+particularly true for applications which only perform read-only
+operations (in which case this value should be set to 0) or only
+perform operations using read/write transactions (in which case this
+value should be set to 1.0).
+
+**IdleSessionRefreshDelay**
+
+Default value: 15 minutes
+
+Sessions that are unused for long periods of time can be invalidated
```
amanda-tarafa: Maybe before the whole explanation, a simple sentence like : "This is the time a session needs to be idle before it is considered for refresh."

apis/Google.Cloud.Spanner.Data/docs/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+of `SessionPoolOptions` that it uses whenever it creates a new
+`SessionPool`. It's easy to modify the options used by
+the default `SessionPoolManager`. For example, to change the minimum
+number of sessions in each session pool:
+
+```csharp
+SessionPoolManager.Default.SessionOptions.MinimumPooledSessions = 100;
+```
+
+Any changes to options should be made before the Spanner operation.
```
amanda-tarafa: nit: to _the_ options

apis/Google.Cloud.Spanner.Data/docs/configuration.md
```diff
@@ -0,0 +1,180 @@
...
+Changes to options after a `SessionPool` has become active *may*
+still take effect, but that varies by option and may change over time.
+
+## Using a non-default SessionPoolManager
+
+An alternative to modifying the options used by the default
+`SessionPoolManager` to create a new manager with a new set of
```
amanda-tarafa: nit: _is_ to create

apis/Google.Cloud.Spanner.Data/docs/migration-to-2.md
```diff
@@ -0,0 +1,103 @@
...
+  - `MaximumPooledSessions` has no direct equivalent in 2.0; once a session
+  is part of the pool, it will be refreshed as necessary until it is evicted.
+  - `MaximumActiveSessions` is now only present in `SessionPoolOptions`
+  - `MaximumGrpcChannels` is specified in the connection string
+  - `LogLevel` is controlled on a per-logger basis; this property on
+    `SpannerOptions` was equivalent to changing
+    `Logger.DefaultLogger.LogLevel`.
+  - `PoolEvictionDelay` is now only present in `SessionPoolOptions`
+  - `ResourcesExhaustedBehavior` has moved to `SessionPoolOptions`
+  - `Timeout` is specified in the connection string
```
amanda-tarafa: and in SessionPoolOptions as well, at least according to configuration.md (unless these are two different timeouts?)

apis/Google.Cloud.Spanner.Data/docs/migration-to-2.md
```diff
@@ -0,0 +1,103 @@
...
+  anyway. Client code should be able to use the resulting
+  `Task<DateTime>` more easily in most cases.
+- `SpannerConnection` and `SpannerConnectionStringBuilder` no longer
+  expose `GetCredentials` method. There is no direct equivalent in 2.0;
+  please file an issue if you rely on this functionality.
+- `SpannerConnection.ClearPooledResourcesAsync` is replaced by
+  `SpannerConnection.ShutdownSessionPoolAsync`. After this is called,
+  no further sessions can be created with the associated session pool.
+  If you need to control sessions in a more fine-grained way,
+  you can associate a `SessionPoolManager` with the
```
amanda-tarafa: Does this implies that they should inherit from SessionPoolManager? Because whatever the options on the session pool, shutting down is the same. Basically what do you mean with "control sessions in a more fine-grained way"?

apis/Google.Cloud.Spanner.Data/docs/configuration.md
jskeet: Done.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
jskeet: I've added a sentence - see what you think.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
jskeet: I did intend fast rather than faster - because basically it doesn't have to do anything when the session is retrieved - but I'm fine with faster. Done.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
jskeet: Done.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
jskeet: Good catch.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
jskeet: Well, it's more *any* options, but "the" is okay. Also fixed "the Spanner operation" which didn't make sense.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
jskeet: Done, and also added "object" to make it clearer I'm not suggesting that people write their own classes.

apis/Google.Cloud.Spanner.Data/docs/connection_string.md
jskeet: It would typically be written as False (or True) in the connection string, and I believe that's how it would be formatted if you call ToString on the connection string too.

apis/Google.Cloud.Spanner.Data/docs/connection_string.md
jskeet: Fixed, thanks. (Admittedly I'd be shocked if this were ever needed. It's really just for completeness.)

apis/Google.Cloud.Spanner.Data/docs/migration-to-2.md
jskeet: They *are* different timeouts, and were always separate. The one that was in SpannerOptions has only moved to the connection string.

apis/Google.Cloud.Spanner.Data/docs/migration-to-2.md
jskeet: I've tried to clarify this a bit. See what you think.

apis/Google.Cloud.Spanner.Data/docs/connection_string.md
amanda-tarafa: OK

apis/Google.Cloud.Spanner.Data/docs/migration-to-2.md
amanda-tarafa: OK

jskeet:
(Will merge on green.)

PR #2670, opened 2018-11-07
Remove a FIXME by simplifying the code

jskeet:
Open() makes sure we have a SessionPool, but that's all. It doesn't
need to allocate a session, therefore it doesn't matter whether or
not we have a database. We can validate quite a lot of things
without that. Admittedly if the project or instance is invalid, we
won't spot that at this point, but that's not so bad. It means we
have more consistency: an open connection always has a session pool.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
@@ -511,7 +511,7 @@ namespace Google.Cloud.Spanner.Data
         }
 
         /// <summary>
-        /// Opens the connection (which includes acquiring a SessionPool, unless there's no database)
+        /// Opens the connection, which involves acquiring a SessionPool,
         /// and potentially enlists the connection in the current transaction.
         /// </summary>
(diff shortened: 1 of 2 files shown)


jskeet:
(I'm running the integration tests to check that this doesn't break anything.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -523,14 +523,6 @@ private Task OpenAsyncImpl(Action transactionEnlister, CancellationToken cancell
             return ExecuteHelper.WithErrorTranslationAndProfiling(
                 async () =>
                 {
-                    // FIXME: Prevent opening instead? It's an annoyingly inconsistent state (with no _sessionPool).
```
amanda-tarafa: nit: On the method summary remove the text in bold:
(which includes acquiring a SessionPool**, unless there's no database**)

jskeet: Well spotted :) Fixed that; will wait for the integration tests to pass, then merge.

PR #2679, opened 2018-11-09
Provide a logger to SpannerSettings, and use it.

jskeet:
(API documentation messages still need rewriting.) The sole "entry
point" for configuring a logger when using ADO.NET is now the
SessionPoolManager.

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
@@ -56,14 +56,13 @@ namespace Google.Cloud.Spanner.V1
         /// </summary>
         /// <param name="client">The client to use for this session pool. Must not be null.</param>
         /// <param name="options">The options for this session pool. Must not be null.</param>
-        /// <param name="logger">The logger to use. May be null, in which case the default logger is used.</param>
-        public SessionPool(SpannerClient client, SessionPoolOptions options, Logger logger)
+        public SessionPool(SpannerClient client, SessionPoolOptions options)
         {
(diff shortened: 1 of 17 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
```diff
@@ -56,14 +56,13 @@ public sealed partial class SessionPool
...
-        /// <param name="logger">The logger to use. May be null, in which case the default logger is used.</param>
-        public SessionPool(SpannerClient client, SessionPoolOptions options, Logger logger)
+        public SessionPool(SpannerClient client, SessionPoolOptions options)
         {
             _client = GaxPreconditions.CheckNotNull(client, nameof(client));
             _clock = _client.Settings.Clock ?? SystemClock.Instance;
             _scheduler = _client.Settings.Scheduler ?? SystemScheduler.Instance;
             Options = GaxPreconditions.CheckNotNull(options, nameof(options));
-            _logger = logger ?? Logger.DefaultLogger;
+            _logger = client.Settings.Logger; // Just to avoid fetching it all the time
```
amanda-tarafa: But it can change in `Settings`, right? None of the other classes does this, so this might introduce incosistencies in logging if the logger is changed midflight (it shouldn't probably but... it can).

jskeet: Yes, in theory this could create an inconsistency. I think it's okay though as:

- We don't expose a SpannerClient or its settings anywhere that I can think of
- We're expecting the logging to be enabled in consultation with Google anyway, with suitable hand-holding

If this were a more general-purpose logging API I think that would be a more important concern.

PR #2690, opened 2018-11-12
Add delays in bucket labels tests

jskeet:
This is an attempt to counter the flakiness we've seen recently

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ModifyBucketLabelsTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ModifyBucketLabelsTest.cs
@@ -269,10 +269,14 @@ namespace Google.Cloud.Storage.V1.IntegrationTests
             Assert.Empty(result);
         }
 
-        // TODO: Move these somewhere common, if this proves a useful pattern.
         private static T RunMaybeAsync<T>(bool runAsync, Func<T> sync, Func<Task<T>> async)
+        {
             // Run the async func separately to explicitly avoid synchronization etc.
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ModifyBucketLabelsTest.cs
```diff
@@ -269,10 +269,14 @@ public void ClearBucketLabels_NoLabelsBefore(bool runAsync)
...
         }
 
-        // TODO: Move these somewhere common, if this proves a useful pattern.
         private static T RunMaybeAsync<T>(bool runAsync, Func<T> sync, Func<Task<T>> async)
+        {
             // Run the async func separately to explicitly avoid synchronization etc.
-            => runAsync ? Task.Run(() => async().Result).Result : sync();
+            var result = runAsync? Task.Run(() => async().Result).Result : sync();
+            // Sleep, as everything in this test modifies buckets, and we can't do that very frequently.
+            StorageFixture.SleepAfterBucketCreateDelete();
```
amanda-tarafa: I think this will work as long as none of the tests do several updates themselves. I've taken a quick look at the tests, and maybe the `_RetrySucceeds` or `_RetryFails` might still hit quota issues? Not sure though.

jskeet: It's possible they'll still have problems - it's unclear. I *suspect* the quota is checked on an aggregate basis and that we're actually sleeping for longer than we need to, so this is "good enough". We can tweak it if we need to, of course. Will merge for now.

Issue #2702, opened 2018-11-13
Storage labels tests are flaky


jskeet:
We run into quota issues when running integration tests in CI. (Not as often on a developer machine, oddly...)

Looking into it; will skip the tests for now.

jskeet:
This was fixed in #2717

PR #2703, opened 2018-11-13
Skip flaky tests until we have more information

jskeet:
See #2702


jskeet:
Merging as I'd discussed this with Amanda offline :)

PR #2707, opened 2018-11-14
Test marker for common integration tests

jskeet:
If this works and updates the internal tool correctly, I'll roll it
out for all other tests.


jskeet:
Will chat internally :)

PR #2709, opened 2018-11-15
Fix CopyObject by using Done instead of RewriteToken to detect completion

### LINKED ISSUE #2710: CopyObject fails with an Invalid Request error
CopyObject uses the "object rewrite" API call, which can require multiple requests before completing.

In 2.3.0-beta02 and earlier, we detected the completion of the call via the rewrite token being absent in the response.
Recently it looks like the service has started returning an empty rewrite token on completion, which meant that we'd make an extra request, which would fail.

The solution is to use the done property in the response instead.

jskeet:
It looks like the server recently changed from not returning a
rewrite token at all to returning an empty one. Using Done is a
better check, and prevents duplicate requests.

Fixes #2710. We should consider whether we want to backport this and release a 2.2.1 package as well.


jskeet:
Adding Chris as a reviewer too as I'd like to get this built and deployed ASAP.

PR #2711, opened 2018-11-15
Storage release 2.3.0-beta02

jskeet:
We need to skip some tests as the release will not be pushed unless all integration and snippet tests pass.
I'm confident these are not client issues, but just server behavior changes that need accounting for after investigation.


jskeet:
Yes, basically I'm taking today to investigate all the Storage issues we've been seeing very carefully.

PR #2712, opened 2018-11-15
Create Google.Cloud.Storage.V1 version 2.2.1

### LINKED ISSUE #2710: CopyObject fails with an Invalid Request error

jskeet:
Commits are:

- Skip integration tests we still expect to fail
- Fix #2710 (the purpose of this release)
- Update the version number


jskeet:
Travis failure is due to the use of --diff.

PR #2723, opened 2018-11-23
Remove unnecessary logging attributes

jskeet:
The fixture has the attribute, so we don't need it on any test that uses the fixture.


jskeet:
I've just rerun the tests, and we still get all the logging. Yay.
Will merge as this PR is so trivial :)

PR #2726, opened 2018-11-24
First pass at Firestore serialization benchmarks

jskeet:
This is mostly just to check that the new serialization code doesn't make things much worse. (It actually improves things significantly in almost all cases.)


jskeet:
(Currently updating to include ExpandoObject and simplify a little. Just running the benchmarks again; will push when they've run successfully.)

jskeet:
Right, repushed now.

PR #2727, opened 2018-11-26
Refactor serialization/deserialization code

jskeet:
This isn't quite a pure refactor, as ValueDeserializer is now a
static class. This does remove one potential (but previously unused)
feature: the ability to deserialize to ExpandoObject using nested
ExpandoObjects for further entities in the document. We can
introduce that later if we really want.

The benchmarks show an improvement to most cases, but not to
dictionary deserialization. That's worse than before, but not
prohibitively so. Once everything is settled, we can look at
optimization. (Basically we want to avoid reflection everywhere we
can.)

This PR is a precursor to custom serialization which will be in a
subsequent PR.


apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ValueDeserializerTest.cs
```diff
@@ -107,15 +104,15 @@ public void EnumOverflow(BclType targetType)
         public static IEnumerable<object[]> InvalidDeserializationCalls = new List<object[]>
         {
             // Inappropriate target types
-            { new Value { IntegerValue = 10L }, typeof(string) },
-            { new Value { ArrayValue = new ArrayValue() }, typeof(string) },
-            { new Value { MapValue = new MapValue() }, typeof(string) },
+//            { new Value { IntegerValue = 10L }, typeof(string) },
```
jskeet: Not sure why I commented these out - will fix now.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/ObjectConverter.cs
```diff
@@ -0,0 +1,67 @@
...
+    /// </summary>
+    internal sealed class ObjectConverter : IFirestoreInternalConverter
+    {
+        public object DeserializeMap(FirestoreDb db, IDictionary<string, Value> values) =>
+            values.ToDictionary(pair => pair.Key, pair => DeserializeValue(db, pair.Value));
+
+        public object DeserializeValue(FirestoreDb db, Value value)
+        {
+            switch (value.ValueTypeCase)
+            {
```
amanda-tarafa: Maybe use the switch only to "calculate" the target type and then do `ValueDeserializer.Deserialize(...)`?? It might be a little overhead to go through all of that, but the code in each of the cases is the same as the one in the corresponding converter, I think.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ValueSerializer.cs
```diff
@@ -128,46 +56,9 @@ internal static Value Serialize(object value)
         internal static Dictionary<string, Value> SerializeMap(object value)
```
amanda-tarafa: Same here, maybe `SerializeToMap`

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ValueDeserializerTest.cs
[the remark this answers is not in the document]
jskeet: This was accidental, and now fixed.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/AttributedTypeConverter.cs
```diff
@@ -81,36 +67,74 @@ private FirestoreDataAttributedType(BclType type)
                 {
                     // This is O(N), but done once per type so should be okay.
```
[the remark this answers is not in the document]
jskeet: It's O(N) per field, so O(N^2) overall, yes.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/IFirestoreInternalConverter.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Populates a map based on the serialized form of a POCO.
+        /// </summary>
+        /// <remarks>
+        /// Logically this could return a dictionary instead, but it's more efficient to populate an existing map.
+        /// </remarks>
+        /// <param name="value">The value to serialize. Never null.</param>
+        /// <param name="map">The map to populate. Never null.</param>
+        void SerializeMap(object value, IDictionary<string, Value> map);
```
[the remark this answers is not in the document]
jskeet: The intention is that it's serializing something that *is* a map. I'll leave it as-is for the moment, but we can revisit all names later. (I intend to keep all of this internal.)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/ObjectConverter.cs
jskeet: Mmm, good point. (We could use an array lookup instead of the switch if we wanted, too.) Will merge with this as-is so we can get a clear comparison, but then see what it does to the benchmarks. I don't mind the duplication if it makes a significant difference, but I'd prefer to get rid of it, yes.

PR #2728, opened 2018-11-26
Vision release (after minor regenerated update)

jskeet:


jskeet:
Fortunately we haven't done any releases with product search before - but regardless of whether we had, it's describing the service behaviour rather than the client library anyway. I agree it's somewhat unusual.

PR #2730, opened 2018-11-27
Remove old prerelease API Vision packages

jskeet:
The features from Google.Cloud.Vision.V1P{1,2}Beta1 are now in Google.Cloud.Vision.V1, so the source for those betas can be removed.

The NuGet packages have been delisted.

This should not affect any users who are still using the packages - they can still be restored from NuGet, they just aren't suggested any more.


jskeet:
The Travis failure (and AppVeyor when it gets that far) is due to the way we build PRs: we find out which projects have changed, and build those. In this case, the two "changed" projects are the deleted ones, which obviously can't be built...

PR #2733, opened 2018-11-27
(Do not merge yet) Generate Dataproc V1Beta2

jskeet:
I don't want to merge this yet as it's not entirely clear that we want to release this package, but I wanted to go through the first steps in case we *do* want it.


jskeet:
Closing for now as it looks like we're going to stick to just V1 for C#.

Issue #2435, opened 2018-08-20
Enable configuration support for Google.Cloud.Diagnostics.AspNetCore


tillig:
When using the `Google.Cloud.Diagnostics.AspNetCore.GoogleDiagnosticsWebHostBuilderExtensions.UseGoogleDiagnostics` extension, there is an assumption that the Google Cloud project ID is hardcoded. This is seen [in the examples](http://googlecloudplatform.github.io/google-cloud-dotnet/docs/Google.Cloud.Diagnostics.AspNetCore/index.html#initializing-google-diagnostics-aspnet-core-20):

```csharp
string projectId = "[Google Cloud Platform project ID]";
var webHost = new WebHostBuilder()
    .UseGoogleDiagnostics(projectId)
    .UseStartup<Startup>()
    .Build();
```

Looking at the extension, it appears this [roughly boils down to a call to `IWebHostBuilder.ConfigureServices(Action<IServiceCollection>)`](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/bdc642b3996a7d2836e645852e73500f351baa56/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/GoogleDiagnosticsWebHostBuilderExtensions.cs#L51) where several things get registered with DI.

There is another overload `IWebHostBuilder.ConfigureServices(Action<WebHostBuilderContext, IServiceCollection>)` where the web host can provide configuration for the app via `WebHostBuilderContext.Configuration`. This allows configuration to be set up then passed in to drive later registrations.

It would be helpful if there was a way to use this overload to avoid having to set up configuration multiple times (one for use in GCP calls, one for the actual app) and avoid hard coding.

For example, something like this:

```csharp
public static IWebHostBuilder UseGoogleDiagnostics(this IWebHostBuilder builder, string projectIdKey = null, string serviceNameKey = null, string serviceVersionKey = null)
{
  builder.ConfigureServices((context, services) =>
  {
    var projectId = projectIdKey == null ? null : context.Configuration[projectIdKey];
    var serviceName = serviceNameKey == null ? null : context.Configuration[serviceNameKey];
    var serviceVersion = serviceVersionKey == null ? null : context.Configuration[serviceVersionKey];
    // then register the stuff as usual.
  }

  return builder;
}
```

[...]

PR #2741, opened 2018-11-29
Extension for reading monitored resource data from host builder context.

tillig:
Resolves issue #2435.

Added an extension for `IWebHostBuilder` that allows the project ID, service name, and service version information to come from the `WebHostBuilderContext` at registration time rather than having to be passed in directly. This allows the data to be stored in configuration or in other locations.

The `WebHostBuilderContext` overload for the `ConfigureServices` method [was introduced in ASP.NET Core 2.0.0](https://github.com/aspnet/Hosting/commit/2b07e88a581ef2ad1cccd6a68170cf5aba77c6e7) so it's only available for the `netstandard2.0` version of the `Google.Cloud.Diagnostics.AspNetCore` library. The corresponding integration test is only run when targeting `netcoreapp2.0`.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsWebHostTests.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsWebHostTests.cs
@@ -22,8 +22,10 @@ using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net.Http;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


tillig:
Also, something to be aware of: The `NuGetFallbackFolder` that gets installed with the .NET Core SDK doesn't include a set of libraries that will pass NuGet signature validation. This folder gets used when things are using multitargeting and include a target for `netcoreapp1.0` or `netcoreapp1.1` like the unit tests in here. This [causes a lot of errors to be raised in newer .NET Core SDK versions](https://github.com/NuGet/Home/issues/7414) and uses of NuGet because they've started trying to validate those signatures during the build... and when they don't validate, package restore fails. I gather they recommend targeting `netcoreapp1.2` or higher to work around the issue.

I did not include that change here. Instead I just temporarily cleared out my `NuGetFallbackFolder` to force NuGet to download fresh/correct packages for everything.

amanda-tarafa:
Thanks @tillig, I'll be reviwing this today.

amanda-tarafa:
@tillig as with the problem with signature validation with some of the packages, I'm not seing this. I even checked out your PR and building and running tests is OK. And Diagnostics is not the only library that targets several netcoreapp versions and I haven't seen that before nor, as far as I know, anyone else from the team. I'm using VS 15.9.3 (as I saw on the issue you shared that you are using as well). For now I don't think we'll change our targetting. But thanks for raising, at leas if it starts happening we know where to look.

amanda-tarafa:
Mergin now!

PR #2742, opened 2018-11-30
Add a lot of tracking test names for Spanner integration tests

jskeet:
Also adds a test for writing then reading a null bytes field, which we'd missed before.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
@@ -70,6 +70,30 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
             }
         }
 
+        // [START spanner_test_write_read_bool]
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -297,6 +366,7 @@ public async Task WriteRandomBytes()
                 Assert.Equal(0, recordedValues.Count);
             }
         }
+        // [END spanner_test_write_read_random_bytes]
 
         [Fact]
```
amanda-tarafa: This one doesn't have START-END. It might have been on purpose though.

And there are some others in this file as well that doesn't have START-END (lines 241-261)

jskeet: Yup, this is only for tracking the tests that are required across multiple languages. We have some extra tests for ourselves.

PR #2745, opened 2018-11-30
Make URL signing methods accepting DateTimeOffset? obsolete

jskeet:
In the future, non-expiring signed URLs will not be supported. In a future PR, we'll probably turn "infinite expiry" into "expires in 2038" as a workaround for this.

Bonus for @frankyn: I was wrong about this *defaulting* to no expiry. At least users are having to explicitly pass null at the moment. Any code passing a non-null value (with a non-null compile-time type) will start using the non-obsolete overload automatically.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
@@ -37,6 +37,9 @@ namespace Google.Cloud.Storage.V1
     /// </remarks>
     public sealed partial class UrlSigner
     {
+        private const string InfiniteExpiryObsoleteMessage =
+            "This overload is obsolete, as non-expiring signed URLs are a bug in the service that can cause security issues and should no longer be used. Please use the overload accepting a DateTimeOffset instead.";
+
         private static readonly ISigner s_v2Signer = new V2Signer();
         private static readonly ISigner s_v4Signer = new V4Signer();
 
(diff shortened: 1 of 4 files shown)


jskeet:
Looking into the build breakage now

jskeet:
I've now no idea why it's lost a whole file! Editing again...

jskeet:
Thanks Amanda - will give @frankyn a bit of a chance to comment as well - particularly on the message in UrlSigner.cs.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
```diff
@@ -37,6 +37,9 @@ namespace Google.Cloud.Storage.V1
     /// </remarks>
     public sealed partial class UrlSigner
     {
+        private const string InfiniteExpiryObsoleteMessage =
+            "This overload is obsolete, as non-expiring signed URLs are no longer supported. Please use the overload accepting a DateTimeOffset instead.";
+
```
frankyn: Technically they will still be supported until the bug is fixed. Is it more that the C# client will no longer support it until it is fixed. Or is this considered too verbose?

Suggestion:
> This overload is obsolete, as non-expiring signed URLs are a bug in the service that can cause security issues and should no longer be used. Please use the overload accepting a DateTimeOffset instead.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
```diff
@@ -239,6 +243,59 @@ internal UrlSigner WithSigningVersion(SigningVersion signingVersion)
...
+        /// in a context that should not block.
+        /// </para>
+        /// <para>
+        /// When the <paramref name="request"/> is specified, some of its headers will be included in the signed URL's
+        /// signature, and therefore must be included in requests made with the created URL. These are the Content-MD5 and
+        /// Content-Type content headers as well as any content or request header with a name starting with "x-goog-".
+        /// </para>
+        /// <para>
+        /// If <paramref name="request"/> is null, no headers are included in the signed URL's signature, so any requests
+        /// made with the created URL must not contain Content-MD5, Content-Type, or any header starting with "x-goog-".
```
frankyn: Is this w.r.t V2 or V4? IIUC you should still be able to pass in Content-Type in a V4 request if it's not part of the string_to_sign given signed headers.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
```diff
@@ -239,6 +243,59 @@ internal UrlSigner WithSigningVersion(SigningVersion signingVersion)
...
+        /// </para>
+        /// <para>
+        /// Note that if the entity is encrypted with customer-supplied encryption keys (see
+        /// https://cloud.google.com/storage/docs/encryption for more information), the <b>x-goog-encryption-algorithm</b>,
+        /// <b>x-goog-encryption-key</b>, and <b>x-goog-encryption-key-sha256</b> headers will be required when making the
+        /// request. However, only the x-goog-encryption-algorithm header is included in the signature for the signed URL.
```
frankyn: I might be wrong, but a developer could include `x-goog-encryption-key` into Signed Headers or is this more about security concerns with CSEK?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
jskeet: Fine by me. I wasn't sure how we wanted to message the service change. Done.

(We'll want to change the documentation again a bit in the second phase where we move to making "non-expiring" just expire in 2038, but that can be on the parameter.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
jskeet: This is the general documentation - but it's copied and pasted from other overloads. I'm going to propose that when we start supporting V4, we move it all to a documentation page that the remarks can reference.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
jskeet: We explicitly don't include x-goog-encryption-key in the signed headers, as per step 3 of https://cloud.google.com/storage/docs/access-control/signed-urls#about-canonical-extension-headers

jskeet:
Right, changed the message as requested - will merge on green (using my admin rights) as I believe I've now addressed Frank's comments.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
frankyn: Sg, thanks.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
frankyn: Nice! I missed that. Thanks @jskeet!

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
frankyn: Sg, thank you!

PR #2746, opened 2018-11-30
using code snippet in https://cloud.google.com/pubsub/docs/emulator

kenfesta:
Adding region tags for use in https://cloud.google.com/pubsub/docs/emulator

### DIFF
--- a/tools/Google.Cloud.Docs.Snippets/FaqSnippets.cs
+++ b/tools/Google.Cloud.Docs.Snippets/FaqSnippets.cs
@@ -27,6 +27,7 @@ namespace Google.Cloud.Tools.Snippets
         public void Emulator()
         {
             // Sample: Emulator
+            // [START pubsub_use_emulator]
             // For example, "localhost:8615"
             string emulatorHostAndPort = Environment.GetEnvironmentVariable("PUBSUB_EMULATOR_HOST");
 
@@ -37,6 +38,7 @@ namespace Google.Cloud.Tools.Snippets
             {
                 Console.WriteLine(topic.Name);
             }
+            // [START pubsub_use_emulator]
             // End sample
         }
         


jskeet:
The text from these snippets is produced *directly* in https://googleapis.github.io/google-cloud-dotnet/docs/faq.html. This pull request would start including those comments in our documentation, which we wouldn't want.

If we want to use the same samples across the two documentation systems, we'll need to modify at least one of the systems - we can't use the existing syntax for both systems in a single example.

jskeet:
A note on this - it would be *relatively* easy to change our system to just ignore any line containing "// [START" or "// [END". That may be simpler than changing the GCP-wide system. Let me know if you want me to take that on.

tools/Google.Cloud.Docs.Snippets/FaqSnippets.cs
```diff
@@ -37,6 +38,7 @@ public void Emulator()
             {
                 Console.WriteLine(topic.Name);
             }
+            // [START pubsub_use_emulator]
```
jskeet: Presumably this is intended to be END rather than START

jskeet:
(Travis failure is a failure to install .NET Core - nothing to do with this PR)

tools/Google.Cloud.Docs.Snippets/FaqSnippets.cs
kenfesta: Yes, thanks!

jskeet: I've merged a tool change, so once this is fixed I'll be able to merge.

jskeet:
What's the status of this? I'd like to either merge it after the minor fix, or close it if it's not going to be useful to you for a while.

jskeet:
What's the status of this? I'd like to either merge it after the minor fix, or close it if it's not going to be useful to you for a while.

kir-titievsky:
Reviewers: any pointers on getting this merged? Something we can do?

jskeet:
@kir-titievsky: Yes, I'm just waiting for the requested typo change to be made.

kenfesta:
Will create a pull request directly from the GUI editor.

PR #2749, opened 2018-12-03
Use "maximum 32-bit seconds Unix time" as "infinite" expiry

jskeet:
Signed URLs without an expiry time are not intended to work - the
expiry time is intended to be a required query parameter. This
change comes as close to being backward-compatible as is feasible.

cc @frankyn


frankyn:
Thank you both!

PR #2751, opened 2018-12-03
Skip lines containing // [START or // [END

jskeet:
This will allow DevRel to use our samples too. We should probably
have more of a consistent single system at some point, but this is a
simple short-term fix.


jskeet:
The latter - we're not expecting the DevRel system to change at all.

Issue #2694, opened 2018-11-12
I never expected "foreach (var obj in objectsInBucket)" to throw an exception.


SurferJeffAtGoogle:
#### Environment details

  - OS:

Windows.

  - .NET version:

All.
  
- Package name and version:
```
    <PackageReference Include="Google.Cloud.Storage.V1" Version="2.3.0-beta02" />
```
#### Steps to reproduce

Full code here:
https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/5d0fd9bab0d23a801c0dd3546ba815a865e80c5e/testutil/TestBucket.cs

I never expected the foreach statement in the code below to throw an exception.

```
                PagedEnumerable<Google.Apis.Storage.v1.Data.Objects,
                    Google.Apis.Storage.v1.Data.Object> objectsInBucket;
                try
                {
                    objectsInBucket = _storage.ListObjects(BucketName);
                }
                catch (Google.GoogleApiException e)
                when (e.Error.Code == 404)
                {
                    return;  // Bucket does not exist.  Ok.
                }

                // Try to delete each object in the bucket.
                foreach (var obj in objectsInBucket)
```

If there are no objects in the bucket, I expect the iterator to be empty.  If the bucket does not exist, I expect the call to ListObjects() to throw an exception.

SurferJeffAtGoogle:
Here's the exception:
```
[xUnit.net 00:01:00.2953631]       Google.GoogleApiException : Google.Apis.Requests.RequestError
[xUnit.net 00:01:00.2955119]       Not Found [404]
[xUnit.net 00:01:00.2955827]       Errors [
[xUnit.net 00:01:00.2956416]       	Message[Not Found] Location[ - ] Reason[notFound] Domain[global]
[xUnit.net 00:01:00.2959794]       ]
[xUnit.net 00:01:00.2960785]       
```

jskeet:
Hmm. I do have sympathy for this, but making the operation "fully lazy" makes various other aspects simpler and more consistent. It's also consistent with other lazy loading scenarios, e.g. LINQ against a database, where constructing the `IQueryable<T>` won't fail if the table has been deleted. I do understand that's a less likely scenario than a bucket not being present. (Note that this isn't just for storage - *all* our APIs that use pagination will behave the same way.)

Additionally, we can't change the design now even if after careful consideration we decided we wanted to, as moving when the exception is thrown would be a breaking change.

[...]

PR #2753, opened 2018-12-03
First part of clarifying the behavior of list operations

jskeet:
After comments on this first commit (the guide and Storage only)
have been addressed, I'll add another commit handling BigQuery.

Addresses #2694.

@SurferJeffAtGoogle please review.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.ListBuckets.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.ListBuckets.cs
@@ -26,7 +26,10 @@ namespace Google.Cloud.Storage.V1
         /// </summary>
         /// <remarks>
         /// This method fetches the buckets lazily, making requests to the underlying service
-        /// for a page of results at a time, as required.
+        /// for a page of results at a time, as required. No network requests are made until the returned sequence is enumerated.
+        /// This means that any exception due to an invalid request will be deferred until that time. Callers should be prepared
+        /// for exceptions to be thrown while enumerating the results. In addition to failures due to invalid requests, network
+        /// or service failures can cause exceptions even after the first results have been returned.
         /// </remarks>
         /// <param name="projectId">The ID of the project to list the buckets from. Must not be null.</param>
         /// <param name="options">The options for the operation. May be null, in which case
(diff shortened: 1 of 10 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.ListBuckets.cs
```diff
@@ -26,7 +26,8 @@ public abstract partial class StorageClient
         /// </summary>
         /// <remarks>
         /// This method fetches the buckets lazily, making requests to the underlying service
-        /// for a page of results at a time, as required.
+        /// for a page of results at a time, as required. No network requests are made until the returned sequence
+        /// is enumerated, so any exception due to an invalid request will be deferred until that time.
```
SurferJeffAtGoogle: I copied the paragraph:

> This method fetches the buckets lazily, making requests to the underlying service for a page of results at a time, as required. No network requests are made until the returned sequence is enumerated, so any exception due to an invalid request will be deferred until that time.

into [Hemingway app](http://hemingwayapp.com/), and it scores like this:

Readability
Grade 13
OK.  Aim for 9.

I also think we should aim for 9.  For many readers, English is not their first language, and for readers with a college education, simpler language will communicate more quickly.

One phrase I had to reread to parse correctly:
>  for a page of results at a time

At what time?  Oh, one *page at a time.*

Also,

> so any exception due to an invalid request will be deferred until that time

It took me a while to translate this sentence into specific advice for myself: when I try to pull exceptions out of the enumeration later, it may throw an exception.  I need to prepare to catch it.

I tried rewriting the paragraph for 5 minutes, but I'm still not satisfied with my results.  I can put some more time and thought into it and propose alternate text, if you like.

jskeet: This gets to grade 9, basically by splitting the new sentence:

>  This method fetches the buckets lazily, making requests to the underlying service for a page of results at a time, as required. No network requests are made until the returned sequence is enumerated. This means that any exception due to an invalid request will be deferred until that time. 

I'm personally okay with the "for a page at a time" part - I *suspect* the line break was working against you, and of course that wouldn't be in the docs the user is reading. I can't think of any alternatives.

In terms of the advice, we could add this to the end:

> Callers should be prepared for exceptions to be thrown while enumerating the results. In addition to failures due to invalid requests, network or service failures can cause exceptions even after the first results have been returned.

I wouldn't want users thinking that it's only retrieving the *first* item that can cause a problem, even if that's the most likely place.

SurferJeffAtGoogle: With the addendum, I like it!

jskeet: Great! Will fix storage and amend this first commit, then roll out a second commit for BigQuery, then merge. Yay.

jskeet:
@amanda-tarafa: I've pushed the changes now - PTAL to make sure I haven't done something totally stupid. The comments are cut/pasted everywhere, so just checking a smattering in each commit should be fine.

jskeet:
Whoops, looks like I accidentally included a project file change. Will fix that now!

PR #2757, opened 2018-12-05
(NOT FOR SUBMISSION) Example of what a client builder might look like

jskeet:
I'll create a second PR for DatastoreDb that has all kinds of interesting bits around emulator support, project ID etc - but that can be separate.
This is a more "vanilla" example.

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
@@ -343,7 +343,7 @@ namespace Google.Api.Gax.Grpc
         public Task<TClient> BuildAsync(CancellationToken cancellationToken = default)
         {
             Validate();
-            return BuildImplAsync();
+            return BuildImplAsync(cancellationToken);
         }
 
         /// <summary>
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
@@ -354,7 +354,7 @@ namespace Google.Api.Gax.Grpc
         /// <summary>
         /// Builds the resulting client asynchronously. Validation will already have been performed by the time this is called.
         /// </summary>
-        protected abstract Task<TClient> BuildImplAsync(CancellationToken cancellationToken = default);
+        protected abstract Task<TClient> BuildImplAsync(CancellationToken cancellationToken);
 
     }
 }
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
@@ -38,9 +38,9 @@ namespace Google.Api.Gax.Grpc
         public ChannelCredentials ChannelCredentials { get; private set; }
 
         /// <summary>
-        /// The credential file to use, or null if credentials are being provided in a different way.
+        /// The path to the credentials file to use, or null if credentials are being provided in a different way.
         /// </summary>
-        public string CredentialsFile { get; private set; }
+        public string CredentialsPath { get; private set; }
 
         /// <summary>
         /// The credentials to use as JSON, or null if credentials are being provided in a different way.
@@ -71,12 +71,12 @@ namespace Google.Api.Gax.Grpc
         protected ClientBuilder(TBuilder builder)
         {
             Endpoint = builder.Endpoint;
-            ChannelCredentials = builder.ChannelCredentials;
             Scopes = builder.Scopes;
-            CredentialsFile = builder.CredentialsFile;
+            ChannelCredentials = builder.ChannelCredentials;
+            CredentialsPath = builder.CredentialsPath;
             JsonCredentials = builder.JsonCredentials;
-            CallInvoker = CallInvoker;
-            ChannelCredentials = null;
+            TokenAccessMethod = builder.TokenAccessMethod;
+            CallInvoker = builder.CallInvoker;
         }
 
         /// <summary>
@@ -86,7 +86,7 @@ namespace Google.Api.Gax.Grpc
         /// </summary>
         protected virtual void ClearCredentials()
         {
-            CredentialsFile = null;
+            CredentialsPath = null;
             JsonCredentials = null;
             TokenAccessMethod = null;
             CallInvoker = null;
@@ -139,14 +139,14 @@ namespace Google.Api.Gax.Grpc
         /// <remarks>
         /// This method overrides any credentials specified with previous calls.
         /// </remarks>
-        /// <param name="file">The credential file to use, or null to use the default credentials.</param>
-        /// <returns>A builder with the specified credential file.</returns>
-        public TBuilder WithCredentialsFile(string file)
+        /// <param name="path">The path to the credentials file to use, or null to use the default credentials.</param>
+        /// <returns>A builder with the specified credentials path.</returns>
+        public TBuilder WithCredentialsPath(string path)
         {
             // Note: don't load this until we build, so that any scope changes will be effective.
             var clone = Clone();
             clone.ClearCredentials();
-            clone.CredentialsFile = file;
+            clone.CredentialsPath = path;
             return clone;
         }
 
@@ -196,7 +196,7 @@ namespace Google.Api.Gax.Grpc
         /// Creates a new builder based on this one, but with the specified gRPC call invoker.
         /// </summary>
         /// <para>
-        /// This method removes any credentials specified with previous calls.
+        /// This method removes any credentials and scopes specified with previous calls.
         /// </para>
         /// <param name="callInvoker">The gRPC call invoker to use, or null to use the default call invoker.</param>
         /// <returns>A builder with the specified call invoker.</returns>
@@ -205,6 +205,7 @@ namespace Google.Api.Gax.Grpc
             var clone = Clone();
             clone.ClearCredentials();
             clone.CallInvoker = callInvoker;
+            clone.Scopes = null;
             return clone;
         }
 
@@ -251,9 +252,9 @@ namespace Google.Api.Gax.Grpc
                 return ChannelCredentials;
             }
             GoogleCredential unscoped;
-            if (CredentialsFile != null)
+            if (CredentialsPath != null)
             {
-                unscoped = GoogleCredential.FromFile(CredentialsFile);
+                unscoped = GoogleCredential.FromFile(CredentialsPath);
             }
             else if (JsonCredentials != null)
             {
@@ -277,10 +278,10 @@ namespace Google.Api.Gax.Grpc
                 return ChannelCredentials;
             }
             GoogleCredential unscoped;
-            if (CredentialsFile != null)
+            if (CredentialsPath != null)
             {
                 // TODO: Use an async overload when one is available
-                unscoped = GoogleCredential.FromFile(CredentialsFile);
+                unscoped = GoogleCredential.FromFile(CredentialsPath);
             }
             else if (JsonCredentials != null)
             {
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
@@ -1,0 +1,322 @@
+﻿/*
+ * Copyright 2018 Google LLC
+ * Use of this source code is governed by a BSD-style
+ * license that can be found in the LICENSE file or at
+ * https://developers.google.com/open-source/licenses/bsd
+ */
+using Google.Apis.Auth.OAuth2;
+using Grpc.Auth;
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// Base class for API-specific builders.
+    /// </summary>
+    /// <typeparam name="TBuilder">The type of this builder, for chaining purposes</typeparam>
+    /// <typeparam name="TClient">The type of client created by this builder.</typeparam>
+    public abstract class ClientBuilder<TBuilder, TClient> where TBuilder : ClientBuilder<TBuilder, TClient>
+    {
+        /// <summary>
+        /// The endpoint to connect to, or null to use the default endpoint.
+        /// </summary>
+        public ServiceEndpoint Endpoint { get; private set; }
+
+        /// <summary>
+        /// The scopes to use, or null to use the default scopes.
+        /// </summary>
+        public IReadOnlyList<string> Scopes { get; private set; }
+
+        /// <summary>
+        /// The channel credentials to use, or null if credentials are being provided in a different way.
+        /// </summary>
+        public ChannelCredentials ChannelCredentials { get; private set; }
+
+        /// <summary>
+        /// The credential file to use, or null if credentials are being provided in a different way.
+        /// </summary>
+        public string CredentialsFile { get; private set; }
+
+        /// <summary>
+        /// The credentials to use as JSON, or null if credentials are being provided in a different way.
+        /// </summary>
+        public string JsonCredentials { get; private set; }
+
+        /// <summary>
+        /// The token access method to use, or null if credentials are being provided in a different way.
+        /// </summary>
+        public Func<string, CancellationToken, Task<string>> TokenAccessMethod { get; private set; }
+
+        /// <summary>
+        /// The call invoker to use, or null to create the call invoker when the client is built.
+        /// </summary>
+        public CallInvoker CallInvoker { get; private set; }
+
+        /// <summary>
+        /// Creates a new instance with no settings.
+        /// </summary>
+        protected ClientBuilder()
+        {
+        }
+        
+        /// <summary>
+        /// Creates a clone of the given builder, with respect to settings in this class.
+        /// </summary>
+        /// <param name="builder">The builder to clone</param>
+        protected ClientBuilder(TBuilder builder)
+        {
+            Endpoint = builder.Endpoint;
+            ChannelCredentials = builder.ChannelCredentials;
+            Scopes = builder.Scopes;
+            CredentialsFile = builder.CredentialsFile;
+            JsonCredentials = builder.JsonCredentials;
+            CallInvoker = CallInvoker;
+            ChannelCredentials = null;
+        }
+
+        /// <summary>
+        /// This method should be called by all "With" methods that change the credentials.
+        /// If a subclass supports additional forms of credentials, it should override this method
+        /// to clear them and call this implementation.
+        /// </summary>
+        protected virtual void ClearCredentials()
+        {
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+
+        /// <summary>
+        /// The channel credentials to use, or null if credentials are being provided in a different way.
+        /// </summary>
+        public ChannelCredentials ChannelCredentials { get; private set; }
+
+        /// <summary>
+        /// The credential file to use, or null if credentials are being provided in a different way.
+        /// </summary>
+        public string CredentialsFile { get; private set; }
```
chrisdunelm: Might want to name this `CredentialsFilePath` of something, to make it clear it's a file-system path. (Assuming that's what it actually is!)

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+        {
+        }
+        
+        /// <summary>
+        /// Creates a clone of the given builder, with respect to settings in this class.
+        /// </summary>
+        /// <param name="builder">The builder to clone</param>
+        protected ClientBuilder(TBuilder builder)
+        {
+            Endpoint = builder.Endpoint;
```
chrisdunelm: super-nit: order these the same as the properties.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+            ChannelCredentials = builder.ChannelCredentials;
+            Scopes = builder.Scopes;
+            CredentialsFile = builder.CredentialsFile;
+            JsonCredentials = builder.JsonCredentials;
+            CallInvoker = CallInvoker;
+            ChannelCredentials = null;
```
chrisdunelm: Why `null`?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+        /// </summary>
+        protected ClientBuilder()
+        {
+        }
+        
+        /// <summary>
+        /// Creates a clone of the given builder, with respect to settings in this class.
+        /// </summary>
+        /// <param name="builder">The builder to clone</param>
+        protected ClientBuilder(TBuilder builder)
```
chrisdunelm: Is `TokenAccessMethod` missing for a reason?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+
+        /// <summary>
+        /// Creates a new builder based on this one, but with the specified credential scopes.
+        /// </summary>
+        /// <param name="scopes">The scopes to use when creating credentials, or null to use the default set of scopes.</param>
+        /// <returns>A builder with the specified scopes.</returns>
+        public TBuilder WithScopes(IEnumerable<string> scopes)
+        {
+            var clone = Clone();
+            clone.Scopes = scopes?.ToList().AsReadOnly();
```
chrisdunelm: nit: Is `.AsReadOnly()` necessary? It's only ever exposed as an `IReadOnlyList<>`

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+        /// <returns>A builder with the specified JSON credentials.</returns>
+        public TBuilder WithJsonCredentials(string json)
+        {
+            var clone = Clone();
+            clone.ClearCredentials();
+            clone.JsonCredentials = json;
+            return clone;
+        }
+
+        // TODO: Handle the fact that GoogleCredential implements ITokenAccess explicitly. That means the sample here doesn't work
```
chrisdunelm: We can make whatever changes to the auth code to help with this.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientBuilder.cs
```diff
@@ -0,0 +1,74 @@
...
+        public DatastoreClientBuilder() : base()
+        {
+        }
+
+        /// <summary>
+        /// Returns a clone of the settings in this builder, or null if no settings have been set. Changes to the
+        /// returned object do not affect the settings in this builder.
+        /// </summary>
+        /// <returns>A clone of the settings in this builder, or null.</returns>
+        public DatastoreSettings CloneSettings() => _settings?.Clone();
```
chrisdunelm: Hmm, I'm uncertain about this.
To match the base class pattern, this should be a property called `Settings`.
I'm guessing that you've done it this way to (a) make it clear by the name that it's a clone of the settings; and (b) made it a method because it's not returning the same instance each time.
However, in some ways I'd prefer it to be the property, then in the xmldoc make it clear that it clones. I found the name `CloneSettings` initially confusing.

Alternatively we could code-gen a second, immutable, settings class; which is returned here. This second class could either use composition, or could be a base class of the mutable settings class...

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+            {
+                unscoped = GoogleCredential.FromFile(CredentialsFile);
+            }
+            else if (JsonCredentials != null)
+            {
+                unscoped = GoogleCredential.FromJson(JsonCredentials);
+            }
+            else
+            {
+                unscoped = GoogleCredential.GetApplicationDefault();
```
amanda-tarafa: I would put this in a virtual GetDefaultCredentials() method and call here the GetDefaultCredentials() method in its place.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+
+        /// <summary>
+        /// Creates a new builder based on this one, but with the specified gRPC call invoker.
+        /// </summary>
+        /// <para>
+        /// This method removes any credentials specified with previous calls.
+        /// </para>
+        /// <param name="callInvoker">The gRPC call invoker to use, or null to use the default call invoker.</param>
+        /// <returns>A builder with the specified call invoker.</returns>
+        public TBuilder WithCallInvoker(CallInvoker callInvoker)
```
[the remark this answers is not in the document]
jskeet: Yes, probably. Will do.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+        /// </summary>
+        /// <param name="builder">The builder to clone</param>
+        protected ClientBuilder(TBuilder builder)
+        {
+            Endpoint = builder.Endpoint;
+            ChannelCredentials = builder.ChannelCredentials;
+            Scopes = builder.Scopes;
+            CredentialsFile = builder.CredentialsFile;
+            JsonCredentials = builder.JsonCredentials;
+            CallInvoker = CallInvoker;
```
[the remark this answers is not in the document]
jskeet: Yup, will fix.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,322 @@
...
+            ChannelCredentials = null;
+        }
+
+        /// <summary>
+        /// This method should be called by all "With" methods that change the credentials.
+        /// If a subclass supports additional forms of credentials, it should override this method
+        /// to clear them and call this implementation.
+        /// </summary>
+        protected virtual void ClearCredentials()
```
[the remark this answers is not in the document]
jskeet: Hmm. We could do that, and just replace the whole credential settings reference with a new one - the properties would need to proxy through. I don't think we'd want to make that the public API though.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Any particular reason? I could see it perhaps for testing, but that's all.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Yes, that's what it is. But having made the change, I don't like it :( Two reasons:

- FilePath sounds a bit redundant. It's a filename or a path
- Path sounds like it *might* be a set of places to look, like the PATH environment variable

`File.*` methods all take a parameter called `path`, so `CredentialsPath` might be okay. Or `CredentialsFilename` perhaps. But I don't like the mixture.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Done.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Nope, fixed.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: It could be cast.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Great. I've had a couple of ideas, but we can explore those later.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientBuilder.cs
jskeet: I think I'd like to stick with a method. Would `GetClonedSettings()` be better?

jskeet:
I *think* this is now what we decided on. Let me know if I've missed something.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,357 @@
...
+            Endpoint = builder.Endpoint;
+            Scopes = builder.Scopes;
+            ChannelCredentials = builder.ChannelCredentials;
+            CredentialsPath = builder.CredentialsPath;
+            JsonCredentials = builder.JsonCredentials;
+            TokenAccessMethod = builder.TokenAccessMethod;
+            CallInvoker = builder.CallInvoker;
+        }
+
+        // TODO: WithEndpoint(string) and WithEndpoint(string, int)?
```
amanda-tarafa: If ServiceEndpoint already has constructors or Create methods with these overloads, maybe you can skip this TODO or leave for way into the future.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,357 @@
...
+        /// <param name="values">The values to check for nullity.</param>
+        /// <exception cref="InvalidOperationException">More than one value is null.</exception>
+        protected void ValidateAtMostOneNotNull(string message, params object[] values)
+        {
+            int notNull = values.Count(v => v == null);
+            GaxPreconditions.CheckState(notNull < 2, message);
+        }
+
+        /// <summary>
+        /// Validates that if <paramref name="controlling"/> is not null, then no value in <paramref name="values"/> is null.
```
amanda-tarafa: nit: ... "then all values in value are null"

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,357 @@
...
+        /// Builds the resulting client asynchronously.
+        /// </summary>
+        public Task<TClient> BuildAsync(CancellationToken cancellationToken = default)
+        {
+            Validate();
+            return BuildImplAsync();
+        }
+
+        /// <summary>
+        /// Builds the resulting client. Validation will
```
amanda-tarafa: nit: Validation will...............

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,357 @@
...
+    /// Base class for API-specific builders.
+    /// </summary>
+    /// <typeparam name="TBuilder">The type of this builder, for chaining purposes</typeparam>
+    /// <typeparam name="TClient">The type of client created by this builder.</typeparam>
+    public abstract class ClientBuilder<TBuilder, TClient> where TBuilder : ClientBuilder<TBuilder, TClient>
+    {
+        /// <summary>
+        /// The endpoint to connect to, or null to use the default endpoint.
+        /// </summary>
+        public ServiceEndpoint Endpoint { get; private set; }
```
amanda-tarafa: Just to be sure, both ServiceEndpoint, ChannelCredentials and CallInvoker are inmutable, right? If they are not we should be consistent and follow the same pattern we are using for Settings on builder's implementations.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientBuilder.cs
```diff
@@ -0,0 +1,83 @@
...
+            }
+
+            /// <summary>
+            /// Returns a clone of the settings in this builder, or null if no settings have been set. Changes to the
+            /// returned object do not affect the settings in this builder.
+            /// </summary>
+            /// <returns>A clone of the settings in this builder, or null.</returns>
+            public DatastoreSettings GetClonedSettings() => _settings?.Clone();
+
+            private Builder(Builder builder) : base(builder)
```
amanda-tarafa: Super nit: can you put both constructors together?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: ServiceEndpoint is. ChannelCredentials and CallInvoker are abstract classes - it's possible that the caller will provide a mutable version, but we don't have any way of cloning them, and I think that's basically a corner case a user could only get into deliberately.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: ServiceEndpoint has a (string, int) constructor. It should probably have a Parse(string) method as well. But even then, I think it would be good to provide the methods for the sake of convenience. It's very likely that if someone *does* have an endpoint, they've either got a string or a string and integer from a configuration file.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Whoops, will fix.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Thanks!

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientBuilder.cs
jskeet: Will do, yes.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
amanda-tarafa: OK, if the user does that on their own, then, that's fine.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
amanda-tarafa: OK

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,360 @@
...
+
+        /// <summary>
+        /// The channel credentials to use, or null if credentials are being provided in a different way.
+        /// </summary>
+        public ChannelCredentials ChannelCredentials { get; private set; }
+
+        /// <summary>
+        /// The path to the credentials file to use, or null if credentials are being provided in a different way.
+        /// </summary>
+        public string CredentialsPath { get; private set; }
```
SurferJeffAtGoogle: Are these credentials also encoded as Json?  I see confusion between the name of this property and the one below.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,360 @@
...
+
+        /// <summary>
+        /// The credentials to use as JSON, or null if credentials are being provided in a different way.
+        /// </summary>
+        public string JsonCredentials { get; private set; }
```
SurferJeffAtGoogle: Why is the setter private?

jskeet: Having said that, we could go in a very different direction, making the builder totally mutable and just using properties. That wasn't going to be feasible back when changing one credential source would wipe out the others etc, but it's more realistic now. I'll add an alternative section to the design doc and get feedback. All this back and forth takes a bit longer, but I want to get it right :)

jskeet:
Closing as we're almost certainly going go with the property-based one.

PR #2759, opened 2018-12-06
Add Google Cloud Scheduler (no release yet)

jskeet:
We may well want a few snippets or integration tests just to check it works - this is all autogenerated so far.


jskeet:
Right - will merge as-is, and add some snippets (as a handy form of integration test) in a new PR.

jskeet:
Right - will merge as-is, and add some snippets (as a handy form of integration test) in a new PR.

jskeet:
Hmm. Turns out this only works if the project has an AppEngine app, and that has a location which is supported by Cloud Scheduler. (Mine is in the wrong location...) So I think we'll leave it without any integration tests right now...

amanda-tarafa:
Does that means that the generated Snippets won't pass?

jskeet:
The generated snippets aren't executable tests.

amanda-tarafa:
Ah OK, good to know!

PR #2761, opened 2018-12-10
Regenerate APIs

jskeet:


apis/Google.Cloud.Dataproc.V1/Google.Cloud.Dataproc.V1/WorkflowTemplateServiceClient.cs
```diff
@@ -1621,13 +1621,13 @@ public virtual WorkflowTemplateService.WorkflowTemplateServiceClient GrpcClient
         /// A pageable sequence of <see cref="WorkflowTemplate"/> resources.
         /// </returns>
         public virtual gax::PagedEnumerable<ListWorkflowTemplatesResponse, WorkflowTemplate> ListWorkflowTemplates(
-            string parent,
+            RegionName parent,
```
chrisdunelm: These are breaking changes.
I assume the previous version hasn't been GA released?

jskeet: It's effectively undoing the changes in commit 84d3f33ba1f37453c55376f182c567af79cf92e7. But no, it isn't GA anyway :)

Issue #2764, opened 2018-12-11
BigQuery: System.OutOfMemoryException thrown when deleting data from large tables


jaredhughes1012:
When executing "Delete" queries on large BigQuery tables (100 million+ rows), a System.OutOfMemoryException is being thrown from the BigQuery .NET driver. Occasionally, this will cause the process to terminate unexpectedly instead of throwing an exception.

An example table with this issue has 1.2 billion rows, 26 columns, and 292 GB of data

#### Environment details

  - OS: Ubuntu 18.04
  - .NET version: .NET Core 2.1.401
  - Package name and version: Google.Cloud.BigQuery.V2 1.3.0-beta05

#### Steps to reproduce

  1. Create a table with at least 100 million rows of data in BigQuery
  2. Execute the following code:
```
var credential = GetGoogleCredential();
using (var client = await BigQueryClient.CreateAsync("projectId", credential)) {
   client.ExecuteQuery("DELETE FROM `projectId.dataset.table` WHERE TRUE");
}
```
#### Stack Trace
[...]

Issue #2559, opened 2018-10-04
BigQueryResults.TotalRows throws InvalidOperationException


ef-computas:
version: Google.Cloud.BigQuery.V2 1.3.0-beta05 and Google.Cloud.BigQuery.V2 1.2.0

Running a query that selects rows from a source table and appends to a destination table:

```
var options = new QueryOptions
{
    UseLegacySql = false,
    AllowLargeResults = true,
    DestinationTable = destinationTableRef,
    JobLocation = "EU",
    WriteDisposition = WriteDisposition.WriteAppend,
    ProjectId = projectId,
    DefaultDataset = client.GetDatasetReference(projectId, datasetId),
    Priority = QueryPriority.Interactive,
    DryRun = false
};

var query = @"SELECT *
                   FROM `" + sourceTable + "` " +
                   @"WHERE _PARTITIONTIME <= TIMESTAMP(DATE_SUB(CURRENT_DATE(), INTERVAL 1 DAY))";

var bigQueryResults = client
                .ExecuteQuery(query, null, queryOptions)
                .ThrowOnAnyError();

var totalRows = bigQueryResults.TotalRows; // this throws: InvalidOperationException: Nullable object must have a value. 
```
The query seem to run ok, but an exception is thrown when attempting to access bigQueryResults.TotalRows:

System.InvalidOperationException: Nullable object must have a value.
  at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
  at Google.Cloud.BigQuery.V2.BigQueryResults.get_TotalRows()

jskeet:
Hmm... that's worrying. That suggests the underlying `TotalRows` property is null. I don't know why that would be. I'd expect that to always bye populated on a successful query. I'll have to look into this. I'm afraid I'm travelling over the next couple of days, which is going to delay things somewhat, although I'll see if I can find pockets of time.

Does this happen every time, or just occasionally? It would be good to be able to reproduce this myself. If you access the rows themselves, do you get the data you expect?

amanda-tarafa:
@ef-computas I haven't been able to reproduce this issue. It would be of great help if you could provide more info to try and reproduce as faithfully as possible. For instance:

- How many rows are you expecting to get from your query?
- Have you tried accessing some of the other properties on `bigQueryResults`, what do you get? Specifically as @jskeet asked, can you access the rows themselves, and you get what you expected?
- How long does it take the query to (apparently) execute properly?

[...]

PR #2765, opened 2018-12-12
Provide nullable row counts

jskeet:
I'm open to alternative names instead of `SafeTotalRows` - or indeed alternative approaches.

I won't merge this PR until I've performed further investigation of #2764, but feedback is still welcomed.


jskeet:
Merging now as this provides a helpful workaround for #2559, even though it doesn't affect #2764.

PR #2769, opened 2018-12-13
(NOT FOR SUBMISSION) Example of what a client builder based on properties might look like

jskeet:


apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+    {
+        /// <summary>
+        /// The endpoint to connect to, or null to use the default endpoint.
+        /// </summary>
+        public ServiceEndpoint Endpoint { get; set; }
+
+        /// <summary>
+        /// The scopes to use, or null to use the default scopes.
+        /// </summary>
+        public IReadOnlyList<string> Scopes { get; set; }
```
amanda-tarafa: nit: Makes no sense for this to be readonly now.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// The credentials to use as a JSON string, or null if credentials are being provided in a different way.
+        /// </summary>
+        public string JsonCredentials { get; set; }
+
+        /// <summary>
+        /// The token access method to use, or null if credentials are being provided in a different way.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// To use a GoogleCredential for credentials, call this method using a method group conversion, e.g.
```
amanda-tarafa: nit: ", set this property to a method group conversion" (and change the example)

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// <c>builder = builder.WithTokenAccessMethod(credential.GetAccessTokenForRequestAsync);</c>
+        /// </para>
+        /// </remarks>
+        public Func<string, CancellationToken, Task<string>> TokenAccessMethod { get; set; }
+
+        /// <summary>
+        /// The call invoker to use, or null to create the call invoker when the client is built.
```
amanda-tarafa: nit: Maybe add: "or null to create the call invoker from all other properties..."

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+            }
+            else if (JsonCredentials != null)
+            {
+                unscoped = GoogleCredential.FromJson(JsonCredentials);
+            }
+            else
+            {
+                unscoped = GoogleCredential.GetApplicationDefault();
+            }
+            return unscoped.CreateScoped(Scopes ?? GetDefaultScopes()).ToChannelCredentials();
```
amanda-tarafa: If CreateScoped does not clone or copy Scopes in some way, this might be the time to turn it into a readonly list.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+            else if (JsonCredentials != null)
+            {
+                unscoped = GoogleCredential.FromJson(JsonCredentials);
+            }
+            else
+            {
+                // TODO: Use cancellationToken when an overload accepting one is available
+                unscoped = await GoogleCredential.GetApplicationDefaultAsync().ConfigureAwait(false);
+            }
+            return unscoped.CreateScoped(Scopes ?? GetDefaultScopes()).ToChannelCredentials();
```
amanda-tarafa: Same here with potentially making Scopes a read only list.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Whoops, yes, will do.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Makes sense, will do.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: I think it still does, but for a different reason: we don't *need* to be able to write to it. We could potentially make it `IEnumerable<string>` instead, admittedly.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/ClientBuilder.cs
jskeet: Good point - it looks like it really doesn't do that. We don't particularly need it to be read-only, but we should create a one-time copy.

jskeet:
Closing as this is now in GAX 2.7.0-beta01

PR #2770, opened 2018-12-14
(Not for submission) Potential Datastore emulator configuration

jskeet:
This builds on #2769, but with a DatastoreDbBuilder. Sample:

```csharp
var db = new DatastoreDbBuilder
{
    ProjectId = "my-project",
    EmulatorBehavior = EmulatorBehavior.ProductionOrEmulator
}.Build();
```

We probably want to review the whole interaction between the base class and subclasses in terms of validation, but that's relatively straightforward.

(We might also want a `DatastoreDbBuilder` constructor accepting a project ID.)

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
@@ -1,0 +1,247 @@
+﻿// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using Google.Api.Gax.Grpc;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Datastore.V1
+{
+    // Note: this file would be generated.
+
+    public partial class DatastoreDb
+    {
+        /// <summary>
+        /// Creates a new <see cref="DatastoreDbBuilder"/> instance. This method is primarily present for discoverability.
+        /// </summary>
+        /// <returns>A new <see cref="DatastoreDbBuilder"/> instance.</returns>
+        public DatastoreDbBuilder CreateBuilder() => new DatastoreDbBuilder();
+
+    }
+
+    /// <summary>
+    /// Builder class for <see cref="DatastoreClient"/> to provide simple configuration of credentials, endpoint etc.
+    /// </summary>
+    public sealed partial class DatastoreDbBuilder : ClientBuilder<DatastoreDb>
+    {
+        /// <summary>
+        /// The settings to use for RPCs, or null for the default settings.
+        /// </summary>
+        public DatastoreSettings Settings { get; set; }
+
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+                        ApplyEmulatorConfiguration();
+                        break;
+                    case EmulatorDetection.ProductionOrEmulator:
+                        if (emulatorConfig.Endpoint != null)
+                        {
+                            ApplyEmulatorConfiguration();
+                        }
+                        else
+                        {
+                            GaxPreconditions.CheckState(emulatorConfig.ProjectId == null, "{0} should not be set when {1} is not", EmulatorConfiguration.EmulatorProjectVariable, EmulatorConfiguration.EmulatorProjectVariable);
```
amanda-tarafa: nit: The second format parameter should the one for Endpoint not the one for ProjectId.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+                        }
+                        break;
+                }
+
+                void ApplyEmulatorConfiguration()
+                {
+                    // TODO: Do we want to require this? Or should a configured emulator override any credentials?
```
amanda-tarafa: I'd say we don't want to require this, so that client code can stay the same in any environment. If there's an emulator configured and we should use it, then, we use that overriding.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+                    GaxPreconditions.CheckState(clientBuilder.ChannelCredentials == null, "{0} should not be set when connecting to an emulator", nameof(clientBuilder.ChannelCredentials));
+                    GaxPreconditions.CheckState(clientBuilder.CredentialsPath == null, "{0} should not be set when connecting to an emulator", nameof(clientBuilder.CredentialsPath));
+                    GaxPreconditions.CheckState(clientBuilder.JsonCredentials == null, "{0} should not be set when connecting to an emulator", nameof(clientBuilder.JsonCredentials));
+                    GaxPreconditions.CheckState(clientBuilder.Scopes == null, "{0} should not be set when connecting to an emulator", nameof(clientBuilder.Scopes));
+                    GaxPreconditions.CheckState(clientBuilder.TokenAccessMethod == null, "{0} should not be set when connecting to an emulator", nameof(clientBuilder.TokenAccessMethod));
+                    clientBuilder.Endpoint = emulatorConfig.Endpoint;
+                    clientBuilder.ChannelCredentials = Grpc.Core.ChannelCredentials.Insecure;
+                    projectId = projectId ?? emulatorConfig.ProjectId;
```
amanda-tarafa: In turn, I would invert this to:
emulatorConfig.ProjectId ?? projectId
This method after all is the one that applies the emulator configuration so that should have priority.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+                {
+                    projectId = null;
+                }
+                var endpoint = string.IsNullOrEmpty(hostAndPort) ? null : ParseEndpoint(hostAndPort);
+                return new EmulatorConfiguration(endpoint, projectId);
+            }
+
+            // TODO: Move this to ServiceEndpoint, and add lots of tests.
+            private static ServiceEndpoint ParseEndpoint(string text)
+            {
```
amanda-tarafa: Maybe try using System.Uri or System.UriBuilder instead of parsing. I'm almost certain it should work.

jskeet:
Closing this PR as we don't need any more eyes on it for now. The branch will remain though. Will reopen the PR when there are tests etc.

PR #2775, opened 2018-12-18
Swallow exceptions when trying to complete a stream in Firestore listeners

jskeet:
(And then release)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/WatchStream.cs
```diff
@@ -180,11 +180,19 @@ async Task CloseStreamAsync()
...
+                        var completeTask = underlyingStream.TryWriteCompleteAsync();
+                        if (completeTask != null)
+                        {
+                            await completeTask.ConfigureAwait(false);
+                        }
+                    }
+                    catch (RpcException)
                     {
-                        await completeTask.ConfigureAwait(false);
+                        // Swallow gRPC errors when trying to "complete" the stream. This may be in response to the network connection
```
amanda-tarafa: Just wondering if it makes sense to do the same thing that is done on line 170, check for e.Status and extend the backoff. Not sure that 
a) we will ever get a ResourceExhausted in this case
b) if we get it, the Resource in ResourceExhausted references the underlyingStream or some backend resource. If the latter then maybe we would want to extend the backoff here.

jskeet: We're only likely to experience this in response to another error anyway, so I don't think we need to extend the backoff - we wouldn't want to extend the backoff twice for a single logical problem.

Issue #2763, opened 2018-12-11
PubSub: Unobserved InvalidOperationException in AsyncSingleRecvQueue


lkts:
Hi, 

My application that uses PubSub has logging handler for `TaskScheduler.UnobservedTaskException`. 
Recently i have found out that such exceptions are constantly produced with stacktrace leading to PubSub (can be up to hundred in 10 minutes).
Exception:
```
System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. 
As a result, the unobserved exception was rethrown by the finalizer thread. (An attempt was made to transition a task to a final state when it had already completed.) 
---> System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.
   at Google.Cloud.PubSub.V1.SubscriberClientImpl.AsyncSingleRecvQueue`1.Enqueue(T item)
   at Google.Cloud.PubSub.V1.SubscriberClientImpl.SingleChannel.<>c__DisplayClass39_0.<Add>b__0(Task _)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
   --- End of inner exception stack trace ---
---> (Inner Exception #0) System.InvalidOperationException: An attempt was made to transition a task to a final state when it had already completed.
   at Google.Cloud.PubSub.V1.SubscriberClientImpl.AsyncSingleRecvQueue`1.Enqueue(T item)
   at Google.Cloud.PubSub.V1.SubscriberClientImpl.SingleChannel.<>c__DisplayClass39_0.<Add>b__0(Task _)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)<---
```

Looks like race in code below since it is not in `lock`.
https://github.com/googleapis/google-cloud-dotnet/blob/2dc1fb754463a7b08c58e999e3b40c28689c8818/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs#L663

I believe there are no functional bugs because of that but ideally library should not produce such things :)

Thanks.

[...]

PR #2785, opened 2019-01-03
PubSub: Avoid unobserved InvalidOperationException in AsyncSingleRecvQueue

lkts:
Fixes https://github.com/googleapis/google-cloud-dotnet/issues/2763.

This fixes unobserved exception which happened when `SetResult` was called multiple times.

~~Comment in code mentioned that problem was that continuations of `TaskCompletionSource` can run synchonously so i changed just that (i may be missing some details why it won't work). Also it is generally [recommended](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#always-create-taskcompletionsourcet-with-taskcreationoptionsruncontinuationsasynchronously) to use `TaskCreationOptions.RunContinuationsAsynchronously`.~~ - this does not build

Another approach it to use [TrySetResult](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcompletionsource-1.trysetresult?view=netcore-1.0) with current approach.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
@@ -659,7 +659,8 @@ namespace Google.Cloud.PubSub.V1
                 }
                 if (tcs != null)
                 {
-                    // Don't run in lock, as it may execute continuations synchonously.
+                    // Don't run in lock, as it may execute continuations synchronously.
+                    // Use TrySetResult as this can be called multiple times on the same tcs instance
                     tcs.TrySetResult(0);
                 }
             }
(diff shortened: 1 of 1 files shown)


jskeet:
I'll assign this to Chris as he's responsible for the PubSub code. He's on vacation until Monday, but I'm sure he'll look at it then.

lkts:
@chrisdunelm @jskeet can you describe what your release approach is? When can we expect this fix to be released?

jskeet:
@lkts: It depends on exactly what the situation is. In this case I'll let Chris think about it - I'd suggest we could probably do a 1.1.0-beta01 release fairly soon, as that doesn't stop us from adding more features for a 1.1.0 GA.

lkts:
cool, thanks

chrisdunelm:
1.1.0-beta01 is now released with this fix (and an updated gRPC dependency).

PR #2784, opened 2019-01-03
[Diagnostics] Adding Propagate Exceptions Snippets to ASP.NET Core

amanda-tarafa:
For helping troubleshoot issues similar to #2630.
(Diagnostics ASP.NET does not offer explicit Logging, only Error Reporting and Trace, so this does not apply)

PR #2789, opened 2019-01-07
Fixes integration tests fails.

amanda-tarafa:
This PR is a fix on #2784, which is making the nightly integration tests fail.

The problem is that different .Net Framework versions attempt to log with the configured logger at different points. The test has been changed to expect the propagated exception at any point and to check that the exception thrown is indeed due to the artificially introduced error.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/LoggingSnippets.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/LoggingSnippets.cs
@@ -16,6 +16,7 @@ using Google.Cloud.ClientTesting;
 using Google.Cloud.Diagnostics.Common;
 using Google.Cloud.Diagnostics.Common.IntegrationTests;
 using Google.Cloud.Logging.Type;
+using Grpc.Core;
 using Microsoft.AspNetCore.Builder;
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/LoggingSnippets.cs
```diff
@@ -69,11 +70,18 @@ public async Task Logs_Information()
...
-                await Assert.ThrowsAsync<AggregateException>(() => client.GetAsync($"/LoggingSamples/LogInformation/{_testId}"));
-            }
+                using (TestServer server = new TestServer(GetExceptionPropagatingHostBuilder()))
+                using (HttpClient client = server.CreateClient())
+                {
+                    await client.GetAsync($"/LoggingSamples/LogInformation/{_testId}");
+                }
+            });
+
+            var rpcException = aggregateEx.InnerException as RpcException;
```
jskeet: I'd just cast to RpcException here. That way we'll get to see the actual exception type if it's wrong, rather than just Assert.NotNull failing.

amanda-tarafa: Yep, done.

amanda-tarafa:
Thanks, will wait for green and merge.

PR #2797, opened 2019-01-09
Adding a snippet for wildcard queries.

amanda-tarafa:


amanda-tarafa:
Yes, I thought about that as well, only thing is this is not a proper table name, but one with a wildcard that only works in this context, so I though that in the end it would need to be harcoded somewhere, because we cannot build it ourselves. Writing it down in things to chat about next week.
Thanks! I'll wait for green and merge.

PR #2798, opened 2019-01-11
Create pubsub channels with correct unique-per-channel id

chrisdunelm:
This bug only becomes apparent with gRPC >=1.15.0 when the sub-channel sharing code changed.


chrisdunelm:
Yes, the options are compared by value, so each channel requires an option that is unique (the GUID in this case) to ensure subchannel sharing doesn't occur.

PR #2815, opened 2019-01-17
Update Storage reference and add tests for Bucket Policy Only

jskeet:
This is currently a whitelisted API, but should still be fine for our CI systems.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/BucketPolicyOnlyTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/BucketPolicyOnlyTest.cs
@@ -1,0 +1,129 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Apis.Storage.v1.Data;
+using Google.Cloud.ClientTesting;
+using System;
+using System.IO;
+using System.Net;
+using Xunit;
+
+namespace Google.Cloud.Storage.V1.IntegrationTests
+{
+    [Collection(nameof(StorageFixture))]
+    public class BucketPolicyOnlyTest
+    {
+        private readonly StorageFixture _fixture;
+
+        public BucketPolicyOnlyTest(StorageFixture fixture) => _fixture = fixture;
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


jskeet:
Thanks Amanda - will wait for Frank to have a look as well before merging.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/BucketPolicyOnlyTest.cs
```diff
@@ -0,0 +1,129 @@
...
+                Assert.InRange(lockedTime.Value.ToUniversalTime(), expectedLockedTime.AddMinutes(-10), expectedLockedTime.AddMinutes(10));
+            }
+
+            void UploadObjectSucceeds_DownloadObjectFails()
+            {
+                string objectName = IdGenerator.FromGuid();
+                var objectData = TestHelpers.GenerateData(512);
+                client.UploadObject(bucketName, objectName, "", objectData);
+                Assert.Throws<GoogleApiException>(() => client.DownloadObject(bucketName, objectName, new MemoryStream()));
+            }
```
frankyn: Does this test using granular permissions when a service account has permission to upload but not download? Not sure what changed to manage the permissions though.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/BucketPolicyOnlyTest.cs
```diff
@@ -0,0 +1,129 @@
...
+            {
+                client.PatchBucket(CreateBucketRepresentation(bucketName, true));
+
+                var fetched = client.GetBucket(bucketName);
+                Assert.True(fetched.IamConfiguration.BucketPolicyOnly.Enabled);
+                var lockedTime = fetched.IamConfiguration.BucketPolicyOnly.LockedTime;
+                Assert.NotNull(lockedTime);
+
+                var expectedLockedTime = DateTime.UtcNow.AddDays(90);
+                Assert.InRange(lockedTime.Value.ToUniversalTime(), expectedLockedTime.AddMinutes(-10), expectedLockedTime.AddMinutes(10));
```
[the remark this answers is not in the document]
jskeet: I checked in the doc that the value was expected to be stable. If it proves flaky for whatever reason, I'll take it out.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/BucketPolicyOnlyTest.cs
jskeet: I don't know the details, to be honest - but I believe this is the expected (by the Storage team :) behaviour for the case where BPO is enabled after bucket creation.

frankyn:
@jskeet I have one more ask. Based on a discussion with Jesse on the Java client, they hit an issue with projection and patching a bucket to enable BPO. I searched the code for C# GCS client and it doesn't use a default value for projection, but I don't trust myself haha. 

Could you verify that `projection=full` isn't used by default in requests?

jskeet:
@frankyn: Confirmed. We only apply a projection if the user explicitly asks us to.

frankyn:
Thank you!

PR #2818, opened 2019-01-17
Remove Google.Cloud.Firestore.V1Beta1

jskeet:
We will no longer need to depend on this, so let's remove it to avoid confusion.
The change to generateapis.sh is due to a rename in the googleapis repo.

Also updates the docs to refer to Google.Cloud.Firestore.V1


jskeet:
CI failures are due to the diff suggesting that it should try to build the now-removed projects. Can be safely ignored :)

Issue #2812, opened 2019-01-16
[BigQuery]: Integration test flakiness for jobs created with null valued labels.


amanda-tarafa:
Since early last week (Jan. 7th 2019) we've started seeing flakiness in all tests including the creation of jobs (query, copy, extract, ...) with null valued labels. After creation we get these jobs back using the get operation and the JobReference. Up until the flakiness we were getting the jobs with just the two non null valued labels, as where now we are sometimes getting the job with three labels, including the null valued one.

How to reproduce:

- Create a job with the following labels:
    "one_label" -> null
    "another-label-2" -> "label_value_2"
    "yet_another_label" -> ""
- Using the JobReference from the job created above, run the following code. The test will end up failing.
       
 ```
        [Fact]
        public void GetJob_JobLabels()
        {
            // We will run the test until it fails.
            while (true)
            {
                var client = BigQueryClient.Create("my-project-id");
                JobReference jobReference = // JobReference from the job created above.
                var job = client.GetJob(jobReference);

                output.WriteLine($"{DateTime.UtcNow} - Checking Labels for job with reference: ProjectID: {job.Reference?.ProjectId}, JobID: {job.Reference?.JobId}, Location: {job.Reference?.Location}.");

                var labels = job.Resource?.Configuration?.Labels;
                Assert.NotNull(labels);
                foreach (var label in labels)
                {
                    output.WriteLine($"{label.Key} -> {label.Value}");
                }
                // The test will eventually fail (in less than 10 minutes always) because of the
                // API side flakiness.
                Assert.Equal(2, labels.Count);
            }
        }
```

- This issue is also reproducible on the API Explorer, by using the same JobReference as above and trying the job get operation, most of the requests come back with only the two non null valued labels, but from time to time we get one with three labels including the null valued one.

This is a bug on the API side, we are opening this issue to keep track on our side and mitigate the flakiness if possible.

amanda-tarafa:
I've now created an internal issue for the BigQuery API team to look into this. This issue is blocked until we hear from them.

amanda-tarafa:
The internal issue has been fixed by the BigQuery API team. I've reactivated the flaky tests in #2819. I'll leave this issue open until Monday to check that no flakiness happens during the weekend.

amanda-tarafa:
Closing as this has now been working properly during the weekend.

PR #2819, opened 2019-01-18
Reactivating the null valued labels on job creation tests.

amanda-tarafa:
The BigQuery API team seems to have fixed the bug.
See #2812


amanda-tarafa:
Yes it was! Waiting on green and merging.

PR #2831, opened 2019-01-25
Enable clustering at table creation time

### LINKED ISSUE #2823: BigQuery: Add support for creating clustered tables
Problem:
The BigQuery library does not support creating tables with clustering enabled.

Solution:
There is already a class named Clustering that exists in various places in the library. It would be best if an instance of that class could be added as a property to CreateTableOptions as it is implemented for TimePartitioning.

Context:
Library version: Google.Cloud.BigQuery.V2 1.3.0-beta06

We are using central BigQuery tables as a template for creating new tables at runtime. We pull down these template tables and use their schemas to create a new table with a different project, dataset, and/or id. We'd like to extend our logic to copy time partitioning (which is already supported) and clustering.

This request appears to be similar to the feature request for time partitioning https://github.com/googleapis/google-cloud-dotnet/issues/692

jskeet:
Note that although clustering can currently only be specified for
partitioned tables, that may be relaxed - so the client doesn't
attempt to validate it.

Fixes #2823.


jskeet:
Fixing build - not sure how that happened...

Issue #2830, opened 2019-01-25
Suggestion for additional Firestore sentinel attributes


codeaid:
I've been successfully using the Firestore library and so far most of the features have been a pleasure to use. One thing that I feel could be slightly improved is mapping of specialised document fields, such as `Id`, `CreateTime` and `UpdateTime`.

Currently the `FirestoreProperty` attribute allows us to specify which user properties should be stored in the document and which shouldn't, however, there is no way to automatically map the time document was created or its identifier.

The project where this library is used in is a rather standard web API with a plethora of CRUD endpoints. Because it returns models for use in an SPA I need them to include the ID, creation time and last updated time.

All my models implement an `IModel` interface that has the following signature:

```csharp
public interface IModel
{
    string Id { get; set; }
    DateTime? Created { get; set; }
    DateTime? Modified { get; set; }
}
```

I also have a custom `Store<T>` class that manages my Firestore data and has methods like the following ones (amongst others):

```csharp
public abstract class Store<T> : IStore<T> where T : class, IModel
{
    public virtual async Task<T> CreateAsync(T model, string documentId = null) { ... }
    public virtual async Task<T> UpdateAsync(T model) { ... }
    public virtual async Task<T> FindByIdAsync(string id) { ... }
}
```

Because there is no way to directly map specific `DocumentSnapshot` properties to the model, this is what I ended up doing in my `CreateAsync` method:

```csharp
var document = string.IsNullOrWhiteSpace(documentId)
    ? Collection.Document()
    : Collection.Document(documentId);

var writeResult = await document.CreateAsync(model, cancellationToken);
var dateTime = writeResult.UpdateTime.ToDateTime();

// Populate model fields from the newly created document.
model.Id = document.Id;
model.Created = dateTime;
model.Modified = dateTime;

return model;
```

This I suppose is not too bad as it allows me to avoid an extra read which is a priced operation. So instead all my properties on the model are mapped with `FirestoreProperty` and I populate `Id`, `Created` and `Modified` manually. One thing I'd like to have is a `CreateTime` on the `WriteResult` (at least when returned from `CreateAsync`). It doesn't make that much sense syntactically that I get an update time when creating a document. Or, if not, at least rename it to be operation-agnostic (e.g. `WriteTime`).

[...]

PR #2832, opened 2019-01-25
Clarify ServerTimestampAttribute documentation

jskeet:
This is an attempt to make the documentation clearer, as requested in #2830 

cc @codeaid

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ServerTimestampAttribute.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ServerTimestampAttribute.cs
@@ -17,9 +17,9 @@ using System;
 namespace Google.Cloud.Firestore
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


jskeet:
(Will wait for feedback from @codeaid before merging)

codeaid:
Overall I'm happy for this to be merged. One thing I would personally change is _"local"_ to _"user-defined"_ (or something similar). I can see how _"local"_ could be confusing as the sentence does not have a specific context. That's just nitpicking though so good to go as is either way.

jskeet:
"Local" is intended to mean "client-side" - i.e. in .NET rather than at the server. Will see if I can clarify that a bit more - I don't think "user-defined" is right necessarily, as the local value could be the one that the server last returned.

jskeet:
Okay, I've made a change - it doesn't sound terribly elegant, but I think it's at least clear.

codeaid:
Looks great! Good to go. Thanks for that!

PR #2835, opened 2019-01-25
Having the RateLimiter work properly when everything needs to be traced and there are many parallel requests.

### LINKED ISSUE #2834: [Diagnostics] RateLimiter not working properly when almost everything needs to be traced and there are many parallel requests.
I will shortly submit a PR with a fix and a test.

Bug found while solving #2791.

amanda-tarafa:
Fixes #2834 .

The problem is that to determined wheter we needed to trace or not we were relying too much on no other requests coming in at the same time, and this won't work when we either need to trace almost everything and have a lot of parallel requests.

The new test wasn't passing before the fix.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/RateLimiterTest.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Trace/RateLimiterTest.cs
@@ -12,13 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using Moq;
 using System;
-using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using System.Threading;
-using System.Threading.Tasks;
 using Xunit;
 
 namespace Google.Cloud.Diagnostics.Common.Tests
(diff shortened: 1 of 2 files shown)


amanda-tarafa:
Agree that this would be a very contested lock, specially in the scenario that we are trying to fix, i.e. when we receive a lot of requests at "the same time".
Agree as well that for `_fixedDelayMillis == 0` we don't need a check, I'll fix that.
I'm not sure though that reverting to what we had for all other cases is right, I think that for every `_fixedDelayMillis` value there's a high enough amount of requests that we can receive in say `2 * _fixedDelayMillis` millis that would make us skip one of the 2 traces we should be storing in that case. The higher the value of `_fixedDelayMillis` the higher the amount of requests that we need to receive to not sample enough traces, but I think that for instace, `_fixedDelayMillis = 1` will fail more or less the same as `_fixedDelayMillis = 0`.
Let me try and write a test for that.

jskeet:
Sorry, I wasn't clear - I meant that with an optimization to avoid the lock entirely when `fixedDelayMillis` is 0, the rest of the code is fine. But we'll need two tests - one with 1000qps and one with more (i.e. infinite).

amanda-tarafa:
Ah OK, writing the tests then.

amanda-tarafa:
Done, I already had the "infinite" test, added one for exactly 1000Qps and optimized for when we need to trace all. PTAL.

Issue #2791, opened 2019-01-07
Google.Cloud.Diagnostics: ISpan is doing blocking I/O


erdalsivri:
We are using Google.Cloud.Diagnostics.{Common,AspNetCore} packages for our Asp.Net Core web application. Google Cloud tracer seems to be doing a blocking flush to send traces to the server. On production we've identified several hundreds of threads having Google.Cloud.Diagnostics.Common.FlushableConsumerBase in their stack trace. 

We are using CloudTraceMiddleware, which calls `Dispose` on the `span` object. Here is the complete call chain: 

`CloudTraceMiddleware.Invoke` -> `Span.Dispose` -> `SimpleManagerTracer.EndSpan` -> `SimpleManagerTracer.Flush` -> `IConsumer<TraceProto>.Receive`

There is an async version of the `Receive` method but it is not used by the `Span` class. Is there a way to make spans make non-blocking calls?

#### Environment details

  - Windows
  - .Net Core 2.1
  - Google.Cloud.Diagnostics.AspNetCore 3.0.0-beta06

Thanks!

amanda-tarafa:
I'll take a look at this tomorrow.

erdalsivri:
My colleagues pointed out that the library is doing a timed buffering by default. We use the default settings so the interval seems to be 5 seconds. `TimedBufferingConsumer` is starting a timer and calls `Flush` every 5 seconds. `Flush` is acquiring a semaphore and doing a blocking I/O to send the traces and releasing the semaphore. I don't think sending traces collected over a period of 5 seconds will take a lot of time so even if the I/O is blocking, since it is done only in 5 seconds, it doesn't explain the 500 blocked threads we've seen in production. However, since we are using default settings, we are ignoring exceptions so maybe when the tracer for some reason starts receiving errors and long delays, timer will still keep ticking and new threads will block on semaphore . This is just a wild guess though. 

`TimedBufferingConsumer` wraps the `Flush` call inside a try-catch block and ignores all exceptions so difficult to understand if there is in fact a problem while trying to communicate to Google trace servers. I am following feature request #2182.

amanda-tarafa:
Hi @erdalsivri, here are some notes on your questions:

> `CloudTraceMiddleware.Invoke -> Span.Dispose -> SimpleManagedTracer.EndSpan -> SimpleManagerTracer.Flush -> IConsumer<TraceProto>.Receive`
> 
[...]

PR #2836, opened 2019-01-25
Stop holding the semaphore when sending traces to Stackdriver.

amanda-tarafa:
This means that client's requests won't be delayed by flushing traces to the server.

Should fix or drastically improve #2791.

This is not yet ready to merge, integration tests need to be adjusted when #2834 is fixed. But apart from integration tests, the rest shouldn't change so it can be reviewed.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
@@ -16,6 +16,7 @@ using Google.Cloud.ClientTesting;
 using Google.Cloud.Diagnostics.Common;
 using Google.Cloud.Diagnostics.Common.IntegrationTests;
 using Google.Protobuf.WellKnownTypes;
+using Grpc.Core;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Mvc;
@@ -267,7 +268,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
             var traceUri = $"/Trace/{nameof(TraceController.Trace)}/{_testId}";
             var traceStackTraceUri = $"/Trace/{nameof(TraceController.TraceStackTrace)}/{_testId}";
 
-            using (var server = new TestServer(new WebHostBuilder().UseStartup<TraceTestBufferHighQpsApplication>()))
+            using (var server = new TestServer(new WebHostBuilder().UseStartup<TraceTestTinyBufferHighQpsApplication>()))
             using (var client = server.CreateClient())
             {
                 // Make a trace with a small span that will not cause the buffer to flush.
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SizedBufferingConsumer.cs
```diff
@@ -69,61 +64,25 @@ public static SizedBufferingConsumer<T> Create(IConsumer<T> consumer, Func<T, in
         public override void Dispose() => Flush();
 
         /// <inheritdoc />
-        protected override void ReceiveWithSemaphoreHeld(IEnumerable<T> items)
+        protected override bool ReceiveWithSemaphoreHeld(IEnumerable<T> items)
```
moserware: Given the return type change, I wonder if a `Try` prefix might make more sense? Although, it's not necessarily clear what `true` means without looking at the code.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
```diff
@@ -72,44 +68,19 @@ public override void Dispose()
...
-
-        /// <inheritdoc />
-        protected override Task FlushAsyncWithSemaphoreHeldAsync(CancellationToken cancellationToken)
-        {
-            if (_items.Count == 0)
-            {
-                return CommonUtils.CompletedTask;
-            }
-
             IList<T> old = _items;
```
moserware: Note that with an `Interlocked.Exchange`, you might be able to do this without needing the semaphore at all.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -315,6 +315,54 @@ public async Task Trace_IgnoreHealthChecks()
...
+                for (int i = 0; i < requests; i++)
+                {
+                    responseTasks.Add(client.GetAsync(uri));
+                }
+
+                await Task.WhenAll(responseTasks);
+            }
+
+            // TODO: When #2834 is fixed, we should always get as many traces as requests.
+            var traces = _polling.GetTraces(uri, _startTime, minEntries: (9 / 10) * requests);
```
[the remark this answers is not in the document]
amanda-tarafa: Duh! Already fixed and will push shortly. When #2834 is pushed I will remove the range anyways. Thanks!

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -315,6 +315,54 @@ public async Task Trace_IgnoreHealthChecks()
...
+                    responseTasks.Add(client.GetAsync(uri));
+                }
+
+                await Task.WhenAll(responseTasks);
+            }
+
+            // TODO: When #2834 is fixed, we should always get as many traces as requests.
+            // Althouhg in the case of a sized buffer we might be missing the last couple of traces
+            // that din't managed to fill the buffer. But this can be calculated and taken into account exactly.
+            var traces = _polling.GetTraces(uri, _startTime, minEntries: (9 / 10) * requests);
```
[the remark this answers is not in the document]
amanda-tarafa: Same as above.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SizedBufferingConsumer.cs
amanda-tarafa: I agree that after the return type change the method name is not the best so I'm super open to a better name, but I wouldn't use `Try`, in C# using the `Try` prefix in a method returning `bool` is more associated with the returned value indicating whether the operation was succesfull or not, which is not the case, since the receiving will always be succesfull (unless exception) and the returning value in this case indicates whether the caller needs to flush the buffer or not. I'm not super worried about this because this is a `protected` method inside an `internal` class, and the meaning of the returned value is already indicated in docs. That said, if a better name comes up, happy to use it.
Oh, and also, I've been wanting to refactor (basically decoupling the buffering strategy from receiving and flushing) some of this for a while now, it probably won't happen soon, but it'll happen at some point.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
amanda-tarafa: Maybe in the `TimedBufferingConsumer` implementation it's possible, but remember that we are inheriting from `FlushableConsumerBase`, which needs to be general and grab the same semaphore for receiving items and for getting them for flushing. Actually `SizedBufferingConsumer`, also inhereting from `FlushableConsumerBase` does a little bit more in its implementation of this method, that is setting the buffer size to zero.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -315,6 +315,57 @@ public async Task Trace_IgnoreHealthChecks()
...
+            {
+                for (int i = 0; i < requests; i++)
+                {
+                    responseTasks.Add(client.GetAsync(uri));
+                }
+
+                await Task.WhenAll(responseTasks);
+            }
+
+            // TODO: When #2834 is fixed, we should always get as many traces as requests.
```
jskeet: This is now fixed, so presumably the code can be simplified.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -315,6 +315,57 @@ public async Task Trace_IgnoreHealthChecks()
...
```
jskeet: And again

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -315,6 +315,57 @@ public async Task Trace_IgnoreHealthChecks()
...
+            // Althouhg in the case of a sized buffer we might be missing the last couple of traces
```
jskeet: Typos in here to fix if #2834 being fixed doesn't mean we can get rid of the whole comment:

- Althouhg => although
- din't => didn't

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -360,6 +411,18 @@ public class TraceTestNoBufferLowQpsApplication : AbstractTraceTestApplication
...
+    /// <summary>
+    /// A web application to test <see cref="CloudTraceMiddleware"/> and associated classes.
+    /// This app uses a timed buffer and will sample 1,000,000 QPS.
+    /// This will allow all calls to be traced.
+    /// </summary>
+    public class TraceTestTimedBufferHighQpsApplication : AbstractTraceTestApplication
+    {
+        public override BufferOptions GetBufferOptions() => BufferOptions.TimedBuffer(TimeSpan.FromMilliseconds(500));
+
+        public override double GetSampleRate() => 1000000;
```
jskeet: Possibly use 1_000_000 for clarity.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SizedBufferingConsumer.cs
```diff
@@ -69,61 +64,25 @@ public static SizedBufferingConsumer<T> Create(IConsumer<T> consumer, Func<T, in
...
-            foreach (T item in items)
-            {
-                _size += _sizer(item);
-                _items.Add(item);
-                if (_size >= _bufferSize)
-                {
-                    await FlushAsyncWithSemaphoreHeldAsync(cancellationToken).ConfigureAwait(false);
-                }
-            }
+            return _size >= _bufferSize;
```
jskeet: One change here is that if we get a *lot* of items here, it would previously have caused multiple flushes - now it'll only cause one. Is that okay?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SizedBufferingConsumer.cs
```diff
@@ -44,11 +39,11 @@ internal class SizedBufferingConsumer<T> : FlushableConsumerBase<T>
         private int _size;
 
         private SizedBufferingConsumer(IConsumer<T> consumer, Func<T, int> sizer, int bufferSize)
+            :base(consumer)
```
jskeet: Nit: space after the colon

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/SizedBufferingConsumer.cs
amanda-tarafa: Yes, that's OK.
By default we only send one trace at a time to the `SizedBufferingConsumer`, I don't even think client code can configure the tracer so that this is not so. The method receives an `IEnumerable` because we implement the `IConsumer` interface, and some of its other implementors (i.e the consumers that this one flushes to) do receive several traces at the same time.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -358,26 +357,89 @@ public async Task Trace_SizedBuffer_Stress()
...
+            using (var server = new TestServer(new WebHostBuilder().UseStartup<TraceTestTinyBufferHighQpsPropagateApplication>()))
+            using (var client = server.CreateClient())
+            {
+                for (int i = 0; i < requests; i++)
+                {
+                    responseTasks.Add(client.GetAsync(uri));
+                }
+
+                // With such a tiny buffer (it fills with at most two traces), we have a lot of concurrent calls to the
+                // backend trying to send traces. Some of these will timeout, some of these will be succesfull.
```
jskeet: Nit - succesfull => successful

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Trace/TraceTest.cs
```diff
@@ -358,26 +357,89 @@ public async Task Trace_SizedBuffer_Stress()
...
+                // By default we don't propagate the exception but we are doing so for the purpose of this test.
```
jskeet: It's not clear to me where we're setting it up to propagate exceptions. Maybe include that in the comment?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/FlushableConsumerBase.cs
```diff
@@ -150,14 +151,14 @@ private void Flush(IEnumerable<T> items)
         /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
         /// <param name="items">The items to flush.</param>
         /// <returns>A task representing the asynchronous operation.</returns>
-        private Task FlushAsync(IEnumerable<T> items, CancellationToken cancellationToken = default)
+        private async Task FlushAsync(IEnumerable<T> items, CancellationToken cancellationToken = default)
```
jskeet: I don't see why we've got this change. Awaiting a completed task won't do anything - you could just return. So why make the method async?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/FlushableConsumerBase.cs
```diff
@@ -150,14 +151,14 @@ private void Flush(IEnumerable<T> items)
...
         {
             if (items == null || !items.Any())
             {
-                return CommonUtils.CompletedTask;
+                await CommonUtils.CompletedTask.ConfigureAwait(false);
```
jskeet: This is a semantic change - we used to return without having called ReceiveAsync. With this change, you're awaiting a completed task and then calling ReceiveAsync anyway.

amanda-tarafa:
(I'm looking at the tests fails now. I'm almost certain I ran them locally.)

PR #2837, opened 2019-01-28
Prevent logging null log entry labels

henkmollema:
A log entry which contains a label with an empty value results in an `ArgumentException` when serializing it to a protobuf message. This can result in pretty ugly bugs, for example: https://github.com/aspnet/AspNetCore/issues/3460. 

This PR prevents adding en empty user name in `UserLogEntryLabelProvider` and an empty environment name in `EnvironmentNameLogEntryLabelProvider` as well as removing any `null` or empty labels in custom log entry label providers.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
@@ -360,6 +360,26 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
         }
 
         [Fact]
+        public void Log_SkipsNullLabels()
+        {
+            Predicate<IEnumerable<LogEntry>> matcher = logEntries =>
+            {
+                LogEntry entry = logEntries.Single();
+                return !entry.Labels.Any();
+            };
+
+            var mockServiceProvider = new Mock<IServiceProvider>();
+            mockServiceProvider.Setup(sp => sp.GetService(typeof(IEnumerable<ILogEntryLabelProvider>)))
+                .Returns(new ILogEntryLabelProvider[] { new EmptyLogEntryLabelProvider() });
+
+            var mockConsumer = new Mock<IConsumer<LogEntry>>();
+            mockConsumer.Setup(c => c.Receive(Match.Create(matcher)));
+            var logger = GetLogger(mockConsumer.Object, LogLevel.Information, serviceProvider: mockServiceProvider.Object, logName: _baseLogName);
+            logger.LogInformation(_logMessage);
+            mockConsumer.VerifyAll();
+        }
+
+        [Fact]
         public void Log_Labels_DefaultLabelsFirst()
         {
             var labels = new Dictionary<string, string> { { "some-key", "some-value" } };
@@ -406,6 +426,15 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
         public void Invoke(Dictionary<string, string> labels)
         {
             labels["Bar"] = "World";
+        }
+    }
+
+    internal class EmptyLogEntryLabelProvider : ILogEntryLabelProvider
+    {
+        public void Invoke(Dictionary<string, string> labels)
+        {
+            labels["Null"] = null;
+            labels["Empty"] = string.Empty;
         }
     }
 }
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/LabelProviders/UserLogEntryLabelProviderTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/LabelProviders/UserLogEntryLabelProviderTest.cs
@@ -57,31 +57,6 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
         }
 
         [Fact]
-        public void AddsUserNameLabelWithEmptyUserName()
-        {
-            // Arrange
-            var mockHttpContext = new Mock<HttpContext>();
-            mockHttpContext.Setup(x => x.User.Identity.IsAuthenticated).Returns(true);
-
-            var mockHttpContextAccessor = new Mock<IHttpContextAccessor>();
-            mockHttpContextAccessor.Setup(x => x.HttpContext).Returns(mockHttpContext.Object);
-
-            var instance = new UserLogEntryLabelProvider(mockHttpContextAccessor.Object);
-            var labels = new Dictionary<string, string>();
-
-            // Act
-            instance.Invoke(labels);
-
-            // Assert
-            var expected = new Dictionary<string, string>
-            {
-                { "user_authenticated", true.ToString() },
-                { "user_name", null }
-            };
-            Assert.Equal(expected, labels);
-        }
-
-        [Fact]
         public void AddsUserNameLabelWithUserName()
         {
             // Arrange
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
@@ -194,7 +194,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore
                 Severity = logLevel.ToLogSeverity(),
                 Timestamp = Timestamp.FromDateTime(_clock.GetCurrentDateTimeUtc()),
                 JsonPayload = jsonStruct,
-                Labels = { labels },
+                Labels = { labels.Where(kvp => !string.IsNullOrEmpty(kvp.Value)).ToDictionary(kvp => kvp.Key, kvp => kvp.Value) },
                 Trace = GetTraceName() ?? "",
             };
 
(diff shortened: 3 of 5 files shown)


amanda-tarafa:
@henkmollema I'm reviwing, just slower than usual because this week we are in events and conferences.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LabelProviders/EnvironmentNameLogEntryLabelProvider.cs
```diff
@@ -38,7 +38,7 @@ public EnvironmentNameLogEntryLabelProvider(IHostingEnvironment hostingEnvironme
         /// <inheritdoc/>
         public void Invoke(Dictionary<string, string> labels)
         {
-            labels["aspnetcore_environment"] = _hostingEnvironment.EnvironmentName;
+            labels["aspnetcore_environment"] = _hostingEnvironment.EnvironmentName ?? "N/A";
```
amanda-tarafa: If the `EnvironmentName` is `null`, just don't add the label, similar to what you are doing on `UserLogEntryLabelProvider`.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
```diff
@@ -194,7 +194,7 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
                 Severity = logLevel.ToLogSeverity(),
                 Timestamp = Timestamp.FromDateTime(_clock.GetCurrentDateTimeUtc()),
                 JsonPayload = jsonStruct,
-                Labels = { labels },
+                Labels = { labels.Where(kvp => !string.IsNullOrEmpty(kvp.Value)).ToDictionary(kvp => kvp.Key, kvp => kvp.Value) },
```
amanda-tarafa: I don't like silently removing labels that client code have set, even if their value is `null` and sending the label will make the request fail. I rather the request fail and have the client realize that they cannot send a `null` value and fix their code.
With the changes on the providers below, we make sure that none of our providers are sending labels with null values. Any other labels client code sets, either through options or custom providers, they should take care of not sending null values.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/LabelProviders/UserLogEntryLabelProviderTest.cs
```diff
@@ -56,31 +56,6 @@ public void AddsUserAuthenticatedLabel()
...
-            var labels = new Dictionary<string, string>();
-
-            // Act
-            instance.Invoke(labels);
-
-            // Assert
-            var expected = new Dictionary<string, string>
-            {
-                { "user_authenticated", true.ToString() },
-                { "user_name", null }
```
amanda-tarafa: Instead of removing this test entirely, just expect the "user_authenticated" label.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LabelProviders/EnvironmentNameLogEntryLabelProvider.cs
amanda-tarafa: And also, add a test for when the `EnvironmentName` is null to check that there's no label added.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
```diff
@@ -359,6 +359,26 @@ public void Log_Labels()
             mockConsumer.VerifyAll();
         }
 
+        [Fact]
```
amanda-tarafa: This test might not be needed since is better if we don't silently remove labels.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
henkmollema: I agree. However, the point of this PR was to make `GoogleLogger.Log()` never throw as that leads to all sorts of unexpected behavior. Also throwing from loggers is discouraged, see https://github.com/aspnet/AspNetCore/issues/3460#issuecomment-416447852 for reference. I once had an error that the username was not populated correctly so the Google logger failed, which caused my app to completely crash. Took me a while to figure out the root cause. That issue was the reason for me to create this PR.

So I'm not sure what the right approach is here. I agree that failing silently is bad and can be annoying, but the consequences of throwing exceptions might be worse. What do you think about this?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LabelProviders/EnvironmentNameLogEntryLabelProvider.cs
henkmollema: The downside of not adding labels (rather than using a default value) is that the logs without those labels don't show up when you include the label in a query. Logs without the `aspnetcore_environment` label won't show up when querying `labels.aspnetcore_environment!="Production"`. I think that in the case of this particular label (the hosting environment name) it's not that big of an issue as it's almost always said, but I've hit this "issue" a few times with my own labels. Anyway, I will change it.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/LabelProviders/UserLogEntryLabelProviderTest.cs
henkmollema: There is already a `AddsUserAuthenticatedLabel` test which tests that.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
amanda-tarafa: I see this PR as more of, "let's make sure we don't introduce any Exception causing issues" than let's make sure no Exception is thrown ever, there are probably several other ways in which `GoogleLogger.Log()` can throw that have nothing to do with labels. But I do agree with "loggers should not throw" and the fact that when they do it's hard to diagnose. So, what do you think if we do the following:

- Don't remove the null labels.
- Let's wrap the whole of `GoogleLogger.Log()` in a try-catch and honour `_loggerOptions.RetryOptions.ExceptionHandling`, if it's set to `Ignore`, we swallow the Exception else we rethrow? Note that we don't have to honour the other Retry settings because for transient server faults `_consumer.Receive` will already do that and anyway, whatever Exceptions are thrown in `GoogleLogger.Log()` are not really transient so it makes no sense to retry. This approach will yield a different result from yours in that when there are null labels nothing will be logged, but I'm fine with that, I think users will find that more suspicious and can knowingly turn on the propagation of Exceptions to check what's happening.
- Document `ILogEntryLabelProvider` explicitly saying that null keys or values are not allowed.
- Document LoggerOptions.Labels explicitly saying that null keys or values are not allowed. Actually we can probably even check that there are no null labels and throw here since this Exception will be trown at app Startup and not when attempting to log.

What do you think?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LabelProviders/EnvironmentNameLogEntryLabelProvider.cs
henkmollema: Thanks, I didn't know that!

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/LabelProviders/UserLogEntryLabelProviderTest.cs
henkmollema: You're right. I'll revert the delete and adjust the test.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
henkmollema: That sounds good. I will look at implementing this approach.

amanda-tarafa:
@henkmollema Would it be OK if I do the changes I requested earlier? We want to do a GA release soon and want to include these changes there?

henkmollema:
@amanda-tarafa sure, I won't have time for this until this Saturday. If that's too long, please feel free to go ahead. I'm sorry I didn't get to this earlier.

amanda-tarafa:
I'll do it then, and don't apologize please :) your contributions are already more than apreciated. I'll ping you when it's ready so you can take a look. Thanks!

henkmollema:
Thanks Amanda! 😄

amanda-tarafa:
Closing in favour of #2917

PR #2839, opened 2019-01-30
bug fix Timestamp.cs

zxu123:
bug fix


jskeet:
Thanks for the PR - rather than accept this as-is, would you be happy for me to create a new PR which also includes a test that currently fails? Or would you like to add a test to this PR?

PR #2847, opened 2019-02-04
Regenerate all APIs

jskeet:
(Will assign a reviewer once it's green.)


apis/Google.Cloud.Asset.V1Beta1/Google.Cloud.Asset.V1Beta1/AssetService.cs
```diff
@@ -646,9 +651,12 @@ public sealed partial class BatchGetAssetsHistoryRequest : pb::IMessage<BatchGet
     public const int ReadTimeWindowFieldNumber = 4;
     private global::Google.Cloud.Asset.V1Beta1.TimeWindow readTimeWindow_;
     /// <summary>
-    /// Required. The time window for the asset history. The start time is
-    /// required. The returned results contain all temporal assets whose time
-    /// window overlap with read_time_window.
+    /// Optional. The time window for the asset history. Both start_time and
```
amanda-tarafa: Just flagging: I think this is breaking, because now the `start_time` if set needs to be after a given date, as where before it needed to be set but not in a particular range. This will break everyone who has `start_time` set before that given date.

apis/Google.Cloud.Asset.V1Beta1/Google.Cloud.Asset.V1Beta1/Assets.cs
```diff
@@ -270,7 +270,7 @@ public sealed partial class TemporalAsset : pb::IMessage<TemporalAsset> {
   }
 
   /// <summary>
-  /// A time window of [start_time, end_time).
+  /// A time window of (start_time, end_time].
```
amanda-tarafa: This is breaking as well.

jskeet:
The API changes are okay to be breaking, in that both the API and the library are in beta.

Issue #2634, opened 2018-10-24
It's hard for users to find their logs on console.cloud.google.com.


SurferJeffAtGoogle:
After collecting friction logs from many people, I've observed a common issue that users suffer is that it's difficult to find their logs in console.cloud.google.com.  If the app is running locally, then they appear under Global.  If the app is running on App Engine flex, then the logs appear under App Engine.

Admittedly, this client library is far removed from the user experience of console.cloud.google.com.  But I do see an opportunity to help users avoid this pain:  Write a log statement (which appears on stdout when the user runs in development mode) that tells them exactly how to see their logs.  Something like:
```
Logs are being written to Stackdriver and can be seen at https://console.cloud.google.com/logs/viewer?resource=global
```
Like https://github.com/googleapis/google-cloud-dotnet/issues/2630, it will take some cleverness to avoid infinite recursion.

amanda-tarafa:
Shouldn't that information be part of normal user documentation instead? We can describe the different scenarios in which logs end up under one or another "environment label".

SurferJeffAtGoogle:
Yes, it should and it is.  But it's just really hard for users to find the logs.  3 of the 5 of us who recorded friction logs had trouble finding our logs in console.cloud.google.com.

jskeet:
Drive-by thought: suppose our *samples* had something that wrote the URL to the console. If we expose an easy way of getting it printed to the console, but without writing to the console by default, then users following the samples will see it, but then can easily remove it once they know where to look.

Does that work, or do you think people don't follow the samples closely enough?

amanda-tarafa:
Hi @SurferJeffAtGoogle, I'm working on this now and we were thinking of adding some more data to the URL (apart from the monitored resource) so that users can truly land on the logs being logged. This data would include the project ID, the log level, the name of the log, etc. With this info included on the URL users would only need to change on the UI the time spans they want to see logs for. Are you OK with this?

amanda-tarafa:
@SurferJeffAtGoogle I just created #2860. Take a look and leave comments, specially if you think the snippets are enough.

JustinBeckwith:
@jskeet the flagging of this as a P2, even though it was a feature request, was causing it to go out of SLO.  I went ahead and removed that label - just wanted to make sure it's ok with you :)

jskeet:
@JustinBeckwith: Fine by me. I thought everything had to have a priority :)

jskeet:
[...]

PR #2860, opened 2019-02-07
[Diagnostics] - Showing URL where logs will go

amanda-tarafa:
Addresses #2634

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/LoggingSnippets.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/LoggingSnippets.cs
@@ -36,7 +36,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Snippets
     [SnippetOutputCollector]
     public class LoggingSnippetsTests
     {
-        private const string _expectedGcpLogBaseUrl = "https://console.cloud.google.com/logs/viewer";
+        private const string ExpectedGcpLogBaseUrl = "https://console.cloud.google.com/logs/viewer";
         private static readonly LogEntryPolling s_polling = new LogEntryPolling();
 
         private readonly string _testId;
@@ -54,20 +54,29 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Snippets
         [Fact]
         public void GetsLogViewingUrl()
         {
-            // No need to make an actual call to the controller. Just starting up the server
-            // creates several loggers.
-            using (TestServer server = new TestServer(GetUrlWriterHostBuilder())) { }
+            TextWriter oldConsoleOut = Console.Out;
+            string writtenInfo;
+            using (StringWriter writer = new StringWriter())
+            {
+                Console.SetOut(writer);
 
-            string expectedProject = TestEnvironment.GetTestProjectId();
-            string writtenInfo = LoggingTestApplicationWriteUrl.textWriter.ToString();
+                // No need to make an actual call to the controller. Just starting up the server
+                // creates several loggers.
+                using (TestServer server = new TestServer(GetUrlWriterHostBuilder())) { }
+
+                writtenInfo = writer.ToString();
+
+                Console.SetOut(oldConsoleOut);
+            }
 
             int urlStart = writtenInfo.LastIndexOf("http", StringComparison.OrdinalIgnoreCase);
-            Assert.NotInRange(urlStart, -1, -1);
+            Assert.NotEqual(urlStart, -1);
 
             // We know the URL is the last to be written.
             string url = writtenInfo.Substring(urlStart);
+            string expectedProject = TestEnvironment.GetTestProjectId();
 
-            Assert.StartsWith(_expectedGcpLogBaseUrl, url);
+            Assert.StartsWith(ExpectedGcpLogBaseUrl, url);
             Assert.Contains($"resource={MonitoredResourceBuilder.FromPlatform().Type}", url);
             Assert.Contains($"project={expectedProject}", url);
             Assert.Contains("minLogLevel=200", url);
@@ -173,14 +182,14 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Snippets
         {
 #if NETCOREAPP2_0
             string projectId = TestEnvironment.GetTestProjectId();
-            TextWriter textWriter = LoggingTestApplicationWriteUrl.textWriter;
+
             // Sample: RegisterGoogleLoggerWriteUrl2
             return new WebHostBuilder()
                 .ConfigureServices(services =>
                 {
                     // Once created, the GoogleLogger will write the URL where logs can be found
-                    // to the given textWriter.
-                    LoggerOptions loggerOptions = LoggerOptions.Create(loggerDiagnosticsOutput: textWriter);
+                    // to a given System.IO.TextWriter, for instance System.Console.Out.
+                    LoggerOptions loggerOptions = LoggerOptions.Create(loggerDiagnosticsOutput: Console.Out);
 
                     // Replace projectId with your Google Cloud Project ID.
                     services.AddSingleton<ILoggerProvider>(sp => GoogleLoggerProvider.Create(sp, projectId, loggerOptions));
@@ -272,7 +281,6 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Snippets
     internal class LoggingTestApplicationWriteUrl
     {
         private static readonly string ProjectId = TestEnvironment.GetTestProjectId();
-        internal static readonly TextWriter textWriter = new StringWriter();
 
         // Sample: RegisterGoogleLoggerWriteUrl
         public void ConfigureServices(IServiceCollection services)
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/LoggingSnippets.cs
```diff
@@ -34,6 +36,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Snippets
     [SnippetOutputCollector]
     public class LoggingSnippetsTests
     {
+        private const string _expectedGcpLogBaseUrl = "https://console.cloud.google.com/logs/viewer";
```
jskeet: As this is a const, I'd just use `ExpectedGcpLogBaseUrl` as the name.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/LoggingSnippets.cs
```diff
@@ -211,6 +264,43 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
...
 
+    /// <summary>
+    /// A simple web application to test the <see cref="GoogleLogger"/>
+    /// and associated classes. Configure <see cref="GoogleLogger"/> to
+    /// write the URL where logs can be found.
+    /// </summary>
+    internal class LoggingTestApplicationWriteUrl
+    {
+        private static readonly string ProjectId = TestEnvironment.GetTestProjectId();
+        internal static readonly TextWriter textWriter = new StringWriter();
```
jskeet: Could we make this a property (perhaps just `Writer`?) rather than a field? I know it's only for tests, but non-private fields simply make me twitchy :)

jskeet: Secondly, this being static makes me nervous. I suspect if we have multiple tests, they could give false positives. (I don't know enough about ASP.NET Core infrastructure to see immediately how to avoid that, but I suspect there's a way.)

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
```diff
@@ -34,10 +35,11 @@ public class GoogleLoggerTest
         private const string _logName = "log-name";
         private const string _baseLogName = "aspnetcore";
         private const string _projectId = "pid";
+        private const string _expectedGcpLogBaseUrl = "https://console.cloud.google.com/logs/viewer";
```
jskeet: Same change as before - or could we just make the one in GoogleLogger internal?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Logging/GoogleLoggerTest.cs
```diff
@@ -46,12 +48,64 @@ private string Formatter(string message, Exception ex)
...
+        {
+            consumer = consumer ?? new Mock<IConsumer<LogEntry>>(MockBehavior.Strict).Object;
+            monitoredResource = monitoredResource ?? MonitoredResourceBuilder.GlobalResource;
+            logTarget = logTarget ?? s_defaultLogTarget;
+            LoggerOptions options = LoggerOptions.Create(logLevel, logName, labels, monitoredResource);
+            return new GoogleLogger(consumer, logTarget, options, _logName, s_clock, serviceProvider);
+        }
+
+        [Fact]
+        public void GetsGcpConsoleLogsUrl()
```
jskeet: I think the "s" in Gets should be removed for all of these tests

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
```diff
@@ -226,5 +235,39 @@ internal string GetTraceName()
...
+        /// this <see cref="GoogleLogger"/> can be seen on the GCP Stackdriver Logging Console.
+        /// </summary>
+        public Uri GetGcpConsoleLogsUrl()
+        {
+            IList<string> parameters = new List<string>
+            {
+                $"resource={_loggerOptions.MonitoredResource.Type}",
+                $"minLogLevel={(int)_loggerOptions.LogLevel.ToLogSeverity()}",
+                $"logName={_fullLogName}",
+                _logTarget.Kind == LogTargetKind.Project ? $"project={_logTarget.ProjectId}" :
```
jskeet: I think this would be easier to format outside the collection initializer:

```text
string target =
    _logTarget.Kind == LogTargetKind.Project ? $"project={_logTarget.ProjectId}" :
    _logTarget.Kind == LogTargetKind.Organization ? $"organizationId={_logTarget.OrganizationId}" :
    throw new InvalidOperationException($"Unrecognized LogTargetKind: {_logTarget.Kind}");
```

Then use `target` in `parameters`

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
```diff
@@ -226,5 +235,39 @@ internal string GetTraceName()
...
+                Query = string.Join("&", parameters)
+            }.Uri;
+        }
+
+        private void WriteCreationDiagnostics()
+        {
+            // Explicitly not catching exceptions.
+            // This should only be activated for diagnostics purposes so in that case
+            // we shouldn't throw exceptions.
+            if (_loggerOptions.LoggerDiagnosticsOutput != null)
```
jskeet: Just use the null conditional operator?

```csharp
_loggerOptions.LoggerDiagnosticsOutput?.WriteLine($"{DateTime.UtcNow} - GoogleLogger created. Logs written at: {GetGcpConsoleLogsUrl()}");
_loggerOptions.LoggerDiagnosticsOutput?.Flush();
```

I'd be tempted to explicitly use string.Format and use the invariant culture as well, so that the date always gets formatted the same way. Ideally, do that with `yyyy-MM-dd'T'HH:mm:ss` as the format, too...

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
```diff
@@ -74,20 +85,24 @@ public sealed class LoggerOptions
...
+        /// the logs written with these options can be found.</param>
         public static LoggerOptions Create(
             LogLevel logLevel = LogLevel.Information,
             string logName = null,
             Dictionary<string, string> labels = null,
             MonitoredResource monitoredResource = null,
             BufferOptions bufferOptions = null,
-            RetryOptions retryOptions = null)
+            RetryOptions retryOptions = null,
+            TextWriter loggerDiagnosticsOutput = null)
```
jskeet: This is a binary breaking change - am I right in saying we're okay with that at the moment, as we're still pre-3.0-GA?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Snippets/LoggingSnippets.cs
```diff
@@ -211,6 +264,43 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
...
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddMvc();
+        }
+
+        public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
+        {
+            // Once created, the GoogleLogger will write the URL where logs can be found to the given
+            // textWriter.
+            LoggerOptions loggerOptions = LoggerOptions.Create(loggerDiagnosticsOutput: textWriter);
```
jskeet: If this is a snippet, I'd probably just use `Console.Out` here instead. In many cases that will be absolutely fine. It's harder to test, but I'd be tempted to ignore that for now.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
amanda-tarafa: Yes, this is all pre 3.0 GA

amanda-tarafa:
Commit pushed, PTAL @jskeet @SurferJeffAtGoogle , thanks!

amanda-tarafa:
Thanks. Will rebase and give @SurferJeffAtGoogle a little bit more time before merging.

SurferJeffAtGoogle:
No matter where/how we add it to documentation and samples, I think very few people will find the additional setting:
```
            // Once created, the GoogleLogger will write the URL where logs can be found
            // to a given System.IO.TextWriter, for instance System.Console.Out.
            LoggerOptions loggerOptions = LoggerOptions.Create(loggerDiagnosticsOutput: Console.Out);
```
and set it accordingly.  By default, can we set it to true when ASPNETCORE_ENVIRONMENT is set to Development?

jskeet:
I don't *think* we know whether it's a development environment at the point where we're building the host. That's detected later on. While we *could* test for the explicit environment variable, I wouldn't be surprised if there were other ways of setting it to development mode.

I would expect that users will need to read *some* documentation in order to get started at all - typically ASP.NET Core doesn't seem to be inherently discoverable - so I'd have thought it would be a matter of making sure that we include this in *every* example we provide. I'm not sure there's a better way, even though I agree with the concern :(

jskeet:
Mind you, the [documentation for environments](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/environments?view=aspnetcore-2.2) does specifically talk about the environment variable.
It feels pretty fragile to use this, but it's possible better than nothing.

Could I suggest that I merge this PR, then add the environment detection in a later one?

jskeet:
The tests have started failing after rebasing. I'll need to look more closely into what's going on here.

jskeet:
I'm going to merge this now, then do another NuGet beta release, while we consider how to make this as discoverable as possible. I've been reflecting on this further after working on personal projects - it can end up being complicated as you may not want it to log to Stackdriver *at all* in Development environment...

PR #2862, opened 2019-02-08
Smoke test and snippets for Google.Cloud.Talent.V4Beta1

jskeet:


jskeet:
I generally prefer to test against the oldest framework we support, as it's more likely that we'll break compatibility with that, than that Microsoft will break compatibility with us (modulo recent bits about .NET Core 3.0).
That said, a new discussion about what to support wouldn't go amiss - in particular, I'd be happy for all our tests to support netcoreapp2.1 as that gives better stack traces for async code.

PR #2866, opened 2019-02-11
Project changes

jskeet:
The first, second and fourth commits are artisanal commits, lovingly hand-crafted from only the finest naturally-occurring and sustainably-harvested 1s and 0s.

The third commit is autogenerated.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -88,8 +88,8 @@
     "description": "Common code used by Bigtable V2 APIs",
     "tags": [ "Bigtable" ],
     "dependencies": {
-      "Google.Api.Gax": "2.6.0-beta02",
-      "Google.Protobuf": "3.6.0"
+      "Google.Api.Gax": "default",
+      "Google.Protobuf": "3.6.1"
     }
   },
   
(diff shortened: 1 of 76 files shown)


apis/apis.json
```diff
@@ -446,13 +428,13 @@
     "tags": [ "NLog", "Logging", "Stackdriver" ],
     "dependencies": {
       "NLog": "4.5.0",
-      "Google.Api.Gax.Grpc": "2.5.0",
+      "Google.Api.Gax.Grpc": "default",
```
chrisdunelm: I'm unclear on why this NLog package specifies The "Google.Api.Gax.Grpc" (as default), but the "Google.Cloud.Monitoring.V3" (for example) now doesn't specify a "Google.Api.Gax.Grpc" dependency at all?

They are both beta packages, so I'd expect them to do the same thing?

jskeet: The NLog package is of type "other", so it doesn't get any packages by default. Any "type=grpc" packages get GAX gRPC automatically.

chrisdunelm: OK, yes, that makes sense :)

Issue #2872, opened 2019-02-13
Revisit SDK dependencies


jskeet:
Currently our SDK dependencies make it slightly hard for other folks to run our build and test:

- We use netcoreapp1.0 for tests, which means installing the .NET Core 1.0 runtime
- We have 2.1.400 in global.json, which is quite a long way behind the curve, and is relatively hard to install on Linux (2.1.500 is the latest in 2.1, and the rest of the world is on 2.0)

We should at least *consider* moving to 2.2.100 in global.json, and possibly testing against netcoreapp2.1 as that gets nicer async stack traces as well. Whether we drop the netcoreapp1.0 tests or not is a tricky decision - we don't want our tests to take much longer, but we also don't want to inadvertently break support.

jskeet:
This is now fixed, between #2874 and #2875.

PR #2874, opened 2019-02-13
Update .NET SDK requirement to 2.2.100

jskeet:
This is the latest SDK minor version, which makes it easier for users who want to just "clone, install latest, build" to do so.

(If we want to stay on 2.1.x, we'd want to move to 2.1.500 instead, as that's the latest 2.1, and what is install on Linux by just following the regular instructions.)

This PR is part of #2872.


.travis.yml
```diff
@@ -5,14 +5,14 @@ env:
...
-dist: trusty
+dotnet: 2.2.104
+dist: xenial
 
 # Install the .NET Core 1.0 runtime as that's what we test against
 addons:
   apt:
     sources:
-    - sourceline: 'deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-ubuntu-trusty-prod trusty main'
+    - sourceline: 'deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-ubuntu-xenial-prod trusty main'
```
amanda-tarafa: question: there's still a "trusty" at the end of the new line, before main, not sure if that is correct.

jskeet: I'm surprised that even works! Will edit and see if it still does afterwards...

jskeet:
Basically 2.2.100 is sort of the minimum "SDK API" version we support. Any 2.2.1xx is compatible with that, so I figured I'd use the latest one I know about (which includes security patches) for Travis.

PR #2885, opened 2019-02-19
Add DLP v2 snippet

jskeet:
This is based on the v2beta1 snippet we used to have, but the API changed significantly.


apis/Google.Cloud.Dlp.V2/Google.Cloud.Dlp.V2.Snippets/DlpServiceClientSnippets.cs
```diff
@@ -0,0 +1,66 @@
+// Copyright 2017, Google Inc. All rights reserved.
```
amanda-tarafa: nit: 2019

jskeet: Whoops, thanks - will fix.

PR #2891, opened 2019-02-20
A simple tool to show the release status of our packages

jskeet:
This helps to highlight packages that we should consider moving to GA.

We can make it more subtle later on, of course.

### DIFF
--- a/tools/Google.Cloud.Tools.ShowReleaseStatus/Program.cs
+++ b/tools/Google.Cloud.Tools.ShowReleaseStatus/Program.cs
@@ -1,0 +1,110 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Tools.Common;
+using LibGit2Sharp;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Google.Cloud.Tools.ShowReleaseStatus
+{
+    /// <summary>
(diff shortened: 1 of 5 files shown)


jskeet:
Ping for cursory review?

tools/Google.Cloud.Tools.ShowReleaseStatus/Program.cs
```diff
@@ -0,0 +1,110 @@
...
+                Console.WriteLine("No releases");
+            }
+            else
+            {
+                // Only display a prerelease if it's more recent than the last GA
+                if (lastPrerelease != null && (lastGa is null || lastPrerelease.Timestamp > lastGa.Timestamp))
+                {
+                    Console.WriteLine($"Last prerelease: {lastPrerelease}");
+                }
+                if (lastGa != null)
```
amanda-tarafa: Nit: I think an else would do the job and is clearer

tools/Google.Cloud.Tools.ShowReleaseStatus/Program.cs
```diff
@@ -0,0 +1,110 @@
...
+            var lastGa = releases.FirstOrDefault(release => release.IsGa);
+            var lastPrerelease = releases.FirstOrDefault(release => !release.IsGa);
+
+            if (attentionOnly && !NeedsAttention(api, lastPrerelease, lastGa))
+            {
+                return;
+            }
+
+            Console.WriteLine(api.Id);
+            if (lastGa is null && lastPrerelease is null)
```
amanda-tarafa: Can this ever happen? I mean would we add a library in the catalog and not release or prerelease it for any reason?

jskeet: Yup - for example, Google.Cloud.IncidentResourceManagement.V1Alpha2 hasn't been released yet.

tools/Google.Cloud.Tools.ShowReleaseStatus/Program.cs
jskeet: No, these aren't mutually exclusive. We have three different kinds of output here:

- Just a prerelease
- Just a GA (either because there were no prereleases, or there haven't been any since the GA)
- Both prerelease and GA

jskeet:
Sorry that it wasn't assigned - I *thought* I'd assigned it to both you and Chris, and that maybe you were each waiting for the other :) Will merge now.

PR #2904, opened 2019-02-26
Change to logging tests to defer all validation until the end

jskeet:
This has two big advantages:

- Shorter running time
- Less flakiness in terms of quota

There are disadvantages too, in terms of readability if the tests
fail:

- If any test fails, all tests fail as the validation is done in Dispose
- With N tests and F failures, we end up with N*F stack traces due
  to the way the failures are reported

I think we can reasonably assume that most of the time, everything's
going to pass - we can take the additional diagnostic pain in the
rare failure case for the benefit in the success case.


jskeet:
Currently running all integration tests on all platforms - previously I've just rerun LoggingTest.

PR #2912, opened 2019-03-01
Minor bugfix to ClientTesting GrpcInfo

chrisdunelm:


chrisdunelm:
Unfortunately this doesn't affect the pubsub issue, but should probably be fixed anyway!

PR #2918, opened 2019-03-06
Spanner Batch DML

jskeet:
(From private repo.)

Running tests now against public endpoint.


jskeet:
(Note that this is *without* having regenerated the API source. That can come in a separate PR.)

jskeet:
(Removed Amanda due to leave - whoops!) Chris, this is really just in need of a very quick skim. I reviewed the code originally.

PR #2922, opened 2019-03-06
Regenerate PubSub

jskeet:
This was all one regeneration, but I've separated it into three commits. The first two are safe, I believe. The last one needs much more careful checking - we may need a partial class to avoid removing previously-published members.


chrisdunelm:
The second commit changes retry parameters, so is theoretically breaking. I'm not sure how much this will affect users in practice.

chrisdunelm:
Commits 1 and 3 don't appear to be breaking?

jskeet:
Whoops, thought I'd reordered the second and third. Let's chat about the retry one tomorrow.

jskeet:
Would it be worth me removing commit 2 and then merging, so we can discuss the retry changes on their own in a second PR?

chrisdunelm:
Yes, that sounds like a good plan.

jskeet:
Removed retry changes - please approve and I'll merge on green.

PR #2923, opened 2019-03-07
Regenerate PubSub: retry changes

jskeet:
I *think* these are breaking, so we need to work out what to do about them...

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/ObsoleteRetryFilters.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/ObsoleteRetryFilters.cs
@@ -1,0 +1,49 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax.Grpc;
+using Grpc.Core;
+using System;
+
+namespace Google.Cloud.PubSub.V1
+{
+    // This file contains retry filters which have been removed from the GAPIC configuration,
+    // but still need to be present for backward compatibility purposes.
+
+    public partial class PublisherServiceApiSettings
+    {
+        /// <summary>
+        /// Legacy property for a retry filter for status codes of Aborted,
+        /// Cancelled, DeadlineExceeded, Internal, ResourceExhausted, Unavailable and Unknown.
+        /// This property is no longer used in any default settings, and is only present
+        /// for backward compatibility purposes.
+        /// </summary>
+        [Obsolete("This property is no longer used in any settings")]
+        public static Predicate<RpcException> OnePlusDeliveryRetryFilter { get; } =
+            RetrySettings.FilterForStatusCodes(StatusCode.Aborted, StatusCode.Cancelled, StatusCode.DeadlineExceeded, StatusCode.Internal, StatusCode.ResourceExhausted, StatusCode.Unavailable, StatusCode.Unknown);
+    }
+
+    public partial class SubscriberServiceApiSettings
+    {
+        /// <summary>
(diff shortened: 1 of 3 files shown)


jskeet:
(I'm expecting that we'll need to talk about this in a VC.)

chrisdunelm:
Looks like Publisher has renamed `OnePlusDeliveryRetryFilter` to `PublishRetryFilter`.
And Subscriber has removed `PullRetryFilter`, and created `NonIdempotentRetryFilter`.
There are various changes to the status-codes in the filters; and some methods have changed filters.

jskeet:
I've added a partial class for compatibility. PTAL.

lvernaillen:
@jskeet I guess this will change synchronous pull behavior when returnImmediately is set to false.
We use this to do a throttling of our own. When there are no messages available the call waits (blocks) until new messages are available (or total expiration is hit).

The default pull settings have a timeout backoff of 12s and a total expiration of 10 minutes.
Before this change those default pull settings had a PullRetryFilter which retries on StatusCode.DeadlineExceeded (among others).

Assume there are no messages available.

Old behavior:

no message available so after 12s a "DeadLine Exceeded" RpcException occurs.
the pull is retried behind the scenes (with some jittered RetryBackoff delay in between) so the call site does not see this "DeadLine Exceeded" RpcException (but it can be seen in GRPC tracing)! It is simply still waiting for new messages.
after some retries the total expiration of 10 minutes is hit.
the pull call now fails with "DeadLine Exceeded" RpcException that is passed to the call site.
New behavior:

no message available so after 12s a "DeadLine Exceeded" RpcException occurs.
the pull is not retried and immediately fails with "DeadLine Exceeded" RpcException that is passed to the call site.
So where before there was a 10 minutes time frame (total expiration) where the pull call was waiting for new messages, it is now down to 12 seconds.

Is this the desired behavior?
What is the reasoning behind the removal of Deadline Exceeded as a retryable status when pulling?

It seems a deliberate decision when I notice it in the [Regenerate all clients, without retrying on DeadlineExceeded](https://github.com/googleapis/google-cloud-dotnet/commit/877f04bb6a47d51399d2e4945e60f05a2b9097d0
) commit as well.

jskeet:
@chrisdunelm could you assess @lvernaillen's comment? I don't know enough the PubSub part specifically.

In terms of not retrying `DeadlineExceeded`, I wasn't present in the meeting where it was discussed, but I *believe* the expectation is that if a deadline has been exceed, often retrying is unlikely to work but will create additional pressure on the service making it even less likely that other traffic will succeed. I can ask other colleagues for more detail about the decision if you want.

jskeet:
Note that the global removal of deadline exceeded has now been reverted, as clients in some languages had started seeing more errors. I wouldn't like to speculate on the long-term situation.

chrisdunelm:
Looks like this is a moot point now it's being reverted.
I also wasn't present during the deadline_exceeded discussions, so can't add any insight here.

But my understanding is that @lvernaillen's assessment is correct. A longer deadline can be specified in the calling code to increase it from 12 seconds.

lvernaillen:
The revert of the global removal of DeadlineExceeded does not add DeadlineExceeded back to the retry filter that is used by default in the Subscriber's PullSettings. Since that was removed in another commit (the one from this pull request).

We can off course always use custom PullSettings.
But why is there such a difference between Publisher and Subscriber?

**Publisher's PublishSettings:**
- `PublishRetryFilter` (not made obsolete)
- Retries on `Aborted`, `Cancelled`, `DeadlineExceeded`, `Internal`, `ResourceExhausted`, `Unavailable`, `Unknown`

**Subscriber's PullSettings:**
- `PullRetryFilter `made obsolete (used to retry on `DeadlineExceeded`, `Internal`, `ResourceExhausted`, `Unavailable`)
- Replaced with `IdempotentRetryFilter`
- Now only retries on `Aborted`, `Unavailable`, `Unknown`

**Question**
- When `DeadlineExceeded` is globally added back to `IdempotentRetryFilter` (the revert), shouldn't it be added to the one in Subscriber as well?
- When `DeadlineExceeded` is globally added back to `IdempotentRetryFilter` (the revert), in the Publisher it is added back to `PublishRetryFilter`. Shouldn't it be added to it's `IdempotentRetryFilter` as well?
- When `PullRetryFilter` is made obsolete, shouldn't `PublishRetryFilter` be made obsolete as well and use  `IdempotentRetryFilter` instead?


On a side note, the Subscriber's PullSettings with an `IdempotentRetryFilter` includes status code `Unknown`. That was not included in the old `PullRetryFilter`. This seems good as we get `Grpc.Core.RpcException: Status(StatusCode=Unknown, Detail="Stream removed")` from time to time from which we can recover by retrying the call ourselves.

PR #2927, opened 2019-03-07
Add snippets for Cloud Scheduler

jskeet:
These also act as a smoke test. They require a new environment variable, TEST_PROJECT_LOCATION, to be set.

### DIFF
--- a/apis/Google.Cloud.Scheduler.V1/Google.Cloud.Scheduler.V1.Snippets/CloudSchedulerClientSnippets.cs
+++ b/apis/Google.Cloud.Scheduler.V1/Google.Cloud.Scheduler.V1.Snippets/CloudSchedulerClientSnippets.cs
@@ -1,0 +1,70 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Scheduler.V1/Google.Cloud.Scheduler.V1.Snippets/CloudSchedulerClientSnippets.cs
```diff
@@ -0,0 +1,70 @@
...
+                {
+                    HttpMethod = HttpMethod.Get,
+                    Uri = "https://www.google.com"
+                }
+            };
+            Job createdJob = client.CreateJob(parent, job);
+            Console.WriteLine($"Created job details: {createdJob}");
+            // End sample
+
+            client.DeleteJob(createdJob.JobName);
```
chrisdunelm: Not in a `finally` - I guess it doesn't really matter.

jskeet: That would only be relevant if the Console.WriteLine failed... any earlier failure, and we wouldn't have a job to delete!

PR #2928, opened 2019-03-07
[DO NOT MERGE] Very first micro-generator run

chrisdunelm:


chrisdunelm:
This is from the very first run of the micro-generator, so there are many things to check and improve.

PR #2934, opened 2019-03-11
Initial version of a version-compatibility tool

chrisdunelm:
Incomplete, but just enough to start being useful.

### DIFF
--- a/tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
+++ b/tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
@@ -1,0 +1,136 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace Class
+{
+    namespace NoChange.A { public class C { public C() { } } }
+    namespace NoChange.B { public class C { public C() { } } }
+
+    namespace ToStatic.A { public class C { public C() { } } }
+    namespace ToStatic.B { public static class C { } }
+
+    namespace ToStaticNoCtor.A { public class C { private C() { } } }
+    namespace ToStaticNoCtor.B { public static class C { } }
+
+    namespace ToAbstract.A { public class C { public C() { } } }
+    namespace ToAbstract.B { public abstract class C { public C() { } } }
+
+    namespace ToAbstractNoCtor.A { public class C { protected C() { } } }
+    namespace ToAbstractNoCtor.B { public abstract class C { protected C() { } } }
+
+    namespace ToSealed.A { public class C { public C() { } } }
+    namespace ToSealed.B { public sealed class C { public C() { } } }
+
+    namespace FromStatic.A { public static class C { } }
+    namespace FromStatic.B { public class C { protected C() { } } }
+
+    namespace FromStaticNoCtor.A { public static class C { } }
+    namespace FromStaticNoCtor.B { public class C { private C() { } } }
+
+    namespace FromAbstract.A { public abstract class C { public C() { } } }
+    namespace FromAbstract.B { public class C { public C() { } } }
+
+    namespace FromAbstractNoCtor.A { public abstract class C { private C() { } } }
+    namespace FromAbstractNoCtor.B { public class C { private C() { } } }
+
+    namespace FromSealed.A { public sealed class C { public C() { } } }
+    namespace FromSealed.B { public class C { public C() { } } }
+
+    namespace FromSealedNoCtor.A { public sealed class C { private C() { } } }
+    namespace FromSealedNoCtor.B { public class C { private C() { } } }
+
+    namespace BaseTypeFromObjectOk.A { public class C { } }
+    namespace BaseTypeFromObjectOk.B { public class C : List<int> { } }
+
+    namespace BaseTypeFromObjectAbstract.A { public abstract class C { } }
+    namespace BaseTypeFromObjectAbstract.B { public abstract class C : List<int> { } }
+
+    namespace BaseTypeChanged.A { public class C : List<string> { } }
+    namespace BaseTypeChanged.B { public class C : List<int> { } }
+
+    public interface I1 { }
+    internal interface I2 { }
+
+    namespace InterfaceAdded.A { public abstract class C { } }
+    namespace InterfaceAdded.B { public abstract class C : I1 { } }
+
+    namespace InterfaceRemoved.A { public abstract class C : I1 { } }
+    namespace InterfaceRemoved.B { public abstract class C { } }
+
+    namespace NonExportedInterfaceAdded.A { public abstract class C { } }
+    namespace NonExportedInterfaceAdded.B { public abstract class C : I2 { } }
+
+    namespace NonExportedInterfaceRemoved.A { public abstract class C : I2 { } }
+    namespace NonExportedInterfaceRemoved.B { public abstract class C { } }
+
+    namespace GenericConstraintAdded1.A { public class C<A> { } }
+    namespace GenericConstraintAdded1.B { public class C<A> where A : new() { } }
+
+    namespace GenericConstraintAdded2.A { public class C<A> { } }
+    namespace GenericConstraintAdded2.B { public class C<A> where A : class { } }
+
+    namespace GenericConstraintAdded3.A { public class C<A> { } }
+    namespace GenericConstraintAdded3.B { public class C<A> where A : struct { } }
+
+    namespace GenericConstraintAdded4.A { public class C<A> { } }
+    namespace GenericConstraintAdded4.B { public class C<A> where A : IList<A> { } }
+
+    namespace GenericConstraintRemoved1.A { public class C<A> where A : new() { } }
+    namespace GenericConstraintRemoved1.B { public class C<A> { } }
+
+    namespace GenericConstraintRemoved2.A { public class C<A> where A : class { } }
+    namespace GenericConstraintRemoved2.B { public class C<A> { } }
+
+    namespace GenericConstraintRemoved3.A { public class C<A> where A : struct { } }
+    namespace GenericConstraintRemoved3.B { public class C<A> { } }
+
+    namespace GenericConstraintRemoved4.A { public class C<A> where A : IList<A> { } }
+    namespace GenericConstraintRemoved4.B { public class C<A> { } }
+}
+
+namespace Google.Cloud.Tools.VersionCompat.Tests
+{
+    public class ClassTest : TestBase
+    {
+        [Fact] public void NoChange() => TestNone();
+        // TOOD: Enable this test after ctors are dealt with properly.
+        // [Fact] public void ToStatic() => TestMajor(Cause.ClassModifierChanged);
+        [Fact] public void ToStaticNoCtor() => TestNone();
+        [Fact] public void ToAbstract() => TestMajor(Cause.ClassModifierChanged);
+        [Fact] public void ToAbstractNoCtor() => TestNone();
+        [Fact] public void ToSealed() => TestMajor(Cause.ClassModifierChanged);
+        // TOOD: Enable this test after ctors are dealt with properly.
+        // [Fact] public void FromStatic() => TestMinor(Cause.ClassModifierChanged);
+        [Fact] public void FromStaticNoCtor() => TestNone();
+        [Fact] public void FromAbstract() => TestMinor(Cause.ClassModifierChanged);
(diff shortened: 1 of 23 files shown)


jskeet:
Will try to have a look tomorrow, but I'm in the office so will be travelling half the day...

chrisdunelm:
No rush :)

On a quick test-run, it agreed that the Google.Cloud.Vision.V1 update from v1.2.0 to v1.3.0 was a minor update :)

tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
```diff
@@ -0,0 +1,136 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Collections.Generic;
+using Xunit;
+
+namespace Class
+{
+    namespace NoChange.A { public class C { public C() { } } }
```
jskeet: Why all the parameterless constructors? Can't we just let the compiler generate them?

tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
```diff
@@ -0,0 +1,136 @@
...
+    namespace GenericConstraintRemoved4.A { public class C<A> where A : IList<A> { } }
+    namespace GenericConstraintRemoved4.B { public class C<A> { } }
+}
+
+namespace Google.Cloud.Tools.VersionCompat.Tests
+{
+    public class ClassTest : TestBase
+    {
+        [Fact] public void NoChange() => TestNone();
+        // TOOD: Enable this test after ctors are dealt with properly.
```
jskeet: Nit: TOOD => TODO, or better, use `[Fact(Skip = "...")]` instead

tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
```diff
@@ -0,0 +1,136 @@
...
+        // [Fact] public void ToStatic() => TestMajor(Cause.ClassModifierChanged);
+        [Fact] public void ToStaticNoCtor() => TestNone();
```
jskeet: No, changing a class from non-static to static is a breaking change. It can't be used as a type argument or a variable type.

tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
```diff
@@ -0,0 +1,136 @@
...
+        [Fact] public void ToAbstractNoCtor() => TestNone();
+        [Fact] public void ToSealed() => TestMajor(Cause.ClassModifierChanged);
+        // TOOD: Enable this test after ctors are dealt with properly.
+        // [Fact] public void FromStatic() => TestMinor(Cause.ClassModifierChanged);
+        [Fact] public void FromStaticNoCtor() => TestNone();
+        [Fact] public void FromAbstract() => TestMinor(Cause.ClassModifierChanged);
+        [Fact] public void FromAbstractNoCtor() => TestNone();
+        [Fact] public void FromSealed() => TestMinor(Cause.ClassModifierChanged);
+        [Fact] public void FromSealedNoCtor() => TestNone();
+        [Fact] public void BaseTypeFromObjectOk() => TestMinor(Cause.ClassBaseClassChanged);
```
jskeet: I think we'll need documentation justifying a fair number of these. That can come later, of course.

tools/Google.Cloud.Tools.VersionCompat.Tests/TestBase.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            var s = new MemoryStream(s_dllBytes);
+            // Must be read twice, as each assembly definition will be independently mutated during the test.
+            s.Position = 0;
+            var older = AssemblyDefinition.ReadAssembly(s);
+            s.Position = 0;
+            var newer = AssemblyDefinition.ReadAssembly(s);
+            // Use this convention for test namespaces.
+            var typeName = GetType().Name;
+            var testNamespace = $"{typeName.Substring(0, typeName.Length - 4)}.{callerMemberName}";
```
jskeet: Why -4? A comment would help here.

tools/Google.Cloud.Tools.VersionCompat.Tests/TopLevelTest.cs
```diff
@@ -0,0 +1,77 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Xunit;
+
+namespace TopLevel
+{
+    namespace ClassUnchanged.A { public class C { } }
```
jskeet: It's a pity we can't add attributes to express the expected result inline. I'll think about whether there's any nice way we could do this - unless there's some more fundamental issue that I'm missing that would stop us from doing it.

tools/Google.Cloud.Tools.VersionCompat/Cause.cs
```diff
@@ -0,0 +1,50 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Google.Cloud.Tools.VersionCompat
+{
+    public enum Cause
```
jskeet: I think we may end up wanting more fine-grained causes, or possibly hierarchical, if we're going to allow configuration to drive this. Obviously that's a longer-term issue though.

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/SameTypeComparer.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                return false;
+            }
+            if (x.FullName != y.FullName)
+            {
+                return false;
+            }
+            // If names are the same, then they will have the same number of generic parameters
+            // because the generic parameter count is encoded in the name. E.g. List`1 for a List<AnyType>
+            for (int i =0; i<x.GenericParameters.Count;i++)
```
jskeet: Nit: spaces before/after operators etc

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/SamePropertyComparer.cs
```diff
@@ -0,0 +1,54 @@
...
+            }
+            if (x.Name != y.Name)
+            {
+                return false;
+            }
+            if (x.Parameters.Count != y.Parameters.Count)
+            {
+                return false;
+            }
+            foreach (var (xParam, yParam) in x.Parameters.Zip(y.Parameters))
```
jskeet: I believe this will consider `out`, `ref` and `in` parameters as equal, which possibly isn't ideal.

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/SamePropertyComparer.cs
```diff
@@ -0,0 +1,54 @@
...
+            {
+                if (!SameTypeComparer.Instance.Equals(xParam.ParameterType, yParam.ParameterType))
+                {
+                    return false;
+                }
+            }
```
jskeet: Potentially consider default values as well?

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/SameTypeComparer.cs
jskeet: Or implement with `return x.GenericParameters.SequenceEqual(y.GenericParameters, this))`

tools/Google.Cloud.Tools.VersionCompat/Detectors/Class.cs
```diff
@@ -0,0 +1,194 @@
...
+                .Concat(cls.BaseType())
+                .Concat(classAndInterface.ImplementedInterfaces())
+                .Concat(classAndInterface.GenericConstraints())
+                .Concat(cls.Methods())
+                .Concat(cls.Properties());
+        }
+
+        private Class(TypeDefinition o, TypeDefinition n) => (_o, _n) = (o, n);
+
+        private readonly TypeDefinition _o;
```
jskeet: Rename to more useful names? I assume this is "original" and "new"?

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/SamePropertyComparer.cs
jskeet: Ah, I see now it's deliberate. Possibly add a comment explaining this?

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/SamePropertyComparer.cs
jskeet: (This probably comes under the same category as the above...)

tools/Google.Cloud.Tools.VersionCompat/Detectors/ClassAndInterface.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+        public ClassAndInterface(TypeDefinition o, TypeDefinition n) => (_o, _n) = (o, n);
+
+        private readonly TypeDefinition _o;
+        private readonly TypeDefinition _n;
+
+        public IEnumerable<Diff> ImplementedInterfaces()
+        {
+            // Treat any change to exported implemented interfaces is breaking.
+            // This isn't correct, as adding or removing an interface which an another interface itself implements is OK.
```
jskeet: And I'd say that adding an implemented interface in a concrete class is okay anyway... it would be okay for a type to add an implementation of `IEquatable<T>` for example.

Yes, in the strictest sense that could be breaking by changing overload resolution - but I suspect that given *any* public API change I could find some way it's breaking.

It may be that we should increase the separation between "Here are the semantic differences" and "This is what I consider to be breaking and what isn't" pretty soon, as I suspect the more code we write, the more there will be to change...

tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
chrisdunelm: Originally it was to make it very clear that there *is* a public ctor; but I agree it's mostly noise. Removed.

tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
chrisdunelm: Change to `Skip`

tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
chrisdunelm: Thanks. Good point. Done.

tools/Google.Cloud.Tools.VersionCompat.Tests/ClassTest.cs
chrisdunelm: Yes, completely agree. We need to be able to review them all easily and check everything is correct. Sometimes things really aren't very obvious!!

tools/Google.Cloud.Tools.VersionCompat.Tests/TestBase.cs
chrisdunelm: Done.

tools/Google.Cloud.Tools.VersionCompat.Tests/TopLevelTest.cs
chrisdunelm: Yes, agreed this could be nice, although I can't see an obvious way to do that right now!

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/SamePropertyComparer.cs
chrisdunelm: Comments added to each of the comparers.

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/SamePropertyComparer.cs
chrisdunelm: Yes, default values are deliberately not included here. Comment added.

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/SameTypeComparer.cs
chrisdunelm: Done

tools/Google.Cloud.Tools.VersionCompat/Detectors/Class.cs
chrisdunelm: I'd rather keep the names very short as they are used a lot throughout the code.
Comment added to briefly explain what they mean.

tools/Google.Cloud.Tools.VersionCompat/Detectors/ClassAndInterface.cs
chrisdunelm: Yes, I defined this as major breaking due to the overload ambiguity issue you mention.

However, I agree we need to define a separation between "theoretically breaking" and "pragmatically this is OK".

I think I'll leave this as it is until we have a better-defined process.

PR #2937, opened 2019-03-12
Try a different test style

chrisdunelm:


chrisdunelm:
Jon, do you think this testing style (in `TopLevelTest.cs`) is an improvement?
It brings the two parts of each test together, but it a little more verbose.

chrisdunelm:
The tests are already in the same file as the namespaces; although in a different root namespace.
I'll put the test-namespaces in the same namespaces as the test, just to make it as simple as possible, then we'll have the namespaces and tests in the same file, with a little verbosity as (easily) possible.

jskeet:
Ah, I'd misread the original diff. It's not as urgent as I'd thought before. Will still try to think about alternative options though...

chrisdunelm:
Please could you LGTM? I can't merge otherwise; and this minor change is a minor improvement ;)

PR #2938, opened 2019-03-12
[Storage] Adding missing conformance tests

frankyn:
Hi Jon!

Adding two conformance tests with expected values.

I do have the following question:
All CSEK headers are supported using V4 signed URLs. V2 has a restriction where only the `X-Goog-Encryption-Algorithm` header needs to be in the string to sign.


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json
```diff
@@ -21,6 +21,19 @@
...
+        "description": "POST for resumable uploads",
+        "bucket": "test-bucket",
+        "object": "test-object",
+        "method": "POST",
+        "expiration": 10,
+        "headers": {
+            "x-goog-resumable": [ "start" ]
+        },
+        "timestamp": "20190201T090000Z",
+        "expectedUrl": "https://storage.googleapis.com/test-bucket/test-object?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=test-iam-credentials%40dummy-project-id.iam.gserviceaccount.com%2F20190201%2Fauto%2Fstorage%2Fgoog4_request&X-Goog-Date=20190201T090000Z&X-Goog-Expires=10&X-Goog-SignedHeaders=host%3Bx-goog-resumable&x-goog-signature=7e9049ca7fcef224d92ab315e06bd54981577dcb7ff7eb67d1dddbe97ed07eb43bdd878534a5a56fccde24a9f63c0dbb822a5993b98b89c1519cb939f5edeb98711c826c0ea34cf17512c9dd0e1e2b8a793dd6840c9dbe272328a0b2f62b394a6253ed620f70ea46fd29be071bf9fa85a66d0c233d608b2f362a3322a57ae36fbf7d1f700001a933f315e4a653127a8c5ab48ab6146c2e6ed03c01f940161ce341eadcc88e88142428f88a085f938330083da6f1ed4af4c85d27a2bd213512662d5d04adfa89bef2e10e875789c9a1167bd296952d280afa2a9514bf92271a0f614f3f0cccb81925075db1195542642d5f9530c5560041e0c48d07e05ab42de0"
```
jskeet: The test is failing, at least in part because this should be X-Goog-Signature. But the signature looks different as well... I'll pull the PR and investigate.

(Note that we have an alternative way of specifying resumable uploads for C# as well, but I'd expect this to work too, I think. I'll have to refresh my memory.)

jskeet:
Okay, with the slight change to the casing, the signature is (as expected) still incorrect. The important intermediate values I've got are:

**Canonical request**

```text
POST
/test-bucket/test-object
X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=test-iam-credentials%40dummy-project-id.iam.gserviceaccount.com%2F20190201%2Fauto%2Fstorage%2Fgoog4_request&X-Goog-Date=20190201T090000Z&X-Goog-Expires=10&X-Goog-SignedHeaders=host%3Bx-goog-resumable
host:storage.googleapis.com
x-goog-resumable:start

host;x-goog-resumable
UNSIGNED-PAYLOAD
```

**Blob to sign**

```text
GOOG4-RSA-SHA256
20190201T090000Z
20190201/auto/storage/goog4_request
877f8b40179d2753296f2fd6de815ab40503c7a3c446a7b44aa4e74422ff4daf
```

I'm not quite sure what you mean about the CSEK part, but my CSEK integration tests are passing. I don't include any of the CSEK headers in the signature - when I did so, the integration tests failed. I don't know whether that's a problem in my code or the server implementation...

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json
frankyn: Face palm, mistakenly used GET instead of POST. Fixed. Thank you for sharing the intermediate steps!

frankyn:
Thanks @jskeet. 

Clarifying my question
For v2 the GCS backend does not allow `x-goog-encryption-key` & `x-goog-encryption-key-sha256` are not allowed to be part of the signature, but expects `x-goog-encryption-algorithm`.

For v4 the GCS backend AFAIU, does not have the restriction imposed by v2. 

I do see the [v4 client implementation imposes a restriction](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs#L173-L175) for the following three values: `x-goog-encryption-key` & `x-goog-encryption-key-sha256` & `x-goog-encryption-algorithm`.

> I'm not quite sure what you mean about the CSEK part, but my CSEK integration tests are passing. I don't include any of the CSEK headers in the signature - when I did so, the integration tests failed.

Did you sign all three in the request?

jskeet:
> Did you sign all three in the request only with v4?

No, that "continue" statement shows that those headers are *not* included in the signature.

PR #2939, opened 2019-03-13
Fix BigQueryRow TimestampConverter rounding issue

bcampbell-prosper:
Fixed rounding errors in TimestampConverter from double to DateTime for dates in the far future and updated scalar test with a value that fails for double and passes for decimal

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryRowTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryRowTest.cs
@@ -49,7 +49,7 @@ namespace Google.Cloud.BigQuery.V2.Tests
                     new TableCell { V = "AQI=" }, // 1, 2
                     new TableCell { V = "2.5" },
                     new TableCell { V = "text" },
-                    new TableCell { V = "1477566580.5" }, // 2016-10-27T11:09:40.500Z
+                    new TableCell { V = "95617584000.059" }, // 5000-01-01T00:00:00.059Z
                     new TableCell { V = "2017-08-09" },
                     new TableCell { V = "2017-08-09T12:34:56.123" },
                     new TableCell { V = "12:34:56.123" },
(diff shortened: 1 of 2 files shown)


jskeet:
Thanks for your contribution. Can I ask how you've observed this? I'm only asking because if the BigQuery servers are actually computing values using 64-bit binary floating point arithmetic in the first place, this might be "adding false precision" in a way. I'll need to validate with the team that we *should* be parsing this as a decimal.

jskeet:
I've just realized that there's another potential issue with this - an example in the documentation is "2.53402300800e11" which I suspect won't be handled by `decimal.Parse`. I'll need to validate whether the server *actually* returns values like that.

jskeet:
Additionally, could I check - if you *have* seen this, is it only when using legacy SQL? My impression is that the "unix seconds since epoch" is only used for the legacy SQL timestamp representation, not standard SQL. That doesn't mean we shouldn't fix it, of course, just trying to clarify the scope of the problem.

jskeet:
I've just tried this, inserting values 1 microsecond apart in the year 5000, and the values aren't being returned in the JSON precisely. (They *are* correct in storage.) They get returned in ~16 microsecond "chunks", which I strongly suspect means that basically they're being converted to `double` values before serialization... parsing as a decimal won't help on that front :(

bcampbell-prosper:
I signed it!

bcampbell-prosper:
I'm using the client library in my own app and one of my tests is a round trip of timestamp values with a random number of ticks.  Additionally, we use the value 9999-12-31 for our version_end_date for all type 2 dimension tables in our dimensional data warehouse.  Usually we use is_current_record = 1 for the "current" record but it's plausible an analyst would do "version_end_date = '9999-12-31'".

bcampbell-prosper:
Description in the [documentation for TIMESTAMP](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#timestamp-type) is "Represents an absolute point in time, with microsecond precision." which I don't think double precision floating point is being faithful to.  I realized after submitting this change that BQ is definitely representing this value as `double` under the hood though and had the same feeling about "adding false precision".

bcampbell-prosper:
I'm glad to submit a bug ticket somewhere if you'd like.

bcampbell-prosper:
`decimal.Parse` does indeed fail with "2.53402300800e11"

jskeet:
I've raised this internally already, but I have doubt that the backend implementation will change. I'll look into it more though.
(I think the scientific notation may be a red herring btw - I haven't seen any values coming back that way.)

Unless the backend changes, I don't think this change will help against the real service, even if fixes unit test problems.

Let's keep this open for now though, and I'll keep looking into options.

bcampbell-prosper:
I signed it!

jskeet:
Okay, I've looked a little bit further, and the server *is* giving the result with 64-bit precision, so it's not like there's any more precision to be gained in this format.
In the JSON, the result is presented as strings - so it's *possible* that more precision could be supplied, with an expectation that clients that couldn't handle the extra precision would effectively ignore it.

I'll file an issue internally to request either a change to include the extra precision, or at least documentation to indicate this limitation.

jskeet:
I've followed up internally and requested that this is highlighted in the documentation. I don't think the behavior is likely to change at this point.

In terms of the 9999-12-31 value, that should be fine - it's well within the range where the number of *whole seconds* can be represented accurately by a double.

I'm going to close this PR now as I believe we've done everything we can within this repo. Thanks very much for raising the issue though!

PR #2942, opened 2019-03-13
Fix generic type handling

chrisdunelm:


tools/Google.Cloud.Tools.VersionCompat/CecilUtils/ShowExtensions.cs
```diff
@@ -30,6 +30,18 @@ private static string Show(this IList<GenericParameter> ps)
             return "";
         }
 
+        private static string ShowGenArgs(this IList<TypeReference> types)
```
jskeet: Any reason for not using the conditional ?: operator and an expression bodied method here?

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/ShowExtensions.cs
```diff
@@ -43,13 +55,26 @@ private static string Show(this ParameterDefinition p)
...
                 sb.Append(" = ");
-                sb.Append(p.Constant ?? "null"); // TODO: I suspect `Constant` is not thw right way to do this.
+                sb.Append(p.Constant ?? "null"); // TODO: I suspect `Constant` is not the right way to do this.
             }
             return sb.ToString();
         }
 
-        public static string Show(this TypeReference type) =>
-            type.FullName == "System.Void" ? "void" : $"{type.FullName.Split('`')[0].Replace('/', '.')}{type.GenericParameters.Show()}";
+        public static string Show(this TypeReference type)
```
jskeet: As a nit, I think Format would be a better name than Show. But that's not really a matter of this PR.

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/ShowExtensions.cs
chrisdunelm: No reason, done :)

tools/Google.Cloud.Tools.VersionCompat/CecilUtils/ShowExtensions.cs
chrisdunelm: ack

PR #2949, opened 2019-03-15
Check method and property modifiers (virtual, abstract, sealed)

chrisdunelm:


chrisdunelm:
Actually, I'll close the PR, it doesn't handle the separate get/set modifiers properly.

PR #2966, opened 2019-03-21
Get rid of common issues from the legacy generator

jskeet:
(By applying a patch with manually-created changes after generation.)


jskeet:
If it doesn't, we scrap it. Simples :)

PR #2974, opened 2019-03-22
storage: convert Header val Array->Str

frankyn:
Hi @jskeet,

After reviewing a few implementations for multi-value headers ([S3 Java SDK](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/util/StringUtils.html#appendCompactedString-java.lang.StringBuilder-java.lang.String-), GCS backend).

The following is happening. 

```
# Provided by @quartzmo 
header_value.strip.gsub(/\s\s+/, " ")
```

1. Strip leading and trailing spaces.
1. Substitute contiguous white space for a single space.

Example 1:
```
"  abc  ,  def  ,  ghi  jkl  " -> "abc , def , ghi jkl"
```

The C# client will clean and concatenate multi-value headers:
```
"  abc  ,  def  ,  ghi  jkl  " -> "abc, def, ghi jkl"
```

The C# result will be accepted by the GCS backend, I just don't think they should be part of conformance tests in their current form. In this PR, I updated multi-value header Arrays into String equivalents.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json
@@ -1,136 +1,149 @@
 ﻿[
-    {
-        "description": "Simple GET",
-        "bucket": "test-bucket",
-        "object": "test-object",
-        "method": "GET",
-        "expiration": 10,
-        "timestamp": "20190201T090000Z",
-        "expectedUrl": "https://storage.googleapis.com/test-bucket/test-object?X-Goog-Algorithm=GOOG4-RSA-SHA256&X-Goog-Credential=test-iam-credentials%40dummy-project-id.iam.gserviceaccount.com%2F20190201%2Fauto%2Fstorage%2Fgoog4_request&X-Goog-Date=20190201T090000Z&X-Goog-Expires=10&X-Goog-SignedHeaders=host&X-Goog-Signature=95e6a13d43a1d1962e667f17397f2b80ac9bdd1669210d5e08e0135df9dff4e56113485dbe429ca2266487b9d1796ebdee2d7cf682a6ef3bb9fbb4c351686fba90d7b621cf1c4eb1fdf126460dd25fa0837dfdde0a9fd98662ce60844c458448fb2b352c203d9969cb74efa4bdb742287744a4f2308afa4af0e0773f55e32e92973619249214b97283b2daa14195244444e33f938138d1e5f561088ce8011f4986dda33a556412594db7c12fc40e1ff3f1bedeb7a42f5bcda0b9567f17f65855f65071fabb88ea12371877f3f77f10e1466fff6ff6973b74a933322ff0949ce357e20abe96c3dd5cfab42c9c83e740a4d32b9e11e146f0eb3404d2e975896f74"
-    },
+  {
+      "description": "Simple GET",
+      "bucket": "test-bucket",
+      "object": "test-object",
+      "method": "GET",
+      "expiration": 10,
+      "timestamp": "20190201T090000Z",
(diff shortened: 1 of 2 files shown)


jskeet:
I'm fine with losing the comments, and we might want to do that in a separate PR, but I'm not yet convinced about the multi-value headers. (And if we *do* make that change, we'll need to add another commit to this PR so that the test still works.)

Your example of

```text
"  abc  ,  def  ,  ghi  jkl  " -> "abc, def, ghi jkl"
```

is treating it as *one* value that happens to include commas - and at that point, C# will behave the same way as all the other languages.

The question is what happens if you specify multiple header values for the same name, so that the HTTP request itself looks like this:

```text
Foo:    abc   
Foo:   def  
Foo:    ghi
Foo:   jkl   
```

(There are trailing spaces as well.)

If a URL signer *just* joins those values together, I believe they'll create an invalid signature. For platforms that support specifying multiple values at both the URL signer and HTTP client level, I believe that's worth testing. If C# has the only URL signer API that allows multiple values to be specified, then I'm happy to add separate conformance tests just for C#. But if there are other platforms that allow it too, I think it's worth having in the common case. We could have a separate representation in the eventual proto, if that makes things easier for clients to skip (and makes simpler tests easier to express).

Do we know which platforms *do* allow this in the URL signing API?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
```diff
@@ -170,10 +170,10 @@ void AddHeaders(SortedDictionary<string, string> canonicalized, IDictionary<stri
                                 continue;
                             }
                             var headerName = pair.Key.ToLowerInvariant();
-                            var value = string.Join(", ", pair.Value.Select(PrepareHeaderValue)).Trim();
+                            var value = string.Join(",", pair.Value.Select(PrepareHeaderValue)).Trim();
```
jskeet: The previous code definitely worked with integration tests. Do you want me to pull this branch to check that they still pass?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
```diff
@@ -1,11 +1,11 @@
 ﻿// Copyright 2018 Google LLC
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
-//
```
jskeet: Please revert these whitespace changes.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerV4TestData.json
```diff
@@ -1,136 +1,149 @@
 ﻿[
-    {
```
jskeet: At this stage, what are the actual semantic changes in the JSON file? I'd prefer not to have these whitespace changes if there's no purpose to them. I'm happy to have a canonicalized format for the JSON files when we get to a conformance repo, potentially after some discussion about that formatting (I'm not sure why there's a two space indent for "object with array" but a four space indent for "field within object" here) - but let's work out the semantics first.

frankyn:
Thanks @jskeet, I split up the PRs to address each issue separately. 

> The question is what happens if you specify multiple header values for the same name, so that the HTTP request itself looks like this:

GCS will join then with a comma with no space on either side. 

```
# Incoming Request
Foo:    abc   
Foo:   def  

# GCS backend
foo:abc,def
```

Given this behavior the C# client would need to update the signer to not add a space after the comma. (added in [9782e13](https://github.com/googleapis/google-cloud-dotnet/pull/2974/commits/9782e135382b5632c7c7197797acfc5bcc586676))

> Do we know which platforms do allow this in the URL signing API?
I wasn't looking at `Foo` being defined multiple times with different values in a single request. I wasn't reviewing this test case correctly.

For example, reviewing the Ruby implementation. The surface API doesn't add a special helper for multi-valued headers, it allows a user to define a Hash with key/value pair. The key is case sensitive so multiple values to the same key with different cases is not handled correctly and @quartzmo is taking a look at it. 

Other languages may encounter the same issue as well and it makes sense to keep as-is. 

### About Changes
1. Updated the UrlSigner code to remove the space after the comma. (it's still broken I'll look at that next).
2. Updated the Conformance test for multiple values with a new signed URL example. 
3. Added an additional Conformance test for multi values in a single value.

jskeet:
Hmm - if that's the behavior of the GCS server, I'm confused about how the current integration tests are passing. I'll pull your PR and see whether the same tests *still* pass - if they don't, that suggests that the server isn't behaving in that way. Note that the URL v2 signer uses spaces too, although that's got some more complicated behaviour.

frankyn:
Thanks @jskeet, to clarify the update of removing space after the comma is to keep it consistent with what the GCS backend does when the same header name is used multiple times with different values.

That's a misstep on my part. It's not required to change to work. 

The GCS backend will accept "Foo: value1, value2, value3" so the integration tests won't break.

jskeet:
The integration tests *have* been broken by this change (I've just run it).

However, it's entirely possible that this depends on how the *platform* then sends an HTTP request where the user has specified multiple values. It may be that `HttpClient` (and its underlying infrastructure) is doing the concatenation. Checking now.

jskeet:
Right, it looks like that is indeed what's happening - `HttpClient` is sending a single value.
At this point, perhaps we should take the multiple value option out of the conformance *unit* tests, but recommend that each library that supports multiple values has appropriate *integration* tests to check that it follows its platform's behaviour correctly. What do you think?

frankyn:
Closing PR, after discussing this with @jskeet. 

There's an expectation that the V4Signing C# code is mimicking how the HTTPClient will handle provided headers in a subsequent request. 

The next step will be to move out the conformance tests for multi-valued headers as an ask for integration tests instead. 

The new multi-valued inline conformance test will be re-added in another PR as well.

PR #2981, opened 2019-03-26
Change targets for diagnostics libraries

### LINKED ISSUE #2959: Consider making Google.Cloud.Diagnostics.AspNetCore 3.0 only support ASP.NET Core 2.x
Currently Google.Cloud.Diagnostics.AspNetCore targets netstandard1.5, netstandard2.0 and net46. They have different dependencies, like this:

```xml
  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.5' OR  '$(TargetFramework)' == 'net46' ">
    <PackageReference Include="Microsoft.AspNetCore.Http" Version="1.0.4" />
    <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="1.0.4" />
    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="1.0.5" />
    <PackageReference Include="Microsoft.Extensions.Logging" Version="1.0.2" />
    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="1.0.2" />
  </ItemGroup>
  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.0.2" />
    <PackageReference Include="Microsoft.AspNetCore.Http.Extensions" Version="2.0.2" />
    <PackageReference Include="Microsoft.AspNetCore.Hosting.Abstractions" Version="2.0.2" />
    <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.1" />
    <PackageReference Include="Microsoft.Extensions.Logging.Debug" Version="2.0.1" />
  </ItemGroup>
```

This strikes me as confusing and potentially error-prone. It means that anyone using .NET desktop will end up with dependencies on ASP.NET Core 1.0, even if their application may target ASP.NET Core 2.x, for example. In reverse, it's *possible* that someone using .NET Core 2.0 may still actually be using ASP.NET Core 1.x, which this configuration would fail with.

Proposal: *just* have a target framework of netstandard2.0, and just use the existing ASP.NET Core 2.x dependencies (possibly updating them).

This is obviously a hugely breaking change - which is why it's important to make the decision *before* Google.Cloud.Diagnostics.AspNetCore hits 3.0 GA.

Cc users for comment:

- @SurferJeffAtGoogle
- @iantalarico 
- @henkmollema 
- @amanda-tarafa

jskeet:
After the change:

- Common: netstandard2.0
- AspNet: net461
- AspNetCore: netstandard2.0

Everything is tested under net461 and (except for AspNet) netcoreapp2.1.

Fixes #2959.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.IntegrationTests/Google.Cloud.Diagnostics.AspNet.IntegrationTests.csproj
+++ b/apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.IntegrationTests/Google.Cloud.Diagnostics.AspNet.IntegrationTests.csproj
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net452</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net452</TargetFrameworks>
+    <TargetFrameworks>net461</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net461</TargetFrameworks>
     <LangVersion>latest</LangVersion>
     <IsPackable>false</IsPackable>
     <AssemblyOriginatorKeyFile>../../GoogleApis.snk</AssemblyOriginatorKeyFile>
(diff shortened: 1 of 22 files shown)


jskeet:
(Will assign after it's green...)

apis/Google.Cloud.Diagnostics.AspNet/Google.Cloud.Diagnostics.AspNet.IntegrationTests/Google.Cloud.Diagnostics.AspNet.IntegrationTests.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net452</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net452</TargetFrameworks>
+    <TargetFrameworks>net461</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net461</TargetFrameworks>
```
chrisdunelm: There's a few redundant lines like this.
I wasn't sure if we wanted to keep them for consistency with csproj files where they *are* needed.

jskeet: Yup, these are generated by generateprojects.sh. I *think* I've removed all the handwritten conditional ones.

PR #2984, opened 2019-03-27
Fix trailing slash in UrlSigner.V4 conformance test.

remyabel:
Context: https://github.com/googleapis/google-cloud-cpp/issues/2350

We ran into an issue where passing an empty object to the url signer did not match the expected output. I identified the problem was the trailing slash. The code only checks for `null`, but string can be either empty or `null`. This fixes that.


frankyn:
@jskeet conformance tests will need to be updated once more to remove the trailing slash. 

GCS team don't have a preference but the gsutil was design without a trailing slash. To keep everything consistent let's remove the trailing slash when listing objects. 

Thanks @remyabel for sending this PR to kick off the changes downstream.

PR #2989, opened 2019-03-29
Only target netstandard2.0 for Spanner libraries

### LINKED ISSUE #2988: Consider making Spanner libraries v2 only target netstandard2.0
This is similar to #2959, but for Spanner.

Currently we have the following targeting:

- Spanner.Data: netstandard1.5;netstandard2.0;net45
- Spanner.Common.V1: netstandard1.5;net45
- Spanner.V1: netstandard1.5;netstandard2.0;net45
- Spanner.Admin.Database.V1: netstandard1.5;net45
- Spanner.Admin.Instance.V1: netstandard1.5;net45

Proposal: make all of these *just* target netstandard2.0.

This makes it simpler to reason about and simpler to test.

cc for input:

@FransBouma
@tomerpeled
@alyaros
@jerarl
@SurferJeffAtGoogle
@chrisdunelm 
@amanda-tarafa 

We may well want to change to make this the target for anything that hasn't already gone GA...

jskeet:
This would fix #2988 if we wanted to do it.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -657,7 +657,7 @@
 
   {
     "id": "Google.Cloud.Spanner.Admin.Database.V1",
-    "targetFrameworks": "netstandard2.0",
+    "targetFrameworks": "netstandard2.0;net45",
     "testTargetFrameworks": "netcoreapp2.0;net461",
     "productName": "Google Cloud Spanner Database Administration",
     "productUrl": "https://cloud.google.com/spanner/",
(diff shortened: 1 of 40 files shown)


apis/apis.json
```diff
@@ -621,7 +621,7 @@
 
   {
     "id": "Google.Cloud.Spanner.Admin.Database.V1",
-    "targetFrameworks": "netstandard2.0",
+    "targetFrameworks": "netstandard2.0;net45",
     "testTargetFrameworks": "netcoreapp2.0;net461",
```
amanda-tarafa: After bringing net45 back shouldn't tests run against net45 as well?

jskeet: Testing net461 will run the tests on the .NET framework - and IIRC, even specifying net45 here would still run under "the latest .NET framework installed" due to the way installations work.

jskeet:
(Adding "needs work" although really it's just "needs more input")

jskeet:
I think we're ready to merge this - PTAL.

PR #2990, opened 2019-04-01
Regenerate APIs (changes to Dialogflow and Bigtable)

jskeet:
Will self-review, as this is purely autogenerated and the Bigtable changes have been looked over already.


jskeet:
Assigned to Chris now just for the third commit, to check that I'm being sensible :)

jskeet:
Humbug - a little more to do in terms of dependencies. Fixing...

PR #2998, opened 2019-04-03
Datastore emulator support

### LINKED ISSUE #2499: Add support for Datastore emulator

jskeet:
Worth reviewing one commit at a time.

There's a TODO with a possible fix (which needs a GAX change) in DatastoreDbBuilder; thoughts on that possible fix very welcome.

We *could* merge this now (or after addressing the TODO) so long as we don't need to create a new GA release of Datastore before we create a GA release of GAX and ideally the generator for the builder.

If merged, would fix #2499.

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientBuilder.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientBuilder.cs
@@ -1,0 +1,82 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax.Grpc;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Datastore.V1
+{
+    // Note: the generator would just generate an internal property to start with, so there'd be no need for this
+    // partial.
+    public partial class DatastoreClient
+    {
+        internal static ChannelPool ChannelPool => s_channelPool;
+    }
+
+    // Note: this partial class would be generated.
+    /// <summary>
+    /// Builder class for <see cref="DatastoreClient"/> to provide simple configuration of credentials, endpoint etc.
+    /// </summary>
+    public sealed partial class DatastoreClientBuilder : ClientBuilderBase<DatastoreClient>
+    {
+        /// <summary>
+        /// The settings to use for RPCs, or null for the default settings.
+        /// </summary>
+        public DatastoreSettings Settings { get; set; }
+
+        /// <inheritdoc />
+        public override DatastoreClient Build()
+        {
+            Validate();
+            var callInvoker = CreateCallInvoker();
+            return DatastoreClient.Create(callInvoker, Settings);
+        }
+
+        /// <inheritdoc />
+        public override async Task<DatastoreClient> BuildAsync(CancellationToken cancellationToken = default)
+        {
+            Validate();
+            var callInvoker = await CreateCallInvokerAsync(cancellationToken).ConfigureAwait(false);
+            return DatastoreClient.Create(callInvoker, Settings);
+        }
+
+        /// <inheritdoc />
+        protected override ServiceEndpoint GetDefaultEndpoint() => DatastoreClient.DefaultEndpoint;
+
+        /// <inheritdoc />
+        protected override IReadOnlyList<string> GetDefaultScopes() => DatastoreClient.DefaultScopes;
+
+        /// <inheritdoc />
+        protected override ChannelPool GetChannelPool() => DatastoreClient.ChannelPool;
+    }
+
+    // Note: this partial class wouldn't be generated
+    public sealed partial class DatastoreClientBuilder : ClientBuilderBase<DatastoreClient>
+    {
+        /// <summary>
+        /// Creates a new instance with no settings.
+        /// </summary>
+        public DatastoreClientBuilder()
+        {
+        }
+
+        internal DatastoreClientBuilder(DatastoreDbBuilder builder)
+        {
+            Settings = builder.Settings;
+            CopyCommonSettings(builder);
+        }
+    }
+}
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
@@ -1,0 +1,188 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using Google.Api.Gax.Grpc;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Datastore.V1
+{
+    /// <summary>
+    /// Builder class for <see cref="DatastoreClient"/> to provide simple configuration of credentials, endpoint etc.
+    /// </summary>
+    public sealed partial class DatastoreDbBuilder : ClientBuilderBase<DatastoreDb>
+    {
+        /// <summary>
+        /// The settings to use for RPCs, or null for the default settings.
+        /// </summary>
+        public DatastoreSettings Settings { get; set; }
+
+        private EmulatorDetection _emulatorDetection;
+
+        /// <summary>
+        /// Specifies how the builder responds to the presence of emulator environment variables as described
+        /// by https://cloud.google.com/datastore/docs/tools/datastore-emulator.
+        /// </summary>
+        /// <remarks>
+        /// This property defaults to <see cref="EmulatorDetection.None"/>, meaning that environment variables are
+        /// ignored.
+        /// </remarks>
+        public EmulatorDetection EmulatorDetection
+        {
+            get => _emulatorDetection;
+            set => _emulatorDetection = GaxPreconditions.CheckEnumValue(value, nameof(value));
+        }
+
+        /// <summary>
+        /// The project ID, or null if this has not yet been configured. Note that a project ID must be configured
+        /// (or detected via an emulator environment variable) before building the <see cref="DatastoreDb"/>.
+        /// </summary>
+        public string ProjectId { get; set; }
+        
+        /// <summary>
+        /// Common code for handling the emulator configuration.
+        /// </summary>
+        private ConfiguredBuilder PrepareBuilder()
+        {
+            var clientBuilder = new DatastoreClientBuilder(this);
+            string projectId = ProjectId;
+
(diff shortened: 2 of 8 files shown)


apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
+﻿// Copyright 2018 Google LLC
```
chrisdunelm: nit: 2019

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+using Google.Api.Gax.Grpc;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Datastore.V1
+{
+    // Note: this file would be generated.
```
chrisdunelm: What would generate this file? I don't think `DatastoreDb` is generated at all is it?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// by https://cloud.google.com/datastore/docs/tools/datastore-emulator.
+        /// </summary>
+        /// <remarks>
+        /// This property defaults to <see cref="EmulatorDetection.None"/>, meaning that environment variables are
+        /// ignored.
+        /// </remarks>
+        public EmulatorDetection EmulatorDetection
+        {
+            get => _emulatorDetection;
+            set => _emulatorDetection = GaxPreconditions.CheckEnumValue(value, nameof(value));
```
chrisdunelm: Is is correct to validate this here?
Isn't the idiom to only validate on build?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+        /// <summary>
+        /// Common code for handling the emulator configuration.
+        /// </summary>
+        private ConfiguredBuilder PrepareBuilder()
+        {
+            // TODO: This is slightly awkward, as if a new property is added to ClientBuilderBase, this code will need to be updated.
+            // An alternative would be for Build to end up setting Endpoint and ChannelCredentials on *this* object, which
+            // might also be surprising, and would prevent Build() from being called again. Hmm.
+            // Possible solution: a ClientBuilderBase constructor accepting another ClientBuilderBase, which copies all properties.
+            // We could then have an internal DatastoreClientBuilder constructor accepting a DatastoreDbBuilder.
```
chrisdunelm: Agreed that it isn't great as it is.
Yes, I think your suggested alternative is a better approach.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+            PrepareBuilder().Build();
+
+        /// <inheritdoc />
+        public override Task<DatastoreDb> BuildAsync(CancellationToken cancellationToken = default) =>
+            PrepareBuilder().BuildAsync(cancellationToken);
+
+        // We never end up using these methods, at least with the current implementation
+        /// <inheritdoc />
+        protected override ServiceEndpoint GetDefaultEndpoint() =>
+            throw new InvalidOperationException($"This method should never execute in DatastoreDbBuilder");
```
chrisdunelm: nit: `$` not required (and following two methods)

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+            }
+
+            internal static EmulatorConfiguration FromEnvironment()
+            {
+                // Treat present-but-empty environment variables as if they were absent.
+                string hostAndPort = Environment.GetEnvironmentVariable(EmulatorHostVariable)?.Trim();
+                string projectId = Environment.GetEnvironmentVariable(EmulatorProjectVariable)?.Trim();
+                if (hostAndPort == "")
+                {
+                    hostAndPort= null;
```
chrisdunelm: nit: missing space.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+                Endpoint = endpoint;
+                ProjectId = projectId;
+            }
+
+            internal static EmulatorConfiguration FromEnvironment()
+            {
+                // Treat present-but-empty environment variables as if they were absent.
+                string hostAndPort = Environment.GetEnvironmentVariable(EmulatorHostVariable)?.Trim();
+                string projectId = Environment.GetEnvironmentVariable(EmulatorProjectVariable)?.Trim();
+                if (hostAndPort == "")
```
chrisdunelm: Is this necessary? This is only used below in a `string.IsNullOrEmpty(...)` call.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+                {
+                    hostAndPort= null;
+                }
+                if (projectId == "")
```
chrisdunelm: nit: And this could be moved to `projectId == "" ? null : projectId` in the ctor call.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/EmulatorDetection.cs
```diff
@@ -0,0 +1,50 @@
+﻿// Copyright 2018 Google LLC
```
chrisdunelm: nit: 2019

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/EmulatorDetection.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// <summary>
+        /// Ignore environment variables entirely.
+        /// </summary>
+        None = 0,
+
+        /// <summary>
+        /// Always connect to the production servers, but throw an exception if
+        /// environment variables exist that would suggest the emulator is expected.
+        /// </summary>
+        ProductionOnlyThrowOnEmulatorEnvironment = 1,
```
chrisdunelm: I slightly dislike the inconsistent naming between this and `EnumatorOnly`.
Either they should both have the `ThrowOn...` suffix or neither should.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -0,0 +1,247 @@
...
+                        GaxPreconditions.CheckState(
+                            emulatorConfig.ProjectId == null && emulatorConfig.Endpoint == null,
+                            "Emulator environment variables are set, contrary to use of {0}.{1}",
+                            nameof(EmulatorDetection), nameof(EmulatorDetection.ProductionOnlyThrowOnEmulatorEnvironment));
+                        break;
+                    case EmulatorDetection.EmulatorOnly:
+                        GaxPreconditions.CheckState(
+                            emulatorConfig.Endpoint != null,
+                            "Expected {0} environment variable to be set", EmulatorConfiguration.EmulatorHostVariable);
+                        ApplyEmulatorConfiguration();
```
chrisdunelm: Does `enumelatorConfig.ProjectId` need checking as not null?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientBuilder.cs
```diff
@@ -27,7 +27,7 @@ public partial class DatastoreClient
         /// Creates a new <see cref="DatastoreClientBuilder"/> instance. This method is primarily present for discoverability.
         /// </summary>
         /// <returns>A new <see cref="DatastoreClientBuilder"/> instance.</returns>
-        public DatastoreClientBuilder CreateBuilder() => new DatastoreClientBuilder();
+        public static DatastoreClientBuilder CreateBuilder() => new DatastoreClientBuilder();
```
chrisdunelm: Agreed that these methods should probably go. They'd never be idiomatically used.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientBuilder.cs
```diff
@@ -0,0 +1,73 @@
+﻿// Copyright 2018 Google LLC
```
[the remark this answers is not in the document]
jskeet: Thanks, will fix all of these.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientBuilder.cs
jskeet: We can always add them later, of course. They'd really only be present for discoverability.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
jskeet: You're right. Will remove that.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
jskeet: Not for things that *can* be validated independently. If the value isn't a valid enum value, it can never be correct, so it's okay to validate here, IMO.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
jskeet: Righto, will get on that.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
jskeet: Nope - it's okay for just the endpoint to be set, if the project ID has been set in the builder.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
jskeet: Thanks, will fix.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
jskeet: Hmm... suspect not. Will fix.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
jskeet: Done.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/EmulatorDetection.cs
jskeet: Hmm. Will make it just ProductionOnly for now. We have a while before this will go GA, and should probably do another review of the API before then anyway.

jskeet:
I've rebased this into a couple of commits and made a couple of fixups.

Please rereview just for the public API; the plan is to then merge, update (with other APIs) to GAX 2.7.0-beta04, then release a new beta.

PR #3014, opened 2019-04-11
Allow conversion from integer values to double in Firestore deserialization

### LINKED ISSUE #3013: Firestore SDK think a number is integer instead of double
Getting "Unable to convert integer value to System.Double" on ConverterBase.cs :line 94

#### Steps to reproduce

1. create a document using firestore C# sdk with Double field
2. modify the field using Google Firestore console or cloud function but set the number to whole number.
3. retrieve the same document using C# sdk.
If you only modify the fields through C# sdk, it will always work, even if you set it as whole number using SDK.

jskeet:
Fixes #3013

Note: I'm not yet sure that we want to merge this, but I want to get it ready if we *do* want to merge it.


jskeet:
Added an extra commit for the reverse conversions, as well as handling float as well as double. PTAL - just at the last commit.

PR #3015, opened 2019-04-12
Add prefix to leading-digit log parameters and refactor GoogleLogger

jskeet:
Best reviewed one commit at a time

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
@@ -119,21 +119,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore
 
                 Struct jsonStruct = CreateJsonPayload(eventId, state, exception, message);
 
-                Dictionary<string, string> labels;
-                var labelProviders = GetLabelProviders()?.ToArray();
-                if (labelProviders?.Length > 0)
-                {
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
```diff
@@ -217,6 +206,32 @@ public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Except
                 _consumer.Receive(new[] { entry });
             }
             catch (Exception) when (_loggerOptions.RetryOptions.ExceptionHandling == ExceptionHandling.Ignore) { }
+
+            Struct CreateStruct(IEnumerable<KeyValuePair<string, object>> fields)
+            {
+                Struct fieldsStruct = new Struct();
+                foreach (var pair in fields)
```
chrisdunelm: This only prefixes with `_` if *all* chars are digits.
Why not prefix with `_` if just the first char is a digit?

jskeet: Hmm... good point, I think that would work - and be simpler, of course. Will fix. (I'll skip any empty fields at the same time, as I don't think they'd be valid.)

jskeet: Added one extra commit - PTAL. (It's very small :)

PR #3020, opened 2019-04-16
Use CommandTimeout to control just how long Read() can take to complete.

### LINKED ISSUE #2809: Separate query timeout from other RPC timeouts in Spanner
Currently we have a single default timeout in SpannerConnectionStringBuilder. That defaults to 1 minute, which isn't always appropriate - a large query may well take more than 1 minute to complete. Currently we retry DeadlineExceeded when reading anyway, but a) it's not clear whether we should; b) it would be better not to have to.

I suspect that we need a second timeout in SpannerConnectionStringBuilder such as StreamingTimeout. The tricky part is then working out what should be used in SpannerCommand, as we don't always know the command type.

jskeet:
This mimics SqlDataReader, and also solves the issue of long-running queries - we don't need to worry about them taking more than the default time (1 minute) to complete, as that's now only applied on a per-read basis.

Fixes #2809.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
@@ -273,55 +273,54 @@ namespace Google.Cloud.Spanner.Data
         public override Task<bool> ReadAsync(CancellationToken cancellationToken) =>
             ExecuteHelper.WithErrorTranslationAndProfiling(async () =>
                 {
-                    // Note: the using statement ensures that the timer tasks is closed appropriately; otherwise
-                    // we could flood the system with timers.
                     using (var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(_readTimeoutSeconds)))
                     {
                         var timeoutToken = timeoutCts.Token;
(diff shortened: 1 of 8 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -267,44 +270,62 @@ public override bool NextResult()
...
-                    var first = await _resultSet.NextAsync(cancellationToken).ConfigureAwait(false);
-                    if (first == null)
+                    // Note: the using statement ensures that the timer tasks is closed appropriately; otherwise
+                    // we could flood the system with timers.
+                    using (var timeoutCts = new CancellationTokenSource(TimeSpan.FromSeconds(_readTimeoutSeconds)))
                     {
-                        // If this is the first thing we've tried to read, then we know there are no rows.
-                        if (_hasRows == null)
+                        var timeoutToken = timeoutCts.Token;
+                        var effectiveToken = !cancellationToken.CanBeCanceled ? timeoutToken
```
chrisdunelm: `effectiveToken` isn't being disposed of; which it probably should be if it's a linked source.

chrisdunelm:
Hmm, CI disagrees with me ;)

jskeet:
That's fine - it's a test that I hadn't updated in terms of what's retriable. Will fix that and amend the commit.

PR #3021, opened 2019-04-17
Always dispose of a linked CancellationTokenSource

jskeet:
This does involve always *creating* a CTS when previously we might have avoided it, but that's probably insignificant.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Transaction.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Transaction.cs
@@ -76,8 +76,10 @@ namespace Google.Cloud.Firestore
         {
             GaxPreconditions.CheckNotNull(documentReference, nameof(documentReference));
             GaxPreconditions.CheckState(_writes.IsEmpty, "Firestore transactions require all reads to be executed before all writes.");
-            CancellationToken effectiveToken = GetEffectiveCancellationToken(cancellationToken);
-            return documentReference.GetSnapshotAsync(TransactionId, effectiveToken);
+            using (var cts = CancellationTokenSource.CreateLinkedTokenSource(CancellationToken, cancellationToken))
+            {
+                return documentReference.GetSnapshotAsync(TransactionId, cts.Token);
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Transaction.cs
```diff
@@ -76,8 +76,10 @@ public Task<DocumentSnapshot> GetSnapshotAsync(DocumentReference documentReferen
         {
             GaxPreconditions.CheckNotNull(documentReference, nameof(documentReference));
             GaxPreconditions.CheckState(_writes.IsEmpty, "Firestore transactions require all reads to be executed before all writes.");
-            CancellationToken effectiveToken = GetEffectiveCancellationToken(cancellationToken);
-            return documentReference.GetSnapshotAsync(TransactionId, effectiveToken);
+            using (var cts = CancellationTokenSource.CreateLinkedTokenSource(CancellationToken, cancellationToken))
```
chrisdunelm: I agree it's probably insignificant, but it would be fairly simple to arrange this so it still only creates a linked source when necessary.
E.g. use something like

```C#
using (GetEffectiveCancellationToken(cancellationToken, out var effectiveToken) { ... }

IDisposable GetEffectiveCancellationToken(CancellationToken token, out CancellationToken effectiveToken)
{
    if (!CancellationToken.CanBeCanceled)
    {
        effectiveToken = token;
        return null;
    }
    else if (!other.CanBeCanceled)
    {
        effectiveToken = CancellationToken;
        return null;
    }
    var linkedCts = CancellationTokenSource.CreateLinkedTokenSource(CancellationToken, other);
    effectiveToken = linkedCts.Token;
    return linkedCts;
}
```

jskeet: Hmm. I think I'll stick with it as it is for now - the fact that I can't remember off the top of my head what happens with a using statement and a null resource suggests it's already a little tricksy. (We could have a no-op IDisposable singleton for that case, admittedly.)

chrisdunelm: It does the right thing, and doesn't attempt to call `Dispose()` on it :)

Issue #3022, opened 2019-04-17
Consider removing netstandard1.5 support from Firestore


jskeet:
This is similar to #2988 and #2959, but for all the Firestore libraries. We need to consciously make this decision before the Firestore libraries go GA. (Removing a target framework would be a breaking change.)

The plan would be to remove the target of netstandard1.5, which would mean it couldn't be used from .NET Core 1.x.

I don't currently have usage data for the API to indicate whether it's *being* used from .NET Core 1.x. I'm trying to work out why that is.

We're going to consider this approach more broadly for all APIs, but it's somewhat more important for Firestore due to the manually-written code, which may well increase over time. Only requiring support for net45 and netstandard2.0 gives us a wider underlying API surface to use.

I'd love feedback from other Firestore users, hence the tagging of users who've previously asked questions or reported issues. If you're able to indicate whether this would be a problem or not for you, that would be great!

- @alexdoan102
- @codeaid
- @veatcha
- @gnpretorius 
- @nicksav
- @pauloevpr
- @DenSmoke
- @wikes82
- @hiranya911
- @jadekler
- @LeXa777
- @joshuabooker
- @Thaina
- @sasog23
- @enriko-riba
- @Watanyou
- @MrMonotone
- @Eamonn-Duffy
- @ricsmania
- @chalber
- @gbaccetta
- @babduljameel
- @mhh12121
- @ryanmats
- @Hussamalsh
- @deepthi-bhattachar
- @AnidjarDavid
- @Sohojoe

DenSmoke:
Currently I'm using Firestore libraries on .NET Core 2.2 and 3.0 preview projects and .NET Standard 2.1 in some cases, so it wouldn't be a problem

veatcha:
No issues here...

wikes82:
Same here, no issues

pauloevpr:
No issues here.

jskeet:
Given the positive responses so far, I'm removing it from the beta release that I'm about to create. This isn't the final word: even if we go GA without netstandard1.5, we can add it back later as a non-breaking change. We can't remove it if it's in there for 1.0 though.

jskeet:
Closing this for now as a *fait accompli* - we can reopen for more careful consideration if anyone has any problems.

PR #3026, opened 2019-04-18
Remove netstandard1.5 from the Firestore API

jskeet:
This is a speculative fix for #3022; we can re-add it later on as a non-breaking change.


chrisdunelm:
Looks like `Google.Cloud.Firestore.Benchmarks` also needs updating :)

jskeet:
Oh humbug... thanks!

jskeet:
Righto, it's green now - PTAL.

Issue #2876, opened 2019-02-14
Running query performance is affected by use of ListRows vs GetQueryResults


Zhenya-Ren:
#### Environment details

* OS: Windows 10 64 bit

* .NET version: .Net Framework 4.5

* Package name and version:
  Google.Cloud.BigQuery.V2.dll 1.2.0
  Google.Api.Gax.Rest 2.5.0
  Google.Apis.Bigquery.v2 1.36.1.1404
  Google.Api.Gax  2.5.0
  Google.Apis.Auth 1.36.1
  Newtonsoft.Json 12.0.1.22727
  Google.Apis 1.36.1
  Google.Apis.Core 1.36.1

#### Steps to reproduce

  1. We make a query to air_quality_annual_summary table of 1 gb. The query returns all columns, 10000 rows. Running the query in the web ui (beta) takes around 2.5 seconds and 0.02 sec if cached. 
  2. We ran the query via api, like this:

`BigQueryResults results = this.client.ExecuteQuery(query, null);`

Running this takes around 5-6 seconds and a bit more than 3 seconds if cached, which is probably fine due to sending and parsing the result. Or? Is is as expected?

3. Once we received BigQueryResults we iterate through the rows.

`foreach (BigQueryRow row in results)
{
   // do something quick
}`

This iteration takes another 5-6 seconds. I looked in the code - this happens because when running the query, the query is saved into destination table. When iterating the rows are read from destination table with ListRows method. This unfortunately takes twice as long as it could be.

4. We now use a workaround to speed things up:

`var job = this.client.CreateQueryJob(query, null);`
`JobsResource.GetQueryResultsRequest queryResults = this.client.Service.Jobs.GetQueryResults(job.Reference.ProjectId, job.Reference.JobId).Execute();`

// Running query takes again around 5-6 seconds for non-cached and around 3 seconds for cached.

`foreach (TableRow row in queryResults.Rows)
{
 // do something quick
}`

// Iteration takes around 100 ms regardless if data was cached or not.

So the performance is 5 vs 10 seconds for not cached data and 3 vs 8 seconds for cached data, which is quite a big difference.

Yes, it is true that if the returned result would be more than 10 mb, than we would have to read the following pages and it would take longer time (this I haven't tested), however many queries are smaller than that.

It feels like it would be a good option to be able to choose if destination table should be used.

[...]

PR #3035, opened 2019-04-23
Reimplement BigQuery query execution to use GetQueryResults

jskeet:
Background: we used to use GetQueryResults, then we were asked to use ListRows - but this slows down queries, as observed in #2876. It turns out that GetQueryResults is now the preferred client library usage, so it's worth changing back to it.

(This review is fine to wait until you're back from the US, Chris.)

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryResults.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryResults.cs
@@ -30,8 +30,9 @@ namespace Google.Cloud.BigQuery.V2
     public sealed class BigQueryResults : IEnumerable<BigQueryRow>
     {
         private readonly BigQueryClient _client;
-        private readonly ListRowsOptions _options;
+        private readonly GetQueryResultsOptions _options;
         private readonly GetQueryResultsResponse _response;
+        private readonly Dictionary<string, int> _fieldNames;
 
         /// <summary>
         /// The reference to the query job.
@@ -71,6 +72,11 @@ namespace Google.Cloud.BigQuery.V2
         public long? NumDmlAffectedRows => _response.NumDmlAffectedRows;
 
         /// <summary>
+        /// The rows in the response, or an empty sequence if the response contains no rows.
+        /// </summary>
+        private IEnumerable<BigQueryRow> ResponseRows => ConvertResponseRows(_response);
+
+        /// <summary>
         /// Returns an asynchronous sequence of rows from this set of results.
         /// </summary>
         /// <remarks>
@@ -78,14 +84,7 @@ namespace Google.Cloud.BigQuery.V2
         /// ambiguity.</para>
         /// </remarks>
         /// <returns>An asynchronous sequence of rows from this set of results.</returns>
-        public IAsyncEnumerable<BigQueryRow> GetRowsAsync() => _client.ListRowsAsync(TableReference, Schema, _options);
-
-        /// <summary>
-        /// Returns an iterator over the query results.
-        /// </summary>
-        /// <returns>An iterator over the query results.</returns>
-        public IEnumerator<BigQueryRow> GetEnumerator() =>
-            _client.ListRows(TableReference, Schema, _options).GetEnumerator();
+        public IAsyncEnumerable<BigQueryRow> GetRowsAsync() => new AsyncRowEnumerable(this);
 
         /// <summary>
         /// Returns an iterator over the query results.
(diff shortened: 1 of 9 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryResults.cs
```diff
@@ -122,10 +163,32 @@ public BigQueryResults(BigQueryClient client, GetQueryResultsResponse response,
...
-            // Make sure we start off by trying to read the right page size...
-            var options = GetOptionsWithPageSize(pageSize);
-            var page = _client.ListRows(TableReference, Schema, options).ReadPage(pageSize);
-            return new BigQueryPage(page, Schema, JobReference, TableReference);
+            GetQueryResultsOptions clonedOptions = _options?.Clone() ?? new GetQueryResultsOptions();
+            List<BigQueryRow> rows = new List<BigQueryRow>(pageSize);
+
+            // Work out whether to use the response we've already got, or create a new one.
+            GetQueryResultsResponse response = _response;
+            if ((clonedOptions.PageSize == null || clonedOptions.PageSize > pageSize) && _response.Rows?.Count > pageSize)
```
chrisdunelm: I'm unclear why this isn't just `if (_response.Rows?.Count > pageSize)`. If `_response` contains more than `pageSize` rows, doesn't it always need re-starting?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryResults.cs
```diff
@@ -192,5 +260,58 @@ public BigQueryResults ThrowOnAnyError()
...
+            public AsyncRowEnumerable(BigQueryResults parent) => _parent = parent;
+
+            public IAsyncEnumerator<BigQueryRow> GetEnumerator() => new AsyncRowEnumerator(_parent);
+        }
+
+        private sealed class AsyncRowEnumerator : IAsyncEnumerator<BigQueryRow>
+        {
+            private readonly GetQueryResultsOptions _options;
+            private readonly BigQueryResults _parent;
+            private IEnumerator<BigQueryRow> _underlyingIterator;
```
chrisdunelm: nit: rename to `_underlyingEnumerator`?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryResults.cs
```diff
@@ -192,5 +260,58 @@ public BigQueryResults ThrowOnAnyError()
...
+                _options.StartIndex = null;
+                _options.PageToken = parent._response.PageToken;
+                _underlyingIterator = parent.ResponseRows.GetEnumerator();
+            }
+
+            public BigQueryRow Current => _underlyingIterator.Current;
+
+            public async Task<bool> MoveNext(CancellationToken cancellationToken)
+            {
+                cancellationToken.ThrowIfCancellationRequested();
```
chrisdunelm: Should this be inside the `while` loop below?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryResults.cs
jskeet: Yes, I think you're right. Will simplify it in both places.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryResults.cs
jskeet: I guess. I prefer the term "iterator" over "enumerator", but that's just a personal thing. Done.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryResults.cs
jskeet: We at least need one call *outside* the loop, so that it can still be cancelled even if we're iterating over the content of an existing response. Within the loop, I don't *think* we need one as I'd expect the cancellation token usage in the RPC to be enough.

jskeet:
Added another commit addressing the comments, but would appreciate a look at them if you have time - particularly my reasoning around *not* changing the cancellation check placement.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryResults.cs
chrisdunelm: Ah yes, I had missed that `cancellationToken` is passed into `GetRawQueryResultsAsync(...)`.

PR #3038, opened 2019-04-25
GA releases for Scheduler and Tasks

jskeet:
After this is merged, we'd delist Tasks.V2Beta2, Tasks.V2Beta3 and Scheduler.V1Beta1. (Those are the beta *API* packages, not the beta releases for the GA package.)


jskeet:
Note: the CI failures are to be expected from this set of commits. They generally happen when deleting source code. I'll separate them out into a separate PR for simplicity.

jskeet:
I've separated out the removal parts into #3039 so *this* PR should pass CI.

PR #3039, opened 2019-04-25
Remove beta APIs of Tasks and Scheduler

jskeet:
I expect CI to fail with this PR as it will think that the beta APIs have been changed and therefore need to be built... it'll all be fine afterwards though.


jskeet:
(Will wait for Dan's feedback on #3038 before merging both and releasing.)

PR #3040, opened 2019-04-25
Beta of pubsub

chrisdunelm:


chrisdunelm:
The changes are some extra fields in the protos from the regeneration in #2980

PR #3046, opened 2019-05-01
Update dependencies now that GAX 2.7.0 has been released

jskeet:
Will wait for this to go green, then review it carefully myself, before assigning to Chris. (Won't be today...)

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -199,7 +199,7 @@
       "Microsoft.AspNet.WebApi.Core": "5.2.7",
       "Microsoft.AspNet.WebPages": "3.2.7",
       "Microsoft.AspNet.Mvc": "5.2.7",
-      "Grpc.Core": "1.20.1"
+      "Grpc.Core": "default"
     },
     "testDependencies": {
       "Google.Cloud.Diagnostics.Common.Tests": "project",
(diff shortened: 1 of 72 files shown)


jskeet:
Okay, it's ready for review. The first commit is the manual part, and the commit message explains what I've attempted to do :)

apis/apis.json
```diff
@@ -226,12 +219,12 @@
     "dependencies": {
       "Google.Cloud.Diagnostics.Common": "project",
       "Google.Cloud.Diagnostics.AspNetCore.Analyzers": "1.0.0-beta02",
-      "Grpc.Core": "1.19.0"
+      "Grpc.Core": "1.20.1"
```
chrisdunelm: Should references to `Grpc.Core` be `default`, rather than a specific version number?
`Google.Cloud.Firestore` uses `default`.

jskeet: Good point - this could absolutely be the default. It's fine to be default for any non-beta version. Will fix.

jskeet: I've added another commit for this; it doesn't change the project files (which is exactly what I expected). Will merge on green.

chrisdunelm:
Still LGTM :)

PR #3050, opened 2019-05-02
Temporarily disable all HMAC calls

jskeet:
(The service has disabled this non-GA feature temporarily, causing all tests to fail.)

Note: I'll self-merge with admin super-powers if Chris doesn't have time to review today; I want to get this in before the integration tests run internally tonight.


jskeet:
Integration tests passed locally.

PR #3051, opened 2019-05-02
Script to check compatibility between two branches

jskeet:
Currently this compares all APIs. We may want something more granular eventually, but this is a start...

### DIFF
--- a/checkcompatibility.sh
+++ b/checkcompatibility.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+set -e
+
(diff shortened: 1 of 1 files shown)


jskeet:
Ping for a review?

chrisdunelm:
Sorry! Reviewing now

jskeet:
Yup - I'm assuming there's no requirement to get back to the original branch (which may not have been master). We can always fix that later if necessary.

PR #3054, opened 2019-05-03
Fix manually written snippets, which are now ambiguous

jskeet:
I'm still running the builddocs.sh script locally, and won't merge until that's complete, but it's looking good so far.


jskeet:
Monitoring needs fixing too...

Issue #3061, opened 2019-05-08
Wrong Date type formatter for date 0001-01-01


Zhenya-Ren:
#### Environment details

OS: Windows 10 64 bit

.NET version: .Net Framework 4.5

Package name and version:
Google.Cloud.BigQuery.V2.dll 1.2.0
Google.Api.Gax.Rest 2.5.0
Google.Apis.Bigquery.v2 1.36.1.1404
Google.Api.Gax 2.5.0
Google.Apis.Auth 1.36.1
Newtonsoft.Json 12.0.1.22727
Google.Apis 1.36.1
Google.Apis.Core 1.36.1

#### Steps to reproduce

  1. The following query is an example from BigQuery documentation:

SELECT NULL AS x, '' AS y, STRUCT(false AS a, DATE '0001-01-01' AS b) AS s

  2. Running this query in console or with GetQueryResults method in API gives the json result:

[
  {
    "x": null,
    "y": "",
    "s": {
      "a": "false",
      "b": "1-01-01"
    }
  }
]

3. To convert string result to DateTime, the following converter is used by BigQueryRow:

private static readonly Func<string, DateTime> DateConverter = v => DateTime.ParseExact(v, "yyyy-MM-dd", CultureInfo.InvariantCulture);

4. This converter is however not valid for "1-01-01" string.

It still works fine for recent dates like '2017-01-01'. And I think, it did work before. So, perhaps from some point the server started to return "1-01-01" instead of "0001-01-01".

jskeet:
I'm pleased to say this has now been fixed server-side. Thank you so much for reporting it.

PR #3062, opened 2019-05-08
Integration test for #3061

jskeet:
This isn't ready to be submitted yet, because it fails. I'm going to investigate whether there needs to be a server-side change, or whether we should fix this locally.


jskeet:
(Fix is merged, but we're waiting for it to roll out.)

jskeet:
The test now passes! Ready to merge after review.

PR #3071, opened 2019-05-10
Regenerate Language service

jskeet:
This has nasty embedded HTML comments, but we won't release the package until the internal Markdown version has been published and we've regenerated from that anyway. This way we avoid having to discard these changes every time we regenerate.


jskeet:
It looks like extra flattenings have been introduced that conflict with our existing partial classes. Will look at this next week.

jskeet:
Closing for now; we still need to decide what to do with this.

PR #3076, opened 2019-05-15
Add Grafeas V1 API

jskeet:
Will check it all myself first, then assign to Chris.


jskeet:
We're still discussing naming and what exactly should be in the library.

jskeet:
Closing in favour of #3085, but will leave the branch present.

PR #3083, opened 2019-05-17
Dispose correctly, and properly ignore cancellation

chrisdunelm:


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
```diff
@@ -515,24 +515,25 @@ private void DelaySendCurrentBatch()
             if (_batchDelayThreshold is TimeSpan batchDelayThreshold)
             {
                 // read cancellation token here, in case the current batch changes before the task below starts.
```
jskeet: Nit - I know it wasn't in this PR, but could we have a capital "R" for "Read"?

chrisdunelm: done!

PR #3103, opened 2019-06-04
Pattern Conversions in Log4Net custom labels

### LINKED ISSUE #2983: Log4Net StackdriverAppender: Use %property values in custom labels
Attempting to use %property syntax on custom label values. It shows up in Stackdriver like this:

```
labels: {
  Key:  "%property{AccountID}"   
...
 }
```
I want the value inserted at runtime, similar to what I can do with the message:

`<conversionPattern value="%property{AccountID}` ...

I'd like to label my logs with some context from the application.

```
<customLabel>
      <key value="AccountID" />
      <value value="%property{AccountID}" />
</customLabel>
```

Put the dynamic data into the log message directly.

 **Additional context**
- [I think this is where a replacement would happen.](https://github.com/googleapis/google-cloud-dotnet/blob/b36872c110349c3d73106d333396c9e24dc494e2/apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs#L315)
- I saw that adding new labels dynamically was prevented completely in #1036. I would just be looking to change the values dynamically.

chrisdunelm:
Fixes #2983

### DIFF
--- a/apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
+++ b/apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
@@ -203,7 +203,8 @@ namespace Google.Cloud.Logging.Log4Net
                 Severity = LogSeverity.Warning,
                 LogName = _logName,
                 Resource = _resource,
-                // This does not use patterns in the custom labels if they are enabled.
+                // Patterns included in custom labels will not be used in this "logs lost" entry.
+                // The pattern itself will be logged, regardless of the "UsePatternWithinCustomLabels" setting.
                 // This is acceptable as most patterns will be irrelevant in this context.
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -190,6 +203,8 @@ public override void ActivateOptions()
                 Severity = LogSeverity.Warning,
                 LogName = _logName,
                 Resource = _resource,
+                // This does not use patterns in the custom labels if they are enabled.
```
jskeet: I don't really understand this comment. Could you clarify it a bit? Admittedly I don't know about the feature very well...

chrisdunelm: Clarification attempted

PR #3116, opened 2019-06-10
Regenerate all APIs with new generator, including ClientBuilders

jskeet:
(Only the second and third commits need reviewing)

### DIFF
--- a/apis/Google.Cloud.Kms.V1/Google.Cloud.Kms.V1/Google.Cloud.Kms.V1.csproj
+++ b/apis/Google.Cloud.Kms.V1/Google.Cloud.Kms.V1/Google.Cloud.Kms.V1.csproj
@@ -25,7 +25,7 @@
     <PackageReference Include="ConfigureAwaitChecker.Analyzer" Version="1.0.1" PrivateAssets="All" />
     <PackageReference Include="Google.Api.Gax.Grpc" Version="2.7.0" />
     <PackageReference Include="Google.Cloud.Iam.V1" Version="1.1.0" />
-    <PackageReference Include="Grpc.Core" Version="1.18.0" PrivateAssets="None" />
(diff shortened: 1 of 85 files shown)


chrisdunelm:
The build is failing with `The type or namespace name 'ClientBuilderBase<>' does not exist in the namespace 'Google.Api.Gax.Grpc' (are you missing an assembly reference?)`.
Is the wrong version of GAX being referenced?

jskeet:
Ah - Bigtable is still failing because there's a custom generation step there. I'll need to look at that separately.

jskeet:
(The others were due to project dependency issues.)

chrisdunelm:
(Problem with Grpc.Core version in KMS)

apis/Google.Cloud.Kms.V1/Google.Cloud.Kms.V1/Google.Cloud.Kms.V1.csproj
```diff
@@ -25,7 +25,7 @@
     <PackageReference Include="ConfigureAwaitChecker.Analyzer" Version="1.0.1" PrivateAssets="All" />
     <PackageReference Include="Google.Api.Gax.Grpc" Version="2.7.0" />
     <PackageReference Include="Google.Cloud.Iam.V1" Version="1.1.0" />
-    <PackageReference Include="Grpc.Core" Version="1.18.0" PrivateAssets="None" />
+    <PackageReference Include="Grpc.Core" Version="1.20.1" PrivateAssets="None" />
```
chrisdunelm: Why not version 1.21.0?

jskeet: Because the default specified by the project generator is 1.20.1. We can update that in the future of course.

PR #3122, opened 2019-06-11
Disable the ordering-keys feature in publisher

chrisdunelm:


chrisdunelm:
Commit message updated. Will merge on green.

PR #3130, opened 2019-06-13
Regenerate all APIs

jskeet:


jskeet:
The Dialogflow generated unit tests are failing (!). I'll look at whether this is the tests or the generator that's having a problem.

jskeet:
Interesting. It's about the x-goog-request-params header in the SetAgent method, if the request has no Agent present.

For now I'll add a null-safe operator, but I'll raise another issue to work out what we should actually do.

PR #3133, opened 2019-06-14
Fix error in handling non-ordered batch errors

chrisdunelm:


chrisdunelm:
The integration test now passes.

PR #3135, opened 2019-06-14
Introduce client builders for Bigtable

### LINKED ISSUE #3117: Support client builders for Bigtable
They don't work out of the box due to Bigtable using a call invokers pool instead of a channel pool. I'm sure it's fixable; we just to handle it carefully.
As a temporary measure, I'm removing the builder to unblock all other APIs.

jskeet:
BigtableClientBuilder has no settings beyond BigtableServiceApiClientBuilder; it's just more convenient.

(The GAX update is required to get at channel options.)

Fixes #3117.


jskeet:
Definitely worth reviewing one commit at a time :)

PR #3137, opened 2019-06-14
Fix release process

jskeet:
The real build is done in releasebuild, so that's where we need to process the build log.


jskeet:
(The CI will take ages on this, but will be irrelevant - so don't wait for that :)

PR #3151, opened 2019-06-19
Introduce new attribute to populate the Firestore document ID on deserialization

### LINKED ISSUE #3042: FirestoreDataAttribute and FirestorePropertyAttribute are really useful, and would be even more useful with one an Id Attribute.
The class I'm storing in firestore looks like this:
```cs
    [FirestoreData]
    public class Book
    {
        [Key]
        [BindNever]
        public string Id { get; set; }

        [Required]
        [FirestoreProperty]
        public string Title { get; set; }

        [FirestoreProperty]
        public string Author { get; set; }

        [Display(Name = "Date Published")]
        [DataType(DataType.Date)]
        [FirestoreProperty]
        public DateTime? PublishedDate { get; set; }

        [FirestoreProperty]
        public string ImageUrl { get; set; }

        [DataType(DataType.MultilineText)]
        [FirestoreProperty]
        public string Description { get; set; }
    }
```

But I don't see a way to bind the `Id` field to Firestore's document Id, so I have to write code to propagate the id from the snapshot to my Book class.

For example:

```cs
            List<Book> bookList = new List<Book>();
            var snapshot = await _books.Offset(nextPageStart).Limit(pageSize)
                .GetSnapshotAsync();
            foreach (DocumentSnapshot docSnapshot in snapshot.Documents) 
            {
                var book = docSnapshot.ConvertTo<Book>();
                book.Id = docSnapshot.Id;
                bookList.Add(book);
            }
            return new BookList()
            {
                Books = bookList,
                NextPageToken = bookList.Count == pageSize ? 
                    (nextPageStart + pageSize).ToString() : null
            };
```

Instead, I wish I could write:
```cs
            List<Book> bookList = new List<Book>();
            var snapshot = await _books.Offset(nextPageStart).Limit(pageSize)
                .GetSnapshotAsync();
            return new BookList()
            {
                Books = snapshot.Documents.Select(doc => doc.ConvertTo<Book>()),
                NextPageToken = bookList.Count == pageSize ? 
                    (nextPageStart + pageSize).ToString() : null
            };
```

I imagine another property I could add to the Book class definition like:

```cs
    [FirestoreData]
    public class Book
    {
        [Key]
        [BindNever]
        [FirestoreId]
        public string Id { get; set; }
```

jskeet:
Fixes #3042

Note: unlike the Java attribute, this is only effective at the "root" type (i.e. the type we convert the document to). Propagating the document reference all the way down through nested data would be very invasive, for little practical benefit.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/AttributedIdAssigner.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/AttributedIdAssigner.cs
@@ -1,0 +1,120 @@
+﻿// Copyright 2019, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Reflection;
+using BclType = System.Type;
+
+namespace Google.Cloud.Firestore.Converters
+{
+    internal sealed class AttributedIdAssigner
+    {
+        private static readonly ConcurrentDictionary<BclType, AttributedIdAssigner> s_assigners =
+            new ConcurrentDictionary<BclType, AttributedIdAssigner>();
+
+        private readonly List<PropertyInfo> _idProperties;
+        private readonly List<PropertyInfo> _referenceProperties;
+
+        private AttributedIdAssigner(List<PropertyInfo> idProperties, List<PropertyInfo> referenceProperties)
+        {
+            _idProperties = idProperties;
+            _referenceProperties = referenceProperties;
+        }
+
+        private void AssignId(object value, DocumentReference reference)
+        {
+            // TODO: Create delegates instead of using reflection on every invocation?
+
+            foreach (var property in _idProperties)
+            {
+                property.SetValue(value, reference.Id);
+            }
+
+            foreach (var property in _referenceProperties)
+            {
+                property.SetValue(value, reference);
+            }
+        }
+
+        internal static void MaybeAssignId(object value, DocumentReference reference)
+        {
+            if (value == null)
+            {
+                return;
+            }
+            var assigner = s_assigners.GetOrAdd(value.GetType(), MaybeCreateAssigner);
+            assigner?.AssignId(value, reference);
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDocumentIdAttribute.cs
```diff
@@ -0,0 +1,37 @@
...
+    /// This attribute is ignored when serializing a document to Firestore.
+    /// This attribute must not be applied on a property which also has <see cref="FirestorePropertyAttribute"/>.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Property)]
+    public sealed class FirestoreDocumentIdAttribute : Attribute
+    {
+        /// <summary>
+        /// Creates an instance of the attribute.
+        /// </summary>
+        public FirestoreDocumentIdAttribute()
```
chrisdunelm: Why the explicit parameterless ctor?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/AttributedIdAssigner.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (value == null)
+            {
+                return;
+            }
+            var assigner = s_assigners.GetOrAdd(value.GetType(), MaybeCreateAssigner);
+            assigner?.AssignId(value, reference);
+        }
+
+        /// <summary>
+        /// Returns an "ID assigner" suitable
```
chrisdunelm: Is this sentence complete?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/AttributedIdAssigner.cs
```diff
@@ -0,0 +1,119 @@
...
+                return;
+            }
+            var assigner = s_assigners.GetOrAdd(value.GetType(), MaybeCreateAssigner);
+            assigner?.AssignId(value, reference);
+        }
+
+        /// <summary>
+        /// Returns an "ID assigner" suitable 
+        /// </summary>
+        /// <param name="type"></param>
```
chrisdunelm: xmldoc for the `type` param and return?

jskeet: Will do.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/AttributedIdAssigner.cs
jskeet: Whoops, will fix.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDocumentIdAttribute.cs
jskeet: It means we get documentation instead of an empty message.

PR #3160, opened 2019-06-20
Allow FirestoreDocumentId to be specified anywhere

jskeet:
Suggest reviewing one commit at a time


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/DeserializationContext.cs
```diff
@@ -24,15 +24,20 @@ internal sealed class DeserializationContext
         /// <summary>
         /// The database containing the document being deserialized. This is never null.
         /// </summary>
-        internal FirestoreDb Database { get; }
+        internal FirestoreDb Database => DocumentReference.Database;
+
+        /// <summary>
+        /// The document being deserialized.
```
chrisdunelm: nit: Add "This is never null"

jskeet: Thanks, will do.

PR #3177, opened 2019-07-01
Changes to API generation

jskeet:
- Python scripts to provide information from the API catalog for shell scripts
  - getapifield.py fetches a single field for a single package
  - listapis.py lists all the packages, with options to select a specific field, and/or to filter by presence or exact value of a field
- All package information is within the API catalog; generateapis.sh consults this
- generateapis.sh now allows you to specify which packages you want to generate
- apis.json is now strictly-compliant JSON, with a README.md to explain it
- Process documentation is updated

### DIFF
--- a/NEWAPI.md
+++ b/NEWAPI.md
@@ -10,6 +10,7 @@ Prerequisites:
 - `git`
 - `bash` (on Windows, the version that comes with Git for Windows)
 - `wget`
+- Python 3 on the path as `python`
 - Java 9+ (until the microgenerators are in use for everything)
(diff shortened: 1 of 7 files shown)


jskeet:
Note: running generateapis.sh detects no changes (as I've merged #3176 - previously it generated the same set of changes).

Next steps will be to get rid of the special-casing for IAM, OsLogin and Logging, and fold in microgeneration as well. But let's get this one done first :)

jskeet:
Side-note: we'll probably need to change the shell script to call "python3" instead of "python" when on Linux. We can put a $PYTHON variable in tool_versions.sh when we need to.

PR #3184, opened 2019-07-02
Allow generation script to handle different generators

jskeet:
While it's still relatively ugly, there's no less special-casing going on.


jskeet:
Note that the diff isn't terribly useful here, due to the amount of moving around :(

PR #3198, opened 2019-07-03
Update .NET Core SDK version

jskeet:
(And revert an earlier change.)

I believe 2.2.300 should be fine everywhere. It's relatively fiddly to install 2.2.100 on Linux from scratch.


jskeet:
Closing this, as I'm changing kokoro instead.

PR #3206, opened 2019-07-03
Fix the link to Cloud Tasks

flagbug:
Noticed that the link doesn't work, fixed it up


jskeet:
I know it seems odd to require a CLA just for a small fix like this - if you'd prefer not to sign it, I can ask our Open Source Program Office whether it's really required for a readme fix. But if you're happy to sign it, that would be simpler :)

Thank you either way!

jskeet:
Fixing this as part of another PR which is editing the README already.

PR #3215, opened 2019-07-06
Add .NET Core 3.0 target for diagnostics libraries

henkmollema:
Starting a draft PR with a stab to add .NET Core 3 target for the diagnostics libraries and to figure out whether it's feasible to get this out to NuGet since the libraries are still in beta. Currently we're unable to get diagnostics working for .NET Core 3.0 apps hosted on GCP due to some breaking changes in .NET Core 3.0 which affect the diagnostics libraries. It would be nice if we could start trying out .NET Core 3.0 apps with diagnostics on GCP since an RC is scheduled for July and GA in September ([source](https://github.com/dotnet/core/blob/master/roadmap.md)).

I'm trying to get .NET Core 3.0 builds running on Travis and AppVeyor. However, this might not be desirable for all the other projects in this repo. We should figure out what to do here.

What's in this PR:
- Cross-compile for .NET Core 3.0 and .NET Standard (both `Common` and `AspNetCore`)
- Upgrade ASP.NET Core packages to latest 2.1 versions (LTS) for the NET Standard 2.0 target since 2.0 is EOL
- Utilize `AddHttpContextAccessor` extension added in 2.1
- Removed use of internal API which were removed in ASP.NET Core 3.0, mainly logging and HTTP context stuff
- Added `#ifdefs` for the new ASP.NET Core 3.0 hosting interfaces
- Updated to latest versions of other dependencies
- Remove unnecessary packages and other MSBuild stuff

cc @jskeet

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests.csproj
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests.csproj
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.1;net461</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.0;netcoreapp2.1;net461</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp3.0;netcoreapp2.1</TargetFrameworks>
     <LangVersion>latest</LangVersion>
     <IsPackable>false</IsPackable>
     <AssemblyOriginatorKeyFile>../../GoogleApis.snk</AssemblyOriginatorKeyFile>
@@ -13,7 +13,7 @@
   <PropertyGroup>
     <DebugSymbols>true</DebugSymbols>
     <DebugType Condition="'$(TargetFramework)' == 'net461'">pdbonly</DebugType>
-    <DebugType Condition="'$(TargetFramework)' == 'netcoreapp2.1'">portable</DebugType>
+    <DebugType Condition="'$(TargetFramework)' == 'netcoreapp2.1' Or '$(TargetFramework)' == 'netcoreapp3.0'">portable</DebugType>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Google.Api.Gax.Testing" Version="2.9.0" />
@@ -30,10 +30,13 @@
     <Reference Condition="'$(TargetFramework)' == 'net452'" Include="Microsoft.CSharp" />
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
   </ItemGroup>
-  <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.0.0" />
-    <PackageReference Include="Microsoft.AspNetCore.Mvc.ViewFeatures" Version="2.0.0" />
-    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="2.0.0" />
+  <ItemGroup Condition="'$(TargetFramework)' != 'netcoreapp3.0'">
+    <PackageReference Include="Microsoft.AspNetCore.Mvc" Version="2.1.3" />
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.ViewFeatures" Version="2.1.3" />
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="2.1.1" />
     <PackageReference Include="Microsoft.Extensions.Http" Version="2.1.1" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.0'">
+    <PackageReference Include="Microsoft.AspNetCore.TestHost" Version="3.0.0-preview6.19307.2" />
   </ItemGroup>
 </Project>
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
@@ -256,7 +256,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
 
         public virtual void ConfigureServices(IServiceCollection services)
         {
-            services.TryAddSingleton<IHttpContextAccessor, HttpContextAccessor>();
+            services.AddHttpContextAccessor();
             services.AddGoogleTrace(options =>
             {
                 options.ProjectId = ProjectId;
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Trace/LabelsTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Trace/LabelsTest.cs
@@ -33,14 +33,14 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
         }
 
         [Fact]
-        public void FromDefaultHttpRequest()
+        public void FromHttpRequest()
         {
             var request = new DefaultHttpRequest(new DefaultHttpContext { TraceIdentifier = "trace-id" });
             request.ContentLength = 123;
             request.Host = new HostString("google.com");
             request.Method = "PUT";
 
-            var labels = Labels.FromDefaultHttpRequest(request);
+            var labels = Labels.FromHttpRequest(request);
             Assert.Equal(4, labels.Count);
             Assert.Equal("123", labels[LabelsCommon.HttpRequestSize]);
             Assert.Equal("google.com", labels[LabelsCommon.HttpHost]);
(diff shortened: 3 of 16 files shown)


jskeet:
Thanks Henk. I'm on vacation on Monday but I'll try to look at this on Tuesday. It would be really nice to get a GA of the diagnostics libraries out at some point, although there's also the aspect of logging and error reporting being handled by fluentd (which is what I'm using for NodaTime for example). Glad to have your contributions, whatever happens!

henkmollema:
Alright! Enjoy your vacation.

jskeet:
Taking a step back a bit - do you think it's definitely the right approach to have a single package that targets both ASP.NET Core 2 and ASP.NET Core 3? Perhaps we should split it into two packages instead? `#if` always makes me nervous - for example, I'd expect it to be feasible to run ASP.NET Core 2 but from .NET Core 3.0, and that could make things somewhat odd in this library. (We've seen that in the past with folks using ASP.NET Core 1 from desktop .NET, and we ended up with the wrong types being used.)

jskeet:
(Rebasing may reduce the number of diffs, by the way - I've recently updated a bunch of dependencies in general.)

henkmollema:
I'm not sold on multi-targeting yet, but I'm not sure whether splitting packages is the right approach either. Running ASP.NET Core 2.x on .NET Core 3.0 seems unlikely to me as ASP.NET Core is shipped as part of the shared/in-box framework. E.g. you get ASP.NET Core just by targeting `netcoreapp3.0` and using the Web SDK, there are no package references. But like you said, choosing the runtime packages based on the target framework has led to issues in the past and it might bite us again.

I agree about the `#if`'s. However, they're only used in a small utility class to add the environment name to the logging labels which happens to use an interface which is renamed in .NET Core 3.0. We might even choose to remove the class altogether as it's not used by default. It will break users who add it manually though.

Personally I lean towards the multi-targeting approach and guide users in the right direction via the docs. Splitting packages seems as much work (relatively) as I guess that means code sharing or duplication?

jskeet:
Yes, there'd definitely be a lot of code duplication with different packages. I'm happy with aiming towards multitargeting to start with, and changing course if necessary - so I'm very happy you've done a lot of the work here :)

henkmollema:
Great! I'll rebase this branch tomorrow.

Do you have an idea how to get the build working with .NET Core 3.0? I'm having issues with both Travis (installing a .NET Core 3.0 'snap' didn't seem to do the trick) and AppVeyor (the `Visual Studio 2019 Preview` contains an old preview version of .NET Core). Also, building for .NET Core 3 is probably not desirable (yet) for the complete repo.

jskeet:
In Noda Time we install the SDK as part of the build - when you've rebased your branch I may clone it and tinker with that, if you're happy for me to do so (and push a commit to it). It may end up being worth merging this work into a new branch until .NET Core 3 is actually released.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Trace/LabelsTest.cs
```diff
@@ -1,11 +1,11 @@
 ﻿// Copyright 2017 Google Inc. All Rights Reserved.
-// 
+//
```
jskeet: Please undo all these whitespace changes.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
```diff
@@ -256,7 +256,7 @@ public LoggerTestApplication()
 
         public virtual void ConfigureServices(IServiceCollection services)
         {
-            services.TryAddSingleton<IHttpContextAccessor, HttpContextAccessor>();
+            services.AddHttpContextAccessor();
```
jskeet: Is this a change that is specific to the .NET Core 3.0 support? It would be good to isolate the changes as much as possible - if there are parts that we can do separately, that would help.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
```
jskeet: These project files are "mostly auto-generated" based on apis.json. Once we've worked out what they need to look like, we'll need to change the generator and apis.json to make sure we can generate the right thing.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.csproj
```diff
@@ -2,8 +2,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <Version>3.0.0-beta13</Version>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.0;netstandard2.0</TargetFrameworks>
```
jskeet: It feels odd for a library to target .NET Core App rather than .NET Standard. Is that definitely necessary? I'd expect netstandard2.1 for .NET Core 3.0 support.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
```diff
@@ -200,7 +199,7 @@ private Struct CreateJsonPayload<TState>(EventId eventId, TState state, Exceptio
             while (currentLogScope != null)
             {
                 // Determine if the state of the scope are format params
-                if (currentLogScope.State is FormattedLogValues scopeFormatParams)
+                if (currentLogScope.State is IReadOnlyList<KeyValuePair<string, object>> scopeFormatParams)
```
jskeet: This looks like something that could be done in a separate PR - is that the case?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
henkmollema: This method was added (by me 😄) in ASP.NET Core 2.1 and we can use this now since I've upgraded the package dependencies to 2.1 (as that is the LTS release). I could do the 2.1 upgrade along with these changes in a separate PR though.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
henkmollema: The `FormattedLogValues` type has been made `internal` in 3.0. Figured this change could be applied to both the 2.x and 3.0 version.

henkmollema:
I've rebased and simplified the changeset a bit.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.csproj
jskeet: I know ASP.NET Core 3 will only run on .NET Core, but I'd still expect most packages to target netstandard2.1. I guess it depends on whether we're depending on anything that targets netcoreapp3.0 directly. Currently the PR depends on Microsoft.AspNetCore.App, but I don't know whether it's best to depend on that from *libraries* as opposed to from apps themselves.

henkmollema: I'm not sure if referencing Microsoft.AspNetCore.App is the right thing too. But there are no other runtime packages for ASP.NET Core which you can reference. Only the Microsoft.Extensions.* packages are published as netstandard2.0 packages for .NET Core, no the Microsoft.AspNetCore.* ones.

jskeet: Interesting - that package contains the reference assemblies, by the looks of it.

henkmollema: I doubt there is a NuGet package of `Microsoft.AspNetCore.App` anymore, the last one is preview3. My understanding is that they've added ASP.NET Core to the shared framework of .NET Core 3. And you can explicitly reference it by adding the `<FrameworkReference>` element.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
jskeet: Right - I've just checked, and ASP.NET Core 2.0 has been out of support by MS for over a year now, so yes, let's update to 2.1 in a separate PR. Are you happy to do that, or would you like me to? (You're in a better position to know which of these changes will work in 2.1, but I realize this is my job!)

henkmollema: I can do that in a separate PR. I might not have time until next week though, is that a OK?

jskeet: Fine by me - thanks in advance!

jskeet:
Hey Hank,

Just checking on the status of this. It looks like it's got quite a lot of the code that was in #3264... do you have the bandwidth to pare this down so we could merge it (into a separate branch)? Let me know if you'd like me to do that.

henkmollema:
Hi Jon. I'll either rebase this branch or start a new one and picking changes from this one. I hope I'll come around to it this week.

jskeet:
Hi Henk,

Given that quite a lot of water has gone under the bridge now, would it be worth closing this PR? I suspect it will make more sense to revisit this after we've taken our major version bump - in particular as you can't use Microsoft's gRPC implementation in ASP.NET Core alongside our current libraries (which is part of the point of the version bump).

henkmollema:
Hey Jon,

Seems fine to me. Most of the work in this PR has been merged through others PRs. Also running this library on .NET Core 3.0 is unblocked with the current .NET Standard 2.0 package. Let's reconsider this later indeed.

PR #3216, opened 2019-07-10
Huge dependency update

jskeet:
- Use GAX 2.9.0 by default
- Use protoc 3.8.0 and gRPC 1.22.0
- Update nunit etc
- Remove unnecessary manually-specified dependencies, and update those that still need to be manually-specified

There are three hand-written commits and three generated commits here (alternating) - only the hand-written changes need reviewing.

I'm half expecting a build break from this... definitely wait for green before reviewing!


chrisdunelm:
"Package Microsoft.CodeAnalysis.CSharp.Workspaces 3.1.0 is not compatible with netstandard1.3" :(

jskeet:
Added a few extra commits to try to fix things up.
May run all the integration tests after that before merging, just to be sure...

jskeet:
Travis is green - PTAL. Will probably merge then anyway, and let the overnight integration tests do their thing.

PR #3218, opened 2019-07-10
Release many of the things!

jskeet:
I'll wait until the repo-wide integration tests have passed before even merging this, as otherwise there's little chance of actually getting the releases through. (Few of these APIs have many integration tests, mind you.)

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -6,12 +6,14 @@
     "serviceYaml": "asset_v1.yaml",
     "productName": "Google Cloud Asset Inventory",
     "productUrl": "https://cloud.google.com/resource-manager/docs/cloud-asset-inventory/overview",
-    "version": "1.0.0-beta01",
+    "version": "1.0.0",
     "type": "grpc",
     "description": "Recommended Google client library to access the Google Cloud Asset Inventory API (v1).",
     "dependencies": {
+      "Google.Api.Gax.Grpc": "2.9.0",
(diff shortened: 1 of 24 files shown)


jskeet:
Integration tests passed yesterday (after big dependency changes), so I'm fairly confident that they won't make these releases fail. PTAL.

apis/apis.json
```diff
@@ -586,12 +614,14 @@
...
     "productUrl": "https://cloud.google.com/compute/docs/instances/managing-instance-access",
-    "version": "1.0.0-beta02",
+    "version": "1.0.0",
     "type": "grpc",
     "description": "Recommended Google client library to manage OS login configuration for Google account users.",
     "tags": [ "OsLogin", "Login" ],
     "dependencies": {
-      "Google.Cloud.OsLogin.Common": "project"
+      "Google.Api.Gax.Grpc": "2.9.0",
+      "Google.Cloud.OsLogin.Common": "project",
```
chrisdunelm: Should this "project" value be a specific version number in a GA release?

jskeet: It can't be at the moment, because Google.Cloud.OsLogin.Common hasn't been released yet. The release tagger checks that both of them are being released at the same time. After 1.0.0 has been released, we can then modify both OsLogin.V1 and OsLogin.V1Beta1 to refer to it.

PR #3250, opened 2019-07-21
Use print() function in both Python 2 and Python 3

cclauss:
Legacy __print__ statements are syntax errors in Python 3 but __print()__ function works as expected in both Python 2 and Python 3.


jskeet:
Thanks - I'm not sure whether we'll actually keep this emulator (it's very old and may well be stale) but it doesn't hurt to be ready for Python3.

PR #3261, opened 2019-07-24
Only write diagnostic log output once

### LINKED ISSUE #3258: Logger diagnostic output writes far too often
I've just configured my logger diagnostics to write to Console.Out like this:

```csharp
LoggerOptions.Create(loggerDiagnosticsOutput: Console.Out)
```

This has resulted in a morass of log entries. We should almost certainly only write once per log provider.

jskeet:
Fixes #3258.

(This is ready for review, but I won't merge it yet anyway - I want to try to get a bunch of issues fixed, then test them in a real application before merging, just to be on the safe side.)

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLoggerProvider.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLoggerProvider.cs
@@ -50,6 +50,15 @@ namespace Google.Cloud.Diagnostics.AspNetCore
             _logTarget = GaxPreconditions.CheckNotNull(logTarget, nameof(logTarget));
             _loggerOptions = GaxPreconditions.CheckNotNull(loggerOptions, nameof(loggerOptions));
             _serviceProvider = serviceProvider;
+
+            var writer = loggerOptions.LoggerDiagnosticsOutput;
+            if (writer != null)
+            {
+                // The log name is the ASP.NET Core log name, not the "/projects/xyz/logs/abc" log name in the resource.
+                // We don't currently use this in the diagnostics, but if we ever start to do so, SampleLogName seems
+                // like a reasonably clear example.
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLoggerProvider.cs
```diff
@@ -50,6 +50,12 @@ internal GoogleLoggerProvider(IConsumer<LogEntry> consumer, LogTarget logTarget,
             _logTarget = GaxPreconditions.CheckNotNull(logTarget, nameof(logTarget));
             _loggerOptions = GaxPreconditions.CheckNotNull(loggerOptions, nameof(loggerOptions));
             _serviceProvider = serviceProvider;
+
+            var writer = loggerOptions.LoggerDiagnosticsOutput;
+            if (writer != null)
+            {
+                ((GoogleLogger) CreateLogger("SampleLogName")).WriteDiagnostics(writer);
```
chrisdunelm: Is "SampleLogName" correct here?

jskeet: Yes. It's not currently used in the method anyway, but I figured it would be a reasonable value to end up being written to diagnostics if we changed things so it *was* used.
(This isn't the log name in the monitored resource - it's the ASP.NET Core log name. Very confusing, I know.)

Would a comment of basically the above be useful?

chrisdunelm: Yes, please do add a comment - it's very non-obvious!

jskeet: Done. PTAL.

PR #3262, opened 2019-07-24
Allow the UseGoogleDiagnostics method to supply options

jskeet:
Without this, when options are required for troubleshooting, the user has to change their configuration approach entirely.

@henkmollema If you have the time to review this as well, that would be much appreciated.

Note that this does break *binary* compatibility, but should be fine for *source* compatibility.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsWebHostTests.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsWebHostTests.cs
@@ -84,11 +84,12 @@ namespace Google.Cloud.Diagnostics.AspNetCore.IntegrationTests
                     { "build_id", "some-build-id" }
                 }
             };
-
+            // We won't be able to detect the right monitored resource, so specify it explicitly.
+            var loggerOptions = LoggerOptions.Create(monitoredResource: resource);
             var webHostBuilder = new WebHostBuilder()
                 .ConfigureServices(services => services.AddMvcCore())
                 .Configure(app => app.UseMvcWithDefaultRoute())
-                .UseGoogleDiagnostics(monitoredResource: resource);
+                .UseGoogleDiagnostics(TestEnvironment.GetTestProjectId(), EntryData.Service, EntryData.Version, loggerOptions);
 
             using (var server = new TestServer(webHostBuilder))
             using (var client = server.CreateClient())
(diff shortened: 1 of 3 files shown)


jskeet:
There's nothing to stop us adding that now if we want to... or even later on, if that's useful. But I've actually *reduced* the number of overloads :) As all of these are optional parameters, I don't think it'll cause a problem for users, who I expect to only provide a single option in most cases.

As a heads-up in case I hadn't mentioned it before, I expect this library to be relatively rarely used. Ideally, there should be a more flexible ASP.NET Core console logger so we can just format some JSON to be written to the console - avoiding all the project ID detection etc. That takes care of error reporting and logging, and the OpenTelemetry project can replace the tracing part. This library will only necessary when those options don't work out for whatever reason. Now that doesn't mean this library isn't important - but it does mean I'm keen on getting it into a better state so we can ship 3.0 soonish, rather than trying to make too many higher level changes.

jskeet:
Note that we could later add a pair of methods with required parameter accepting either a `DiagnosticsOptions` or a `Func<WebHostBuilderContext, DiagnosticsOptions>` if we want to - those would just call these methods, and that would still be source and binary compatible, if we're careful.

henkmollema:
I noticed your removed an overload :). What I meant is that perhaps one single method with an optional `DiagnosticsOptions` parameter would have been better. However two overloads are not that bad at all and, like you said, it's pretty easy to configure just one options now.

> As a heads-up in case I hadn't mentioned it before, I expect this library to be relatively rarely used. Ideally, there should be a more flexible ASP.NET Core console logger so we can just format some JSON to be written to the console - avoiding all the project ID detection etc.

Is it possible to utilize the structured logging (with parameters, log severity, etc.) using the console logger?

> Now that doesn't mean this library isn't important - but it does mean I'm keen on getting it into a better state so we can ship 3.0 soonish, rather than trying to make too many higher level changes.

Yeah it would be nice to finally get 3.0 out. Are you planning to release it before .NET Core 3.0 RTM lands?

jskeet:
> Is it possible to utilize the structured logging (with parameters, log severity, etc.) using the console logger?

If you write appropriate JSON to the console, Stackdriver picks it up automatically. That's what I do in nodatime.org now - it's really simple.

I filed [an issue](https://github.com/aspnet/AspNetCore/issues/12237) requesting better out-of-the-box support for this; I suspect we'd still need a Stackdriver-specific package to create the appropriate JSON, but that wouldn't need to depend on *any* of this code.

> Yeah it would be nice to finally get 3.0 out. Are you planning to release it before .NET Core 3.0 RTM lands?

I hope so, yes. We could then release 3.1 of this package to target ASP.NET Core 3.0 when that's released - does that work for you?

henkmollema:
> If you write appropriate JSON to the console, Stackdriver picks it up automatically. That's what I do in nodatime.org now - it's really simple.

I see, nice, didn't know that. 

> I hope so, yes. We could then release 3.1 of this package to target ASP.NET Core 3.0 when that's released - does that work for you?
Yes, that's great.

PR #3264, opened 2019-07-24
Update Diagnostics to ASP.NET Core 2.1

henkmollema:
@jskeet these are the 2.1 changes requested here https://github.com/googleapis/google-cloud-dotnet/pull/3215#discussion_r302497546.

Updates the ASP.NET Core runtime packages to the latest 2.1 LTS release as well as some other package updates.

PR #3265, opened 2019-07-24
Move test dependencies for ASP.NET Core diagnostics into apis.json

jskeet:
This should make #3264 simpler.

cc @henkmollema


henkmollema:
Thanks! Should the separated dependencies in `Google.Cloud.Diagnostics.AspNetCore.csproj` be added to apis.json as well?

https://github.com/googleapis/google-cloud-dotnet/blob/4bf62d0c32e94798e768111105588a1e9d68ca45/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.csproj#L34-L40

jskeet:
Hadn't spotted those - yes, will do that now. It'll make it clearer when reviewing apis.json.

(Your PR will need to update those, then run generateprojects.sh to propagate the change.)

jskeet:
Done in a second commit.

henkmollema:
> (Your PR will need to update those, then run generateprojects.sh to propagate the change.)

Do I need to do this if I rebase my branch after this PR is merged?

jskeet:
> Do I need to do this if I rebase my branch after this PR is merged?

Yes, because you're going to want to change some of the version numbers (I'd expect, anyway). Any dependencies which are *just* moved by your current PR should be fine after the rebase, but I'd assumed all these 2.0.x will want to be 2.1.x.

henkmollema:
Ah I see 👍

PR #3315, opened 2019-07-24
Add validation that we haven't broken apis.json

jskeet:


jskeet:
First commit adds a check in Travis (which currently goes red).
Second commit fixes apis.json so it goes green.

jskeet:
I'm going to merge this without review, so I can rerun autosynth and clear out all the issues.

Issue #3259, opened 2019-07-23
support for UseGoogleDiagnostics with CreateHostBuilder


WellspringCS:
Thanks for stopping by to let us know something could be better!

 **Is your feature request related to a problem? Please describe.**
as noted here https://docs.microsoft.com/en-us/aspnet/core/migration/22-to-30?view=aspnetcore-2.2&tabs=visual-studio, `CreateHostBuilder` is the go-forward recommended approach in Program.cs, with `CreateWebHostBuilder` being deprecated.

I am disappointed to find what appears to be no way to call `UseGoogleDiagnostics` via `CreateHostBuilder`.

 **Describe the solution you'd like**
Support for ASP.NET Core 3.0 recommended design next week instead of next year. (Can't hurt for me to ask...)

 **Describe alternatives you've considered**
I thought about MS DOS 4.0 but cannot find the floppies.

 **Additional context**
I'll send cute emoticons when my dreams are granted to me by the Google fairies.

jskeet:
"Next week" is asking a bit much, but I'll look into it. There are several more urgent aspects of the diagnostics library that I want to address first, but I'm all for supporting the 3.0 way of doing things - *as well as* the old way. (I don't see any reason to break code that's already working.)

jskeet:
I prototyped this earlier today, and I suspect the production code will actually be trivial - but I got stuck on the testing part.
I'm going to have another go when all our other changes have landed, but it *may* be within a week after all...

WellspringCS:
John, I'd read that most of the people benefiting from your help are relative newcomers (to Stack Exchange, admittedly). So either that makes me a noob or you're not living up to your billing! (My silly way of saying Thank You!)

Thank you, indeed.

In other news... I have not been idle. I decided to try to implement the API using the legacy `CreateWebHostBuilder`. (I really wasn't expecting you to turn around my request so quickly!)

Bad news is, I got weird errors.

So I went back a step and, using [http://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.Diagnostics.AspNetCore/index.html](url) as guide, I was able to implement Stackdriver logging with netcoreapp2.2 -- happy and no errors on first swing. No exhaustive effort, not exhaustively tested, to be sure... but something rudimentary did work without errors.

However in my full 3.0 project (which includes GraphQL and IdentityServer4 authentication, etc. etc.) that same code produces errors.

[...]

PR #3318, opened 2019-07-25
Remove internal type usage to unblock .NET Core 3

henkmollema:
Unblocks .NET Core 3.0 usage as requested in #3259 without targetting .NET Core 3.0 yet, by removing the use of deleted/internal types in .NET Core 3.0.

I did a small test and got it working on a .NET Core 3.0 app with the following host builder setup:

```cs
public static IHostBuilder CreateHostBuilder(string[] args) =>
    Host.CreateDefaultBuilder(args)
        .ConfigureWebHostDefaults(webBuilder =>
        {
            webBuilder.UseGoogleDiagnostics();
            webBuilder.UseStartup<Startup>();
        });
```

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Trace/LabelsTest.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.Tests/Trace/LabelsTest.cs
@@ -40,7 +40,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
             request.Host = new HostString("google.com");
             request.Method = "PUT";
 
-            var labels = Labels.FromDefaultHttpRequest(request);
+            var labels = Labels.FromHttpRequest(request);
             Assert.Equal(4, labels.Count);
             Assert.Equal("123", labels[LabelsCommon.HttpRequestSize]);
             Assert.Equal("google.com", labels[LabelsCommon.HttpHost]);
@@ -54,7 +54,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore.Tests
             var response = new DefaultHttpResponse(new DefaultHttpContext());
             response.StatusCode = 404;
 
-            var labels = Labels.FromDefaultHttpResponse(response);
+            var labels = Labels.FromHttpResponse(response);
             Assert.Equal(1, labels.Count);
             Assert.Equal("404", labels[LabelsCommon.HttpStatusCode]);
         }
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
@@ -19,7 +19,6 @@ using Google.Protobuf.WellKnownTypes;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Internal;
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -199,7 +198,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore
             while (currentLogScope != null)
             {
                 // Determine if the state of the scope are format params
-                if (currentLogScope.State is FormattedLogValues scopeFormatParams)
+                if (currentLogScope.State is IEnumerable<KeyValuePair<string, object>> scopeFormatParams)
                 {
                     scopeParamsList.Add(CreateStructValue(scopeFormatParams));
                 }
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/Labels.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Trace/Labels.cs
@@ -14,7 +14,6 @@
 
 using Google.Api.Gax;
 using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.Http.Internal;
 using System.Collections.Generic;
 using System.Linq;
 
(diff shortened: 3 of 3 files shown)


jskeet:
Great, thanks!

henkmollema:
Is there another (beta) release plannend anytime soon? This would unblock a few of our projects as well.

jskeet:
I've got a few more issues to address today/tomorrow, then I'll do a beta release - probably using a beta of GAX that unblocks Cloud Run, too.

henkmollema:
Perfect, thank you 👍

jskeet:
Just as a caveat, if I don't get a release done today, it'll probably be Monday - I'm out on Friday morning, and I generally don't like releasing on a Friday anyway, in terms of risk. Is Monday okay for you, or would you like me to prioritize releasing today?

henkmollema:
Monday is fine!

jskeet:
3.0.0-beta14 is now out - please let me know if you run into any problems.

jskeet:
Hmm... have just run into a weird error with UseGoogleDiagnostics where I'm providing a project ID and it doesn't think I am. I'm going to investigate briefly, but I may need to unlist 3.0.0-beta14, hopefully creating a beta15 later today. Sorry!

jskeet:
Okay, the problem isn't as bad as I thought it was - it's just `ErrorReportingOptions.Create`. It would be nice to address that, but I'm not too worried right now, so beta14 can stay :)

WellspringCS:
I am off to the races! Thank you!!

In recognition that your day is winding down, I'll post this note of thanks and continue testing/working with it. I see it working and what remains may be nothing more than my learning curve (having the dickens of a time getting a JSON object into that jsonPayload field, for example).

jskeet:
@WellspringCS: I've got several hours left of working time :) It may be worth filing a new issue with how far you've got, and what's going wrong at the moment. If I don't get to look today, I can at least look tomorrow.

WellspringCS:
I hesitate to file one just yet because I'm not sure how to give you a repeatable problem. But here's the sort of thing that's dogging me right now--this one I happened to catch perfectly in the log...

image

I started up the API, hit the endpoint, got confused by what I saw. Hit it again. Second time I saw what I'd expected the first time around. A "before" and an "after" -- but for some reason on the first pass the log does not report what (deterministically) was surely logged in both cases. Code involved is in the ASP.NET Core 3.0 configure() pipeline....

```
  app.Use(async (context, next) =>
            {
                logger.LogInformation($"authenticated! {context.Request.Path} (before)");
                await next();
                logger.LogInformation($"authenticated! {context.Request.Path} (after)");
            });
```
And as you can see, another logged item showed up (UserService) on the second pass. Also deterministic, so should have been present on the first pass. I hit the endpoint via Postman the same way both times.

But that's not much to go on for you, I'd guess. I will keep digging.

jskeet:
Understood - thanks for your patience as an early adopter of the combination :)

WellspringCS:
Thank you one last time, @jskeet and @henkmollema -- all is working well, from what I am seeing.

If I were to point out some difficulties along the way, it'd be:

(a) documentation
After playing around I came to realize that this catchall `webBuilder.UseGoogleDiagnostics();` means that other bits of code in Configure() and ConfigureServices() (e.g. `services.AddLogging()`, `services.AddGoogleTrace()`, `app.UseGoogleExceptionLogging()`, `loggerFactory.AddGoogle(Configuration["GoogleProjectId"])` -- I'm just looking through my code for things that ended up commented out... )  are not required.

(b) nuances of getting stuff into the Errors section of StackDriver

The following code shows what I had to do if I wanted to take, for example, an error passed in via a SPA and get it into the Errors, as opposed to the Log.

```
if (logSev >= LogSeverity.Error) {
    try {
        // await exceptionLogger.LogAsync(new Exception(...) doesn't work!!
        // Have to **THROW** ...
        throw (new Exception(
            $"Browser threw error {myLogInfo.TextPayload} with tags {myLogInfo.JsonTags}.")
            { HResult = (int) logSev });
    } catch (Exception ex) {
        await exceptionLogger.LogAsync(ex); // this posts error where it belongs
    }
    // and then we proceed on our merry way.
    } else { // normal logging...
        LogName logName = new LogName(configuration["GoogleProjectId"], "Browser");
        LogEntry logEntry = new LogEntry {
            LogName = logName.ToString(),
            Severity = logSev,
            JsonPayload = jsonPayload //,
            //TextPayload = myLogInfo.TextPayload
            // If I pass TextPayload, the JsonPayload value is ignored!
       };
       MonitoredResource resource = new MonitoredResource { Type = "global" };
       await client.WriteLogEntriesAsync(LogNameOneof.From(logName)
            , resource, entryLabels,
           new [] { logEntry }, _retryAWhile);
    }
```
and

(c) documentation of jsonPayload vs TextPayload
I searched high and low but did not find much info on these two items. I never figured out how to load jsonPayload with variable JSON data. Then I discovered (as my comments above attest) that I could provide one or the other, but not both. I ended up landing on jsonPayload, but I was only able to parse out a list of key value pairs via the entryLabels in the WriteLogEntriesAsync(). That worked, so I called it good and gave up on doing the same for jsonPayload.

Anyway, I'm new to everything here, so errors due to ignorance I'll own in advance.

Thanks much, though -- I'm very excited to put StackDriver to use in this project I'm working on, and many others to come.

jskeet:
a) We do explicitly say in the [documentation](http://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.Diagnostics.AspNetCore/) that: 

> The easiest way to initialize Google Diagnostics services is using the UseGoogleDiagnostics extentension method on IWebHostBuilder. This configures Logging, Tracing and Error Reporting middleware.

We could call out that everything underneath that is only if you don't want to use `UseGoogleDiagnostics`, I guess...

b) It's not clear to me why you're writing the log entries directly at all. The point of this package is to do all that for you. But the error reporter *does* use logging to report the error, as that's the preferred way of reporting errors in Stackdriver

c) Yes, JsonPayload and TextPayload are alternatives. You should almost always use JsonPayload.

WellspringCS:
> We could call out that everything underneath that is only if you don't want to use `UseGoogleDiagnostics`, I guess...

Yeah, it was totally my ignorance. I realized late in the game that this feature really did replace all the rest. My documentation suggestion on that front is there "for the next noob" :)

> It's not clear to me why you're writing the log entries directly at all. 
For the SPA calling in with intent to simply log something, was my thinking, but your comment has me pondering whether in some fashion I have missed the forest for the trees. [Goes away to re-think life decisions...]

cheers

PR #3319, opened 2019-07-25
Remove service name and version requirement for error reporting

### LINKED ISSUE #3263: Allow defaulting for service name and version in diagnostics
One common mistake when configuring diagnostics is to not specify a service name and service version. Error reporting requires these to be non-null. We try to detect them from the platform, but they're only provided by GAE.
We should default to a reasonable value if neither the user nor the platform provides them.

jskeet:
It's fine just not to provide them when logging: the error reporting UI works with no problems, etc.

Fixes #3263.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ProjectTest.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ProjectTest.cs
@@ -24,7 +24,7 @@ namespace Google.Cloud.Diagnostics.Common.Tests
         public void GetAndCheckProjectId()
         {
             var projectId = "pid";
-            var resourceProjectId = "pid";
+            var resourceProjectId = "resource-pid";
             var monitoredResource = new MonitoredResource
             {
                 Type = "some-type",
@@ -57,7 +57,7 @@ namespace Google.Cloud.Diagnostics.Common.Tests
         public void GetAndCheckServiceName()
         {
             var serviceName = "my-app";
-            var resourceServiceName = "my-app";
+            var resourceServiceName = "resource-my-app";
             var monitoredResource = new MonitoredResource
             {
                 Type = "some-type",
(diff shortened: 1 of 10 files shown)


jskeet:
(I know this is an area you're not terribly familiar with, Chris - fortunately it's just the same "stop checking" again and again...)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ProjectTest.cs
```diff
@@ -86,6 +86,38 @@ public void GetAndCheckServiceName()
             Assert.Equal("No Google App Engine service name was passed in or detected.", ex.Message);
         }
 
+        [Fact]
+        public void GetServiceName()
+        {
+            var serviceName = "my-app";
+            var resourceServiceName = "my-app";
```
chrisdunelm: These should not be the same string ;)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ProjectTest.cs
```diff
@@ -119,6 +151,38 @@ public void GetAndCheckServiceVersion()
             Assert.Equal("No Google App Engine service version was passed in or detected.", ex.Message);
         }
 
+        [Fact]
+        public void GetServiceVersion()
+        {
+            var serviceVersion = "1.0.0";
```
chrisdunelm: Neither should these.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/ProjectTest.cs
jskeet: This was c-n-p, so I'm sure it's the same everywhere - will fix them all.

PR #3323, opened 2019-07-25
Update ASP.NET Core diagnostics documentation

### LINKED ISSUE #3256: Diagnostic docs improvements
Our ASP.NET Core docs have a section for "(ASP.NET Core 2.0+)" - but that's all we support now, so we should remove or at least edit the earlier section.

jskeet:
Fixes #3256 (as far as we need to).


jskeet:
Thanks for the LGTM - I'm just running the snippet tests locally to check they still pass before merging.

jskeet:
Hmm... the LogsInformation test isn't working, and I don't know why. Will investigate.

jskeet:
Logs_Information2 is broken too... at least when running on my machine. Will see whether they were already broken before, which could be a different matter... (Will do this all tomorrow though.)

jskeet:
The test passed this morning... I was messing with my configuration quite a bit yesterday, so I suspect I may have been in the wrong window without the right credentials. Merging.

PR #3326, opened 2019-07-26
Release diagnostics packages 3.0.0-beta14

jskeet:
- Target ASP.NET Core 2.1
- Remove uses of types not present in ASP.NET Core 3.0
- Make service name and service version optional when configuring
  error reporting
- Add options (for each diagnostic service) as optional parameters
  in UseGoogleDiagnostics
- Modify the diagnostic output for logging to only write once


jskeet:
While I normally don't like releasing on a Friday, releasing *first thing* on a Friday isn't so bad - and I'm around all weekend in case anything goes wrong and we need to roll back.

jskeet:
Integration tests failed overnight - I'm going to try to work out why before we do the release, so this will probably be Monday after all.

jskeet:
Okay, now that that bug has been fixed (when #3330 is merged) and as it didn't involve a *production* change, I'm still happy to get this out today.

jskeet:
(Except that a Diagnostics.Common test that's passing locally is failing on Kokoro. So I guess it's not happening after all...)

PR #3331, opened 2019-07-26
Improve test diagnostics

jskeet:
I still don't know why this is failing on Kokoro (and only Kokoro) but this should help us find out. (And one semi-bug-fix.)


jskeet:
Will merge, then run Kokoro again to try to work out what's wrong.

PR #3332, opened 2019-07-26
Change predicate into multiple assertions

jskeet:
This actually *fixes the test* (according to Kokoro) but I don't understand why. Either way, it's better.


jskeet:
Okay, it doesn't actually fix the test - I accidentally ran on Unix.

This only fails on Windows, and it's because we're now not getting a line number. I don't know why. (And it's fine locally.)

jskeet:
(Merged as it's better code anyway, but still trying to work out what's changed.)

PR #3333, opened 2019-07-26
Reinstate property specifying DebugType in tests

jskeet:
Without this, the error reporting test fails in Windows


jskeet:
Will run this in Kokoro before merging.

PR #3334, opened 2019-07-29
FirestoreDb-registered custom converters

### LINKED ISSUE #3255: FireStore: Register converters on client
Currently the FireStore converters are stored inside of a static cache that can only be accessed internally by the library. Can we change it so that the cache is instead on a client and allow the user add converters to this client cache?

My reasoning for this is that currently the only way to add converters for FireStore is by adding attributes to the POCO objects, which I could prefer to avoid. One, it requires me to reference FireStore in a part of our project where I don't feel it is required. Secondly, it may make it difficult when trying to use POCO's from an external library where one does not have access to the code base to add FireStore attributes.

By putting the converter cache on the client and allowing the public to add converters opens up the possibility of writing converters without having to rely on the FireStore attributes.

jskeet:
This PR will eventually fix #3255.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Benchmarks/ValueSerializerBenchmark.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Benchmarks/ValueSerializerBenchmark.cs
@@ -20,36 +20,38 @@ namespace Google.Cloud.Firestore.Benchmarks
 {
     public class ValueSerializerBenchmark
     {
+        private static readonly SerializationContext _serializationContext = new SerializationContext(null);
+
         [Benchmark]
         public Dictionary<string, Value> SerializeMap_Attributed() =>
-            ValueSerializer.SerializeMap(SampleData.Attributed);
+            ValueSerializer.SerializeMap(_serializationContext, SampleData.Attributed);
 
         [Benchmark]
         public Value Serialize_Attributed() =>
-            ValueSerializer.Serialize(SampleData.Attributed);
+            ValueSerializer.Serialize(_serializationContext, SampleData.Attributed);
 
         [Benchmark]
         public Dictionary<string, Value> SerializeMap_Anonymous() =>
-            ValueSerializer.SerializeMap(SampleData.Anonymous);
+            ValueSerializer.SerializeMap(_serializationContext, SampleData.Anonymous);
 
         [Benchmark]
         public Value Serialize_Anonymous() =>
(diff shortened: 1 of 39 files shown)


jskeet:
I'm now talking about this with the Firestore team, and the more I've written about it, the more I'd really quite like it in. Thoughts on overall API design and feature usefulness are welcome as well as the implementation details.

jskeet:
Firestore team is now happy, so we're ready for review. I'll need to update the docs when it's in, but I might wait until I've finished another couple of Firestore serialization features first, and update the docs in one go.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Benchmarks/ValueSerializerBenchmark.cs
```diff
@@ -20,36 +20,38 @@ namespace Google.Cloud.Firestore.Benchmarks
 {
     public class ValueSerializerBenchmark
     {
+        private static readonly SerializationContext _serializationContext = new SerializationContext(null);
```
chrisdunelm: `= SerializationContext.Default` ?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/SerializationContext.cs
```diff
@@ -24,11 +25,22 @@ namespace Google.Cloud.Firestore
     /// </summary>
     internal sealed class SerializationContext
     {
+        // Important note: if this ever contains more state than just the converters,
+        // e.g. a reference to the Database, then WithWarningLogger will need to be adjusted accordingly.
+        
+        private readonly IReadOnlyDictionary<BclType, IFirestoreInternalConverter> _customConverters;
```
chrisdunelm: nit: place this *after* the static `Default` property?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Benchmarks/ValueSerializerBenchmark.cs
jskeet: Good idea. I probably hadn't created SerializationContext.Default when I made that change :)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/SerializationContext.cs
jskeet: Ooh yes, definitely.

PR #3340, opened 2019-07-31
Make generateapis.sh aware of SYNTHTOOL_GOOGLEAPIS

### LINKED ISSUE #3339: Doesn't support env var --> SYNTHTOOL_GOOGLEAPIS (local googleapis)
It doesn't support $SYNTHTOOL_GOOGLEAPIS (env var that points to local googleapis/googleapis).  Other languages supports $SYNTHTOOL_GOOGLEAPIS, and I thnk it might be good idea to add this as option and keep it consistent throughout all languages.

Add an option to read $SYNTHTOOL_GOOGLEAPIS (points to local googleapis) if that env variable exists.

jskeet:
Expectations:
- SYNTHTOOL_GOOLEAPIS is either a path relative to the google-cloud-dotnet root, or an absolute path, without spaces
- Other generator code (e.g. pre/post-build fix-ups) doesn't need to find the protos

The second expectation is known to be violated in the Google.LongRunning package, but that's probably okay for now.

This does as much as we're likely to want to do to fix #3339.

### DIFF
--- a/generateapis.sh
+++ b/generateapis.sh
@@ -20,6 +20,12 @@ declare -r CORE_PROTOS_ROOT=$PROTOBUF_TOOLS_ROOT/tools
 # - Java 9
 
 OUTDIR=tmp
+if [[ "$SYNTHTOOL_GOOGLEAPIS" != "" ]]
+then
+  declare -r GOOGLEAPIS="$SYNTHTOOL_GOOGLEAPIS"
+else
+  declare -r GOOGLEAPIS=googleapis
+fi
 
 fetch_github_repos() {
   if [ -d "gapic-generator" ]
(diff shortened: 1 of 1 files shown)


generateapis.sh
```diff
@@ -40,21 +46,26 @@ fetch_github_repos() {
     git clone --recursive https://github.com/googleapis/gapic-generator-csharp
   fi
 
-  if [ -d "googleapis" ]
+  if [[ "$SYNTHTOOL_GOOGLEAPIS" != "" ]]
```
chrisdunelm: I'm not very familiar with the context of the PR, but I was expecting this to simply replace "googleapis" with $GOOGLEAPIS ?

chrisdunelm: OK, so shouldn't it be `==`, not `!=`?

And what's going on - you're meant to be on vacation??

jskeet: Was just idly checking emails, but won't look at detail of PR now :)

jskeet: Whoops - yes, that should indeed be ==. Fixed, and removed the echo lower down.

PR #3344, opened 2019-08-06
Generate csproj files for samples

hzyi-google:
In `tools/Google.Cloud.Tools.ProjectGenerator/Program.cs`, when a `*.Samples` directory is detected, generate a `.csproj` files in this directory for samples.

The project file is very similar to that of the snippets, with one more dependency on a public package `CommandLineParser`.

We don't have any published samples yet (there are no `Google.Cloud.{Api}.{Version}.Samples` yet), but being able to generate a project file for future samples makes testing much easier for us.

### DIFF
--- a/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
+++ b/tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
@@ -90,6 +90,11 @@ namespace Google.Cloud.Tools.ProjectGenerator
             { SourceLinkPackage, "2.8.3" }
         };
 
+        private static readonly Dictionary<string, string> CommonSampleDependencies = new Dictionary<string, string>
+        {
+            {"CommandLineParser", "2.6.0"}
+        };
+
         private const string CompatibilityAnalyzer = "Microsoft.DotNet.Analyzers.Compatibility";
         private const string ConfigureAwaitAnalyzer = "ConfigureAwaitChecker.Analyzer";
         private const string SourceLinkPackage = "SourceLink.Create.CommandLine";
@@ -125,7 +130,7 @@ namespace Google.Cloud.Tools.ProjectGenerator
                 }
                 foreach (var api in apis)
                 {
-                    // GenerateSolutionFiles(Path.Combine(root, "apis", api.Id), api);
+                    GenerateSolutionFiles(Path.Combine(root, "apis", api.Id), api);
                 }
                 foreach (var api in apis)
                 {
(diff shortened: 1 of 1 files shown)


hzyi-google:
PTAL @jskeet @chrisdunelm 
cc/ @googleapis/samplegen @ munkhuushmgl

jskeet:
I'm on vacation until Wednesday, but I can look on Thursday. I think I want to know more context about the samples though - we already have the snippets projects and a whole separate samples repo. Do we definitely need a third source? If you could point me to any doc internally, that would be great.

hzyi-google:
Hi @jskeet - no worries. Having this change live on a branch already unblocks us for the moment. Also, we are working on docs! But I'll explain things a bit here.

_(also summon @beccasaurus who would explain this better than I do)_

The snippets projects, the `dotnet-docs-samples` repo and this samples projects serve different purposes:

- the snippets projects: 

  - these seem to be the equivalent of the in-code documentation samples (the samplegen team refers to these as **"in-code samples"**) of other languages (like Java or Node), and they serve as the source for generating [language-specific documentation](http://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.Asset.V1/api/Google.Cloud.Asset.V1.AssetServiceClient.html) or a reference for people working with the client library source code.

- the dotnet-docs-samples: 

  - These are manually written, standalone runnable samples, and they are used to backup [examples on cloudsite](https://cloud.google.com/vision/docs/libraries#client-libraries-usage-csharp) under the `How-to Guides` section (the samplegen team refer to these as **"standalone samples"**).

- the samples projects that we want to add to this repo: 

  - They are the sample standalone samples as those in `dotnet-docs-samples`, but rather than being manually written, they are auto-generated. And because they are generated, making them live closer to the generated client libraries instead of in a separate repo will make processes like testing, generation and publishing much easier.

Please let me know if any part of these is confusing. I'll link the doc here too when ready.

tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
```diff
@@ -90,6 +90,11 @@ public class Program
             { SourceLinkPackage, "2.8.3" }
         };
 
+        private static readonly Dictionary<string, string> CommonSampleDependencies = new Dictionary<string, string>
+        {
+            {"CommandLineParser", "2.6.0"}
```
jskeet: Nit: please add spaces after the `{` and before `}` as per the earlier code.

hzyi-google:
Thank you!

PR #3359, opened 2019-08-12
Prototype of changes for IAM

jskeet:


apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2.Tests/IamVersionTest.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    public class IamVersionTest
+    {
+        [Theory]
+        [InlineData(null, 1, 1)]
+        [InlineData(null, null, null)]
+        [InlineData(3, 1, 1)]
+        [InlineData(3, null, 3)]
+        [InlineData(1, 3, 3)]
+        [InlineData(1, null, 1)]
```
frankyn: settingsVersion - set by the client library at the point of release.
callerRequestVersion - The version a user provides when performing a GetIamPolicy Request and meant to override the settingsVersion if provided.
expectedRpcVersion - Is the version the client is sending to the service.

Could you please confirm these definitions?

apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClientPartial.cs
```diff
@@ -0,0 +1,55 @@
...
+using Google.Cloud.Iam.V1;
+
+namespace Google.Cloud.Bigtable.Admin.V2
+{
+    public partial class BigtableInstanceAdminSettings
+    {
+        /// <summary>
+        /// The version of IAM policies that the client will indicate that it supports in RPCs.
+        /// This defaults to a null value, meaning that requests will not be modified at all.
+        /// If a value is already present in a request, it is not overwritten.
```
frankyn: How does this get flipped between `null`, `1` or `3` version values?

apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClientPartial.cs
```diff
@@ -0,0 +1,55 @@
...
+            supportedIamPolicyVersion = effectiveSettings.SupportedIamPolicyVersion;
+        }
+
+        partial void Modify_GetIamPolicyRequest(ref GetIamPolicyRequest request, ref CallSettings settings)
+        {
+            if (supportedIamPolicyVersion == null)
+            {
+                return;
+            }
+            // TODO: Set the appropriate field
```
frankyn: How is this used?

apis/Google.Cloud.Iam.V1/Google.Cloud.Iam.V1.Tests/PolicyTest.cs
```diff
@@ -154,5 +155,19 @@ public void RemoveRoleMember_NoRole()
             Assert.False(policy.RemoveRoleMember("target", "d"));
             Assert.Equal(expected, policy);
         }
+
+        [Fact]
+        public void AddRoleMember_LaterVersion()
+        {
+            var policy = new Policy { Version = 2 };
```
frankyn: Please update to `Version = 3`

apis/Google.Cloud.Iam.V1/Google.Cloud.Iam.V1/PolicyPartial.cs
```diff
@@ -77,5 +89,8 @@ public bool RemoveRoleMember(string role, string member)
         }
 
         private Binding FindRole(string role) => Bindings.FirstOrDefault(binding => binding.Role == role);
+
+        private void ValidatePolicyVersion() =>
+            GaxPreconditions.CheckState(Version <= 1, "Helper methods cannot be invoked on policies with version {0}. For more information, see FIXME.", Version);
```
frankyn: How is `Version` set in this case?

apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2.Tests/IamVersionTest.cs
jskeet: Not quite. `settingsVersion` is taken from the settings used to create the instance of the client. The *default* is set at the point of release, but the user can specify settings themselves - just as we do in this test. This allows them to create a client which will *always* use version 3, for example (modulo per-request settings)... whereas the `callerRequestVersion` is specified just for a *single* request.

apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClientPartial.cs
jskeet: It's a property - a user just sets it on an instance: `settings.SupportedIamPolicyVersion = 3;`

apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClientPartial.cs
jskeet: Do you mean what calls it? It's a partial method implementation. We generate partial methods for all RPCs, and they can be implemented within partial classes like this.

TL;DR: It's C# magic ;)

apis/Google.Cloud.Iam.V1/Google.Cloud.Iam.V1.Tests/PolicyTest.cs
jskeet: I deliberately left this as 2, as we're testing the `<=` behavior. It doesn't matter that we'll never *see* a version 2 in reality... it makes the test unambiguous. (Basically, when I have comparisons to test, I try to test the closest values to the boundary in each direction.)

apis/Google.Cloud.Iam.V1/Google.Cloud.Iam.V1/PolicyPartial.cs
jskeet: This is just a partial class for the proto - so it's "whatever the proto property has been set to". That could be in client side code for whatever reason, or a policy resource returned by a server.

apis/Google.Cloud.Bigtable.Admin.V2/Google.Cloud.Bigtable.Admin.V2/BigtableInstanceAdminClientPartial.cs
jskeet: If they're using IAM v1, yes. But I'm not expecting that to be the common way of accessing IAM by that time.

broady: Fair point!

jskeet:
@jkwlui @frankyn @broady I've now updated this with the complete implementation.

Notes:

- We always clone a request if we're going to modify it (and check whether or not we're going to need to before doing so)
- We now handle SetIamPolicy as well as GetIamPolicy; the two differ in terms of the logic for setting the version number. For "get", we respect any manually-specified value in the request; we can't really do that for "set" because an operation of "fetch policy, add condition, set policy" is very likely to end up with a policy version of 1 in the proto, but it will need to be 3.

apis/Google.Cloud.Iam.V1/Google.Cloud.Iam.V1/PolicyPartial.cs
```diff
@@ -26,14 +26,20 @@ public partial class Policy
...
         /// not already exist, it is created.
+        /// This method will fail with an <see cref="InvalidOperationException"/>
+        /// if it is called on a Policy with a <see cref="Version"/> greater than 1,
+        /// as that indicates a more complicated policy than this method is prepared
+        /// to handle. Changes to such policies must be made manually.
         /// </summary>
         /// <param name="role">The role to add the member to. Must not be null or empty.</param>
         /// <param name="member">The member to add to the role. Must not be null or empty.</param>
         /// <returns><c>true</c> if the policy was changed; <c>false</c> if the member already existed in the role.</returns>
+        /// <exception cref="InvalidOperationException">The <see cref="Version"/> is greater than 1.</exception>
```
broady: From my reading, this also only happens if SupportedIamPolicyVersion is set to 3, right?
That is, without any code modification (SupportedIamPolicyVersion=3), existing programs will not get an exception *here*, but when they *fetch* a policy with Version=3 and SupportedIamPolicyVersion=default.

Though I think it's probably safer to make the helpers invalid/throw exception when SupportedIamPolicyVersion=3

jskeet: Yes, this shouldn't happen unless you request a V3 policy explicitly.

But:

- You could request a V3 policy explicitly without realizing that it effectively invalidates the use of these methods
- There could be multiple teams in the same company, where one team passes a policy that it's fetched with v3 support to code written by a team that hasn't even heard of conditions

So as you say, it's safer to throw here as well. Very much by design.

broady: I think it's also possible to:

* Request with SupportedIamPolicyVersion=3
* Policy returned by server is Version=1
* Modify bindings (e.g., add a binding with a condition)
* Use the v1-only helpers

No exception is thrown because the version in the policy is only set to 3 in ModifyRequest

Yes, quite an edge case.

jskeet: > But would be prevented by setting Policy=3 in GetIamPolicy rather than SetIamPolicy.

I don't see how. The scenario you described *does* put Policy=3 in GetIamPolicy. Or are you saying that we should modify the Policy object returned by the server to be v3 even if it doesn't contain any conditions? If that's the right thing to do, the server should do it. (And we should persuade the IAM team of that.)

I feel I may have misunderstood what you're suggesting though. Could you clarify the difference between what's in this PR and what you'd like to be in there?

broady: > Or are you saying that we should modify the Policy object returned by the server to be v3 even if it doesn't contain any conditions?

Yes, that's what I was saying. I meant "PolicyVersion=3" (PolicyVersion being different to RequestedPolicyVersion).

> If that's the right thing to do, the server should do it. (And we should persuade the IAM team of that.)

Maybe. If the underlying bindings were not directly editable, then it would be a valid signal to throw a runtime error. Here's a valid way it could be used (pseudocode):
```
policy = GetIamPolicy(RequestedPolicyVersion=3)

assert(policy.PolicyVersion == 1) // there are no conditions in this policy
policy.AddMember("user:foo@foo.com", "reader") // a "V1" helper

policy.Raw()[0].Condition = ...

policy.AddMember("user:foo@foo.com", "reader") // throws an exception because the call to `Raw()` above means the policy may not be compatible with V1 helpers.
```

Does that make sense? I know we're deep in the weeds with corner cases and hypotheticals, just trying to make sure we are on the safe/conservative side.

jskeet: I don't think that's feasible, really. The bindings are already directly visible - we can't start breaking that now.
I'd also say that for better or worse, the way the API is structured discourages this.

Now we could make the helper methods detect the presence of a condition in any binding, of course - that would be conservative, but it might also be not-conservative-enough if (and I hope this doesn't happen) another v3-only aspect is introduced as well.

Fundamentally all of this feels fragile, but I think using just the version is at least a simple-to-understand approach.

apis/Google.Cloud.Iam.V1/Google.Cloud.Iam.V1/PolicyPartial.cs
jskeet: Note: we still need to resolve this FIXME, which requires some documentation to exist somewhere

jskeet:
Closing in favour of #3610.

PR #3361, opened 2019-08-13
Translate V3 API

jskeet:
(Won't merge until the sample works.)


jskeet:
Will merge on green - the snippet passes for a whitelisted account. We won't release until it's available without whitelisting.

PR #3364, opened 2019-08-13
Write complete commit messages in release notes

jskeet:


jskeet:
(Self-approving as it's trivial and for a tool.)

PR #3388, opened 2019-08-22
Fix requester pays integration tests

jskeet:
The storage.objectCreator role doesn't give read permission.


jskeet:
(I'm going to run this test in Kokoro before merging.)

jskeet:
Kokoro has just passed, so I'm merging.

PR #3391, opened 2019-08-23
Support for IAM conditions in Storage when fetching bucket policies

jskeet:
The implementation can't be completed yet as the request field doesn't exist in the REST representation, but the TODO shows where it would be.

Sample usage:

```csharp
var policy = client.GetBucketIamPolicy(bucket, new GetBucketIamPolicyOptions { SupportedIamPolicyVersion = 3 });
```

Currently I don't propose adding a client-wide user-settable default; we can do that later as an extra feature if it's requested.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/GetBucketIamPolicyOptionsTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/GetBucketIamPolicyOptionsTest.cs
@@ -26,6 +26,7 @@ namespace Google.Cloud.Storage.V1.Tests
             var options = new GetBucketIamPolicyOptions();
             options.ModifyRequest(request);
             Assert.Null(request.UserProject);
+            Assert.Null(request.OptionsRequestedPolicyVersion);
         }
 
         [Fact]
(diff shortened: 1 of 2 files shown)


jskeet:
@jkwlui @frankyn @broady I've now updated this with the complete implementation.

Unfortunately we can't clone the policy, so in this case if the user specifies a version in a Set request, we'll modify the object they pass in. I've documented this so it's clear, but it's still a little unfortunate.

broady:
Did you remove the version checks on the existing helpers?

jskeet:
> Did you remove the version checks on the existing helpers?

That's in a [different PR](#3359). This PR is just around Storage, which wraps the REST library.

jskeet:
@jkwlui @frankyn @broady This is now just for GetBucketIamPolicy, with a renamed property as discussed.

I believe that this is now production-ready, barring any timing concerns around when we want to make it available. PTAL.

frankyn:
This LGTM, thanks Jon. I want to get checkoff from IAM Team on the C# design document then we can merge it away.

frankyn:
IAM Team checked off high-level expectations for this design.
Thanks Jon, @jkwlui is on point for IAM Conditions releases moving forward.

jskeet:
@frankyn @jkwlui @broady: Does that mean this is okay to merge? Note that we'll want to do another release reasonably soon... if we're confident in the client library API surface, I think it's okay if this turns out not to be customer-useful for a little while.

jskeet:
@amanda-tarafa: We're now okay to merge this, so please review when you're back. Let me know if you need more context.

PR #3396, opened 2019-08-27
Promote the FirestoreEnumNameConverter to the public API

### LINKED ISSUE #2842: the .net Firsestore sdk converts ENUM to Integer, but the Android SDK converts enums to String
the .net Firsestore sdk converts ENUM to Integer, but the Android SDK converts enums to String.

we need a way, on .net to store enum as string, to match the compatibility with our android app.
==> is there an custom attribute we an use to tell .net to store enums as string ?  (like android sdk)

jskeet:
This was previously just a proof-of-concept, but it's useful to put
into the public API once. At the moment it's relatively crude, but
we can add customization later if requested. In particular, I suspect we may want:

- Case-insensitive comparisons
- Customization of the name returned based on a custom attribute

Fixes #2842


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreEnumNameConverter.cs
```diff
@@ -0,0 +1,89 @@
...
+                // This may overwrite an existing value, but that's okay. We explicitly don't
+                // make any guarantees as to which name is returned for a value with multiple names.
+                s_valueToNameMap[value] = name;
+            }
+        }
+        
+        /// <inheritdoc />
+        public T FromFirestore(object value)
+        {
+            string name = (string) value;
```
chrisdunelm: nit: format in the same way as the `ToFirestore` method?
```
... =>
    s_nameToValueMap.TryGetValue((string)value, out var result) ? result : ...
```
```

File: apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreEnumNameConverter.cs [hunk 2]
```diff
@@ -0,0 +1,89 @@
...
+            if (s_nameToValueMap.TryGetValue(name, out var result))
+            {
+                return result;
+            }
+            throw new ArgumentException($"Unknown name {name} for enum {typeof(T).FullName}");
```
chrisdunelm: nit: Put `{name}` in quotes for improved clarity of error message?
(and below for `{value}`)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreEnumNameConverter.cs
jskeet: Had wondered about that - thought it might be a bit much to have the cast in there too, but very happy to be guided by you here :)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreEnumNameConverter.cs
jskeet: Will do.

PR #3411, opened 2019-09-03
Implement conversion for named value tuples, with limitations

### LINKED ISSUE #2787: FirestoreData or FirestoreProperty support for value type tuples
## Problem
As part of your documents in Firestore, you will often have multiple nested objects. In most cases, these objects will be defined as classes within your solution, however in some cases these might simply be  container objects holding a subset of data to simplify queries. In this latter scenario, value type tuples are great as it avoids you cluttering your solution with tons of classes. an example: 

## Example
Using the old course and student nugget, you might have a sub collection within a course called "EnrolledStudents" which is a list of student id's. It would be handy to store the name, surname and major of each student so you don't have to link back to the main student object. This class could look as follows: 

```
    [FirestoreData]
    public class EnrolledStudent
    {
        [FirestoreProperty]
        public string StudentId { get; set; }
        [FirestoreProperty]
        public DateTime EnrollmentDate { get; set; }

        public (string Name, string Email, string Major) StudentInfo;
    }
```

Currently you cannot decorate the tuple with FirestoreProperty or FirestoreData. As you can see, this avoids you having to create the "StudentInfo" class which is essentially just a subset of the Student class. And in other cases you might only want Name & Email.

## Other options

You could just list the properties as part of the main object i.e. StudentName, StudentEmail, StudentMajor and will work as expected, however it's less tidy and concise. 

Let me know if unclear or more info is needed (or if this is in fact already possible) 

P.S. Great work on this library and Firestore in general.

jskeet:
Fixes #2787

Documentation for all the recent converter changes will come in a separate PR.

Please let me know if there are any tests you think I've missed!

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/ValueTupleConverter.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/ValueTupleConverter.cs
@@ -1,0 +1,444 @@
+﻿// Copyright 2019, Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Firestore.V1;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using BclType = System.Type;
+
+namespace Google.Cloud.Firestore.Converters
+{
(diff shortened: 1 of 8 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/ValueTupleConverter.cs
```diff
@@ -0,0 +1,444 @@
...
+        }
+
+        private interface ITupleAccessor
+        {
+            object this[int index] { get; set; }
+            object Value { get; set; }            
+        }
+
+        // Tuple accessor implementations. These all look the same, 
+        private sealed class TupleAccessor<T1> : ITupleAccessor
```
chrisdunelm: Might using `System.Runtime.CompilerServices.ITuple` mean these `TupleAccessor<...>` classes are not required?

jskeet: It would indeed. Unfortunately it's not available in netstandard2.0, despite the docs :( (It's in 2.1.)

chrisdunelm: Oh right, that's unfortunate (and wrongly documented!).

PR #3422, opened 2019-09-05
Release Google.Cloud.Asset.V1Beta1 version 1.0.0-beta02

jskeet:
This is to test our autorelease process


jskeet:
Expectation: when the autorelease bot next runs, it will create a GitHub release, and kick off a release job

PR #3448, opened 2019-09-16
Support custom exit codes for major/minor changes

chrisdunelm:


chrisdunelm:
Or should it simply always return 0 on identical; 1 on minor change; 2 on major change?
So no flags at all?

jskeet:
Let me ponder that. I think we'd very often want it to not give an error code for minor changes, as they're rarely a problem...

chrisdunelm:
@jskeet please take another look - the second commit is a different potential way of controlling exit-code behaviour.

PR #3505, opened 2019-09-18
Changes to Travis jobs and API regexes

jskeet:
We name the jobs explicitly now, which improves readability and
means it's simpler to add new ones.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -60,20 +60,35 @@ done
 # If no APIs were specified explicitly, build all of them (and tools on Windows)
 if [[ ${#apis[@]} -eq 0 ]]
 then
-  apis=(${tools} $(echo apis/Google.* | sed 's/apis\///g'))
+  apis=(${tools} $(find apis -mindepth 1 -maxdepth 1 -type d | sed 's/apis\///g'))
 fi
 
 # If we were given an API filter regex, apply it now.
 if [[ "$apiregex" != "" ]]
 then
   filteredapis=()
-  for api in ${apis[*]}
-  do
-    if [[ "$api" =~ $apiregex ]]
-    then
-      filteredapis+=($api)
-    fi
-  done
+  # This is a hack to allow ! to negate the regex.
+  # Bash regular expressions don't allow for lookahead, so this is the
+  # simplest way of doing it.
+  if [[ $apiregex == !* ]]
+  then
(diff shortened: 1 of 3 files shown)


jskeet:
Sample output (end of log):

```text
=====================
Changes detected
=====================
Checking Google.Cloud.Monitoring.V3 for changes
Major changes:
Class 'Google.Cloud.Monitoring.V3.UptimeCheckServiceClient'; method 'System.Threading.Tasks.Task<Google.Cloud.Monitoring.V3.UptimeCheckServiceClient> CreateAsync(Google.Api.Gax.Grpc.ServiceEndpoint endpoint = null, Google.Cloud.Monitoring.V3.UptimeCheckServiceSettings settings = null)' removed.
Diff level: Major
---------------------
Checking Google.Cloud.Storage.V1 for changes
Minor changes:
Class 'Google.Cloud.Storage.V1.NewClass' added.
Diff level: Minor
---------------------
Checking Google.Cloud.TextToSpeech.V1 for changes
Diff level: Identical
---------------------
Google.Cloud.VideoIntelligence.V1 was deleted
---------------------
SomeNewApi is new
---------------------
The command "./travis-changes.sh" exited with 0.
Done. Your build exited with 0.
```

jskeet:
(We may eventually want to have a "brief" mode in VersionCompat that emits short names rather than fully-qualified names, but that can wait).

jskeet:
This is only a starting point. We're not failing the build on major changes or anything like that. But it should make it easy to see what's changed in a PR.

build.sh
```diff
@@ -60,20 +60,35 @@ done
...
-    then
-      filteredapis+=($api)
-    fi
-  done
+  # This is a hack to allow ! to negate the regex.
+  # Bash regular expressions don't allow for lookahead, so this is the
+  # simplest way of doing it.
+  if [[ $apiregex == !* ]]
+  then
+    apiregex=$(echo $apiregex | sed s/^!//g)
```
amanda-tarafa: Can't you just do this inside the if, the rest of the code seems the same for if and else clauses.

jskeet: The tricky part is negating the regex condition. We could have an extra flag saying whether or not to negate, but then the condition becomes complex. It felt like duplication was the lesser of two evils here.

Issue #3433, opened 2019-09-11
Run the breaking change detector on PRs


jskeet:
For APIs which have already gone GA, it would be useful to run our breaking change detector against the published version.
Will need to look to see how tricky that is.

jskeet:
VersionCompat feature requests:

- Load from NuGet
- Split into breaking / non-breaking
- Allow "only show breaking"
- Allow "return error code on breaking"

jskeet:
Added assignment to Chris for the feature requests; I'll implement the integration part.

jskeet:
Rules for what to compare against, based on version in the API catalog (which indicates the most recent release, unless this change *is* a release):

- `x.0.0-*`: None
- `x.0.*-*`: `x.0.0`
- `x.y.*`: `x.y.0`
- `x.y.*-*`: `x.(y-1).0`

Need to handle "this change is a release" separately somehow.

jskeet:
See #3505 for the start of this work.

jskeet:
It's possible that what we've got now is fine for PRs, but that if we detect version changes in a PR, *then* we should check against the previous versions.

jskeet:
Fixed by #3572

PR #3507, opened 2019-09-18
Add flag to control Output behaviour

chrisdunelm:
Updates #3433

### DIFF
--- a/tools/Google.Cloud.Tools.VersionCompat/Program.cs
+++ b/tools/Google.Cloud.Tools.VersionCompat/Program.cs
@@ -93,12 +93,29 @@ namespace Google.Cloud.Tools.VersionCompat
             MajorAndMinor = 2,
         }
 
+        [Flags]
+        private enum OutputBehavior
+        {
+            None = 0,
+            Major = 1,
+            Minor = 2,
+
+            // To allow all combinations to be set with a single flag value
+            MajorAndMinor = 3,
+        }
+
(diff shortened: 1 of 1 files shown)


tools/Google.Cloud.Tools.VersionCompat/Program.cs
```diff
@@ -93,12 +93,26 @@ private enum ExitCodeBehavior
             MajorAndMinor = 2,
         }
 
+        private enum OutputBehavior
```
jskeet: Possibly make this a flags enum, then you can conditionalize it more easily?

chrisdunelm: done

tools/Google.Cloud.Tools.VersionCompat/Program.cs
```diff
@@ -137,7 +154,7 @@ static int Main(string[] args)
 
             var diffs = Check(assem0, assem1, null);
 
-            if (diffs.Major.Any())
+            if ((options.OutputBehavior & OutputBehavior.Major) != 0 && diffs.Major.Any())
```
jskeet: Could use Enum.HasFlag, but that could come later.

chrisdunelm: ack :) I'd forgotten that existed!

PR #3515, opened 2019-09-19
Generate diff in autosynth

jskeet:


jskeet:
I suggest reviewing one commit at a time :)

PR #3530, opened 2019-09-23
Fix dialogflow

jskeet:


jskeet:
Self-merging to get this in ahead of autosynth

PR #3531, opened 2019-09-23
Revert "Remove a Create method from Monitoring"

jskeet:
Reverts an accidental commit within googleapis/google-cloud-dotnet#3530


jskeet:
This is getting too messy...

PR #3567, opened 2019-09-28
[Spanner] Batch session create support

amanda-tarafa:
Adds support for  Batch Session create in Spanner. Can't be merged until batch affinity is supported by Grpc.Gcp (see [this issue](https://github.com/GoogleCloudPlatform/grpc-gcp-csharp/issues/25)).

@jskeet could you take a look an see if this is going in the right direction? 
Note that:
- Options for batch size and session creation requests are being used as I understood you last proposed but we might need to look further into this.
- Affinity configuration for batch create sessions is included in this PR but commented out, until supported by Grpc.Gcp.
- Some unit tests became slightly slower because of this PR, so their timeout had to be slightly increased.
- Integration tests are passing, which is somewhat surprising and means at least one of the following:
    - Session - channel affinity is desired but not mandatory.
    - We are testing with just one gRPC channel (will check this on monday).
    - We are lucky.


apis/Google.Cloud.Spanner.Data/docs/configuration.md
```diff
@@ -142,11 +142,23 @@ server, including retries. This setting is applied to calls to
 create, refresh and delete sessions, as well as beginning
 transactions.
 
-Applications rarely need to change this setting
+Applications rarely need to change this setting.
+
+**CreateSessionBatchMaximumSize**
```
jskeet: Maybe this should be CreateSessionMaximumBatchSize? That sounds better to me (in that it's the "maximum batch size" for "create session" but I could probably be convinced :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
```diff
@@ -162,18 +163,39 @@ public TimeSpan Timeout
...
         /// <para>
-        /// This value must be positive. The default value is 10.
+        /// This value must be positive. This value should always be greater than or equal to <see cref="CreateSessionBatchMaximumSize"/>
+        /// The default value is 50.
         /// </para>
         public int MaximumConcurrentSessionCreates
         {
             get => _maximumConcurrentSessionCreates;
-            set => _maximumConcurrentSessionCreates = GaxPreconditions.CheckArgumentRange(value, nameof(value), 1, int.MaxValue);
+            set => _maximumConcurrentSessionCreates = GaxPreconditions.CheckArgumentRange(value, nameof(value), CreateSessionBatchMaximumSize, int.MaxValue);
```
jskeet: There's a problem with having these preconditions - it means you can end up with an object initializer which is valid in one ordering and invalid in another.

If MaximumConcurrentSessionCreates is less than CreateSessionBatchMaximumSize, it just means we'll always use a single batch - that's not unreasonable, really.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -48,15 +48,22 @@ internal sealed class TargetedSessionPool : SessionPoolBase
             // Tasks waiting for the pool to reach its minimum size or become unhealthy.
             private readonly LinkedList<TaskCompletionSource<int>> _minimumSizeWaiters = new LinkedList<TaskCompletionSource<int>>();
 
+            // The size of the last unhealthy batch.
+            // If 0, the pool is healthy.
             // Assume the pool is healthy to start with.
-            private int _healthy = 1;
-            private bool Healthy => Interlocked.CompareExchange(ref _healthy, 0, 0) == 1;
+            private int _lastUnhealthyBatch;
```
jskeet: Possibly `_lastUnhealthyBatchSize` to make it clear that it's the size? (Ditto the property.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -48,15 +48,22 @@ internal sealed class TargetedSessionPool : SessionPoolBase
...
+            private bool Healthy => Interlocked.CompareExchange(ref _lastUnhealthyBatch, 0, 0) == 0;
+            private int LastUnhealthyBatch => Interlocked.CompareExchange(ref _lastUnhealthyBatch, -1, -1);
 
             // The task (completion source) associated with shutting down the session pool, if any.
             // Note that this task has no cancellation token associated with it, even if the original call did.
             private TaskCompletionSource<int> _shutdownTask = null;
             private bool Shutdown => Interlocked.CompareExchange(ref _shutdownTask, null, null) != null;
 
+            // The task (completion source) associated with nursing the pool back to health, if any.
+            // The pool can only be being nursed back to health by at most one task at any give moment.
```
jskeet: Nit: give => given

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -242,10 +250,20 @@ private Task<PooledSession> GetSessionAcquisitionTask(ModeOneofCase transactionM
...
+                        // This means basically to try and nurse the pool back to health.
                     }
                 }
                 // Effectively the "else" block above, but outside the lock
-                return CreatePooledSessionAsync(cancellationToken);
+
+                // Just to hide the "await" here so we don't have to make this method async.
+                // Because we cannot await inside the lock, the only way to make this method
+                // async is to have it return Task<Task<PooledSession>> which is weird.
+                return Task.Run(async () =>
```
jskeet: I think we should be able to do this just by having an async method - I don't think we need Task.Run

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -242,10 +250,20 @@ private Task<PooledSession> GetSessionAcquisitionTask(ModeOneofCase transactionM
...
+                {
+                    await GetNursePoolBackToHealthTask(cancellationToken).ConfigureAwait(false);
+                    // If we reached this point the call above dind't fail and the pool is healthy.
```
jskeet: Nit: dind't => didn't (or make it "The nursing task succeeded and the pool is healthy"

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -518,44 +537,57 @@ private void StartAcquisitionTasksIfNecessary()
...
+                // How many more sessions do we need to create in order to get the pool to the minimum size, after satisfying
+                // pending callers, assuming that all the in-flight session requests succeed, and there are no more requests?
+                int newSessionsToSatisfyPool = (pendingAcquisitionCount + minPoolSize) - (poolSize + inFlightRequests);
+                // How many more sessions *can* we create without going over the maximum number of active sessions?
+                int maxPossibleNewSessions = maxActiveSessions - liveOrRequestedSessionCount;
+
+                int actualNewSessions = Math.Min(newSessionsToSatisfyPool, maxPossibleNewSessions);
+
+                // Now we need to know how many batch session create requests we need to make.
+                int partialBatchSize = actualNewSessions % createSessionBatchMaxSize;
```
jskeet: You could probably use Math.DivRem for this.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -518,44 +537,57 @@ private void StartAcquisitionTasksIfNecessary()
...
+                int totalWholeBatches = actualNewSessions / createSessionBatchMaxSize;
 
-                int actualNewRequests = Math.Min(newRequestsToSatisfyPool, maxAvailableRequests);
+                for (int i = 0; i < totalWholeBatches; i++)
+                {
+                    Parent.ConsumeBackgroundTask(PrepareNewSessionsAsync(createSessionBatchMaxSize), "session creation");
+                }
 
-                for (int i = 0; i < actualNewRequests; i++)
+                if(partialBatchSize > 0)
```
jskeet: Nit: space after `if` (but irrelevant if the above suggestion is taken).

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -518,44 +537,57 @@ private void StartAcquisitionTasksIfNecessary()
...
+                int maxPossibleNewSessions = maxActiveSessions - liveOrRequestedSessionCount;
+
+                int actualNewSessions = Math.Min(newSessionsToSatisfyPool, maxPossibleNewSessions);
+
+                // Now we need to know how many batch session create requests we need to make.
+                int partialBatchSize = actualNewSessions % createSessionBatchMaxSize;
+                int totalWholeBatches = actualNewSessions / createSessionBatchMaxSize;
 
-                int actualNewRequests = Math.Min(newRequestsToSatisfyPool, maxAvailableRequests);
+                for (int i = 0; i < totalWholeBatches; i++)
```
jskeet: Alternative option for this - a single `while` loop, such as:

```csharp
// Renamed from actualNewSessions above
int remainingNewSessions = Math.Min(newSessionsToSatisfyPool, maxPossibleNewSessions);
while (remainingNewSessions > 0)
{
    int batchSize = Math.Min(remainingNewSessions, createSessionBatchMaxSize);
    Parent.ConsumeBackgroundTask(PrepareNewSessionsAsync(batchSize), "session creation");
    remainingNewSessions -= batchSize;
}
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
jskeet: Also, it's possible that the above succeeds because the pool is shutdown. Is that still okay?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -518,44 +537,57 @@ private void StartAcquisitionTasksIfNecessary()
...
+                        {
+                            ReleaseInactiveSession(session, maybeCreateReadWriteTransaction: true);
+                        }
                     }
                     // Note: we expect this to always actually be an RpcException, but we don't want to end up unhealthy
                     // with a lot of 
                     catch (Exception e)
                     {
-                        Parent._logger.Warn(() => $"Failed to create session for {_databaseName}", e);
+                        Parent._logger.Warn(() => $"Failed to batch create sessions for {_databaseName}", e);
```
jskeet: Should this block set _lastUnhealthyBatch?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -576,29 +608,96 @@ async Task PrepareNewSessionAsync()
...
+                        foreach (var session in sessions)
+                        {
+                            ReleaseInactiveSession(session, maybeCreateReadWriteTransaction: true);
+                        }
+
+                        // We have succesfully nurse the pool back to health.
+                        newTcs.TrySetResult(1);
+                    }
+                    // Note: we expect this to always actually be an RpcException, but we don't want to end up unhealthy
+                    // with a lot of
```
jskeet: Unfinished comment

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -576,29 +608,96 @@ async Task PrepareNewSessionAsync()
...
+                    // Note: we expect this to always actually be an RpcException, but we don't want to end up unhealthy
+                    // with a lot of 
+                    catch (Exception e)
+                    {
+                        // We couldn't nurse the pool back to health.
+                        newTcs.TrySetException(e);
+
+                        Parent._logger.Warn(() => $"Failed to batch create session for {_databaseName}", e);
+
+                        // If we are here it's because the pool was unhealthy, so we are not receiving and don't have
```
jskeet: I don't understand this comment. I suspect I could after reading more context, but if it can be expanded so I don't have to, that would be great :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -576,29 +608,96 @@ async Task PrepareNewSessionAsync()
...
+                    try
+                    {
+                        // Not using the cancellation token here so this can go on even if the caller cancels the call.
+                        var sessions = await CreatePooledSessionsAsync(LastUnhealthyBatch, CancellationToken.None).ConfigureAwait(false);
+                        foreach (var session in sessions)
+                        {
+                            ReleaseInactiveSession(session, maybeCreateReadWriteTransaction: true);
+                        }
+
+                        // We have succesfully nurse the pool back to health.
```
jskeet: Nit: nurse => nursed

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -576,29 +608,96 @@ async Task PrepareNewSessionAsync()
...
+                        newTcs.TrySetResult(1);
+                    }
+                    // Note: we expect this to always actually be an RpcException, but we don't want to end up unhealthy
+                    // with a lot of 
+                    catch (Exception e)
+                    {
+                        // We couldn't nurse the pool back to health.
```
jskeet: Possibly clarify this comment to say we want to propagate the exception to anyone waiting for the nursing task. (I was originally looking for a throw statement. I understand why there isn't one, but I didn't get it from this comment.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
jskeet: Ah, it looks like that's handled by CreatePooledSessionsAsync. It might be worth mentioning that here in a comment. (It's slightly odd for the logging to be separate from the update. Maybe we should move this log into CreatePooledSessionsAsync.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -576,29 +608,96 @@ async Task PrepareNewSessionAsync()
                         minimumSizeWaiters.ForEach(tcs => tcs.TrySetException(e));
                     }
                 }
+            } 
+
+            private Task GetNursePoolBackToHealthTask(CancellationToken cancellationToken)
```
jskeet: I'm slightly nervous about this, just because it's a big new chunk of code. Was the previous code simpler because we could always just create a single new session?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
```diff
@@ -275,26 +277,26 @@ public async Task AcquireAsync_MaxActiveSessions_Fail()
             [Fact(Timeout = TestTimeoutMilliseconds)]
             public async Task AcquireAsync_MaxActiveSessions_Block()
             {
-                var pool = CreatePool(false);
-                var client = (SessionTestingSpannerClient)pool.Client;
-                pool.Options.WaitOnResourcesExhausted = ResourcesExhaustedBehavior.Block;
+                    var pool = CreatePool(false);
```
jskeet: Undo this extra indentation?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Actually I think the previous code would have benefited from something like this as well. When the pool was unhealthy, we attempted to create a single new session for each caller trying to acquire a session. If these were "a lot" and depending on the options setup, we could have ended up either with more than the allowed number of sessions, or potentially "choking" with individual new session requests an already unhealthy pool/service. When creating sessions in batch, the chances of ending up with more sessions than allowed is greater, that's why I went this way. 
This method is really not that different from the `StartSessionCreationTasksIfNecessary.PrepareNewSessionsAsync` method, it does the same thing in different circumstances.
Does that sound good?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: I actually copied this one on purpose from `StartSessionCreationTasksIfNecessary.PrepareNewSessionsAsync`, it's unfinished there as well and I can't think of what to write next :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Done.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Done.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
amanda-tarafa: True, and I've hated when that has happened to me.
And true as well that is no big deal to remove that validation.

Done.

apis/Google.Cloud.Spanner.Data/docs/configuration.md
amanda-tarafa: Done

jskeet:
Thanks for addressing the comments - will look when I know I'll have at least half an hour of solid concentration :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -518,45 +538,48 @@ private void StartAcquisitionTasksIfNecessary()
...
+                int maxPossibleNewSessions = maxActiveSessions - liveOrRequestedSessionCount;
 
-                for (int i = 0; i < actualNewRequests; i++)
+                int remainingNewSessions = Math.Min(newSessionsToSatisfyPool, maxPossibleNewSessions);
+                while (remainingNewSessions > 0)
                 {
-                    Parent.ConsumeBackgroundTask(PrepareNewSessionAsync(), "session creation");
+                    int batchSize = Math.Min(remainingNewSessions, createSessionBatchMaxSize);
+                    Parent.ConsumeBackgroundTask(PrepareNewSessionsAsync(batchSize), "session creation");
+                    remainingNewSessions -= batchSize;
```
skuruppu: It's possible that BatchCreateSessions will return fewer than the session count asked for. So I would do this computation in the callback in the async function below.

amanda-tarafa: I think we can actually do better. We can explicitly call StartSessionCreationTasksIfNecessary whenever a BatchCreateSession returns less sessions than requested. @jskeet, what do you think?

amanda-tarafa:
This is now rebased onto #3577.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -36,7 +36,7 @@ internal sealed class TargetedSessionPool : SessionPoolBase
             // Read-only state
             private readonly object _lock = new object();
             private readonly DatabaseName _databaseName;
-            private readonly CreateSessionRequest _createSessionRequest;
+            private readonly BatchCreateSessionsRequest _batchCreateSessionRequest;
```
jskeet: It may be worth adding a comment to note that this should be cloned each time it's used.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -518,45 +538,48 @@ private void StartAcquisitionTasksIfNecessary()
...
-                int actualNewRequests = Math.Min(newRequestsToSatisfyPool, maxAvailableRequests);
+                // How many more sessions do we need to create in order to get the pool to the minimum size, after satisfying
+                // pending callers, assuming that all the in-flight session requests succeed, and there are no more requests?
+                int newSessionsToSatisfyPool = (pendingAcquisitionCount + minPoolSize) - (poolSize + inFlightRequests);
+                // How many more sessions *can* we create without going over the maximum number of active sessions?
+                int maxPossibleNewSessions = maxActiveSessions - liveOrRequestedSessionCount;
 
-                for (int i = 0; i < actualNewRequests; i++)
+                int remainingNewSessions = Math.Min(newSessionsToSatisfyPool, maxPossibleNewSessions);
+                while (remainingNewSessions > 0)
```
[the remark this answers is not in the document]
jskeet: Yes, that seems reasonable. We may be able to do without a bunch of local variables that way, too.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
jskeet: I think it's worth doing whatever is simplest here, to be honest. Do we know how often BatchCreateSessions returns fewer sessions than requested? If it's relatively rare, I'd be happy with "this can occasionally be inefficient, but will never result in session starvation".

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
jskeet: Okay, that sounds fine.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -624,29 +719,33 @@ private async Task<PooledSession> CreatePooledSessionAsync(CancellationToken can
...
+                }
                 finally
                 {
                     // Atomically set _healthy and determine whether we were previously healthy, but only if either we've succeeded,
                     // or we failed for a reason other than cancellation. We don't want to go unhealthy just because a caller cancelled
                     // a cancellation token before we had chance to create the session.
                     if (success || !canceled)
                     {
-                        bool wasHealthy = Interlocked.Exchange(ref _healthy, success ? 1 : 0) == 1;
+                        bool wasHealthy = Interlocked.Exchange(ref _lastUnhealthyBatchSize, success ? 0 : batchSize) == 0;
```
jskeet: The comment in 733 needs to be updated too.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -624,29 +719,33 @@ private async Task<PooledSession> CreatePooledSessionAsync(CancellationToken can
...
                             Parent._logger.Info(() => $"Session pool for {_databaseName} is now {(success ? "healthy" : "unhealthy")}.");
                         }
                     }
                     // If this call failed, we can make another attempt.
-                    if (!success)
-                    {
-                        Interlocked.Decrement(ref _liveOrRequestedSessionCount);
-                    }
-                    Interlocked.Decrement(ref _inFlightSessionCreationCount);
+                    // If it succeeded update the value with the actual sessions created.
```
jskeet: This comment suggests there should be an "if" here, which there isn't.
I think probably these two comments need to be rewritten along the lines of "update the state of the world to the new reality based on how many sessions were returned, which may be 0".

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
```diff
@@ -162,13 +163,32 @@ public TimeSpan Timeout
...
+        }
+
         /// <summary>
         /// The maximum number of session create operations allowed to occur simultaneously per session pool.
         /// Spanner has limits on the number of sessions that can be created concurrently without affecting performance.
         /// This value is not typically changed.
         /// </summary>
         /// <para>
-        /// This value must be positive. The default value is 10.
+        /// This value must be positive. The default value is 50.
```
jskeet: This feels odd to me. If we're using batches, I'd expect us to use *fewer* concurrent operations, not more.

amanda-tarafa:
@jskeet I think this is now ready for merge after review.
New changes included in the 4 last commits.
PTAL, thanks!

jskeet:
The AppVeyor tests have failed - could you have a look at the log and consider whether it's worrying?

amanda-tarafa:
On it.

amanda-tarafa:
It's a timeout, not of the FakeScheduller but of the actual test. I'll increase it some and push a new commit.

amanda-tarafa:
All green now. Not a worrying timeout, just that some unit tests might take a little longer, which is understandable since we have changed the way in which sessions are created.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -537,48 +544,37 @@ private void StartSessionCreationTasksIfNecessary()
...
+                    // pending callers, assuming that all the in-flight session requests succeed, and there are no more requests?
+                    int newSessionsToSatisfyPool = (_pendingAcquisitions.Count + Options.MinimumPooledSessions) - (poolSize + InFlightSessionCreationCount);
+                    // How many more sessions *can* we create without going over the maximum number of active sessions?
+                    int maxPossibleNewSessions = Options.MaximumActiveSessions - liveOrRequestedSessionCount;
+
+                    int remainingNewSessions = Math.Min(newSessionsToSatisfyPool, maxPossibleNewSessions);
+                    while (remainingNewSessions > 0)
+                    {
+                        int batchSize = Math.Min(remainingNewSessions, Options.CreateSessionMaximumBatchSize);
+                        Parent.ConsumeBackgroundTask(PrepareNewSessionsAsync(batchSize), "session creation");
```
jskeet: I'm slightly nervous now, because we're calling PrepareNewSessionsAsync in the lock, which calls CreatePooledSessionsAsync in the lock... that's a lot of code (in terms of scope) to execute in the lock, and I'm worried that we might end up deadlocking.

It might be okay, but I wonder how feasible it would be to have a loop of "acquire lock, work out how many sessions to create, release lock, call into other code, loop again" until the number of sessions to create is 0? Feel free to argue against this, or to perhaps argue to do it in a separate PR.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -673,7 +669,22 @@ private Task GetNursePoolBackToHealthTask(CancellationToken cancellationToken)
...
-                return newTcs.WithCancellationToken(cancellationToken);
+                return WithCancellationToken(newTcs, cancellationToken);
+
+                async Task WithCancellationToken(TaskCompletionSource<int> source, CancellationToken token)
+                {
+                    TaskCompletionSource<int> cancellable = new TaskCompletionSource<int>();
+                    token.Register(() => cancellable.TrySetCanceled());
+                    try
+                    {
+                        int dummyResult = await source.Task.ConfigureAwait(false);
```
jskeet: This doesn't propagate cancellation immediately - it'll still wait for the original task first. The final result will be the same, but with very different timing than what I'd expect. We might need a VC about this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
```diff
@@ -299,6 +299,226 @@ public async Task AcquireAsync_MaxActiveSessions_Block()
...
+                Assert.Equal(13, stats.ActiveSessionCount);
+                Assert.Equal(3, stats.PendingAcquisitionCount);
+                Assert.Equal(0, stats.ReadPoolCount);
+                Assert.Equal(0, stats.ReadWritePoolCount);
+                Assert.Equal(13, stats.InFlightCreationCount);
+
+                // Make the create session tasks fail and notify the waiting callers.
+                await client.Scheduler.RunAsync(TimeSpan.FromSeconds(5));
+
+                int succesfulAcquisitions = 0;
```
jskeet: Nit: succesful => successful

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
```diff
@@ -299,6 +299,226 @@ public async Task AcquireAsync_MaxActiveSessions_Block()
...
+            public async Task AcquireAsync_NursesBackToHealth()
+            {
+                var pool = CreatePool(false);
+                var client = (SessionTestingSpannerClient)pool.Client;
+                var options = pool.Options;
+
+                client.FailAllRpcs = true;
+                client.FailAllRpcsDelay = TimeSpan.FromSeconds(1);
+
+                // In total MinimumSizePool + 1 sessions will be created (or tried to in our case)
```
jskeet: Is the +1 because of the "trying to nurse back to health"? If so, that's probably worth saying in the comment.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
```diff
@@ -299,6 +299,226 @@ public async Task AcquireAsync_MaxActiveSessions_Block()
...
+                // pool to become healthy first.
+                Assert.Equal(0, stats.PendingAcquisitionCount);
+                Assert.Equal(0, stats.ReadPoolCount);
+                Assert.Equal(0, stats.ReadWritePoolCount);
+                // Nursing back to health attempts to create the last failed batch.
+                // That will be either a batch of maximum size or a smaller batch with the remainder
+                // whichever failed last.
+                Assert.True(stats.InFlightCreationCount == 5 || stats.InFlightCreationCount == 1);
+
+                // Run enough to have the pool become healthy and all the requested sessions acquiered.
```
jskeet: Nit: acquiered => acquired

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
jskeet: Ping on this.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
jskeet: Okay. Let's defer to another PR. Maybe add a TODO?

(My main cause for concern is that reasoning about how much work happens in the lock is really tricky when it calls async methods, due to the "execute synchronously until the first await that needs to pause" aspect. It's odd to have "this code is sometimes executed in the lock and sometimes not".)

amanda-tarafa: Adding a TODO.
(And agree with your concern as well, I'll start a PR straight away to refactor)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPoolOptions.cs
amanda-tarafa: Done here and in configuration.md as well.

PR #3570, opened 2019-10-01
Add support for NuGet packages to VersionCompat

jskeet:
This includes a slight refactoring that will make the project easier to use from other code too.

### DIFF
--- a/tools/Google.Cloud.Tools.VersionCompat/Program.cs
+++ b/tools/Google.Cloud.Tools.VersionCompat/Program.cs
@@ -77,13 +77,15 @@ namespace Google.Cloud.Tools.VersionCompat
         /// <param name="version">The version of the package to load. May be null, in which case the latest stable version is downloaded.</param>
         /// <param name="tfm">The target framework to find within the package. May be null, in which case "netstandard2.0" is assumed.</param>
         /// <param name="assemblyName">The name of the assembly to find within the package. May be null, in which case it's assumed to be the same as the package.</param>
-        /// <returns></returns>
+        /// <returns>The assembly definition loaded from the given package.</returns>
         public async static Task<AssemblyDefinition> LoadPackageAsync(string package, string version, string tfm, string assemblyName)
         {
+            tfm = tfm ?? "netstandard2.0";
+            assemblyName = assemblyName ?? package;
             var client = new HttpClient();
             // Handily, this automatically loads the latest stable release if version is null.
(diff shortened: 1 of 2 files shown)


tools/Google.Cloud.Tools.VersionCompat/Program.cs
```diff
@@ -66,24 +70,78 @@ public static DiffResult Check(AssemblyDefinition older, AssemblyDefinition newe
...
+        public async static Task<AssemblyDefinition> LoadPackageAsync(string package, string version, string tfm, string assemblyName)
+        {
+            var client = new HttpClient();
+            // Handily, this automatically loads the latest stable release if version is null.
+            var url = $"https://www.nuget.org/api/v2/package/{package}/{version}";
+            var response = await client.GetAsync(url);
+            response.EnsureSuccessStatusCode();
+            var bytes = await response.Content.ReadAsByteArrayAsync();
+
+            string expectedPath = $"lib/{tfm}/{assemblyName}.dll";
```
chrisdunelm: I expect this will fail if a tfm isn't given.
Which doesn't matter for now, but should be fixed if specifying a tfm really is optional.

tools/Google.Cloud.Tools.VersionCompat/Program.cs
```diff
@@ -66,24 +70,78 @@ public static DiffResult Check(AssemblyDefinition older, AssemblyDefinition newe
...
+                    if (entry.Key == expectedPath)
+                    {
+                        var path = entry.Key.Substring(4);
+                        string targetFramework = Path.GetDirectoryName(path);
+                        using (var stream = entry.OpenEntryStream())
+                        {
+                            // Mono.Cecil requires the stream to be seekable. It's simplest
+                            // just to copy the whole DLL to a MemoryStream and pass that to Cecil.
+                            var ms = new MemoryStream();
+                            stream.CopyTo(ms);
```
chrisdunelm: CopyToAsync() ??

tools/Google.Cloud.Tools.VersionCompat/Program.cs
jskeet: Fixed - both `tfm` and `assemblyName` weren't properly optional.

tools/Google.Cloud.Tools.VersionCompat/Program.cs
jskeet: Done.

PR #3572, opened 2019-10-02
Check for new releases, and validate their compatibility

jskeet:
We compare the API catalog with the tags in git, to work out what would be released.

We then perform zero, one or two compatibility checks:

- If there's a previous release we need to be compatible with:
  - A new major release can break anything
  - A new minor release (even a beta) must be compatible with the previous
    minor release (minor changes only)
  - A new patch release must be identical to the .0 patch level for the same major/minor
- If there's a different previous release, show the changes from
  that in an advisory sense. For example, it's good to know what's
  changed between betas, even though it's okay for some of those changes to be breaking.

### DIFF
--- a/detect-pr-changes.sh
+++ b/detect-pr-changes.sh
@@ -11,6 +11,7 @@ mkdir tmpgit/new
 
 # First build everything, so we can get straight to the good stuff at the end of the log.
 dotnet build tools/Google.Cloud.Tools.VersionCompat
+dotnet build tools/Google.Cloud.Tools.CheckVersionCompatibility
 
 for api in $apis
 do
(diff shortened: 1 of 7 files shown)


jskeet:
I can split this into multiple commits if that's desirable, btw. But it's not a huge amount of code anyway :)

detect-pr-changes.sh
```diff
@@ -11,6 +11,7 @@ mkdir tmpgit/new
 
 # First build everything, so we can get straight to the good stuff at the end of the log.
 dotnet build tools/Google.Cloud.Tools.VersionCompat
```
amanda-tarafa: Do you still need this here? Wouldn't this one get built anyway because of the dependency CheckVersionCompatibility has on it?

jskeet: Yes, but I think I'd rather be explicit. If we ever remove the dependency in CheckVersionCompatibility (e.g. by making the two console apps depend on a single common library) it would be easy to forget to put this back.

jskeet:
Merging now without Chris's review - would still welcome a post-merge review.

PR #3577, opened 2019-10-03
[Spanner] Store and give out sessions in LIFO instead of FIFO style.

amanda-tarafa:
Apparently there's a 30 second cache server side that stores somo of the session's information, so it's more efficient to hand them out LIFO style.

@skuruppu will create an issue for this.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.cs
@@ -91,6 +91,58 @@ namespace Google.Cloud.Spanner.V1.Tests
         }
 
         [Fact(Timeout = TestTimeoutMilliseconds)]
+        public async Task ScheduledMaintenanceEvictsSessions_DifferentEvictionTimes()
+        {
+            var client = new SessionTestingSpannerClient();
+            var options = new SessionPoolOptions
+            {
+                // We'll never actually hit a refresh, as the eviction delay is shorter.
+                IdleSessionRefreshDelay = TimeSpan.FromMinutes(30),
+                PoolEvictionDelay = TimeSpan.FromMinutes(3),
+                MaintenanceLoopDelay = TimeSpan.FromMinutes(1),
+                SessionEvictionJitter = RetrySettings.NoJitter,
+                MinimumPooledSessions = 10,
+                MaximumConcurrentSessionCreates = 20,
+                WriteSessionsFraction = 0
+            };
+            var sessionPool = new SessionPool(client, options);
+            var acquisitionTask = sessionPool.AcquireSessionAsync(s_sampleDatabaseName, new TransactionOptions(), default);
+
+            await client.Scheduler.RunAsync(TimeSpan.FromMinutes(1));
+
+            // Our session should be ready, the pool should be up to size, and we should
+            // have created 11 sessions in total.
+            var session = await acquisitionTask;
+            var stats = sessionPool.GetStatisticsSnapshot(s_sampleDatabaseName);
+            Assert.Equal(10, stats.ReadPoolCount);
+            Assert.Equal(11, client.SessionsCreated);
+            Assert.Equal(0, client.SessionsDeleted);
+
+            // Force the creation of a newer session by acquiring one.
+            // The new one will be created to satisfy the minimum size of the pool
+            // and will sit on top of the stack.
+            // First move the pool to T=2, so that the new session will be created in T=3
+            // so that its eviction time will be T=6 so as to make sure that it's not
+            // being evicted when we check at T=5.
+            await client.Scheduler.RunAsync(TimeSpan.FromMinutes(1));
+            acquisitionTask = sessionPool.AcquireSessionAsync(s_sampleDatabaseName, new TransactionOptions(), default);
+            await client.Scheduler.RunAsync(TimeSpan.FromMinutes(1));
+            session = await acquisitionTask;
+            stats = sessionPool.GetStatisticsSnapshot(s_sampleDatabaseName);
+            Assert.Equal(10, stats.ReadPoolCount);
+            Assert.Equal(12, client.SessionsCreated);
+            Assert.Equal(0, client.SessionsDeleted);
+
+            // If we allow the maintenance pool to run until T=5 minutes, we should have evicted
+            // all the old 9 sessions in the pool and replaced them with 9 new ones.
+            await client.Scheduler.RunAsync(TimeSpan.FromMinutes(2));
+            stats = sessionPool.GetStatisticsSnapshot(s_sampleDatabaseName);
+            Assert.Equal(10, stats.ReadPoolCount);
+            Assert.Equal(21, client.SessionsCreated);
+            Assert.Equal(9, client.SessionsDeleted);
+        }
+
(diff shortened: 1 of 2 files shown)


skuruppu:
I created #3578. I don't have sufficient repo perms to attach this PR to the issue.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -483,8 +483,8 @@ void RemoveStaleOrExpiredItemsFromQueue(ConcurrentQueue<PooledSession> queue)
                         }
                         else
                         {
-                            // The session is fine - put it back on the queue.
-                            queue.Enqueue(session);
+                            // The session is fine - put it back on the stack.
+                            stack.Push(session);
```
jskeet: We can't do this - we'll end up pushing and popping the same element multiple times.
We could do with a dequeue really - but otherwise, we'll need to work out how to check everything but end up with it in the right order.

Before fixing this, it would be good to have a test for this case (that fails with the current code) if possible.

amanda-tarafa: Ohh, rigth, totally went over this. Will right test first and fix after. Thanks for catching!

amanda-tarafa: This is now fixed.
Before fixing, the new test failed as expected.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.cs
```diff
@@ -90,6 +90,58 @@ public async Task ScheduledMaintenanceEvictsSessions()
...
+            var stats = sessionPool.GetStatisticsSnapshot(s_sampleDatabaseName);
+            Assert.Equal(10, stats.ReadPoolCount);
+            Assert.Equal(11, client.SessionsCreated);
+            Assert.Equal(0, client.SessionsDeleted);
+
+            // Force the creation of a newer session by acquiring one.
+            // The new one will be created to satisfy the minimum size of the pool
+            // and will sit on top of the stack.
+            // First move the pool to T=2, so that the new session will be created in T=3
+            // so that it's eviction time will be T=6 so as to make sure that it's not
```
jskeet: Nit - not overly bothered if it's fixed or not. The first "it's" in this line should be "its".

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -464,27 +464,29 @@ private void EvictAndRefreshSessions()
...
-                    int count = stack.Count;
-                    for (int i = 0; i < count; i++)
+                    if (!stack.IsEmpty)
                     {
-                        // This shouldn't happen, as nothing else should touch the stack while we're in a
-                        // lock, but we don't want to throw an exception
-                        if (!stack.TryPop(out var session))
+                        int count = stack.Count;
+                        PooledSession[] sessions = new PooledSession[count];
+                        if (stack.TryPopRange(sessions) > 0)
```
jskeet: I think it would be better to have:

```csharp
int count;
PooledSession[] sessions = new PooledSession[stack.Count];
if ((count = stack.TryPopRange(sessions)) > 0)
{
}
```

That way it's clearer that we won't be trying to use an array element which wasn't actually popped.
Probably not a problem due to the locking, but might make it more robust if we change the locking strategy later.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -464,27 +464,29 @@ private void EvictAndRefreshSessions()
...
-                        }
-                        else if (session.RequiresRefresh)
-                        {
-                            staleSessions.AddLast(session);
-                        }
-                        else
-                        {
-                            // The session is fine - put it back on the stack.
-                            stack.Push(session);
+                            for (int i = count - 1; i >= 0; i--)
```
jskeet: It's probably worth adding a comment here to explain what's going on, in terms of the order of elements in TryPopRange and what we're trying to achieve.
It's top-of-mind now, but it may not be in a year's time :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.cs
amanda-tarafa: Done

amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Yep, done.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Yep, done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Actually, undoing this, `TryPopRange` fails when the stack is empty. It calls [this](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentstack-1.trypoprange?view=netframework-4.8#System_Collections_Concurrent_ConcurrentStack_1_TryPopRange__0___System_Int32_System_Int32_) overload internally that fails [because `startIndex` is the same as the array length](https://referencesource.microsoft.com/#mscorlib/system/Collections/Concurrent/ConcurrentStack.cs,0c76dfd482eb5d09)

jskeet: Right, so we still need to avoid doing anything if the stack is empty, but the previous code would have failed too.
My point is we should use the return value of TryPopRange to know how many items we really need to handle.

amanda-tarafa: Agreed, I'm doing a combination of both codes. Will flag when done.

amanda-tarafa: @jskeet PTAL, thanks!!

PR #3588, opened 2019-10-08
Fix Language version

jskeet:
It should be 1.4.0-beta00 because we're not releasing it.

This will fix Travis, but we probably need to think about whether the detect-pr-changes.sh needs to be set to not run after a PR is merged.


jskeet:
Second part done in #3589.

PR #3589, opened 2019-10-08
Change the detect-pr-changes job to only run on PRs

jskeet:
When we create this PR, we should still see detect-pr-changes running, and it will fail until #3588 is merged.


jskeet:
Second part done in #3589.

jskeet:
Change detection still running, but broken, as expected. This is a start :)

jskeet:
Hmm. It's possible that this hasn't actually worked. We'll see over time.

PR #3609, opened 2019-10-10
Alternative approach for IAM v3 in GAPIC libraries

jskeet:


frankyn:
@jkwlui @jskeet and I discussed this option and leaning towards non moving forward with it to align language GAPIC and Manuals as close as possible.

Additional helper methods to make this easier to use can be addressed at a later time. 

Pending check off from @broady.

PR #3626, opened 2019-10-15
[Spanner] Adds support for retrying Spanner transactions on commit abort.

amanda-tarafa:
@jskeet just a quick look for now, sanity checking and that everything is where it should be. Taking a look a commit at a time might be helpfull.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
@@ -15,13 +15,11 @@
 using Google.Cloud.ClientTesting;
 using Google.Cloud.Spanner.Data.CommonTesting;
 using Google.Cloud.Spanner.V1;
-using Microsoft.Practices.EnterpriseLibrary.TransientFaultHandling;
 using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Linq;
 using System.Threading.Tasks;
-using System.Transactions;
 using Xunit;
 
 // All samples now use a connection string variable declared before the start of the snippet.
@@ -99,19 +97,24 @@ namespace Google.Cloud.Spanner.Data.Snippets
         {
             string connectionString = _fixture.ConnectionString;
 
-            await RetryHelpers.ExecuteWithRetryAsync(async () =>
+            // Sample: InsertDataAsync
+            // Additional: RunWithRetriableTransactionAsync
+            using (SpannerConnection connection = new SpannerConnection(connectionString))
             {
-                // Sample: InsertDataAsync
-                using (SpannerConnection connection = new SpannerConnection(connectionString))
-                {
-                    await connection.OpenAsync();
+                await connection.OpenAsync();
 
+                // If the transaction is aborted, RunWithRetriableTransactionAsync will
+                // retry the whole unit of work with a fresh transaction each time.
+                await connection.RunWithRetriableTransactionAsync(async transaction =>
+                {
                     SpannerCommand cmd = connection.CreateInsertCommand("TestTable");
                     SpannerParameter keyParameter = cmd.Parameters.Add("Key", SpannerDbType.String);
                     SpannerParameter stringValueParameter = cmd.Parameters.Add("StringValue", SpannerDbType.String);
                     SpannerParameter int64ValueParameter = cmd.Parameters.Add("Int64Value", SpannerDbType.Int64);
+                    cmd.Transaction = transaction;
 
-                    // This executes 5 distinct transactions with one row written per transaction.
+                    // This executes 5 distinct insert commands using the retriable transaction.
+                    // The mutations will be effective once the transaction has committed successfully.
                     for (int i = 0; i < 5; i++)
                     {
                         keyParameter.Value = Guid.NewGuid().ToString("N");
@@ -120,9 +123,9 @@ namespace Google.Cloud.Spanner.Data.Snippets
                         int rowsAffected = await cmd.ExecuteNonQueryAsync();
                         Console.WriteLine($"{rowsAffected} rows written...");
                     }
-                }
-                // End sample
-            });
+                });
+            }
+            // End sample
         }
 
         [Fact]
@@ -130,21 +133,26 @@ namespace Google.Cloud.Spanner.Data.Snippets
         {
             string connectionString = _fixture.ConnectionString;
 
-            await RetryHelpers.ExecuteWithRetryAsync(async () =>
+            // Sample: Dml
+            using (SpannerConnection connection = new SpannerConnection(connectionString))
             {
-                // Sample: Dml
-                using (SpannerConnection connection = new SpannerConnection(connectionString))
-                {
-                    await connection.OpenAsync();
+                await connection.OpenAsync();
 
+                // If the transaction is aborted, RunWithRetriableTransactionAsync will
+                // retry the whole unit of work with a fresh transaction each time.
+                // Please be aware that the whole unit of work needs to be prepared
+                // to be called more than once.
+                await connection.RunWithRetriableTransactionAsync(async transaction =>
+                {
                     SpannerCommand cmd = connection.CreateDmlCommand(
                         "UPDATE TestTable SET StringValue='Updated' WHERE Int64Value=@value");
                     cmd.Parameters.Add("value", SpannerDbType.Int64, 10);
+                    cmd.Transaction = transaction;
                     int rowsAffected = await cmd.ExecuteNonQueryAsync();
                     Console.WriteLine($"{rowsAffected} rows updated...");
-                }
-                // End sample
-            });
+                });
+            }
+            // End sample
         }
 
         [Fact]
@@ -174,14 +182,18 @@ namespace Google.Cloud.Spanner.Data.Snippets
         {
             string connectionString = _fixture.ConnectionString;
 
-            await RetryHelpers.ExecuteWithRetryAsync(async () =>
+            // Sample: BatchDml
+            using (SpannerConnection connection = new SpannerConnection(connectionString))
             {
-                // Sample: BatchDml
-                using (SpannerConnection connection = new SpannerConnection(connectionString))
-                {
-                    await connection.OpenAsync();
+                await connection.OpenAsync();
 
-                    SpannerBatchCommand cmd = connection.CreateBatchDmlCommand();
+                // If the transaction is aborted, RunWithRetriableTransactionAsync will
+                // retry the whole unit of work with a fresh transaction each time.
+                // Please be aware that the whole unit of work needs to be prepared
+                // to be called more than once.
+                await connection.RunWithRetriableTransactionAsync(async (transaction) =>
+                {
+                    SpannerBatchCommand cmd = transaction.CreateBatchDmlCommand();
 
                     cmd.Add(
                         "UPDATE TestTable SET StringValue='Updated' WHERE Int64Value=@value",
@@ -194,9 +206,9 @@ namespace Google.Cloud.Spanner.Data.Snippets
                     IEnumerable<long> rowsAffected = await cmd.ExecuteNonQueryAsync();
                     Console.WriteLine($"{rowsAffected.ElementAt(0)} rows updated...");
                     Console.WriteLine($"{rowsAffected.ElementAt(1)} rows deleted...");
-                }
-                // End sample
-            });
+                });
+            }
+            // End sample
         }
 
         [Fact]
@@ -204,57 +216,72 @@ namespace Google.Cloud.Spanner.Data.Snippets
         {
             string connectionString = _fixture.ConnectionString;
 
-            await RetryHelpers.ExecuteWithRetryAsync(async () =>
+            // Sample: CommitTimestamp
+            using (SpannerConnection connection = new SpannerConnection(connectionString))
             {
-                // Sample: CommitTimestamp
-                using (SpannerConnection connection = new SpannerConnection(connectionString))
+                await connection.OpenAsync();
+
+                string createTableStatement =
+                    @"CREATE TABLE UsersHistory (
+                    Id INT64 NOT NULL,
+                    CommitTs TIMESTAMP NOT NULL OPTIONS
+                        (allow_commit_timestamp=true),
+                    Name STRING(MAX) NOT NULL,
+                    Email STRING(MAX),
+                    Deleted BOOL NOT NULL,
+                    ) PRIMARY KEY(Id, CommitTs DESC)";
+
+                await connection.CreateDdlCommand(createTableStatement).ExecuteNonQueryAsync();
+
+                // Create a command for inserting rows.
+                SpannerCommand cmd = connection.CreateInsertCommand("UsersHistory",
+                    new SpannerParameterCollection
+                    {
+                        { "Id", SpannerDbType.Int64 },
+                        { "CommitTs", SpannerDbType.Timestamp, SpannerParameter.CommitTimestamp },
+                        { "Name", SpannerDbType.String },
+                        { "Deleted", SpannerDbType.Bool , false}
+                    });
+
+                int rowsAffected = 0;
+
+                // If the transaction is aborted, RunWithRetriableTransactionAsync will
+                // retry the whole unit of work with a fresh transaction each time.
+                // Please be aware that the whole unit of work needs to be prepared
+                // to be called more than once.
+                await connection.RunWithRetriableTransactionAsync(async transaction =>
                 {
-                    await connection.OpenAsync();
-
-                    string createTableStatement =
-                        @"CREATE TABLE UsersHistory (
-                        Id INT64 NOT NULL,
-                        CommitTs TIMESTAMP NOT NULL OPTIONS
-                            (allow_commit_timestamp=true),
-                        Name STRING(MAX) NOT NULL,
-                        Email STRING(MAX),
-                        Deleted BOOL NOT NULL,
-                      ) PRIMARY KEY(Id, CommitTs DESC)";
-
-                    await connection.CreateDdlCommand(createTableStatement).ExecuteNonQueryAsync();
-
                     // Insert a first row
-                    SpannerCommand cmd = connection.CreateInsertCommand("UsersHistory",
-                        new SpannerParameterCollection
-                        {
-                            { "Id", SpannerDbType.Int64, 10L },
-                            { "CommitTs", SpannerDbType.Timestamp, SpannerParameter.CommitTimestamp },
-                            { "Name", SpannerDbType.String, "Demo 1" },
-                            { "Deleted", SpannerDbType.Bool, false }
-                        });
-                    int rowsAffected = await cmd.ExecuteNonQueryAsync();
+                    cmd.Parameters["Id"].Value = 10L;
+                    cmd.Parameters["Name"].Value = "Demo 1";
+                    cmd.Transaction = transaction;
+                    rowsAffected += await cmd.ExecuteNonQueryAsync();
+                });
 
+                await connection.RunWithRetriableTransactionAsync(async transaction =>
+                {
                     // Insert a second row
                     cmd.Parameters["Id"].Value = 11L;
                     cmd.Parameters["Name"].Value = "Demo 2";
                     rowsAffected += await cmd.ExecuteNonQueryAsync();
-                    Console.WriteLine($"{rowsAffected} rows written...");
+                });
 
-                    // Display the inserted values
-                    SpannerCommand selectCmd = connection.CreateSelectCommand("SELECT * FROM UsersHistory");
-                    using (SpannerDataReader reader = await selectCmd.ExecuteReaderAsync())
+                Console.WriteLine($"{rowsAffected} rows written...");
+
+                // Display the inserted values
+                SpannerCommand selectCmd = connection.CreateSelectCommand("SELECT * FROM UsersHistory");
+                using (SpannerDataReader reader = await selectCmd.ExecuteReaderAsync())
+                {
+                    while (await reader.ReadAsync())
                     {
-                        while (await reader.ReadAsync())
-                        {
-                            long id = reader.GetFieldValue<long>("Id");
-                            string name = reader.GetFieldValue<string>("Name");
-                            DateTime timestamp = reader.GetFieldValue<DateTime>("CommitTs");
-                            Console.WriteLine($"{id}: {name} - {timestamp:HH:mm:ss.ffffff}");
-                        }
+                        long id = reader.GetFieldValue<long>("Id");
+                        string name = reader.GetFieldValue<string>("Name");
+                        DateTime timestamp = reader.GetFieldValue<DateTime>("CommitTs");
+                        Console.WriteLine($"{id}: {name} - {timestamp:HH:mm:ss.ffffff}");
                     }
                 }
-                // End sample
-            });
+            }
+            // End sample
         }
 
         [Fact]
@@ -262,15 +289,20 @@ namespace Google.Cloud.Spanner.Data.Snippets
         {
             string connectionString = _fixture.ConnectionString;
 
-            await RetryHelpers.ExecuteWithRetryAsync(async () =>
+            // Sample: ReadUpdateDeleteAsync
+            // Additional: CreateUpdateCommand
+            // Additional: CreateDeleteCommand
+            // Additional: CreateSelectCommand
+            using (SpannerConnection connection = new SpannerConnection(connectionString))
             {
-                // Sample: ReadUpdateDeleteAsync
-                // Additional: CreateUpdateCommand
-                // Additional: CreateDeleteCommand
-                // Additional: CreateSelectCommand
-                using (SpannerConnection connection = new SpannerConnection(connectionString))
+                await connection.OpenAsync();
+
+                // If the transaction is aborted, RunWithRetriableTransactionAsync will
+                // retry the whole unit of work with a fresh transaction each time.
+                // Please be aware that the whole unit of work needs to be prepared
+                // to be called more than once.
+                await connection.RunWithRetriableTransactionAsync(async (transaction) =>
                 {
-                    await connection.OpenAsync();
 
                     // Read the first two keys in the database.
                     List<string> keys = new List<string>();
@@ -294,62 +326,9 @@ namespace Google.Cloud.Spanner.Data.Snippets
                     SpannerCommand deleteCmd = connection.CreateDeleteCommand("TestTable");
                     deleteCmd.Parameters.Add("Key", SpannerDbType.String, keys[1]);
                     await deleteCmd.ExecuteNonQueryAsync();
-                }
-                // End sample
-            });
-        }
-
-        // Sample: SpannerFaultDetectionStrategy
-        private class SpannerFaultDetectionStrategy : ITransientErrorDetectionStrategy
-        {
-            /// <inheritdoc />
-            public bool IsTransient(Exception ex) => ex.IsTransientSpannerFault();
-        }
-        // End sample
-
-        [Fact]
-        public async Task TransactionAsync()
-        {
-            string connectionString = _fixture.ConnectionString;
-
-            // Sample: TransactionAsync
-            // Additional: BeginTransactionAsync
-            RetryPolicy<SpannerFaultDetectionStrategy> retryPolicy =
-                new RetryPolicy<SpannerFaultDetectionStrategy>(RetryStrategy.DefaultExponential);
-
-            await retryPolicy.ExecuteAsync(
-                async () =>
-                {
-                    using (SpannerConnection connection = new SpannerConnection(connectionString))
-                    {
-                        await connection.OpenAsync();
-
-                        using (SpannerTransaction transaction = await connection.BeginTransactionAsync())
-                        {
-                            SpannerCommand cmd = connection.CreateInsertCommand(
-                                "TestTable", new SpannerParameterCollection
-                                {
-                                    {"Key", SpannerDbType.String},
-                                    {"StringValue", SpannerDbType.String},
-                                    {"Int64Value", SpannerDbType.Int64}
-                                });
-                            cmd.Transaction = transaction;
-
-                            // This executes a single transactions with alls row written at once during CommitAsync().
-                            // If a transient fault occurs, this entire method is re-run.
-                            for (int i = 0; i < 5; i++)
-                            {
-                                cmd.Parameters["Key"].Value = Guid.NewGuid().ToString("N");
-                                cmd.Parameters["StringValue"].Value = $"StringValue{i}";
-                                cmd.Parameters["Int64Value"].Value = i;
-                                await cmd.ExecuteNonQueryAsync();
-                            }
-
-                            await transaction.CommitAsync();
-                        }
-                    }
                 });
-            // End sample
+                // End sample
+            }
         }
 
         [Fact]
@@ -447,50 +426,6 @@ namespace Google.Cloud.Spanner.Data.Snippets
                     Console.WriteLine(stats);
                 }
             }
-            // End sample
-        }
-
-        [Fact]
-        public async Task TransactionScopeAsync()
-        {
-            string connectionString = _fixture.ConnectionString;
-
-            // Sample: TransactionScopeAsync
-            // Additional: CreateInsertCommand
-            RetryPolicy<SpannerFaultDetectionStrategy> retryPolicy =
-                new RetryPolicy<SpannerFaultDetectionStrategy>(RetryStrategy.DefaultExponential);
-
-            await retryPolicy.ExecuteAsync(
-                async () =>
-                {
-                    using (TransactionScope scope = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled))
-                    {
-                        using (SpannerConnection connection = new SpannerConnection(connectionString))
-                        {
-                            await connection.OpenAsync();
-
-                            SpannerCommand cmd = connection.CreateInsertCommand(
-                                "TestTable", new SpannerParameterCollection
-                                {
-                                    {"Key", SpannerDbType.String},
-                                    {"StringValue", SpannerDbType.String},
-                                    {"Int64Value", SpannerDbType.Int64}
-                                });
-
-                            // This executes a single transactions with alls row written at once during scope.Complete().
-                            // If a transient fault occurs, this entire method is re-run.
-                            for (int i = 0; i < 5; i++)
-                            {
-                                cmd.Parameters["Key"].Value = Guid.NewGuid().ToString("N");
-                                cmd.Parameters["StringValue"].Value = $"StringValue{i}";
-                                cmd.Parameters["Int64Value"].Value = i;
-                                await cmd.ExecuteNonQueryAsync();
-                            }
-
-                            scope.Complete();
-                        }
-                    }
-                });
             // End sample
         }
 
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/RetriableTransactionOptionsTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/RetriableTransactionOptionsTests.cs
@@ -1,0 +1,64 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Linq;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    public class RetriableTransactionOptionsTests
+    {
+        [Fact]
+        public void DefaultJitter()
+        {
+            var options = RetriableTransactionOptions.CreateDefault();
+
+            var delay = TimeSpan.FromTicks(TimeSpan.TicksPerMillisecond);
+            var delayTicks = (int)delay.Ticks;
+            // The jitter on RetriableTransactionOptions should jitter between
+            // delay and delay * 2.
+            var expectedMinDelayTicks = delayTicks;
+            var expectedMaxDelayTicks = delayTicks * 2;
+            var expectedMedianDelayTicks = (expectedMinDelayTicks + expectedMaxDelayTicks) / 2;
+
+            var delays = Enumerable.Range(0, 10_000)
+                .Select(_ => options.Jitter(delay))
+                .ToList();
+            var underMedianCount = delays.LongCount(ts => ts.Ticks <= expectedMedianDelayTicks);
+
+            Assert.InRange(underMedianCount, 4_000, 6_000);
+        }
+
+        [Fact]
+        public void NoJitter()
+        {
+            var options = RetriableTransactionOptions.CreateWithNoJitter();
+            var delay = TimeSpan.FromTicks(TimeSpan.TicksPerMillisecond);
+            var delayTicks = (int)delay.Ticks;
+            var expectedJitteredTicks = delayTicks;
+
+            // This shouldn't jitter at all, 50_000 seems a good enough sample
+            // to test that.
+            var delays = Enumerable.Range(0, 50_000)
+                .Select(_ => options.Jitter(delay))
+                .ToList();
+            var min = delays.Min();
+            var max = delays.Max();
+
+            Assert.Equal(expectedJitteredTicks, min.Ticks);
+            Assert.Equal(expectedJitteredTicks, max.Ticks);
+        }
+    }
+}
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
@@ -1,0 +1,328 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax.Testing;
+using Google.Cloud.Spanner.Common.V1;
+using Google.Cloud.Spanner.V1;
+using Google.Cloud.Spanner.V1.Internal.Logging;
+using Google.Cloud.Spanner.V1.Tests;
+using Grpc.Core;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    public partial class SpannerConnectionTests
+    {
+        public sealed class TransactionAbortedRetryTests
+        {
+            // Combining backoff and jittering for retries
+            // when the RpcException doesn't specify a wait time.
+            // InitialDelay = 250ms, BackoffMultiplier = 2, JitterMultiplier = 2
+            // Those are all hard-coded, client code can't change them.
+            private const int InitialDelayBackoffMs = 250;
+            private const int BackoffMultiplier = 2;
+            private const int JitterMultiplier = 2;
+
+            private const int ExceptionRetryDelayMs = 300;
+            // We respect the retry from the Exception
+            private const int ExceptionBackoffMultiplier = 1;
+
+            [Fact]
+            public async Task FirstCallSucceeds()
+            {
+                var spannerClientMock = SpannerClientHelpers
+                    .CreateMockClient(Logger.DefaultLogger, MockBehavior.Strict)
+                    .SetupBatchCreateSessionsAsync()
+                    .SetupBeginTransactionAsync()
+                    .SetupExecuteBatchDmlAsync()
+                    .SetupCommitAsync();
+                SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+
+                var scheduler = (FakeScheduler)connection.Builder.SessionPoolManager.SpannerSettings.Scheduler;
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var result = await connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction));
+                    Assert.Equal(1, result);
+                });
+
+                callee.AssertBackoffTimesInRange(time0);
+                callee.AssertLastCallTime(scheduler.Clock.GetCurrentDateTimeUtc());
+            }
+
+            [Fact]
+            public async Task CommitAbortsTwice()
+            {
+                var spannerClientMock = SpannerClientHelpers
+                    .CreateMockClient(Logger.DefaultLogger, MockBehavior.Strict)
+                    .SetupBatchCreateSessionsAsync()
+                    .SetupBeginTransactionAsync()
+                    .SetupExecuteBatchDmlAsync()
+                    .SetupCommitAsync_Fails(failures: 2, statusCode: StatusCode.Aborted)
+                    .SetupRollbackAsync();
+
+                SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+
+                var scheduler = (FakeScheduler)connection.Builder.SessionPoolManager.SpannerSettings.Scheduler;
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var result = await connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction));
+                    Assert.Equal(3, result);
+                });
+
+                callee.AssertBackoffTimesInRange(time0);
+                callee.AssertLastCallTime(scheduler.Clock.GetCurrentDateTimeUtc());
+            }
+
+            // The transaction can be aborted on commit or on any of the operations.
+            [Fact]
+            public async Task BatchDmlAbortsTwice()
+            {
+                var spannerClientMock = SpannerClientHelpers
+                    .CreateMockClient(Logger.DefaultLogger, MockBehavior.Strict)
+                    .SetupBatchCreateSessionsAsync()
+                    .SetupBeginTransactionAsync()
+                    .SetupExecuteBatchDmlAsync_Fails(failures: 2, statusCode: StatusCode.Aborted)
+                    .SetupCommitAsync()
+                    .SetupRollbackAsync();
+
+                SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+
+                var scheduler = (FakeScheduler)connection.Builder.SessionPoolManager.SpannerSettings.Scheduler;
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var result = await connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction));
+                    Assert.Equal(3, result);
+                });
+
+                callee.AssertBackoffTimesInRange(time0);
+                callee.AssertLastCallTime(scheduler.Clock.GetCurrentDateTimeUtc());
+            }
+
+            [Fact]
+            public async Task CommitAbortsTwice_RecommendedDelay()
+            {
+                var spannerClientMock = SpannerClientHelpers
+                    .CreateMockClient(Logger.DefaultLogger, MockBehavior.Strict)
+                    .SetupBatchCreateSessionsAsync()
+                    .SetupBeginTransactionAsync()
+                    .SetupExecuteBatchDmlAsync()
+                    .SetupCommitAsync_Fails(failures: 2, statusCode: StatusCode.Aborted, exceptionRetryDelay: TimeSpan.FromMilliseconds(ExceptionRetryDelayMs))
+                    .SetupRollbackAsync();
+
+                SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+
+                var scheduler = (FakeScheduler)connection.Builder.SessionPoolManager.SpannerSettings.Scheduler;
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler, initialDelayMs: ExceptionRetryDelayMs, backoffMultiplier: ExceptionBackoffMultiplier);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var result = await connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction));
+                    Assert.Equal(3, result);
+                });
+
+                callee.AssertBackoffTimesInRange(time0);
+                callee.AssertLastCallTime(scheduler.Clock.GetCurrentDateTimeUtc());
+            }
+
+            [Fact]
+            public async Task CommitAbortsAlways_RespectsOverallDeadline()
+            {
+                var spannerClientMock = SpannerClientHelpers
+                    .CreateMockClient(Logger.DefaultLogger, MockBehavior.Strict)
+                    .SetupBatchCreateSessionsAsync()
+                    .SetupBeginTransactionAsync()
+                    .SetupExecuteBatchDmlAsync()
+                    .SetupCommitAsync_FailsAlways(statusCode: StatusCode.Aborted)
+                    .SetupRollbackAsync();
+
+                SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+
+                var scheduler = (FakeScheduler)connection.Builder.SessionPoolManager.SpannerSettings.Scheduler;
+                // This test needs a little bit more of real time, else it's flaky.
+                scheduler.RealTimeTimeout = TimeSpan.FromSeconds(60);
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var exception = await Assert.ThrowsAsync<SpannerException>(() => connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction)));
+                    Assert.True(exception.IsRetryable && !exception.SessionExpired);
+                    Assert.Contains("Bang!", exception.InnerException.Message);
+                });
+
+                // The minimum calls that can be made (assuming maximum jitter always) in that time is 60 * 60 / 32 * 2 which is 56.25,
+                // plus 1 because the first one has no delay.
+                // The maximum number of calls that can be made in 1 hour:
+                // - Given that 2^7 * 250 = 32000, the first 8 calls take, at a minimum 31_750ms, approx 32s.
+                // - The rest will happen at most (60 * 60 - 32) / 32 which is 111.5
+                callee.AssertCalledInRange(57, 120);
+                // The overall deadline is of 1 hour. The maximum backoff delay is of 32s.
+                // Because of jitter retries can stop at any time after 60mins - 64s, let's give it 2 minutes of range.
+                Assert.InRange(scheduler.Clock.GetCurrentDateTimeUtc(), time0.AddMinutes(58), time0.AddMinutes(60));
+            }
+
+            [Fact]
+            public async Task CommitFailsOtherThanAborted()
+            {
+                var spannerClientMock = SpannerClientHelpers
+                    .CreateMockClient(Logger.DefaultLogger, MockBehavior.Strict)
+                    .SetupBatchCreateSessionsAsync()
+                    .SetupBeginTransactionAsync()
+                    .SetupExecuteBatchDmlAsync()
+                    .SetupCommitAsync_Fails(failures: 1, StatusCode.Unknown)
+                    .SetupRollbackAsync();
+
+                SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+
+                var scheduler = (FakeScheduler)connection.Builder.SessionPoolManager.SpannerSettings.Scheduler;
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var exception = await Assert.ThrowsAsync<SpannerException>(
+                        () => connection.RunWithRetriableTransactionAsync(
+                            transaction => callee.DatabaseWorkAsync(connection, transaction)));
+                    Assert.Contains("Bang!", exception.InnerException.Message);
+                });
+
+                callee.AssertBackoffTimesInRange(time0);
+                callee.AssertLastCallTime(scheduler.Clock.GetCurrentDateTimeUtc());
+            }
+
+            [Fact]
+            public async Task WorkFails()
+            {
+                var spannerClientMock = SpannerClientHelpers
+                    .CreateMockClient(Logger.DefaultLogger, MockBehavior.Strict)
+                    .SetupBatchCreateSessionsAsync()
+                    .SetupBeginTransactionAsync()
+                    .SetupRollbackAsync();
+
+                SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+
+                var scheduler = (FakeScheduler)connection.Builder.SessionPoolManager.SpannerSettings.Scheduler;
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var exception = await Assert.ThrowsAsync<InvalidOperationException>(() => connection.RunWithRetriableTransactionAsync(callee.Fails));
+                    Assert.Contains("Bang!", exception.Message);
+                });
+
+                callee.AssertBackoffTimesInRange(time0);
+                callee.AssertLastCallTime(scheduler.Clock.GetCurrentDateTimeUtc());
+            }
+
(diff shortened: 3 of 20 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerException.cs
```diff
@@ -84,6 +86,25 @@ public bool IsRetryable
             }
         }
 
+        /// <summary>
+        /// For retryable <see cref="SpannerException"/>, this value represents
+        /// the recommended minimum retry delay. It might or might not
+        /// be present, even if this exeption is retryable.
```
jskeet: Nit in case I miss it later: exeption => exception

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerException.cs
```diff
@@ -84,6 +86,25 @@ public bool IsRetryable
...
+        /// </summary>
+        public TimeSpan? RecommendedRetryDelay { get; }
+
+        /// <summary>
+        /// For retrying aborted commits, we need to diferentiate whether the operation
+        /// was aborted due to a session expiry. If the session expired we cannot
```
jskeet: Can we not retry, or can we just not retry with the same session? I thought it was just more efficient if we reuse the sesssion.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerException.cs
```diff
@@ -101,7 +122,20 @@ internal SpannerException(ErrorCode code, RpcException innerException)
...
+
+            SessionExpired = innerException.IsSessionExpiredError();
             ErrorCode = innerException.IsSessionExpiredError() ? ErrorCode.Aborted : code;
+
+            Metadata.Entry retryInfoEntry = innerException.Trailers.FirstOrDefault(
+                entry => s_retryInfoMetadataKey.Equals(entry.Key, StringComparison.InvariantCultureIgnoreCase));
+            if (retryInfoEntry != null)
+            {
+                RetryInfo retryInfo = RetryInfo.Parser.ParseFrom(retryInfoEntry.ValueBytes);
+                if (retryInfo?.RetryDelay?.Nanos > 0)
```
jskeet: Why do we not want to use a retry delay of 0? (This may well be moving code I wrote before, but we should check :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerRetryHelpers.cs
```diff
@@ -0,0 +1,75 @@
...
+            while (true)
+            {
+                try
+                {
+                    var result = await fn().ConfigureAwait(false);
+                    return result;
+                }
+                catch (SpannerException e) when (retryOptions.ExceptionFilter(e))
+                {
+                    // If there's a recommended retry dely specified on the exception
```
jskeet: dely => delay

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -143,6 +199,26 @@ internal static PooledSession FromSessionName(SessionPool.ISessionPool pool, Ses
         /// </remarks>
         public void Dispose() => ReleaseToPool(false);
 
+        /// <summary>
+        /// Usually disposing of a <see cref="PooledSession"/> instance should release its
+        /// underlying session to the pool. But when we create a new instance of <see cref="PooledSession"/>
+        /// from an existing instance of <see cref="PooledSession"/>, that is, with the same underlying session,
+        /// the original instance shouldn't be allowed to be disposed off normally so that the underlying session
```
jskeet: Nit: off => of

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -412,8 +415,37 @@ private async Task<PooledSession> BeginTransactionAsync(PooledSession session, T
...
+                else
+                {
+                    try
+                    {
+                        // Let's try to begin a new transaction for this same session.
+                        return BeginTransactionAsync(session, transactionOptions, cancellationToken, true);
+                    }
+                    catch(RpcException e)
+                    {
+                        Parent._logger.Warn("Failed to create transaction for acquiered session", e);
```
jskeet: Nit: acquiered => acquired

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -412,8 +415,37 @@ private async Task<PooledSession> BeginTransactionAsync(PooledSession session, T
...
+                        // Failed to create the transaction for the session.
+                        // Release the session back to the pool.
```
jskeet: Possibly add that we'll have another go with another session?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -412,8 +415,37 @@ private async Task<PooledSession> BeginTransactionAsync(PooledSession session, T
...
+                        session.ReleaseToPool(false);
+                    }
                 }
+                // If we are here we need to acquiere a new session.
```
jskeet: Nit: acquiere => acquire

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -90,6 +90,21 @@ public override string ConnectionString
         [Category("Data")]
         public string SpannerInstance => Builder.SpannerInstance;
 
+        /// <summary>
```
jskeet: We should probably ask the Spanner team about this. There's too much I don't know.

If we expose it at all though, I'd be tempted to make it a parameter on the method instead of having it as a property. (With a suitable default if null is passed in.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -261,6 +276,76 @@ public SpannerTransaction BeginReadOnlyTransaction(TransactionId transactionId)
...
+        /// </summary>
+        /// <param name="asyncWork">The work to perform in each transaction attempt.</param>
+        /// <param name="cancellationToken">An optional token for canceling the call.</param>
+        /// <returns>The value returned by <paramref name="asyncWork"/> if the transaction commits successfully.</returns>
+        /// <remarks><paramref name="asyncWork"/> will be fully run whenever the commit on 
+        /// the <see cref="SpannerTransaction"/> that it receives as a parameter aborts. It must be prepared
+        /// to be called more than once.
+        /// A new <see cref="SpannerTransaction"/> will be passed to <paramref name="asyncWork"/>
+        /// each time it is rerun.
+        /// <paramref name="asyncWork"/> doens't need to handle the lifecycle of the <see cref="SpannerTransaction"/>,
```
jskeet: Nit: doens't => doesn't

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -261,6 +276,76 @@ public SpannerTransaction BeginReadOnlyTransaction(TransactionId transactionId)
...
+                        {
+                            await transaction.RollbackAsync(cancellationToken).ConfigureAwait(false);
+                            // Throw, the retry helper will know when to retry.
+                            throw;
+                        }
+                        finally
+                        {
+                            if (transaction != null)
+                            {
+                                // Let's make sure that the associated session is not released to the pool
```
jskeet: Why? I may be missing something.

amanda-tarafa:
Thannks! Will address comments and questions, and get ready for merge. Will flag when I'm done for a full review.

amanda-tarafa:
Travis failed because the compatibility checker failed to install dotnet. Rerunning.

amanda-tarafa:
This is now ready for review. A few notes:

- Reviewing one commit at a time might be helpful.
- I din't update the integration tests in DMLTests.cs to use the new RunWithRetriableTransaction... method. I created a new RetriableTransactionTests.cs instead. I figured that even though the recommended way to do writes is with a retriable transaction, client code can still use implicit transactions or manage their own transactions so it is worth it to keep those tests as well.
- I did update all the snippets that were writing data so that they'd use the RunWithRetriableTransaction... methods. Those are the docs, and we should show the recommended way of using read-write transactions.
- I removed the snippets having to do with System.Transactions.TransactionScope since that's not compatible with the recommended RunWithRetriableTransaction... The integrations tests for TransactionScope are unchanged.

amanda-tarafa:
Checking CI now.

amanda-tarafa:
Checking CI now.

amanda-tarafa:
CI is now green.

jskeet:
Okay, I'm re-reviewing from scratch, one commit at a time :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerException.cs
```diff
@@ -84,6 +86,25 @@ public bool IsRetryable
...
+        /// the recommended minimum retry delay. It might or might not
+        /// be present, even if this exception is retryable.
+        /// </summary>
+        public TimeSpan? RecommendedRetryDelay { get; }
+
+        /// <summary>
+        /// For retrying aborted commits, we need to diferentiate whether the operation
+        /// was aborted due to a session expiry. If the session expired we cannot
+        /// retry automatically because we cannot reuse the session.
+        /// Note that it is us who are setting the <see cref="ErrorCode"/> to <see cref="ErrorCode.Aborted"/>
```
jskeet: I think this part would be better as a non-XML comment, just in case this property is ever made public.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerException.cs
jskeet: Okay, I think I agree :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerRetryHelpers.cs
```diff
@@ -0,0 +1,109 @@
...
+                    {
+                        throw;
+                    }
+                    await scheduler.Delay(actualDelay, cancellationToken).ConfigureAwait(false);
+                    retryDelay = retryOptions.BackoffSettings.NextDelay(retryDelay);
+                }
+            }
+        }
+
+        private class MinimumBoundJitter : IJitter
```
jskeet: I don't think it's a good idea to implement `IJitter` here. It doesn't fulfill the same contract, as is evident by the fact that the parameter is `minDelay` whereas in `IJitter` it's `maxDelay`. I think just a simple method (passing in `retryOptions.DelayJitterMultiplier`) would be better. Apologies for having missed this before.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerRetryHelpers.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// Note that this is not the same as the retrying mechanism for each
+    /// of the RPCs, but for a higher level kind of operation that may include
+    /// one or several RPCs.
+    /// A good example of a higher level operation is the retrying of a 
+    /// transaction commit on abort, which requieres the retrying of all 
+    /// the transaction work and of the commit.
+    /// TODO(atarafamas, jskeet): Consider making public or entirely eliminating once we know 
+    /// what we can reuse from GAX retry helpers.
+    /// </summary>
+    internal static class SpannerRetryHelpers
```
jskeet: Just as a heads-up, we'll be changing some of the retry code in GAX soon. Maybe we want to bear this (and some Firestore code) in mind when doing that, to make it more amenable to reuse.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerRetryHelpers.cs
```diff
@@ -0,0 +1,109 @@
...
+                try
+                {
+                    var result = await fn().ConfigureAwait(false);
+                    return result;
+                }
+                catch (SpannerException e) when (retryOptions.ExceptionFilter(e))
+                {
+                    // If there's a recommended retry delay specified on the exception
+                    // we should at least wait for that.
+                    if (e.RecommendedRetryDelay.HasValue && e.RecommendedRetryDelay.Value > retryDelay)
```
jskeet: It's not clear to me whether we need any JITter on top of that at all.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerRetryHelpers.cs
```diff
@@ -0,0 +1,109 @@
...
+                    return minDelay;
+                }
+
+                double jitterMultiplier;
+                lock (_lock)
+                {
+                    jitterMultiplier = s_random.NextDouble();
+                }
+
+                var minDelayTicks = minDelay.Ticks;
```
jskeet: I don't think I fully understand what this code is meant to do, but I suspect it can be simplified. It'll be clearer with more comments about the purpose of the class/method.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerRetryOptions.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace Google.Cloud.Spanner.Data
+{
+    /// <summary>
+    /// Options to use with <see cref="SpannerRetryHelpers"/>.
+    /// TODO: Consider making public or entirely eliminating once we know 
+    /// what we can reuse from GAX retry helpers.
+    /// </summary>
+    internal sealed class SpannerRetryOptions
+    {
+        internal SpannerRetryOptions(
```
jskeet: Do we need all of this anyway? For example, do we ever really need to change the exception filter?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerRetryHelpersTests.cs
```diff
@@ -0,0 +1,249 @@
...
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Xunit;
+using Status = Grpc.Core.Status;
+
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    public class SpannerRetryHelpersTests
```
jskeet: Note: I haven't looked through the tests carefully, partly because I suspect they'll change based on other comments.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
```diff
@@ -521,6 +509,80 @@ public async Task AcquireAsync_NursesBackToHealth_OriginalCallerCancelled()
...
+            [Fact(Timeout = TestTimeoutMilliseconds)]
+            public async Task RefreshTransaction_FailsForNone()
+            {
+                var pool = CreatePool(false);
+                var client = (SessionTestingSpannerClient)pool.Client;
+
+                await client.Scheduler.RunAsync(async () =>
+                {
+                    var originalSession = await pool.AcquireSessionAsync(
+                        new TransactionOptions (),
```
jskeet: Nit: remove space before `()`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SessionPoolTests.TargetedSessionPoolTests.cs
```diff
@@ -822,13 +884,18 @@ public async Task ShutdownPoolAsync_CannotAcquireAfterShutdown()
...
-                var tasks = Enumerable
-                    .Range(0, pool.Options.MaximumActiveSessions)
-                    .Select(_ => pool.AcquireSessionAsync(new TransactionOptions(), default))
-                    .ToList();
+                var tasks = AcquireSessionTasks(pool, pool.Options.MaximumActiveSessions);
                 await Task.WhenAll(tasks);
                 return tasks.Select(t => t.Result).ToList();
             }
+
+            private List<Task<PooledSession>> AcquireSessionTasks(TargetedSessionPool pool, int sessionsToAcquiere)
```
jskeet: Optional nit: use expression-bodied member?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -58,6 +58,24 @@ public sealed class PooledSession : IDisposable
         /// </summary>
         internal ModeOneofCase TransactionMode { get; }
 
+        internal TransactionOptions TransactionOptions
```
jskeet: Optional nit: might want to make this a method, given that it creates a new instance each time.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.ISessionPool.cs
```diff
@@ -27,6 +27,7 @@ internal interface ISessionPool
         {
             SpannerClient Client { get; }
             IClock Clock { get; }
+            Task<PooledSession> WithFreshTransactionOrNewAsync(PooledSession session, TransactionOptions transactionOptions, CancellationToken cancellationToken);
```
jskeet: I'm glad ISessionPool is internal, otherwise this would have been a breaking change :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -427,11 +427,14 @@ private async Task TryCreateReadWriteTransactionAndReturnToPool(PooledSession se
                 ReleaseInactiveSession(session, maybeCreateReadWriteTransaction: false);
             }
 
-            private async Task<PooledSession> BeginTransactionAsync(PooledSession session, TransactionOptions options, CancellationToken cancellationToken)
+            private async Task<PooledSession> BeginTransactionAsync(PooledSession session, TransactionOptions options, CancellationToken cancellationToken, bool isSessionAcquired = false)
```
jskeet: I don't fully understand what this parameter is yet. It's a little odd that nothing else has changed in the method, apart from whether or not the book-keeping. Maybe add a comment explaining this more?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -443,8 +446,38 @@ private async Task<PooledSession> BeginTransactionAsync(PooledSession session, T
...
+                    if (!isSessionAcquired)
+                    {
+                        Interlocked.Decrement(ref _inFlightSessionCreationCount);
+                    }
+                }
+            }
+
+            public override Task<PooledSession> WithFreshTransactionOrNewAsync(PooledSession session, TransactionOptions transactionOptions, CancellationToken cancellationToken)
+            {
+                if(session.RequiresRefresh)
```
jskeet: Nit: space after `if`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -261,6 +274,104 @@ public SpannerTransaction BeginReadOnlyTransaction(TransactionId transactionId)
         public Task<SpannerTransaction> BeginTransactionAsync(CancellationToken cancellationToken = default) =>
             BeginTransactionImplAsync(s_readWriteTransactionOptions, TransactionMode.ReadWrite, cancellationToken);
 
+        /// <summary>
+        /// Executes a read-write transaction, with retries as necessary.
```
jskeet: Possibly clarify in "remarks" what's meant by "as necessary"? (It's only retrying certain things.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -261,6 +274,104 @@ public SpannerTransaction BeginReadOnlyTransaction(TransactionId transactionId)
...
+        /// to be called more than once.
+        /// A new <see cref="SpannerTransaction"/> will be passed to <paramref name="asyncWork"/>
+        /// each time it is rerun.
+        /// <paramref name="asyncWork"/> doesn't need to handle the lifecycle of the <see cref="SpannerTransaction"/>,
+        /// it will be automatically committed after <paramref name="asyncWork"/> has finished or rollbacked if an 
+        /// <see cref="Exception"/> (other than because the transaction commit aborted) is thrown by <paramref name="asyncWork"/>.</remarks>
+        public async Task RunWithRetriableTransactionAsync(Func<SpannerTransaction, Task> asyncWork, CancellationToken cancellationToken = default)
+        {
+            GaxPreconditions.CheckNotNull(asyncWork, nameof(asyncWork));
+            await RunWithRetriableTransactionAsync(async (transaction) =>
```
jskeet: Nit on style: remove parentheses from `(transaction)`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -261,6 +274,104 @@ public SpannerTransaction BeginReadOnlyTransaction(TransactionId transactionId)
...
+                    Builder.SessionPoolManager.SpannerSettings.Clock ?? SystemClock.Instance,
+                    Builder.SessionPoolManager.SpannerSettings.Scheduler ?? SystemScheduler.Instance,
+                    cancellationToken).ConfigureAwait(false);
+            }
+            finally
+            {
+                session?.Dispose();
+            }
+
+            async Task<TResult> CommitAttempt()
```
jskeet: It feels like this could be in SpannerRetryHelpers. It's possible there's too much context here to easily move it there, but it would be nice to make the retry helper class be more helpful :) That can definitely happen in a later PR though - maybe add a TODO for now?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -261,6 +274,104 @@ public SpannerTransaction BeginReadOnlyTransaction(TransactionId transactionId)
...
+        /// it will be automatically committed after <paramref name="asyncWork"/> has finished or rollbacked if an 
+        /// <see cref="Exception"/> (other than because the transaction commit aborted) is thrown by <paramref name="asyncWork"/>.</remarks>
+        public async Task<TResult> RunWithRetriableTransactionAsync<TResult>(Func<SpannerTransaction, Task<TResult>> asyncWork, CancellationToken cancellationToken = default)
+        {
+            GaxPreconditions.CheckNotNull(asyncWork, nameof(asyncWork));
+            PooledSession session = null;
+            try
+            {
+                await OpenAsync(cancellationToken).ConfigureAwait(false);
+                return await SpannerRetryHelpers.ExecuteWithRetryAsync(
```
jskeet: I think it would be worth noting in here that `CommitAttempt` modifies the `session` variable, otherwise this looks a bit odd - it looks like session will always be null.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -99,17 +97,20 @@ public async Task InsertDataAsync()
...
             {
-                // Sample: InsertDataAsync
-                using (SpannerConnection connection = new SpannerConnection(connectionString))
-                {
-                    await connection.OpenAsync();
+                await connection.OpenAsync();
 
+                // If the transaction is aborted, RunWithRetriableTransactionAsync will
+                // retry the whole unit of work with a fresh transaction each time.
+                await connection.RunWithRetriableTransactionAsync(async (transaction) =>
```
jskeet: Nit: remove parentheses from `(transaction)`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -99,17 +97,20 @@ public async Task InsertDataAsync()
...
+                {
                     SpannerCommand cmd = connection.CreateInsertCommand("TestTable");
                     SpannerParameter keyParameter = cmd.Parameters.Add("Key", SpannerDbType.String);
                     SpannerParameter stringValueParameter = cmd.Parameters.Add("StringValue", SpannerDbType.String);
                     SpannerParameter int64ValueParameter = cmd.Parameters.Add("Int64Value", SpannerDbType.Int64);
+                    cmd.Transaction = transaction;
```
jskeet: The comment below this line is now incorrect, isn't it?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -294,61 +315,46 @@ public async Task ReadUpdateDeleteAsync()
...
-
-                            await transaction.CommitAsync();
-                        }
+                            {"Key", SpannerDbType.String},
+                            {"StringValue", SpannerDbType.String},
+                            {"Int64Value", SpannerDbType.Int64}
+                        });
+                    cmd.Transaction = transaction;
+
+                    // This executes a single transactions with alls row written at once during CommitAsync().
```
jskeet: alls row => all rows

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -294,61 +315,46 @@ public async Task ReadUpdateDeleteAsync()
...
-        public async Task TransactionAsync()
+        public async Task RetriableTransactionAsync()
         {
             string connectionString = _fixture.ConnectionString;
 
-            // Sample: TransactionAsync
-            // Additional: BeginTransactionAsync
-            RetryPolicy<SpannerFaultDetectionStrategy> retryPolicy =
-                new RetryPolicy<SpannerFaultDetectionStrategy>(RetryStrategy.DefaultExponential);
+            // Sample: RetriableTransactionAsync
```
jskeet: Is this now just the same as InsertDataAsync?

apis/Google.Cloud.Spanner.Data/docs/index.md
```diff
@@ -128,29 +146,20 @@ Cloud Spanner is fully ACID compliant.
...
-{{sample:SpannerConnection.SpannerFaultDetectionStrategy}}
-
- You should execute the entire transaction within a RetryPolicy. Generally, we recommend
- using exponential backoff as shown below.
+Please read the [Cloud Spanner user documentation](https://cloud.google.com/spanner/docs/reference/rest/v1/TransactionOptions)
+for details on when and why transactions can be aborted.
 
 {{sample:SpannerConnection.TransactionAsync}}
 
-## TransactionScope support (.NET 4.5+, Single resource only)
```
jskeet: Let's talk offline about this being removed. I think it's reasonable to say we still support TransactionScope, but that it won't retry transactions automatically. If we don't want to support it at all any more, maybe we should think about removing it for the upcoming breaking change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerRetryOptions.cs
amanda-tarafa: For retrying aborted commints:

- We shouldn't ever change the exception filter.
- As per discussion with the Spanner team, the backoff config is not modifiable by client code either.

But, I was trying to write a more general use kind of retry helper in case there are more "high level" retries we needed/wanted to offer, something that in the future, if needed across libraries, we could consider moving to Gax making the appropiate changes.

That said, I have no problem hardwiring some or all of this, but then I'd also like to have the retry method be specifically for aborted transactions and not for anything else. It would be weird to have the general method use the hardwire config (for instance, for the filter) that only really makes sense for aborted transactions.

Or should we consider modifying Gax to extract general retry logic that can be reused on a per call retry and also on higher level retries?

jskeet: My experience is that there tends to be enough that's "just a bit different" in each application-level retry that it's not worth trying to encapsulate too much of it.
If there are aspects of GAX that can be easily exposed, I'm open to that.

amanda-tarafa:
@jskeet this is ready for review.
The first and last commit are mostly nit and minor fixes.
The second commit contains most new changes, since we agreed on retry logic specific for aborted transactions.
PTAL.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -143,6 +181,26 @@ internal static PooledSession FromSessionName(SessionPool.ISessionPool pool, Ses
...
+        /// is not released back to the pool.
+        /// </summary>
+        /// <returns><code>true</code> if the <see cref="PooledSession"/> had not been disposed before.
+        /// <code>false</code> otherwise.</returns>
+        private bool MarkAsDisposed()
+        {
+            bool wasDisposed = Interlocked.Exchange(ref _disposed, 1) == 1;
+            if (!wasDisposed)
+            {
+                GC.SuppressFinalize(this);
```
jskeet: I don't think we even have a finalizer now - this may have been left over from a time that we did. Not suggesting a change right now, but we may be able to simplify this method to a single expression.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -114,10 +114,48 @@ internal static PooledSession FromSessionName(SessionPool.ISessionPool pool, Ses
...
+        /// This method will always try to get a fresh transaction for this session.
+        /// If the session has expired or it fails to get a fresh transaction, then it will
+        /// acquire a session in the normal way.
+        /// This <see cref="PooledSession"/> instance will be disposed of to ensure that all operations
+        /// with the underlying session are done through the new instance.
+        /// </summary>
+        /// <returns>A new instance of <see cref="PooledSession"/>.</returns>
+        /// <exception cref="InvalidOperationException">If this <see cref="PooledSession.TransactionMode"/>
+        /// is <see cref="ModeOneofCase.None"/>.</exception>
+        /// <remarks>
```
jskeet: Nit: I'd move the remarks to just below summary.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -295,6 +351,21 @@ internal Task<Transaction> BeginTransactionAsync(BeginTransactionRequest request
             return RecordSuccessAndExpiredSessions(Client.BeginTransactionAsync(request, callSettings));
         }
 
+        internal TransactionOptions GetTransactionOptions()
```
jskeet: These won't always be complete options - for example, I think read-only transactions usually have a staleness bound. (There may be other things too.)
Not sure how much of a problem this is. We should probably talk about this. Also, could this method be private?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
jskeet: Yes, better thanks.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -99,19 +97,24 @@ public async Task InsertDataAsync()
...
+                {
                     SpannerCommand cmd = connection.CreateInsertCommand("TestTable");
                     SpannerParameter keyParameter = cmd.Parameters.Add("Key", SpannerDbType.String);
                     SpannerParameter stringValueParameter = cmd.Parameters.Add("StringValue", SpannerDbType.String);
                     SpannerParameter int64ValueParameter = cmd.Parameters.Add("Int64Value", SpannerDbType.Int64);
+                    cmd.Transaction = transaction;
 
-                    // This executes 5 distinct transactions with one row written per transaction.
+                    // This executes 5 distinct insert commands using the retriable transaction.
+                    // The mutations will be effective once the transaction has commited successfully.
```
jskeet: Nit: commited => committed

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -120,31 +123,36 @@ public async Task InsertDataAsync()
...
-                using (SpannerConnection connection = new SpannerConnection(connectionString))
-                {
-                    await connection.OpenAsync();
+                await connection.OpenAsync();
 
+                // If the transaction is aborted, RunWithRetriableTransactionAsync will
+                // retry the whole unit of work with a fresh transaction each time.
+                // Please be aware that the whole unit of work needs to be prepared
+                // to be called more than once.
+                await connection.RunWithRetriableTransactionAsync(async (transaction) =>
```
jskeet: Nit: take transaction out of parentheses?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -120,31 +123,36 @@ public async Task InsertDataAsync()
...
+                {
                     SpannerCommand cmd = connection.CreateDmlCommand(
-                        "UPDATE TestTable SET StringValue='Updated' WHERE Int64Value=@value");
+                    "UPDATE TestTable SET StringValue='Updated' WHERE Int64Value=@value");
```
jskeet: It's worth reindenting this, I think.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -194,44 +206,49 @@ public async Task BatchDml()
...
+                    Deleted BOOL NOT NULL,
+                    ) PRIMARY KEY(Id, CommitTs DESC)";
+
+                await connection.CreateDdlCommand(createTableStatement).ExecuteNonQueryAsync();
+
+                // If the transaction is aborted, RunWithRetriableTransactionAsync will
+                // retry the whole unit of work with a fresh transaction each time.
+                // Please be aware that the whole unit of work needs to be prepared
+                // to be called more than once.
+                await connection.RunWithRetriableTransactionAsync(async (transaction) =>
```
jskeet: Again, remove parentheses around the parameter.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets/SpannerConnectionSnippets.cs
```diff
@@ -242,6 +259,7 @@ public async Task CommitTimestampAsync()
 
                     // Display the inserted values
                     SpannerCommand selectCmd = connection.CreateSelectCommand("SELECT * FROM UsersHistory");
+                    selectCmd.Transaction = transaction;
```
jskeet: I don't know whether we really want to do the display part within the same transaction. It's not even clear to me whether the first two should be in the same transaction.
Aside from anything else, if they haven't been committed yet, I don't know what the commit timestamp would be... I suspect we should have two retriable bits, then execute the SELECT not in a transaction at all. That should show us two different commit timestamps.

apis/Google.Cloud.Spanner.Data/docs/index.md
```diff
@@ -77,8 +83,14 @@ visible to all subsequent statements in the batch.
 When one of the statemens in the batch fails, execution is halted and all subsequent statements are not
 executed. A `SpannerBatchNonQueryException` will be thrown that contains both information about the error
 and the number of rows affected by each of the statements that executed before the failed one. If you are executing
-the batch DML command inside a transaction you can simply commit the transaction when `SpannerBatchNonQueryException`
+the batch DML command inside a non retriable transaction you can simply commit the transaction when `SpannerBatchNonQueryException`
```
jskeet: Nit: hyphenate non-retriable

apis/Google.Cloud.Spanner.Data/docs/index.md
```diff
@@ -128,28 +146,19 @@ Cloud Spanner is fully ACID compliant.
...
+can be aborted due to causing a deadlock. In this case, the entire transaction should be
+retried. The methods `SpannerConnection.RunWithRetriableTransactionAsync(...)` will take care of
+automatically retrying the entire callback passed as a paremeter using a fresh transaction
+if previous attempts aborted.
 
-Use the extension method IsTransientSpannerFault() on Exception to create your detection
-strategy for the TransientFaultApplication Block.
+- The callback doesn't need to handle the transaction lifecycle.
+- The callback needs to set the transaction received on all the commands it creates.
+- The callback might be called more than once so it needs to be prepared for it.
```
jskeet: Possibly "prepared" is an unfortunate term, due to the idea of preparing statements etc. How about:

"The callback might be called more than once, so the code should be written bearing this in mind."

apis/Google.Cloud.Spanner.Data/docs/index.md
jskeet: Let's make sure we do that. I suspect we really do need to keep TransactionScope support - which is in netstandard2.0, so we don't need to say ".NET 4.5+" any more - but that we need to emphasize the retry aspect.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    public partial class SpannerConnectionTests
+    {
+        public sealed class TransactionAbortedRetryTests
+        {
+            // Combining backoff and jittering for retries
+            // when the RpcException doesn't specify a wait time.
+            // InitialDelay = 250ms, BackoffMultiplier = 2, JitterMultiplier = 2
+            // Those are all hardwire, client code can't change them.
```
jskeet: Nit: hardwire => hardwired or hard-coded perhaps.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+            private const int SecondCallLowBo = 250;
```
jskeet: Possibly expand "Bo" to "Backoff"? It took me a little while to work out what was meant.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var result = await connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction));
+                    Assert.Equal(1, result);
+                });
+
+                callee.AssertCallTimesInrange(new Tuple<DateTime, DateTime>(time0, time0));
```
jskeet: If you want to use ValueTuple rather than Tuple, I'm fine to add a dependency. I'm also happy to commit as it is, then I'll have a look at simplifying in another PR :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var result = await connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction));
+                    Assert.Equal(3, result);
+                });
+
+                callee.AssertCallTimesInrange(
+                    new Tuple<DateTime, DateTime>(time0, time0),
+                    new Tuple<DateTime, DateTime>(time0.AddMilliseconds(SecondCallLowBo), time0.AddMilliseconds(SecondCallHighBo)),
```
jskeet: If you created TimeSpan equivalents of the constants, you could just use `time0 + SecondCallLowBackoff` etc. Get rid of 18 calls to AddMilliseconds :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var result = await connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction));
+                    Assert.Equal(3, result);
+                });
+
+                callee.AssertCallTimesInrange(
```
jskeet: I wonder whether this might be refactored to AssertBackoffTimesInRange or somthing like that?
In particular, that could be stricter - because the third time will depend on the second time, if you see what I mean. (We can chat about this.)

jskeet: All of this could be done in another PR though :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+                var scheduler = (FakeScheduler)connection.Builder.SessionPoolManager.SpannerSettings.Scheduler;
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var exception = await Assert.ThrowsAsync<SpannerException>(() => connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction)));
+                    Assert.True(exception.IsRetryable && !exception.SessionExpired);
+                    Assert.Contains("Bang!", exception.InnerException.Message);
+
```
jskeet: Nit: remove empty line.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+                });
+
+                // The overall deadline is of 1 hour. The maximum backoff delay is of 32s.
+                // Because of jitter retries can stop at any time after 60mins - 64s, let's give it 2 minutes of range.
+                Assert.InRange(scheduler.Clock.GetCurrentDateTimeUtc(), time0.AddMinutes(58), time0.AddMinutes(60));
```
jskeet: Can we check that there's been at least 110 (or whatever the theoretical minimum is) calls?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+
+                SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+
+                var scheduler = (FakeScheduler)connection.Builder.SessionPoolManager.SpannerSettings.Scheduler;
+                var time0 = scheduler.Clock.GetCurrentDateTimeUtc();
+                var callee = new Callee(scheduler);
+
+                await scheduler.RunAsync(async () =>
+                {
+                    var exception = await Assert.ThrowsAsync<SpannerException>(() => connection.RunWithRetriableTransactionAsync(transaction => callee.DatabaseWorkAsync(connection, transaction)));
```
jskeet: Nit: possibly wrap the line just after `ThrowsAsync<SpannerException>(`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+                    return _callTimes.Count;
+                }
+
+                public Task<int> Fails(SpannerTransaction transaction)
+                {
+                    _callTimes.Add(_scheduler.Clock.GetCurrentDateTimeUtc());
+                    throw new InvalidOperationException("Bang!");
+                }
+
+                public void AssertCallTimesInrange(params Tuple<DateTime, DateTime>[] times)
```
jskeet: Nit: capitalize the R of Range - but we may change how all of this works anyway :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -0,0 +1,337 @@
...
+                {
+                    for (int i = 0; i < times.Length; i++)
```
jskeet: Should we check that `times` is equal to `_callTimes.Count`?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SpannerClientHelpers.cs
```diff
@@ -12,30 +12,225 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Api.Gax.Grpc;
```
jskeet: In order to speed up this review, I'm not reviewing this file right now.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
```diff
@@ -0,0 +1,135 @@
...
+
+        internal RetriableTransaction(SpannerConnection connection, IClock clock, IScheduler scheduler, RetriableTransactionOptions options = null)
+        {
+            _connection = GaxPreconditions.CheckNotNull(connection, nameof(connection));
+            _clock = GaxPreconditions.CheckNotNull(clock, nameof(clock));
+            _scheduler = GaxPreconditions.CheckNotNull(scheduler, nameof(scheduler));
+            _options = options ?? RetriableTransactionOptions.CreateDefault();
+        }
+
+        public async Task<TResult> RunAsync<TResult>(Func<SpannerTransaction, Task<TResult>> asyncWork, CancellationToken cancellationToken = default)
```
jskeet: Maybe make the method internal, given that it's effectively internal anyway?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
```diff
@@ -0,0 +1,135 @@
...
+                        {
+                            session = await (session?.WithFreshTransactionOrNewAsync(cancellationToken) ?? _connection.AcquireReadWriteSessionAsync(cancellationToken)).ConfigureAwait(false);
+                            transaction = new SpannerTransaction(_connection, TransactionMode.ReadWrite, session, null);
+
+                            TResult result = await asyncWork(transaction).ConfigureAwait(false);
+                            await transaction.CommitAsync(cancellationToken).ConfigureAwait(false);
+
+                            return result;
+                        }
+                        catch when (transaction != null)
```
jskeet: It took me a little while to work out how transaction *could* be null. Maybe add a comment?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
```diff
@@ -0,0 +1,135 @@
...
+                        {
+                            try
+                            {
+                                await transaction.RollbackAsync(cancellationToken).ConfigureAwait(false);
+                            }
+                            catch (Exception e)
+                            {
+                                // If the commit actually aborted, calling Rollback will fail.
```
jskeet: I don't suppose we can tell whether that's the case and avoid even trying the rollback?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
```diff
@@ -0,0 +1,135 @@
...
+                                // We don't want that or any other rollback exception to propogate,
```
jskeet: Nit: propogate => propagate

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
```diff
@@ -0,0 +1,135 @@
...
+
+                            // Throw, the retry helper will know when to retry.
+                            throw;
+                        }
+                        finally
+                        {
+                            if (transaction != null)
+                            {
+                                // Let's make sure that the associated session is not released to the pool
+                                // because we'll be attempting to get a fresh transaction for this same session first.
```
jskeet: A few concerns here:

- Don't we want to release the session if the transaction has succeeded? We'll still be executing this code.
- What about if the transaction failed, but we're not going to retry either due to deadlines or it not being an abort?

jskeet: Ah, my guess is that both of these are handled by the final `session?.Dispose()` call - please add a comment to avoid silly people like me getting confused again :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
```diff
@@ -0,0 +1,135 @@
...
+                try
+                {
+                    var result = await fn().ConfigureAwait(false);
+                    return result;
+                }
+                catch (SpannerException e) when (e.IsRetryable && !e.SessionExpired)
+                {
+                    // If there's a recommended retry delay specified on the exception
+                    // we should respect it.
+                    if (e.RecommendedRetryDelay.HasValue)
```
jskeet: How about

```csharp
retryDelay = e.RecommendedRetryDelay ?? retryDelay;
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
```diff
@@ -0,0 +1,135 @@
...
+                    }
+
+                    TimeSpan actualDelay = _options.Jitter(retryDelay);
+                    DateTime expectedRetryTime = _clock.GetCurrentDateTimeUtc() + actualDelay;
+                    if (expectedRetryTime > overallDeadline)
+                    {
+                        throw;
+                    }
+                    await _scheduler.Delay(actualDelay, cancellationToken).ConfigureAwait(false);
+                    retryDelay = _options.NextDelay(retryDelay);
```
jskeet: If we took a recommended delay, do we definitely want to double *that* rather than "the delay we'd have been using otherwise"?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransactionOptions.cs
```diff
@@ -0,0 +1,71 @@
...
+                maxDelay: TimeSpan.FromSeconds(32),
+                delayMultiplier: 2);
+        private static readonly Expiration s_defaultOverallExpiry = Expiration.FromTimeout(TimeSpan.FromHours(1));
+
+        internal readonly Func<TimeSpan, TimeSpan> _jitter;
+
+        private RetriableTransactionOptions(Func<TimeSpan, TimeSpan> jitter = null) =>
+            _jitter = jitter ?? MinimumBoundJitter;
+
+        public static RetriableTransactionOptions CreateDefault() => new RetriableTransactionOptions();
```
jskeet: Make all of these internal for clarity?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransactionOptions.cs
```diff
@@ -0,0 +1,71 @@
...
+using Google.Api.Gax;
+using Google.Api.Gax.Grpc;
+using System;
+
+namespace Google.Cloud.Spanner.Data
+{
+    internal sealed class RetriableTransactionOptions
+    {
+        private static readonly BackoffSettings s_defaultBackoff = new BackoffSettings(
+                delay: TimeSpan.FromMilliseconds(250),
```
jskeet: Nit: outdent by 4

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -261,6 +260,103 @@ public SpannerTransaction BeginReadOnlyTransaction(TransactionId transactionId)
...
             BeginTransactionImplAsync(s_readWriteTransactionOptions, TransactionMode.ReadWrite, cancellationToken);
 
+        /// <summary>
+        /// Executes a read-write transaction, with retries as necessary.
+        /// The work to perform in each transaction attempt is defined by <paramref name="asyncWork"/>.
+        /// </summary>
+        /// <param name="asyncWork">The work to perform in each transaction attempt.</param>
+        /// <param name="cancellationToken">An optional token for canceling the call.</param>
+        /// <returns>The value returned by <paramref name="asyncWork"/> if the transaction commits successfully.</returns>
+        /// <remarks><paramref name="asyncWork"/> will be fully retried whenever the <see cref="SpannerTransaction"/>
```
jskeet: Nit: move remarks to just after summary (and in the method below.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerException.cs
```diff
@@ -101,7 +119,20 @@ internal SpannerException(ErrorCode code, RpcException innerException)
...
+
+            SessionExpired = innerException.IsSessionExpiredError();
             ErrorCode = innerException.IsSessionExpiredError() ? ErrorCode.Aborted : code;
+
+            Metadata.Entry retryInfoEntry = innerException.Trailers.FirstOrDefault(
+                entry => s_retryInfoMetadataKey.Equals(entry.Key, StringComparison.InvariantCultureIgnoreCase));
+            if (retryInfoEntry != null)
+            {
+                RetryInfo retryInfo = RetryInfo.Parser.ParseFrom(retryInfoEntry.ValueBytes);
+                if (retryInfo?.RetryDelay?.Nanos >= 0)
```
jskeet: I suspect this was copied from some of my code somewhere, but I think it's slightly broken - if the server specifies a whole number of seconds, this won't pick that up.
I'd convert it to a TimeSpan unconditionally, and only set the RecommendedRetryDelay if that's not TimeSpan.Zero.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
amanda-tarafa: No, we really can't tell if it was the commit that failed or not, or we could but with uglier code than this I think. I did changed the comment though, because rollback will fail if the commit was attempted and failed, but not only on aborted but other possible failures.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
amanda-tarafa: Done :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
amanda-tarafa: Yep!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
amanda-tarafa: Apparently, having a retry backoff configuration is just for being on the very safe side, because on aborted the server will always recommend a delay.
I'm fine either way, but although I couldn't swear on it right now, I think I did this because it's what Java was doing.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransactionOptions.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransactionOptions.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerException.cs
amanda-tarafa: Done. I'll look the code that was doing this then, because it probably needs to be fixed as well.

amanda-tarafa:
All comments addressed on the last 3 commits (I will squash those into the first 3 later). I'm running tests locally. PTAL.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/RetriableTransactionOptionsTests.cs
```diff
@@ -24,38 +24,44 @@ public class RetriableTransactionOptionsTests
...
             var options = RetriableTransactionOptions.CreateDefault();
+
             var delay = TimeSpan.FromTicks(TimeSpan.TicksPerMillisecond);
+            var delayTicks = (int)delay.Ticks;
+            // The jitter on RetriableTransactionOptions should jitter between
+            // delay and delay * 2.
+            var expectedMinDelayTicks = delayTicks;
+            var expectedMaxDelayTicks = delayTicks * 2;
+            var possibleDelayValues = expectedMaxDelayTicks - expectedMinDelayTicks;
+            var expectedMedianDelayTicks = expectedMinDelayTicks + possibleDelayValues / 2;
```
jskeet: I think this would be simpler as:

```csharp
var expectedMedianDelayTicks = (expectedMinDelayTicks + expectedMaxDelayTicks) / 2;
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/RetriableTransactionOptionsTests.cs
```diff
@@ -24,38 +24,44 @@ public class RetriableTransactionOptionsTests
...
 
-            var delays = Enumerable.Range(0, (int)TimeSpan.TicksPerMillisecond)
+            var delays = Enumerable.Range(0, possibleDelayValues)
```
jskeet: This is still using a value which is fundamentally "a number of ticks" as "the number of iterations to perform" - they don't seem related to me, or at least, I don't understand why they would be.

It feels like the number of iterations could be any reasonably large number (even just 1000) - all we're checking is that at least one value is in the lower half and at least one value is in the upper half. I may be missing something though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
```diff
@@ -183,12 +181,17 @@ public async Task CommitAbortsAlways_RespectsOverallDeadline()
...
                     Assert.Contains("Bang!", exception.InnerException.Message);
-
                 });
 
                 // The overall deadline is of 1 hour. The maximum backoff delay is of 32s.
                 // Because of jitter retries can stop at any time after 60mins - 64s, let's give it 2 minutes of range.
                 Assert.InRange(scheduler.Clock.GetCurrentDateTimeUtc(), time0.AddMinutes(58), time0.AddMinutes(60));
+                // The minimum calls that can be made (assuming maximum jitter always) in that time is 60 * 60 / 32 * 2 which is 56.25,
+                // plus 1 because the first one has no delay.
+                // The maximum number of calls tha can be made in 1 hour:
```
jskeet: Nit: tha => that, and on the line below, aprox => approx

amanda-tarafa:
> As discussed, we should make this only valid for ReadWrite transactions. (It would be okay for PartitionedDml as well, as that has no extra details.)
> We should probably think about what we'd do if TransactionOptions.ReadWrite ever gained any fields though... one simple option would be to make this method _accept_ a TransactionOptions, which we could check was of the same type as the existing one. It may be simpler to do that now rather than later.

I've changed the method to take the parameter and do the check. Added a test for that as well.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/RetriableTransactionOptionsTests.cs
amanda-tarafa: Changed!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/RetriableTransactionOptionsTests.cs
amanda-tarafa: > This is still using a value which is fundamentally "a number of ticks" as "the number of iterations to perform" - they don't seem related to me, or at least, I don't understand why they would be.

But ticks is as granular as TimeSpan would go. So, when we jitter, the quantity of possible values we can ever get is always the number of ticks between the jitter min and max bounds.
In a perfect random distribution we would get each of the possible values if we had max - min iterations, and that's why, yes, this is fundamentally a number of ticks, but it represents the amount of different values so it's ok to use it as the number of iterations to be performed. What is true is that, if the random generator is good, this tests is really loose, and we could probably do with way less iterations, or with being way more strict on our checks.

I would leave it as is, but have no problem with putting just a number there, LMK.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionTests.AbortedTransactionRetryTests.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/RetriableTransactionOptionsTests.cs
jskeet: I'd definitely prefer to have just a number. Even with a perfect random number generator, you wouldn't necessarily get each value exactly once - just like if you roll a fair die 6 times, you wouldn't always expect to get 1, 2, 3, 4, 5, 6.
(To put it another way - if that *were* the case, the last sample *wouldn't* be random, because you'd be able to predict the exact value - that would be weird!)

It might be reasonable to take, say, 10,000 samples and assert that between 4,000 and 6,000 of them should be in the lower half of the range - that would be a reasonable test that it's probably genuinely linear.

amanda-tarafa: Yep, true, it would be a perfect distribution, but not random :).
Changing the test now.

amanda-tarafa:
Rerunning Travis, it failed because it couldn't download something from Nuget.

amanda-tarafa:
Last rebase, will wait for green before merging.

PR #3634, opened 2019-10-17
Migrate Redis to the micro-generator

jskeet:
This will show up in the Travis log as a major change due to effectively reverting the addition of FailoverInstance calls with nullable parameters. However, it's *not* a major change compared with the released package. (Validated locally; will be validated again in CI before a new release.)


jskeet:
Ah yes - I need to add an import for the smoke test, just in V1Beta1. Easy fix.

PR #3639, opened 2019-10-17
Regenerate all micro-generated snippets

jskeet:
Most of these are just changes in case, but a few (e.g.
Translate.V3) involve properly filling in generics, and List
snippets now have the right references.


jskeet:
(Assigned to myself as this is just generated code.)

PR #3646, opened 2019-10-22
Update OS Login versions to indicate the end of 1.x

jskeet:
Also add how we're doing this to our process document.

Alternative suggestions for how to handle this are welcome. We
*could* just use 2.0.0-beta00, but it might be too easy to miss that
when doing the "upgrade everything" change later.


jskeet:
(Would like feedback from both of you if possible.)

PROCESSES.md
```diff
@@ -92,6 +92,13 @@ nothing else, for clarity. Include both the `apis.json` change and
 the project file changes that occur when the project generator has
 been run, in the same commit.
 
+As a final very rare case, a version number such as
+"1.99.99-alpha00" is used to indicate that a breaking change has
```
amanda-tarafa: X.99.99-alpha00??

PROCESSES.md
```diff
@@ -92,6 +92,13 @@ nothing else, for clarity. Include both the `apis.json` change and
...
+already been made, and the intention is for the next release to be a
+new major version - but that there are still significant changes
```
amanda-tarafa: "a new major version (X+1)"?

apis/Google.Cloud.OsLogin.V1Beta/.repo-metadata.json
```diff
@@ -1,4 +1,4 @@
 {
   "distribution_name": "Google.Cloud.OsLogin.V1Beta",
-  "release_level": "beta"
+  "release_level": "ga"
```
amanda-tarafa: This is now a little bit confusing, I get it, it's just confusing.

jskeet: Eek, that shouldn't be there. Will need to fix that.

PROCESSES.md
jskeet: I think it's easier to keep it concrete - the "such as" makes it clear that it's an example, I reckon. Let's see what Chris thinks.

apis/Google.Cloud.OsLogin.V1Beta/.repo-metadata.json
jskeet: Ah, this is tricky, actually. The repo metadata generator can't know what's *been* released when all it knows is what *won't* be released. It's assuming that because there's a 1.99, there must have been a 1.98.
Humbug. Will need to think about this more.

amanda-tarafa: Yes, I was just going to add a comment around this, that is that we don't do 1.1.99-beta00 instead we do 1.2.00-beta00, so why can't we do 2.0.0-alpha00?

jskeet: We could do that - we'd just need to be really careful not to then bump it to 3.0.0 when we do the major version bump.

I would also note that with 2.0.0-alpha00, this would probably *still* say "Ah, there's a GA available". I think we probably need an override for that.

jskeet:
Maybe we're over-thinking this, and this is such a rare situation that we should just "know" not to release OS Login for now. The release process would fail if we tried anyway, given the breaking changes...

amanda-tarafa:
Agree.

jskeet:
Right, closing this PR. I think it was worth experimenting with, but it's clearly too complicated to be worth it.

PR #3648, opened 2019-10-22
Add instance ctor compatibility. And fix method-comparer bug

chrisdunelm:

PR #3652, opened 2019-10-23
Fix snippets by specifying which TopicName we need

jskeet:


amanda-tarafa:
(Travis failed, but when installing dotnetcore)

chrisdunelm:
I've restarted travis.

This is blocking merging #3648 ;)

PR #3655, opened 2019-10-23
Use the microgenerator for Scheduler.V1

jskeet:


jskeet:
In terms of "things still to do" here for the major version bump:

- LocationName is "local" to this package
- There are the obsolete settings
- A one-line patch is required to allow TopicName.Parse to compile

PR #3657, opened 2019-10-23
Regenerate all APIs (other than OS Login) with new property handling

jskeet:
(This generated surprisingly few changes.)


jskeet:
Will merge on green if the change level is "identical" as I expect

PR #3658, opened 2019-10-23
Move OS Login to the microgenerator

jskeet:
Commits 1 and 3 are manual; 2 and 4 are generated.


chrisdunelm:
This currently has a lot of major changes, which I wasn't expecting as it's had a GA release.
Am I misunderstanding something?

jskeet:
This is the API we *don't* plan to release before the major version bump. The expectation is that all the breaking changes are due to the resource names being renamed; I'll check through them tomorrow to make sure that's the case.

Issue #3667, opened 2019-10-25
Storage: Add support for Archive storage class


tritone:
GCS is adding a storage class called "archive" to the [existing options for object storage](https://cloud.google.com/storage/docs/storage-classes) (coldline, standard, etc). Support for this should be added to this client library. This has already been implemented in the C++ client library and there is a PR in python; reference PR [here](https://github.com/googleapis/google-cloud-python/issues/9532)

This should be straightforward as ARCHIVE should just mirror existing valid storage classes. I see the following places where changes are required in order to accomplish this:
* https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClasses.cs (add new storage class).
* https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/StorageClassTest.cs (consider adding Archive class to the test somehow).

@jskeet let me know who you think would make sense to take this one (I'm also happy to try and pick it up myself since it's simple, but it would also be my first work in C#, hah).

amanda-tarafa:
We are reverting #3674 until the feature becomes available in the backend.
Please reopen this issue when that happens.

tritone:
Hi @jskeet and @amanda-tarafa , thanks for your patience on this. We are now no longer blocked on this! If you could re-open the PR, that would be great. We would want to merge and release this change sometime during the week from 1/2-1/8 (once the release freeze has ended).

jskeet:
Neither myself nor Amanda are working until January 2nd, but we'll sort it then.

tritone:
Sounds good, thanks! (and hope you're having a nice break!)

jskeet:
This is now released in Google.Cloud.Storage.V1 version 2.5.0.

tritone:
Fantastic; thanks so much @jskeet and @amanda-tarafa !

PR #3674, opened 2019-10-28
[Storage] Adds support for Archive storage class.

amanda-tarafa:
Addresses #3667.


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClasses.cs
```diff
@@ -55,5 +55,10 @@ public static class StorageClasses
         /// and better performance/availability.
         /// </summary>
         public const string DurableReducedAvailability = "DURABLE_REDUCED_AVAILABILITY";
+
```
tritone: Let's re-sort these as follows: Standard, Nearline, Coldine, Archive. Then Regional, Multi-regional, and Durable Reduced Availability.

The latter 3 are all considered legacy/deprecated at this point, so putting them at the bottom and adding comments to that effect (that they are accepted but not recommended for new implementations) would be great.

PR #3677, opened 2019-10-29
Update storage conformance tests and fix slash-escaping issue

jskeet:


apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/Proto/Tests.cs
```diff
@@ -908,7 +908,7 @@ public sealed partial class GetTest : pb::IMessage<GetTest> {
       }
       if (other.request_ != null) {
         if (request_ == null) {
```
amanda-tarafa: I know these are all generated, but it's a little "weird" to use request_ for the null check and the Request for assigning the value. For these and all the others.

jskeet: Yup, this has changed many times over the course of the protobuf lifetime.

PR #3681, opened 2019-10-30
Migrate Spanner to the microgenerator

jskeet:
(From a branch; do not merge yet.)


jskeet:
This was generated against a separate branch of googleapis - I've just triggered propagation of a change to github to the master branch; will wait until that's gone through and regenerate (which should be a no-op) before merging.

jskeet:
(Done, and it is indeed a no-op. Yay!)

PR #3686, opened 2019-10-31
samplegen: implement ApplicationChooser

hzyi-google:
- Implement an `ApplicationChooser` with extensive copy and paste from Jon's `DemoCode` repo (non-interactive mode only)
- Update the project generator to generator project files for samples that reference the `ApplicationChooser`, as well as configures the `StartupObject` to `Google.Cloud.SomeApi.V1.Samples.Program`
- An upcoming PR in gapic-generator to generate the `Program.cs` per API will come shortly

### DIFF
--- a/tools/Google.Cloud.SampleUtil/ApplicationChooser.cs
+++ b/tools/Google.Cloud.SampleUtil/ApplicationChooser.cs
@@ -21,8 +21,23 @@ using System.Threading.Tasks;
 
 namespace Google.Cloud.SampleUtil
 {
+    /// <summary>
+    /// Class allowing a user to choose a sample to run, after
+    /// examining an assembly for all classes containing a static Main method, either
+    /// parameterless or with a string array parameter.
+    /// The ApplicationChooser is designed to run in both interactive mode (not 
+    /// implemented yet) and non-interactive mode.
+    /// In the interactive mode, the options are presented in a list ordered 
+    /// first by those with a Description attribute (by that description) and 
+    /// then those without (ordered by name).
+    /// In the non-interactive mode, the ApplicationChooser will invoke the sample
+    /// with the sample file name (the first positional commandline argument).
+    /// </summary>
     public class ApplicationChooser
     {
+        /// <summary>
+        /// Looks up a sample based on its name, and invoke it with the remaining arguments.
+        /// </summary>	
         public static void Run(string[] args)
         {
             if (args.Length == 0)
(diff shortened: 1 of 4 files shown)


tools/Google.Cloud.SampleUtil/Google.Cloud.SampleUtil.csproj
```diff
@@ -0,0 +1,15 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
hzyi-google: I have no idea about what needs to be included in this file... PTAL

hzyi-google:
@jskeet PTAL
cc/ @googleapis/samplegen

jskeet:
I'm on vacation tomorrow, but will look at this on Monday.

tools/Google.Cloud.SampleUtil/ApplicationChooser.cs
```diff
@@ -0,0 +1,125 @@
...
+    public class ApplicationChooser
+    {
+        /// <summary>
+        /// Looks up a sample based on its name, and invoke it with the remaining arguments.
+        /// </summary>	
+        public static void Run(string[] args)
+        {
+            if (args.Length == 0)
+            {
+                throw new NotImplementedException("Interative mode isn't implemented yet.");
```
jskeet: Nit: Interative => Interactive

tools/Google.Cloud.SampleUtil/ApplicationChooser.cs
```diff
@@ -0,0 +1,125 @@
...
+            string sample = args[0];
+            var assembly = Assembly.GetEntryAssembly();
+
+            // Create a dictionary, so that we can easily look up a sample entry point based on
+            // the filename of the sample.
+            var samples = assembly
+                          .DefinedTypes
+                          .Select(t => GetEntryPoint(t))
+                          .Where(ep => ep != null && ep != assembly.EntryPoint)
+                          .ToDictionary(ep => ep.DeclaringType.Name.Replace("Main", string.Empty), ep => ep);
```
jskeet: I think it might be better to write a method that trims "Main" from the end if it exists - otherwise this could cause problems if we ever have samples to do with "Maintenance" for example.

tools/Google.Cloud.SampleUtil/Google.Cloud.SampleUtil.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <PackageId>Google.Cloud.SampleUtil</PackageId>
```
jskeet: No need for this one. The description isn't really *necessary* either, but I think it's worth keeping.

Please add

```xml
<IsPackable>false</IsPackable>
```

Other than that, this looks fine.

tools/Google.Cloud.SampleUtil/ApplicationChooser.cs
hzyi-google: Good point!

Issue #3675, opened 2019-10-29
Proposal: remove EF Core provider for Spanner


jskeet:
We currently have an alpha version of an Entity Framework Core provider for Spanner, along with a GA version of an ADO.NET provider.

The EF Core provider has not been reviewed and maintained as thoroughly as the ADO.NET provider, and probably needs significant effort to bring it from alpha to beta and then GA. At the moment, we have no plans to put in that effort.

Proposal:

- Delist the packages in NuGet
- Remove the source code from GitHub in a single commit (so we can easily revert it later if we decide to bring the project back)

cc @meteatamel, @JustinBeckwith, @chrisdunelm, @SurferJeffAtGoogle, @jsimonweb, @amanda-tarafa 

(Note: unless there are any objections, I plan to implement this on Friday November 8th.)

SurferJeffAtGoogle:
The proposal SGTM.

jskeet:
This has now been done.

PR #3703, opened 2019-11-06
Remove Spanner Entity Framework Core provider source

jskeet:
Implements proposal #3675.

(This PR will only be merged if there are no objections on that proposal by Friday.)


jskeet:
The CI failures are expected - our CI scripts try to test "the projects that have changed" and aren't geared up for project deletion. It's so rare that it's not worth doing.

Issue #3705, opened 2019-11-07
FirestoreProperty: allow ignoring null values


kinex:
I would like to prevent serializing null values when using `DocumentReference.SetAsync` to keep documents cleaner and to save storage space. Something like this would be useful:
```c#
[FirestoreProperty(nullValueHandling: NullValueHandling.Ignore)]
public string Name { get; set; }
```
Is there any way to do this currently?

amanda-tarafa:
Hi @kinex, you can create your own custom converter to handle this.
Take a look at the docs on [Costum Converters](http://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.Firestore/datamodel.html#custom-converters).

kinex:
Hi @amanda-tarafa, thanks for the idea. I noticed documentation says "conversion methods never receive null references". Anyway, I tried this since ignoring also empty string would be possibly ok in my case:

```c#
public class IgnoreNullConverter : IFirestoreConverter<string>
{
    public object ToFirestore(string value)
    {
        if (string.IsNullOrEmpty(value))
        {
            return FieldValue.Delete;
        }
        return value;
    }

    public string FromFirestore(object value)
    {
        return value as string;
    }
}
```
But this throws exception `System.ArgumentException: 'Delete cannot appear in document data when overwriting Arg_ParamName_Name'`. I guess the exception refers to the name parameter which I am providing to FirestoreProperty. I cannot remove that parameter so it seems this approach does not work.

amanda-tarafa:
I'm writing a sample now. But basically you need to write a [Costum Document Converter](http://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.Firestore/datamodel.html#document-converters) (sorry I thought I had been explicit about that but I see I wasn't). As you well point out other Custom Converters don't support null values.

amanda-tarafa:
@kinex Take a look at #3707 which should be of help. I understand though that if you want to ignore null values for all string properties in all your models, you might need to end up writing several document converters and that might be cumbersome.

kinex:
@amanda-tarafa Thanks for the code sample! That is useful at least for simple projects. In my case there are around 20 data classes where I would like to ignore null values for specific fields which may be strings or references to some other classes.

I actually started implementing a generic version of a converter based on your code sample, but it started to look too complex risking both performance and stability. So I decided to give up and keep the null values for now. 

[...]

PR #3707, opened 2019-11-07
[Firestore] Snippet for writing a Document Converter that ignores null values.

amanda-tarafa:
Although it doesn't fully addresses #3705 it might help.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Snippets/DataModelSnippets.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Snippets/DataModelSnippets.cs
@@ -273,6 +273,78 @@ namespace Google.Cloud.Firestore.Snippets
         }
         // End sample
 
+        // Sample: ComplexConverterIgnoresNull
+        [FirestoreData(ConverterType = typeof(CustomPersonConverter))]
+        public class CustomPerson
+        {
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Snippets/DataModelSnippets.cs
```diff
@@ -273,6 +273,78 @@ public CustomCity FromFirestore(object value)
...
+            }
+
+            public CustomPerson FromFirestore(object value)
+            {
+                if (value is IDictionary<string, object> map)
+                {
+                    // Any exceptions thrown here will be propagated directly. You may wish to be more
+                    // careful, if you need to control the exact exception thrown used when the
+                    // data is incomplete or has the wrong type.
+                    return new CustomPerson(
```
jskeet: I think it would be worth extracting these out to locals first. I'd probably use `TryGetValue` rather than `ContainsKey` followed by the indexer, too.

amanda-tarafa:
Agree with closing this PR, its main purpose was to show the sample (and prove it worked). It will still be referenced by the issue so anyone landing there can see this code. Will address your comments and close afterwards.

jskeet:
The PR will still exist being closed, but the code itself won't be present unless you keep your branch around... which I guess you could do for a while, and then delete it if things change.

PR #3724, opened 2019-11-11
Update Google.Cloud.Container.V1 to use the microgenerator

jskeet:


jskeet:
Added a commit addressing review comments - PTAL.

PR #3741, opened 2019-11-13
Regenerate Bigtable.V2 with GAPIC v2 config

jskeet:


jskeet:
Abandoning in favour of #3747.

PR #3750, opened 2019-11-14
Regenerate error reporting

jskeet:


jskeet:
This is badly broken. We're rolling back the proto change. Closing this PR.

PR #3754, opened 2019-11-14
Adds clustering data to the info returned about BQ tables

abair34:
While trying to use the API to return back the list of clustering columns for a table I noticed BigQueryTable.Resource.Clustering was empty.  It looks as if it was not being mapped to the returned data from the API call although the Clustering property existed in the table class.  

This commit added a unit test and populates the table's clustering info.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
@@ -606,9 +606,11 @@ namespace Google.Cloud.BigQuery.V2.Tests
                 TableReference = new TableReference { DatasetId = "dataset", TableId = "id", ProjectId = "project" },
                 TimePartitioning = new TimePartitioning { ExpirationMs = 10000, Type = "DAY" },
                 Type = "VIEW",
-                View = new TableList.TablesData.ViewData { UseLegacySql = true }
+                View = new TableList.TablesData.ViewData { UseLegacySql = true },
+                Clustering = new Clustering { Fields = new[] { "ClusterCol1", "ClusterCol2" } },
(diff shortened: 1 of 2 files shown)


abair34:
@googlebot I signed it!

jskeet:
Thanks for this. It looks like there are a few other properties exposed in the `TableList.TablesData` that would be worth copying. I'm happy either for you to amend this PR, or we (@amanda-tarafa or I) could do it all in a new PR - up to you!

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
```diff
@@ -597,8 +597,7 @@ public void CreateJob_Labels()
         [Fact]
         public void ListTablesConversion()
         {
-            var listResource = new TableList.TablesData
-            {
+            var listResource = new TableList.TablesData {
```
jskeet: Please revert this change - it was fine as it was before.

abair34: Ahh good eye!  I've got my VS formatting options set to put open {'s on the same line (probably not in alignment with the project).  I'll make sure to update them.  I'm new to contributing on github in general so I appreciate the help!  Thanks @jskeet!

jskeet: Let me know if you need a hand with the git aspects of it. Once you've created an extra commit for this bit, if you use `git rebase -i master` then change the second and third lines to have "f" at the start instead of "pick", it will squash them all together. Then just run `git push -f` to update the PR.

amanda-tarafa:
This has now been released in [Google.Cloud.BigQuery.V2 version 1.4.0-beta07](https://www.nuget.org/packages/Google.Cloud.BigQuery.V2/1.4.0-beta07).
Thanks for the PR!

Issue #3749, opened 2019-11-14
Update hand written BigQuery Options


jskeet:
The author of https://github.com/googleapis/google-api-dotnet-client/issues/1476 is unable to use `BigQueryClient.InsertRow` (etc) due to the lack of an option (`SkipInvalidRows`). There may be other options we wish to add to `InsertOptions`.

PR #3758, opened 2019-11-15
[BigQuery] Updates InsertOptions and CreateDatasetOptions

amanda-tarafa:
Partially addresses #3749.
Fully addresses googleapis/google-api-dotnet-client#1476.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
@@ -70,18 +70,87 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
             Assert.Contains(rowsAfter, r => (string)r["player"] == "Lisa");
         }
 
-        [Fact]
-        public void InsertRow_BadData()
+        public static IEnumerable<object[]> BadDataThrowsOptions
+        {
+            get
+            {
+                yield return new object[] { null };
+                yield return new object[] { new InsertOptions() };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = false } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = true } };
+                yield return new object[] { new InsertOptions { AllowUnknownFields = false } };
+                yield return new object[] { new InsertOptions { AllowUnknownFields = true } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = false, AllowUnknownFields = false } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = false, AllowUnknownFields = true } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = true, AllowUnknownFields = false } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = true, AllowUnknownFields = true } };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(BadDataThrowsOptions))]
+        public void InsertRow_BadData_Throws(InsertOptions options)
         {
             var client = BigQueryClient.Create(_fixture.ProjectId);
             var dataset = client.GetDataset(_fixture.DatasetId);
-            var table = dataset.GetTable(_fixture.HighScoreTableId);
-            var row = new BigQueryInsertRow { { "noSuchField", 10 } };
-            Assert.Throws<GoogleApiException>(() => table.InsertRow(row));
+            // Don't insert into a table used by other tests...
+            var table = dataset.CreateTable(
+                _fixture.CreateTableId(),
+                new TableSchemaBuilder { { "year", BigQueryDbType.Int64 } }.Build());
+            var rows = new BigQueryInsertRow[]
+            {
+                new BigQueryInsertRow { { "noSuchField", 10 } },
+                new BigQueryInsertRow { {"year", "Unknown"} }
+            };
+            Assert.Throws<GoogleApiException>(() => table.InsertRows(rows, options));
+        }
+
+        public static IEnumerable<object[]> BadDataSilentOptions
+        {
+            get
+            {
+                yield return new object[] { new InsertOptions { SuppressInsertErrors = true } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = false, SuppressInsertErrors = true } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = true, SuppressInsertErrors = true } };
+                yield return new object[] { new InsertOptions { AllowUnknownFields = false, SuppressInsertErrors = true } };
+                yield return new object[] { new InsertOptions { AllowUnknownFields = true, SuppressInsertErrors = true } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = false, AllowUnknownFields = false, SuppressInsertErrors = true } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = false, AllowUnknownFields = true, SuppressInsertErrors = true } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = true, AllowUnknownFields = false, SuppressInsertErrors = true } };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = true, AllowUnknownFields = true, SuppressInsertErrors = true } };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(BadDataSilentOptions))]
+        public void InsertRow_BadData_Silent(InsertOptions options)
+        {
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+            var dataset = client.GetDataset(_fixture.DatasetId);
+            // Don't insert into a table used by other tests...
+            var table = dataset.CreateTable(
+                _fixture.CreateTableId(),
+                new TableSchemaBuilder { { "year", BigQueryDbType.Int64 } }.Build());
+            var rows = new BigQueryInsertRow[]
+            {
+                new BigQueryInsertRow { { "noSuchField", 10 } },
+                new BigQueryInsertRow { {"year", "Unknown"} }
+            };
+            table.InsertRows(rows, options);
         }
 
         [Fact]
-        public void InsertRow_BadData_IgnoreBadData()
+        public void InsertRow_BadTable()
+        {
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+            var options = new InsertOptions { AllowUnknownFields = true, SkipInvalidRows = true, SuppressInsertErrors = true };
+            var row = new BigQueryInsertRow { { "noSuchField", 10 } };
+            // This should still throw because the error is not on inserting a specific row.
+            var exception = Assert.Throws<GoogleApiException>(() => client.InsertRow(_fixture.DatasetId, "noSuchTable", row));
+        }
+
+        [Fact]
+        public void InsertRow_BadData_IgnoreExtraColumn()
         {
             var client = BigQueryClient.Create(_fixture.ProjectId);
             var dataset = client.GetDataset(_fixture.DatasetId);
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/InsertOptions.cs
```diff
@@ -30,12 +30,48 @@ public sealed class InsertOptions
...
+        /// </summary>
+        public string TemplateSuffix { get; set; }
+
         internal void ModifyRequest(TableDataInsertAllRequest body)
         {
             if (AllowUnknownFields != null)
             {
                 body.IgnoreUnknownValues = AllowUnknownFields;
             }
+            if(SkipInvalidRows != null)
```
jskeet: Nit: space after "if" here and below.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/CreateDatasetOptionsTest.cs
```diff
@@ -25,12 +25,15 @@ public class CreateDatasetOptionsTest
         [Fact]
         public void ModifyRequest()
         {
+            EncryptionConfiguration config = new EncryptionConfiguration() { KmsKeyName = "dataset_default_key" };
```
jskeet: Nit: remove the `()` for the object initializer.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
```diff
@@ -70,14 +70,31 @@ public void InsertRows()
...
 
-        [Fact]
-        public void InsertRow_BadData()
+        public static IEnumerable<object[]> BadDataThrowsOptions
+        {
+            get
+            {
+                yield return new object[] { null };
+                yield return new object[] { new InsertOptions() };
+                yield return new object[] { new InsertOptions() { SkipInvalidRows = false } };
```
jskeet: Nit: remove () from the object initializers from these lines.
(At least, I tend to think they're a bit ugly. But if you actively prefer them, that's fine. :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
```diff
@@ -96,6 +113,49 @@ public void InsertRow_BadData_IgnoreBadData()
...
+                    new BigQueryInsertRow { { "name", "Jane Doe" } }
+                };
+
+            var options = new InsertOptions { SkipInvalidRows = true, ThrowOnSkippedInvalidRows = true };
+            var exception = Assert.Throws<GoogleApiException>(() => _fixture.InsertAndWait(table, () => table.InsertRows(rows, options), 1));
+            var error = exception.Error.Errors.Single();
+            Assert.Equal("noSuchField", error.Location);
+            Assert.Contains("Row 0", error.Message);
+        }
+
```
jskeet: Nit: remove extra blank line.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
```diff
@@ -96,6 +113,49 @@ public void InsertRow_BadData_IgnoreBadData()
...
+        {
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+            var dataset = client.GetDataset(_fixture.DatasetId);
+            // Don't insert into a table used by other tests...
+            var table = dataset.CreateTable(
+                _fixture.CreateTableId(),
+                new TableSchemaBuilder { { "name", BigQueryDbType.String } }.Build());
+
+            var rows = new BigQueryInsertRow[]
+                {
```
jskeet: Nit: outdent array initializer (and in the test above)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/InsertOptions.cs
```diff
@@ -30,12 +30,48 @@ public sealed class InsertOptions
...
+
+        /// <summary>
+        /// If <see cref="SkipInvalidRows"/> is set to <c>true</c>, then you can use
+        /// this option to control whether the insertion silently skips the invalid rows
+        /// or if it throws a <see cref="GoogleApiException"/> with information on each of the
+        /// rows that failed. Either way, as long as <see cref="SkipInvalidRows"/> is set to
+        /// <c>true</c> all valid rows will be inserted.
+        /// The default value is false.
+        /// </summary>
+        public bool ThrowOnSkippedInvalidRows { get; set; }
```
jskeet: Hmm. I'm not sure about this. It will make it hard to tell the difference between "Yes, really the server threw an error - we haven't inserted" and "No, everything's been inserted". It's a shame we don't return the insertion response, to be honest.

Do we know if other languages have this?

How would you feel about not doing this part yet, but starting a feature request for it and getting feedback from the BQ team?

jskeet:
(I'm mostly fine with this - just not sure about the throwing aspect.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/InsertOptions.cs
amanda-tarafa: Done

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/InsertOptions.cs
amanda-tarafa: Yep, I thought about this as well, agree that returning the response would have been better.

I did it like this in the end because I found the original feature request issue (I'll share that internally), and the main use case they were trying to solve was for:
- Skip bad rows knowing why they had been skipped
- For the ones that had been skipped because of unknown fields, attempt a new insert with `AllowUnknownFields` set to true, which is always silent.
True that would have been solved by always throwing, but I think that by default I would expect not throwing anything when `SkipInvalidRows` is set to true.

I'm fine to move this to another PR and chat with the BQ team about it. we just need to make a decision about what to do when SkipInvalidRows is set to true. I'm torn:

- I **like** skipping silently, but that doesn't cover use cases in which knowing what was skipped and why is important, so maybe the _right_ thing is to throw if skipped.

And I just looked at Java and they are returning an [`InsertAllResponse`](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-bigquery/src/main/java/com/google/cloud/bigquery/InsertAllResponse.java) which can contain errors for skipped rows.

jskeet: How about we go without the extra property, temporarily making it hard to detect this, but then change the return type in our "next major version" bump?

amanda-tarafa: Yes, very nice!!

> temporarily making it hard to detect this

So, without the extra property, should we skip rows silently (and it's hard to detect what was skipped) or throw when skipped (and it's hard to distinguish skipping erros from something else failing)? I prefer being silent but I'm fine either way.

jskeet: If possible, I think we should be silent, but it's not obvious how feasible that is. We don't want to skip *all* errors just because we've asked to skip invalid rows. Maybe the table isn't present at all! I think we probably need to think about that more. Presumably if we leave all other code (e.g. HandleInsertAllResponse) as it was, it will throw. I suspect that's okay, and what we *actually* need in the next version is an option of "Don't throw on errors; return the result to me instead."

amanda-tarafa:
@jskeet This is ready for re-review. PTAL. Thanks!

jskeet:
Right, will look tomorrow morning. It may not be as early as I'd previously thought - I'd got my weeks mixed up. But basically once I've reached an airport lounge tomorrow, I'll look :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.InsertData.cs
```diff
@@ -144,13 +144,17 @@ public override void InsertRows(TableReference tableReference, IEnumerable<BigQu
...
-            HandleInsertAllResponse(response);
+            HandleInsertAllResponse(response, options);
         }
 
-        private void HandleInsertAllResponse(TableDataInsertAllResponse response)
+        private void HandleInsertAllResponse(TableDataInsertAllResponse response, InsertOptions options)
         {
             var errors = response.InsertErrors;
-            if (errors?.Count > 0)
+            bool shouldThrow = options == null ||
```
jskeet: I would simplify this - we should throw if we haven't been told not to. If we keep the options entirely independent, it makes it easier to reason about for both us and callers. That'll be particularly important if there are other skip options later.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/InsertOptions.cs
```diff
@@ -30,12 +30,48 @@ public sealed class InsertOptions
         /// </summary>
         public bool? AllowUnknownFields { get; set; }
 
+        /// <summary>
+        /// Insert all valid rows of a request, even if invalid rows exist. The
+        /// default value is false, which causes the entire request to fail if any invalid
```
jskeet: I wouldn't say the default is false - I'd just say what the default behaviour is. (Because the default value of this property is null rather than false.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/InsertOptions.cs
```diff
@@ -30,12 +30,48 @@ public sealed class InsertOptions
...
+        /// rows exist.
+        /// Use in combination with <see cref="SuppressInsertErrors"/> to obtain
+        /// information about the rows that were skipped.
+        /// </summary>
+        public bool? SkipInvalidRows { get; set; }
+
+        /// <summary>
+        /// If <see cref="SkipInvalidRows"/> is set to <c>true</c>, then you can use
```
jskeet: Assuming you agree about simplifying the behaviour, we can simplify this description too - it should just say that insert errors don't lead to an exception. You could always put that more detailed information will be available in a future release if you want.

(I'm thinking that in the future, the only change will be to return the InsertAllRowsResponse - we should still throw by default, even when SkipInvalidRows is true, to avoid it being a tricky breaking change. And users may want to set this even if they're not skipping rows, to perform their own handling.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.InsertData.cs
amanda-tarafa: Yes, I thought about making them independent as well. The reason I didn't was because then there might be cases in which not even good rows are inserted but nothing is thrown either, albeit at the request of the caller, and allowing that case seemed somewhat odd.
But agree that making them independent will simplify things a lot. I'll make a point of explaining the Supress... property better.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/InsertOptions.cs
amanda-tarafa: Done.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/InsertOptions.cs
amanda-tarafa: Done updating the description.

And really agree with the idea of keeping this for the future as you described.

PR #3759, opened 2019-11-15
Restore the ability to pass a cancellation token in for query streaming

jskeet:
Users could already cancel the stream using a cancellation token
when calling GetAsyncEnumerator, but it may be convenient to do it
when creating the stream.


jskeet:
Cracking - merging, will then rebase to keep it with the other one.

PR #3766, opened 2019-11-18
Migrating Trace.V2 to GAPIC v2 and microgenerator

jskeet:
(We have to do this in one step as there's no equivalent of "use common resource names for monolithic generator with gapic v2 config".)


jskeet:
(Ready for review now - sorry for the brokenness before; had forgotten to git add the new files.)

PR #3767, opened 2019-11-18
Move Error Reporting to the microgenerator

jskeet:
Expected breaking changes:

- GroupName renamed to ErrorGroupName
- Missing retry methods/properties

This is okay as this is only a beta API/package.


jskeet:
Right, ready for review :)

apis/Google.Cloud.ErrorReporting.V1Beta1/Google.Cloud.ErrorReporting.V1Beta1/ReportErrorsServiceClient.g.cs
```diff
@@ -253,7 +253,7 @@ public static ReportErrorsServiceClient Create(grpccore::CallInvoker callInvoker
         public virtual ReportErrorsService.ReportErrorsServiceClient GrpcClient => throw new sys::NotImplementedException();
 
         /// <summary>
-        /// Report an individual error event.
+        /// Report an individual error @event.
```
chrisdunelm: Don't want `@` here, but it's probably OK for now. I'll fix the generator soon, then this can be re-gened.

jskeet: Yup, that's definitely simpler than trying to fix that. Well spotted though :)

PR #3776, opened 2019-11-21
Update Tasks.V2 to use the microgenerator

jskeet:
This involves quite a few steps in order to maintain backward compatibility.

For Tasks.V2Beta3, I'm going to allow all the breaking changes to be breaking changes.


jskeet:
Commits 1, 3, 4 and 5 were "at least partially manual" so worth reviewing. Commit 2 (the big one) was purely generated.

jskeet:
I don't understand the details of why it's not generated now, but I'm not bothered :)

PR #3777, opened 2019-11-21
Move Tasks.V2Beta3 to the microgenerator

jskeet:
Breaking changes are expected and acceptable.


jskeet:
Breaking in all the expected ways:

- Retry methods/properties removed
- IAM method parameters now just IResourceName
- LocationName is now Google.Api.Gax.ResourceNames.LocationName

Other than LocationName, I don't expect this to really break users anyway, and LocationName is just a matter of using a different import.

PR #3781, opened 2019-11-25
Rework EnumConverter.

mikelehen:
This avoids generics (and is perhaps a bit simpler).  The key difference is that I use `Enum.ToObject()` to convert the numeric value to the TargetType enum instead of having to use generics and a cast.  But I also ended up reworking the whole class to get rid of the `PrimitiveConverter` classes since it seemed simpler to just use a couple `switch` / `case` statements now.

Context: I had to rework the fork of this code that we're using in the Unity SDK due to Unity / iOS limitations.

I don't actually care too much if you want to accept this PR or not but I wanted to at least see if you have any thoughts / feedback on the approach.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/EnumConverter.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/EnumConverter.cs
@@ -26,11 +26,11 @@ namespace Google.Cloud.Firestore.Converters
     /// </summary>
     internal sealed class EnumConverter : ConverterBase
     {
-        private TypeCode typeCode;
+        private readonly TypeCode _typeCode;
 
         internal EnumConverter(BclType targetType) : base(targetType)
         {
-            typeCode = BclType.GetTypeCode(TargetType);
+            _typeCode = BclType.GetTypeCode(TargetType);
         }
 
         protected override object DeserializeInteger(DeserializationContext context, long value)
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/EnumConverter.cs
```diff
@@ -26,94 +26,65 @@ namespace Google.Cloud.Firestore.Converters
     /// </summary>
     internal sealed class EnumConverter : ConverterBase
     {
-        private static readonly ConcurrentDictionary<BclType, PrimitiveConverter> s_primitiveConverters =
-            new ConcurrentDictionary<BclType, PrimitiveConverter>();
-
-        private readonly PrimitiveConverter _converter;
+        private TypeCode typeCode;
```
jskeet: Could you rename this to `_typeCode` and make it readonly please?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/EnumConverter.cs
```diff
@@ -26,94 +26,65 @@ namespace Google.Cloud.Firestore.Converters
...
+        }
 
-            private sealed class Int64EnumConverter<T> : PrimitiveConverter
-            {
-                internal override long EnumToInt64(object value) => (long) value;
-                internal override object Int64ToEnum(long value) => (T) (object) value;
+        public override Value Serialize(SerializationContext context, object value) =>
+            new Value { IntegerValue = EnumToInt64(value) };
+
+        private long EnumToInt64(object value) {
```
jskeet: Please fix the formatting to "brace at start of line" here and in the next line.
For the cases, I'm fine with the existing formatting, or the return statement could be on the same line as the case.

jskeet: (And all of this for Int64ToEnumBaseType too :)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/EnumConverter.cs
mikelehen: Woops, good catch.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Converters/EnumConverter.cs
mikelehen: Done.  The return statements on same line look nice.  Thanks.

PR #3784, opened 2019-11-26
Migrate Google.Cloud.Datastore.V1 to micro-generator

jskeet:


apis/apis.json
```diff
@@ -166,9 +166,9 @@
 
   {
     "id": "Google.Cloud.Datastore.V1",
-    "generator": "gapic",
+    "generator": "micro",
     "protoPath": "google/datastore/v1",
-    "serviceYaml": "datastore.yaml",
+    "serviceYaml": "datastore_v1.yaml",
```
chrisdunelm: Why is the `serviceYaml` still needed?

jskeet: Whoops, it's not. Will remove.

PR #3786, opened 2019-11-27
Migrate Google.Cloud.Vision.V1 to the microgenerator

jskeet:


jskeet:
(This doesn't have a GAPIC v2 config yet, but it seems to be fine.)

PR #3788, opened 2019-11-28
Remove Google.Cloud.Irm.V1Alpha2

jskeet:
This has never been released as a NuGet package; we'll wait for the beta API.

Note: I expect CI to fail because it will detect Google.Cloud.Irm.V1Alpha2 as a "change" and try to build it.


jskeet:
(CI has failed in exactly the way predicted, with no other problems.)

Issue #3783, opened 2019-11-25
[Firestore] Support for "in" queries


lluismontero:
Earlier this month Firestore announced support for **"in"** queries (https://firebase.googleblog.com/2019/11/cloud-firestore-now-supports-in-queries.html). 

In many of the client libraries the option is already available, but not in the .NET SDK (https://cloud.google.com/firestore/docs/query-data/queries#in_and_array-contains-any). Is there any ETA when the "in" operator will be available?

jskeet:
Hi @lluismontero, thanks for the heads-up. I haven't talked with the Firestore team about this, but I'd imagine we'd want to support it as soon as we can. Just to set expectations, it definitely won't be before Q1 2020... but I'll see if we can do it then :)

lluismontero:
Hi @jskeet, thanks for your response. Is there any workarround to run that kind of queries?

jskeet:
@lluismontero: Not without digging lower down than I suspect you want to go. (I mean, we do provide the underlying `V1.FirestoreClient` so you could make the query call yourself, but it'll be pretty painful, I'm afraid. It's a big cliff between the abstraction layers.)

lluismontero:
Thanks anyway, I'll keep an eye on the SDK updates

jskeet:
We've now released Google.Cloud.Firestore version 1.1.0 which contains this functionality. So I was wrong on the "definitely not before Q1 2020"...

Issue #3397, opened 2019-08-28
[Firestore] ListDocumentsAsync against emulator resulting in PermissionDenied


emmekappa:
I'm running Firestore emulator on MacOS 10.14.6 from gcloud sdk version 260.0.0.

And I'm using the following code to connect to the emulator:

`return FirestoreDb.Create("na",
                FirestoreClient.Create(new Channel("127.0.0.1", 8080, ChannelCredentials.Insecure)));`

Every `ListDocumentsAsync()` operations results in 

`Grpc.Core.RpcException : Status(StatusCode=PermissionDenied, Detail="Metadata operations require admin authentication.")`

Looks the same problem described in https://github.com/firebase/firebase-tools/issues/1363

emmekappa:
Found a way to workaround the issue by expliciting the `FirestoreSettings` adding `Baerer owner` as `Authorization` header.

Here's now the code to create a `FirestoreDb` able to list documents:

```
var firestoreSettings = new FirestoreSettings()
{
   CallSettings = CallSettings.FromHeaderMutation(x => x.Add("Authorization", "Bearer owner"))
};
return FirestoreDb.Create("na",
   FirestoreClient.Create(new Channel("127.0.0.1", 8080, ChannelCredentials.Insecure), firestoreSettings));
```

jskeet:
That's interesting, thanks. I'll treat this issue as a feature request to make that easier to do. We might even end up with a simple way of configuring `FirestoreDbBuilder` to use emulator in the same way that we do for DatastoreDb.

Thanks for filing and then following up!

emmekappa:
Yes @jskeet, it would be great to have the same OOB emulator support already implemented in DatastoreDb.

jskeet:
This is now implemented in Google.Cloud.Firestore version 1.1.0-beta02.
See http://googleapis.github.io/google-cloud-dotnet/docs/Google.Cloud.Firestore/ for an example.

PR #3792, opened 2019-12-03
Release Google.Cloud.Firestore 1.1.0

jskeet:
Changes since 1.0.0
---

- Support for In and ArrayContainsAny queries (#3783)
- Firestore emulator support (#3397)
- Conversion support for named value tuples (#2787)
- FirestoreDbBuilder for simplified configuration beyond defaults
- Per-FirestoreDb converter customization (#3255)
- Public FirestoreEnumNameConverter type (#2842)
- Document snapshot timestamp propagation attributes (#2830)

Changes since 1.1.0-beta02
---

- Support for In and ArrayContainsAny queries


jskeet:
(I've gone straight to 1.1.0 to avoid another beta followed by an almost-immediate GA. I'm pretty confident in the API here - and we'll be starting 2.0 releases in January anyway.)

PR #3805, opened 2019-12-06
Make builds quieter

jskeet:
We really don't need to see the compiler version hundreds of times


jskeet:
Hold up - I think I've got a way of making it quieter still :) Just testing.

jskeet:
(This is now ready for review.)

PR #3813, opened 2019-12-09
Release Google.Cloud.Container.V1 version 1.1.0

jskeet:
Changes since 1.0.0:

- Some retry settings are now obsolete, and will be removed in the next major version
- Added support for many cluster options
- Added ListUsableSubnetworks methods

### DIFF
--- a/apis/Google.Cloud.Container.V1/docs/history.md
+++ b/apis/Google.Cloud.Container.V1/docs/history.md
@@ -1,0 +1,11 @@
+# Version history
+
+# Version 1.1.0, released 2019-12-09
+
+- Some retry settings are now obsolete, and will be removed in the next major version
+- Added support for many cluster options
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Container.V1/docs/history.md
```diff
@@ -0,0 +1,11 @@
+# Version history
+
+# Version 1.1.0, released 2019-12-09
+
+- Some retry settings are now obsolete, and will be removed in the next major version
```
amanda-tarafa: These don't have link to the commits, just flagging in case it is relevant.

jskeet: Nope, that's fine. (I'll keep the commit links in some cases, but I don't think they always need to be there.)

PR #3815, opened 2019-12-09
Release Google.Cloud.DevTools.Common version 1.1.0

jskeet:
Aside from updated comments, this is just a dependency update.

### DIFF
--- a/apis/Google.Cloud.DevTools.Common/docs/history.md
+++ b/apis/Google.Cloud.DevTools.Common/docs/history.md
@@ -1,0 +1,10 @@
+# Version history
+
+# Version 1.1.0, released 2019-12-09
+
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.DevTools.Common/docs/history.md
```diff
@@ -0,0 +1,10 @@
+# Version history
+
+# Version 1.1.0, released 2019-12-09
+
+- [Commit 518728a](https://github.com/googleapis/google-cloud-dotnet/commit/518728a): A few minor tidy-ups of comments
```
amanda-tarafa: Is this something that we'd want in version history? I would say no.

jskeet: I'd normally agree, but it shows that at least *something* has happened other than the dependency versions being updated :)

Issue #2957, opened 2019-03-19
Google.Cloud.Translation.V2 - Translate Html


OpenSpacesAndPlaces:
#### Environment details

  - OS: Windows
  - .NET version: 4.5.2
  - Package name and version:
Google.Cloud.Translation.V2

#### Steps to reproduce
Large Html snippet to translate

Code
GoogleCredential gc = GoogleCredential.FromJson(System.IO.File.ReadAllText(sDrivePath));
TranslationClient tcClient = TranslationClient.Create(gc, TranslationModel.ServiceDefault);
TranslationResult result = tcClient.TranslateHtml(sHtml, GetLangCodeCookie(), LanguageCodes.English,TranslationModel.ServiceDefault);

Error

```
[UriFormatException: Invalid URI: The Uri string is too long.]
   System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind) +6912000
   System.Uri..ctor(String uriString) +21
   Google.Apis.Requests.RequestBuilder.UriJoin(Uri baseUri, String path) +192
   Google.Apis.Requests.RequestBuilder.BuildUri() +279
   Google.Apis.Requests.RequestBuilder.CreateRequest() +49
   Google.Apis.Requests.ClientServiceRequest`1.CreateRequest(Nullable`1 overrideGZipEnabled) +26
   Google.Apis.Requests.<ExecuteUnparsedAsync>d__33.MoveNext() +69
   System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw() +26
   Google.Apis.Requests.ClientServiceRequest`1.Execute() +149
   Google.Cloud.Translation.V2.TranslationClientImpl.TranslateHtml(IEnumerable`1 htmlItems, String targetLanguage, String sourceLanguage, Nullable`1 model) +150
```

Related reading would suggest the library code does not switch from GET to POST for large requests:
https://stackoverflow.com/questions/9432422/google-translate-v2-cannot-hanlde-large-text-translations-from-c-sharp

jskeet:
Hmm. I thought it did, but I'll investigate. Thanks for reporting.

OpenSpacesAndPlaces:
If there is an override or alternate methods I missed - that works too.
Though I wasn't seeing anything on a cursory run through.

jskeet:
Just so I can reproduce this, how large is the file? (We have an integration test to check a 7K file - I'll try making that 10x larger just as a first pass to reproduce.)

[...]

PR #3818, opened 2019-12-09
Release Google.Cloud.Translation.V2 version 1.2.0

jskeet:
Changes since 1.1.0:

- [Commit 5c5afff](https://github.com/googleapis/google-cloud-dotnet/commit/5c5afff): Added client builders for simplified configuration
- [Commit 4c1905b](https://github.com/googleapis/google-cloud-dotnet/commit/4c1905b): Use Translate method instead of List method, to avoid URI construction limits. Fixes [issue 2957](https://github.com/googleapis/google-cloud-dotnet/issues/2957) as far as we can.


jskeet:
Eek - the history log seems to be missing. Will find it...

jskeet:
Right, I've added the version history back now. PTAL :)

PR #3841, opened 2019-12-10
Release Google.Cloud.DevTools.ContainerAnalysis.V1 and Grafeas.V1 version 1.0.0

jskeet:
This is the first GA release of these two packages.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -226,13 +226,15 @@
     "serviceYaml": "containeranalysis_v1.yaml",
     "productName": "Google Container Analysis",
     "productUrl": "https://cloud.google.com/container-registry/docs/container-analysis/",
-    "version": "1.0.0-beta05",
+    "version": "1.0.0",
     "type": "grpc",
     "description": "Recommended Google client library to access the Google Cloud Container Analysis API.",
     "tags": [ "analysis", "artifact", "metadata", "grafeas" ],
     "dependencies": {
-      "Google.Cloud.Iam.V1": "1.2.0",
-      "Grafeas.V1": "project"
+      "Google.Api.Gax.Grpc": "2.10.0",
+      "Google.Cloud.Iam.V1": "1.3.0",
+      "Grafeas.V1": "project",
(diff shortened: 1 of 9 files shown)


apis/apis.json
```diff
@@ -226,13 +226,15 @@
...
+    "version": "1.0.0",
     "type": "grpc",
     "description": "Recommended Google client library to access the Google Cloud Container Analysis API.",
     "tags": [ "analysis", "artifact", "metadata", "grafeas" ],
     "dependencies": {
-      "Google.Cloud.Iam.V1": "1.2.0",
-      "Grafeas.V1": "project"
+      "Google.Api.Gax.Grpc": "2.10.0",
+      "Google.Cloud.Iam.V1": "1.3.0",
+      "Grafeas.V1": "project",
```
amanda-tarafa: Just flagging in case it's a problem, but this is not a explicti version dependency, but to the project, and the release is not a beta.

docs/root/index.md
```diff
@@ -63,14 +64,13 @@ GA:
   - Additionally, the following libraries are available for access to beta API functionality:
     - [Google.Cloud.Vision.V1P1Beta1](Google.Cloud.Vision.V1P1Beta1/index.html)
     - [Google.Cloud.Vision.V1P2Beta1](Google.Cloud.Vision.V1P2Beta1/index.html)
+- [Grafeas.V1](Grafeas.V1/index.html)
```
amanda-tarafa: The indent seems to be wrong.

apis/apis.json
jskeet: That's okay because it's being released at the same time. (It would fail otherwise.)

docs/root/index.md
jskeet: Nope, it's just that DLP just above has a two-level-nested list. It's the same as other top-level packages.

jskeet:
Whoops - looks like I may not have run generateprojects at all... odd!

PR #3849, opened 2019-12-10
Release Google.Cloud.Bigtable.Common.V2 version 1.1.0

jskeet:
Changes since 1.0.0:

- Resource name types have Format methods


jskeet:
(Doing Common first will make it easier to do the other ones separately, I think.)

PR #3864, opened 2019-12-11
Add release notes for remaining APIs

jskeet:
Many of these are just links to other packages. For the Spanner admin APIs we *may* want to add more detailed histories at some point, but it's better to have something than nothing.

### DIFF
--- a/apis/Google.Cloud.Firestore/docs/history.md
+++ b/apis/Google.Cloud.Firestore/docs/history.md
@@ -1,0 +1,15 @@
+# Version history
+
+# Version 1.1.0, released 2019-12-03
+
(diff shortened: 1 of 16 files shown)


apis/Google.Cloud.Firestore/docs/history.md
```diff
@@ -0,0 +1,15 @@
+# Version history
+
+# Version 1.1.0, released 2019-12-03
```
amanda-tarafa: These are missing de links. Maybe add the links or delete the references to the issues/PRs

jskeet: Will add the links. These ones were copied from a commit.

PR #3904, opened 2019-12-16
Migrate Google.LongRunning to the microgenerator (next-major-version only)

jskeet:
This currently involves patching the proto - I'm shepherding the same change upstream, but there are some internal failures (which don't affect us) to investigate.


jskeet:
Will look into the conflict

jskeet:
(Fixed conflict; ready for review.)

PR #3908, opened 2019-12-16
Migrate Google.Cloud.SecurityCenter.V1 to the microgenerator

jskeet:


jskeet:
If you haven't already started looking at this, don't do so for now - I want to change one of the commits :)

jskeet:
Okay, it's now as I want it to be - PTAL.

PR #3914, opened 2019-12-16
Release Google.Cloud.Speech.V1P1Beta1 version 1.0.0-beta05

jskeet:
This contains no API surface changes, just a fix to not attempt to retry streaming calls (which isn't supported).


jskeet:
(Self-reviewing - will merge on green.)

PR #3917, opened 2019-12-17
Update release process documentation.

jskeet:


PROCESSES.md
```diff
@@ -129,32 +129,114 @@ project itself:
...
 references from APIs being released now to APIs that *aren't* being
 released. Without this check, it's possible for a released version
 to depend on unreleased changes.
+
+**Building and publishing the release**
+
+On a Google corp machine, trigger a Kokoro release build. (This
+can only be performed by Googlers.)
+
+If you want to release anything other than the current head of
```
amanda-tarafa: Aint this what we usually want? The way is written it seems like we usually release master's current head. I tend to always specify the commitish (tagreleases.sh always "returns" the commitish) just in case someone has merged something in between. I think it's safer as general instructions to say "always specify the commitish".

jskeet: I almost always trigger it immediately after running tagreleases.sh, and just let it go with HEAD. It's never caused a problem :) I think it's easier to let that be the default approach - especially as if there are multiple people within in the repo, they're likely to be us!

PR #3920, opened 2019-12-17
Pubsub: Update default settings; add maximum total lease extension

chrisdunelm:


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -207,10 +218,15 @@ internal void Validate()
         public static TimeSpan MinimumAckExtensionWindow { get; } = TimeSpan.FromMilliseconds(50);
 
         /// <summary>
-        /// The default message ACKnowlegdment extension window of 15 seconds.
+        /// The default message ACKnowlegdement extension window of 15 seconds.
```
jskeet: There's still a typo here - gd instead of dg

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1099,34 +1119,105 @@ private async Task ProcessPullMessagesAsync(List<ReceivedMessage> msgs, HashSet<
...
                     {
-                        lock (_lock)
+                        cancellation.Dispose();
+                        lock (msgIds)
                         {
-                            _extendQueue.Enqueue(msgIds.Select(x => new TimedId(_extendThrottleHigh + 1, x)));
+                            msgIds.Clear();
                         }
-                    }
+                    }));
```
jskeet: Should the method return at this point? Presumably it's just about feasible for _maxExtensionDuration to be one tick, so it could be disposed by the end of the block.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1099,34 +1119,105 @@ private async Task ProcessPullMessagesAsync(List<ReceivedMessage> msgs, HashSet<
...
+                }
+
+                public void Cancel()
+                {
+                    CancellationTokenSource cts2;
+                    lock (_lock)
+                    {
+                        cts2 = _cts;
+                    }
+                    cts2?.Cancel();
```
jskeet: Should this dispose as well? I'm a little lost as to what this is all doing.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1099,34 +1119,105 @@ private async Task ProcessPullMessagesAsync(List<ReceivedMessage> msgs, HashSet<
...
+                            {
+                                _extendQueue.Enqueue(msgIds.Select(x => new TimedId(_extendThrottleHigh + 1, x)));
+                            }
+                        }
+                    }
+                    if (anyMsgIds)
+                    {
+                        // Ids have been added to _extendQueue, so trigger a push.
+                        _eventPush.Set();
+                        // Some ids still exist, schedule another extension.
```
jskeet: Possibly edit this comment to add "with the same deadline for how long the lease can be extended"?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
jskeet: (Comments on this and Dispose would be useful.)

jskeet: Reading through everything else, it looks like it doesn't *have* to be disposed, but I suspect it would be good to do so.

PR #4025, opened 2020-01-03
API regeneration

jskeet:
(Note that this is on the next-major-version branch.)

First commit is KMS only; second commit is "regenerate everything".

The only manual aspects are the deleted files in the first commit.


apis/Google.Cloud.OsLogin.V1/Google.Cloud.OsLogin.V1/OsLoginServiceClient.g.cs
```diff
@@ -656,10 +656,10 @@ public static OsLoginServiceClient Create(grpccore::CallInvoker callInvoker, OsL
         /// </param>
         /// <param name="callSettings">If not null, applies overrides to this RPC call.</param>
         /// <returns>The RPC response.</returns>
-        public virtual LoginProfile GetLoginProfile(gcoc::UserName name, gaxgrpc::CallSettings callSettings = null) =>
+        public virtual LoginProfile GetLoginProfile(gcoc::PosixAccountName name, gaxgrpc::CallSettings callSettings = null) =>
```
chrisdunelm: Are these (seemingly breaking) changes expected?

jskeet: Good question. PosixAccountName is correct as far as I'm aware - I don't know why it would previously have been generated as UserName. Will look into it and report back.

jskeet: Ooh - it looks like it's actually a bug in the generator. Here's the definition: https://github.com/googleapis/googleapis/blob/4d45a6399e9444fbddaeb1c86aabfde210723714/google/cloud/oslogin/v1/oslogin.proto#L142

Note that PosixAccountName is the *child* type - not the resource type. So it was right before.

chrisdunelm: Yes, it was a generator bug :( It's now fixed :)

jskeet:
I'm closing this PR as we have a bunch of other APIs to migrate to the microgenerator; I'll redo the whole "regenerate everything" after that.

PR #4027, opened 2020-01-03
Migrate Google.Cloud.Firestore.Admin.V1 to the microgenerator

jskeet:


apis/Google.Cloud.Firestore.Admin.V1/Google.Cloud.Firestore.Admin.V1.SmokeTests/FirestoreAdminSmokeTest.cs
```diff
@@ -33,10 +33,10 @@ public static int Main(string[] args)
             FirestoreAdminClient client = FirestoreAdminClient.Create();
 
             // Initialize request argument
-            ParentName parentName = new ParentName(projectId, "(default)", "collection");
+            CollectionGroupName collectionGroupName = CollectionGroupName.FromProjectDatabaseCollection(projectId, "(default)", "collection");
```
chrisdunelm: This change looks like it's the result of a breaking change in the generated code?

jskeet: Yes. That's okay, given that we won't be doing this in the master branch.
The "parent name" name was specified in the GAPIC config before: https://github.com/googleapis/googleapis/blob/91ef2d9dd69807b0b79555f22566fb2d81e49ff9/google/firestore/admin/v1/firestore_gapic.yaml#L24

PR #4037, opened 2020-01-06
Update Firestore manual code to reflect GAX changes

jskeet:
(The QueryTest change seems to be just reverting a previous change, but that wouldn't have compiled. Not sure what happened there, but this is correct.)

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/RetryHelper.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/RetryHelper.cs
@@ -31,35 +31,41 @@ namespace Google.Cloud.Firestore
             Func<TRequest, CallSettings, Task<TResponse>> fn,
             TRequest request, CallSettings callSettings, IClock clock, IScheduler scheduler)
         {
-            RetrySettings retrySettings = callSettings.Timing?.Retry;
+            RetrySettings retrySettings = callSettings.Retry;
             if (retrySettings == null)
             {
                 return await fn(request, callSettings).ConfigureAwait(false);
             }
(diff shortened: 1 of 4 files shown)


jskeet:
Integration tests are failing - looking into that now.

jskeet:
Think I've fixed it - will update the PR.

jskeet:
Done, ready for review again :)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/RetryHelper.cs
```diff
@@ -31,35 +31,41 @@ internal static class RetryHelper
...
-            DateTime? overallDeadline = retrySettings.TotalExpiration.CalculateDeadline(clock);
-            TimeSpan retryDelay = retrySettings.RetryBackoff.Delay;
-            TimeSpan callTimeout = retrySettings.TimeoutBackoff.Delay;
+            DateTime? overallDeadline = callSettings.Expiration.CalculateDeadline(clock);
+            // Every attempt should use the same deadline, calculated from the start of the call.
+            if (callSettings.Expiration?.Type == ExpirationType.Timeout)
+            {
+                callSettings = callSettings.WithDeadline(overallDeadline.Value);
+            }
+            // Remove retry from the call settings we pass into the function, so that the settings
```
amanda-tarafa: This is almost certainly that I lack context, but I don't understand this comment.
What I get is:

- Streaming calls are not retried (which makes sense).
- But then, how do we get here for a streaming call with a not null callSettings.Retry? And why?

jskeet: We get here from FirestoreDb, which passes in call settings that *do* have a retry, precisely because this code has to react to the retry bit.

PR #4068, opened 2020-01-07
Migrate Asset.V1 to the microgenerator

jskeet:


apis/Google.Cloud.Asset.V1/Google.Cloud.Asset.V1.SmokeTests/AssetServiceSmokeTest.cs
```diff
@@ -48,7 +40,7 @@ public static int Main(string[] args)
             // Initialize request argument(s)
             var request = new BatchGetAssetsHistoryRequest
             {
-                ParentAsProjectName = new ProjectName(projectId),
+                ParentAsResourceName = new ProjectName(projectId),
```
chrisdunelm: Do you know why this property name has changed? I would usually expect the naming to remain unchanged?

jskeet: The type has changed, from `ProjectName` to `IResourceName`, so the name change makes sense.

PR #4070, opened 2020-01-07
Add local GAX to the repo.

amanda-tarafa:
So that we can share it and don't have to be creating the folder for every branch.


jskeet:
If you run build.sh, that fetches and builds GAX already - it means we don't need to keep refreshing it. I only didn't do that in the generator because the dependency on LongRunning ended up being tricksy.
I *think* I'd prefer to keep things the way they are, but I'm happy to be persuaded.

amanda-tarafa:
Ah no, that's great. I'll run build.sh then. And closing now.

PR #4073, opened 2020-01-08
[BigQuery] Support for specifying specific fields when fetching a table or listing rows.

amanda-tarafa:
Partially addresses #3749 

First commit adds support for partial table schema retrieval.
Second commit adds support for partial row (data) retrieval. Semantic breaking change here, so pushing into next major version branch.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/TablesTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/TablesTest.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using Google.Apis.Bigquery.v2.Data;
+using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
@@ -308,14 +309,26 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
             Assert.Equal(totalCount - 2, remainder.Count);
         }
 
-        [Fact]
-        public void GetTable()
+        public static IEnumerable<object[]> FullSchemaOptions
+        {
+            get
+            {
+                yield return new object[] { null };
+                // Null or empty SelectedFields return the whole schema.
+                yield return new object[] { new GetTableOptions { SelectedFields = null } };
+                yield return new object[] { new GetTableOptions { SelectedFields = string.Empty } };
+            }
+        }
+
+        [Theory]
+        [MemberData(nameof(FullSchemaOptions))]
+        public void GetTable_FullSchema(GetTableOptions options)
         {
             var client = BigQueryClient.Create(_fixture.ProjectId);
             string datasetId = _fixture.DatasetId;
             string tableId = _fixture.HighScoreExtendedTableId;
 
-            var table = client.GetTable(datasetId, tableId);
+            var table = client.GetTable(datasetId, tableId, options);
 
             Assert.Equal(4, table.Schema.Fields.Count);
             Assert.Equal("player", table.Schema.Fields[0].Name);
@@ -374,6 +387,8 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
             };
             var table = client.GetTable(datasetId, tableId, options);
 
+            // We still have 4 top level fields because nesting is preserved
+            // when requesting a partial table schema.
             Assert.Equal(4, table.Schema.Fields.Count);
 
             Assert.Equal("age", table.Schema.Fields[0].Name);
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/TablesTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/TablesTest.cs
@@ -308,5 +308,86 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
             Assert.Equal(totalCount - 2, remainder.Count);
         }
 
+        [Fact]
+        public void GetTable()
+        {
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+            string datasetId = _fixture.DatasetId;
+            string tableId = _fixture.HighScoreExtendedTableId;
+
+            var table = client.GetTable(datasetId, tableId);
+
+            Assert.Equal(4, table.Schema.Fields.Count);
+            Assert.Equal("player", table.Schema.Fields[0].Name);
+            Assert.Equal("gameStarted", table.Schema.Fields[1].Name);
+            Assert.Equal("score", table.Schema.Fields[2].Name);
+            Assert.Equal("gameFinished", table.Schema.Fields[3].Name);
+        }
+
+        [Fact]
+        public void GetTable_PartialSchema_SimpleTypes()
+        {
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+            string datasetId = _fixture.DatasetId;
+            string tableId = _fixture.HighScoreExtendedTableId;
+            var options = new GetTableOptions
+            {
+                SelectedFields = "player,score"
+            };
+            var table = client.GetTable(datasetId, tableId, options);
+
+            Assert.Equal(2, table.Schema.Fields.Count);
+            Assert.Equal("player", table.Schema.Fields[0].Name);
+            Assert.Equal("score", table.Schema.Fields[1].Name);
+        }
+
+        [Fact]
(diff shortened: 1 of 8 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetTableOptions.cs
```diff
@@ -22,8 +22,19 @@ namespace Google.Cloud.BigQuery.V2
     /// </summary>
     public sealed class GetTableOptions
     {
+        /// <summary>
+        /// Comma separated list of schema fields to return.
+        /// When set allows to obtain a partial table schema.
+        /// If not set, all fields, that is, the full table schema, will be returned.
```
jskeet: I'd probably simplify this to "When this property is null, the full table schema is returned."

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/TablesTest.cs
```diff
@@ -308,5 +308,86 @@ public void ListTables_ResumeWithPageToken()
...
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+            string datasetId = _fixture.DatasetId;
+            string tableId = _fixture.PeopleTableId;
+            var options = new GetTableOptions
+            {
+                SelectedFields = "age,gender,phoneNumber,children.gender,children.age"
+            };
+            var table = client.GetTable(datasetId, tableId, options);
+
+            Assert.Equal(4, table.Schema.Fields.Count);
```
jskeet: Possibly add a comment to highlight that we've got five entries above, but only four fields, due to nesting?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.Queries.cs
```diff
@@ -180,10 +184,14 @@ private JobsResource.InsertRequest CreateInsertQueryJobRequest(string sql, IEnum
             return CreateInsertJobRequest(new JobConfiguration { Query = jobConfigurationQuery, DryRun = options?.DryRun }, options);
         }
 
-        private TabledataResource.ListRequest CreateListRequest(TableReference tableReference, ListRowsOptions options)
+        private TabledataResource.ListRequest CreateListRequest(TableReference tableReference, ListRowsOptions options, TableSchema schema)
         {
             var request = Service.Tabledata.List(tableReference.ProjectId, tableReference.DatasetId, tableReference.TableId);
             options?.ModifyRequest(request);
+            if (schema != null)
```
jskeet: Possibly just use `request.SelectedFields = schema?.BuildSelectedFields();`? It's fine (I believe) to set `request.SelectedFields` to null redundantly.

jskeet: Or call `BuildSelectedFields` unconditionally, given that it looks like it will handle null input.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/TableSchemaExtensions.cs
```diff
@@ -40,5 +40,38 @@ internal static class TableSchemaExtensions
             }
             return ret;
         }
+
+        internal static string BuildSelectedFields(this TableSchema schema)
+        {
+            if (!(schema?.Fields?.Count > 0))
```
jskeet: This is where the null handling matters. I would suggest that if `schema` is null, we should return a null string to indicate "all fields". It's only if we've got a non-null schema with null/empty fields that we want to say "don't return any fields" I think - but we should add a clear comment about that, too :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/TableSchemaExtensions.cs
```diff
@@ -40,5 +40,38 @@ internal static class TableSchemaExtensions
...
+            var built = new List<string>();
+            var currentField = new List<string>();
+            BuildSelectedFields(schema.Fields, currentField, built);
+            return string.Join(",", built);
+        }
+
+        private static void BuildSelectedFields(IList<TableFieldSchema> fieldSchemas, IList<string> currentField, IList<string> built)
+        {
+            // Whatever we have in currentField is terminal.
+            // Build that an we are done.
```
jskeet: Nit: an => and

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/TableSchemaExtensions.cs
```diff
@@ -40,5 +40,38 @@ internal static class TableSchemaExtensions
...
+
+            var built = new List<string>();
+            var currentField = new List<string>();
+            BuildSelectedFields(schema.Fields, currentField, built);
+            return string.Join(",", built);
+        }
+
+        private static void BuildSelectedFields(IList<TableFieldSchema> fieldSchemas, IList<string> currentField, IList<string> built)
+        {
+            // Whatever we have in currentField is terminal.
```
jskeet: Possibly move this comment into the body of the if statement?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/TableSchemaExtensions.cs
```diff
@@ -40,5 +40,38 @@ internal static class TableSchemaExtensions
...
+            // Build that an we are done.
+            if (!(fieldSchemas?.Count > 0))
+            {
+                built.Add(string.Join(".", currentField));
+            }
+            else
+            {
+                int removeAtIndex = currentField.Count;
```
jskeet: Is there any reason not to just remove the last element and *then* do the adding? (The idea is to replace the complex field with all the field paths, right?)
Either way, a comment explaining the logic would help.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/TableSchemaExtensions.cs
```diff
@@ -40,5 +40,38 @@ internal static class TableSchemaExtensions
...
+                foreach(var field in fieldSchemas)
```
jskeet: Nit: space after foreach.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/TableSchemaExtensions.cs
```diff
@@ -43,33 +43,52 @@ internal static class TableSchemaExtensions
...
             }
             else
             {
-                int removeAtIndex = currentField.Count;
-                foreach(var field in fieldSchemas)
+                // For each field at this level, we'll add it to the current path
+                // and then go in depth into its possible nested fields.
+                // When we return from the recursive call, we can remove this field
+                // from the current path. It will always be at the last index, which
+                // will always be one more than what we have right now.
```
jskeet: It's still not clear to me why we can't just remove the last entry *before* adding the child entries. (That would be very slightly more efficient, but that's not particularly important.)

amanda-tarafa: Because the one I'm removing is the parent of those child entries, it is part of the path to get to each of the terminal child ones. I add it inside the loop, before doing the recursive call and remove it right after returning. I might be missing something though?

jskeet: Ah, okay, I think I see now. It might be a bit clearer with method documentation for what the various parameters mean. Maybe we should merge as is, then I should spend a bit more time understanding what this does, then propose documentation? (It's only a private method, so it's not urgent.)

amanda-tarafa: Okay, this is just like an indepth traversal of a forest, only I dind't use a stack, which would probably have made it clearer, because when I get to a terminal field I want to build the path wich is FIFO and not LIFO, I could have reversed the stack etc., but with the list I get both easier, FIFO for building the path, LIFO for backtracking.
I'll merge now, but happy to go over this later for improvements and docs.

PR #4094, opened 2020-01-08
Regenerate all APIs and tidy up

jskeet:
This will pull in multiple microgenerator API changes, including:

- Resource name redesign (with factory methods, and better multipattern support)
- The updated GAX retry settings
- The use of WithGoogleRequestParam instead of WithCallSettingsOverlay
- Exclusing obsolete properties from snippets and tests


jskeet:
I would suggest just a fairly casual review. I'm going to check that all the micro-generated APIs build. (The old ones won't at the moment due to GAX changes.)

jskeet:
Note to self:

```sh
./build.sh $(python tools/listapis.py apis/apis.json --test=generator=micro)
```

jskeet:
Ah - I need to regenerate some common resources, e.g. Google.Cloud.Bigtable.Common.V2.TableName.

jskeet:
- Bigtable.V2 still won't build as the test project depends on Bigtable.Admin.V2, which is on the monolith (so has old GAX Backoffsettings etc).
- The Container.V1 tests fail for reasons I'm going to look into imminently.

The rest now builds.
Adding Amanda as a reviewer for the fourth commit, which is about Spanner.

PR #4100, opened 2020-01-09
Migrate ContainerAnalysis.V1 to the microgenerator on the master branch

jskeet:
This is a breaking change, but we won't need to release updates to this library imminently anyway.

The migration happened on next-major-version yesterday, but the GAPIC v2 config doesn't work for this API - I've tried digging into why, but it's all a bit tortuous. It's simpler just to move to the microgenerator now.


jskeet:
I don't understand why there are merge conflicts, but I'll sort them out.

PR #4104, opened 2020-01-10
Tweaks to Grafeas generation

jskeet:
The first two commits (the small ones!) are manual; the third is generated.


generateapis.sh
```diff
@@ -67,12 +68,19 @@ generate_microgenerator() {
...
-  if [[ -f "$API_OUT_DIR/$1/CommonResourcesConfig.json" ]]
+  if [[ -f "$API_OUT_DIR/$PACKAGE_ID/CommonResourcesConfig.json" ]]
   then
-    COMMON_RESOURCES_CONFIG=$API_OUT_DIR/$1/CommonResourcesConfig.json
+    COMMON_RESOURCES_CONFIG=$API_OUT_DIR/$PACKAGE_ID/CommonResourcesConfig.json
   fi
   COMMON_RESOURCES_OPTION=--gapic_opt=common-resources-config=$COMMON_RESOURCES_CONFIG
   
+  # Only specify common resource protos for GCP APIs
+  COMMON_RESOURCES_PROTO=
```
chrisdunelm: Are you sure this form of conditional assignment works??

jskeet: Whoops, no - that should be `COMMON_RESOURCE_PROTOS=$GOOGLEAPIS/google/cloud/common_resources.proto`. Will fix!

(I'm now testing by generating Speech, seeing it fail, fixing, seeing it pass.)

PR #4178, opened 2020-01-16
[BigQuery]: Models CRUD operations

amanda-tarafa:
Reviewing one commit at a time will help.

I'm asking internally about Model.ETag and concurrent updates. Leaving the skipped integration test as is until I hear something.


amanda-tarafa:
This is now green. Problem was, before C#8 interpolated verbatim strings had to be specified with `$@` in that order (not `@$`). And this was building locally because I'm on C#8, but Travis and AppVeyor are not.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.ModelCrud.cs
```diff
@@ -0,0 +1,148 @@
...
+namespace Google.Cloud.BigQuery.V2
+{
+    public partial class BigQueryClientImpl
+    {
+        // Visible for testing
+        internal sealed class ModelPageManager : IPageManager<ListRequest, ListModelsResponse, BigQueryModel>
+        {
+            private readonly BigQueryClient _client;
+
+            internal ModelPageManager(BigQueryClient client)
```
jskeet: Maybe use an expression-bodied constructor now? (Possibly commit this as is, then we can create another PR that does this for all the page managers...)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryModel.cs
```diff
@@ -61,5 +63,64 @@ public BigQueryModel(BigQueryClient client, Model resource)
...
+        /// </summary>
+        /// <remarks>
+        /// This method delegates to <see cref="BigQueryClient.PatchModel(ModelReference, Model, PatchModelOptions)"/>.
+        /// </remarks>
+        /// <param name="resource">The resource to patch with. Must not be null.</param>
+        /// <param name="matchETag">If true, the etag from <see cref="Resource"/> is propagated into <paramref name="resource"/> for
+        /// optimistic concurrency. Otherwise, <paramref name="resource"/> is left unchanged.</param>
+        /// <param name="options">The options for the operation. May be null, in which case defaults will be supplied.</param>
+        /// <returns>The updated model.</returns>
+        public BigQueryModel Patch(Model resource, bool matchETag, PatchModelOptions options = null)
```
jskeet: I'm assuming this is what we've done elsewhere? Will think more carefully if this is "new thinking" :)

amanda-tarafa: Yes, it was done for Table and Dataset which are the ones that could be patched/updated until now.
But, I've had to skip [the test](https://github.com/googleapis/google-cloud-dotnet/blob/c118c5eae2be64b2d4c9944451270c6308772789/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/ModelTests.cs#L53) that test for ETag being used for avoiding concurrent modifications in Model because it is not passing. Similar tests are passing for Table and Dataset.
I updated the internal bug asking @shollyman about this.
Maybe we can remove the matchETag parameter from all, given that these methods already accept a resource, and take the breaking change (for Table and Dataset) into the next version?

jskeet: I think it's useful to have the matchETag parameter, personally. I *think* we've had it from the start, and it's useful in terms of the difference between "only patch I know it's the latest" or "unconditionally patch".
It would be good to know why the test is failing though.

amanda-tarafa: OK, I don't feel very strongly about keeping the matchETag parameter or not. But I'll wait for @shollyman to say something before merging. I believe that this is simply a bug on the API that's not taking the ETag into account (I believe the test and our production code are OK, it's mostly the same as the one for Table and Dataset), but I'll wait in case they are not planning on using ETag for concurrent modifications in Model, because then having the parameter does not make sense for Model.

amanda-tarafa:
Mergin after having created #4232

PR #4183, opened 2020-01-20
CI .NET SDK update.

amanda-tarafa:
This updates both Travis and AppVeyor.

I believe we also needed the testing runtime update in Travis since we are no longer testing against .NET Core 1.* but against .NET Core 2.1, but not entirely sure about this.


jskeet:
We can't do this yet in the master branch - Kokoro used for Autosynth won't support .NET Core 3.1.

I suggest making the change in the GAX repo first, as Kokoro doesn't need that. It will also have a quicker turnaround.
I believe you're right about not testing against 1.x any more - and certainly that'll be the case moving forward.

amanda-tarafa:
Closing this PR but keeping the branch since these seem like the correct changes.

Issue #4199, opened 2020-01-28
[BigQuery] Discussion about when to throw in insert operations (revisit query operations as well)


amanda-tarafa:
#4200 Makes insert methods return the response object (or a wrapper) with information about rows that failed to insert. These methods were previously void, so the only way to obtain information about failed rows was through a GoogleApiException that was thrown. The SuppressInsertErrors option controlled when the exception was thrown or not, false by default, if set to true the exception wouldn't be thrown. This made more sense when used in conjunction with the SkipInvalidRows options set to true, which meant that the caller was OK with a partial insert (and not knowing about failed rows).
Now that the insert methods are not void should we keep the SuppressInsertErrors option? Main worry is about side effects for code that never set this option and thus has been expecting an exception in all cases except when an insert is completely successful, i.e. all rows are inserted. Code like this will still build, but might not function properly.

amanda-tarafa:
Options are, I think (all of these are about insert failures because of incorrect rows, all other failures, like network, access, etc. will be thrown regardless).

1. Leave as is in #4200: for all but complete success throw, unless SuppressInsertErrors is true.
**Pros**: Code that has never set SuppressInsertErrors won't be affectedm they'll get the same exceptions as always. This is the least breaking.
**Cons**: To me, now SkipInvalidRows and SuppressInsertErrors seem redundant. If I already set SkipInvalidRows I'm saying that I'm OK with a partial insert, so I don't expect to get an exception when that happens, or to have to set another option so that I don't get the exception.
2. Remove SuppressInsertErrors and never throw, the response object contains enough information so as to know whether the failure was complete or partial and also has some helper ThrowOnAnyError() methods.
**Pros**: We always get the best info possible, which is contained in the response object and not the exception.
**Cons**: It's not idiomatic to not throw when the insert failed completely.
This will break all clients that were relying on an exception on any insert errors.
3. Remove SuppressInsertErrors and only throw when no rows were inserted because either SkipInvalidRows is false and some rows were invalid or all rows were invalid (that is SkipInvalidRows means we are OK with a partial insert not with a full failure).
**Pros**: To me this seems the more idiomatic solution.
**Cons**: The exception contains less information than the response, and when the exception is thrown there is no way to access the response (maybe include the response in the exception, but where).
This will break all clients that were relying on exceptions on partial failures.

Personally I would favour 3 except for the fact that we don't include all the info in the exception. If we don't find a good way to include the response on the exception then I'd favour 2.
[...]

PR #4200, opened 2020-01-28
[BigQuery] Proposal: Have insert methods return an insert response wrapper.

amanda-tarafa:
The API insert operation returns a response object with information about the rows that contained errors, if any, which is specially relevant if the option SkipInvalidRows is set to true. In that case the insert operation is partially successful and the response contains information about the rows that were not inserted and why.
The client library insert operations were, until now, void. The only way to obtain part of the information contained in the reponse object has been trough an exception that was thrown depending on another option SuppressInsertErrors. Even then, the information contained in the exception was less machine readable than the information in the response, for instance, in the exception, the index of each of the erred rows is part of a message and cannot be accessed by itself.

In this proposal the BigQueryInsertResults wraps the response object. I believe this wrapper is in line with more or less everything in the library but I also understand that it is more code for us to maintain so I'd be fine with simple returning the response object or a simpler wrapper.

In this proposal the SuppressInsertErrors options has been, that is, unless it is set to true, insert methods will throw unless **all** rows are inserted. Even if SkipInvalidRows is set to true, if some rows were skipped, the methods would throw. For a discussion around this see #4199 .

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
@@ -30,6 +30,12 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
             _fixture = fixture;
         }
 
+        private void AssertAllRowsInserted(BigQueryInsertResults insertResult)
+        {
+            Assert.Equal(BigQueryInsertStatus.AllRowsInserted, insertResult.Status);
+            Assert.Empty(insertResult.Errors);
+        }
+
         [Fact]
         public void InsertRow()
         {
@@ -39,7 +45,9 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
 
             var row = BuildRow("Joe", 100, new DateTime(2016, 4, 26, 11, 43, 1, DateTimeKind.Utc));
 
-            _fixture.InsertAndWait(table, () => table.InsertRow(row), 1);
+            var insertResult = _fixture.InsertAndWait(table, () => table.InsertRow(row), 1);
+
+            AssertAllRowsInserted(insertResult);
 
             var rowsAfter = table.ListRows();
             var fetched = rowsAfter.Single(r => (string)r["player"] == "Joe");
@@ -60,7 +68,9 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
                 BuildRow("Lisa", 90, new DateTime(2011, 10, 12, 0, 0, 0, DateTimeKind.Utc))
             };
 
-            _fixture.InsertAndWait(table, () => table.InsertRows(rows), 2);
+            var insertResult = _fixture.InsertAndWait(table, () => table.InsertRows(rows), 2);
+
+            AssertAllRowsInserted(insertResult);
 
             var rowsAfter = table.ListRows().ToList();
             Assert.Contains(rowsAfter, r => (string)r["player"] == "Jenny");
@@ -81,10 +91,12 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
                 BuildRow("Henry", 90, new DateTime(2011, 10, 12, 0, 0, 0, DateTimeKind.Utc))
             };
 
-            _fixture.InsertAndWait(table, () => table.InsertRows(rows, options), 2);
+            var insertResult = _fixture.InsertAndWait(table, () => table.InsertRows(rows, options), 2);
 
             Assert.Null(rows[0].InsertId);
             Assert.Null(rows[1].InsertId);
+
+            AssertAllRowsInserted(insertResult);
 
             var rowsAfter = table.ListRows().ToList();
             Assert.Contains(rowsAfter, r => (string)r["player"] == "Helen");
@@ -95,22 +107,25 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
         {
             get
             {
-                yield return new object[] { null };
-                yield return new object[] { new InsertOptions() };
-                yield return new object[] { new InsertOptions { SkipInvalidRows = false } };
-                yield return new object[] { new InsertOptions { SkipInvalidRows = true } };
-                yield return new object[] { new InsertOptions { AllowUnknownFields = false } };
-                yield return new object[] { new InsertOptions { AllowUnknownFields = true } };
-                yield return new object[] { new InsertOptions { SkipInvalidRows = false, AllowUnknownFields = false } };
-                yield return new object[] { new InsertOptions { SkipInvalidRows = false, AllowUnknownFields = true } };
-                yield return new object[] { new InsertOptions { SkipInvalidRows = true, AllowUnknownFields = false } };
-                yield return new object[] { new InsertOptions { SkipInvalidRows = true, AllowUnknownFields = true } };
+                int[] bothRowsIndexes = new int[] { 0, 1 };
+                int[] oneRowIndex = new int[] { 1 };
+
+                yield return new object[] { null, bothRowsIndexes };
+                yield return new object[] { new InsertOptions(), bothRowsIndexes };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = false }, bothRowsIndexes };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = true }, bothRowsIndexes };
+                yield return new object[] { new InsertOptions { AllowUnknownFields = false }, bothRowsIndexes };
+                yield return new object[] { new InsertOptions { AllowUnknownFields = true }, bothRowsIndexes };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = false, AllowUnknownFields = false }, bothRowsIndexes };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = false, AllowUnknownFields = true }, bothRowsIndexes };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = true, AllowUnknownFields = false }, bothRowsIndexes };
+                yield return new object[] { new InsertOptions { SkipInvalidRows = true, AllowUnknownFields = true }, oneRowIndex };
             }
         }
 
         [Theory]
         [MemberData(nameof(BadDataThrowsOptions))]
-        public void InsertRow_BadData_Throws(InsertOptions options)
+        public void InsertRow_BadData_Throws(InsertOptions options, int[] errorRowsIndexes)
         {
             var client = BigQueryClient.Create(_fixture.ProjectId);
             var dataset = client.GetDataset(_fixture.DatasetId);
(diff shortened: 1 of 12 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryInsertRowErrors.cs
```diff
@@ -0,0 +1,91 @@
...
+            OriginalRow = originalRow;
+        }
+
+        /// <summary>
+        /// Returns an interator over the insert errors.
+        /// </summary>
+        /// <returns>An iterator over the insert errors.</returns>
+        public IEnumerator<SingleError> GetEnumerator()
+        {
+            string messageSuffix = $"Error in row {OriginalRowIndex?.ToString() ?? "unknown"}.";
```
jskeet: Nit: this is a prefix, not a suffix :)

amanda-tarafa:
> Just a review of the production code from the first commit. (I haven't reviewed the tests or the second commit yet. This review will be spotty, basically - let me know if you'd rather I waited until I can do it all in one go!)

Yes, this is totally fine. This is the most important bit anyway. I'll address your comments now, thanks!

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryInsertResults.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// </summary>
+        /// <param name="client">The client used for the insert request. Must not be null.</param>
+        /// <param name="options">The options used for the insert request. May be null.</param>
+        /// <param name="originalRows">The rows whose insert was attempted. Must not be null.</param>
+        /// <param name="insertResponse">The response obtained after attempting the insert. Must not be null.</param>
+        public BigQueryInsertResults(BigQueryClient client, InsertOptions options, IEnumerable<BigQueryInsertRow> originalRows, TableDataInsertAllResponse insertResponse)
+        {
+            GaxPreconditions.CheckNotNull(insertResponse, nameof(insertResponse));
+            _client = GaxPreconditions.CheckNotNull(client, nameof(client));
+            _originalRows = GaxPreconditions.CheckNotNull(originalRows, nameof(originalRows)).ToList();
```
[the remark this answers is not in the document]
jskeet: If that's all we expose, that would be fine.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/InsertTest.cs
```diff
@@ -185,7 +216,11 @@ public void InsertRow_BadData_IgnoreExtraColumn()
             // Even though SuppressInsertErrors is false, this won't throw
             // because server side, unknown fields are ignored silently.
             var options = new InsertOptions { AllowUnknownFields = true };
-            _fixture.InsertAndWait(table, () => table.InsertRow(row, options), 1);
+            var insertResult = _fixture.InsertAndWait(table, () => table.InsertRow(row, options), 1);
+
+            Assert.False(insertResult.HasErrors);
```
jskeet: I think we probably want a helper method for this set of assertions.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryInsertResultsTests.cs
```diff
@@ -0,0 +1,426 @@
...
+                new BigQueryInsertRow(),
+            };
+
+            List<BigQueryInsertRowErrors> expectedInsertRowErrors = new List<BigQueryInsertRowErrors>
+            {
+                new BigQueryInsertRowErrors(rows[1], new List<InsertErrorsData> { row1List1, row1List2 }),
+                new BigQueryInsertRowErrors(rows[5], new List<InsertErrorsData> { row5List1 }),
+                new BigQueryInsertRowErrors(rows[6], new List<InsertErrorsData> { row6List1 }),
+            };
+            List<SingleError> expectedSingleErrors = (from rowError in expectedInsertRowErrors
```
jskeet: I'd replace the RHS with `expectedInsertRowErrors.SelectMany(rowError => rowError).ToList()`

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryInsertResultsTests.cs
```diff
@@ -0,0 +1,426 @@
...
+            BigQueryInsertResults results = new BigQueryInsertResults(
+                _client,
+                new InsertOptions { SkipInvalidRows = true },
+                rows,
+                response);
+
+            Assert.Equal(BigQueryInsertStatus.SomeRowsInserted, results.Status);
+            Assert.True(results.HasErrors);
+            Assert.Same(results, results.ThrowOnNoneInserted());
+            AssertException(
```
jskeet: Perhaps the Assert.Throws call could be in AssertException? Then it could use a method group conversion and read really nicely:

```csharp
AssertException(results.ThrowOnNoneInserted);
AssertException(results.ThrowOnNotAllInserted);
AssertException(results.ThrowOnAnyError);
```

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryInsertResultsTests.cs
```diff
@@ -0,0 +1,426 @@
...
+            ErrorProto row1Error2 = new ErrorProto { Location = "field_2", Message = "message_2" };
+            InsertErrorsData row1List1 = new InsertErrorsData { Index = 1, Errors = new List<ErrorProto> { row1Error1, row1Error2 } };
+
+            ErrorProto row5Error1 = new ErrorProto { Location = "field_3", Reason = "reason_3" };
+            InsertErrorsData row5List1 = new InsertErrorsData { Index = 5, Errors = new List<ErrorProto> { row5Error1 } };
+
+            InsertErrorsData row6List1 = new InsertErrorsData { Index = 6 };
+
+            ErrorProto row1Error3 = new ErrorProto { Location = "field_4", Reason = "reason_4", Message = "message_4" };
+            InsertErrorsData row1List2 = new InsertErrorsData { Index = 1, Errors = new List<ErrorProto> { row1Error3 } };
```
jskeet: I don't currently understand how there would be two separate sets of errors from the same row. I hope I'll understand it later...

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryInsertResultsTests.cs
```diff
@@ -0,0 +1,426 @@
...
+            AssertException(
+                Assert.Throws<GoogleApiException>(() => results.ThrowOnNoneInserted()));
+            AssertException(
+                Assert.Throws<GoogleApiException>(() => results.ThrowOnNotAllInserted()));
+            AssertException(
+                Assert.Throws<GoogleApiException>(() => results.ThrowOnAnyError()));
+
+            Assert.Equal(expectedInsertRowErrors, results.Errors.ToList(), new BigQueryInsertRowErrorEqualityComparer());
+
+            void AssertException(GoogleApiException exception)
```
jskeet: Can we add some parameters and hoist this method out of the individual tests?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryInsertResultsTests.cs
```diff
@@ -0,0 +1,426 @@
...
+                new BigQueryInsertRow(),
+                new BigQueryInsertRow(),
+            };
+
+            List<BigQueryInsertRowErrors> expectedInsertRowErrors = new List<BigQueryInsertRowErrors>
+            {
+                new BigQueryInsertRowErrors(rows[1], new List<InsertErrorsData> { row1List1, row1List2 }),
+                new BigQueryInsertRowErrors(rows[5], new List<InsertErrorsData> { row5List1 }),
+                new BigQueryInsertRowErrors(rows[6], new List<InsertErrorsData> { row6List1 }),
+            };
```
jskeet: Why have we only got errors for three rows, if no rows were inserted?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryInsertResultsTests.cs
```diff
@@ -0,0 +1,426 @@
...
+                InsertErrors = new List<InsertErrorsData>
+                {
+                    row1List1,
+                    row5List1,
+                    row6List1,
+                    row1List2
+                }
+            };
+
+            IList<BigQueryInsertRow> rows = new List<BigQueryInsertRow>
```
jskeet: We might want a helper method for this, rather than taking 11 lines in each test. Or just use `Enumerable.Range(0, 8).Select(_ => new BigQueryInsertRow()).ToList()`

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryInsertRowErrorsTests.cs
```diff
@@ -0,0 +1,162 @@
...
+            {
+                new InsertErrorsData
+                {
+                    Index = null,
+                    Errors = new List<ErrorProto>
+                    {
+                        new ErrorProto
+                        {
+                            Reason = "Very bad reason",
+                            Message = "A very bad error occured"
```
jskeet: Nit: occurred, here and below

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryInsertRowErrorsTests.cs
```diff
@@ -0,0 +1,162 @@
...
+                },
+                new SingleError
+                {
+                    Reason = "Reason 3",
+                    Message = "Error in row 1. A very bad error 3 occured",
+                    Location = "field_3"
+                }
+            };
+
+            // This is allowed in case there's no information (index)
```
jskeet: We've passed null as the first constructor argument in all the tests. Shouldn't we have some tests for non-null arguments?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.InsertData.cs
```diff
@@ -136,39 +135,20 @@ private BigQueryJob UploadData(JobConfigurationLoad loadConfiguration, Stream in
         }
 
         /// <inheritdoc />
-        public override void InsertRows(TableReference tableReference, IEnumerable<BigQueryInsertRow> rows, InsertOptions options = null)
+        public override BigQueryInsertResults InsertRows(TableReference tableReference, IEnumerable<BigQueryInsertRow> rows, InsertOptions options = null)
         {
             var request = CreateInsertAllRequest(tableReference, rows, options, out bool hasRows);
```
jskeet: This is where I'm expecting something will be creating a List<BigQueryInsertRow>` internally. We may be able to be more efficient. I'll look once most of this is in, as it'll be simpler to work out then.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.InsertData.cs
```diff
@@ -136,39 +135,20 @@ private BigQueryJob UploadData(JobConfigurationLoad loadConfiguration, Stream in
...
-        }
-
-        private void HandleInsertAllResponse(TableDataInsertAllResponse response, InsertOptions options)
-        {
-            var errors = response.InsertErrors;
-            bool shouldThrow = options == null || !options.SuppressInsertErrors;
-            if (errors?.Count > 0 && shouldThrow)
+            TableDataInsertAllResponse response = request.Execute();
+            BigQueryInsertResults result = new BigQueryInsertResults(this, options, rows, response);
+            if(options?.SuppressInsertErrors == true)
```
jskeet: `return options?.SuppressInsertErrors == true ? result : result.ThrowOnAnyError();`

jskeet: Or move it into an internal helper method in BigQueryInsertResults accepting a `bool?`, and call it from here and the async one.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryInsertResults.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// </summary>
+        public BigQueryInsertStatus Status
+        {
+            get
+            {
+                if (!HasErrors)
+                {
+                    return BigQueryInsertStatus.AllRowsInserted;
+                }
+                if (_errors.Count >= _originalRowsCount ||
```
jskeet: Unwrap this? I think it would be simpler to read as a single line.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryInsertResults.cs
```diff
@@ -0,0 +1,137 @@
...
+        private readonly BigQueryClient _client;
+        private readonly InsertOptions _options;
+        private readonly int _originalRowsCount;
+        private IReadOnlyList<BigQueryInsertRowErrors> _errors;
+
+        /// <summary>
+        /// Status of the insert request.
+        /// See <see cref="BigQueryInsertStatus"/> for a detailed explanation.
+        /// </summary>
+        public BigQueryInsertStatus Status
```
jskeet: While I'm okay having the status, I wonder whether having an integer property for the number of rows inserted would be useful.

Is there ever a case where we provide 10 rows, there are no errors, but only 5 rows are inserted, e.g. because of duplicate IDs?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryInsertResults.cs
```diff
@@ -0,0 +1,137 @@
...
+                   _options?.SkipInvalidRows != true)
+                {
+                    return BigQueryInsertStatus.NoRowsInserted;
+                }
+
+                return BigQueryInsertStatus.SomeRowsInserted;
+            }
+        }
+
+        /// <summary>
```
jskeet: Not sure this is particularly useful - anyone can use `_errors.Any()` if they want, for example.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryInsertResults.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// see <see cref="Status"/> if you need to know the exact
+        /// outcome of the insert request.
+        /// </summary>
+        public bool HasErrors => _errors.Count > 0;
+
+        /// <summary>
+        /// Returns a sequence over rows with insert errors.
+        /// Each element represents a row with possibly multiple errors associated to it.
+        /// </summary>
+        /// <returns> A sequence over rows with insert errors.</returns>
```
jskeet: Nit: remove leading space.

Also, let's say that it will never be null. (And if that's not the case now, let's make sure it is :)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryInsertResults.cs
```diff
@@ -0,0 +1,137 @@
...
+        /// </summary>
+        /// <param name="client">The client used for the insert request. Must not be null.</param>
+        /// <param name="options">The options used for the insert request. May be null.</param>
+        /// <param name="originalRows">The rows whose insert was attempted. Must not be null.</param>
+        /// <param name="insertResponse">The response obtained after attempting the insert. Must not be null.</param>
+        public BigQueryInsertResults(BigQueryClient client, InsertOptions options, IEnumerable<BigQueryInsertRow> originalRows, TableDataInsertAllResponse insertResponse)
+        {
+            GaxPreconditions.CheckNotNull(insertResponse, nameof(insertResponse));
+            _client = GaxPreconditions.CheckNotNull(client, nameof(client));
+            var rows = GaxPreconditions.CheckNotNull(originalRows, nameof(originalRows)).ToList();
```
jskeet: I would hope we can do this without calling ToList. Will look after this is in.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryInsertRowErrors.cs
```diff
@@ -0,0 +1,91 @@
...
+            OriginalRow = originalRow;
+        }
+
+        /// <summary>
+        /// Returns an interator over the insert errors.
+        /// </summary>
+        /// <returns>An iterator over the insert errors.</returns>
+        public IEnumerator<SingleError> GetEnumerator()
+        {
+            string messagePrefi = $"Error in row {OriginalRowIndex?.ToString() ?? "unknown"}.";
```
jskeet: Nit: Prefi => Prefix

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.InsertData.cs
```diff
@@ -341,5 +321,71 @@ private InsertAllRequest CreateInsertAllRequest(TableReference tableReference, I
             RetryHandler.MarkAsRetriable(request);
             return request;
         }
+
+        /// <summary>
+        /// Wrapper list so that we can be more efficient and reuse the BigQueryInsertRow
+        /// list.
+        /// This list is really read only only the type needs to be IList.
```
jskeet: Nit: only only

(It would be good to say *why* it has to be `IList<T>` - I assume it's because of the type of `TableDataInsertAllRequest.Rows`.

PR #4205, opened 2020-01-30
Update Grafeas now that it has a Project resource

### LINKED ISSUE #4102: Grafeas.V1 should have its own Project resource type
Currently the [Grafeas proto](https://github.com/googleapis/googleapis/blob/master/grafeas/v1/grafeas.proto) file uses `"cloudresourcemanager.googleapis.com/Project"` as the resource type for the parent of various things.
Given that Grafeas is vendor-neutral, this should really be `"grafeas.io/Project"`. That requires a bit of work in the generation script. I'll get it working locally before trying to work out how to get the proto change in.

jskeet:
Fixes #4102 

(There'll be more regeneration soon, but I regenerated for now to prove that it does actually work...)


jskeet:
Hold off reviewing - will do this after the client factory generation change is merged.

jskeet:
Closing for now, in fact... we need to work out what to do about GrafeasClientBuilder, which is removed currently (due to there being no default endpoint).

PR #4207, opened 2020-01-31
feat: add support to Cloud Spanner client to connect to an emulator

skuruppu:
If the SPANNER_EMULATOR_HOST environment variable is set, the connection will be made to the given host and port. The channel credentials will be set to use insecure credentials.

Usage:
export SPANNER_EMULATOR_HOST="localhost:9010"


skuruppu:
@amanda-tarafa and @jskeet, I am unable to run the ITs or snippets yet due to not having an emulator to run against. But I thought I'd ask you for early feedback given that this is the first C# PR I've worked on and I'm not yet familiar with this code.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -50,10 +50,14 @@ public sealed class SpannerConnectionStringBuilder : DbConnectionStringBuilder
         private const string DataSourceKeyword = "Data Source";
         private const string UseClrDefaultForNullKeyword = "UseClrDefaultForNull";
         private const string EnableGetSchemaTableKeyword = "EnableGetSchemaTable";
+        private const string EmulatorHostVariable = "SPANNER_EMULATOR_HOST";
 
         private InstanceName _instanceName;
         private DatabaseName _databaseName;
 
+        private string _emulatorHost;
+        private Int32 _emulatorPort;
```
amanda-tarafa: nit: use `int` instead of `Int32`, `long` instead of `Int64`, `bool` instead of `Boolean`, etc.
Also, you can use `int? _emulatorPort`, `int?` represents the nullable int type.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -138,6 +142,35 @@ private string ValidatedDataSource(string dataSource)
             return dataSource;
         }
 
+        // Parse the emulator host and port if the SPANNER_EMULATOR_HOST environment variable is set.
+        private void ParseEmulatorHostAndPort()
+        {
+            // Emulator host and port have already been parsed.
+            if (!string.IsNullOrEmpty(_emulatorHost))
```
amanda-tarafa: nit: only need to check for null here, all instance variables are initialized to the type's default value and that is null for string, which is a reference type.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -148,21 +181,37 @@ private string ValidatedDataSource(string dataSource)
...
-        /// host will be used.
+        /// host will be used. If the SPANNER_EMULATOR_HOST environment variable is set, the host specified
+        /// in its value will be used as the default host.
         /// </summary>
         public string Host
         {
-            get => GetValueOrDefault(nameof(Host), SpannerClient.DefaultEndpoint.Host);
+            get
+            {
+                if (ContainsKey(nameof(Host)))
```
amanda-tarafa: So, if the Host is explicitly set it takes presedence over the emulator?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -148,21 +181,37 @@ private string ValidatedDataSource(string dataSource)
...
+        /// in the value will be used as the default port.
         /// </summary>
         public int Port
         {
-            get => GetInt32OrDefault(nameof(Port), 1, 65535, SpannerClient.DefaultEndpoint.Port);
+            get
+            {
+                ParseEmulatorHostAndPort();
+                return GetInt32OrDefault(nameof(Port), 1, 65535,
+                    (string.IsNullOrEmpty(_emulatorHost)) ? SpannerClient.DefaultEndpoint.Port : _emulatorPort);
```
amanda-tarafa: nit: You can check for null only, and if you change the type of `_emulatorPort` to `int?` you can do `_emulatorPort ?? Spanner.DefaultEndpoint.Port`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -148,21 +181,37 @@ private string ValidatedDataSource(string dataSource)
...
-        /// port will be used.
+        /// port will be used.  If the SPANNER_EMULATOR_HOST environment variable is set, the port specified
+        /// in the value will be used as the default port.
         /// </summary>
         public int Port
         {
-            get => GetInt32OrDefault(nameof(Port), 1, 65535, SpannerClient.DefaultEndpoint.Port);
+            get
+            {
+                ParseEmulatorHostAndPort();
```
amanda-tarafa: Why are Port and Host implemented differently? For clarity and consistency they should be implemented the same way.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -334,14 +383,19 @@ public SpannerConnectionStringBuilder(string connectionString, ChannelCredential
         public SpannerConnectionStringBuilder(string connectionString, ChannelCredentials credentials, SessionPoolManager sessionPoolManager)
         {
             ConnectionString = GaxPreconditions.CheckNotNull(connectionString, nameof(connectionString));
-            CredentialOverride = credentials;
+            // If SPANNER_EMULATOR_HOST env var is set, ignore any provided credentials and use insecure credentials.
+            ParseEmulatorHostAndPort();
+            CredentialOverride = string.IsNullOrEmpty(_emulatorHost) ? credentials : ChannelCredentials.Insecure;
```
amanda-tarafa: I believe this is wrong. If the Host and the Port were already specified in the connection string, the emulator will be ignored even if it is set (that's how Host and Port are implemented, the specified values take precedence over the emulator values).
In here you cannot override the credentials if the emulator is set, you can override them if the emulator is set and Host and Port were not set in the connection string.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -334,14 +383,19 @@ public SpannerConnectionStringBuilder(string connectionString, ChannelCredential
         public SpannerConnectionStringBuilder(string connectionString, ChannelCredentials credentials, SessionPoolManager sessionPoolManager)
         {
             ConnectionString = GaxPreconditions.CheckNotNull(connectionString, nameof(connectionString));
-            CredentialOverride = credentials;
+            // If SPANNER_EMULATOR_HOST env var is set, ignore any provided credentials and use insecure credentials.
+            ParseEmulatorHostAndPort();
```
amanda-tarafa: You parse here, but also you keep parsing whenever the Host or the Port are requested, so it might be that at the time of construction the emulator was not set, but that it is set when Host or Port are requested, and then the credentials would be wrong.
I would only parse on construction.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -334,14 +383,19 @@ public SpannerConnectionStringBuilder(string connectionString, ChannelCredential
...
+            CredentialOverride = string.IsNullOrEmpty(_emulatorHost) ? credentials : ChannelCredentials.Insecure;
             SessionPoolManager = GaxPreconditions.CheckNotNull(sessionPoolManager, nameof(sessionPoolManager));
         }
 
         /// <summary>
         /// Creates a new <see cref="SpannerConnectionStringBuilder"/>.
         /// </summary>
-        public SpannerConnectionStringBuilder() { }
+        public SpannerConnectionStringBuilder() : this("")
```
amanda-tarafa: null, not "". In general avoid the empty string as a signal for an unknown value. It carries the same problems as a 0 for an unknown int (thankfully we now have int?).

skuruppu:
> I reviewed the PR as is. But please consider changing this implementation to match support for emulators in other client libraries. As examples:
> Firestore: #3697
> Datastore: #2998

Thank you for the advice @amanda-tarafa and @jskeet. I changed the implementation to match that of existing implementations.

@snehashah16 was however concerned about this implementation. The reason why we we're using an environment variable is because we want users to be able to use the emulator without having to recompile the application. It seems that this is impossible to do once the enum is introduced because customers would need to recompile their app.

The particular use case that @snehashah16 was most concerned about is customers who have a Travis CI setup for their production app. They would want to run tests against the emulator but run the app against prod. They would need to maintain two binaries. Or they can by set the connection to use EmulatorDetection.ProductionOrEmulator but that seems to defeat the purpose of having this enum to protect against folks accidentally spending a lot of money on Cloud Spanner.

Would like to hear your thoughts about this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
skuruppu: No longer making this change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
skuruppu: No longer using this logic.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
skuruppu: No longer relevant since the implementation has changed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
skuruppu: No longer relevant since the implementation changed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
skuruppu: #TIL

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
skuruppu: No longer an issue since the implementation has changed.

skuruppu:
@amanda-tarafa @jskeet I have another question. Is there a way to get the Instance and Database admin clients to respect the environment variable? Or is it the case that they're all auto-generated so the only way to set the endpoint is for customers to do the following:

```
string hostAndPort = Environment.GetEnvironmentVariable("SPANNER_EMULATOR_HOST")?.Trim();
ServiceEndpoint emulator = ServiceEndpoint.Parse(hostAndPort);
Channel channel = new Channel(
        emulator.Host, emulator.Port, Grpc.Core.ChannelCredentials.Insecure);
InstanceAdminClient client = InstanceAdminClient.Create(channel);
```

amanda-tarafa:
> The reason why we we're using an environment variable is because we want users to be able to use the emulator without having to recompile the application. It seems that this is impossible to do once the enum is introduced because customers would need to recompile their app.

Hi @skuruppu, in those cases, the client code would use the `EmulatorDetector.ProductionOrEmulator` enum value to indicate a preference for connecting to the emulator if the env variable is present, else to production. Which is basically no different to your original approach.
Also, client code could have other configuration means, say it can detect the environment is running in, and then, depending on that, it could set one `EmulatorDetector` value or the other.

Basically, always using `EmulatorDetector.ProductionOrEmulator` is equivalent to your original approach. The rest gives client code extra granularity and control, if they need it.

amanda-tarafa:
> @amanda-tarafa @jskeet I have another question. Is there a way to get the Instance and Database admin clients to respect the environment variable? 

@jskeet please confirm that you are OK with the following approach and that I'm not missing anything.

@skuruppu Even though both of those libraries are completely autogenerated, we can add custom bits of code in new files that won't get overwritten when the library is regenarated. For Database admin client (the same applies to Instance admin client):

The class `DatabaseAdminClientBuilder` is a partial class which means you can create a new file with a partial declaration of it, the result is no different than adding some code to the generated file itself except for the fact that the added code won't be deleted when regenerating.
This class inherits from `ClientBuilderBase<TClient>` which has a method `protected virtual CallInvoker CreateCallInvoker()` (and its async counterpart). This means that in your custom partial class declaration you can override this method (luckily it is not overriden by the generated code) . 
So, what I would do is, add and `EmulatorDetector` enum (the same as you did for Spanner) to the Database admin client. Create a partial class definetion for `DatabaseAdminClientBuilder` and in this partial definition add an `EmulatorDetection` property and override the `CreateCallInvoker` and `CreateCallInvokerAsync` methods. When overriding the methods you need to check the `EmulatorDetection` property and if the emulator should be used, then create a `CallInvoker` that points to the emulator. Else just call the base implementation of `CreateCallInvoker` so it does as usual.

Let me know if I'm not being clear, I can probably create a draft PR for one of the clients that you can then finish off and use as a sample for the other admin client.

jskeet:
I'm afraid I won't get to this until Friday at the very earliest, and it's more likely to be next week. Sorry - I'm just juggling too many other things at the moment.

jskeet:
@amanda-tarafa: I think by the time we've got lots of different EmulatorDetection enums, we should probably put this into GAX in the next major version - along with at least *some* helper methods (although the different emulators have slightly different environment variables etc). But that's mostly for the future.

One tricky part is that some uses of just Spanner.Data implicitly connect to the admin APIs. I wonder whether we should make *only Spanner.Data* know about emulators, but provide an API for accessing the admin APIs from that. Then we can have a single idea of whether or not we're talking to emulators, and be consistent in that usage.

I'd also suggest that the partial classes here are probably more fiddly than I'd be happy asking @skuruppu to come up with, as they involve a fair amount of knowledge of our client builder hierarchy. One of us should probably take this on, as a more efficient way of getting this done.

amanda-tarafa:
+1 to adding support for emulators in GAX. I was thinking even if it made sense to add an annotation for this so the microgenerators could include the code themselves. Don't know if this might be out of scope for annotations though.

About adding it only to Spanner.Data, I'm concerned about people who might only need to depend on the admin APIs now having to depend on Spanner.Data for emulator support? This will have them depending on ADO.NET even when they might not need it. The usage of the admin APIs from Spanner.Data should not be a problem, we can propagate the setting from Data to the  admin libs. The problem might be with an application actively using both and forgetting to set the same emulator support. That probably would fail fast but...  Documentation? An analyzer even? A small Spanner.EmulatorSupport package that all Spanner libraries depend on and allows for centralized emulator management?

I'm happy to submit a PR for one of the admin APIs and at least we can see how it'd look like with the partial classes etc. I can have that in on Monday.

jskeet:
Hmm. If we add a common emulator enum to GAX, we can depend on it in the next major version... I wonder whether we should do all this work in that branch. I think we should at least make it feasible in Spanner.Data, and then all three clients would need to be lined up - so we do need some kind of commonality. But yes, I take your point about "admin only" users - if each client supports it directly, then Spanner.Data can just do the same thing for all of them.

skuruppu:
Thanks @amanda-tarafa and @jskeet for the options. I have started work on adding the partial class using the [Bigtable version](https://github.com/googleapis/google-cloud-dotnet/blob/5ab2822deb4c5baffa1658ca3b959c8d725e9a14/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs#L198) of this as an example. I haven't got it working yet and won't get around to it till tomorrow afternoon so if you get to it before me, then that's totally fine. I realized that the lack of admin client support here was what was stopping me from getting the integration tests running with the emulator so that's why I started working on Amanda's idea last Friday.

As for adding the emulator enum to GAX, +1000 for that idea because I'm not a fan of redefining the same enum in multiple places (3 places just in the Spanner client for each API).

When is the next major version going out? I think we want this work to be in a release before the end of March at the latest.

jskeet:
We're hoping to release a beta of GAX v3 and all libraries in the next week or two, and ideally go to GA within a few weeks of that - around mid-March, perhaps, but it will absolutely depend on feedback. I think we could commit to a beta being available with this functionality being availbale before the end of March, but I wouldn't like to *guarantee* a GA release - particularly if there are other features to get done and merge in.

Having said that, I think it would be much better to work on the basis of the next major version branch rather than doing a load of work in this branch, only for that work to be effectively wasted pretty soon.

Amanda is going to put together a PR to get as much common code as possible in GAX, and then we can work on the Spanner-specific changes in this repo, in the next-major-version branch. I'd advise any other work trying to get this done in parallel.

Thinking about the user experience, I'd expect the following:

- A connection string option where the value matches the enum name, e.g. EmulatorDetection=ProductionOrEmulator; this would affect all clients created by Google.Cloud.Spanner.Data
- Properties in the client builders for each individual client (Spanner.V1, Spanner.Admin.Instance.V1, Spanner.Admin.Database.V1) for users not using the ADO.NET integration

jskeet:
The support for emulators is now in GAX. I'll create a PR to add that support to the builders for the three clients involved, and we can then revisit how that should be integrated into Spanner.Data.

skuruppu:
> The support for emulators is now in GAX. I'll create a PR to add that support to the builders for the three clients involved, and we can then revisit how that should be integrated into Spanner.Data.

Thanks @jskeet. I will close this PR then and I can open a new one based on what needs to be done.

jskeet:
@skuruppu: I'm happy for you to create a PR (bearing in mind that I changed the name of the partial methods to InterceptBuild and InterceptBuildAsync), or I could probably do so myself more speedily - it depends on whether you want to use this as an opportunity to dig into the code more.

skuruppu:
Thanks @jskeet, I'd like to give it a try myself. Having said that, I won't get around to it for about two weeks given all the rest of the work going on. Hope that's ok.

jskeet:
That's fine by me, if we're okay in terms of timelines for the feature.

PR #4211, opened 2020-01-31
DLQ support in subscriber client

chrisdunelm:


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PubsubExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Delivery attempt counter is 1 + (the sum of number of NACKs and number of
+        /// ack_deadline exceeds) for this message.
+        /// </remarks>
+        /// <param name="msg"></param>
+        /// <returns>The delivery attempt; or <c>null</c> if the subscription does not
+        /// have dead-letter queues enabled.</returns>
+        public static int? GetDeliveryAttempt(this PubsubMessage msg) =>
+            msg.Attributes.TryGetValue(SubscriberClientImpl.DeliveryAttemptAttrKey, out var value) ? (int?)int.Parse(value) : null;
```
jskeet: Are we okay with this throwing if the value exists, but isn't an integer? I don't know whether specifying the invariant culture would make any difference or not for integers, but it might be worth doing to remove any doubt.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1099,6 +1101,10 @@ private async Task ProcessPullMessagesAsync(List<ReceivedMessage> msgs, HashSet<
                             _softStopCts.Token.ThrowIfCancellationRequested();
                             _userHandlerInFlight += 1;
                         }
+                        if (msg.DeliveryAttempt > 0)
+                        {
+                            msg.Message.Attributes.Add(DeliveryAttemptAttrKey, msg.DeliveryAttempt.ToString());
```
jskeet: Again, consider specifying the invariant culture.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1099,6 +1101,10 @@ private async Task ProcessPullMessagesAsync(List<ReceivedMessage> msgs, HashSet<
                             _softStopCts.Token.ThrowIfCancellationRequested();
                             _userHandlerInFlight += 1;
                         }
+                        if (msg.DeliveryAttempt > 0)
```
jskeet: Okay, so this is part of a ReceivedMessage in the proto, and we're putting it into the PubsubMessage, which will be reused.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
jskeet: Should you be using the indexer rather than Add? Won't this throw if the PubsubMessage already has the attribute?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PubsubExtensions.cs
chrisdunelm: Done

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
chrisdunelm: ack

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
chrisdunelm: done

chrisdunelm: done

chrisdunelm:
@jskeet PTAL, the integration test is now done; and it passes for me locally.

jskeet:
Just a ping to remind you that I've approved this, as I think I did so after you'd finished work on Friday. (If there's some other reason for not merging, that's fine of course.)

PR #4230, opened 2020-02-07
Generate and release Google.Cloud.ManagedIdentities.V1

jskeet:
Note that while we don't normally release on a Friday, I'm not expecting high initial take-up, and we can release early in the day - so there's not much risk.


jskeet:
Grr... accidentally closed without merging, due to a rebase conflict. Will create a new PR and self-merge.

PR #4264, opened 2020-02-13
Adjust retry in API library code to use new retry mechanisms

jskeet:


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
```diff
@@ -105,6 +105,8 @@ async Task<TResult> CommitAttempt()
 
         internal async Task<TResult> ExecuteWithRetryAsync<TResult>(Func<Task<TResult>> fn, CancellationToken cancellationToken)
         {
+            // Note: the way that we sometimes use the recommended retry delay, and base
```
amanda-tarafa: What about a `Func<Exception, TimeSpan?> RecommendedBackoff` in `RetrySettings`.

In `RetryAttempt` what is now `JitteredBackoff` can become `DefaultJitteredBackoff` and we can add `RecommendedBackoff` (that is set in `ShouldRetry` using the function). The Backoff methods would use `RecommendedBackoff ?? DefaultJitteredBackoff`.

When building the sequence , after the `yield return` we use `RetryAttempt.RecommendeBackoff ?? backoff` to calculate the new backoff. Since we are past the `yield return`, `RetryAttempt.ShouldRetry` has already been executed and `RetryAttempt.RecommendedBackoff` has been set (if any).

I get this might be a little bit more complicated than what we may want right now.
And also, happy to do this later in another PR.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
```diff
@@ -203,8 +204,9 @@ internal class RetryState
             private readonly IScheduler _scheduler;
             private readonly RetrySettings _retrySettings;
             private readonly int _maxConsecutiveErrors;
-            private TimeSpan _nextBackoff;
-            // TODO: Possibly use RetrySettings.MaxAttempts instead of having this separately?
+            private IEnumerator<TimeSpan> _retrySettingsBackoffs;
+            // Note that we can't use RetrySettings.MaxAttempts, as our sequence of backoffs is not advanced
```
amanda-tarafa: I think my proposal above would solve this one as well if it's OK for next backoff to be calculated from the recommended delay, which I think it's OK because the server is probably sending the recommended delay always so we'll really never use the calculated backoffs.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
jskeet: I think that's a bit complicated for what's effectively a one-off situation. I'm a big fan of only putting things into common libraries when we see the same pattern several times.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
jskeet: Yes, it would probably solve this one too. If we see the same thing in *other* APIs, I'd be happy to add it to GAX at that point.

PR #4267, opened 2020-02-13
Fix Bigtable to use GrpcChannelOptions instead of the ChannelOption from Grpc.Core

jskeet:


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -210,7 +197,7 @@ protected override CallInvoker CreateCallInvoker()
                 return CallInvoker;
             }
             var endpoint = Endpoint ?? GetDefaultEndpoint();
-            var channelOptions = Settings.CreateChannelOptions().Concat(GetChannelOptions()).ToList();
+            var channelOptions = Settings.CreateChannelOptions().MergedWith(GetChannelOptions());
```
amanda-tarafa: Not sure if this will make a difference but if I remember correctly MergeWith ovewrote the former with that latter if an option was on both? Just flagging

jskeet: Yup, it probably makes sense to do this the other way round. Will fix that now. (I doubt that it'll ever be an issue, but...)

PR #4296, opened 2020-02-17
Next build steps

jskeet:
With a local copy of the potential GAX 3.0.0-beta01, this all builds and tests pass as far as Firestore - which I'll look at fixing next.

(I won't merge these changes until I've actually released GAX 3.0.0-beta01 of course.)

Best reviewed one commit at a time.


jskeet:
Firestore is now fixed in an extra commit. Silly mistake.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Version>3.0.0-alpha00</Version>
+    <Version>3.0.0-beta01</Version>
```
chrisdunelm: I'm unclear why `Version` here has moved from `alpha00` to `beta01`, whereas others (e.g. `Google.Cloud.Debugger.V2`) are still on alpha.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/WatchStream.cs
```diff
@@ -206,10 +206,15 @@ async Task CloseStreamAsync()
...
             // Create a new enumerator for the retry attempt sequence, starting with a backoff of zero.
-            IEnumerator<RetryAttempt> CreateRetryAttemptSequence() =>
-                RetryAttempt
+            IEnumerator<RetryAttempt> CreateRetryAttemptSequence()
+            {
+                var iterator = RetryAttempt
                     .CreateRetrySequence(_backoffSettings, _scheduler, initialBackoffOverride: TimeSpan.Zero)
                     .GetEnumerator();
+                // Make sure Current is already valid
+                iterator.MoveNext();
```
chrisdunelm: I'm a bit surprised this is required. What's consuming this that assumes the enumerator has already had `.MoveNext()` called?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.csproj
jskeet: Whoops - that one was a mistake. That's because I did a find and replace of 3.0.0-alpha00 to 3.0.0-beta01 for GAX, and accidentally caught this too. Will fix in an extra commit.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/WatchStream.cs
jskeet: All of the rest of the code :) We do a MoveNext *after* retrying. I could move that instead, but I think I'll keep it here at least for now.

jskeet:
Merging now, even though CI is still running. I expect the change detection to fail on Travis, as the old version expects a locally-built GAX and the new doesn't. It'll be easiest to sort everything out once this is done.

PR #4318, opened 2020-02-17
[Do Not Merge]: Adds channel/call credential clash tests.

amanda-tarafa:
tl;dr We don't respect call credentials in gRPC.

- I've added integration tests in Spanner.V1 (any other gRPC library would have been the same, I just know Spanner) to show credential clashing.
- I tried to add these same tests to GAX gRPC but I cannot easily create a test gRPC service that uses Google credentials to authenticate/authorize.
- I have two service accounts, one who is project owner (`_withSpannerAccess`) and another one that is only a Stackdriver log reader (`_withoutSpannerAccess`).
- The first two tests are for sanity checking that the accounts are as described. Only channel credentials are used and the results are as expected when using each of the accounts.
- The second two tests are for checking credential clashing, they fail in a way that indicates that only the channel credential is being used.
  - GoodChannelCred_BadCallCred shouldn't get a call through, but it does.
  - BadChannelCred_GoodCallCred should get a call through, but it doesn't.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/V1/CredentialsTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/V1/CredentialsTests.cs
@@ -19,6 +19,7 @@ using Grpc.Auth;
 using Grpc.Core;
 using System;
 using System.Linq;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace Google.Cloud.Spanner.V1.IntegrationTests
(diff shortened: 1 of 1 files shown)


jskeet:
Yikes. Good catch. This is on the "old" master branch, so using GAX v3, yes? We should definitely look at this closely.

Have you seen the same thing using calls other than List? I can imagine there being some subtleties due to List operations being deferred... it's possible we're not respecting CallSettings at all there, but that other calls aren't broken.

amanda-tarafa:
I'll try later with other call types and see what I get. I'm already looking at Gax to see where the call credentials might be falling through the cracks.

amanda-tarafa:
Also the Travis failure I know what it is, it's unrelated. I'll send a PR later so don't worry about it.

jskeet:
Thought I'd added a comment on this, but apparently not - perhaps you could add "Skip" to these tests, rebase to just a single commit, then we can merge?

amanda-tarafa:
Yep, that sounds good, will skip only the ones that are not working as they should.

amanda-tarafa:
Temporarily closing until I've set up some restrited access credentials. Will keep the branch until then though.

PR #4363, opened 2020-02-19
Add a small script to check which tags have NuGet packages

jskeet:
(This is useful when performing mass releases, as it's quite easy to miss a release if you're not careful.)

As a side note, we seem to be accumulating scripts... maybe "occasional" ones like this should go in a separate directory.


jskeet:
Yup, that's right.

PR #4381, opened 2020-02-19
Remove Google.Cloud.Diagnostics.AspNet source code

jskeet:
The package remains listed in nuget.org, and it makes sense to keep it in the README; it's perfectly usable, we're just not taking it any further.

If we get requests to build a version of it targeting GAX v3, we can revert this commit at that point.


amanda-tarafa:
Travis is failing thouhg, I think maybe because the folder is not deleted?

jskeet:
Travis is failing because it *is* deleted. It tries to build "things that have changed" - but in this case, the change is a deletion. This always happens when we delete a project, but it's never worth fixing!

amanda-tarafa:
Ahhh, makes sense!

PR #4385, opened 2020-02-19
Fix log4net tests by limiting list requests to recent log entries

jskeet:


chrisdunelm:
This appears to be a change to NLog, not log4net?

jskeet:
That's very strange. My branches must have got mixed up somehow. Will fix.

PR #4388, opened 2020-02-19
Introduce a "release manager" to make our lives easier

jskeet:
(./prepare-release.sh is now just a simple way of invoking the release manager.)

The idea is to simplify the various steps involved in releasing,
without taking away any flexibility. ReleaseManager can do the
following things:

- Set the version of an API by editing apis.json and regenerating projects
- Show you which APIs have changed versions in the working directory
- Show you the changes (compatibility etc) for all changed APIs
- Update the version history for all changed APIs
- If you've only changed a single API (common case) create the git commit for it

So a common flow would be something like this:

- git checkout -b release-storage
- ./prepare-release.sh set-version Google.Cloud.Storage.V1 3.0.0-beta01
- ./prepare-release.sh show *(just to check we did the right thing)*
- ./prepare-release.sh update-history
- ./prepare-releases.sh compare
- Manually fix up the history file a bit (path is shown in update-history)
- ./prepare-releases.sh commit
- git push, after amending the commit if necessary
- Then create a PR etc as normal - this tooling doesn't affect that flow

Note that we've only had to type in the API ID and version once in that whole flow :)

Limitations:

- The tool doesn't update README.md or docs/root/index.md
- New APIs still require apis.json to be edited manually

### DIFF
--- a/prepare-release.sh
+++ b/prepare-release.sh
@@ -2,23 +2,4 @@
 
 set -e
 
-if [ -z "$1" ]
-then
-  echo "Please specify the API you're releasing, e.g. Google.Cloud.Storage.V1"
-  exit 1
-fi
-
-# Make sure we have all the tags locally.
-git fetch --tags -f -q upstream master
-
-echo "Building $1"
-./build.sh $1 --notests
-echo ""
-
-echo "Updating release notes at:"
-echo "apis/$1/docs/history.md"
-dotnet run -p tools/Google.Cloud.Tools.UpdateReleaseNotes -- $1
-echo ""
-
-echo "Checking version compatibility"
-dotnet run -p tools/Google.Cloud.Tools.CheckVersionCompatibility -- $1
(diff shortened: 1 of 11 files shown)


jskeet:
Note: the code itself isn't as important as the flow here. It feels right to me (having released lots of libraries very recently) but I may be missing something.
Ideally I'd like to get this merged fairly soon, then experiment as we go... if it breaks, we just have to do stuff manually.

jskeet:
Sample commit as generated by the tool, btw:

```text
Release Google.Cloud.Storage.V1 version 3.0.0-beta01

Changes in this release

- [Commit ee62bbc](https://github.com/googleapis/google-cloud-dotnet/commit/ee62bbc): Manual code changes for updating to GAX v3
```

jskeet:
I believe it *should* work for first releases (and will create the history file automatically), although we'll need to add the entry to apis.json manually.

You're right that it doesn't update the README.md or docs/root/index.md automatically. Will add that as a limitation. I'd like to get all of that fully automated separately, in fact.

PR #4393, opened 2020-02-20
Display the version change in set-version command

jskeet:


jskeet:
Hmm.. hold off reviewing for a minute. I don't quite understand how release note formatting works yet. Sometimes it's being formatted as Markdown, and sometimes not.

jskeet:
Ah, I see the problem. It's https://github.community/t5/How-to-use-Git-and-GitHub/Link-markdown-not-working-in-tag-release-notes/td-p/12517.
We should put the unwrapping code into TagReleases, and keep the commit wrapped, I think. Will remove the second commit.

jskeet:
(This PR is ready to review, and is now trivial :)

PR #4394, opened 2020-02-20
Explicitly set a body when creating a release

jskeet:
This body is constructed from the most recent commit, but with the
lines unwrapped - GitHub doesn't follow regular Markdown (ignoring
line breaks) when displaying releases.

We detect blank lines, lines that are list items etc and break
accordingly. As an example, this list:

- First
- Second
  - Subitem
* Third

... should still render as a list.


jskeet:
It's going to be hard to actually test this before we next need to do a release. But that might be very soon :)

PR #4411, opened 2020-02-24
[Spanner] Improvements to nursing pool back to health.

amanda-tarafa:
This is way cleaner.
All tests are green on local.


jskeet:
We now have tests in Google.Apis and it seems to be fine. I guess the only difference is that this won't log errors in the same way that it used to with ConsumeBackgroundTask.

amanda-tarafa:
Yes, although we were doing everyting in a try catch in the background task so I think we wouldn't have logged any exceptions anyway, maybe for a very edge case of something happening whe starting the task or similar?
Are you OK with me mergin then? I'm fine either way, the only thing that's really needed is the using statement, and I can add that to the existing code in another PR.

jskeet:
Yes, I'm fine for you to merge :)

PR #4413, opened 2020-02-25
Use the reference assemblies package to build cross-platform

jskeet:


jskeet:
Chris: would you like me to create a PR to make the projects generated by the microgenerator do this too?

PR #4451, opened 2020-03-04
Migrate projects from .NET Core 2.2 to 3.1

jskeet:
These aren't test projects, so we might as well move them to 3.1 rather than down to 2.1.


jskeet:
(I believe this will fix CI for your Storage change, Amanda.)

jskeet:
(First commit wouldn't have helped - second one should!)

PR #4510, opened 2020-03-04
[Storage] V4 fixes and new features.

### LINKED ISSUE #4190: [Storage] Fix V4 Signature (Formatting Inconsistencies)
Tracking issue for work defined in companion document.

[New Client Request - GCS SigV4 Formatting Inconsistencies](https://docs.google.com/document/d/11OpYYj2FMFsWe-slIKK-JQC97CUh4CYOWeraR0wuImo).

amanda-tarafa:
Fixes #4190, #4191, #4192, #4193.

Each commit is either a fix or one of the new features.

@frankyn For virtual hosted style and scheme, it's not clear to me if V2 should support it (or if we should invest the time on it), we don't currently do.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerTest.OptionsTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerTest.OptionsTest.cs
@@ -31,6 +31,9 @@ namespace Google.Cloud.Storage.V1.Tests
                 Assert.Null(options.Expiration);
                 Assert.Equal(TimeSpan.FromMinutes(1), options.Duration);
                 Assert.Equal(SigningVersion.Default, options.SigningVersion);
+                Assert.Equal(UrlStyle.PathStyle, options.UrlStyle);
+                Assert.Equal("https", options.Scheme);
+                Assert.Null(options.BucketBoundHostname);
             }
 
             [Fact]
@@ -43,6 +46,9 @@ namespace Google.Cloud.Storage.V1.Tests
                 Assert.Null(options.Duration);
                 Assert.Equal(now, options.Expiration);
                 Assert.Equal(SigningVersion.Default, options.SigningVersion);
+                Assert.Equal(UrlStyle.PathStyle, options.UrlStyle);
+                Assert.Equal("https", options.Scheme);
+                Assert.Null(options.BucketBoundHostname);
             }
 
             [Fact]
@@ -107,6 +113,71 @@ namespace Google.Cloud.Storage.V1.Tests
 
                 Assert.NotSame(options, newVersion);
                 Assert.Equal(SigningVersion.V4, newVersion.SigningVersion);
+            }
+
+            [Fact]
+            public void WithBucketBoundHost()
+            {
+                var options = Options.FromDuration(TimeSpan.FromMinutes(1));
+
+                var newUrlStyle = options.WithBucketBoundHostname("my.bucket.bound.domain");
+
+                Assert.NotSame(options, newUrlStyle);
+                Assert.Equal(UrlStyle.BucketBoundHostname, newUrlStyle.UrlStyle);
+                Assert.Equal("my.bucket.bound.domain", newUrlStyle.BucketBoundHostname);
+            }
+
+            [Fact]
+            public void WithUrlStyle_Valid()
+            {
+                var options = Options.FromDuration(TimeSpan.FromMinutes(1));
+
+                var newUrlStyle = options.WithUrlStyle(UrlStyle.VirtualHostedStyle);
+
+                Assert.NotSame(options, newUrlStyle);
+                Assert.Equal(UrlStyle.VirtualHostedStyle, newUrlStyle.UrlStyle);
+                Assert.Null(newUrlStyle.BucketBoundHostname);
+            }
+
+            [Fact]
+            public void WithUrlStyle_BucketBoundHostname()
+            {
+                var options = Options.FromDuration(TimeSpan.FromMinutes(1));
+
+                Assert.Throws<ArgumentException>(() => options.WithUrlStyle(UrlStyle.BucketBoundHostname));
+            }
+
+            [Fact]
+            public void WithUrlStyle_FromBucketBoundHostnameToOther()
+            {
+                var options = Options
+                    .FromDuration(TimeSpan.FromMinutes(1))
+                    .WithBucketBoundHostname("my.bucket.bound.domain");
+
+                var newUrlStyle = options.WithUrlStyle(UrlStyle.VirtualHostedStyle);
+
+                Assert.NotSame(options, newUrlStyle);
+                Assert.Equal(UrlStyle.VirtualHostedStyle, newUrlStyle.UrlStyle);
+                Assert.Null(newUrlStyle.BucketBoundHostname);
+            }
+
+            [Fact]
+            public void WithScheme()
+            {
+                var options = Options.FromDuration(TimeSpan.FromMinutes(1));
+
+                var newScheme = options.WithScheme("http");
+
+                Assert.NotSame(options, newScheme);
+                Assert.Equal("http", newScheme.Scheme);
+            }
+
+            [Fact]
+            public void WithScheme_InvalidThrows()
+            {
+                var options = Options.FromDuration(TimeSpan.FromMinutes(1));
+
+                Assert.Throws<ArgumentException>(() => options.WithScheme("ftp"));
             }
         }
     }
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerTest.RequestTemplateTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerTest.RequestTemplateTest.cs
@@ -44,6 +44,7 @@ namespace Google.Cloud.Storage.V1.Tests
                 Assert.Null(requestTemplate.ObjectName);
                 Assert.Empty(requestTemplate.RequestHeaders);
                 Assert.Empty(requestTemplate.ContentHeaders);
+                Assert.Empty(requestTemplate.QueryParameters);
                 Assert.Equal(HttpMethod.Get, requestTemplate.HttpMethod);
             }
 
@@ -93,7 +94,7 @@ namespace Google.Cloud.Storage.V1.Tests
 
                 var newRequestTemplate = requestTemplate.WithRequestHeaders(headers);
 
-                var expectedHeaders = ToExpectedHeaders(headers);
+                var expectedHeaders = ToExpectedEntries(headers);
                 Assert.NotSame(requestTemplate, newRequestTemplate);
                 Assert.Equal(expectedHeaders, newRequestTemplate.RequestHeaders);
             }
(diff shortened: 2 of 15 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerTest.RequestTemplateTest.cs
```diff
@@ -110,12 +111,29 @@ public void WithContentHeaders()
...
 
-            private IReadOnlyDictionary<string, IReadOnlyCollection<string>> ToExpectedHeaders(
+            [Fact]
+            public void WithQueryParameters()
+            {
+                var requestTemplate = RequestTemplate.FromBucket("bucket-name");
+
+                var queryParameters = new Dictionary<string, IEnumerable<string>>
+                {
+                    { "query1", new List<string> { "value1", "value2"} },
```
jskeet: Nit: space after "value2"

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerTest.RequestTemplateTest.cs
```diff
@@ -110,12 +111,29 @@ public void WithContentHeaders()
...
+                    { "query2", new List<string> {"value3"} }
```
jskeet: Nit: spaces around "value3"

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.RequestTemplate.cs
```diff
@@ -205,7 +216,27 @@ public sealed class RequestTemplate
...
+
+            /// <summary>
+            /// Returns a new request template with all the same settings as this one
+            /// except for the query parameters.
+            /// </summary>
+            /// <param name="queryParameters">The new query parameters. May be null, in which case
+            /// an empty collection will be used.</param>
+            /// <returns>A new request template.</returns>
+            public RequestTemplate WithQueryParameters(IEnumerable<KeyValuePair<string, IEnumerable<string>>> queryParameters) =>
+                new RequestTemplate(Bucket, ObjectName, HttpMethod, RequestHeaders, ContentHeaders, queryParameters.AsReadOnly());
```
jskeet: Calling `AsReadOnly()` doesn't make this immutable - nor does the IReadOnlyDictionary-based overload above.
If we want to be able to rely on this not changing under our feet, we should clone it.

That's possibly overkill, especially as these will normally be short-lived objects - it's just worth being aware of.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.Options.cs
```diff
@@ -47,7 +47,19 @@ public sealed class Options
...
+            /// The style of signed URL to generate.
+            /// Defaults to <see cref="UrlStyle.Path"/>.
+            /// </summary>
+            public UrlStyle UrlStyle { get; }
+
+            /// <summary>
+            /// The Scheme to use for the request. Usually http or https.
+            /// Defaults to https.
+            /// </summary>
+            public string Scheme { get; }
```
jskeet: This feels like it might be better in the template. As discussed, the dividing line is pretty fuzzy.
I'm happy to leave it here if you're reasonably happy that it belongs here - it's just worth you considering it again, in case. (You're closer to the code here, and probably have a better sense of this than I do.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.Options.cs
```diff
@@ -94,23 +108,40 @@ private Options(TimeSpan? duration, DateTimeOffset? expiration, SigningVersion v
...
-                new Options(Duration, Expiration, GaxPreconditions.CheckEnumValue(version, nameof(version)));
+                new Options(Duration, Expiration, GaxPreconditions.CheckEnumValue(version, nameof(version)), UrlStyle, Scheme);
+
+            /// <summary>
+            /// Returns a new set of options with the same values as this one except for the
+            /// <see cref="UrlStyle"/> value.
+            /// </summary>
+            /// <param name="urlStyle">The new url style.</param>
+            /// <returns>A new set ofoptions with the given url style.</returns>
+            public Options WithUrlStyle(UrlStyle urlStyle) =>
```
jskeet: I think we should validate that the style is a known enum value here.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
```diff
@@ -68,10 +68,15 @@ private struct SigningState
                 private string _resourcePath;
                 private string _canonicalQueryString;
                 internal byte[] _blobToSign;
+                private string _scheme;
+                private string _host;
 
                 internal SigningState(RequestTemplate template, Options options, IBlobSigner blobSigner, IClock clock)
                 {
                     options = options.ToExpiration(clock);
+                    _host = $"{(urlStyle == UrlStyle.VirtualHosted ? $"{bucket}." : string.Empty)}{StorageHost}";
```
jskeet: I think this would actually be simpler without the interpolated string literal:

```csharp
_host = urlStyle == UrlStyle.VirtualHosted ? $"{bucket}.{StorageHost}" : StorageHost;
```

It duplicates the StorageHost part, but avoids nested interpolated string literals and just-for-the-parser parentheses.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.Options.cs
```diff
@@ -94,23 +108,40 @@ private Options(TimeSpan? duration, DateTimeOffset? expiration, SigningVersion v
...
+                new Options(Duration, Expiration, SigningVersion, urlStyle, Scheme);
+
+            /// <summary>
+            /// Returns a new set of options with the same values as this one except for the scheme.
+            /// </summary>
+            /// <param name="scheme">The new scheme.</param>
+            /// <returns>A new set of options with the given scheme.</returns>
+            public Options WithScheme(string scheme) =>
```
jskeet: Is there anything other than http or https that would work? If not, should we validate that? Also document the null behavior.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/UrlSignerTest.OptionsTest.cs
```diff
@@ -114,14 +116,49 @@ public void WithSigningVersion()
             }
 
             [Fact]
-            public void WithUrlStyle()
+            public void WithBucketStyleDomain()
```
jskeet: WithBucketStyleDomain => WithBucketBoundDomain?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
```diff
@@ -73,10 +73,24 @@ private struct SigningState
 
                 internal SigningState(RequestTemplate template, Options options, IBlobSigner blobSigner, IClock clock)
                 {
+                    _resourcePath = string.Empty;
+                    switch (options.UrlStyle)
```
jskeet: It's a pity that a switch expression doesn't really help here, given that you could now use it!

amanda-tarafa: Oh yes it does help, switch expressions work well with unnamed tuples. Take a look.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.Options.cs
jskeet: Fine by me.

frankyn:
One quick comment as I missed  your question @amanda-tarafa,

> @frankyn For virtual hosted style and scheme, it's not clear to me if V2 should support it (or if we should invest the time on it), we don't currently do.

Scheme and virutal hosted style both work with V2 as well. I've asked other maintainers to support these in v2 as well to reduce gotchas between versions for end-users.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.Options.cs
frankyn: Pending discussion from @houglum, but I'm fairly certain these are the only two values for scheme.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.Options.cs
```diff
@@ -47,18 +47,73 @@ public sealed class Options
             /// </summary>
             public SigningVersion SigningVersion { get; }
 
-            private Options(TimeSpan? duration, DateTimeOffset? expiration, SigningVersion version)
```
frankyn: For SigningVersion, given the rework of this implementation, could you update to default to V4?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.RequestTemplate.cs
```diff
@@ -115,18 +115,34 @@ public sealed class RequestTemplate
...
+            /// The query parameters which will be included with the request. This will never be null.
+            /// If null is specified in <see cref="WithQueryParameters(IEnumerable{KeyValuePair{string, IEnumerable{string}}})"/>
+            /// then an empty dictionary will be used.
+            /// </summary>
+            /// <remarks>
+            /// When using <see cref="SigningVersion.V2"/> query parameters are not included in the signature
+            /// and <see cref="UrlSigner.Sign(RequestTemplate, Options)"/> and
+            /// <see cref="UrlSigner.SignAsync(RequestTemplate, Options, System.Threading.CancellationToken)"/>
+            /// will throw <see cref="ArgumentException"/> if <see cref="RequestTemplate.QueryParameters"/> is not
+            /// an empty collection.
```
frankyn: This is a little confusing to me because `The query parameters which will be included with the request. This will never be null.` where as this states: `When using <see cref="SigningVersion.V2"/> query parameters are not included in the signature... will throw <see cref="ArgumentException"/> if <see cref="RequestTemplate.QueryParameters"/> is not an empty collection`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.Options.cs
amanda-tarafa: Done

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.RequestTemplate.cs
amanda-tarafa: But a null reference is not the same as a reference to an empty collection.

- `RequestTemplate.QueryParameters` will never be null because even if a null is passed to `WithQueryParameters`, an empty dictionary will be set. This is basically to make calling code cleaner by not having to deal with the special case "null".
- When using V2, if custom query parameters are specified, that is, if `RequestTemplate.QueryParameters` is referencing other than an empty collection (it will never be a null reference because of the previous point) then the signing method will throw.

I've added a short phrase at the end of the comment to try and help clarify, or let me know exactly what is confusing for you and what wording would help make it clearer.

PR #4512, opened 2020-03-05
feat: [Spanner] Add query options support

skuruppu:
Adds the ability to set `QueryOptions` when running Cloud Spanner queries.
For now, only setting the `query_optimizer_version` is added.

`QueryOptions` can be configured through the following mechanisms.
1. At the `SpannerConnection` level.
2. Through the `SPANNER_OPTIMIZER_VERSION` environment variable.
3. At a query level.

1. Options set at an environment variable level will override options
configured at the `SpannerConnection` level.
2. Options set at a query-level will override options set at either the
`SpannerConnection` or environment variable level.

If no options are set, the optimizer version will default to:
1. The optimizer version the database is pinned to.
2. If the database is not pinned to a specific version, then the Cloud Spanner
backend will use the "latest" version.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
@@ -33,17 +33,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         public QueryOptionsTests(ReadTableFixture fixture) =>
             _fixture = fixture;
 
-        private async Task<T> ExecuteAsync<T>(string sql)
-        {
-            using (var connection = _fixture.GetConnection())
-            {
-                var cmd = connection.CreateSelectCommand(sql);
-                var result = await cmd.ExecuteScalarAsync<T>();
-                return result;
-            }
-        }
-
-        // [START spanner_test_single_key_read_with_query_options]
+        // [START spanner_test_single_key_read_with_conn_level_query_options]
         [Fact]
         public async Task PointReadWithConnectionLevelQueryOptions()
         {
@@ -61,7 +51,27 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
                 }
             }
         }
-        // [END spanner_test_single_key_read_with_query_options]
+        // [END spanner_test_single_key_read_with_conn_level_query_options]
+
+        // [START spanner_test_single_key_read_with_cmd_level_query_options]
+        [Fact]
+        public async Task PointReadWithQueryLevelOptions()
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                var cmd = connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName} WHERE Key = 'k1'");
+                cmd.QueryOptions = new QueryOptions().WithOptimizerVersion("1");
+                using (var reader = await cmd.ExecuteReaderAsync())
+                {
+                    Assert.True(await reader.ReadAsync());
+                    Assert.Equal("k1", reader.GetString(0));
+                    Assert.Equal("v1", reader.GetString(1));
+
+                    Assert.False(await reader.ReadAsync());
+                }
+            }
+        }
+        // [END spanner_test_single_key_read_with_cmd_level_query_options]
 
         // [START spanner_test_single_key_read_with_invalid_query_options]
         [Fact]
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
@@ -1,0 +1,84 @@
+﻿// Copyright 2020 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Apis.Auth.OAuth2;
+using Google.Cloud.Spanner.Data.CommonTesting;
+using Google.Cloud.Spanner.V1;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.IntegrationTests
+{
+    [Collection(nameof(ReadTableFixture))]
+    [CommonTestDiagnostics]
+    public class QueryOptionsTests
+    {
+        private readonly ReadTableFixture _fixture;
+
+        public QueryOptionsTests(ReadTableFixture fixture) =>
+            _fixture = fixture;
+
+        private async Task<T> ExecuteAsync<T>(string sql)
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                var cmd = connection.CreateSelectCommand(sql);
+                var result = await cmd.ExecuteScalarAsync<T>();
+                return result;
+            }
+        }
+
+        // [START spanner_test_single_key_read_with_query_options]
+        [Fact]
+        public async Task PointReadWithConnectionLevelQueryOptions()
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                connection.QueryOptions = new QueryOptions().WithOptimizerVersion("1");
+                var cmd = connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName} WHERE Key = 'k1'");
+                using (var reader = await cmd.ExecuteReaderAsync())
+                {
+                    Assert.True(await reader.ReadAsync());
+                    Assert.Equal("k1", reader.GetString(0));
+                    Assert.Equal("v1", reader.GetString(1));
+
+                    Assert.False(await reader.ReadAsync());
+                }
+            }
+        }
+        // [END spanner_test_single_key_read_with_query_options]
+
+        // [START spanner_test_single_key_read_with_invalid_query_options]
+        [Fact]
+        public async Task PointReadWithInvalidConnectionLevelQueryOptions()
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                connection.QueryOptions = new QueryOptions().WithOptimizerVersion("invalid");
+                var cmd = connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName} WHERE Key = 'k1'");
+                using (var reader = await cmd.ExecuteReaderAsync())
+                {
+                    var e = await Assert.ThrowsAsync<SpannerException>(() => reader.ReadAsync());
+                    Assert.Equal(ErrorCode.InvalidArgument, e.ErrorCode);
+                    Assert.False(e.IsTransientSpannerFault());
+                }
+            }
+        }
+        // [END spanner_test_single_key_read_with_invalid_query_options]
+    }
+}
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
@@ -39,7 +39,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         {
             using (var connection = _fixture.GetConnection())
             {
-                connection.QueryOptions = new QueryOptions().WithOptimizerVersion("1");
+                connection.QueryOptions = QueryOptions.Empty.WithOptimizerVersion("1");
                 var cmd = connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName} WHERE Key = 'k1'");
                 using (var reader = await cmd.ExecuteReaderAsync())
                 {
@@ -60,7 +60,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
             using (var connection = _fixture.GetConnection())
             {
                 var cmd = connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName} WHERE Key = 'k1'");
-                cmd.QueryOptions = new QueryOptions().WithOptimizerVersion("1");
+                cmd.QueryOptions = QueryOptions.Empty.WithOptimizerVersion("1");
                 using (var reader = await cmd.ExecuteReaderAsync())
                 {
                     Assert.True(await reader.ReadAsync());
@@ -79,7 +79,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         {
             using (var connection = _fixture.GetConnection())
             {
-                connection.QueryOptions = new QueryOptions().WithOptimizerVersion("invalid");
+                connection.QueryOptions = QueryOptions.Empty.WithOptimizerVersion("invalid");
                 var cmd = connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName} WHERE Key = 'k1'");
                 using (var reader = await cmd.ExecuteReaderAsync())
                 {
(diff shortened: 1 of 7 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -219,6 +220,34 @@ public override UpdateRowSource UpdatedRowSource
...
+
+            _queryOptions.OptimizerVersion = options.OptimizerVersion;
+        }
+
+        /// <summary>
+        /// Query options to use when running SQL and streaming SQL commands.
+        /// When a new set of options is set, the new fields will be merged
+        /// into the fields in the existing options.
+        /// </summary>
+        public QueryOptions QueryOptions
```
jskeet: Right, this gets back to the earlier GetEffectiveOptions() discussion.

It feels a little odd that if I set the QueryOptions property and then fetch it again, I might not get the same result. The behavior of this code would be odd too:

```csharp
var command = new SqlCommand();
var options1 = new QueryOptions { OptimizerVersion = "foo" };
var options2 = new QueryOptions { OptimizerVersion = "bar" };

command.QueryOptions = options1;
var returnedOptions = command.QueryOptions;
Console.WriteLine(returnedOptions.OptimizerVersion); // foo

command.QueryOptions = options2;
Console.WriteLine(returnedOptions.OptimizerVersion); // bar
```

The mutation of an existing object feels odd.

It also feels odd that once I've set query options with an optimizer version, there's nothing I can do to clear it - setting QueryOptions to null doesn't clear the property itself, and setting it to a QueryOptions with an empty optimizer version has no effect.

I think it would be simpler to understand if each place where a QueryOptions can be specified had a simple automatically-implemented property with no logic, and we put the merging logic into the place where the request is constructed. We'd need to work out exactly what that merging logic *is*, admittedly - if the connection specifies a query optimizer version, can the command effectively remove that by specifying a non-null QueryOptions property, but with an empty query optimizer version? That would seem reasonably sensible to me.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -85,5 +91,100 @@ public void CloneUpdateWithParameters()
             Assert.True(command.Parameters.SequenceEqual(command2.Parameters));
         }
 
+        /*
```
skuruppu: Tests are commented out for now since I ran out of time to setup the mock. Will complete this next week.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
skuruppu: I hope the new implementation is consistent with your idea.

> We'd need to work out exactly what that merging logic is, admittedly - if the connection specifies a query optimizer version, can the command effectively remove that by specifying a non-null QueryOptions property, but with an empty query optimizer version?

I've asked the backend team on what the expected behavior is. But from the implementations for the other languages, we decided to go with the approach of only merge if a value is set. If a field/env var is null/empty then we assume that it was not set. To "unset" a value set at the connection level, you can explicitly set the default value. E.g. OptimizerVersion = "latest". This is what the backend assumes if no value is set.

jskeet:
Thanks @skuruppu - I'm at a conference today, but will look at this first thing Monday (assuming I don't get time to look today). @amanda-tarafa can also look in the meantime, of course :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -337,6 +339,43 @@ private List<Mutation> GetMutations()
...
+                {
+                    optimizerVersion = Connection.QueryOptions.OptimizerVersion;
+                }
+
+                if (string.IsNullOrEmpty(optimizerVersion))
+                {
+                    return null;
+                }
+
+                var queryOptionsProto = new V1.ExecuteSqlRequest.Types.QueryOptions();
```
jskeet: This can be simplified to use an object initializer:

```csharp
return new V1.ExecuteSqlRequest.Types.QueryOptions { OptimizerVersion = optimizerVersion };
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -349,10 +388,7 @@ private ExecuteSqlRequest GetExecuteSqlRequest()
                     Sql = CommandTextBuilder.ToString()
                 };
 
-                if (QueryOptions != null)
-                {
-                    request.QueryOptions = QueryOptions.ToProto();
-                }
+                request.QueryOptions = GetEffectiveQueryOptions();
```
jskeet: You can put this inside the above object initializer:

```csharp
var request = new ExecuteSqlRequest
{
    Sql = CommandTextBuilder.ToString(),
    QueryOptions = GetEffectiveQueryOptions()
};
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -220,32 +220,13 @@ public override UpdateRowSource UpdatedRowSource
...
         /// <summary>
         /// Query options to use when running SQL and streaming SQL commands.
-        /// When a new set of options is set, the new fields will be merged
-        /// into the fields in the existing options.
         /// </summary>
         public QueryOptions QueryOptions
         {
             get => _queryOptions;
-            set => MergeQueryOptions(value);
+            set => _queryOptions = (QueryOptions) value;
```
jskeet: At this point I don't believe you need the `_queryOptions` field at all. Just use:

```csharp
/// <summary>
/// Query options to use when running SQL and streaming SQL commands.
/// </summary>
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -138,10 +119,7 @@ private void ApplyQueryOptionsFromEnvironment()
         public QueryOptions QueryOptions
         {
             get => _queryOptions;
-            set {
-                _queryOptions = (QueryOptions) value;
-                ApplyQueryOptionsFromEnvironment();
-            }
+            set => _queryOptions = (QueryOptions) value;
```
jskeet: Ditto comment from SpannerCommand.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/QueryOptions.cs
```diff
@@ -84,11 +84,6 @@ public QueryOptions() : this(new V1.ExecuteSqlRequest.Types.QueryOptions())
             return new QueryOptions(proto.Clone());
         }
 
-        /// <summary>
```
jskeet: If you don't really need the proto form, it might be worth just having the OptimizerVersion string property. You can still have a method to construct an instance from a proto if you want.
I don't much mind.

skuruppu: I actually reverted this change because @larkee made me realize that merging protos is much easier than explicitly checking for each field. So leaving the protos for now but let me know if you think there's a better way to do this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
skuruppu: Done.

I actually simplified the whole function and made it a bit more future-proof by using the proto merge functionality. That way when we add new options, the merging will happen automatically. The only thing to update manually would be reading the new env vars.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -85,5 +92,120 @@ public void CloneUpdateWithParameters()
...
+            spannerClientMock
+                .SetupBatchCreateSessionsAsync()
+                .Setup(client => client.ExecuteStreamingSql(
+                    It.Is<ExecuteSqlRequest>(
+                        request => request.QueryOptions.OptimizerVersion == optimizerVersion),
+                    It.IsAny<CallSettings>()))
+                .Returns<ExecuteSqlRequest, CallSettings>((request, _) =>
+                {
+                    // Returning null since we can't return a SpannerClient.ExecuteStreamingSqlStream()
+                    // since it's an abstract class.
```
skuruppu: I couldn't find a type to return an empty result here so had to resort to null.

skuruppu:
I'm a bit uncertain about adding an integration test to this given that the backend won't respond in a noticeable way if a valid version was given. But if an invalid version was given it will let us know.

@jskeet let me know your thoughts on whether this type of IT is typically added in this repo.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
jskeet: You could find the concrete class involved. Personally I generally avoid using mocks, preferring fakes - but I can see how a mock is possibly simpler here.

The lambda expression can be just `(request, _) => null` though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/QueryOptions.cs
```diff
@@ -0,0 +1,108 @@
...
+{
+    /// <summary>
+    /// Immutable class representing query options.
+    /// </summary>
+    public sealed class QueryOptions : IEquatable<QueryOptions>
+    {
+        /// <summary>
+        /// The query optimizer version configured in the options.
+        /// </summary>
+        public string OptimizerVersion
```
jskeet: Nit: simplify to

```csharp
public string OptimizerVersion => Proto.OptimizerVersion;
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/QueryOptions.cs
```diff
@@ -0,0 +1,108 @@
...
+            return new QueryOptions(proto.Clone());
+        }
+
+        /// <summary>
+        /// Get proto version of the query options.
+        /// </summary>
+        public V1.ExecuteSqlRequest.Types.QueryOptions ToProto() => Proto.Clone();
+
+        /// <inheritdoc />
+        public bool Equals(QueryOptions other)
```
jskeet: I'd probably write this as:

```csharp
public bool Equals(QueryOptions other) => other is object && OptimizerVersion == other.OptimizerVersion;
```

(As other.OptimizerVersion can never return null, and OptimizerVersion can never be null, you *could* just return `OptimizerVersion == other?.OptimizerVersion`, but the reasoning required makes that harder to understand.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
```diff
@@ -0,0 +1,84 @@
...
+    [Collection(nameof(ReadTableFixture))]
+    [CommonTestDiagnostics]
+    public class QueryOptionsTests
+    {
+        private readonly ReadTableFixture _fixture;
+
+        public QueryOptionsTests(ReadTableFixture fixture) =>
+            _fixture = fixture;
+
+        private async Task<T> ExecuteAsync<T>(string sql)
```
jskeet: I don't think this is actually used, is it?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                var cmd = connection.CreateSelectCommand(sql);
+                var result = await cmd.ExecuteScalarAsync<T>();
+                return result;
+            }
+        }
+
+        // [START spanner_test_single_key_read_with_query_options]
```
jskeet: It probably wouldn't hurt to add equivalent command-level option tests; I'll leave that as a call for you to make though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -85,5 +92,114 @@ public void CloneUpdateWithParameters()
...
+            connection.QueryOptions = queryOptions;
+
+            var command = connection.CreateSelectCommand("SELECT * FROM FOO");
+            using (var reader = command.ExecuteReader())
+            {
+                // Do nothing.
+            }
+
+            // Set the environment back.
+            Environment.SetEnvironmentVariable(optimizerVersionVariable, savedOptimizerVersion);
```
jskeet: It would be good to do this in a `finally` block. You might even want to write a method accepting an `Action` that will set the environment variable, execute the action, then reset the environment variable.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -85,5 +92,114 @@ public void CloneUpdateWithParameters()
...
+            const string envOptimizerVersion = "2";
+            Environment.SetEnvironmentVariable(optimizerVersionVariable, envOptimizerVersion);
+
+            var cmdOptimizerVersion = "3";
+            // Optimizer version set at a command level has higher precedence
+            // than version set through the connection or the environment
+            // variable.
+            Mock<SpannerClient> spannerClientMock = SetupExecuteStreamingSql(cmdOptimizerVersion);
+
+            const string optimizerVersion = "1";
```
jskeet: Nit: rename to `connOptimizerVersion` to clarify? The other two variables are pleasantly clear :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/QueryOptions.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// overrides the default optimizer version for query execution.</para>
+        /// </remarks>
+        /// <param name="optimizerVersion">Optimizer version to set.</param>
+        public QueryOptions WithOptimizerVersion(string optimizerVersion)
+        {
+           var protoCopy = Proto.Clone();
+           protoCopy.OptimizerVersion = optimizerVersion;
+           return new QueryOptions(protoCopy);
+        }
+
```
jskeet: Very optional nit: remove extra blank line.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/QueryOptions.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// The proto representation of the query options. Must not be mutated
+        /// or exposed publicly.
+        /// </summary>
+        internal V1.ExecuteSqlRequest.Types.QueryOptions Proto { get; }
+
+        private QueryOptions(V1.ExecuteSqlRequest.Types.QueryOptions proto) => Proto = proto;
+
+        /// <summary>
```
jskeet: An optional alternative to this would be to have a static `Empty` property, e.g.

```csharp
public static QueryOptions Empty { get; } = new QueryOptions(new V1.ExecuteSqlRequest.Types.QueryOptions());
```

Then code which is currently `new QueryOptions().WithOptimizerVersion(...)` would become `QueryOptions.Empty.WithOptimizerVersion(...)` which *might* be clearer/simpler.

+amanda-tarafa do you have a preference on this one?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -110,6 +110,12 @@ public override ConnectionState State
 
         internal bool IsOpen => (State & ConnectionState.Open) == ConnectionState.Open;
 
+        /// <summary>
```
jskeet: Currently this is only available as a property in the connection. Do you think we want people to be able to provide it in the connection string itself? I don't know what the expected use is likely to be.

We could potentially do that later.

jskeet:
Great, thanks for all of that. Are you now happy for me to squash into a single commit and merge?

skuruppu:
> Great, thanks for all of that. Are you now happy for me to squash into a single commit and merge?

All good, sorry for the late reply. I will squash and merge now.

skuruppu:
@jskeet I have to ask you about cutting a release with these changes. What process should I follow?

jskeet:
@skuruppu: Will email you internally.

PR #4517, opened 2020-03-05
Use new GAX emulator support for Firestore and Datastore

jskeet:


apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -140,29 +85,37 @@ void ApplyEmulatorConfiguration()
...
-                string projectId = Environment.GetEnvironmentVariable(EmulatorProjectVariable)?.Trim();
-                var endpoint = string.IsNullOrEmpty(hostAndPort) ? null : hostAndPort;
-                return new EmulatorConfiguration(endpoint, projectId == "" ? null : projectId);
-            }
+            var clientBuilder = new DatastoreClientBuilder(this);
+            clientBuilder.Settings = Settings;
+            return new ConfiguredBuilder(ProjectId, NamespaceId, clientBuilder);
+        }
+
+        private ClientBuilderBase<DatastoreDb> MaybeUseEmulator()
```
amanda-tarafa: This method is private, can't you just return DatastoreDbBuilder to save you the cast in the method above?

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -140,29 +85,37 @@ void ApplyEmulatorConfiguration()
...
-            internal string Endpoint { get; }
-            internal string ProjectId { get; }
-
-            private EmulatorConfiguration(string endpoint, string projectId)
+            var emulatorBuilder = (DatastoreDbBuilder) MaybeUseEmulator();
+            if (emulatorBuilder is object)
             {
-                Endpoint = endpoint;
-                ProjectId = projectId;
+                return emulatorBuilder.PrepareBuilder();
```
amanda-tarafa: This is a recursive call?, it will work as long as EmulatorDetection defaults to None I think.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDbBuilder.cs
```diff
@@ -96,13 +103,19 @@ public override FirestoreDb Build()
         /// <inheritdoc />
         public override async Task<FirestoreDb> BuildAsync(CancellationToken cancellationToken = default)
         {
+            var emulatorBuilder = MaybeUseEmulator();
+            if (emulatorBuilder is object)
+            {
+                return emulatorBuilder.Build();
```
amanda-tarafa: should you call BuildAsycn here?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDbBuilder.cs
```diff
@@ -147,19 +160,14 @@ private static FirestoreSettings AddGcclVersionHeader(FirestoreSettings settings
             return clone;
         }
 
-        private void ApplyEmulatorSettings(FirestoreClientBuilder clientBuilder)
+        private ClientBuilderBase<FirestoreDb> MaybeUseEmulator()
```
amanda-tarafa: Same here, maybe just return FirestoreDbBuilder.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -140,29 +85,37 @@ void ApplyEmulatorConfiguration()
...
+        }
+
+        private ClientBuilderBase<DatastoreDb> MaybeUseEmulator()
+        {
+            var environment = GetEmulatorEnvironment(s_requiredEmulatorVariables, s_allEmulatorVariables);
+            return environment == null ? null :
+                new DatastoreDbBuilder
+                {
+                    Endpoint = environment[s_emulatorHostVariable],
+                    Settings = Settings,
```
amanda-tarafa: Should the settings be cloned?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDbBuilder.cs
```diff
@@ -63,31 +66,35 @@ public sealed class FirestoreDbBuilder : ClientBuilderBase<FirestoreDb>
...
+            set => base.EmulatorDetection = value;
         }
 
         /// <inheritdoc />
         public override FirestoreDb Build()
         {
+            var emulatorBuilder = MaybeUseEmulator();
+            if (emulatorBuilder is object)
+            {
+                return emulatorBuilder.Build();
```
amanda-tarafa: Same comment as above, I think this is a recursive call that will only work while EmulatorDetection defaults to None. It's fine by me, just maybe leave a note somewhere in case we ever change the default and start seeing well, stack overflow :0.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
jskeet: I don't think we need to in this case, because the builder is just being used immediately - it's not like anything can change things later for *that* builder any more than they can for *this* builder.

Inlining this method btw, as we don't really need it to be separate.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDbBuilder.cs
jskeet: Yes, definitely. Thanks!

jskeet:
Right, this is pretty much the same code you saw before, but now using GAX 3.0.0-beta03.

amanda-tarafa:
Travis is failing, a cancellation token that hasn't been specified.

jskeet:
Thanks, will have a look.

jskeet:
Fixed - love it when analyzers can spot problems :)

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreDbBuilder.cs
```diff
@@ -140,29 +85,32 @@ void ApplyEmulatorConfiguration()
...
             {
-                Endpoint = endpoint;
-                ProjectId = projectId;
+                // Note: this is a recursive call, but because EmulatorDetection defaults to None,
+                // it will only recurse once.
+                return new DatastoreDbBuilder
+                {
+                    Endpoint = environment[s_emulatorHostVariable],
+                    Settings = Settings,
+                    ProjectId = ProjectId ?? environment[s_emulatorProjectVariable],
```
amanda-tarafa: I know it was like this before, but intuitively I would have though the emulator project ID would have precedence when using the emulator. Flagging just in case.

jskeet: I *think* it makes more sense to make code take precedence here, but I do see your point. Let's wait and see if anyone flags it :)

PR #4518, opened 2020-03-05
Update to the most recent stable version of DotCover

jskeet:
This works with .NET Core 3.1


jskeet:
(The rest should probably be updated too... the GRPC_VERSION isn't as scary as it looks, as this is just the generator, which I don't think has changed significantly. Will do all all that in a separate PR though.)

PR #4535, opened 2020-03-10
Add Google.Cloud.ServiceDirectory.V1Beta1

jskeet:
We'll need to provide a product URL before releasing the first package, but this gets most of the way there.


jskeet:
Ah, no - it was a failure to rerun generateprojects. Repushing; will merge on green.

PR #4539, opened 2020-03-11
Prototype of how we might configure emulators in Spanner.

jskeet:
Note:

- This requires a generator change in order to allow partial classes to intercept Build/BuildAsync methods
- I looked into putting common code into the common project, but the fact that every builder type has a different Settings property makes it very tricky

### DIFF
--- a/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.g.cs
+++ b/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.g.cs
@@ -404,8 +404,18 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         /// <summary>The settings to use for RPCs, or <c>null</c> for the default settings.</summary>
         public DatabaseAdminSettings Settings { get; set; }
 
+        partial void PartialBuild(ref DatabaseAdminClient client);
+        partial void PartialBuildAsync(st::CancellationToken cancellationToken, ref stt::Task<DatabaseAdminClient> task);
+
         /// <inheritdoc/>
         public override DatabaseAdminClient Build()
+        {
+            DatabaseAdminClient client = null;
+            PartialBuild(ref client);
+            return client ?? BuildImpl();
+        }
+
+        private DatabaseAdminClient BuildImpl()
         {
             Validate();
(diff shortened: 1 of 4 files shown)


jskeet:
Chris: please see whether you think the generator change makes sense.
Amanda: please review in general
Shanika: please check that this corresponds to your understanding of what's needed to connect to a Spanner emulator

Once we're all happy with this prototype, the next steps would be:

- Make the change in the generator
- Roll out partial classes for SpannerClientBuilder, DatabaseAdminClientBuilder, InstanceAdminClientBuilder
- Add an EmulatorDetection property to SpannerConnection (and SpannerConnectionString, and the session pool key etc)

jskeet:
Oh, and this is currently entirely untested :)

apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.g.cs
```diff
@@ -404,16 +404,33 @@ public sealed partial class DatabaseAdminClientBuilder : gaxgrpc::ClientBuilderB
         /// <summary>The settings to use for RPCs, or <c>null</c> for the default settings.</summary>
         public DatabaseAdminSettings Settings { get; set; }
 
+        partial void PartialBuild(ref DatabaseAdminClient client);
```
amanda-tarafa: nit: OnBuild?

jskeet: Maybe. It's trying to act as a potential replacement rather than just a response to the existing code though. We can definitely bike shed on names when we're all happy with the overall idea :)

skuruppu:
Thanks @jskeet. These are our requirements:

1. Set the endpoint to what's in the env var. I assume `GetEmulatorEnvironment()` would return a ServiceEndpoint?
2. Credentials are set to Insecure.

I think you covered both of these in this prototype.

jskeet:
(But GetEmulatorEnvironment returns a `Dictionary<string, string>`, basically mapping all the requested environment variables. ServiceEndpoint has been removed from GAX - we now just use a string as the endpoint.)

jskeet:
Closing this pull request, but I'll leave the branch available for a while for reference.
The generator change has now been made, with the partial methods now called InterceptBuild and InterceptBuildAsync.

PR #4540, opened 2020-03-11
Release Google.Cloud.ServiceDirectory.V1Beta1 version 1.0.0-beta01

jskeet:
Changes in this release:

Initial beta release.


jskeet:
Note that the change to index.md is because we don't currently have any product documentation, which the description would link to. We'll reinstate that part when there's public documentation.

PR #4542, opened 2020-03-11
Regenerate all APIs to pick up generator changes

jskeet:
This commit only contains changes to the client builders, which now
have partial methods to allow interception of the Build/BuildAsync
methods.


jskeet:
Self-reviewing, as this is purely generated. No API surface changes - merging.

PR #4545, opened 2020-03-12
[BigQuery] Removes SafeTotalRows from BigQueryResults.

amanda-tarafa:
* TotalRows is now ulong? instead of ulong.
  * SafeTotalRows has been removed.


amanda-tarafa:
Merging now, we can always rollback this change if it causes a lot of pain.

PR #4552, opened 2020-03-16
Fix BigQuery docs build

jskeet:
The snippet references needed to be fixed up to include the new optional resource parameters.


jskeet:
No problem - but I won't merge this until #4554 is in, and I can add a second commit to make the equivalent changes.

PR #4557, opened 2020-03-16
Update docfx generation

jskeet:
- Use docfx 2.50
- Use the googleapis.dev docs for Google.Apis.Auth etc


jskeet:
The other docs aren't yet separately on googleapis.dev. I'd like to get there, but docfx has problems with links in some cases, and it's also "just something else to get round to".

PR #4639, opened 2020-03-19
First stage of fixing existing docs

jskeet:
Plan:

- Run Kokoro with this commit. We should see output saying that it
  *would* upload docs for 2 APIs, but it's just a dry run.
- Revert the uploaddocs.sh change in another commit, so we can actually do it for
  those 2 APIs, and wait for those APIs to be published.
- Do it for the rest of the APIs.


jskeet:
Closing now this is fixed. Will leave the branch around for a while.

PR #4640, opened 2020-03-19
Generate a table of APIs in the README

jskeet:
This loses significant nuance in the manually-maintained version, but is more scalable.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -230,7 +230,7 @@
     "id": "Google.Cloud.Debugger.V2",
     "generator": "micro",
     "protoPath": "google/devtools/clouddebugger/v2",
-    "productName": "Stackdriver Debugger",
+    "productName": "Google Cloud Debugger",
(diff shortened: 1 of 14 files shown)


jskeet:
Note that I'll do the docs/root/index.md file after this (in a similar way) but I wanted to get this one done first.

apis/apis.json
```diff
@@ -289,7 +289,7 @@
     "type": "other",
     "targetFrameworks": "netstandard2.0",
     "testTargetFrameworks": "netcoreapp2.1;net461",
-    "description": "Google Stackdriver Instrumentation Libraries for ASP.NET Core.",
+    "description": "Google Cloud Logging Instrumentation Libraries for ASP.NET Core.",
```
amanda-tarafa: Should be Logging, Trace and Error Reporting. For all Google.Cloud.Diagnostics.*

jskeet: Will fix this as a follow-up commit (in this PR still).

PR #4644, opened 2020-03-20
Generate docs/root/index.md file in same way as README.md

jskeet:


tools/Google.Cloud.Tools.ProjectGenerator/Program.cs
```diff
@@ -177,7 +178,23 @@ static void ReformatApiCatalog()
         static void RewriteReadme(string root, List<ApiMetadata> apis)
         {
             var readmePath = Path.Combine(root, "README.md");
-            var existing = File.ReadAllLines(readmePath).ToList();
+            RewriteApiTable(readmePath, apis, api => $"https://googleapis.dev/dotnet/{api.Id}/latest");
```
amanda-tarafa: This is what I was suggesting (or similar, not sure if api.Version is exactly what needs to go there) while the TOC issue is being fixed. (And in the README at the root of the repo as well).
```suggestion
            RewriteApiTable(readmePath, apis, api => $"https://googleapis.dev/dotnet/{api.Id}/{api.Version}");
```

jskeet: This line *only* affects the README at the root repo. The one in RewriteDocsRootIndex is only a relative URL to GitHub pages.
I'd like to make the README change separately, after we've done a tiny bit more investigation into the TOC issue.

PR #4659, opened 2020-03-25
Update the Dockerfile for .NET Core SDK 3.1.100

codyoss:
This seems to be the required version when using `prepare-release.sh`.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 # Start with the .NET Core SDK
-FROM mcr.microsoft.com/dotnet/core/sdk:3.1.100
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1.100-bionic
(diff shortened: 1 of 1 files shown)


codyoss:
There also appears to be some linker issues /w some of the native bindings: 
```
...Unable to load shared library 'git2-7ce88e6' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable...
```

It seems moving this it `3.1.100-bionic` gets rid of this. Would you like to use that Ubuntu image or set the envar. I can update this PR wither way.

jskeet:
Hmm, that's odd. If using bionic works, I'd go with that. Thanks! (I'm turning laptop off for now, but will be around tomorrow.)

jskeet:
Superceded by #4662.

PR #4758, opened 2020-04-01
Add documentation link for Service Directory

jskeet:
(The documentation wasn't public before, but is now.)


jskeet:
Will merge this after all the autosynth stuff, as I don't want to risk any extra changes :)

PR #4816, opened 2020-04-06
Release Google.Cloud.Dialogflow.V2 version 3.0.0-beta01

jskeet:
Changes in this release:

- [Commit 5a41fb5](https://github.com/googleapis/google-cloud-dotnet/commit/5a41fb5):
  - Change `parent` field's resource_reference to specify child_type instead of type per client library generation requirement;
  - Change Session with its child resource pattern to support both projects/{project}/agent/sessions/{session} and projects/{project}/agent/environments/{environment}/users/{user}/sessions/{session};
  - Fix `method_signature`
  - Regular documentation update

This is a breaking change in terms of resource names, which weren't
correctly defined in the API before. We will release a new GA
version once we're satisfied that no further breaking changes will
be required.

### DIFF
--- a/apis/Google.Cloud.Dialogflow.V2/docs/history.md
+++ b/apis/Google.Cloud.Dialogflow.V2/docs/history.md
@@ -1,5 +1,19 @@
 # Version history
 
+# Version 3.0.0-beta01, released 2020-04-06
+
+- [Commit 5a41fb5](https://github.com/googleapis/google-cloud-dotnet/commit/5a41fb5):
+  - Change `parent` field's resource_reference to specify child_type instead of type per client library generation requirement;
+  - Change Session with its child resource pattern to support both projects/{project}/agent/sessions/{session} and projects/{project}/agent/environments/{environment}/users/{user}/sessions/{session};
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Dialogflow.V2/docs/history.md
```diff
@@ -1,5 +1,18 @@
 # Version history
 
+# Version 3.0.0-beta01, released 2020-04-06
+
+- [Commit 5a41fb5](https://github.com/googleapis/google-cloud-dotnet/commit/5a41fb5):
+  - Change `parent` field's resource_reference to specify child_type instead of type per client library generation requirement;
+  - Change Session with its child resource pattern to support both projects/{project}/agent/sessions/{session} and projects/{project}/agent/environments/{environment}/users/{user}/sessions/{session};
```
amanda-tarafa, line 7: And remove the one requiering location?

jskeet: This was just what's already in the commit, and the Dialogflow team didn't include that bit. I'll add it now, just in the update history.

PR #4842, opened 2020-04-08
Release Asset 1.0 and DataCatalog 1.0

jskeet:
I think this is the end of releases which are "ready to go"


jskeet:
Travis error is just a network one, on the non-cloud build. Merging.

PR #4868, opened 2020-04-16
Fix line-number-sensitive test

jskeet:
(This was broken by adding an import in an earlier PR.)

### DIFF
--- a/apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
+++ b/apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
@@ -346,8 +346,8 @@ namespace Google.Cloud.Logging.Log4Net.Tests
             {
                 Assert.True(string.IsNullOrEmpty(entry0.SourceLocation.File) || entry0.SourceLocation.File.EndsWith("Log4NetTest.cs"),
                 $"Actual 'entry0.SourceLocation.File' = '{entry0.SourceLocation.File}'");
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
```diff
@@ -346,8 +346,8 @@ public async Task SingleLogEntryWithLocation()
             {
                 Assert.True(string.IsNullOrEmpty(entry0.SourceLocation.File) || entry0.SourceLocation.File.EndsWith("Log4NetTest.cs"),
                 $"Actual 'entry0.SourceLocation.File' = '{entry0.SourceLocation.File}'");
-                // Line 44 on dev machine, line 42 on AppVeyor. Don't ask, I don't understand.
-                Assert.True(entry0.SourceLocation.Line == 0L || entry0.SourceLocation.Line == 45L || entry0.SourceLocation.Line == 43L,
+                // The exact line number seems to vary depending on framework (Core or desktop) and CI system.
+                Assert.True(entry0.SourceLocation.Line == 0L || entry0.SourceLocation.Line == 46L || entry0.SourceLocation.Line == 44L,
```
amanda-tarafa, line 350: I'm fine with this as it is, but maybe consider Assert.InRange(..., 0, 100)?

jskeet: I suspect the intention is to keep it a little tighter. Let's keep it like this for now, and we can always loosen it if we find ourselves seeing this happen frequently.

PR #4890, opened 2020-04-21
Move common properties to automatically-imported files

jskeet:


jskeet:
The error is due to deprecation - which would have given us a warning before (I'm surprised we didn't see it) and it's now an error. This is a good thing, catching warnings...

jskeet:
Hmm... that's not actually it. It's due to Microsoft.DotNet.Analyzers.Compatibility.

It's easy to fix though - done in another commit.

jskeet:
Thanks - will merge on green, but definitely not before, and will add a new analyzer issue.

jskeet:
Hmm... the error with Firestore will need a bit more investigation.

PR #4895, opened 2020-04-22
Refactor ReleaseManager into a separate class per command

jskeet:
This is in preparation for creating a new "push" command. I couldn't
bear adding to the mess of a huge Program.cs

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/Program.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/Program.cs
@@ -13,15 +13,10 @@
 // limitations under the License.
 
 using Google.Cloud.Tools.Common;
-using Google.Cloud.Tools.UpdateReleaseNotes;
-using LibGit2Sharp;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using System;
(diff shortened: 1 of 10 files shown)


tools/Google.Cloud.Tools.ReleaseManager/Program.cs
```diff
@@ -31,57 +26,34 @@ namespace Google.Cloud.Tools.ReleaseManager
...
-        private const string CompareCommand = "compare";
-        private const string UpdateHistoryCommand = "update-history";
-        private const string CommitCommand = "commit";
-
         private static int Main(string[] args)
         {
             try
             {
+                var allCommands = typeof(Program).Assembly.GetTypes()
+                    .Where(t => !t.IsAbstract && typeof(ICommand).IsAssignableFrom(t))
```
amanda-tarafa, line 34: Not sure if you'll get the abstract BaseCommand here? Suppose not given that you are creating an instance on the Select and that would have failed for the abstract class.

jskeet: That's what the `!t.IsAbstract` is for :)

PR #4904, opened 2020-04-24
[Storage]: V4 Post Policies

### LINKED ISSUE #4194: [Storage] V4 Signature POST Policies
Tracking issue for work defined in companion document.

[GCS - POST with SigV4](https://docs.google.com/document/d/1zLV5kpDf1gNER-twwt0sLs4BRSnVsPEWVnkhzEbT6SM/edit)

@jskeet @amanda-tarafa

amanda-tarafa:
Fixes #4194, #4905 

Consider reviewing one commit at a time. For now, only the first 3 commits. These include all implementation. I'm holding back on a couple of tests commits until conformance tests are updated and include the bucket in the policy.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
@@ -1,0 +1,340 @@
+﻿// Copyright 2020 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Net;
+
+namespace Google.Cloud.Storage.V1
+{
+    public sealed partial class UrlSigner
+    {
+        /// <summary>
+        /// A standard element that can be added to a <see cref="PostPolicy"/>.
+        /// </summary>
+        public enum PostPolicyStandardElement
+        {
+            /// <summary>
+            /// The ACL that you want to allow for the objects that will be uploaded.
+            /// </summary>
+            Acl = 0,
+
+            /// <summary>
+            /// The name of the bucket that you want to allow uploads to.
+            /// Alternatively, you can include the bucket name when you add 
+            /// the <see cref="PostPolicyStandardElement.Key"/> element.
+            /// </summary>
+            BucketName = 1,
+
+            /// <summary>
+            /// The cache control you want to allow for the objects that will be uploaded.
+            /// </summary>
+            CacheControl = 2,
+
+            /// <summary>
+            /// The content disposition you want to allow for the objects that will be uploaded.
+            /// </summary>
+            ContentDisposition = 3,
+
+            /// <summary>
+            /// The content encoding you want to allow for the objects that will be uploaded.
+            /// </summary>
+            ContentEncoding = 4,
+
+            /// <summary>
+            /// The content length you want to allow for the objects that will be uploaded.
+            /// </summary>
+            ContentLength = 5,
+
+            /// <summary>
+            /// The content type you want to allow for the objects that will be uploaded.
+            /// </summary>
+            ContentType = 6,
+
+            /// <summary>
+            /// The date and time you want to allow before objects that will be uploaded
+            /// are considered stale by the browser.
+            /// The value for this element will be included in the <see cref="PostPolicy"/>
+            /// in ISO8601 basic format. When using this element for <see cref="PostPolicy.SetStartsWith(PostPolicyStandardElement, string)"/>
+            /// you should specify the part of the ISO8601 basic format representation that you want
+            /// values on upload to start with.
+            /// </summary>
+            ObjectExpiry = 7,
+
+            /// <summary>
+            /// The acceptable key name or a prefix for the uploaded objects.
+            /// This element should always be present in a post policy.
+            /// </summary>
+            Key = 8,
+
+            /// <summary>
+            /// The allowed URL to which the client is redirected upon successful upload.
+            /// </summary>
+            SuccessActionRedirect = 9,
+
+            /// <summary>
+            /// The status code returned to the client upon successful upload if 
+            /// <see cref="SuccessActionRedirect"/> is not specified.
+            /// Allowed values are <see cref="HttpStatusCode.OK"/>, <see cref="HttpStatusCode.Created"/>,
+            /// <see cref="HttpStatusCode.NoContent"/>.
+            /// </summary>
+            SuccessActionStatus = 10,
+        }
+
+        /// <summary>
+        /// A custom element that can be added to a <see cref="PostPolicy"/>.
+        /// </summary>
+        public enum PostPolicyCustomElement
+        {
+            /// <summary>
+            /// Custom element used for requiering extra Google metadata to be included
+            /// in the post request. The element name you provide should start with x-goog-meta-.
+            /// </summary>
+            GoogleMetadata = 0,
+
+            /// <summary>
+            /// Custom element used for requiering extra fields to be included in the post request.
+            /// </summary>
+            Custom = 1,
+        }
+
+        /// <summary>
+        /// Represents an unsigned post policy. It needs to be signed with <see cref="UrlSigner.Sign(PostPolicy, Options)"/> 
+        /// before it is used to enforced policies when posting objects via form.
+        /// </summary>
+        public sealed class PostPolicy
+        {
+            private static readonly IPostPolicyElement[] s_knownElements = new IPostPolicyElement[]
+            {
+                Acl.Element,
+                BucketName.Element,
+                CacheControl.Element,
+                ContentDisposition.Element,
+                ContentEncoding.Element,
+                ContentLength.Element,
+                ContentType.Element,
+                ObjectExpiry.Element,
+                Key.Element,
+                SuccessActionRedirect.Element,
+                SuccessActionStatus.Element
+            };
+
+            private static readonly ICustomElementFactory[] s_customElementFactories = new ICustomElementFactory[]
+            {
+                GoogleMetadataElementFactory.Default,
+                CustomElementFactory.Default
+            };
+
+            private readonly SortedDictionary<IPostPolicyElement, IPostPolicyCondition> _conditions;
+
+            private PostPolicy()
+                :this (new ElementComparer()) 
+            { }
+
+            // Only for testing purpuses. Elements in conformance tests are not sorted.
+            private PostPolicy(IComparer<IPostPolicyElement> comparer) =>
+                _conditions = new SortedDictionary<IPostPolicyElement, IPostPolicyCondition>(comparer);
+
+            internal PostPolicy(PostPolicy other) =>
+                _conditions = new SortedDictionary<IPostPolicyElement, IPostPolicyCondition>(
+                    GaxPreconditions.CheckNotNull(other, nameof(other))._conditions, other._conditions.Comparer);
+
+            /// <summary>
+            /// Creates a new <see cref="PostPolicy"/> instance which sets the
+            /// <see cref="PostPolicyStandardElement.BucketName"/> and requires
+            /// an exact match for the <see cref="Key"/> element.
+            /// </summary>
+            /// <param name="bucketName">The exact value for the bucket name.</param>
+            /// <param name="exactKeyValue">The value to match against. Must not be null.</param>
+            /// <returns>A new <see cref="PostPolicy"/> instance.</returns>
+            public static PostPolicy ForBucketAndKey(string bucketName, string exactKeyValue) =>
+                ForBucketAndKey(bucketName, exactKeyValue, new ElementComparer());
+
+            // Same as <see cref="FromExactKey(string)"/> but receiving a comparer for testing purposes.
+            // Conformance tests are not sorted.
+            internal static PostPolicy ForBucketAndKey(string bucketName, string exactKeyValue, IComparer<IPostPolicyElement> comparer)
+            {
+                var postPolicy = new PostPolicy(comparer);
+                postPolicy.SetField(PostPolicyStandardElement.BucketName, bucketName);
+                postPolicy.SetField(PostPolicyStandardElement.Key, exactKeyValue);
+                return postPolicy;
+            }
+
+            /// <summary>
+            /// Creates a new <see cref="PostPolicy"/> instance which sets the
+            /// <see cref="PostPolicyStandardElement.BucketName"/> and requires
+            /// a starts-with match for the <see cref="PostPolicyStandardElement.Key"/> element.
+            /// </summary>
+            /// <param name="bucketName">The exact value for the bucket name.</param>
+            /// <param name="keyPrefix">The value to match against. Must not be null.</param>
+            /// <returns>A new <see cref="PostPolicy"/> instance.</returns>
+            public static PostPolicy ForBucketAndKeyStartsWith(string bucketName, string keyPrefix)
+            {
+                var postPolicy = new PostPolicy();
+                postPolicy.SetField(PostPolicyStandardElement.BucketName, bucketName);
+                postPolicy.SetStartsWith(PostPolicyStandardElement.Key, keyPrefix);
+                return postPolicy;
+            }
+
+            /// <summary>
+            /// Bucket name explicitly enforced by this policy, that is,
+            /// if <see cref="SetField{TElement, TValue}(TElement, TValue)"/> has been called
+            /// for <see cref="PostPolicyStandardElement.BucketName"/>.
+            /// Otherwise, for instance, if the bucket name was specified as part of the
+            /// <see cref="PostPolicyStandardElement.Key"/> element, this will return null.
+            /// </summary>
+            public string Bucket =>
+                _conditions.TryGetValue(BucketName.Element, out IPostPolicyCondition condition) &&
+                condition is ExactMatch<ISupportsExact<string>, string> exactMatch
+                ? exactMatch.Value : null;
+
+            /// <summary>
+            /// Collection of fields as defined by this policy.
+            /// Post requests enforced by this policy should set, for each
+            /// pair in this sequence, the given element to the exact given value.
+            /// </summary>
+            public IReadOnlyDictionary<string, object> Fields =>
+                new ReadOnlyDictionary<string, object>(
+                    _conditions
+                    .Where(pair => pair.Key != BucketName.Element)
+                    .Select(pair => pair.Value)
+                    .OfType<IExactMatch>()
+                    .ToDictionary(cond => cond.Field.Key, cond => cond.Field.Value));
+
+            /// <summary>
+            /// Sets a required field value for <paramref name="element"/>.
+            /// If any condition existed on this instance for <paramref name="element"/> it will
+            /// be overriden with this required field value.
+            /// </summary>
+            /// <typeparam name="TValue">The type of the value to set the element to.</typeparam>
+            /// <param name="element">The element for which to set a required field value.</param>
+            /// <param name="value">The value to set the element to.</param>
+            public void SetField<TValue>(PostPolicyStandardElement element, TValue value)
+            {
+                GaxPreconditions.CheckEnumValue(element, nameof(element));
+                ISupportsExact<TValue> supportsExact = s_knownElements[(int)element] as ISupportsExact<TValue> ??
+                    throw new ArgumentException($"Element {element} does not support field setting for values of {typeof(TValue).FullName}.");
+                SetField(supportsExact, value);
+            }
+
+            /// <summary>
+            /// Sets a starts-with condition for <paramref name="element"/>.
+            /// If any condition existed on this instance for <paramref name="element"/> it will
+            /// be overriden with this starts-with condition.
+            /// </summary>
+            /// <param name="element">The element for which to set an starts-with condition.</param>
+            /// <param name="prefix">The prefix to validate against.</param>
+            public void SetStartsWith(PostPolicyStandardElement element, string prefix)
+            {
+                GaxPreconditions.CheckEnumValue(element, nameof(element));
+                ISupportsStartsWith supportsStartsWith = s_knownElements[(int)element] as ISupportsStartsWith ??
+                    throw new ArgumentException($"Element {element} does not support the starts-with condition.");
+
+                AddOrReplace(supportsStartsWith, new StartsWith<ISupportsStartsWith>(supportsStartsWith, prefix));
+            }
+
+            /// <summary>
+            /// Sets a range condition for <paramref name="element"/>.
+            /// If any condition existed on this instance for <paramref name="element"/> it will
+            /// be overriden with this range condition.
+            /// </summary>
+            /// <typeparam name="TValue">The type of the values defining the range.</typeparam>
+            /// <param name="element">The element for which to set a range condition.</param>
+            /// <param name="min">The range lower bound.</param>
+            /// <param name="max">The range upper bound.</param>
+            public void SetRange<TValue>(PostPolicyStandardElement element, TValue min, TValue max)
+                where TValue : IComparable
+            {
+                GaxPreconditions.CheckEnumValue(element, nameof(element));
+                ISupportsRange<TValue> supportsRange = s_knownElements[(int)element] as ISupportsRange<TValue> ??
+                    throw new ArgumentException($"Element {element} does not support range conditions for values of {typeof(TValue).FullName}.");
+
+                AddOrReplace(supportsRange, new InRange<ISupportsRange<TValue>, TValue>(supportsRange, min, max));
+            }
+
+            /// <summary>
+            /// Sets a required field value for <paramref name="element"/> of name <paramref name="elementName"/>.
+            /// If any condition existed on this instance for <paramref name="element"/> of name <paramref name="elementName"/>
+            /// it will be overriden with this required field value.
+            /// </summary>
+            /// <typeparam name="TValue">The type of the value to set the element to.</typeparam>
+            /// <param name="element">The element for which to set a required field value.</param>
+            /// <param name="elementName">The custom element name.</param>
+            /// <param name="value">The value to set the element to.</param>
+            /// <param name="valueToObject">The function to convert <paramref name="value"/> to the
+            /// representation which with it should be included in the policy. May be null in which case
+            /// <typeparamref name="TValue"/> ToString() will be used.</param>
+            public void SetCustomField<TValue>(PostPolicyCustomElement element, string elementName, TValue value, Func<TValue, object> valueToObject = null)
+            {
+                GaxPreconditions.CheckEnumValue(element, nameof(element));
+
+                ICustomElementFactory factory = s_customElementFactories[(int)element];
+
+                ISupportsExact<TValue> supportsExact = factory.Create(elementName, valueToObject) as ISupportsExact<TValue> ??
+                    throw new ArgumentException($"Element {element} does not support field setting for values of {typeof(TValue).FullName}.");
+
+                SetField(supportsExact, value);
+            }
+
+            /// <summary>
+            /// Sets a starts-with condition for <paramref name="element"/> of name <paramref name="elementName"/>.
+            /// If any condition existed on this instance for <paramref name="element"/> of name <paramref name="elementName"/>
+            /// it will be overriden with this starts-with condition.
+            /// </summary>
+            /// <param name="element">The element for which to set an starts-with condition.</param>
+            /// <param name="elementName">The custom element name.</param>
+            /// <param name="prefix">The prefix to validate against.</param>
+            public void SetCustomStartsWith(PostPolicyCustomElement element, string elementName, string prefix)
+            {
+                GaxPreconditions.CheckEnumValue(element, nameof(element));
+
+                ICustomElementFactory factory = s_customElementFactories[(int)element];
+
+                // For starts-with we don't care about the value type, because we won't use it.
+                ISupportsStartsWith supportsStartsWith = factory.Create<object>(elementName, null) as ISupportsStartsWith ??
+                    throw new ArgumentException($"Element {element} does not support the starts-with condition.");
+
+                AddOrReplace(supportsStartsWith, new StartsWith<ISupportsStartsWith>(supportsStartsWith, prefix));
+            }
+
+            internal void WriteTo(JsonWriter json)
+            {
+                json.WritePropertyName("conditions");
+                json.WriteStartArray();
+                foreach (var condition in _conditions.Values)
+                {
+                    condition.WriteTo(json);
+                }
+                json.WriteEndArray();
+            }
+
+            internal void SetField<TElement, TValue>(TElement element, TValue value)
+                where TElement : class, ISupportsExact<TValue> =>
+                AddOrReplace(element, new ExactMatch<TElement, TValue>(element, value));
+
+            internal void AddOrReplace(IPostPolicyElement element, IPostPolicyCondition condition) =>
+                _conditions[element] = GaxPreconditions.CheckNotNull(condition, nameof(condition));
+
+            private sealed class ElementComparer : IComparer<IPostPolicyElement>
+            {
+                public int Compare(IPostPolicyElement x, IPostPolicyElement y) =>
+                    x.ElementName.CompareTo(y.ElementName);
+            }
+        }
+    }
+}
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyConditions.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyConditions.cs
@@ -1,0 +1,163 @@
+﻿// Copyright 2020 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+
+namespace Google.Cloud.Storage.V1
+{
+    public sealed partial class UrlSigner
+    {
+        /// <summary>
+        /// Represents a matching condition that is used to include
+        /// <see cref="IPostPolicyElement"/> in a <see cref="PostPolicy"/>.
+        /// </summary>
+        internal interface IPostPolicyCondition
+        {
+            /// <summary>
+            /// Writes this condition to the given <see cref="JsonWriter"/>.
+            /// </summary>
+            void WriteTo(JsonWriter json);
+        }
+
+        internal interface IExactMatch : IPostPolicyCondition
+        {
+            /// <summary>
+            /// The element name / value pair that this condition defines.
+            /// </summary>
+            KeyValuePair<string, object> Field { get; }
+        }
+
+        /// <summary>
+        /// Represents the exact match condition for <see cref="IPostPolicyElement"/>
+        /// included in a <see cref="PostPolicy"/>.
+        /// </summary>
+        /// <typeparam name="TElement">The type of the element this condition acts upon.</typeparam>
+        /// <typeparam name="TValue">The type of the values allowed by <see cref="IPostPolicyElement"/>.</typeparam>
+        internal sealed class ExactMatch<TElement, TValue> : IExactMatch
+            where TElement : class, ISupportsExact<TValue>
+        {
+            internal ExactMatch(TElement element, TValue value)
+            {
+                Element = GaxPreconditions.CheckNotNull(element, nameof(element));
+                Value = value;
+            }
+
+            /// <summary>
+            /// The element this condition applies to.
+            /// </summary>
+            public TElement Element { get; }
+
+            /// <inheritdoc/>
+            public KeyValuePair<string, object> Field => 
+                new KeyValuePair<string, object>(Element.ElementName, Element.ToPostPolicyValue(Value));
+
+            /// <summary>
+            /// The exact value to match.
+            /// </summary>
+            public TValue Value { get; }
+
+            /// <inheritdoc/>
+            void IPostPolicyCondition.WriteTo(JsonWriter json)
+            {
+                var field = Field;
+
+                json.WriteStartObject();
+                json.WritePropertyName(field.Key);
+                json.WriteValue(field.Value);
+                json.WriteEndObject();
+            }
+        }
+
+        /// <summary>
+        /// Represents the starts with condition for <see cref="IPostPolicyElement"/>
+        /// included in a <see cref="PostPolicy"/>.
+        /// <typeparam name="TElement">The type of the element this condition acts upon.</typeparam>
+        /// </summary>
+        internal sealed class StartsWith<TElement> : IPostPolicyCondition
+            where TElement : class, ISupportsStartsWith
+        {
+            internal StartsWith(TElement element, string prefix)
+            {
+                Element = GaxPreconditions.CheckNotNull(element, nameof(element));
+                Prefix = GaxPreconditions.CheckNotNull(prefix, nameof(prefix));
+            }
+
+            /// <summary>
+            /// The element this condition applies to.
+            /// </summary>
+            public TElement Element { get; }
+
+            /// <summary>
+            /// The prefix to match against.
+            /// </summary>
+            public string Prefix { get; }
+
+            /// <inheritdoc/>
+            void IPostPolicyCondition.WriteTo(JsonWriter json)
+            {
+                json.WriteStartArray();
+                json.WriteValue("starts-with");
+                json.WriteValue("$"+Element.ElementName);
+                json.WriteValue(Prefix);
+                json.WriteEndArray();
+            }
+        }
+
+        /// <summary>
+        /// Represents the range condition for <see cref="IPostPolicyElement"/>
+        /// included in a <see cref="PostPolicy"/>.
+        /// </summary>
+        /// <typeparam name="TElement">The type of the element this condition acts upon.</typeparam>
+        /// <typeparam name="TValue">
+        /// The type of the values allowed by <see cref="IPostPolicyElement"/>.</typeparam>
+        internal sealed class InRange<TElement, TValue> : IPostPolicyCondition
+            where TElement : class, ISupportsRange<TValue>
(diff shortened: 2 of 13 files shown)


jskeet:
Woot! Will look first thing on Monday.

amanda-tarafa:
The Travis and Appveyor failures are just because I split the tests commits wrong. That'll get fix when I push the rest of the tests commits.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,450 @@
...
+
+            /// <inheritdoc/>
+            string ISupportsExact<string>.ValueToString(string value) => value;
+        }
+
+        /// <summary>
+        /// The name of the bucket that you want to allow uploads to.
+        /// Alternatively, you can include the bucket name in the <see cref="Key"/> element.
+        /// </summary>
+        public sealed class BucketName : ISupportsExact<string>
```
frankyn, line 100: This can be removed given the update to conformance tests.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
```diff
@@ -255,6 +282,34 @@ public async Task<string> SignAsync(string bucket, string objectName, TimeSpan d
...
+        /// </remarks>
+        /// <param name="postPolicy">The post policy to signed and that will be enforced when making the post request.
+        /// Most not be null.</param>
+        /// <param name="options">The options used to generate the signed post policy. Must not be null.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <returns>The signed post policy, which contains all the fields that should be including in the form to post.</returns>
+        public Task<SignedPostPolicy> SignAsync(PostPolicy postPolicy, Options options, CancellationToken cancellationToken = default) =>
+            GetEffectiveSigner(GaxPreconditions.CheckNotNull(options, nameof(options)).SigningVersion).SignAsync(
+                GaxPreconditions.CheckNotNull(postPolicy, nameof(postPolicy)), options, _blobSigner, _clock, cancellationToken);
+
```
frankyn, line 312: @amanda-tarafa could you add examples how the client is used in Tests OR [This document](https://docs.google.com/document/d/1jck8zmQVBVTMY09QCfHKdh7XiY0-qzeDC6uKo1F3J6g/edit?hl=en#heading=h.8t1aqb78dzr)?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,199 @@
...
+            }
+
+            internal string ToJsonString()
+            {
+                GaxPreconditions.CheckState(_expiration.HasValue, "Can't use a policy with no expiration.");
+
+                string expirationText = _expiration.Value.ToUniversalTime().ToString("YYYY-MM-DD'T'HH:MM:SS'Z'", CultureInfo.InvariantCulture);
+                string conditions = string.Join(",", _conditions.Values.Select(cond => cond.ToJsonString()));
+
+                // {\"conditions\":[{\"key\":\"test-object\"},{\"x-goog-date\":\"20200123T043530Z\"},{\"x-goog-credential\":\"test-iam-credentials@dummy-project-id.iam.gserviceaccount.com/20200123/auto/storage/goog4_request\"},{\"x-goog-algorithm\":\"GOOG4-RSA-SHA256\"}],\"expiration\":\"2020-01-23T04:35:40Z\"}
```
frankyn, line 191: conditions willl now include `{\"bucket\":\"bucket-name\"}` to protect against attack vector discussed.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.SignedPostPolicy.cs
```diff
@@ -0,0 +1,83 @@
...
+                    (SignatureElementName, Signature),
+                    (PolicyElementName, EncodedPolicy)
+                });
+
+            /// <summary>
+            /// Collection of start with conditions as defined by this policy.
+            /// Post requests enforced by this policy should set, for each
+            /// pair in this sequence, the given element to a value starting with the given prefix.
+            /// </summary>
+            public IEnumerable<(string element, string prefix)> StartWithConditions =>
```
frankyn, line 70: Could you provide more information on what this helper does?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.SignedPostPolicy.cs
```diff
@@ -0,0 +1,83 @@
...
+                Source.StartWithConditions;
+
+            /// <summary>
+            /// Collection of in range conditions as defined by this policy.
+            /// Post requests enforced by this policy should set, for each
+            /// tuple in this sequence, the given element to a value that's in the range defined by
+            /// min and max values.
+            /// </summary>
+            public IEnumerable<(string element, object min, object max)> InRangeCondition =>
```
frankyn, line 79: Same here: Could you provide more information on what this helper does?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
```diff
@@ -185,6 +185,33 @@ public string Sign(string bucket, string objectName, TimeSpan duration, HttpMeth
...
+        /// </para>
+        /// <para>
+        /// See https://cloud.google.com/storage/docs/xml-api/post-object for more information on signed post policies.
+        /// </para>
+        /// </remarks>
+        /// <param name="postPolicy">The post policy to signed and that will be enforced when making the post request.
+        /// Must not be null.</param>
+        /// <param name="options">The options used to generate the signed post policy. Must not be null.</param>
+        /// <returns>The signed post policy, which contains all the fields that should be including in the form to post.</returns>
+        public SignedPostPolicy Sign(PostPolicy postPolicy, Options options) =>
```
frankyn, line 211: For other languages, we have a consistent method name `GenerateSignedPostPolicyV4`, looking at the design and how it is used similarly to SignedURLs, this may not be possible but wanted to raise it.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,199 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+
+namespace Google.Cloud.Storage.V1
+{
+    public sealed partial class UrlSigner
+    {
+        /// <summary>
+        /// Represents the post policy to be enforced when uploading
```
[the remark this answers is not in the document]
amanda-tarafa: Yes, agreed as it is might be a tad too much. A builder then.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,199 @@
...
+                _conditions.TryGetValue(BucketName.Element, out IPostPolicyCondition condition) &&
+                condition is ExactMatch<BucketName, string> exactMatch
+                ? exactMatch.Value : null;
+
+            /// <summary>
+            /// Collection of fields as defined by this policy.
+            /// Post requests enforced by this policy should set, for each
+            /// pair in this sequence, the given element to the exact given value.
+            /// </summary>
+            public IEnumerable<(string element, string value)> Fields =>
```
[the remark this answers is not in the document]
amanda-tarafa: Compatibility with other languages and terminology used throughout.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,199 @@
...
+                _conditions.Values.OfType<IExactMatch>().Select(f => f.Field);
+
+            /// <summary>
+            /// Collection of start with conditions as defined by this policy.
+            /// Post requests enforced by this policy should set, for each
+            /// pair in this sequence, the given element to a value starting with the given prefix.
+            /// </summary>
+            public IEnumerable<(string element, string prefix)> StartWithConditions =>
+                _conditions.Values.OfType<IStartsWith>().Select(c => c.Condition);
```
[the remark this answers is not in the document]
amanda-tarafa: I think I went overboard with the value tuples, these should all be dictionaries. At most, I can use a value tuple for the range condition (min, max), but happy to add something else for that.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,199 @@
...
+                }
+                return postPolicy;
+            }
+
+            internal string ToJsonString()
+            {
+                GaxPreconditions.CheckState(_expiration.HasValue, "Can't use a policy with no expiration.");
+
+                string expirationText = _expiration.Value.ToUniversalTime().ToString("YYYY-MM-DD'T'HH:MM:SS'Z'", CultureInfo.InvariantCulture);
+                string conditions = string.Join(",", _conditions.Values.Select(cond => cond.ToJsonString()));
```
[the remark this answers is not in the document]
amanda-tarafa: Agreed! I added these bits to get the conformance tests running. My intention was to agree with you which JSON lib we should depend on and I forgot. Json.NET it is from one of your other comments.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
amanda-tarafa: Yes, my implementation already includes them, that's why my tests were failing. Will update tests now and on next push they should be green.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyConditions.cs
```diff
@@ -0,0 +1,174 @@
...
+
+            public (string element, string value) Field =>
+                (Element.ElementName, Element.ValueToString(Value));
+
+            /// <summary>
+            /// The exact value to match.
+            /// </summary>
+            public TValue Value { get; }
+
+            string IPostPolicyCondition.ToJsonString()
```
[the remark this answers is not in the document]
amanda-tarafa: I'll think Json.NET is fine.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyConditions.cs
```diff
@@ -0,0 +1,174 @@
...
+            {
+                var (element, value) = Field;
+                if (Element.RepresentAsString)
+                {
+                    value = $"\\\"{value}\\\"";
```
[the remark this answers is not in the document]
amanda-tarafa: To make tests pass temporarily :(

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,450 @@
...
+using System.Net;
+
+namespace Google.Cloud.Storage.V1
+{
+    public sealed partial class UrlSigner
+    {
+        /// <summary>
+        /// Represents an element that can be included on a <see cref="PostPolicy"/>.
+        /// </summary>
+        public interface IPostPolicyElement
```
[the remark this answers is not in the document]
amanda-tarafa: Not really, will hide.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,450 @@
...
+        {
+            /// <summary>
+            /// The name with which the element is included on a <see cref="PostPolicy"/>.
+            /// </summary>
+            string ElementName { get; }
+
+            /// <summary>
+            /// Whether the element's value should be treated as a string or not.
+            /// Allows to decide whether it needs to be sorrounded by quotations or not.
```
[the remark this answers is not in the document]
amanda-tarafa: I hope this goes way with a proper Json lib.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,450 @@
...
+
+        /// <summary>
+        /// The ACL that you want to allow for the objects that will be uploaded.
+        /// </summary>
+        public sealed class Acl : ISupportsExact<string>, ISupportsStartsWith
+        {
+            /// <summary>
+            /// The default acl element.
+            /// </summary>
+            public static readonly Acl Element = new Acl();
```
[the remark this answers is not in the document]
amanda-tarafa: OK, these will probably be internal now anyway.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
amanda-tarafa: Nope, this is how a a bucket field is added. This is the implementation though and will be hidden from view.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.SignedPostPolicy.cs
amanda-tarafa: From what I understand one of the main use cases of post policies is as follows:

- I have a service that as part of its functionality allows consumers (other apps) to upload files to a bucket.
- My service emits post policies to those other apps to constraint the posts.
- The other apps use the policy to generate, show and validate a form to the final user who chooses the file they want to post.
- If only the fields are returned by the policy, the other apps can only partially validate that whatever the user is posting matches the policy.
- The other apps need also to know the conditions to validate that what the user is uploading is valid before it reaches the server.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.SignedPostPolicy.cs
amanda-tarafa: Same as above.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
```diff
@@ -62,15 +80,15 @@ public string Sign(RequestTemplate requestTemplate, Options options, IBlobSigner
...
             /// of the implementation.
             /// </summary>
-            private struct SigningState
+            private struct UrlSigningState
             {
-                private string _resourcePath;
-                private string _canonicalQueryString;
+                private readonly string _resourcePath;
+                private readonly string _canonicalQueryString;
                 internal byte[] _blobToSign;
```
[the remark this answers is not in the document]
amanda-tarafa: None, I'll make it readonly.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
```diff
@@ -156,6 +174,64 @@ internal SigningState(RequestTemplate template, Options options, IBlobSigner blo
...
             }
 
+            /// <summary>
+            /// State which needs to be carried between the "pre-signing" stage and "post-signing" stages
+            /// of the implementation.
+            /// </summary>
+            private struct PostPolicySigningState
+            {
+                private readonly PostPolicy _policy;
+                private string _encodedPolicy;
```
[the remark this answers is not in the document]
amanda-tarafa: Yes!

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
amanda-tarafa: I would prefer to keep it consistent with the existing URL signing method names if that's allright? @jskeet any preferences?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
amanda-tarafa: Snippets are coming!

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
frankyn: Understood, I agree language consistency is better than across the other 7 languages.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.SignedPostPolicy.cs
frankyn: Gotcha, thanks for the explanation @amanda-tarafa,

I'd hold off on surfacing *Conditions as return values in the policy. I understand the use-case to add them but it wasn't an expectation for implementation and C# would be the only one to do it at the moment.

I'm going to side on hearing customer feedback before introducing more complexity to the surface.

Please revert this surface.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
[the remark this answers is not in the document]
amanda-tarafa: @frankyn asked for all but the Fields to be removed, and I've turned the Fields into a dictionary.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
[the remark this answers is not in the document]
amanda-tarafa: Done.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.SignedPostPolicy.cs
amanda-tarafa: I've remove them, but I honestly think we are only partially helping post policy consumers if we don't include the conditions.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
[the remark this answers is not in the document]
amanda-tarafa: Done

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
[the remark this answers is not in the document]
amanda-tarafa: Done

amanda-tarafa:
The commits don't appear in order for some reason (age probably?) So review 3, 4 and 5 in that order thanks.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,459 @@
...
+            internal Acl() { }
+
+            /// <inheritdoc/>
+            string IPostPolicyElement.ElementName => "acl";
+
+            /// <inheritdoc/>
+            object ISupportsExact<string>.ToPostPolicyValue(string value) => value;
+        }
+
+        /// <summary>
```
jskeet, line 88: I suspect with one class for "exact element" and one derived class for "exact and starts with", using private/protected constructors for the element name, you could remove a lot of duplication.

Very happy to experiment with that later though. That's the nice thing about all of this being internal.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,459 @@
...
+            /// </summary>
+            /// <param name="name">The name for this element that will be lowercased and
+            /// prefixed with x-goog-meta- (if not already so) to obtain the element name. Must not be null.</param>
+            /// <param name="valueToObjectFunc">The function to use to convert <typeparamref name="TValue"/> values
+            /// to the representation with which they'll be included in a <see cref="PostPolicy"/>.
+            /// May be null in which case <typeparamref name="TValue"/> ToString() will be used.</param>
+            internal GoogleMetadata(string name, Func<TValue, object> valueToObjectFunc = null)
+            {
+                _elementName = GaxPreconditions.CheckNotNull(name, nameof(name)).ToLowerInvariant();
+                if (!_elementName.StartsWith(s_elementPrefix))
```
jskeet, line 318: Possibly specify a culture-invariant string comparison here.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,459 @@
...
+        internal sealed class GoogleMetadata<TValue> : ISupportsExact<TValue>, ISupportsStartsWith
+        {
+            private const string s_elementPrefix = "x-goog-meta-";
+
+            private readonly Func<TValue, object> _valueToObjectFunc;
+            private readonly string _elementName;
+
+            /// <summary>
+            /// Creates a new element with the given name.
+            /// If the given name is not prefixed with x-goog-meta- it will be so.
```
jskeet, line 308: This feels a little icky to me. Was it a documented requirement that all the other languages are doing? I think I'd prefer to say "we'll validate that it starts with x-goog-meta".

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,459 @@
...
+            /// prefixed with x-goog-meta- (if not already so) to obtain the element name. Must not be null.</param>
+            /// <param name="valueToObjectFunc">The function to use to convert <typeparamref name="TValue"/> values
+            /// to the representation with which they'll be included in a <see cref="PostPolicy"/>.
+            /// May be null in which case <typeparamref name="TValue"/> ToString() will be used.</param>
+            internal GoogleMetadata(string name, Func<TValue, object> valueToObjectFunc = null)
+            {
+                _elementName = GaxPreconditions.CheckNotNull(name, nameof(name)).ToLowerInvariant();
+                if (!_elementName.StartsWith(s_elementPrefix))
+                {
+                    _elementName = $"{s_elementPrefix}{_elementName}";
```
jskeet, line 320: For this I'd just use `+` rather than an interpolated string literal.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,459 @@
...
+            /// <inheritdoc/>
+            string IPostPolicyElement.ElementName => _elementName;
+
+            /// <inheritdoc/>
+            object ISupportsExact<TValue>.ToPostPolicyValue(TValue value) =>
+                _valueToObjectFunc(value);
+
+            /// <inheritdoc/>
+            public override bool Equals(object obj) =>
+                obj is GoogleMetadata<TValue> other ? _elementName.Equals(other._elementName) : false;
```
jskeet, line 334: I'd write this as `obj is GoogleMetadata<TValue> other && _elementName == other.elementName`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,459 @@
...
+
+            /// <inheritdoc/>
+            string IPostPolicyElement.ElementName => _elementName;
+
+            /// <inheritdoc/>
+            object ISupportsExact<TValue>.ToPostPolicyValue(TValue value) =>
+                _valueToObjectFunc(value);
+
+            /// <inheritdoc/>
+            public override bool Equals(object obj) =>
```
jskeet, line 384: And again. (There may be others I've missed.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicyElements.cs
```diff
@@ -0,0 +1,459 @@
...
+            public static PolicyCreationDateTime Element { get; } = new PolicyCreationDateTime();
+
+            internal PolicyCreationDateTime() { }
+
+            /// <inheritdoc/>
+            string IPostPolicyElement.ElementName => "x-goog-date";
+
+            /// <inheritdoc/>
+            object ISupportsExact<DateTime>.ToPostPolicyValue(DateTime value) =>
+                value.ToUniversalTime().ToString("YYYYMMDD'T'HHMMSS'Z'", CultureInfo.InvariantCulture);
```
jskeet, line 451: As noted in chat, this format string won't work.

Do we definitely want to support arbitrary `DateTime` values? I'd be tempted to either use `DateTimeOffset`, or require a `DateTime` with a kind of UTC. Basically let's stop users from shooting themselves in the feet with system-time-zone-sensitive code.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+using System.Net;
+
+namespace Google.Cloud.Storage.V1
+{
+    public sealed partial class UrlSigner
+    {
+        /// <summary>
+        /// A known element that can be added to a <see cref="PostPolicy"/>.
+        /// </summary>
+        public enum PostPolicyKnownElement
```
jskeet, line 30: Possibly "standard" instead of "known"? (Standard vs custom is a fairly common distinction, I think.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+
+            private readonly IDictionary<IPostPolicyElement, IPostPolicyCondition> _conditions =
+                new Dictionary<IPostPolicyElement, IPostPolicyCondition>();
+
+            private PostPolicy() { }
+
+            internal PostPolicy(PostPolicy other)
+            {
+                GaxPreconditions.CheckNotNull(other, nameof(other));
+                foreach (var condition in other._conditions)
```
jskeet, line 153: An alternative would be to initialize `_conditions` in each constructor, and make this one just use `new Dictionary<...>(other._conditions)`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+                {
+                    _conditions.Add(condition);
+                }
+            }
+
+            /// <summary>
+            /// Creates a new <see cref="PostPolicy"/> instance which requieres
```
jskeet, line 160: Nit: spelling of requires. There are a few of these, I think in a few permutations. Let's catch what we can no, but then do another round of review after merging - I'm happy to go through it locally.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+                return postPolicy;
+            }
+
+            internal string Bucket =>
+                _conditions.TryGetValue(BucketName.Element, out IPostPolicyCondition condition) &&
+                condition is ExactMatch<BucketName, string> exactMatch
+                ? exactMatch.Value : null;
+
+            /// <summary>
+            /// Collection of fields as defined by this policy.
```
jskeet, line 191: It feels like the name and first summary should emphasize the "exact match" part. Do we need this method at all? (I see the StartsWithConditions etc have been removed - could this go too?)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+            public void SetField<TValue>(PostPolicyKnownElement element, TValue value)
+            {
+                GaxPreconditions.CheckEnumValue(element, nameof(element));
+                ISupportsExact<TValue> supportsExact = s_knownElements[(int)element] as ISupportsExact<TValue> ??
+                    throw new ArgumentException($"Element {element} does not support field setting for values of {typeof(TValue).FullName}.");
+                SetField(supportsExact, value);
+            }
+
+            /// <summary>
+            /// Sets a start-with condition for <paramref name="element"/>.
```
jskeet, line 217: Nit: start-with => starts-with

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+            /// If any condition existed on this instance for <paramref name="element"/> it will
+            /// be overriden with this start-with condition.
+            /// </summary>
+            /// <param name="element">The element for which to set an starts-with condition.</param>
+            /// <param name="prefix">The prefix to validate against.</param>
+            public void SetStartsWith(PostPolicyKnownElement element, string prefix)
+            {
+                GaxPreconditions.CheckEnumValue(element, nameof(element));
+                ISupportsStartsWith supportsStartsWith = s_knownElements[(int)element] as ISupportsStartsWith ??
+                    throw new ArgumentException($"Element {element} does not support the start-with condition.");
```
jskeet, line 227: Ditto

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+                ICustomElementFactory factory = s_customElementFactories[(int)element];
+
+                ISupportsExact<TValue> supportsExact = factory.Create(elementName, valueToObject) as ISupportsExact<TValue> ??
+                    throw new ArgumentException($"Element {element} does not support field setting for values of {typeof(TValue).FullName}.");
+
+                SetField(supportsExact, value);
+            }
+
+            /// <summary>
+            /// Sets a start-with condition for <paramref name="element"/> of name <paramref name="elementName"/>.
```
jskeet, line 276: Another start-with - probably worth doing a global search.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+            /// it will be overriden with this required field value.
+            /// </summary>
+            /// <typeparam name="TValue">The type of the value to set the element to.</typeparam>
+            /// <param name="element">The element for which to set a required field value.</param>
+            /// <param name="elementName">The custom element name.</param>
+            /// <param name="value">The value to set the element to.</param>
+            /// <param name="valueToObject">The function to convert <paramref name="value"/> to the
+            /// representation which with it should be included in the policy. May be null in which case
+            /// <typeparamref name="TValue"/> ToString() will be used.</param>
+            public void SetCustomField<TValue>(PostPolicyCustomElement element, string elementName, TValue value, Func<TValue, object> valueToObject = null)
```
jskeet, line 263: Given that there are only two options here (Google metadata or genuinely custom) is the enum worth having rather than just two methods? Or is that part of the language-agnostic design requirements?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+                    throw new ArgumentException($"Element {element} does not support the start-with condition.");
+
+                AddOrReplace(supportsStartsWith, new StartsWith<ISupportsStartsWith>(supportsStartsWith, prefix));
+            }
+
+            internal void WriteTo(JsonWriter json)
+            {
+                json.WritePropertyName("conditions");
+                json.WriteStartArray();
+                foreach (var condition in _conditions.Values)
```
jskeet, line 300: I think it would be best to order the JSON so that it's deterministic.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.SignedPostPolicy.cs
```diff
@@ -0,0 +1,77 @@
...
+        /// making the post request.
+        /// </summary>
+        public sealed class SignedPostPolicy
+        {
+            private const string SignatureElementName = "x-goog-signature";
+            private const string PolicyElementName = "policy";
+
+            internal SignedPostPolicy(PostPolicy source, string encodedPolicy, string signature, DateTimeOffset expiration, Uri url)
+            {
+                Source = GaxPreconditions.CheckNotNull(source, nameof(source));
```
jskeet, line 36: Should we clone the source here? Currently the user could create a policy, get a signed version, add a field, and then the signed policy becomes invalid.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V2Signer.cs
```diff
@@ -44,6 +44,13 @@ public string Sign(RequestTemplate requestTemplate, Options options, IBlobSigner
                 return state.GetResult(signature);
             }
 
+            public SignedPostPolicy Sign(PostPolicy postPolicy, Options options, IBlobSigner blobSigner, IClock clock) =>
+                throw new NotSupportedException($"Post policy signing is not supported by {nameof(SigningVersion)}.{SigningVersion.V2}.");
+
```
jskeet, line 49: Super-nit - remove a blank line.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
```diff
@@ -156,6 +176,78 @@ internal SigningState(RequestTemplate template, Options options, IBlobSigner blo
...
+
+                    StringBuilder sb = new StringBuilder();
+                    StringWriter sw = new StringWriter(sb);
+
+                    using (JsonWriter writer = new JsonTextWriter(sw))
+                    {
+                        writer.WriteStartObject();
+                        policy.WriteTo(writer);
+                        writer.WritePropertyName("expiration");
+                        writer.WriteValue(_expiration.UtcDateTime.ToString("YYYY-MM-DD'T'HH:MM:SS'Z'", CultureInfo.InvariantCulture));
```
jskeet, line 235: Format string needs fixing again.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
```diff
@@ -156,6 +176,78 @@ internal SigningState(RequestTemplate template, Options options, IBlobSigner blo
...
+                    }
+
+                    _blobToSign = Encoding.UTF8.GetBytes(sb.ToString());
+                    _encodedPolicy = FormatHex(_blobToSign);
+                }
+
+                internal SignedPostPolicy GetResult(string signature) =>
+                    new SignedPostPolicy(_policy, _encodedPolicy, signature, _expiration, _url);
+
+                private static string CheckBucketNotNull(string bucket, UrlStyle style) =>
```
jskeet, line 246: Not used in this code? (Presumably a left-over from C&P)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.cs
jskeet: I'm fine with either just `Sign` or `SignPostPolicy` (the latter being reasonable as a "extra specificity" given that this is a **URL**-signer, but being asked to sign something other than a URL). But I definitely prefer both of those over `GenerateSignedPostPolicyV4`.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+
+        /// <summary>
+        /// A custom element that can be added to a <see cref="PostPolicy"/>.
+        /// </summary>
+        public enum PostPolicyCustomElement
+        {
+            /// <summary>
+            /// Custom element used for requiering extra Google metadata to be included
+            /// in the post request. If the element name you provide is not prefixed by
+            /// x-goog-meta-, it will be prefixed with x-goog-meta- before being added to
```
frankyn, line 107: `x-goog-meta-*` will be stored in Object resource custom metadata, for example `x-goog-meta-custom-1` will add a custom metadata key,value pair of `{"custom-1": "value"}`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+                }
+            }
+
+            /// <summary>
+            /// Creates a new <see cref="PostPolicy"/> instance which requieres
+            /// an exact match for the <see cref="Key"/> element.
+            /// </summary>
+            /// <param name="exactKeyValue">The value to match against. Must not be null.</param>
+            /// <returns>A new <see cref="PostPolicy"/> instance.</returns>
+            public static PostPolicy FromExactKey(string exactKeyValue)
```
frankyn, line 165: This looks a bit specific in that only one field is set, but might be misunderstanding.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.V4Signer.cs
```diff
@@ -156,6 +176,78 @@ internal SigningState(RequestTemplate template, Options options, IBlobSigner blo
...
+                    {
+                        throw new ArgumentOutOfRangeException(nameof(options.Expiration), "Expiration must not be greater than 7 days.");
+                    }
+
+                    var datestamp = now.ToString("yyyyMMdd", CultureInfo.InvariantCulture);
+                    string credentialScope = $"{datestamp}/{DefaultRegion}/{ScopeSuffix}";
+
+                    policy.SetField(PolicyCreationDateTime.Element, now);
+                    policy.SetField(PolicyAlgorithm.Element, Algorithm);
+                    policy.SetField(PolicyCredential.Element, $"{blobSigner.Id}/{credentialScope}");
```
frankyn, line 221: I'm confused with fields such as `key`, `acl` are added into the signature in this case, but should be caught with conformance tests if there's an issue.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
```diff
@@ -0,0 +1,315 @@
...
+            /// The ACL that you want to allow for the objects that will be uploaded.
+            /// </summary>
+            Acl = 0,
+
+            /// <summary>
+            /// The name of the bucket that you want to allow uploads to.
+            /// Alternatively, you can include the bucket name when you add 
+            /// the <see cref="PostPolicyKnownElement.Key"/> element.
+            /// </summary>
+            BucketName = 1,
```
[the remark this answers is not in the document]
frankyn: Thanks for clarifying. I'm pending the surface examples to parse the design better, but in other languages Bucket is a required parameter as is Key:

```python
client.generate_signed_post_policy_v4(
    bucket_name="rsaposttest-1579902678-pt5yms55j47r6qy4",
    blob_name="test-object",
    expiration=datetime.datetime.now() + datetime.timedelta(seconds=10),
    fields={"success_action_status": 200},
)
```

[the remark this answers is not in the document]
amanda-tarafa: Ah but I believe the bucket cannot be required because it can be specified as part of the key. In that case it goes into the policy as part of the key so everything is still secured, but there's no need to specify the bucket separately.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
frankyn: +1, specifically about the comment above not about the implementation.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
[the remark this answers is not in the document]
frankyn: Gotcha, the reason why I'm pushing for bucket to be required is to support the following three cases:

1. Virtual Hosted Path: https://bucket-name.storage.googleapis.com
1. Path Style: https://storage.googleapis.com/bucket-name
1. Bucket Bound Hostname: http://cname.domain.tld

Based on discussions with the GCS team and research requiring the bucket name then storing it in policy is adaptable for these three cases where in contrast `key: bucket-name/object-name` will only work for Case 1&2, but not 3.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.PostPolicy.cs
frankyn: Follow-up from discussion above, can it include bucket name as well?

frankyn:
Raised this in examples document but adding it here as well:

Could you update `FromExactKey` to `NewPostPolicy`? It's no longer just the key/object in this case. Otherwise this LGTM. 
```
UrlSigner urlSigner = UrlSigner.FromServiceAccountCredential(credential);

UrlSigner.Options options = UrlSigner.Options
    .FromDuration(TimeSpan.FromSeconds(10))
    .WithSigningVersion(SigningVersion.V4)
    .WithScheme("https");

UrlSigner.PostPolicy postPolicy =
    UrlSigner.PostPolicy.FromExactKey(bucketName, objectName);

UrlSigner.SignedPostPolicy signedPostPolicy = await urlSigner.SignAsync(postPolicy, options);

````

amanda-tarafa:
@frankyn all done. Will wait on green before merging in case you want to take a last look.

PR #4915, opened 2020-04-29
feat: spanner emulator support

skuruppu:
- Allows users to set EmulatorDetection through SpannerConnectionStringBuilder.
- Configures integration tests with EmulatorDetection.EmulatorOrProduction.

### DIFF
--- a/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.g.cs
+++ b/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.g.cs
@@ -398,6 +398,10 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         /// <summary>The settings to use for RPCs, or <c>null</c> for the default settings.</summary>
         public DatabaseAdminSettings Settings { get; set; }
 
+        partial void PartialBuild(ref DatabaseAdminClient client);
+
+        partial void PartialBuildAsync(st::CancellationToken cancellationToken, ref stt::Task<DatabaseAdminClient> task);
+
         partial void InterceptBuild(ref DatabaseAdminClient client);
 
         partial void InterceptBuildAsync(st::CancellationToken cancellationToken, ref stt::Task<DatabaseAdminClient> task);
@@ -406,7 +410,7 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         public override DatabaseAdminClient Build()
         {
             DatabaseAdminClient client = null;
-            InterceptBuild(ref client);
+            PartialBuild(ref client);
             return client ?? BuildImpl();
         }
 
@@ -414,7 +418,7 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         public override stt::Task<DatabaseAdminClient> BuildAsync(st::CancellationToken cancellationToken = default)
         {
             stt::Task<DatabaseAdminClient> task = null;
-            InterceptBuildAsync(cancellationToken, ref task);
+            PartialBuildAsync(cancellationToken, ref task);
             return task ?? BuildAsyncImpl(cancellationToken);
         }
 
--- a/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.g.cs
+++ b/apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.g.cs
@@ -398,10 +398,6 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         /// <summary>The settings to use for RPCs, or <c>null</c> for the default settings.</summary>
         public DatabaseAdminSettings Settings { get; set; }
 
-        partial void PartialBuild(ref DatabaseAdminClient client);
-
-        partial void PartialBuildAsync(st::CancellationToken cancellationToken, ref stt::Task<DatabaseAdminClient> task);
-
         partial void InterceptBuild(ref DatabaseAdminClient client);
 
         partial void InterceptBuildAsync(st::CancellationToken cancellationToken, ref stt::Task<DatabaseAdminClient> task);
@@ -410,7 +406,7 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         public override DatabaseAdminClient Build()
         {
             DatabaseAdminClient client = null;
-            PartialBuild(ref client);
+            InterceptBuild(ref client);
             return client ?? BuildImpl();
         }
 
@@ -418,7 +414,7 @@ namespace Google.Cloud.Spanner.Admin.Database.V1
         public override stt::Task<DatabaseAdminClient> BuildAsync(st::CancellationToken cancellationToken = default)
         {
             stt::Task<DatabaseAdminClient> task = null;
-            PartialBuildAsync(cancellationToken, ref task);
+            InterceptBuildAsync(cancellationToken, ref task);
             return task ?? BuildAsyncImpl(cancellationToken);
         }
 
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Api.Gax;
 using Google.Cloud.ClientTesting;
 using Google.Cloud.Spanner.Common.V1;
 using Google.Cloud.Spanner.V1.Internal.Logging;
@@ -81,7 +82,8 @@ namespace Google.Cloud.Spanner.Data.CommonTesting
             var builder = new SpannerConnectionStringBuilder
             {
                 Host = SpannerHost,
-                DataSource = $"projects/{ProjectId}/instances/{SpannerInstance}"
+                DataSource = $"projects/{ProjectId}/instances/{SpannerInstance}",
+                EmulatorDetection = EmulatorDetection.EmulatorOrProduction
             };
             if (SpannerPort != null)
             {
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
@@ -15,6 +15,7 @@
 using Google.Api.Gax;
 using Google.Cloud.ClientTesting;
 using Google.Cloud.Spanner.Common.V1;
+using Google.Cloud.Spanner.V1;
 using Google.Cloud.Spanner.V1.Internal.Logging;
 using System;
 
@@ -71,6 +72,7 @@ namespace Google.Cloud.Spanner.Data.CommonTesting
         // Connection string without the database, generated from the above properties
         public string NoDbConnectionString { get; }
         public string ProjectId { get; }
+        public SpannerConnectionStringBuilder Builder { get; }
         public DatabaseName DatabaseName { get; }
         internal SpannerClientCreationOptions SpannerClientCreationOptions { get; }
 
@@ -79,25 +81,50 @@ namespace Google.Cloud.Spanner.Data.CommonTesting
             TestLogger.Install();
 
             ProjectId = projectId;
+            // Tests may connect to the emulator if the emulator host environment variable is set.
+            EmulatorDetection emulatorDetection = EmulatorDetection.EmulatorOrProduction;
+
             var builder = new SpannerConnectionStringBuilder
             {
                 Host = SpannerHost,
                 DataSource = $"projects/{ProjectId}/instances/{SpannerInstance}",
-                EmulatorDetection = EmulatorDetection.EmulatorOrProduction
+                EmulatorDetection = emulatorDetection
             };
+
+            SessionPoolOptions options = new SessionPoolOptions();
+
+            // Create a builder that may create a client that connects to the
+            // emulator (if configured) so that we can configure the session
+            // pool. This is necessary because the emulator doesn't support
+            // running concurrent transactions so we need to configure the
+            // session pool to only have one active session at a time so that
+            // we don't prepare multiple concurrent transactions.
+            var clientBuilder = new SpannerClientBuilder
+            {
+                EmulatorDetection = emulatorDetection
+            };
+            var emulatorBuilder = clientBuilder.MaybeCreateEmulatorClientBuilder();
+            if (emulatorBuilder != null)
+            {
+                options.MinimumPooledSessions = 1;
+                options.MaximumActiveSessions = 1;
+            }
+
+            builder.SessionPoolManager = SessionPoolManager.Create(options);
             if (SpannerPort != null)
             {
                 builder.Port = int.Parse(SpannerPort);
             }
             NoDbConnectionString = builder.ConnectionString;
             SpannerClientCreationOptions = new SpannerClientCreationOptions(builder);
-            var databaseBuilder = builder.WithDatabase(SpannerDatabase);
-            ConnectionString = databaseBuilder.ConnectionString;
-            DatabaseName = databaseBuilder.DatabaseName;
+            Builder = builder.WithDatabase(SpannerDatabase);
+            Builder.EmulatorDetection = emulatorDetection;
+            ConnectionString = Builder.ConnectionString;
+            DatabaseName = Builder.DatabaseName;
 
             if (Fresh)
             {
-                using (var connection = new SpannerConnection(NoDbConnectionString))
+                using (var connection = new SpannerConnection(builder))
                 {
                     var createCmd = connection.CreateDdlCommand($"CREATE DATABASE {SpannerDatabase}");
                     createCmd.ExecuteNonQuery();
@@ -116,6 +143,6 @@ namespace Google.Cloud.Spanner.Data.CommonTesting
             return string.IsNullOrEmpty(value) ? defaultValue : value;
         }
         
-        public SpannerConnection GetConnection() => new SpannerConnection(ConnectionString);        
+        public SpannerConnection GetConnection() => new SpannerConnection(Builder);
     }
 }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
@@ -15,7 +15,6 @@
 using Google.Api.Gax;
 using Google.Cloud.ClientTesting;
 using Google.Cloud.Spanner.Common.V1;
-using Google.Cloud.Spanner.V1;
 using Google.Cloud.Spanner.V1.Internal.Logging;
 using System;
 
@@ -72,7 +71,6 @@ namespace Google.Cloud.Spanner.Data.CommonTesting
         // Connection string without the database, generated from the above properties
         public string NoDbConnectionString { get; }
         public string ProjectId { get; }
-        public SpannerConnectionStringBuilder Builder { get; }
         public DatabaseName DatabaseName { get; }
         internal SpannerClientCreationOptions SpannerClientCreationOptions { get; }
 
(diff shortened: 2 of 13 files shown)


skuruppu:
Currently this only addresses the second dot point of the [comment](https://github.com/googleapis/google-cloud-dotnet/pull/4539#issuecomment-597524619). I am still figuring out how to address the third point.

@jskeet I just wanted some quick feedback from you on whether I'm on the right track :) No need to do a detailed review for now.

apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.g.cs
```diff
@@ -412,15 +416,15 @@ public sealed partial class DatabaseAdminClientBuilder : gaxgrpc::ClientBuilderB
         public override DatabaseAdminClient Build()
         {
             DatabaseAdminClient client = null;
-            InterceptBuild(ref client);
+            PartialBuild(ref client);
```
jskeet, line 419: The generated files shouldn't need to be changed. Just keep the method name as `InterceptBuild` and `InterceptBuildAsync`.

apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClientBuilder.cs
```diff
@@ -0,0 +1,57 @@
...
+        partial void PartialBuild(ref DatabaseAdminClient client) => client = MaybeCreateEmulatorClientBuilder()?.Build();
+
+        partial void PartialBuildAsync(CancellationToken cancellationToken, ref Task<DatabaseAdminClient> task) =>
+            task = MaybeCreateEmulatorClientBuilder()?.BuildAsync(cancellationToken);
+
+        private DatabaseAdminClientBuilder MaybeCreateEmulatorClientBuilder()
+        {
+            var emulatorEnvironment = GetEmulatorEnvironment(s_emulatorEnvironmentVariables, s_emulatorEnvironmentVariables);
+            return emulatorEnvironment is null ? null :
+                new DatabaseAdminClientBuilder
```
jskeet, line 49: It may be worth adding a comment to the effect that you're not copying the EmulatorDetection property here, deliberately - that's what keeps it from recursing infinitely.

apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClient.g.cs
skuruppu: Ok cool, I think you also mean to change the name in the builder class. I checked to see if anything else was implementing the Intercept* methods but doesn't seem that way so I think this is safe. It compiles at least.

jskeet: Yes, exactly. They were introduced primarily for this feature, so there really shouldn't be anything else conflicting with it :)

apis/Google.Cloud.Spanner.Admin.Database.V1/Google.Cloud.Spanner.Admin.Database.V1/DatabaseAdminClientBuilder.cs
skuruppu: I added a comment to each builder. Hope this is ok.

skuruppu:
This is ready for review (I do need to fix the unit tests which I'll do now). I ran the ITs and they ran for the most part:

Total tests: 167
     Passed: 92
     Failed: 74
    Skipped: 1

I will go through the failed tests to figure out which ones need to be skipped (due to the emulator not supporting all features that the production backend supports). This will need to be done in a separate PR. If there's anything that can be fixed in terms of the setup, I will do so in this PR.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -69,6 +69,16 @@ static SessionPoolManager()
...
 
+        /// <summary>
+        /// Specifies whether to allow the connection to check for the presence of the emulator
+        /// environment variable.
+        /// </summary>
+        /// <remarks>
+        /// This property defaults to <see cref="EmulatorDetection.None"/>, meaning that the
+        /// environment variable is ignored.
+        /// </remarks>
+        public EmulatorDetection EmulatorDetection { get; set; }
```
jskeet, line 80: I don't think this belongs here. It's generally immutable.

I don't *think* you need to change SessionPoolManager at all - because the emulator detection should instead affect the SpannerClientCreationOptions that are in use.
SpannerClientCreationOptions gets its information from a SpannerConnectionStringBuilder - so it can use the settings there.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -313,7 +323,8 @@ private static async Task<SpannerClient> CreateClientAsync(SpannerClientCreation
             return new SpannerClientBuilder
             {
                 CallInvoker = callInvoker,
-                Settings = spannerSettings
+                Settings = spannerSettings,
+                EmulatorDetection = emulatorDetection
```
jskeet, line 327: Setting a call invoker and emulator detection would be an odd thing to do. (I'm slightly surprised we don't prohibit it.)
If we do the work in SpannerClientCreationOptions (which may mean some code duplication, unfortunately - let's see what we can do) the endpoint and credentials will already be set correctly, so we shouldn't need to make any changes here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -307,6 +307,20 @@ public SessionPoolManager SessionPoolManager
...
 
+        /// <summary>
+        /// Specifies whether to allow the connection to check for the presence of the emulator
+        /// environment variable.
+        /// </summary>
+        /// <remarks>
+        /// This property defaults to <see cref="EmulatorDetection.None"/>, meaning that the
+        /// environment variable is ignored.
+        /// </remarks>
+        public EmulatorDetection EmulatorDetection
```
jskeet, line 318: I think this should be handled in a fashion much more like properties such as MaximumGrpcChannels, so it becomes part of the connection string itself.

skuruppu:
Thanks @jskeet. I agree about doing this as two separate PRs. To preserve the discussion, I may leave this PR as is to continue the work on setting the connection properties. I'll cherry-pick the builders into a separate PR.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
skuruppu: Oh I see, so we shouldn't rely on the automated emulator detection for the data client. We should only do that for the admin clients. So we'll manually read the env var and set the endpoint and credentials. Is the worry that it would confuse users because they would set an endpoint and credentials but we'd be using a different parameters for the actual connection?

jskeet: It's not a matter of "not relying" on the automated emulation code - it's not that we don't trust it. It's a matter of making it fit in well with the rest of the infrastructure.
I'd like to investigate how to do this without code duplication though - for example, it may be possible to add a *public* member to `SpannerClientBuilder` to say "please let me know how you would be configured for emulation" (effectively like the current partial method does) and then use the endpoint and credentials from that.

I think our aim should be for the behavior to be consistent between "using SpannerClientBuilder directly" and "using SpannerConnectionStringBuilder directly" in terms of the connection.

jskeet:
I've merged #4928 now, so it's probably worth rebasing this PR, unless you'd prefer to start a new one.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
skuruppu: All done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
skuruppu: Ok that makes sense. Now we're getting the value from SpannerClientCreationOptions through SpannerConnectionStringBuilder.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
skuruppu: I've made the minimal amount of changes to get this working. But I'm still not updating the endpoint and credentials in the connection string to reflect what is actually being connected to. I think this is where your comment regarding updating the `SpannerClientBuilder` comes in.

The other issue that I want to solve is that if the connection string has EmulatorDetection then users shouldn't be asked to specify credentials. I'll also work on trying to bypass this check.

jskeet:
I'll get to this when I can, but it may go into next week I'm afraid.

skuruppu:
No worries @jskeet. I'm also sporadically working on it when I find bits of time here and there.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -289,6 +290,31 @@ public int Timeout
...
+                if (TryGetValue(EmulatorDetectionKeyword, out object value) &&
+                    (value is EmulatorDetection parsed || (value is string text && EmulatorDetection.TryParse(text, out parsed))))
+                {
+                    return parsed >= EmulatorDetection.None && parsed <= EmulatorDetection.EmulatorOrProduction ? parsed : EmulatorDetection.None;
+                }
+                return EmulatorDetection.None;
+            }
+            set
+            {
+                this[EmulatorDetectionKeyword] = (int) value;
```
jskeet, line 314: Setting it to the integer value will make the connection string very hard to read in string form. I would call ToString - after validating that it's a valid value using `GaxPreconditions.CheckEnumValue`.

skuruppu:
This is what I came up with for now https://github.com/googleapis/google-cloud-dotnet/pull/4915/commits/e98b470ec20d5318fc13040a82c6a104e737711a. I don't think you need to move the CallInvoker to SpannerClientCreationOptions, as long as the endpoint and credentials can be correctly extracted from that class.

My proposed change to SpannerClientBuilder is a bit of a mess since there's no nice way to extract the endpoint and credentials without attempting to build an instance of the class. For now I'm reading the env var manually. But if there was a static version of `GetEmulatorEnvironment` that also accepts an `EmulatorDetection` argument, then it might be cleaner.

Sorry, I know the back-and-forth is taking a bit of time. I'm more than happy for you to take a crack at this but I know you have a lot more stuff going on so I want to help as much as possible.

jskeet:
Building an instance of `SpannerClientBuilder` is the way to go - we want to get the same exact behaviour as if we were specifying the emulator detection in the builder.

I would suggest just making `MaybeCreateEmulatorClientBuilder` public - then you can use that in `SessionPoolManager`. If it returns a non-null value, use the endpoint and credentials from that.

skuruppu:
PTAL @jskeet. I made all the discussed changes.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
```diff
@@ -120,6 +135,19 @@ public override string ToString()
...
+            var builder = new SpannerClientBuilder
+            {
+                EmulatorDetection = EmulatorDetection
+            };
+            var emulatorBuilder = builder.MaybeCreateEmulatorClientBuilder();
+            if (emulatorBuilder != null)
+            {
+                return emulatorBuilder.ChannelCredentials;
+            }
+
```
skuruppu, line 150: I was a bit surprised I had to do this.

I was attempting to call this function conditionally in `SessionPoolManager.CreateClientAsync()` only if `MaybeCreateEmulatorClientBuilder()` returns null. But this call kept happening and returning an error saying that a credentials files is not given  so I assume that it has something to do with the fact that it's async. I was trying to read up on when async functions get called but didn't find a good source that explained my problem.

jskeet:
Will have a look when I get a chance, but it may well be quite late in the week.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilder.cs
```diff
@@ -42,7 +42,16 @@ public new EmulatorDetection EmulatorDetection
         partial void InterceptBuildAsync(CancellationToken cancellationToken, ref Task<SpannerClient> task) =>
             task = MaybeCreateEmulatorClientBuilder()?.BuildAsync(cancellationToken);
 
-        private SpannerClientBuilder MaybeCreateEmulatorClientBuilder()
+        /// <summary>
+        /// May return a builder that will connect to the emulator under certain conditions.
+        /// </summary>
+        /// <remarks>
+        /// Returns null if <see cref="EmulatorDetection"/> is <see cref="EmulatorDetection.None"/>
+        /// or <see cref="EmulatorDetection.ProductionOnly"/>. Otherwise, returns a builder to
```
jskeet, line 50: I'd write this as "it returns null if the emulator should not be used" - it will return null if EmulatorDetection is EmulatorOrProduction but the environment variable isn't configured, for example.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilder.cs
```diff
@@ -42,7 +42,16 @@ public new EmulatorDetection EmulatorDetection
...
+        /// connect to the emulator using the endpoint specified by the emulator host environment
```
jskeet, line 51: I'd probably say it returns a builder with the endpoint and credentials set appropriately for the emulator.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
jskeet: I can't tell what the code looked like when it was failing, but it seems reasonable to me. I don't think this has anything to do with async.

Basically if we want to use the emulator, we need to specify insecure channel credentials - otherwise, we use the normal code.

I'd probably write it in one step though:

```csharp
var emulatorBuilder = new SpannerClientBuilder { EmulatorDetection = EmulatorDetection }.MaybeCreateEmulatorClientBuilder();
...
```

If we use it elsewhere for the endpoint, you might want to put this into a common method.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
```diff
@@ -79,25 +81,50 @@ private SpannerTestDatabase(string projectId)
...
+            // emulator (if configured) so that we can configure the session
+            // pool. This is necessary because the emulator doesn't support
+            // running concurrent transactions so we need to configure the
+            // session pool to only have one active session at a time so that
+            // we don't prepare multiple concurrent transactions.
+            var clientBuilder = new SpannerClientBuilder
+            {
+                EmulatorDetection = emulatorDetection
             };
+            var emulatorBuilder = clientBuilder.MaybeCreateEmulatorClientBuilder();
```
jskeet, line 106: I wonder whether this should be in the production code, if everyone will always need to do that.
Do we know whether this is a long-term restriction?

jskeet: In general, the changes in this file (beyond just setting EmulatorDetection) feel like they're a symptom of us not doing quite enough to "just make it work".

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
```diff
@@ -79,25 +81,50 @@ private SpannerTestDatabase(string projectId)
...
             {
                 builder.Port = int.Parse(SpannerPort);
             }
             NoDbConnectionString = builder.ConnectionString;
             SpannerClientCreationOptions = new SpannerClientCreationOptions(builder);
-            var databaseBuilder = builder.WithDatabase(SpannerDatabase);
-            ConnectionString = databaseBuilder.ConnectionString;
-            DatabaseName = databaseBuilder.DatabaseName;
+            Builder = builder.WithDatabase(SpannerDatabase);
+            Builder.EmulatorDetection = emulatorDetection;
```
jskeet, line 121: We shouldn't need to set the EmulatorDetection again, should we?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
```diff
@@ -116,6 +143,6 @@ private static string GetEnvironmentVariableOrDefault(string name, string defaul
             return string.IsNullOrEmpty(value) ? defaultValue : value;
         }
         
-        public SpannerConnection GetConnection() => new SpannerConnection(ConnectionString);        
+        public SpannerConnection GetConnection() => new SpannerConnection(Builder);
```
jskeet, line 146: Why is this change required? I'd expect the connection string to be fine.

jskeet: Or is this because we've set the SessionPoolManager because of the options? It feels like a lot of this change could go away if we didn't need to set the options explicitly.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
skuruppu: Thanks for the suggestion. Now it's in a separate function and I've cleaned up the calls a bit. Now there's no need to modify SessionPoolManager at all and all the changes are confined to this file.

skuruppu:
I addressed most of the changes except for the ones in https://github.com/googleapis/google-cloud-dotnet/pull/4915/files#diff-450b4cc43d8fb620f621d63b7de37a31 relating to the session pool.

For context, the changes to that file came out of #4970 when I was trying to run the existing ITs against the emulator. To be honest, I haven't quite gotten the whole thing to work yet even with that change.

My preference would be to revert the changes to that file and work on it separately. The highest priority is to add the emulator support to the client since there is a customer that's keen to try this out in C#.

Getting the ITs to work is a bonus so that we can report emulator issues to the dev team and perhaps have a presubmit that runs the ITs against the emulator. But this will require a bunch of work from my side and from the emulator team to fix some of the failing tests.

@jskeet I would be happy for you to try out a fix for the session pool issue. I'm not familiar with it enough to know what to change. But only if you have time. Otherwise, I'll work on it until I figure it out.

I don't know if the emulator team is planning to fix the concurrent transaction issue before GA. I will check with them and get back to you.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilder.cs
```diff
@@ -46,10 +46,9 @@ public new EmulatorDetection EmulatorDetection
         /// May return a builder that will connect to the emulator under certain conditions.
         /// </summary>
         /// <remarks>
-        /// Returns null if <see cref="EmulatorDetection"/> is <see cref="EmulatorDetection.None"/>
-        /// or <see cref="EmulatorDetection.ProductionOnly"/>. Otherwise, returns a builder to
-        /// connect to the emulator using the endpoint specified by the emulator host environment
-        /// variable, and insecure credentials.
+        /// It returns null if the emulator should not be used, i.e. if <see cref="EmulatorDetection"/> is
```
jskeet, line 49: Nit: change the `i.e.` to `e.g.`, as there are other cases where the emulator shouldn't be used. But otherwise, this is fine, thanks.

skuruppu:
All done, thanks for the review feedback @jskeet.

PR #4916, opened 2020-04-29
First pass at service YAML parsing

jskeet:
This is all in aid of longer term API automation. For now, it allows
us to see which APIs we haven't generated - and I'm hoping it won't
be too hard to add an "add" command to Release Manager to add a new
API just based on the proto path. (I'd like to eventually remove a
bunch of our information from the API catalog, taking it from this
metadata instead.)

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -1549,5 +1549,12 @@
         "Grpc.Core": "2.27.0"
       }
     }
+  ],
+  "ignoredPaths": [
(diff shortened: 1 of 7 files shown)


jskeet:
(This is basically prototype code for my own use - I just wanted to get it into source control and off "just my local machine")

jskeet:
Will merge on green - apis.json formatting issues.

PR #4918, opened 2020-04-29
Consume library-related options in ServiceDirectory

jskeet:
... and then use them when adding a new API. So we only need the ID.
If things are wrong, that's usually because options are missing - so
let's print out what we *do* have that's similar.

### DIFF
--- a/tools/Google.Cloud.Tools.MetadataGenerator/ServiceDirectory.cs
+++ b/tools/Google.Cloud.Tools.MetadataGenerator/ServiceDirectory.cs
@@ -17,6 +17,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Text.RegularExpressions;
 
 namespace Google.Cloud.Tools.MetadataGenerator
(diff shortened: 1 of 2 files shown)


jskeet:
Sample output (Google.Cloud.BigQuery.Connection.V1 has a C# namespace option; Google.Cloud.BigQuery.Reservation.V1 doesn't.):

```text
$ ./prepare-release.sh add Google.Cloud.BigQuery.Reservation.V1
Configuration error: No service found for 'Google.Cloud.BigQuery.Reservation.V1'.
Similar possibilities (check options?): Google.Cloud.Bigquery.Reservation.V1

$ ./prepare-release.sh add Google.Cloud.BigQuery.Connection.V1
Added Google.Cloud.BigQuery.Connection.V1 to the API catalog with the following configuration:
{
  "id": "Google.Cloud.BigQuery.Connection.V1",
  "version": "1.0.0-beta00",
  "type": "grpc",
  "productName": "BigQuery Connection API",
  "description": "Allows users to manage BigQuery connections to external data sources.",
  "tags": [],
  "dependencies": {
    "Google.Cloud.Iam.V1": "2.0.0"
  },
  "generator": "micro",
  "protoPath": "google/cloud/bigquery/connection/v1"
}
```

tools/Google.Cloud.Tools.MetadataGenerator/ServiceDirectory.cs
```diff
@@ -119,20 +134,31 @@ public static Service FromServiceConfig(string googleapisRoot, ServiceConfig con
...
-            private static readonly char[] ImportTrimChars = { ' ', '"', ';'};
+            private static readonly HashSet<string> LibraryOptions = new HashSet<string>
+            {
+                "csharp_namespace",
+                "go_package",
+                "java_package",
+                "php_namespace",
+                "ruby_package"
+            };
+            private static readonly char[] ImportAndOptionaValueTrimChars = { ' ', '"', ';'};
```
amanda-tarafa, line 152: Nit: Optiona

jskeet: Whoops, thanks. Fixing.

PR #4921, opened 2020-04-30
docs: add devsite template

tbpg:
Eventually, we should probably move the other template into the `templates` directory. But, I didn't move it for now to keep this PR small.


jskeet:
Thanks - will start experimenting with generating a second docfx,json file.

PR #4923, opened 2020-05-01
Tool to suggest smoke tests (in JSON format)

jskeet:
This will make it easier to have confidence in newly-introduced APIs

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/SuggestSmokeTestsCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/SuggestSmokeTestsCommand.cs
@@ -1,0 +1,164 @@
+﻿// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.Tools.Common;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+using Newtonsoft.Json.Serialization;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Tools.ReleaseManager
+{
+    /// <summary>
+    /// Command that looks at all the RPCs in a client library and considers which ones might be suitable for smoke tests.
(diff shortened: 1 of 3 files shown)


tools/Google.Cloud.Tools.ReleaseManager/SuggestSmokeTestsCommand.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Tools.ReleaseManager
+{
+    /// <summary>
+    /// Command that looks at all the RPCs in a client library and considers which ones might be suitable for smoke tests.
+    /// (This is currently very simplistic, but can become more sophisticated over time.)
+    /// The smoke tests are not written to disk automatically, as they should definitely be checked by hand. For example,
+    /// the Vision API's ProductSearchClient has a PurgeProducts parameterless method - we can't easily tell that that's
+    /// dangerous whereas other parameterless RPCs are safe. (A really sophisticated version of this tool might examine
```
chrisdunelm, line 33: You have an amusingly low bar for "really sophisticated" :)

tools/Google.Cloud.Tools.ReleaseManager/SuggestSmokeTestsCommand.cs
```diff
@@ -0,0 +1,164 @@
...
+                }
+                // If there's a method that accepts just a ProjectName, find its equivalent accepting a string. (There should be exactly one.)
+                if (Overloads.Any(o => o.HasSingleParameterOfType("Google.Api.Gax.ResourceNames.ProjectName")))
+                {
+                    var stringParameter = Overloads.Single(o => o.HasSingleParameterOfType(typeof(string))).RequiredParameters[0];
+                    candidate.Arguments[stringParameter.Name] = "projects/${PROJECT_ID}";
+                    return candidate;
+                }
+
+                // If there's a method that accepts just a LocationName, find its equivalent accepting a string. (There should be exactly one.)
```
chrisdunelm, line 134: Serious errors here, which need fixing before any further progress can be made in this PR.
There should not be a full stop before the open parenthesis, and the first character in the parenthesis should not be capitalized.  The full stop within the parenthesis should then be moved out of the parenthesis.
Alternatively consider not using parenthesis at all; I enjoy using a semicolon for sentence structures such as this - or possibly a hyphen.

tools/Google.Cloud.Tools.ReleaseManager/SuggestSmokeTestsCommand.cs
jskeet: Maybe "complicated" in this case - we'd have to scan the protos to find the right RPC, then parse the HTTP bindings, which we don't currently use.
A less complex but really, really ugly approach would be to construct the default settings for the RPC and see if it's retryable. Assume we don't want to do anything that can't be retried...

tools/Google.Cloud.Tools.ReleaseManager/SuggestSmokeTestsCommand.cs
jskeet: https://www.youtube.com/watch?v=kfVsfOSbJY0

PR #4924, opened 2020-05-01
Tooling reorg

jskeet:
(As described in internal proposal document.)

Google.Cloud.Tools.sln ends up with 12 projects instead of the current 27.


jskeet:
Post-generation fix-up code was still using netcoreapp2.1, which now won't work with the SourceManipulation library. Updated and push an extra commit.

PR #4929, opened 2020-05-05
Fix the autorelease reporter script

jskeet:


jskeet:
Once this is merged, I'll do a release for Bigtable, which needs one.

PR #4931, opened 2020-05-05
Change Bigtable dependency on Bigtable.Common to be a NuGet dependency

jskeet:
(Bigtable.V2 depends on Bigtable.Admin.V2 for tests, so they need to be
consistent in terms of either both depending on Bigtable.Common.V2
via a NuGet reference or via a project reference.)


jskeet:
(Investigating the AppVeyor failure)

PR #4935, opened 2020-05-05
Automate devsite upload

jskeet:


.kokoro/release.sh
```diff
@@ -60,10 +60,10 @@ fi
 # Push documentation, if we've got the right credentials and haven't been asked to skip
 if [[ -f $DOCS_CREDENTIALS && $SKIP_GOOGLEAPISDEV_UPLOAD = "" ]]
 then
-  echo "Uploading documentation to googleapis.dev"
-  ./uploaddocs.sh releasebuild/nuget releasebuild/docs/output $DOCS_CREDENTIALS docs-staging
+  echo "Uploading documentation to googleapis.dev and devsite"
+  ./uploaddocs.sh releasebuild/nuget releasebuild/docs/output $DOCS_CREDENTIALS docs-staging docs-staging-v2
```
tbpg, line 64:
```suggestion
  ./uploaddocs.sh releasebuild/nuget releasebuild/docs/output $DOCS_CREDENTIALS docs-staging docs-staging-v2-staging
```

Want to keep the prod bucket for the final version of docs (couldn't think of a better name).

uploaddocs.sh
```diff
@@ -59,12 +60,19 @@ do
...
+    echo "Generating metadata (googleapis.dev)"
     python -m docuploader create-metadata --name $pkg --version $version --language dotnet --github-repository googleapis/google-cloud-dotnet
     
-    echo "Final upload stage"
-    python -m docuploader upload . --credentials $SERVICE_ACCOUNT_JSON --staging-bucket $STAGING_BUCKET
+    echo "Final upload stage (googleapis.dev)"
+    python -m docuploader upload . --credentials $SERVICE_ACCOUNT_JSON --staging-bucket $GOOGLEAPIS_DEV_STAGING_BUCKET
+
+    cd ../devsite/api
+    echo "Generating metadata (googleapis.dev)"
```
tbpg, line 70:
```suggestion
    echo "Generating metadata (Devsite)"
```

uploaddocs.sh
```diff
@@ -59,12 +60,19 @@ do
...
+    python -m docuploader create-metadata --name $pkg --version $version --language dotnet --github-repository googleapis/google-cloud-dotnet
     
+    echo "Final upload stage (googleapis.dev)"
```
tbpg, line 73:
```suggestion
    echo "Final upload stage (Devsite)"
```

docs/builddocs.sh
```diff
@@ -49,7 +49,7 @@ build_api_docs() {
   # Then build for devsite
   $DOCFX build --logLevel Warning --disableGitFeatures output/$api/devsite-docfx.json | tee errors.txt | grep -v "Invalid file link"
   (! grep --quiet 'Build failed.' errors.txt)
-  mv output/$api/devsite/toc.html output/$api/devsite/_toc.yaml
+  mv output/$api/devsite/api/toc.html output/$api/devsite/api/_toc.yaml
```
tbpg, line 52: How difficult would it be to change the output directory to include the API name referenced in `_rootPath`?

.kokoro/release.sh
jskeet: Whoops - got confused with the double staging.

uploaddocs.sh
jskeet: Doh! Thank you.

docs/builddocs.sh
jskeet: The output directory is irrelevant here, given that we're going *into* that directory to perform the upload. If you want to be "above" the output directory when we do the upload, I think we'll need to do more work - we'll want to plumb it through a little more carefully.

jskeet:
Changes requested are now force-pushed into the existing commits - PTAL.

If we can keep it simple and upload from the api directory for right now, I think that would be useful - we can then work out what needs to happen about directory names etc.

PR #4947, opened 2020-05-06
Create CODEOWNERS

czahedi:
@crwilcox requested teams for repo access across the board for SoDa teams. Adding a CODEOWNERS file with yoshi-dotnet as default and with the storage team.

The style for dotnet repo names is a bit different than the others so I may have done this wrong. If the wildcard won't do what I need to and I need a repo per client, LMK and I'll adjust :)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -9,5 +9,6 @@
 # explicitly taken by someone else.
 *                               @googleapis/yoshi-dotnet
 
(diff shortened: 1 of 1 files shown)


jskeet:
So is the expectation that we *need* a review from the Firestore/Bigtable team as well as from other members of yoshi-dotnet, or is it just "one or other is fine"?

crwilcox:
The current repo configuration shouldn't require an owner review. It will assign the owner as a reviewer. We could require owner review (github setting), but for now my hope was to have FYIs for changes.

PR #4966, opened 2020-05-13
Regenerate all clients with more explicit documentation

jskeet:
The use of "inheritdoc" breaks docfx in many cases. The generator
has been improved to generate more explicit documentation. This PR only
contains changes due to that generator change.


jskeet:
Chris: I suggest you hold off reviewing until the change detection has completed, at which point it should say "identical" in which case we're probably fine :)

chrisdunelm:
sgtm, I'll wait

jskeet:
Done: https://api.travis-ci.org/v3/job/686590926/log.txt
(The non-raw log eats some lines.)

PR #4972, opened 2020-05-14
Fix dependency calculations for docfx generation

### LINKED ISSUE #4965: Improve dependency checking for APIs
Our [Firestore API docs](https://googleapis.dev/dotnet/Google.Cloud.Firestore/latest/api/Google.Cloud.Firestore.FirestoreDbBuilder.html) still don't include Google.Api.Gax.Grpc.
I suspect that's because they depend on Firestore.V1. We should handle transitive dependencies properly. Shouldn't be too hard.

jskeet:
Fixes #4965.

We really need to work on getting everything present in
googleapis.dev and using xrefmaps at some point, but this is
definitely better than what we've got now.


jskeet:
(Tested by building the Google.Cloud.Firestore docs. There's much more in there now!)

PR #4983, opened 2020-05-21
feat: add SpannerNumeric class to support NUMERIC

skuruppu:
`SpannerNumeric` is a fork of [BigQueryNumeric](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryNumeric.cs).

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs
@@ -1,0 +1,482 @@
+﻿// Copyright 2020 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using System;
+using System.Globalization;
+using System.Linq;
+using System.Numerics;
+using System.Text.RegularExpressions;
+
+namespace Google.Cloud.Spanner.V1
+{
+    // TODO: Implement IFormattable?
+
+    /// <summary>
+    /// Representation of the Spanner NUMERIC type, which has 38 digits of precision,
+    /// and a fixed scale of 9 decimal places to the right of the decimal point
+    /// </summary>
+    public struct SpannerNumeric : IEquatable<SpannerNumeric>, IComparable<SpannerNumeric>, IComparable
+    {
+        private static readonly BigInteger s_maxValue = BigInteger.Parse("99999999999999999999999999999999999999");
+        private static readonly BigInteger s_minValue = -s_maxValue;
+
+        private static readonly BigInteger[] s_powersOf10 = Enumerable.Range(0, 28).Select(p => BigInteger.Pow(10, p)).ToArray();
+
+        private static readonly BigInteger s_integerScalingFactor = new BigInteger(1_000_000_000L);
+        // TODO: Don't require a 0 before the decimal point.
+        // TODO: Replace with manual validation if we find this is a performance bottleneck (as it was with field name validation).
+        private static readonly Regex s_validation = new Regex(@"^-?[0-9]+\.?[0-9]*$");
+
+        // Note: the following properties must be declared *after* s_maxValue and s_minValue. Initialization order matters.
+
+        /// <summary>
+        /// Zero represented as a <see cref="SpannerNumeric"/>. This is the default value for the type.
+        /// </summary>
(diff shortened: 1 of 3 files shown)


skuruppu:
I wanted to preserve the commit history and branch the file. But I had no idea how to convince git of my intentions.

jskeet:
I don't think Git has the concept of branching a file. (It knows about renames.) It's fine though :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Numeric.cs
```diff
@@ -0,0 +1,482 @@
...
+
+namespace Google.Cloud.Spanner.V1
+{
+    // TODO: Implement IFormattable?
+
+    /// <summary>
+    /// Representation of the Spanner NUMERIC type, which has 38 digits of precision,
+    /// and a fixed scale of 9 decimal places to the right of the decimal point
+    /// </summary>
+    public struct Numeric : IEquatable<Numeric>, IComparable<Numeric>, IComparable
```
jskeet, line 30: I'm nervous of *just* calling this Numeric. I think I'd prefer to call it `SpannerNumeric`. This fits in with the pattern of [SqlDecimal](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqltypes.sqldecimal?view=dotnet-plat-ext-3.1) etc.

@amanda-tarafa do you have thoughts either way?

jskeet:
The AppVeyor failure is due to the project file format. If you just run `./generateprojects.sh` from the root directory, that should fix it. Let me know if there are further issues.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Numeric.cs
skuruppu: I should've explained my reasoning for this. I thought that if we were to ever move this class to a common place and share it between Spanner and BigQuery then it might be easier if we call it Numeric now. But I don't know if that'll ever happen.

jskeet: If it ever happened, it would be a breaking change anyway - and arguably having a Spanner-specific name means that a general "Numeric" type can be *added* without introducing naming collisions, allowing an easier migration path.

skuruppu: Ok makes sense. I'm happy for this to be called SpannerNumeric. Probably won't get around to doing this till next week though.

In the meantime, if @amanda-tarafa has any thoughts, please chime in.

amanda-tarafa: I agree that for the meantime it makes more sense to call it SpannerNumeric, it goes well with SpannerStruct and with the more general SpannerDbType.
I haven't looked at this PR but as far as I can see this is just adding the numeric class, right? The conversion hooks are yet to be added? I might be misunderstanding something, so take this with a pinch of salt.

skuruppu: I found some time and renamed it to SpannerNumeric.

@amanda-tarafa, I'm just adding the class in this PR. I'll send a separate PR that adds the conversion hooks.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs
```diff
@@ -0,0 +1,482 @@
...
+    public struct SpannerNumeric : IEquatable<SpannerNumeric>, IComparable<SpannerNumeric>, IComparable
+    {
+        private static readonly BigInteger s_maxValue = BigInteger.Parse("99999999999999999999999999999999999999");
+        private static readonly BigInteger s_minValue = -s_maxValue;
+
+        private static readonly BigInteger[] s_powersOf10 = Enumerable.Range(0, 28).Select(p => BigInteger.Pow(10, p)).ToArray();
+
+        private static readonly BigInteger s_integerScalingFactor = new BigInteger(1_000_000_000L);
+        // TODO: Don't require a 0 before the decimal point.
+        // TODO: Replace with manual validation if we find this is a performance bottleneck (as it was with field name validation).
```
amanda-tarafa, line 39: This is something that happened in BigQuery, see #4976. It's OK to leave this TODO here, but reference the PR and change the comment to something like 

> (as it was with BigQuery field name valitions see #4976).

jskeet:
Not sure whether it's worth having the benchmark project for Spanner while the code is the same for both BigQuery and Spanner, to be honest. It's just one more thing to update. It would make more sense if they diverged

amanda-tarafa:
Yes, as long as we don't forget if they ever diverge that'd be fine, not having the benchmarks here.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs
skuruppu: All done.

jskeet:
(I merged as I wasn't sure whether @skuruppu would have access to.)

PR #4996, opened 2020-05-28
Testable emulator detection

jskeet:
This builds on #4995. Only the second commit is new.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
@@ -303,7 +303,7 @@ namespace Google.Cloud.Spanner.Data
             get
             {
                 if (TryGetValue(EmulatorDetectionKeyword, out object value) &&
-                    (value is EmulatorDetection parsed || (value is string text && EmulatorDetection.TryParse(text, out parsed))))
+                    (value is EmulatorDetection parsed || (value is string text && Enum.TryParse(text, out parsed))))
                 {
                     return parsed >= EmulatorDetection.None && parsed <= EmulatorDetection.EmulatorOrProduction ? parsed : EmulatorDetection.None;
                 }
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -303,15 +303,15 @@ public EmulatorDetection EmulatorDetection
...
+                    (value is EmulatorDetection parsed || (value is string text && Enum.TryParse(text, out parsed))))
                 {
                     return parsed >= EmulatorDetection.None && parsed <= EmulatorDetection.EmulatorOrProduction ? parsed : EmulatorDetection.None;
                 }
                 return EmulatorDetection.None;
             }
             set
             {
-                GaxPreconditions.CheckEnumValue((EmulatorDetection)value, "value");
+                GaxPreconditions.CheckEnumValue(value, "value");
```
amanda-tarafa, line 314: nit: nameof(value);

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -303,15 +303,15 @@ public EmulatorDetection EmulatorDetection
             get
```
amanda-tarafa, line 303: nit: get and set here are not consistent. set fails if the value is out of range but get returns None.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
jskeet: Done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
jskeet: That's the same for all the properties. For int ones for example, we have `GetInt32OrDefault`, but `SetInt32WithValidation`.

PR #5002, opened 2020-05-29
Fix integration tests when running on the emulator

jskeet:
Please review one commit at a time.

The last of these commits is the most impactful one (in that it moves from "half the tests fail" to "everything passes") but is the only one I'm worried about in terms of production code. I'm happy for it to be merged for now, and even for betas to be released with it in, but we should think really carefully before including it in a GA release.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
@@ -29,9 +29,10 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         public PartitionedReadTests(PartitionedReadTableFixture fixture) =>
             _fixture = fixture;
 
-        [Fact]
+        [SkippableFact]
         public async Task DistributedReadAsync()
         {
+            Skip.If(_fixture.RunningOnEmulator, "Emulator doesn't support distributed reads yet.");
             int numRows;
             using (var connection = _fixture.GetConnection())
             using (var cmd = connection.CreateSelectCommand($"SELECT COUNT(*) FROM {_fixture.TableName}"))
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerFixtureBase.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerFixtureBase.cs
@@ -33,6 +33,7 @@ namespace Google.Cloud.Spanner.Data.CommonTesting
         public DatabaseName DatabaseName => Database.DatabaseName;
         public SpannerConnection GetConnection() => Database.GetConnection();
         public string ConnectionString => Database.ConnectionString;
+        public bool RunningOnEmulator => SpannerClientCreationOptions.UsesEmulator;
         internal SpannerClientCreationOptions SpannerClientCreationOptions => Database.SpannerClientCreationOptions;
     }
 }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AdminTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AdminTests.cs
@@ -193,7 +193,7 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
 
             using (var connection = new SpannerConnection(_fixture.Database.NoDbConnectionString))
             {
-                var createCmd = connection.CreateDdlCommand($"CREATE DATABASE BadName");
+                var createCmd = connection.CreateDdlCommand($"CREATE DATABASE Bad-Name");
                 await Assert.ThrowsAsync<SpannerException>(() => createCmd.ExecuteNonQueryAsync());
                 createCmd = connection.CreateDdlCommand($"CREATE DATABASE {dbName}");
                 await createCmd.ExecuteNonQueryAsync();
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BatchDmlTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BatchDmlTests.cs
@@ -266,9 +266,10 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
             Assert.Equal(expected, actual);
         }
 
-        [Fact]
+        [SkippableFact]
         public void InvalidDmlBatch_PartialSuccess_ImplicitTransaction()
         {
+            Skip.If(_fixture.RunningOnEmulator, "Leaks a transaction, causing all further tests to fail.");
             string key = _fixture.CreateTestRows();
 
             SpannerBatchNonQueryException exception;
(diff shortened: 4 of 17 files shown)


jskeet:
Whoops - will fix the broken unit test.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
```diff
@@ -95,22 +89,20 @@ async Task<IEnumerable<long>> Impl()
...
-                    return count;
-                }
+                // Importantly, we need to set timeout on the transaction, because
+                // ExecuteMutations on SpannerTransaction doesnt actually hit the network
+                // until you commit or rollback.
+                transaction.CommitTimeout = timeoutSeconds;
+                int count = await ((ISpannerTransaction)transaction)
+                    .ExecuteMutationsAsync(mutations, cancellationToken, timeoutSeconds)
+                    .ConfigureAwait(false);
+                rollbackCheckpoint();
```
amanda-tarafa, line 103: Should this be after the commit? What state is the transaction left in if the commit fails? And it doesn't hurt us to put this after the commit anyway (I think). And same for the others except dml which is of its own.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
```diff
@@ -38,48 +39,41 @@ internal EphemeralTransaction(SpannerConnection connection, TransactionOptions t
...
         public Task<long> ExecuteDmlAsync(ExecuteSqlRequest request, CancellationToken cancellationToken, int timeoutSeconds)
         {
-            return ExecuteHelper.WithErrorTranslationAndProfiling(Impl, "EphemeralTransaction.ExecuteDmlAsync", _connection.Logger);
+            return ExecuteInReadWriteTransactionWithErrorTranslationAndProfiling(Impl, cancellationToken, "EphemeralTransaction.ExecuteDmlAsync");
 
-            async Task<long> Impl()
+            async Task<long> Impl(SpannerTransaction transaction, Action rollbackCheckpoint)
             {
-                using (var transaction = await _connection.BeginTransactionImplAsync(_transactionOptions, TransactionMode.ReadWrite, cancellationToken).ConfigureAwait(false))
+                // TODO: Validate this. Do we really never need to roll back the transaction here?
```
amanda-tarafa, line 46: Might depend on partioned vs. non-partioned?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
```diff
@@ -127,5 +119,45 @@ async Task<ReliableStreamReader> Impl()
...
+                using (var transaction = await _connection.BeginTransactionImplAsync(_transactionOptions, TransactionMode.ReadWrite, cancellationToken).ConfigureAwait(false))
+                {
+                    bool rollback = true;
+                    try
+                    {
+                        return await taskProvider(transaction, () => rollback = false).ConfigureAwait(false);
+                    }
+                    finally
+                    {
+                        if (rollback)
```
amanda-tarafa, line 142: Why not simply catch any exception and _attempt_ a rollback? Is it because of partioned-dml? What happens if we rollback that transaction? I'd say it would rollback to the last commit which is fine?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AssemblyInfo.cs
```diff
@@ -11,9 +11,10 @@
...
 // See the License for the specific language governing permissions and
 // limitations under the License.
+using Google.Cloud.ClientTesting;
 using Xunit;
 
 [assembly: CollectionBehavior(DisableTestParallelization = true)]
 // Uncomment these to run the tests in a predictable order. (By fixture name, then by test case within the fixture.)
-// [assembly: TestCollectionOrderer(AlphabeticalOrderer.TypeName, AlphabeticalOrderer.AssemblyName)]
-// [assembly: TestCaseOrderer(AlphabeticalOrderer.TypeName, AlphabeticalOrderer.AssemblyName)]
+[assembly: TestCollectionOrderer(AlphabeticalOrderer.TypeName, AlphabeticalOrderer.AssemblyName)]
```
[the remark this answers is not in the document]
jskeet: Okay. We can just remove the commit.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
jskeet: Maybe. It's frankly unclear to me.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
jskeet: No, it should be before the commit - because I believe if *commit* fails, then it's automatically rolled back and we don't need to call rollback. Although I suppose there could be a communication problem with means the reason the commit fails is that the server never gets the message.

I'll leave it for now, and we'll see what the Spanner team thinks we should do.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
jskeet: It's because I believe we shouldn't call Rollback if we've already called Commit - that a failed Commit already rolls back.
But let's wait to hear more.

jskeet:
Hmm... I'm now seeing the session label test failing - it's finding 160-165 sessions with the specified label instead of 100. I need to dig into that.

jskeet:
@skuruppu: It looks like the emulator isn't actually filtering as requested on session labels. Do you know anything about this as a limitation? (I'm surprised the tests didn't fail earlier in the day...)

jskeet:
Okay, the session label filtering is a known limitation of the emulator. I'm skipping that test. I've also fixed a bad test which was trying to use "creating a database with a name of BadName" as an example of DDL that would fail. It fails *after the first success per instance* but that's not really good enough. Have changed it to use Bad-Name instead, which is invalid due to using a hyphen.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
```diff
@@ -29,9 +29,10 @@ public class PartitionedReadTests
         public PartitionedReadTests(PartitionedReadTableFixture fixture) =>
             _fixture = fixture;
 
-        [Fact]
+        [SkippableFact]
         public async Task DistributedReadAsync()
         {
+            Skip.If(_fixture.RunningOnEmulator, "Emulator doesn't support distributed reads yet.");
```
skuruppu, line 35: Just as an FYI, this test does pass if you use the daily build of the emulator but I gather that this is not available through gcloud yet.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
skuruppu: Oooh this is something I'm not confident to comment on. I will have to defer to @snehashah16, especially given that this is something that the emulator is expecting but is not an issue on prod.

jskeet:
Okay, I've now:

- Removed the previous "last commit" that changed EphemeralTransaction
- Added another commit to skip BatchDmlTest.InvalidDmlBatch_PartialSuccess_ExplicitTransaction, which leaks a transaction
- Raised #5021 to consider the best fix for leaking transactions
- Added a final commit to skip a Snippet test (so the snippet suite now passes on the emulator too)

The only changes in production code are now:

- Working around the emulator returning a ExecuteBatchDmlResponse without a status, in SpannerTransaction
- Reverting the previous SessionPoolManager change to use special options for the emulator

jskeet:
Will merge on green - thanks!

PR #5023, opened 2020-06-03
Delete AspNetCore diagnostics analyzers and AspNet diagnostics

### LINKED ISSUE #4891: Reevaluate ASP.NET Core analyzer value
The analyzer complicates project generation - we should check that it still adds value. (It may well be that as the installation approach has changed, many of the things it guards against are irrelevant.)

Even if we keep it, we could potentially not generate the project file - or just generate certain parts of it (e.g. version and copyright), making the rest manual.

jskeet:
Fixes #4891


jskeet:
The Travis failure is the normal "trying to build a deleted API" issue.

PR #5043, opened 2020-06-08
Draft of changes for transactions

jskeet:
This assumes that we can always roll back transactions in a purely fire-and-forget manner

No tests yet, but thought I'd get feedback on the general shape.

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
@@ -82,6 +82,7 @@ namespace Google.Cloud.Spanner.V1
 
         private readonly SessionPool.ISessionPool _pool;
         private int _disposed;
+        private int _committedOrRolledBack;
 
         private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, ByteString transactionId, ModeOneofCase transactionMode, DateTime evictionTime, long refreshTicks)
         {
@@ -193,6 +194,14 @@ namespace Google.Cloud.Spanner.V1
         private bool MarkAsDisposed() => Interlocked.Exchange(ref _disposed, 1) != 1;
 
         /// <summary>
+        /// Remembers that the transaction in this session has been successfully commmitted or rolled back.
+        /// Sessions that are disposed but still have a
+        /// </summary>
+        private void MarkAsCommittedOrRolledBack() => Interlocked.Exchange(ref _committedOrRolledBack, 1);
+
+        private bool IsCommittedOrRolledBack() => Interlocked.CompareExchange(ref _committedOrRolledBack, 0, 0) == 1;
+
+        /// <summary>
         /// Returns this session to the session pool from which it was acquired, unless
         /// it has become invalid. This method should only be called once per instance; subsequent
         /// calls are ignored. No other methods can be called after this.
@@ -202,7 +211,9 @@ namespace Google.Cloud.Spanner.V1
         {
             if (MarkAsDisposed())
             {
-                _pool.Release(AfterReset(), forceDelete || ServerExpired || ShouldBeEvicted);
+                // TODO: What about Partitioned DML, which implicitly commits as it goes?
+                ByteString transactionToRollback = IsCommittedOrRolledBack() ? null : TransactionId;
+                _pool.Release(AfterReset(), transactionToRollback, forceDelete || ServerExpired || ShouldBeEvicted);
             }
             else
             {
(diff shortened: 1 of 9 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -487,9 +489,11 @@ public override Task<PooledSession> WithFreshTransactionOrNewAsync(PooledSession
             /// <summary>
             /// Release a session back to the pool (or refresh or evict it) and decrement the number of active sessions.
             /// </summary>
-            public override void Release(PooledSession session, bool deleteSession)
+            public override void Release(PooledSession session, ByteString transactionToRollback, bool deleteSession)
```
amanda-tarafa, line 492: Instead of receiving the transactionId here, which is always going to be the transaction associated with the session, I think I'd prefer passing the session to MaybeRollback... and checking there if the transaction has been commited or not and rollbacking otherwise. It's all in V1 so at most you have to make some properties internal instead of private. I think it'll be cleaner and we avoid the (minimal) risk of passing the wrong transaction to be rollbacked for a given session

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -487,9 +489,11 @@ public override Task<PooledSession> WithFreshTransactionOrNewAsync(PooledSession
...
             {
                 Interlocked.Decrement(ref _activeSessionCount);
+                Parent.MaybeRollbackTransactionFireAndForget(session.SessionName, transactionToRollback);
```
amanda-tarafa, line 495: Can't we add this call in EvictSession and ReleaseInactiveSession instead. Rolling back is something that we'll want to attempt eitherway so it should be part of it (Evict or Releasing) I think.
And we don't have to remember to add the calls to MaybeRollback... whenever we Evict or Release.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
```diff
@@ -233,6 +234,20 @@ private void DeleteSessionFireAndForget(PooledSession session)
             ConsumeBackgroundTask(DeleteSessionAsync(session), "session deletion");
         }
 
+        /// <summary>
+        /// Rolls back the specified transaction, if any, in a fire-and-forget manner.
+        /// </summary>
+        /// <param name="session">The name of the session that owns the transaction.</param>
+        /// <param name="transactionID">The ID of the transaction to rollback. May be null, in which case this method is a no-op.</param>
+        private void MaybeRollbackTransactionFireAndForget(SessionName session, ByteString transactionID)
```
amanda-tarafa, line 242: Here's where I would pass the whole PooledSession, check whether we need to rollback and do it.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -202,7 +211,9 @@ public void ReleaseToPool(bool forceDelete)
         {
             if (MarkAsDisposed())
             {
-                _pool.Release(AfterReset(), forceDelete || ServerExpired || ShouldBeEvicted);
+                // TODO: What about Partitioned DML, which implicitly commits as it goes?
```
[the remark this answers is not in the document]
jskeet: No idea, I'm afraid. We'll need to ask the Spanner team.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
jskeet: We don't actually have the transaction ID here in the case of PooledSession.ReleaseToPool - the session is already a fresh one without a transaction ID. We could change that of course, but I figured this was the minimally-disruptive change.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
jskeet: We don't always. I originally put it in there, but there are cases where we call ReleaseInactiveSession but don't want to roll anything back... e.g. because we've just allocated the transaction in TryCreateReadWriteTransactionAndReturnToPool.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Ah true, we have called AfterReset(). Yep it's fine then.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
amanda-tarafa: Yep, fine as well then.

jskeet:
Deleting in favour of #5065.

PR #5088, opened 2020-06-15
Fixes #5028: Adds options to specify service context values for GoogleLogger.

### LINKED ISSUE #5028: Support ability to specify ServiceContext in GoogleLogger
Hi,

I have recently added Google.Cloud.Diagnostics.AspNetCore logging into an existing application running in GKE (and it's working great).

My assumption is that I don't need to use the library's error reporting capabilities, as all errors that occur within my application are caught and logged with ILogger<T>.LogError(exception, message). These appear properly in cloud console.

Google Cloud Error Reporting automatically picks up these errors from the logs. However, in cloud console, all errors are reported as coming from "gke_instance", which does not help identify where the errors are coming from.  There is an [existing discussion](https://issuetracker.google.com/issues/62828996) about using resource labels from logs to better identify services, but 3 years on it is not supported yet.

I believe we can provide this information to error reporting by adding serviceContext to the JSON payload of log entries, however GoogleLogger does not expose any capabilities that permit configuring this.

I confess I have not used this library for long, and I might be missing something. But if not, could we potentially add support similar to ILogEntryLabelProvider, but for the JSON payload?

Cheers!

amanda-tarafa:


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
```diff
@@ -103,14 +119,87 @@ public sealed class LoggerOptions
             MonitoredResource monitoredResource = null,
             BufferOptions bufferOptions = null,
             RetryOptions retryOptions = null,
-            TextWriter loggerDiagnosticsOutput = null)
+            TextWriter loggerDiagnosticsOutput = null) =>
+            CreateWithServiceContext(logLevel, logName, labels, monitoredResource, bufferOptions, retryOptions, loggerDiagnosticsOutput);
```
jskeet, line 123: Why do we not want to use CreateWithDefaultServiceContext here?

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
```diff
@@ -103,14 +119,87 @@ public sealed class LoggerOptions
...
+        /// <param name="bufferOptions">Optional, the buffer options.  Defaults to a <see cref="BufferType.Timed"/></param>
+        /// <param name="retryOptions">Optional, the retry options.  Defaults to a <see cref="RetryType.None"/></param>
+        /// <param name="loggerDiagnosticsOutput">Optional. If set some logger diagnostics info will be written
+        /// to the given <see cref="TextWriter"/>. Currently the only diagnostics info we provide is the URL where
+        /// the logs written with these options can be found.</param>
+        /// <param name="serviceName">A string that represents the version of the service or the source code that logs are written for.
+        /// When set, it will be included in the <code>serviceContext</code> field of the log entry JSON payload.</param>
+        /// <param name="version">A string that represents the version of the service or the source code that logs are written for.
+        /// When set, it will be included in the <code>serviceContext</code> field of the log entry JSON payload.</param>
+        public static LoggerOptions CreateWithServiceContext(
```
jskeet, line 186: I was expecting the user to be able to supply a `Struct` as the complete service context here. Is that something that might be useful? Can the service context have more than the name and version? I think it's worth working that out during this PR, so we don't end up with more overloads than we want later on.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
```diff
@@ -89,6 +94,17 @@ public sealed class GoogleLogger : ILogger
             _fullLogName = logTarget.GetFullLogName(_loggerOptions.LogName);
             _serviceProvider = serviceProvider;
             _clock = clock ?? SystemClock.Instance;
+            // Create the service context here, it's going to be the same for all entries.
+            if (_loggerOptions.ServiceName != null)
```
jskeet, line 98: Perhaps `LoggerOptions` should have a `CreateServiceContext` or `GetServiceContext` method? (That would help to simplify things if we allow complete customization too.)

amanda-tarafa: OK, I reshuffled the code a little bit. Squashed everything in the original commit so you can review as a whole.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
jskeet: Yup, that works for me.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/AssemblyInfo.cs
```diff
@@ -14,6 +14,7 @@
 using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("DynamicProxyGenAssembly2,PublicKey=0024000004800000940000000602000000240000525341310004000001000100c547cac37abd99c8db225ef2f6c8a3602f3b3606cc9891605d02baa56104f4cfc0734aa39b93bf7852f7d9266654753cc297e7d2edfe0bac1cdcf9f717241550e0a7b191195b7667bb4f64bcb8e2121380fd1d9d46ad2d92d2d15605093924cceaf74c4861eff62abf69b9291ed0a340e113be11e6a7d3113e92484cf7045cc7")]
+[assembly: InternalsVisibleTo("Google.Cloud.Diagnostics.AspNetCore,PublicKey=0024000004800000940000000602000000240000525341310004000001000100afab79952ee22215f12b4e09337e65509c943fbc22d7006bc371d581d0f0ebf0da5d8039aab2607fb68a138a5d80a71bc02b7ebf586dbe1f2493c0ab20423ababfd15ce74d2264a6b37745f3658f016abaad662182aaef634a60f1346fcc45343acab5b6781535a3134818e13fac895a6c106c0480e34bbb06cb123e5583d8d2")]
```
jskeet, line 17: We shouldn't do this - it breaks versioning horribly.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/CommonUtils.cs
```diff
@@ -37,5 +40,21 @@ internal static class CommonUtils
                 .Assembly
                 .GetCustomAttribute<AssemblyInformationalVersionAttribute>()
                 .InformationalVersion;
+
+        internal static Struct CreateServiceContext(string serviceName, string version)
```
jskeet, line 44: Let's make this public. Sure, most other code won't use it - but it's not *dangerous* for it to be public, and it avoids the versioning issue.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
```diff
@@ -296,6 +296,90 @@ public async Task Logging_NoServiceProvider()
...
+            using (var client = server.CreateClient())
+            {
+                await client.GetAsync($"/Main/Warning/{testId}");
+            }
+
+            _fixture.AddValidator(testId, results =>
+            {
+                var jsonFields = results.Single().JsonPayload.Fields;
+                Assert.Equal("my.test.service", jsonFields["serviceContext"].StructValue.Fields["service"].StringValue);
+                Assert.Equal("v1.0.0-alpha01", jsonFields["serviceContext"].StructValue.Fields["version"].StringValue);
```
jskeet, line 317: Just as an aside: if we ever get the Struct.GetValue extension method into Google.Protobuf, this would be:

```csharp
var json = results.Single().JsonPayload;
Assert.Equal("my.test.service", json.GetValue("serviceContext/service");
Assert.Equal("v1.0.0-alpha01", json.GetValue("serviceContext/version");
```

Wouldn't that be nice? :)

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
jskeet: Except that "default" doesn't actually work as a name, given that it's not what happens by default...

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/LoggerOptions.cs
```diff
@@ -103,14 +126,87 @@ public sealed class LoggerOptions
...
+        /// on the value of <see cref="RetryOptions.ExceptionHandling"/>.</param>
+        /// <param name="monitoredResource">Optional, the monitored resource.  The monitored resource will
+        /// be automatically detected if it is not set and will default to the global resource if the detection fails.
+        /// See: https://cloud.google.com/logging/docs/api/v2/resource-list </param>
+        /// <param name="bufferOptions">Optional, the buffer options.  Defaults to a <see cref="BufferType.Timed"/></param>
+        /// <param name="retryOptions">Optional, the retry options.  Defaults to a <see cref="RetryType.None"/></param>
+        /// <param name="loggerDiagnosticsOutput">Optional. If set some logger diagnostics info will be written
+        /// to the given <see cref="TextWriter"/>. Currently the only diagnostics info we provide is the URL where
+        /// the logs written with these options can be found.</param>
+        public static LoggerOptions CreateWithDefaultServiceContext(
```
jskeet, line 152: How about CreateWithDetectedServiceContext to make it clearer that this isn't the default behaviour?

PR #5118, opened 2020-06-26
test: run spanner ITs against the emulator

skuruppu:


skuruppu:
[Example test run](https://travis-ci.org/github/googleapis/google-cloud-dotnet/jobs/702273579).

skuruppu:
Sorry for the noise. It turns out I'm not actually running the integration tests here. I'll fix that.

skuruppu:
Based on the [output](https://travis-ci.org/github/googleapis/google-cloud-dotnet/jobs/702335185), I don't think the tests are running at all. It's too fast.

.travis.yml
```diff
@@ -39,5 +42,5 @@ branches:
 script:
   - python3 tools/listapis.py apis/apis.json > /dev/null || travis_terminate 1
   - touch build_timing_log.txt
-  - ./build.sh --diff --regex $API_REGEX
+  - if [ -z "$SPANNER_EMULATOR_HOST" ]; then ./build.sh --diff --regex $API_REGEX; else ./spanner-ITs-against-emulator.sh; fi
```
skuruppu, line 45: This is where I most need feedback. As a start, I decided not to modify `build.sh` but please let me know if it's better to add an option like `--spanner-ITs` to build.sh and then run this command.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
```diff
@@ -79,6 +83,9 @@ private SpannerTestDatabase(string projectId)
             TestLogger.Install();
 
             ProjectId = projectId;
+
+            MaybeInitializeInstance();
```
skuruppu, line 87: We currently rely on there being a `spannerintegration` instance available in the project. But if we're using the emulator, there are no default instances available and we need to create one.

skuruppu:
Example [output](https://travis-ci.org/github/googleapis/google-cloud-dotnet/jobs/703475172) from the latest run after I fixed the scripts. On my machine, it takes only a minute to run the tests but it seems to be taking considerably longer on Travis.

spanner-ITs-against-emulator.sh
```diff
@@ -0,0 +1,42 @@
...
+# Start the emulator
+./emulator_main --host_port $SPANNER_EMULATOR_HOST &
+
+EMULATOR_PID=$!
+
+# Stop the emulator & clean the environment variable
+trap "kill -15 $EMULATOR_PID; unset SPANNER_EMULATOR_HOST; unset TEST_PROJECT; echo \"Cleanup the emulator\";" EXIT
+
+cd apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests
+dotnet test
```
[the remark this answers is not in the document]
skuruppu: Ah, I didn't realise that we could also run the snippets tests. Found a skipped test here that I was able to unskip. Thanks.

spanner-ITs-against-emulator.sh
```diff
@@ -0,0 +1,42 @@
...
+tar zxvf cloud-spanner-emulator_linux_amd64-${EMULATOR_VERSION}.tar.gz
+chmod u+x emulator_main
+
+# Start the emulator
+./emulator_main --host_port $SPANNER_EMULATOR_HOST &
+
+EMULATOR_PID=$!
+
+# Stop the emulator & clean the environment variable
+trap "kill -15 $EMULATOR_PID; unset SPANNER_EMULATOR_HOST; unset TEST_PROJECT; echo \"Cleanup the emulator\";" EXIT
```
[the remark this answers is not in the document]
skuruppu: Yup, agree. I've removed this now.

spanner-ITs-against-emulator.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# Display commands being run
+set -x
+
+export SPANNER_EMULATOR_HOST=localhost:9010
+export TEST_PROJECT=emulator-test-project
+echo "Running the Cloud Spanner emulator: $SPANNER_EMULATOR_HOST";
+
+# Download the emulator
+EMULATOR_VERSION=0.8.0
```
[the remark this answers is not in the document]
skuruppu: I think they will for the GA launch. I have now changed the script to use gcloud because that was the only way that I could figure out how to create an instance through the command line.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
skuruppu: Now using gcloud in the script so I was able to create an instance on the command line. Reverted these commits.

.travis.yml
skuruppu: Ah I missed that you can configure a separate script per job. Fixed now.

skuruppu:
PTAL @jskeet, I think I got this working with minimal changes. [Output](https://travis-ci.org/github/googleapis/google-cloud-dotnet/jobs/703784604) from the latest run.

spanner-ITs-against-emulator.sh
```diff
@@ -0,0 +1,51 @@
...
+
+export SPANNER_EMULATOR_HOST=localhost:9010
+export TEST_PROJECT=emulator-test-project
+echo "Running the Cloud Spanner emulator: $SPANNER_EMULATOR_HOST";
+
+# Download the Google Cloud SDK and fetch the beta component which contains the
+# emulator.
+GOOGLE_CLOUD_SDK_VERSION=299.0.0
+wget https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${GOOGLE_CLOUD_SDK_VERSION}-linux-x86_64.tar.gz
+tar zxvf google-cloud-sdk-${GOOGLE_CLOUD_SDK_VERSION}-linux-x86_64.tar.gz
```
jskeet, line 28: Please could you remove the `v` from `zxvf` here? The tar file content is dwarfing the rest of the log file.

jskeet: (I'm assuming we do definitely want to download this version explicitly rather than just updating the version that's already installed in Travis?)

spanner-ITs-against-emulator.sh
```diff
@@ -0,0 +1,51 @@
...
+
+export CLOUDSDK_CORE_DISABLE_PROMPTS=1
+./google-cloud-sdk/bin/gcloud config set disable_usage_reporting false
```
jskeet, line 31: Why do we want to enable usage reporting for these tests?

.travis.yml
```diff
@@ -13,6 +13,8 @@ jobs:
     - name: "Change detection"
       if: type = pull_request
       script: ./detect-pr-changes.sh
+    - name: "Spanner Emulator"
+      script: ./spanner-ITs-against-emulator.sh;
```
jskeet, line 17: I will probably noodle on this filename further - I wonder whether putting it under apis/Google.Cloud.Spanner.Data would make sense. But don't worry about that for this PR.
(We'll also want to work out a way of only doing this when something about Spanner has changed, rather than making every PR take 15 minutes to run, but I can do that.)

spanner-ITs-against-emulator.sh
skuruppu: Oh, it's because otherwise there was a prompt. But I think it's fine if it defaults to Yes. I'll remove this.

spanner-ITs-against-emulator.sh
skuruppu: I tried using the already-installed gcloud but the error says:

```
ERROR: (gcloud.components.install) You cannot perform this action because this Cloud SDK installation is managed by an external package manager.
Please consider using a separate installation of the Cloud SDK created through the default mechanism described at: https://cloud.google.com/sdk/
```

So I will instead remove the `v`.

jskeet:
Will merge this now, then create a new PR to move the script (and probably get it to unpack into a tmp directory).

PR #5119, opened 2020-06-26
[apis/Google.Cloud.PubSub.V1] feat: Add support for server-side streaming pull flow control

kamalaboulhosn:


jskeet:
Chris, do you have the bandwidth to review this?

jskeet:
What's the user-visible change here? It looks like it's not a public API surface change. It looks like it's just sending more information to the server, which I guess leads to increased efficiency?
(I'm trying to work out what we'd put in release notes.)

kamalaboulhosn:
Correct, this is not user-visible. It just propagates settings that were already used in the client to the server since the StreamingPullRequest now has theses fields and can use them. For now, this is mostly about correctness for ordered delivery via ordering keys.

jskeet:
@kamalaboulhosn  Thanks. Will work out something to put in the next release notes :)

What's the sense of urgency in terms of including this in a release?

PR #5121, opened 2020-06-29
Improvements to GetOrdinal behavior

### LINKED ISSUE #5114: Using Task.Run in Google.Cloud.Spanner.Data causes remarkable delay
We are developing a server application using Google.Cloud.Spanner.Data with .NET Core 3.1.
The application server is built with ASP.NET Core MVC.

Remarkable delay in the part that uses `Task.Run` at `Google.Cloud.Spanner.Data`  has occurred when load testing sending a large number of requests.
At high load, it takes more than 10 seconds and the transaction times out.

Specifically, call `Task.Run()` in [GetOrdinal()](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs#L225) and in [PopulateMetadata()](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs#L353) calliing from `GetFieldValue()` via `GetSpannerFieldType()`.

So, when I cloned `google-cloud-dotnet` and implemented the async method of `GetOrdinal()` and `GetFieldValue()`,
The noticeable delay mentioned above has been resolved.

How about implementing the Async version for the two methods `GetOrdinal()` and `GetFieldValue()` ?

Now that I've signed the Corporate CLA, I can refine my implementation and contribute.

jskeet:
- Avoid an unnecessary call to Task.Run if we already have metadata but haven't populated the field index
- Add GetOrdinalAsync for async callers who want to avoid Task.Run entirely

Note that GetOrdinal is called by GetFieldValue<T> and this[string], which are called after Read/ReadAsync - this would previously have involved a Task.Run call for a task that completed immediately, whereas now we'll spot that we already have metadata, and complete synchronously.

Fixes #5114

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
@@ -43,7 +43,7 @@ namespace Google.Cloud.Spanner.Data
         private bool? _hasRows = null;
         private readonly List<Value> _innerList = new List<Value>();
         private readonly ReliableStreamReader _resultSet;
-        private readonly ConcurrentDictionary<string, int> _fieldIndex = new ConcurrentDictionary<string, int>();
+        private ConcurrentDictionary<string, int> _fieldIndex;
         private ResultSetMetadata _metadata;
         private readonly IDisposable _resourceToClose;
         private readonly SpannerConversionOptions _conversionOptions;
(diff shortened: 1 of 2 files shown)


jskeet:
I'm running integration tests and snippets locally, but it's looking promising so far.

This was a much nicer fix than I'd expected to be able to create :)

jskeet:
(Integration tests and snippets all passed.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -215,16 +215,52 @@ public Value GetJsonValue(int i)
...
+            {
+                await PopulateMetadataAsync(cancellationToken).ConfigureAwait(false);
+                InitializeFieldIndexFromMetadata();
+            }
+            return _fieldIndex[name];
+        }
+
+        private void InitializeFieldIndexFromMetadata()
+        {
+            _fieldIndex = new ConcurrentDictionary<string, int>();
```
amanda-tarafa, line 252: Consider moving this line inside the if. As it stands, if this method is called without having called PopulateMetada, the index collection gets initialized but not populated, and that means that it won't ever get populated.
Or if this method should always be called after having metadata populated, just throw if thre's no metadata.

jskeet: I'm not sure it should throw though. In at least some places, it looks like it's valid for there to be no metadata. We're not consistent about handling that, but I'd like to avoid failing if that's the case (when we can let the GetOrdinal method through due to an invalid key instead).

How about I just add a comment above the method saying that PopulateMetadata (or async) must already have been called?

amanda-tarafa: Yep, OK, that's fine.

PR #5124, opened 2020-06-30
Adds test for using client side signed JWT with a gRPC client.

amanda-tarafa:
This is green now and has been tested locally with googleapis/google-api-dotnet-client#1592 and it's also green. We can merge now.

It's here for a lack of a better place, any gRPC client is good enough for the test and Spanner is already peculiar in where tests live.
Ideally this test would be in googleapis/google-api-dotnet-client, but all gRPC clients depend on GAX and GAX depends on Google.Apis so we would have to copy the gRPC client and GAX over and keep it up to date.


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/V1/JwtTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        private readonly SpannerDatabaseFixture _fixture;
+
+        public JwtTests(SpannerDatabaseFixture fixture) => _fixture = fixture;
+
+        [Fact]
+        public void JwtAccessToken_GrpcClient()
+        {
+            InstanceAdminClientBuilder builder = new InstanceAdminClientBuilder
+            {
+                // Making sure we have no scopes so that a JWT is sent.
```
jskeet, line 40: Can't you just use `Scopes = new string[0]` in the builder? I believe that would achieve the same thing, but in a simpler way.

amanda-tarafa:
AppVeyor failure due to new line missing at the end of project file. Fixed. Waiting on green before merging.

PR #5136, opened 2020-07-02
Fix console output in Travis logs

jskeet:


apis/Google.Cloud.Speech.V1/Google.Cloud.Speech.V1/RecognitionAudioPartial.cs
```diff
@@ -19,6 +19,8 @@
 using System.Net.Http;
 using System.Threading.Tasks;
 
+// Not a real change
```
chrisdunelm, line 22: Is this meant to be here?

apis/Google.Cloud.Vision.V1/Google.Cloud.Vision.V1/AnnotateImageResponsePartial.cs
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// Not a real change
```
chrisdunelm, line 15: Is this meant to be here?

detect-pr-changes.sh
```diff
@@ -13,11 +13,15 @@ mkdir tmpgit/new
...
+log_header() {
+  echo "\e[1;35m$1\e[0m"
+}
+
 for api in $apis
 do  
   if [[ -d tmpgit/apis/$api/$api && -d apis/$api/$api ]]
   then
-    echo "Building $api"
+    log_header "Building $api"
```
chrisdunelm, line 24: Maybe add a comment explaining what this does?

toolversions.sh
```diff
@@ -161,6 +161,9 @@ install_docfx() {
 
 # Logs to both stdout and a build timing log, allowing
 # post-processing to see how long each part of the build takes.
+# The console log is in magenta to stand out, and doesn't include the
+# timing part.
 log_build_action() {
-  echo "$(date -u -Iseconds) $1" | tee -a $REPO_ROOT/build_timing_log.txt
+  echo -e "\e[1;35m$1\e[0m"
```
chrisdunelm, line 167: Maybe add a comment explaining what this does?

apis/Google.Cloud.Vision.V1/Google.Cloud.Vision.V1/AnnotateImageResponsePartial.cs
jskeet: Not by the time I merge, no. But I need it to be in the PR so that I can check that detect-pr-changes.sh gives the right output.

toolversions.sh
jskeet: The comment a few lines earlier says that the output is in magenta - is that not enough?

detect-pr-changes.sh
jskeet: Yup, will add it to the log_header definition.

jskeet:
(Note that I'm still trying to get this to actually run on Travis, so there may well be several changes to come...)

toolversions.sh
chrisdunelm: Ah yes, I didn't notice that :)

apis/Google.Cloud.Vision.V1/Google.Cloud.Vision.V1/AnnotateImageResponsePartial.cs
chrisdunelm: ok, thanks

PR #5139, opened 2020-07-06
Changes to Spanner integration tests for CI

jskeet:
- Move the script under Google.Cloud.Spanner.Data
- Don't do anything unless the PR contains something involving Spanner
- Unpack gcloud to a tmp directory

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/run-emulator-tests.sh
+++ b/apis/Google.Cloud.Spanner.Data/run-emulator-tests.sh
@@ -17,6 +17,19 @@
 # Fail on any error.
 set -e
 
+# Check if we need to run
+if ! git diff master --name-only | grep -iq spanner
+then
+  echo "Skipping Spanner integration tests; no Spanner-related changes."
+  exit 0
+fi
+
+declare -r REPO_ROOT=$(readlink -f $(dirname ${BASH_SOURCE})/../..)
+
+declare -r REPO_TMP=$REPO_ROOT/tmp
+rm -rf $REPO_TMP
+mkdir $REPO_TMP
+
 export SPANNER_EMULATOR_HOST=localhost:9010
 export TEST_PROJECT=emulator-test-project
 echo "Running the Cloud Spanner emulator: $SPANNER_EMULATOR_HOST";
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.Data/run-emulator-tests.sh
```diff
@@ -17,35 +17,51 @@
...
 # When this shell exits, stop the emulator.
 trap "kill -15 $EMULATOR_PID; echo \"Cleaned up the emulator\";" EXIT
 
-./google-cloud-sdk/bin/gcloud config set auth/disable_credentials true
-./google-cloud-sdk/bin/gcloud config set project ${TEST_PROJECT}
-./google-cloud-sdk/bin/gcloud config set api_endpoint_overrides/spanner http://localhost:9020/
-./google-cloud-sdk/bin/gcloud spanner instances create spannerintegration \
+$REPO_GCLOUD config set auth/disable_credentials true
+$REPO_GCLOUD config set project ${TEST_PROJECT}
+$REPO_GCLOUD config set api_endpoint_overrides/spanner http://localhost:9020/
```
amanda-tarafa, line 60: Wonder if it be a good idea to rollback these config modifications? Would this affect the gcloud client installed in Travis, or just this one that we installed here? I read on the other PR that we couldn't use the Travis one for some reason but not sure if this will affect that one.

amanda-tarafa, line 60: Also by looking at the logs, localhost:9020 seems to be the REST server and localhost:9010 the gRPC one, not sure about whether that's something we'd need to change here, tbh, just flagging.

jskeet: I don't believe we need to roll them back - I'd expect it to only affect the configuration of *this* installation of gcloud. (Not that it matters too much given that it's in a fresh VM.)

jskeet: I wouldn't actually expect any of these to be needed, but we'll see - I've created a new commit that removes them all... if that passes, let's just go with it.

amanda-tarafa: I think they are needed for creating the instance that is done with gcloud, they are not needed for running the actual tests. That's why setting the endpoint to rest is OK, glcoud probably uses rest instead of gRPC.

jskeet: I'm sure I've used gcloud for creating an instance without this, but I suspect there's some oddity in terms of authentication - it does seem to need it. Will merge when back to green.

PR #5172, opened 2020-07-14
Add the ability to push using different NuGet API keys based on package

jskeet:
(Requires an internal change that I've sent to you, Amanda.)

### DIFF
--- a/.kokoro/release.sh
+++ b/.kokoro/release.sh
@@ -17,8 +17,11 @@ cd ..
 
 export GOOGLE_APPLICATION_CREDENTIALS="$KOKORO_KEYSTORE_DIR/73609_cloud-sharp-jenkins-compute-service-account"
 export REQUESTER_PAYS_CREDENTIALS="$KOKORO_KEYSTORE_DIR/73609_gcloud-devel-service-account"
-export DOCS_CREDENTIALS="$KOKORO_KEYSTORE_DIR/73713_docuploader_service_account"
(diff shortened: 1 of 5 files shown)


.kokoro/release.sh
```diff
@@ -51,7 +54,28 @@ then
...
   # Push the changes to nuget.
   cd ./releasebuild/nuget
-  for pkg in *.nupkg; do dotnet nuget push -s https://api.nuget.org/v3/index.json -k $NUGET_API_KEY $pkg; done
+  for pkg in *.nupkg
+  do
+    # Work out just the package ID based on the filename.
+    pkg_id=$(echo $pkg | sed -r 's/^(.*)\.([0-9]+\.[0-9]+\.[0-9]+(-.*)?)\.nupkg$/\1/g')
+    # Work out the package owner based on apis.json and the package ID
+    default_package_owner=$([[ $pkg == Google.Cloud* ]] && echo google-cloud || echo google-api-packages)
+    package_owner=$PYTHON3 ../tools/getapifield.py ../apis/apis.json $pkg_id --default=$default_package_owner
```
amanda-tarafa, line 63: Can't you add the finding of the default owner logic to getapifield.py? Seems like a better place that will allow for more reuse if needed.

jskeet: I'd prefer not to put it into getapifield.py - that was intended to be utterly generic. I could create another Python tool to do it if we wanted to - although the part about the NuGet API key case would still be in here.
Perhaps leave it here until we have a second place we need it? I can't think of reasons we need the package owner other than when pushing packages, at the moment.

amanda-tarafa: Yep, that's fine.

PR #5192, opened 2020-07-17
Prints available disk space during build.

amanda-tarafa:
To try an see whether:

- We are running on smaller machines than we think we are.
- We are actually using up all the space.

Draft while I run Kokoro on this branch to get it right.


amanda-tarafa:
This is now printing disk space, once we see it fail we can print disk space in more places to try and pinpoint what makes it fail.
The two manual Kokoro executions went to the correct machine.

PR #5200, opened 2020-07-23
First prototype of a console formatter for Google Cloud Logging

jskeet:
The majority of this commit is a sample app, which can probably be removed later.

There's just one perfunctory test, and a lot more work to do thinking about exactly how we want the JSON to look - but it's a start.
(There's no hurry here, as we probably don't want to release anything before the ConsoleFormatter API goes GA anyway.)

Screenshots of a before/after:

Vanilla console logging:

before

With the console formatter:

after


amanda-tarafa:
I'll take a look later today. It does look pretty good at first glance.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.csproj
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <Version>1.0.0-alpha00</Version>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <Description>ConsoleFormatter for Google Cloud Logging.</Description>
+    <PackageTags>Logging;Stackdriver;ILogger;Console;Google;Cloud</PackageTags>
```
amanda-tarafa, line 8: Remove Stackdriver? Granted people might still be using that for looking libs up.

jskeet: Yup, that was the only reason for including it. We can remove all the Stackdriver tags from all packages in a couple of years when they're no longer useful at all.

PR #5199, opened 2020-07-21
[Diagnostics]: Associates logs to the correct trace/span if there is one.

amanda-tarafa:
* Fixes #5190

(I'm running integration tests locally).

PR #5201, opened 2020-07-23
[Diagnostics] Fixes some flakiness introduced in #5199.

amanda-tarafa:
In local, one of the tests that uses the server that traces everything was being run first, so the RateLimiter was set to trace everything from the start. In CI it seems that one of the tests that force tracing through the header was run first, so the RateLimiter was set to trace nothing.
Reseting the RateLimiter for each test fixes it (and it's the same we do for TraceTests)


apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/Logging/LoggingTest.cs
```diff
@@ -41,11 +41,18 @@ public class LoggingTest : IClassFixture<LogValidatingFixture>
...
 
         private readonly LogValidatingFixture _fixture;
 
-        public LoggingTest(LogValidatingFixture fixture) => _fixture = fixture;
+        public LoggingTest(LogValidatingFixture fixture)
+        {
+            _fixture = fixture;
+
+            // The rate limiter instance is static and only set once.  If we do not reset it at the
+            // beginning of each tests the qps will not change.  This is dependent on the tests not
```
jskeet, line 52:
```suggestion
            // beginning of each test the qps will not change.  This is dependent on the tests not
```

amanda-tarafa: Done, here, and where I copied it from.

Issue #5181, opened 2020-07-15
spanner: set the google-cloud-resource-prefix header in clients


skuruppu:
The `google-cloud-resource-prefix` header is expected to be set for the data and two admin clients of Cloud Spanner. I found that the header is set in [Firestore](https://github.com/googleapis/google-cloud-dotnet/blob/053434fa2a4d99da8a804d1645e2b634fe64be45/apis/Google.Cloud.Firestore.V1/Google.Cloud.Firestore.V1/FirestoreClientPartial.cs#L54) and [Datastore](https://github.com/googleapis/google-cloud-dotnet/blob/8eda274f0fe6e7fa4301da9d44b2662e25771646/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreClientPartial.cs#L22).

So I can follow that example to set it for the Spanner clients.

jskeet:
Great, thank you. Let me know if anything is unclear. (It's somewhat annoying that we have to do this via manual code, admittedly.)

jskeet:
@skuruppu: Can this issue now be closed? And would you like a new release, or do you want to wait until #5209 has been addressed?

skuruppu:
@jskeet not quite :( It turns out I forgot to cover the StreamingRead API. I also haven't yet handled the admin APIs. I may take me a couple more weeks to complete these tasks. So we should hold out on a release until I do these.

These changes are not related to #5209 so releases can be done independently of that.

skuruppu:
@jskeet it turns out we [don't have hand-written wrappers](https://github.com/googleapis/google-cloud-dotnet/search?q=StreamingRead&unscoped_q=StreamingRead) around the StreamingRead API, which I think means we can't do something like #5255. Instead, we have to ask users to set the header manually. Is my understanding correct?

jskeet:
In vacation until Tuesday - will look then.

jskeet:
Right, now I can get my head into this again... basically we don't support StreamingRead in any client-friendly way at the moment.

Two options to consider:

- Support it in PooledSession, which probably requires additional work to support PartialResultSet; at this point we could add the header ourselves
- Leave it in its current state of "I'd be astonished if you wanted to use the raw API for this"

I don't know the use cases for calling StreamingRead, so I don't know how important it is. I'd suggest going with option 2 for now, but including option 1 into future planning meetings as something to consider. Does that work for you?

skuruppu:
[...]

PR #5203, opened 2020-07-24
feat(spanner): set resource header in data client

skuruppu:
Sets the google-cloud-resource-prefix header in the Spanner data client so that we can support resource-based routing.

Towards #5181

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/ResourcePrefixHeaderTest.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/ResourcePrefixHeaderTest.cs
@@ -1,0 +1,217 @@
+// Copyright 2020, Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Grpc.Core;
+using System;
+using System.Linq;
+using Xunit;
+
+namespace Google.Cloud.Spanner.V1.Tests
+{
+    public class ResourcePrefixHeaderTest
+    {
+        private const string ResourcePrefixHeader = "google-cloud-resource-prefix";
+        private const string SampleDatabaseName = "projects/proj/instances/inst/databases/db";
+        private const string SampleSessionName = "projects/proj/instances/inst/databases/db/sessions/sess";
+
+        [Fact]
+        public void NoHeaderOnCreateSessionIfEmptyDatabaseName()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            client.CreateSession(new CreateSessionRequest());
+            Assert.Null(invoker.Metadata.FirstOrDefault(e => e.Key == ResourcePrefixHeader));
+        }
+
+        [Fact]
+        public void NoHeaderOnCreateSessionIfInvalidDatabaseName()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            client.CreateSession(new CreateSessionRequest { Database = "projects/proj/instances/databases" });
+            Assert.Null(invoker.Metadata.FirstOrDefault(e => e.Key == ResourcePrefixHeader));
+        }
+
+        [Fact]
+        public void NoHeaderOnGetSessionIfEmptySessionName()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            client.GetSession(new GetSessionRequest());
+            Assert.Null(invoker.Metadata.FirstOrDefault(e => e.Key == ResourcePrefixHeader));
+        }
+
+        [Fact]
+        public void NoHeaderOnGetSessionIfInvalidSessionName()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            client.GetSession(new GetSessionRequest { Name = "projects/proj/instances/inst/databases/sessions" });
+            Assert.Null(invoker.Metadata.FirstOrDefault(e => e.Key == ResourcePrefixHeader));
+        }
+
+        [Fact]
+        public void SetsHeaderOnCreateSession()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            client.CreateSession(new CreateSessionRequest { Database = SampleDatabaseName });
+            Metadata.Entry entry = Assert.Single(invoker.Metadata, e => e.Key == ResourcePrefixHeader);
+            Assert.Equal(SampleDatabaseName, entry.Value);
+        }
+
+        [Fact]
+        public void SetsHeaderOnBatchCreateSessions()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            client.BatchCreateSessions(new BatchCreateSessionsRequest { Database = SampleDatabaseName });
+            Metadata.Entry entry = Assert.Single(invoker.Metadata, e => e.Key == ResourcePrefixHeader);
+            Assert.Equal(SampleDatabaseName, entry.Value);
+        }
+
+        [Fact]
+        public void SetsHeaderOnGetSession()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            client.GetSession(new GetSessionRequest { Name = SampleSessionName });
+            Metadata.Entry entry = Assert.Single(invoker.Metadata, e => e.Key == ResourcePrefixHeader);
+            Assert.Equal(SampleDatabaseName, entry.Value);
+        }
+
+        [Fact]
+        public void SetsHeaderOnListSessions()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            client.ListSessions(new ListSessionsRequest { Database = SampleDatabaseName }).AsRawResponses().First();
+            Metadata.Entry entry = Assert.Single(invoker.Metadata, e => e.Key == ResourcePrefixHeader);
+            Assert.Equal(SampleDatabaseName, entry.Value);
+        }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientPartial.cs
```diff
@@ -0,0 +1,150 @@
...
+    public partial class SpannerClientImpl
+    {
+        /// <summary>
+        /// The name of the header used for efficiently routing requests.
+        /// </summary>
+        /// <remarks>
+        /// This should be set to the database resource name ("projects/{projectId}/instances/{instanceId}/databases/{databaseId}") for any RPC.
+        /// For non-streaming calls, <see cref="SpannerClientImpl"/> performs this automatically. This cannot be performed
+        /// automatically for streaming calls due to the separation between initializing the stream and sending requests, so
+        /// client code should set the value in a <see cref="CallSettings"/>. Typically this is performed with either the
```
skuruppu, line 31: This part worries me, I think I have to do this in the hand-written later.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientPartial.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        // Visible for testing
+
+        /// <summary>
+        /// Retrieves the database resource name from a full resource name.
+        /// Validation is performed as far as the database ID but no further; the database ID is deemed to end at the first slash.
+        /// </summary>
+        /// <param name="resource">The resource name, which must start with projects/{project_id}/instances/{instance_id}/databases/{database_id}</param>
+        /// <returns>The database resource name</returns>
+        internal static string GetDatabaseResourceName(string resource)
```
jskeet, line 95: Drive-by as I don't have time to review *right* now - I strongly suspect this could be massively reduced using DatabaseName.TryParse or DatabaseName.Parse.

skuruppu: Ah yes, that's a good idea. I will fix it early next week. Don't worry about reviewing until I fix it.

skuruppu: That was much nicer. I still want to write more comprehensive tests for whether the header is set but at least logic-wise, this is hopefully now closer to the final implementation.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/SpannerClientImplTest.cs
```diff
@@ -0,0 +1,65 @@
...
+    {
+        [Fact]
+        public void ClientSetsResourcePrefixHeader()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            const string session = "projects/proj/instances/inst/databases/db/sessions/sess";
+            client.BeginTransaction(new BeginTransactionRequest { Session = session });
+            var found = false;
+            foreach (Metadata.Entry e in invoker.Metadata)
```
jskeet, line 31: I'd suggest using:

```csharp
Metadata.Entry entry = invoker.Metadata.Single(e => e.Key == "google-cloud-resource-prefix");
Assert.Equal("projects/proj/instances/inst/databases/db", entry.Value);
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientPartial.cs
jskeet: Yes, that would make sense. There are a few places we could do it - I don't think it'll be terribly hard. I'm happy for you to look, or I can do so. I'd suggest doing that as a separate PR, just for simplicity.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientPartial.cs
```diff
@@ -0,0 +1,98 @@
...
+            if (string.IsNullOrEmpty(resource))
+            {
+                return;
+            }
+
+            // Try to parse the resource as a database name.
+            DatabaseName database;
+            if (!DatabaseName.TryParse(resource, out database))
+            {
+                // If that fails, the resource name is likely a session name.
```
jskeet, line 89: Do we definitely want to use Parse rather than TryParse here? If we have a call with a resource which is neither a session or a database name, do we definitely want it to throw here?

Moreover, might it not make sense to use different methods here, one expecting a session name and one expecting a database name? I'd expect us to know which should be used on a per-request basis.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/SpannerClientImplTest.cs
skuruppu: Hmmm I was trying to do something similar by following this [example](https://github.com/googleapis/google-cloud-dotnet/blob/d1056b9a083ab925bbe63053624b79f108d0c3ee/apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Tests/ApiClientHeaderTest.cs#L32). Except the compiler kept telling me that `Metadata` doesn't contain a definition for `FirstOrDefault` and `Single`.

I looked at the [Metadata docs](https://grpc.github.io/grpc/csharp-dotnet/api/Grpc.Core.Metadata.html#methods) and indeed those methods are not defined. Then I thought I would use `Metadata.Contains` but xUnit was unhappy about that.

Some Googling gave me the idea to use Assert.Contains() which finally worked :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientPartial.cs
skuruppu: Yup, happy to put this in a separate PR.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientPartial.cs
skuruppu: I guess you're right and it's overkill to fail the request here if an appropriate resource name is not specified. I guess that's not the job of the client library.

I split the function into two as you suggested and now ignoring the failure to parse.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/SpannerClientImplTest.cs
jskeet: `Metadata` implements `IList<Metadata.Entry>`. `Single` is an extension method defined in LINQ, so `using System.Linq;` would have worked.

However, thinking about it, `Assert.Single` is *really* what you want:

```csharp
Metadata.Entry entry = Assert.Single(invoker.Metadata, e => e.Key == "google-cloud-resource-prefix");
Assert.Equal("projects/proj/instances/inst/databases/db", entry.Value);
```

That way you'll get a better error message if there are no/multiple matches.

skuruppu: You're right. We don't want to set it multiple times. Updated the implementation and also added tests for all methods. There one (ListSessions) that I can't get working so I will look into it tomorrow.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/ResourcePrefixHeaderTest.cs
```diff
@@ -0,0 +1,184 @@
...
+            Assert.Equal(database, entry.Value);
+        }
+
+        [Fact]
+        public void SetsHeaderOnListSessions()
+        {
+            var invoker = new FakeCallInvoker();
+            var client = new SpannerClientBuilder { CallInvoker = invoker }.Build();
+            client.ListSessions(new ListSessionsRequest { Database = database });
+            // TODO(skuruppu): Figure out why the invoker doesn't have Metadata set.
```
jskeet, line 64: I suspect this is because it hasn't actually issued a request yet. List operations are lazy. If you change the above to 

```csharp
client.ListSessions(new ListSessionsRequest { Database = database }).AsRawResponses().First();
```

... then I suspect it'll work. (Or it may throw another exception because your fake doesn't populate the response appropriately... you'll have to see.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/ResourcePrefixHeaderTest.cs
```diff
@@ -0,0 +1,184 @@
...
+// limitations under the License.
+
+using Grpc.Core;
+using System;
+using System.Linq;
+using Xunit;
+
+namespace Google.Cloud.Spanner.V1.Tests
+{
+    public class ResourcePrefixHeaderTest
```
jskeet, line 22: I think you should add a couple of tests (one for database, one for session) for cases where the request contains a badly-formatted resource name. (And another for when it's not populated at all.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/ResourcePrefixHeaderTest.cs
```diff
@@ -0,0 +1,184 @@
...
+    {
+        const string header = "google-cloud-resource-prefix";
+        const string database = "projects/proj/instances/inst/databases/db";
+        const string session = "projects/proj/instances/inst/databases/db/sessions/sess";
```
jskeet, line 26:
```suggestion
        private const string ResourcePrefixHeader = "google-cloud-resource-prefix";
        private const string SampleDatabaseName = "projects/proj/instances/inst/databases/db";
        private const string SampleSessionName = "projects/proj/instances/inst/databases/db/sessions/sess";
```

(If you're using Visual Studio, use F2 to do the renaming and all the uses will change too.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/ResourcePrefixHeaderTest.cs
skuruppu: Thanks heaps for the suggestion, it works now!

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1.Tests/ResourcePrefixHeaderTest.cs
skuruppu: That's a good idea. Sorry I missed this. I added 4 tests.

skuruppu:
PTAL @jskeet. I've addressed all the feedback and got that last test to pass.

skuruppu:
Friendly ping on this PR @jskeet. Just curious to know if there's anything else I should do here. Otherwise I'll move onto supporting the header in streaming RPCs.

skuruppu:
All good, no worries at all @jskeet. Thanks for merging it in. I'll start work on the next piece.

PR #5221, opened 2020-07-30
[Diagnostics]: Trace rate limiter always/never trace.

### LINKED ISSUE #5213: [Diagnostics]: The trace rate limiter cannot be set to trace nothing or everything.
The trace rate limiter for sample tracing is set through QPS (queries per second to trace). Supposedly, for configuring Diagnostics tracing so that no sample tracing is performed (only traces forced through header are done), a small enough possitive double value should be provided for QPS. But even if the smallest possible value is set for QPS, the first call is always traced, and theoretically, if the application runs for really long periods of time, some other subsequent request will be traced as well.

1. This could be fixed by never tracing the first request, but that's semantically breaking.

The real cause of this issue is that really small positive double values, zero and really big positive double values are not allowed, because of the way the rate limiter is implemented, these values end up overflowing a TimeSpan value. So we never really can configure the rate limiter to trace nothing or trace everything in absolute terms (in practical terms we are close).

2. For fixing, allow these extreme values and interpret the TimeSpan overflows as:

 - Never trace when the overflow is in excess, as in, so much time needs to pass between traces that it actually means no tracing.
 - Always trace when the overflow is in defect, as in, so little time needs to pass between traces that it actually means always trace.

This approach has the advantage that is non-breaking.

3. An alternative approach can be considered where the rate limiter implementation does not rely (at least so heavily) on TimeSpan, but that would be at least semantically breaking and will requiere more time for redesign and implement.

@jskeet Thoughts welcome. I'm all for going with the second fix.

amanda-tarafa:
Fixes #5213 

Each commit contains a possible solution. I like the second commit better because it's more general, but I'm fine with the first one as well.
I'll deal with tests after we've chosen one of the alternatives.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/RateLimiter.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/RateLimiter.cs
@@ -23,7 +23,7 @@ namespace Google.Cloud.Diagnostics.Common
     internal sealed class RateLimiter
     {
         /// <summary>A mutex to protect the rate limiter instance.</summary>
-        private static object _instanceMutex = new object();
+        private static readonly object _instanceMutex = new object();
 
         /// <summary>The single rate limiter instance.</summary>
         private static RateLimiter _instance;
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Trace/RateLimiter.cs
```diff
@@ -62,15 +62,18 @@ internal RateLimiter(double qps, ITimer timer)
...
-                _ => Convert.ToInt64(TimeSpan.FromSeconds(1 / qps).TotalMilliseconds)
-            };
+                _fixedDelayMillis = long.MaxValue;
+            }
+            else
+            {
+                _fixedDelayMillis = (1000 / qps) switch
+                {
+                    double milis when milis > TimeSpan.MaxValue.TotalMilliseconds || milis > long.MaxValue => long.MaxValue,
+                    double milis => Convert.ToInt64(milis)
```
jskeet, line 74:
```suggestion
                    double millis when millis > TimeSpan.MaxValue.TotalMilliseconds || millis > long.MaxValue => long.MaxValue,
                    double millis => Convert.ToInt64(millis)
```

amanda-tarafa:
Fixed the typo and added tests.
This is now ready for review.
Integrations tests are running on local.

jskeet:
Will review tomorrow - meetings for the rest of the day :(

PR #5230, opened 2020-08-03
NLog GoogleStackDriverTarget skip reflection on Stream-objects

snakefoot:
Stream-objects properties often throw exceptions, and it is difficult to get useful information by reflection.

### DIFF
--- a/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
+++ b/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
@@ -141,11 +141,12 @@ namespace Google.Cloud.Logging.NLog
             jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(MemberInfo)));
             jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(Assembly)));
             jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(Module)));
+            jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(System.IO.Stream)));
 
             jsonSettings.Error = (sender, args) =>
             {
                 // Serialization of properties that throws exceptions should not break everything
-                InternalLogger.Warn(args.ErrorContext.Error, "GoogleStackdriver(Name={0}): Error serializing exception property: {1}", Name, args.ErrorContext.Member);
+                InternalLogger.Debug(args.ErrorContext.Error, "GoogleStackdriver(Name={0}): Error serializing exception property: {1}", Name, args.ErrorContext.Member);
                 args.ErrorContext.Handled = true;
             };
             var jsonSerializer = Newtonsoft.Json.JsonSerializer.CreateDefault(jsonSettings);


apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
```diff
@@ -141,11 +141,12 @@ protected override void InitializeTarget()
             jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(MemberInfo)));
             jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(Assembly)));
             jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(Module)));
+            jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(System.IO.Stream)));
```
amanda-tarafa, line 144: This is probably fine...

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
```diff
@@ -141,11 +141,12 @@ protected override void InitializeTarget()
             jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(MemberInfo)));
             jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(Assembly)));
             jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(Module)));
+            jsonSettings.Converters.Add(new ToStringJsonConverter(typeof(System.IO.Stream)));
 
             jsonSettings.Error = (sender, args) =>
             {
                 // Serialization of properties that throws exceptions should not break everything
-                InternalLogger.Warn(args.ErrorContext.Error, "GoogleStackdriver(Name={0}): Error serializing exception property: {1}", Name, args.ErrorContext.Member);
+                InternalLogger.Debug(args.ErrorContext.Error, "GoogleStackdriver(Name={0}): Error serializing exception property: {1}", Name, args.ErrorContext.Member);
```
amanda-tarafa, line 149: ... but we might not want to change the level of logging serialization errors at this point.

chrisdunelm:
Thanks @snakefoot for the PR.
I agree with @amanda-tarafa - adding a converter for `System.IO.Stream` is fine, but probably best to revert the `InternalLogger.Warn` change.

snakefoot:
The NLog InternalLogger is for diagnostics


Warn / Error is for serious errors like no access, no connection or something blew up. Thus the entire logging pipeline is broken.


NLog suppresses by default issues caused by bad input data being logged. Ex. invalid message template or missing parameters.


NLog users often wants to monitor these exceptional events. So configure the NLog InternalLogger to monitor for Warn/Error to see when the logging pipeline has broken.


Because users often uses giant frameworks, where they have no control of what is being logged then errors during reflection generates unwanted noise. Making it difficult to monitor the health of the NLog logging pipeline.


I introduced the logging the first place. Now I see the Loglevel was initially wrong. Please let me correct that mistake.




Sent from my Sony Xperia


---- Chris Bacon wrote ----


Thanks @snakefoot<https://github.com/snakefoot> for the PR.
I agree with @amanda-tarafa<https://github.com/amanda-tarafa> - adding a converter for System.IO.Stream is fine, but probably best to revert the InternalLogger.Warn change.

chrisdunelm:
Thanks for the explanation of why the change is sensible.
Yes, this seems like a reasonable thing to do.

PR #5250, opened 2020-08-17
Updates GAX and gRPC default versions.

amanda-tarafa:
I'm not sure if I should update all explicit dependency versions in apis.json or if we should remember to do so when we next release each library?


amanda-tarafa:
It seems I need to regenerate project files so that at least the non-ga projects pick up the change. On it.

amanda-tarafa:
I've added a second commit that regenrates all non-ga projects.

PR #5267, opened 2020-08-26
Ignore "EffectivePackageOwner" when writing out API metadata

jskeet:
This is a derived property; we shouldn't serialize it.


chrisdunelm (approved):
LGTM

PR #5299, opened 2020-09-09
Generate Assured Workloads API

jskeet:
No smoke tests are suggested, but it looks like the API isn't public yet, so I don't want to release it right now.


amanda-tarafa (approved):
LGTM

PR #5308, opened 2020-09-11
Massively speed up the version history update

jskeet:
We don't need to check every file in every commit: if apis/xyz/xyz
has the same sha in the parent of the commit and the "candidate"
commit, then it definitely doesn't contain any changes to API xyz.

This cuts the per-API time down from about 10 seconds to 1 second.

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/UpdateHistoryCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/UpdateHistoryCommand.cs
@@ -77,8 +77,7 @@ namespace Google.Cloud.Tools.ReleaseManager
             var id = api.Id;
             var pathPrefix = $"apis/{id}/{id}/";
             var projectFile = $"apis/{id}/{id}/{id}.csproj";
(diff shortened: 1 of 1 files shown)


tools/Google.Cloud.Tools.ReleaseManager/UpdateHistoryCommand.cs
```diff
@@ -17,6 +17,7 @@
 using LibGit2Sharp;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
```
chrisdunelm, line 20: Is this needed?

chrisdunelm (approved):
LGTM
One questions.

tools/Google.Cloud.Tools.ReleaseManager/UpdateHistoryCommand.cs
jskeet: Nope, sorry - that was left over from using Stopwatch. Removed.

jskeet:
(Will merge on green.)

PR #5314, opened 2020-09-14
[Diagnostics] Sets LogEntry.TraceSampled to true if the entry is traced.

### LINKED ISSUE #5307: TraceSampled
Last December you added support for the [TraceSampled LogEntry field](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Logging.V2/docs/history.md#version-230-released-2019-12-10), but it's unclear how to actually set it. It doesn't appear to get set by default, even if the decision to trace is approved based on provided `TraceOptions`.

Why is this important? I want to use the "View trace details" option in the Cloud Console, and the only way to do that is to set TraceSampled to true - see [here](https://cloud.google.com/logging/docs/view/logs-viewer-interface#trace_data).

amanda-tarafa:
Fixes #5307


chrisdunelm (approved):
LGTM

PR #5318, opened 2020-09-15
Release Google.Cloud.Asset.V1 version 2.4.0

jskeet:
Changes in this release:

- [Commit c9ca107](https://github.com/googleapis/google-cloud-dotnet/commit/c9ca107): Remove AnalyzeIamPolicy and ExportIamPolicyAnalysis RPCs

BREAKING CHANGE: The change here effectively reverts part of [commit 8ef5dc2](https://github.com/googleapis/google-cloud-dotnet/commit/8ef5dc2), so removes part of the public API.

We have taken the decision to publish this as a minor version rather than going to version 3.0, as we don't expect customers to be affected by the nature of the breaking change:

- The affected version (2.3.0) has been delisted, and had only been installed by a small number of customers, as reported on NuGet
- This version *is* compatible with 2.2.0 and earlier versions
- The accidentally-included new APIs were not enabled server-side, so customers could not have started relying on the server implementation yet

We apologise for any confusion this causes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Each package name links to the documentation for that package.
 |---------|----------------|-------------|
 | [Google.Analytics.Admin.V1Alpha](https://googleapis.dev/dotnet/Google.Analytics.Admin.V1Alpha/1.0.0-alpha01) | 1.0.0-alpha01 | Analytics Admin API |
 | [Google.Analytics.Data.V1Alpha](https://googleapis.dev/dotnet/Google.Analytics.Data.V1Alpha/1.0.0-alpha01) | 1.0.0-alpha01 | Google Analytics Data API |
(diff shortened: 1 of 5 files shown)


jskeet:
Investigating the breakage.

jskeet:
The Travis failure is due to it correctly detecting that this is a breaking change in a new minor version. Nice to see the tool working as intended. Will override as an administrator when approved.

chrisdunelm:
It looks like there's a line-ending change for all lines in every file ;) Can this be fixed?

jskeet:
Oh, not again. I need to work out what's going on with this...

jskeet:
Okay, *this* one I've managed to fix - PTAL.

chrisdunelm (approved):
LGTM

amanda-tarafa (approved):
LGTM.
But the compatibility checker is complaining, it won't let you bump to minor because of the breaking change.

jskeet:
> But the compatibility checker is complaining, it won't let you bump to minor because of the breaking change.

Yes, that's working as intended - I had to use admin powers to merge.

PR #5322, opened 2020-09-15
Regenerate all APIs with updated generator

jskeet:
The only relevant change is to enums: we now generate a comma after all enum members, including the last one.


amanda-tarafa (approved):
LGTM

PR #5337, opened 2020-09-22
Disable pretty printing in BigQuery unless explicitly requested

### LINKED ISSUE #5330: BigQuery: audit use of dense JSON encoding
# Request
Review this library to ensure dense encoding (e.g. `prettyPrint=false`) is specified when possible when communicating with the BigQuery API methods.

# Additional Context
Pretty printed JSON responses from server can have significant response size inflation for complex objects.  BigQuery data APIs tend to exacerbate this.  Library can improve performance by using dense encoding, particular for underlying methods like jobs.getQueryResults and tabledata.list that may supply large, complex JSON responses.

Info about `prettyPrint` system API parameter:
https://cloud.google.com/apis/docs/system-parameters

# Internal (for Googlers)
See issue 168725137

jskeet:
This is client-wide, and can only be set on the BigQueryClientBuilder (to avoid an explosion of BigQueryClientImpl constructors in the long run). It is then propagated to all requests. It defaults to false.

In order to test this change properly, I've had to also add a Service property to BigQueryClientBuilder, but that seems appropriate anyway.

Fixes #5330.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientBuilder.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientBuilder.cs
@@ -36,24 +36,48 @@ namespace Google.Cloud.BigQuery.V2
         /// </summary>
         public string DefaultLocation { get; set; }
 
+        /// <summary>
+        /// If set, overrides the default pretty-print setting.
+        /// </summary>
+        public bool? PrettyPrint { get; set; }
+
+        // Note: we may want to make this public in the future. If we do, we should check that no credentials
+        // are set in Validate().
+        /// <summary>
+        /// If set, this is used as the underlying service for the client.
+        /// Note that this should rarely be used outside testing, unless the service comes from another <see cref="BigQueryClient"/>.
+        /// Usually this library constructs a service using specific settings such as the JSON serializer settings provided by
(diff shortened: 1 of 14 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientBuilder.cs
```diff
@@ -66,6 +88,7 @@ public override async Task<BigQueryClient> BuildAsync(CancellationToken cancella
         protected override void Validate()
         {
             base.Validate();
+            // TODO: Validate that if Service is set, no credentials are set?
```
amanda-tarafa, line 91: That would be inline with what we do with the invokers in the gRPC equivalents. Can't Service here just be internal? We can always make it public later on if there's a need for it apart from testing.

amanda-tarafa (approved):
LGTM. Just one comment.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientBuilder.cs
jskeet: We could definitely make it internal, yes - I just don't like the idea that *we* can test stuff but others can't. Still, it means we don't need to worry about the TODO for now. Will remove this TODO, but add a note on the newly-internal property.

PR #5344, opened 2020-09-23
Fix Firestore conformance tests

### LINKED ISSUE #5343: Some Firestore conformance tests are currently ignored
We need to fix the conformance tests, then unskip them.

jskeet:
Fixes #5343.

This should not be merged until we've seen a spec for update transforms.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/WriteBatch.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/WriteBatch.cs
@@ -62,7 +62,7 @@ namespace Google.Cloud.Firestore
             RemoveSentinels(fields, sentinels);
             // Force a write if we've not got any sentinel values. Otherwise, we end up with an empty transform instead,
             // just to specify the precondition.
-            AddUpdateWrites(documentReference, fields, s_emptyFieldPathList, Precondition.MustNotExist, sentinels, sentinels.Count == 0, false);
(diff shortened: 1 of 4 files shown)


amanda-tarafa (commented):
LGTM.
I've added the `do not merge` label and won't really approve so we don't accidentally merge before we have confirmation.

jskeet:
Note: using the precondition to determine whether or not to use an empty update mask may not be a good idea. It *may* be as simple as depending on which method is calling it (so just have an extra flag...)

jskeet:
Amanda: I've added an extra commit to this based on internal guidance. Please could you have a look? If you're happy, we can merge this and we're done :)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/WriteBatch.cs
```diff
@@ -264,7 +244,9 @@ internal async Task<IList<WriteResult>> CommitAsync(ByteString transactionId, Ca
                     Fields = { fields },
                     Name = documentReference.Path,
                 },
-                UpdateMask = includeEmptyUpdateMask || updatePaths.Count > 0 ? new DocumentMask { FieldPaths = { updatePaths.Select(fp => fp.EncodedPath) } } : null,
+                UpdateMask = updatePaths is null
```
jskeet, line 247: I could compute the DocumentMask (or null) outside the object initializer if you'd prefer.

amanda-tarafa: No, this is fine.

amanda-tarafa (approved):
Yep, so much cleaner.

PR #5345, opened 2020-09-23
feat: retry SpannerTransaction on aborted error [WIP]

olavloite:
Adds automatic retries of `Aborted` transactions when using the normal ADO.NET transactions (i.e. not using `SpannerConnection.RunWithRetriableTransactionAsync(..)`).

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableSpannerTransaction.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableSpannerTransaction.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2019 Google LLC
+﻿// Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableSpannerTransaction.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableSpannerTransaction.cs
@@ -1,0 +1,374 @@
+﻿// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using Google.Apis.Auth.OAuth2.Requests;
+using Google.Cloud.Spanner.V1;
+using Google.Cloud.Spanner.V1.Internal.Logging;
+using Google.Protobuf;
+using Google.Rpc;
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Data;
+using System.Data.Common;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Status = Grpc.Core.Status;
+
+namespace Google.Cloud.Spanner.Data
+{
+    internal interface IRetriableStatement
+    {
+        internal Task Retry(RetriableSpannerTransaction transaction, CancellationToken cancellationToken, int timeoutSeconds);
+    }
+
+    internal sealed class RetriableDmlStatement : IRetriableStatement
+    {
+        private readonly ExecuteSqlRequest _request;
+        private readonly long _updateCount;
+
+        internal RetriableDmlStatement(ExecuteSqlRequest request, long updateCount)
+        {
+            this._request = request;
+            this._updateCount = updateCount;
+        }
+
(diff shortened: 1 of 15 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableSpannerTransaction.cs
```diff
@@ -0,0 +1,374 @@
...
+                catch (SpannerException e) when (e.ErrorCode == ErrorCode.Aborted)
+                {
+                    Retry();
+                }
+            }
+        }
+
+        Task<int> ISpannerTransaction.ExecuteMutationsAsync(List<Mutation> mutations, CancellationToken cancellationToken, int timeoutSeconds)
+        {
+            return DoExecuteMutationsAsync(mutations, cancellationToken, timeoutSeconds);
```
amanda-tarafa, line 300: This is the problem right here. These mutations might be based on data that is stale when they get to be executed as part of the commit.
Imagine a read goes through, and we create a mutation based on that read, and add it to the transaction mutations. Then we have to retry because of some other operation was aborted. While we are retrying, someone updates the value we based the mutation on, now our mutation has become stale. Even if we retry the original read in which we based the mutation, the mutation itself won't be updated according to the newly read values, which would make the mutation stale. We would effectively be allowing through mutations that are based on invalid read locks (the original one that we had to retry).

We explored this approach and found it wasn't possible, you really need to take a delegate in that contains the whole sets of intrstuctions that need to be retried.

amanda-tarafa (changes requested):
I've left a comment explaining why this approach is not possible, unless I'm missing something here.
@jskeet, @skuruppu, what do you think?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableSpannerTransaction.cs
olavloite: No, the retry logic is checking for exactly that possibility, and if it detects it, it will abort the transaction after all. Consider the following (pseudo-code) example:

```
SELECT Id, Name FROM Singers WHERE Id=1;
-- The following statement is not valid SQL, but hopefully it's clear what's meant.
CREATE UPDATE_MUTATION Singer (Id, Name) VALUES (1, 'NewName');
COMMIT;
```

Now let's assume that another transaction updates the name of the same Singer to 'SomeOtherName' at the same time and commits. The above transaction will be aborted by Cloud Spanner. The `Aborted` error is in this example returned during the COMMIT call (but any statement could return `Aborted`) and the transaction will have to execute a retry.

The `RetriableSpannerTransaction` keeps track of all statements that have been executed during the transaction and the results that it received. When a retry is needed, it starts a new transaction and executes all statements again on the new transaction and compares the results that it receives during the retry with the results that it received during the initial transaction. If they match, the transaction is allowed to proceed as if nothing happened. If during the retry it receives different data, the retry will abort and the `Aborted` error will be propagated to the client application.

In the above example the first query [will be returned](https://github.com/googleapis/google-cloud-dotnet/blob/c99cc806b78137f032cc5f0674101236b579f6e7/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableSpannerTransaction.cs#L370) as a `SpannerDataReaderWithChecksum`. That reader keeps track of a running checksum of the data it has already seen, and if a retry is needed, it will compare the checksum from the original transaction with the checksum during the retry. In this example the checksums will not match, and the transaction will abort as it should.
If however the specific Singer row had not been updated and the transaction was aborted for some other reason, the retry will succeed.

olavloite: For some more context: This retry strategy is also what is used in the [Cloud Spanner JDBC driver](https://github.com/googleapis/java-spanner/blob/master/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ChecksumResultSet.java).

amanda-tarafa: Then I would argue that this strategy is way more restrictive than the one we have right now in that it will result in less successful retries. Take for instance the checksum data reader, it requires rows to be always returned in exactly the same order, I'm not sure this is guarenteed by Spanner. There's also a requirement for updated number of rows and errors even to be the same every time, just because you can only hold onto the mutations as they were initially. So in a sense this goes further than guaranteeing non-stale mutations, it enforces that the transaction will end up committing only if it can do so exactly the same as it would have the first time. This is in my opinion way more restrictive than usually expected for a retriable transaction.

Also, I wouldn't say that mutations are not going stale. They might not be stale with respect to the pure Spanner data, but they can go "semantically" stale. Say you have several mutations that store DateTime.Now. Depending on how many retries you get in between adding these mutations, the relation between those values will be wildly different and probably not meaningful. Or for instance, we would need to remember to never ever expose the Session object in the SpannerTransaction so that no one can base a mutation on that.

And the checksum overhead, I'd say that's not trivial either in terms of memory use or of computational complexity.

I don't want to sound overly negative or far fetched. But I think this is significant work that comes with a maintenance burden and I honestly don't see it adds much to what we already have. It's only just a little bit more like ADO.NET but it's not even that because ADO.NET does not define a retriable transaction, you still wouldn't be able to use the SpannerConnection as a DbConnection and get to this feature, same as what we have now.
Anyway, those are my two cents, @skuruppu and @jonskeet might think differently.

jskeet:
Just to confirm, I believe we should wait for @skuruppu to comment before proceeding further either way.

jskeet:
Closing this PR now as (having chatted with @skuruppu) we think it would *probably* live in an ORM package - but it's probably worth keeping the branch around so that you've got the code available when it's needed.

PR #5353, opened 2020-09-24
Release Google.Cloud.Asset.V1 version 2.5.0-beta01

amanda-tarafa:
Changes in this release:

- [Commit 0ca05f5](https://github.com/googleapis/google-cloud-dotnet/commit/0ca05f5): Regenerate all APIs using protoc 3.13 and Grpc.Tools 2.31.
Note that the change in gRPC generation is *just possibly* breaking, as it changes a constructor parameter type (to a base type). This is far less likely to break anyone than changing a *method* parameter, as method group conversions aren't involved. It also removes the dependency on Grpc.Core from the generated code.
- [Commit 6eed2d8](https://github.com/googleapis/google-cloud-dotnet/commit/6eed2d8): feat: added support for per type and partition export for Cloud Asset API.

### DIFF
--- a/apis/Google.Cloud.Asset.V1/docs/history.md
+++ b/apis/Google.Cloud.Asset.V1/docs/history.md
@@ -1,5 +1,16 @@
 # Version history
 
+# Version 2.5.0-beta01, released 2020-09-24
+
+- [Commit 0ca05f5](https://github.com/googleapis/google-cloud-dotnet/commit/0ca05f5): Regenerate all APIs using protoc 3.13 and Grpc.Tools 2.31.
+
+  Note that the change in gRPC generation is *just possibly* breaking,
+  as it changes a constructor parameter type (to a base type). This is
+  far less likely to break anyone than changing a *method* parameter,
+  as method group conversions aren't involved. It also removes the
+  dependency on Grpc.Core from the generated code.
+- [Commit 6eed2d8](https://github.com/googleapis/google-cloud-dotnet/commit/6eed2d8): feat: added support for per type and partition export for Cloud Asset API.
+
 # Version 2.4.0, released 2020-09-15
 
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Asset.V1/docs/history.md
```diff
@@ -1,5 +1,16 @@
 # Version history
 
+# Version 2.5.0-beta01, released 2020-09-24
+
+- [Commit 0ca05f5](https://github.com/googleapis/google-cloud-dotnet/commit/0ca05f5):
+  - Regenerate all APIs using protoc 3.13 and Grpc.Tools 2.31
```
amanda-tarafa, line 6: Oops, I'm now removing these bullets.

amanda-tarafa:
Ready for review.
It can very well wait till tomorrow for the tooling fix.

apis/Google.Cloud.Asset.V1/docs/history.md
```diff
@@ -1,5 +1,16 @@
 # Version history
 
+# Version 2.5.0-beta01, released 2020-09-24
+
+- [Commit 0ca05f5](https://github.com/googleapis/google-cloud-dotnet/commit/0ca05f5): Regenerate all APIs using protoc 3.13 and Grpc.Tools 2.31.
+
+  Note that the change in gRPC generation is *just possibly* breaking,
```
jskeet, line 7: We should decide whether or not we actually want to mention this at all. I still haven't thought of any cases where this would break code, even as corner cases, so I'm tempted to remove it entirely.

(If we do include it, I think it should be as a separate paragraph underneath the list, and probably shouldn't have as much detail. I'm happy to suggest specific text.)

amanda-tarafa:
Closing, will recreate tomorrow with some needed tooling changes.

Issue #5275, opened 2020-08-29
Tracing does not work well in Cloud Run with TimedBuffer


EatonZ:
I have a Cloud Run service behind Cloud Load Balancer. The service is a .NET Core API, utilizing this library to log traces.

If I send 5 concurrent requests, only one gets traced. This is confirmed by checking the Trace List in the Cloud Console, and also the HTTP responses - only the first response has a `X-Cloud-Trace-Context` header. If I send the 5 requests, but wait for the previous one to complete before sending the next, all 5 get traced. I looked through the options and I tried specifying `BufferOptions.NoBuffer()` in `UseGoogleDiagnostics`, but that didn't seem to make a difference. Am I missing something, or could this be a bug?

jskeet:
We'll look at this on Tuesday - Monday is a public holiday in the UK.

amanda-tarafa:
@EatonZ Can you provide a minimum but full application that reproduces the problem for you? We do have some tests around concurrent tracing and logging and they seem to work.

EatonZ:
@amanda-tarafa Here you go. It's a lot (sorry!), but it does reproduce the problem.

Please make sure you have gcloud installed and authenticated. You should also have the [latest PowerShell](https://github.com/PowerShell/PowerShell/releases/tag/v7.0.3). The built-in Windows version may not work. Finally, make sure Docker Desktop is installed and running.
1. [Download pack](https://buildpacks.io/docs/install-pack/#windows) and copy beside the .sln file.
2. Create a new GCP project so you can easily dispose of everything when done. Enable the Run and Compute APIs, and [create an Artifact Registry repo](https://console.cloud.google.com/artifacts/create-repo) named `ez-sample`.
3. Open GoogleBuild.ps1 in Notepad, put in your project ID at the top, and create/put in a service account (give it the Cloud Trace Agent role).
4. In GoogleCreateLoadBalancer.ps1, put in your project ID at the top.
5. Create an IP [here](https://console.cloud.google.com/networking/addresses/add). Premium, IPv4, **Global**, named `ez-sample-ip`.
6. Run these 2 commands:
`gcloud auth configure-docker "us-docker.pkg.dev"`
`gcloud auth print-access-token | docker login --username "oauth2accesstoken" --password-stdin "https://us-docker.pkg.dev"`
[...]

PR #5355, opened 2020-09-25
Change Span ID format string to improve Trace list consistency

EatonZ:
When viewing traces in Cloud Console Trace list, the span IDs are displayed as lower-case hex strings without the "0x" prefix.
When logging using this library, the span ID is set to an upper-case hex string with a "0x" prefix. While it isn't wrong, it seems out of place in Trace list and this PR has an easy fix to improve it.

**Without this PR:**

Before 1

Before 2

**With this PR:**

New 1

New 2

As you can see, the organization is still correct, but the span ID is now consistent with the rest of the span IDs shown in Trace list.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/TraceContextForLogEntry.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/TraceContextForLogEntry.cs
@@ -62,7 +62,7 @@ namespace Google.Cloud.Diagnostics.AspNetCore
         internal static TraceContextForLogEntry FromExternalTrace(IServiceProvider serviceProvider) =>
             serviceProvider?.GetService<IExternalTraceProvider>()?.GetCurrentTraceContext(serviceProvider);
 
-        private static string SpanIdToHex(ulong? spanId) => spanId is null ? null : string.Format("0x{0:X}", spanId);
+        private static string SpanIdToHex(ulong? spanId) => spanId is null ? null : $"{spanId:x}";
 
     }
 }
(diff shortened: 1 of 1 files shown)


EatonZ:
NOTE: Looks like there was actually a bug in the original implementation too.
In cases where the span ID started with a 0, it was left out. Updated the format string further to account for that.

jskeet:
Amanda can comment on this more, but I have a concern around compatibility: am I right in thinking customers expecting the existing behavior would be broken by this change? If so, we might need to make this opt-in.

EatonZ:
@jskeet I thought about that too, but I'm not sure what this could break.
The hex number is still technically the same - it's just cosmetically correct now.
This also brings it in line with the example noted in the `spanId` field [here](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry). Also in [your own documentation](https://github.com/googleapis/google-cloud-dotnet/blob/cf7bd95cdb6352d87c53b5ded8c3451397e171de/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/TraceContextForLogEntry.cs#L43). I haven't found any information to suggest that a "0x" suffix is correct in this field.

jskeet:
> I thought about that too, but I'm not sure what this could break.

Anything that's fetching the data later, and expecting it to be in the existing format.

We *could* decide that's okay - I just think it needs careful consideration.

amanda-tarafa:
Hmmm, this might have been my bad. I'm trying to remember why I picked this format but can't, I did explicitly picked it, I maybe just got confused.

- This is potentially breaking because the log filter is based on the string representation of the SpanID, anyone filtering logs based on the span ID value will be broken (that's probably rare use case btw).
- But as it is now, it's a bug. Logs are not being associated to their Span in the backend because of the format mismatch.
  - Integration tests didn't catch this because they are checking that the Span IDs on the span and the log entries are the same by converting the Span ID to this same format.
- Setting the Span ID was only introduced in #5199 and release on August 3rd 2020, we didn't set any Span ID before that. As of today, the v4.1.0 version of the library (which contains this change) has been downloades a little over 7k times, vs the 120k times of the previous version.

For all of the above, I think we should go ahead with the change. bumping minor and adding a comment on release notes.

This PR will break integration tests though which are not run on presubmit. @EatonZ I'll be closing so I can address this and a few of the things that have stemmed from #5275. I'll address this afternoon London time.

PR #5373, opened 2020-09-28
Generate constructors for gRPC APIs, then delete the tool and revert old changes

jskeet:


tools/Google.Cloud.Tools.ConstructorGenerator/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+        static void Main(string[] args)
+        {
+            var catalog = ApiCatalog.Load();
+
+            var generateAwkward = args.Length == 1 && args[0] == "--awkward";
+
+            var apis = catalog.Apis
+                // Only use gRPC APIs
+                .Where(api => api.Type == ApiType.Grpc)
+                // If the minor version is greater than 0, there must be a GA release.
```
amanda-tarafa: What about the ones that are GA on a zero minor? Like 3.0.0 GA? If we don't have any like those, this is fine being a one off.

tools/Google.Cloud.Tools.ConstructorGenerator/Program.cs
```diff
@@ -0,0 +1,128 @@
...
+                    {
+                        // All interpolated to make indentation easier to verify
+                        $"    public partial class {name}",
+                        $"    {{",
+                        $"        public partial class {name}Client",
+                        $"        {{",
+                        $"            /// <summary>Creates a new client using a channel.</summary>",
+                        $"            /// <param name=\"channel\">The channel to use to make remote calls.</param>",
+                        $"            [Obsolete(\"This constructor overload is present for compatibility only, and will be removed in the next major version\")]",
+                        $"            public {name}Client(Channel channel) : base(channel)",
```
amanda-tarafa: Not sure if it should be so that it calls the new constructor. If the new one is now doing stuff that this one did before, then I think this suggestion stands.

```suggestion
                        $"            public {name}Client(Channel channel) : this(channel)",
```

amanda-tarafa (approved):
LGTM, with a couple of comments.

chrisdunelm (approved):
LGTM

tools/Google.Cloud.Tools.ConstructorGenerator/Program.cs
jskeet: 3.0.0 GA is fine - api.IsReleaseVersion will be true. The idea is to generate constructors for "2.1.0-beta01" but not for "2.0.0-beta01".

tools/Google.Cloud.Tools.ConstructorGenerator/Program.cs
jskeet: The previous constructor had `base(channel)` so it's really just doing exactly the same as before the change. (And the other constructor also just uses `base(channel)`, so it makes no odds.)

PR #5377, opened 2020-09-28
[Testing helpers][Diagnostics]: Add namespace checks.

amanda-tarafa:


tools/Google.Cloud.ClientTesting/CodeHealthTester.cs
```diff
@@ -67,5 +67,33 @@ public static void AssertClassesAreSealedOrAbstract(Type sampleType, IEnumerable
             // Force output to show the bad types
             Assert.Equal(new string[0], badTypes.ToList());
         }
+
+        /// <summary>
+        /// Asserts that the assembly containing the given type does not define the disallowed namespaces.
+        /// </summary>
+        public static void AssertNoDisallowedNamespaces(Type sampleType, IEnumerable<string> disallowedNamespeces)
```
jskeet, line 74: If you make this a `params` array, you won't need the `new string[]` in the calling code.

tools/Google.Cloud.ClientTesting/CodeHealthTester.cs
```diff
@@ -67,5 +67,33 @@ public static void AssertClassesAreSealedOrAbstract(Type sampleType, IEnumerable
...
+        {
+            var badNamespaces = sampleType.Assembly.DefinedTypes
+                .Select(t => t.Namespace)
+                .Distinct()
+                .Intersect(disallowedNamespeces);
+
+            // Force output to show the bad namespaces.
+            Assert.Equal(new string[0], badNamespaces.ToList());
```
jskeet, line 82: I think `Assert.Empty(badNamespaces)` would show everything, wouldn't it

I think I'd be tempted to list the types themselves - so use `sampleType.Assembly.DefinedTypes.Where(t => disallowedNamespaces.Contains(t.Namespace))` and the equivalent for AssertOnlyAllowed.

tools/Google.Cloud.ClientTesting/CodeHealthTester.cs
jskeet, line 74: Typo nit (here and below) - namespaces, not namespeces

tools/Google.Cloud.ClientTesting/CodeHealthTester.cs
amanda-tarafa: Done.

amanda-tarafa (commented):
Comments addressed in original commits because it's all there is.

jskeet (approved):

PR #5382, opened 2020-09-30
For all but patch releases, update known dependencies

jskeet:
This change also expands the list of known defaulted dependencies.

(We don't *downgrade* a dependency though.)


amanda-tarafa (approved):
LGTM

PR #5390, opened 2020-10-01
Fix generation by removing a previous tweak

jskeet:
This is a legacy from preivous resource name generation (and API
configuration). The using directive is no longer required - and
actively interferes in this case as `ProjectName` is present in both
Grafeas.V1 and Google.Api.Gax.ResourceNames.


amanda-tarafa (approved):
LGTM

PR #5402, opened 2020-10-05
Delete previously-generated source code before generating

jskeet:
This means that if a .proto file is deleted from googleapis, we'll delete it rather than continuing to have a stale file.


chrisdunelm (approved):
LGTM

PR #5408, opened 2020-10-06
Delete files in apis, not api

jskeet:
Also, make the compgen invocation silent.

(Coincidentally, this was caught by an API changing in a way that *required* an old file to be deleted...)


amanda-tarafa (approved):
LGTM

PR #5422, opened 2020-10-12
Add Workflows and Workflow Executions APIs

jskeet:
Note that this doesn't *release* either API.

Currently the smoke tests are failing for me - I won't merge until that's fixed.
Both APIs have a WorkflowName class as a resource name - I will probably create a Google.Cloud.Workflows.Common.V1Beta (in the same way as we have for Spanner) with resource names in. That can happen after this "naive" generation though.


jskeet:
Smoke test fixed by changing the location in the test.

chrisdunelm (approved):
LGTM

PR #5431, opened 2020-10-13
Update Retail - change from V2 to V2Beta

jskeet:


jskeet:
(AppVeyor failure is due to trying to build the now-removed API.)

chrisdunelm (approved):
LGTM

chrisdunelm:
Just noticed that appveyor is failing with a missing project file.
Is this benign, and caused by the rename?

jskeet:
Yes, precisely that. I thought we'd fix that, but apparently not. Not going to look into it now, given how rarely we remove a directory.

PR #5432, opened 2020-10-13
Workflows common project for a resource name

jskeet:
Once this is in, we can release the Workflows projects.

(At this point I may add ReleaseManager commit support for multiple projects...)


chrisdunelm (approved):
LGTM

PR #5436, opened 2020-10-13
Add option for a smoke test to be skipped

### LINKED ISSUE #5433: Add a "skip" property to smoke test files?
Sometimes we have a smoke test that we expect to work eventually, but doesn't right now. It would be nice to be able to have the smoke test file present, but skip some of the tests when running.

jskeet:
(And apply this to retail.)

Fixes #5433.


amanda-tarafa (approved):
LGTM

PR #5443, opened 2020-10-14
Remove Google.Cloud.Retail.V2Beta

jskeet:
Plans have changed around when client libraries will be needed, and
there seems little point in merging autosynth commits in the meantime.

(I expect this to make AppVeyor fail, btw.)


amanda-tarafa (approved):
LGTM

PR #5448, opened 2020-10-14
Allow for more flexible commit handling

jskeet:
When we know at commit time that we want a different message, allow
that to be specified after a "Version history:" line.

When we only know after the fact (or can't change the commit), allow
overrides to be specified in a JSON file.

If the message is just "skip", the commit is skipped.

(Tested with DLP)

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/History/GitCommit.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/History/GitCommit.cs
@@ -13,7 +13,10 @@
 // limitations under the License.
 
 using LibGit2Sharp;
+using Newtonsoft.Json;
+using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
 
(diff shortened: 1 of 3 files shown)


tools/Google.Cloud.Tools.ReleaseManager/History/GitCommit.cs
```diff
@@ -62,6 +75,21 @@ internal IEnumerable<string> GetHistoryLines()
...
+            // followed by all the lines that should be in version history.
+            int overrideIndex = messageLines.IndexOf(VersionHistoryOverride);
+            if (overrideIndex != -1)
+            {
+                messageLines = messageLines.Skip(overrideIndex + 1).ToList();
+            }
+
+            // A single line of "skip" (usually from an override, either in the commit or in the overrides file)
+            // means we skip the commit entirely.
+            if (messageLines.Count == 1 && messageLines[0].Equals("skip", StringComparison.OrdinalIgnoreCase))
```
amanda-tarafa, line 88: Maybe add a constant for skip VersionHistorySkip, but that can be done later.

amanda-tarafa (approved):
LGTM!
This is great!

tools/Google.Cloud.Tools.ReleaseManager/History/GitCommit.cs
jskeet: Added it as SkipMessage now. Will merge on green.

PR #5454, opened 2020-10-15
Order the service directory entries by proto path

jskeet:
(Previously it was only by luck that we were getting some kind of ordering - but it wasn't fully deterministic.)


chrisdunelm (approved):
LGTM

jskeet:
Will trigger autosynth manually now

PR #5472, opened 2020-10-20
Show the updated history section when doing the updating

jskeet:
When everything is working correctly, this means we can just go
straight to commit, without manually checking the file.


chrisdunelm (approved):
LGTM

PR #5474, opened 2020-10-20
Prototype of passing server-provided fields to custom type converters

### LINKED ISSUE #5470: Support for Firestore Metadata fields (Id, CreateTimestamp, UpdateTimestamp, ReadTimestamp) in IFireStoreConverter<T>
I'm using a custom converter to convert my type to/from a document (IDictionary<string,object>) to avoid decorating my types with attributes. (I want my types to be shared in other libraries where I can't reference the Firestore lib.)

The custom converter works great, except I don't think it exposes the 4 metadata properties you can get using attribute serialization:

- FirestoreDocumentId
- FirestoreDocumentCreateTimestamp
- FirestoreDocumentUpdateTimestamp
- FirestoreDocumentReadTimestamp

It would be awesome of these 4 properties could be passed through the IFirestoreConverter<T> class

I don't see any way to get these values using IFirestoreConverter<T> today.

 **Additional context**
I was exploring this scenario with the below types/converter:
```cs

using AutoMapper;
using Google.Cloud.Firestore;
using System;
using System.Threading.Tasks;

namespace ConsoleApp1
{
    class Program
    {
        public static IMapper MapperInstance = new MapperConfiguration(cfg => { }).CreateMapper();

        static async Task Main(string[] args)
        {
            // Set GOOGLE_APPLICATION_CREDENTIALS env variable here

            FirestoreDbBuilder builder = new FirestoreDbBuilder();
            builder.ConverterRegistry = new ConverterRegistry();
            builder.ConverterRegistry.Add(new Converter());
            builder.ProjectId = "<project id>";

            FirestoreDb db = builder.Build();

            await db.Collection("Testing").AddAsync(new MyFireStoreType
            {
                Prop1 = "foo",
                Prop2 = 2
            });

            var result = await db.Collection("Testing").GetSnapshotAsync();

            MyFireStoreType deserializedType = result[0].ConvertTo<MyFireStoreType>();

            Console.WriteLine("Done");
        }
    }

    public sealed class MyFireStoreType : FireStoreBaseObject
    {
        public string Prop1 { get; set; }
        public int Prop2 { get; set; }
        public bool? Prop3 { get; set; }
    }

    public abstract class FireStoreBaseObject
    {
        public string FirestoreDocumentId { get; set; }

        public DateTimeOffset FirestoreDocumentCreateTimestamp { get; set; }

        public DateTimeOffset FirestoreDocumentUpdateTimestamp { get; set; }

        public DateTimeOffset FirestoreDocumentReadTimestamp { get; set; }
    }

    public class Converter : IFirestoreConverter<MyFireStoreType>
    {
        public MyFireStoreType FromFirestore(object value)
        {
            Dictionary<string, object> valueAsDict = (Dictionary<string, object>)value;
            // Issue - no Id, CreateTimestamp, UpdateTimestamp, ReadTimestamp values available in the above dictionary
            return Program.MapperInstance.Map<MyFireStoreType>(valueAsDict);
        }

        public object ToFirestore(MyFireStoreType value)
        {
            return Program.MapperInstance.Map<Dictionary<string, object>>(value);
        }
    }
}

```

jskeet:
Would fix #5470, if adopted. Needs documentation if we go ahead.

Concerns:

- Is the Firestore team happy with this?
- Should we just work out a way of making the deserialization
  context public?
- Is it okay for the attribute to allow multiple and be inherited?
  (I can imagine a TypeConverterBase specifying it, for example...
  but what if we later have extra properties on it that conflict?)
- Is it okay for this to be very tightly bound to the implementation
  type? (There's no flexibility here - you need to know the keys
  at compile-time. I think that's okay, but...)

cc @schmidt-sebastian for Firestore, @ryandl for verification that it would solve their issue.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ConverterRegistryTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/ConverterRegistryTest.cs
@@ -13,9 +13,7 @@
 // limitations under the License.
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using Xunit;
 using static Google.Cloud.Firestore.Tests.SerializationTestData;
 
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/Converters/CustomConverterTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/Converters/CustomConverterTest.cs
@@ -46,10 +46,64 @@ namespace Google.Cloud.Firestore.Tests.Converters
             Assert.Throws<InvalidOperationException>(() => CustomConverter.ForConverterType(typeof(NullReturningConverter), typeof(int)));
         }
 
+        [Fact]
+        public void DeserializeMap_ServerReturnedProperties()
+        {
+            var db = FirestoreDb.Create("proj", "db", new FakeFirestoreClient());
+            var context = new DeserializationContext(GetSampleSnapshot(db, "doc1"));
+            context.Snapshot.Document.Fields["doc-field"] = ProtoHelpers.CreateValue("FieldValue");
+            var converter = CustomConverter.ForConverterType(typeof(DictionaryHolderConverter), typeof(DictionaryHolder));
+            var holder = (DictionaryHolder) converter.DeserializeMap(context, context.Snapshot.Document.Fields);
+            var dictionary = holder.Dictionary;
+            // The regular field deserialized from the document
+            Assert.Equal("FieldValue", dictionary["doc-field"]);
+
+            // The server-provided values
+            Assert.Equal(context.DocumentReference, dictionary["DocumentId"]);
+            Assert.Equal(context.Snapshot.CreateTime, dictionary["CreateTime"]);
+            Assert.Equal(context.Snapshot.UpdateTime, dictionary["UpdateTime"]);
+            Assert.Equal(context.Snapshot.ReadTime, dictionary["ReadTime"]);
+        }
+
+        [Fact]
+        public void DeserializeValue_WhenMap_ServerReturnedProperties()
+        {
+            var db = FirestoreDb.Create("proj", "db", new FakeFirestoreClient());
+            var context = new DeserializationContext(GetSampleSnapshot(db, "doc1"));
+            var value = ProtoHelpers.CreateMap(("key1", ProtoHelpers.CreateValue("value1")), ("key2", ProtoHelpers.CreateValue("value2")));
+            var converter = CustomConverter.ForConverterType(typeof(DictionaryHolderConverter), typeof(DictionaryHolder));
+            var holder = (DictionaryHolder) converter.DeserializeValue(context, value);
+            var dictionary = holder.Dictionary;
+            // The regular fields deserialized from the map
+            Assert.Equal("value1", dictionary["key1"]);
+            Assert.Equal("value2", dictionary["key2"]);
+
+            // The server-provided values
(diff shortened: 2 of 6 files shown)


jskeet:
Have just noticed that we need to handle DeserializeValue as well as DeserializeMap... adding a second commit.

ryandle:
I have reviewed the changes and, Yes, this would definitely solve my scenario, thanks!

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreConverterConfigurationAttribute.cs
```diff
@@ -0,0 +1,72 @@
+﻿// Copyright 2019, Google LLC
```
amanda-tarafa, line 1: nit: 2020

amanda-tarafa (approved):
LGTM

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreConverterConfigurationAttribute.cs
jskeet: Thanks, will fix.

jskeet:
At a Firestore team member suggestion, I've added an extra commit which moves the attribute to be a method-targeted attribute instead of a type-targeted attribute.

We'd want to rename the attribute and change the comments for it, but I wanted to present the idea first.

cc @ryandle 
cc @amanda-tarafa 
cc @schmidt-sebastian

I'm on the fence at the moment.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreConverterConfigurationAttribute.cs
```diff
@@ -0,0 +1,72 @@
...
+using System;
+using System.Collections.Generic;
+
+namespace Google.Cloud.Firestore
+{
+    /// <summary>
+    /// Provides additional information for how Firestore converters (types implementing <see cref="IFirestoreConverter{T}"/>)
+    /// should be called.
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
```
amanda-tarafa, line 24: Why multiple?

amanda-tarafa (commented):
I think I like it being clearer that this is for deserialization only, althoguh that could also be achieved by changing the name of the class attribute to something like DesrializationConfig...
(Also, I seem to remember that overriding methods will inherit attributes, but not 100% certain about that though)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreConverterConfigurationAttribute.cs
jskeet: No particular reason, and definitely something to ask ourselves. I think it makes sense if it's inherited, in that a subclass may want to specify it, then a superclass add it later... if it's specified in multiple inheritance levels, I think it makes sense to allow it to be specified multiple times at one level too. But I think we should decide "method or class" for sure, then make those decisions (and experiment with overrides etc).

jskeet:
Okay, I've now rebased this to be a single commit that is method-targeted.

I've kept Multiple=true and specified Inherited=true - I think it makes sense to be able to compose the attribute however the user wants, basically.

I've added some basic documentation - I don't think we need a separate sample, although I could write one to match the existing one for document ID etc if we want. Let me know.

Happy to reconsider the attribute name, but otherwise this is good to go.

amanda-tarafa (approved):
LGTM. And agree with allowing full composition

jskeet:
Great - will merge this, then release a beta. If that works out for @ryandle we can then release a GA version in a week or so.

PR #5480, opened 2020-10-21
feat: add support for CommitStats

olavloite:
This draft adds support for returning `CommitStats` for a transaction. `CommitStats` must specifically be requested from the backend to be returned. The current implementation adds access to `CommitStats` as follows:
1. The property `SpannerConnection.ReturnCommitStats` is added and specifies whether transactions that are created on this connection should return commit stats.
1. The property `SpannerTransaction.ReturnCommitStats` is added and specifies whether this specific transaction should return commit stats. The default value is taken from the connection, but can be changed for a specific transaction.
1. The property `SpannerConnetion.LastCommitResponse` is added and contains the response for the last commit that was executed by a transaction on this connection.

The reason for adding the `ReturnCommitStats` property to `SpannerConnection` is that we want to be able to support `CommitStats` for commands that are executed without a transaction (i.e. for `EphemeralTransaction`).

There are a couple of possible alternative solutions that could also be considered:
1. The `ReturnCommitStats` property could be removed from `SpannerConnection` and be added to `SpannerCommand`. Setting it to true for a `SpannerCommand` only makes sense as long as the `SpannerCommand` does not use a transaction, as the command will not automatically commit. Instead, it should in that case be set on the `SpannerTransaction`.
1. The `LastCommitResponse` property could be removed from `SpannerConnection` and added as a `CommitResponse` property to `SpannerCommand` and `SpannerTransaction`.

The advantage of the alternative solution(s) is that the properties are defined directly on the objects that actually execute the commit. The disadvantage is that there are two different ways of requesting and inspecting commit statistics, depending on whether you use an explicit transaction or not.

Having the `ReturnCommitStats` on `SpannerConnection` (both as the only solution, or in combination with a specific property on `SpannerTransaction` and/or `SpannerCommand`), is that we could also allow users to define a value for the property in the connection string. That would allow users to turn commit stats on/off without having to change the code.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
@@ -115,7 +115,12 @@ namespace Google.Cloud.Spanner.Data.CommonTesting
             string value = Environment.GetEnvironmentVariable(name);
             return string.IsNullOrEmpty(value) ? defaultValue : value;
         }
-        
-        public SpannerConnection GetConnection() => new SpannerConnection(ConnectionString);        
+
+        public SpannerConnection GetConnection()
+        {
+            var conn = new SpannerConnection(ConnectionString);
+            conn.ReturnCommitStats = true;
+            return conn;
+        }
     }
 }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/CommitResponse.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/CommitResponse.cs
@@ -1,0 +1,79 @@
+﻿// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Google.Cloud.Spanner.Data
+{
+    /// <summary>
+    /// Immutable class representing a commit response.
+    /// </summary>
+    public sealed class CommitResponse
+    {
+        private const string MISSING_COMMIT_STATS = "The transaction did not return any CommitStats. Set Transaction.ReturnCommitStats=true to request commit statistics.";
+
+        /// <summary>
+        /// The proto representation of the commit response. Must not be mutated
+        /// or exposed publicly.
+        /// </summary>
+        internal V1.CommitResponse Proto { get; set; }
+
+        /// <summary>
+        /// The commit timestamp of the transaction.
+        /// </summary>
+        public DateTime CommitTimestamp { get => Proto.CommitTimestamp.ToDateTime(); }
+
+        /// <summary>
+        /// The number of mutations that was executed by the transaction. This count may not
+        /// exceed 20,000.
+        /// 
+        /// Insert and update operations count with the multiplicity of the number of columns they affect.
+        /// For example, inserting a new record may count as five mutations, if values are inserted into five columns.
+        /// 
+        /// Delete and delete range operations count as one mutation regardless of the number of columns affected.
+        /// 
+        /// Deleting a row from a parent table that has the ON DELETE CASCADE annotation is also counted as one mutation
+        /// regardless of the number of interleaved child rows present. The exception to this is if there are secondary
+        /// indexes defined on rows being deleted, then the changes to the secondary indexes will be counted individually.
+        /// 
+        /// For example, if a table has 2 secondary indexes, deleting a range of rows in the table will count as 1 mutation
+        /// for the table, plus 2 mutations for each row that is deleted because the rows in the secondary index might be
+        /// scattered over the key-space, making it impossible for Cloud Spanner to call a single delete range operation on
+        /// the secondary indexes. Secondary indexes include the foreign keys backing indexes.
+        /// </summary>
+        public long MutationCount
+        {
+            get
+            {
+                GaxPreconditions.CheckState(Proto.CommitStats != null, MISSING_COMMIT_STATS);
+                return Proto.CommitStats.MutationCount;
+            }
+        }
+
+        /// <summary>
+        /// The duration that the commit was delayed due to overloaded servers.
+        /// </summary>
+        public TimeSpan OverloadDelay
+        {
+            get
+            {
+                GaxPreconditions.CheckState(Proto.CommitStats != null, MISSING_COMMIT_STATS);
+                return Proto.CommitStats.OverloadDelay.ToTimeSpan();
+            }
+        }
+    }
+}
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/CommitResponse.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/CommitResponse.cs
@@ -24,13 +24,18 @@ namespace Google.Cloud.Spanner.Data
     /// </summary>
     public sealed class CommitResponse
     {
-        private const string MISSING_COMMIT_STATS = "The transaction did not return any CommitStats. Set Transaction.ReturnCommitStats=true to request commit statistics.";
+        private const string MISSING_COMMIT_STATS = "The transaction did not return any CommitStats. Set SpannerConnection.ReturnCommitStats=true to request commit statistics.";
+
+        internal CommitResponse(V1.CommitResponse proto)
+        {
+            Proto = GaxPreconditions.CheckNotNull(proto, nameof(proto));
+        }
 
         /// <summary>
         /// The proto representation of the commit response. Must not be mutated
         /// or exposed publicly.
         /// </summary>
-        internal V1.CommitResponse Proto { get; set; }
+        internal V1.CommitResponse Proto { get; }
 
         /// <summary>
         /// The commit timestamp of the transaction.
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
@@ -61,6 +61,9 @@ namespace Google.Cloud.Spanner.Data
         // State used for TransactionScope-based transactions.
         private VolatileResourceManager _volatileResourceManager;
 
+        // The response for the last commit on this connection.
+        private CommitResponse _commitResponse;
+
         /// <inheritdoc />
         public override string ConnectionString
         {
@@ -115,6 +118,22 @@ namespace Google.Cloud.Spanner.Data
         /// running SQL and streaming SQL requests.
         /// </summary>
         public QueryOptions QueryOptions { get; set; }
+
+        /// <summary>
+        /// Request CommitStats for all read/write transactions throughout the
+        /// lifetime of the connection.
+        /// </summary>
+        public bool ReturnCommitStats { get; set; }
+
+        /// <summary>
+        /// The response of the last commit on this connection. This includes the
+        /// response of any implicit commits by commands without a transaction.
+        /// </summary>
+        public CommitResponse LastCommitResponse
+        {
+            get => _commitResponse;
+            internal set => _commitResponse = value;
+        }
 
         /// <summary>
         /// Creates a SpannerConnection with no datasource or credential specified.
@@ -284,7 +303,7 @@ namespace Google.Cloud.Spanner.Data
         public async Task<TResult> RunWithRetriableTransactionAsync<TResult>(Func<SpannerTransaction, Task<TResult>> asyncWork, CancellationToken cancellationToken = default)
         {
             GaxPreconditions.CheckNotNull(asyncWork, nameof(asyncWork));
-            
+
             await OpenAsync(cancellationToken).ConfigureAwait(false);
             RetriableTransaction transaction = new RetriableTransaction(
                 this,
@@ -790,7 +809,7 @@ namespace Google.Cloud.Spanner.Data
         private Action GetTransactionEnlister()
         {
             Transaction current = Transaction.Current;
-            return current == null ? (Action) null : () => EnlistTransaction(current);
+            return current == null ? (Action)null : () => EnlistTransaction(current);
         }
 
         /// <summary>
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
@@ -61,8 +61,8 @@ namespace Google.Cloud.Spanner.Data
         // State used for TransactionScope-based transactions.
         private VolatileResourceManager _volatileResourceManager;
 
-        // The response for the last commit on this connection.
-        private CommitResponse _commitResponse;
+        // The response for the last commit (explicit or implicit) on this connection.
+        private CommitResponse _lastCommitResponse;
 
         /// <inheritdoc />
         public override string ConnectionString
@@ -127,12 +127,17 @@ namespace Google.Cloud.Spanner.Data
 
         /// <summary>
         /// The response of the last commit on this connection. This includes the
-        /// response of any implicit commits by commands without a transaction.
+        /// commit response of any implicit commits by commands that are executed
+        /// without a transaction.
+        ///
+        /// The property is null after the execution of a command that uses a
+        /// transaction, and will return the commit response of that transaction
+        /// when the transaction is committed.
         /// </summary>
         public CommitResponse LastCommitResponse
         {
-            get => _commitResponse;
-            internal set => _commitResponse = value;
+            get => _lastCommitResponse;
+            internal set => _lastCommitResponse = value;
         }
 
         /// <summary>
@@ -303,7 +308,7 @@ namespace Google.Cloud.Spanner.Data
         public async Task<TResult> RunWithRetriableTransactionAsync<TResult>(Func<SpannerTransaction, Task<TResult>> asyncWork, CancellationToken cancellationToken = default)
         {
             GaxPreconditions.CheckNotNull(asyncWork, nameof(asyncWork));
-
+            
             await OpenAsync(cancellationToken).ConfigureAwait(false);
             RetriableTransaction transaction = new RetriableTransaction(
                 this,
@@ -809,7 +814,7 @@ namespace Google.Cloud.Spanner.Data
         private Action GetTransactionEnlister()
         {
             Transaction current = Transaction.Current;
-            return current == null ? (Action)null : () => EnlistTransaction(current);
+            return current == null ? (Action) null : () => EnlistTransaction(current);
         }
 
         /// <summary>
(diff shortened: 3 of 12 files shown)


olavloite:
@jskeet @amanda-tarafa 
I would really appreciate your input on the considerations in the description of this PR before continuing with additional tests etc. WDYT?

(The current PR only includes some additional asserts to existing integration tests. I do intend to add separate tests for this feature, but I would like to first get an agreement on what would be the best approach to add this feature.)

@skuruppu Please feel free to also add any additional input you might have.

jskeet:
Will try to have a look later today.

amanda-tarafa:
I'll take a look later as well, thanks!

olavloite:
(The CI error is caused by the emulator, which does not yet support returning commit stats)

amanda-tarafa:
Haven't looked at the PR yet, but here are some initial thoughs based on the PR description.

> we want to be able to support CommitStats for commands that are executed without a transaction (i.e. for EphemeralTransaction)

Is this a hard requirement? I think we could agree that inspecting (in code) commit stats is an advanced feature, so it doesn't seem unreasonable to me to require an explicit transaction to access that feature. If a user is not even setting up an explicit transaction it is probably because they are not interested in what happens with it. If we drop this requirement, the feature surface and its implementation becomes so much simpler and straightforward. But not only that, I'm also worried about how `SpannerConnetion.LastCommitResponse` would behave in a multithreaded environment where the connection is shared and several transactions may be open at any given time.

I would keep `SpannerConnection.ReturnCommitStats` as a default (and as pointed out can be included in the connection string) and `SpannerTransaction.ReturnCommitStats` as a means to override that default for specific transactions, but I would add `ISpannerTransaction.CommitStats` instead of any of the other alternatives to store the actual commit stats, notice that with this approach, the stats would also be available for EphemeralTransactions, only thing is that EphemeralTransactions are not inmediately exposed.

But going further, I would say that the main use of this feature is for diagnostics purposes? Hardly any production code will execute depending on whether the commit stats were this or that? So, the question is, do we really need to expose the commit stats via code? Or could we think of a way of making them easily added to logs and turned on and off from the connection string? We already have a [rudimentary logging mechanism](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Spanner.Data/docs/logging.md) in place for diagnostic purposes, so, could we take advantage of that to expose these stats? And also, at the end of the day, if some user really really needs to access commit stats in code, they can always use the lower level library (custom retries are already exposed via the lower level library).

amanda-tarafa (commented):
Have done a quick review.
Main worry is the same, exposing commit stats through the connection is not going to work on multithreaded environments. I definetely thinkg they belong in the transaction, not the command either, commands can be reused.

olavloite:
@amanda-tarafa 
Thanks for the elaborate answer.

>  But not only that, I'm also worried about how SpannerConnetion.LastCommitResponse would behave in a multithreaded environment

Yes, I was worried about that as well. I tried to find a definite answer to whether a `DbConnection` implementations are normally thread safe or not, but with at best mixed results. If a `SpannerConnection` is (intended to be) shareable between multiple threads, the `LastCommitResponse` is a non-starter.

> Is this [support for ephemeral transactions] a hard requirement?

@skuruppu That's in the end for you to say.

The design document does not specifically mention it, probably because most (all?) other client libraries do not implement ephemeral transactions in the same way as the .NET client does. There are a number of situations that are (very) similar where we do (intend to) support it:
* JDBC driver: The JDBC driver supports autocommit and will support `CommitStats` in that mode as well. This will be done by implementing a `LastCommitResponse` style solution. JDBC connections are by design not thread safe and can have at most one transaction active at any time, so we don't have to worry about shared connections or multiple simultaneous transactions.
* Java, Node, and Go client libraries: All these libraries support some sort of `write(Enumerable<Mutation>)` method. These methods write a list of mutations in what is basically an ephemeral transaction and can return `CommitStats`.

> But going further, I would say that the main use of this feature is for diagnostics purposes?

That is my understanding as well. Users sometimes run into the limit of max 20,000 mutations in one transaction, and as that mutation count is quite difficult to calculate when you have multiple secondary indices etc., this feature will allow users to get a better idea on what the actual mutation count is for specific transactions / statements. Being able to turn this on/off using configuration is not a hard requirement, but could be nice to have to debug an application that occasionally runs into this limit.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Spanner.cs
```diff
@@ -6066,6 +6072,22 @@ public sealed partial class CommitRequest : pb::IMessage<CommitRequest>
       get { return mutations_; }
     }
 
+    /// <summary>Field number for the "return_commit_stats" field.</summary>
```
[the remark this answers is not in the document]
olavloite: The changes to the protos have not yet been released, so I manually regenerated the client. I'll do that in a separate commit the next time.

olavloite:
@amanda-tarafa 

>  I definetely thinkg they belong in the transaction, not the command either, commands can be reused.

I would say that the fact that a command can be reused is not necessarily a blocker for letting a user inspect commit stats and other metadata of the ephemeral transaction that was used for the command, but it would be if a command is potentially reused by multiple threads. I don't think that is the intended usage of a `DbCommand` and/or `SpannerCommand`.
The user would of course have to inspect the property after executing the command, as it would be lost / overwritten by the results of the next execution if the command is reused.

amanda-tarafa:
> @amanda-tarafa
> 
> > I definetely thinkg they belong in the transaction, not the command either, commands can be reused.
> 
> I would say that the fact that a command can be reused is not necessarily a blocker for letting a user inspect commit stats and other metadata of the ephemeral transaction that was used for the command, but it would be if a command is potentially reused by multiple threads. I don't think that is the intended usage of a `DbCommand` and/or `SpannerCommand`.
> The user would of course have to inspect the property after executing the command, as it would be lost / overwritten by the results of the next execution if the command is reused.

Oh no, the command is not intended to be used concurrently by several threads, it wouldn't actually work as expected, as in it is not thread safe. But it can be reused, and I think it'll be confusing to have the commit stats there, in context B you might have access to some commit stats from context A that you don't know anything about, etc. Also, the fact that the commit stats would be populated in the command at different points in time (after the command executed if under ephemeral transactions or after the transaction commited if under explicit transaction) adds yet another level of confusion.

In my view, from a design perspective, the commit stats belong with the transaction. But I still would prefer not even exposing them there, but having a mechanism to log them or not.

olavloite:
> Also, the fact that the commit stats would be populated in the command at different points in time (after the command executed if under ephemeral transactions or after the transaction commited if under explicit transaction) adds yet another level of confusion.

Sorry, I wasn't completely clear on that:
1. I would only populate the `CommitStats` for a command if it was executed using an ephemeral transaction.
1. `CommitStats` for a transaction should definitely go in the `SpannerTransaction` class (considering the fact that the `LastCommitResponse` solution is a non-starter).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
```diff
@@ -115,7 +115,12 @@ private static string GetEnvironmentVariableOrDefault(string name, string defaul
             string value = Environment.GetEnvironmentVariable(name);
             return string.IsNullOrEmpty(value) ? defaultValue : value;
         }
-        
-        public SpannerConnection GetConnection() => new SpannerConnection(ConnectionString);        
+
+        public SpannerConnection GetConnection()
```
jskeet, line 119: I'd probably still use the expression-bodied member here:

```csharp
GetConnection() => new SpannerConnection(ConnectionString) { ReturnCommitStatus = true };
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/CommitResponse.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Text;
+
+namespace Google.Cloud.Spanner.Data
+{
+    /// <summary>
+    /// Immutable class representing a commit response.
+    /// </summary>
+    public sealed class CommitResponse
+    {
+        private const string MISSING_COMMIT_STATS = "The transaction did not return any CommitStats. Set SpannerConnection.ReturnCommitStats=true to request commit statistics.";
```
jskeet, line 27:
```suggestion
        private const string MissingCommitStatsMessage = "The transaction did not return any CommitStats. Set SpannerConnection.ReturnCommitStats=true to request commit statistics.";
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/CommitResponse.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// or exposed publicly.
+        /// </summary>
+        internal V1.CommitResponse Proto { get; }
+
+        /// <summary>
+        /// The commit timestamp of the transaction.
+        /// </summary>
+        public DateTime CommitTimestamp { get => Proto.CommitTimestamp.ToDateTime(); }
+
+        /// <summary>
```
jskeet, line 45: If the intention of the blank lines is to signify multiple paragraphs, it would be best to use `<para>` tags. I suspect that only the first paragraph belongs in `summary` though, and the rest should be in `remarks`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/CommitResponse.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// For example, if a table has 2 secondary indexes, deleting a range of rows in the table will count as 1 mutation
+        /// for the table, plus 2 mutations for each row that is deleted because the rows in the secondary index might be
+        /// scattered over the key-space, making it impossible for Cloud Spanner to call a single delete range operation on
+        /// the secondary indexes. Secondary indexes include the foreign keys backing indexes.
+        /// </summary>
+        public long MutationCount
+        {
+            get
+            {
+                GaxPreconditions.CheckState(Proto.CommitStats != null, MISSING_COMMIT_STATS);
```
jskeet, line 67: I'd be tempted to write a private property that returns Proto.CommitStats after validating it - then these properties (and any others that can up) can be as simple as (say)

```csharp
public long MutationCount => CheckedCommitStats.MutationCount;
```

We may not want to have a separate constant for the exception message then, either.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -116,6 +119,27 @@ public override ConnectionState State
...
+        /// lifetime of the connection.
+        /// </summary>
+        public bool ReturnCommitStats { get; set; }
+
+        /// <summary>
+        /// The response of the last commit on this connection. This includes the
+        /// commit response of any implicit commits by commands that are executed
+        /// without a transaction.
+        ///
+        /// The property is null after the execution of a command that uses a
```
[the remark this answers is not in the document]
jskeet: To be fair, I'm not sure whether we actually expect SpannerConnection itself to be used across multiple threads in parallel. (Certainly SpannerCommand and the result sets don't expect to handle multiple threads.) The normal pattern should be create/use/dispose... but we should at least see what we do elsewhere.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -116,6 +119,27 @@ public override ConnectionState State
...
+        /// transaction, and will return the commit response of that transaction
+        /// when the transaction is committed.
+        /// </summary>
+        public CommitResponse LastCommitResponse
```
jskeet, line 137: This can just be an automatically implemented property, so we don't need the field declaration.

jskeet (commented):
I'm not *too* worried by the threading aspect - but it does feel like an awkward place to put the property.

One alternative might be to have a "commit response sink" or something like that - effectively treat it as an event handler for "I got some commit stats". But I think I need to know more about the use case before judging the API further.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/CommitResponse.cs
```diff
@@ -0,0 +1,84 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Google.Cloud.Spanner.Data
+{
+    /// <summary>
+    /// Immutable class representing a commit response.
+    /// </summary>
+    public sealed class CommitResponse
```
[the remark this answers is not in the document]
skuruppu: There is a chance that we'd add additional fields to the response in the future. So in all languages, we're keeping this super generic to make sure we can accommodate any new response fields. Let me know if this might be overkill. I'm just paranoid because in some languages, this became a breaking change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -116,6 +119,27 @@ public override ConnectionState State
         /// </summary>
         public QueryOptions QueryOptions { get; set; }
 
+        /// <summary>
+        /// Request CommitStats for all read/write transactions throughout the
+        /// lifetime of the connection.
+        /// </summary>
+        public bool ReturnCommitStats { get; set; }
```
skuruppu, line 126: To give some additional context that I may not have made clear in the spec, this feature is mostly required for users to debug their queries. So there's no specific need to set this at a connection level. Users should just be able to set it at a command level if they want to debug a query. But if it makes the implementation simpler to have it at a connection level, then I'm not opposed to it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -137,6 +137,7 @@ internal bool HasMutations
             Mode = mode;
             _session = GaxPreconditions.CheckNotNull(session, nameof(session));
             TimestampBound = timestampBound;
+            ReturnCommitStats = connection.ReturnCommitStats;
```
skuruppu, line 140: If we decide not to set this as a connection level, then you won't need this.

jskeet:
Given @skuruppu's comments, I wonder whether it would make sense to have both `ReturnCommitStats` and `LastCommitResponse` (or `LastCommitStat`) *just* on `SpannerCommand`, and the setting automatically propagated to `SpannerTransaction`.

While a `SpannerCommand` can be reused, that would typically be more obvious in code (it's unlikely that a user will have an instance variable of type `SpannerCommand` for example, whereas they may do so for `SpannerConnection` even if I'd prefer they didn't).

Does the `SpannerCommand`/`SpannerTransaction` interaction work in a way that would allow the public API change *only* to affect `SpannerCommand`? Or are transactions used without the user ever being aware of a `SpannerCommand`? (It's been a while since I've looked at all of this in detail, and a surface-level browse would probably miss things.)

olavloite:
> Given @skuruppu's comments, I wonder whether it would make sense to have both `ReturnCommitStats` and `LastCommitResponse` (or `LastCommitStat`) _just_ on `SpannerCommand`, and the setting automatically propagated to `SpannerTransaction`.
> 
> While a `SpannerCommand` can be reused, that would typically be more obvious in code (it's unlikely that a user will have an instance variable of type `SpannerCommand` for example, whereas they may do so for `SpannerConnection` even if I'd prefer they didn't).
> 
> Does the `SpannerCommand`/`SpannerTransaction` interaction work in a way that would allow the public API change _only_ to affect `SpannerCommand`? Or are transactions used without the user ever being aware of a `SpannerCommand`? (It's been a while since I've looked at all of this in detail, and a surface-level browse would probably miss things.)

I think it would be doable from a technical perspective, but I'm not sure it would be a very user-friendly solution. Providing access to commit stats on `SpannerCommand` makes sense for commands that are executed without a transaction, as it basically means that the command itself is a transaction. For actual transactions, the user will execute multiple `SpannerCommands` within the transaction and then call commit on the transaction. The user would then have to inspect the last command that was executed in the transaction to find the commit stats. (The commit call itself is not a `SpannerCommand`).

jskeet:
> For actual transactions, the user will execute multiple SpannerCommands within the transaction and then call commit on the transaction.

Ah, yes, of course. That makes sense. Thanks.

So to summarise:

- We need it on SpannerCommand because some transactions are implicit
  - We *could* decide that in order to get the stats, you have to be explicit about the transaction. That would simplify things, yes?
- We need it on SpannerTransaction because the stats are effectively transaction-wide

Is that a reasonable summary, or have I still missed things? I do think it would be best *not* to put it on SpannerConnection until we have a clear and reasonable use case for it. Basically I'm in favour of deferring API surface changes until they're necessary, so long as we don't do anything to *prevent* them being added later on. I'd like to take the same approach to SpannerCommand, although I can see that's a slightly harder sell.

amanda-tarafa:
What about Jon's proposal yesterday of a sink or similar. I still like this better because it does away with the commit stats being a property of the command (and of transaction).

- Add an Func<CommitResponse, Task> OnCommitAsync to both the command and the transaction, that gets called when either is commited. This means that the CommitResponse is not part of the state of the command.
- As an alternative, which I like better, it could be Func<IList<CommandInfo>, CommitResponse, Task> OnCommitAsync added to the connection itself. CommandInfo is a new type that contains the description of the command, as in, what type of command, the SQL that got executed and the parameters. With CommandInfo we avoid sending the command itself which could potentially have been changed when it's process by the commit handler.
(And in both cases it would be a Func returning a task so it can be async).

olavloite:
@jskeet Your summary is spot on.

Not supporting the feature for ephemeral transactions would certainly simplify it, but I don't think it is a reasonable thing to do. If a user wants to debug a query that is using an ephemeral transaction in an existing application, it is quite a big ask to require the user to change the existing application code to (temporarily) use explicit transactions at that point.

The event model sound promising, and my preference would also be to implement it as @amanda-tarafa suggested in the second bullet, i.e. as an event on the `SpannerConnection`. That would mean that there is only one place where `CommitResponse` can be accessed, instead of on both `SpannerCommand` and `SpannerTransaction`.

skuruppu:
> We could decide that in order to get the stats, you have to be explicit about the transaction. That would simplify things, yes?

It depends. The idea would be for the user to quickly take some production code, add a parameter to set ReturnCommitStats=true and be able to get a result. But if it's not too tricky to start/end a transaction then I suppose that's fine. We can show how to do it in a code sample.

amanda-tarafa:
> > We could decide that in order to get the stats, you have to be explicit about the transaction. That would simplify things, yes?
> 
> It depends. The idea would be for the user to quickly take some production code, add a parameter to set ReturnCommitStats=true and be able to get a result. But if it's not too tricky to start/end a transaction then I suppose that's fine. We can show how to do it in a code sample.

Risking sounding repetitive, @skuruppu the description of the use case you just gave it's spot on "diagnostics". This seems more and more to be something the user could activate with a flag **in production** and get some logs to inspect.
Adding the information in code as we have been discussing requieres either:

- The user debugging production code almost certainly in a different environment than production, which has zero guatantees of reproducing the issue they want to diagnose (or they have to go through hoops to reproduce the production environment or debug in it).
- Or the user writing code to capture the commit stats in code and push them to a log so they can have the production information available. But we can do this for them, and make the logs available with a flag switch I think.

Would this be something that you'd considered? What to me feels off about this feature on the higher level library is that we have a lot of code to make things transparent for users, hide transactions when they don't explicitly use them, hide transations that retry on failures, etc. And then we want to expose very low level information about those same things that we are hiding. That's why it is so hard design wise to come up with a clean surface, I think.

skuruppu:
@amanda-tarafa I actually added a comment on our spec doc to the feature team to check whether they're ok with this being a logging option. I'll report back with what they say.

I think I agree in terms of debugability that logging this result would be perfectly reasonable. There is a risk though of the behavior being different across client libraries. I wanted to check with the team on whether that would be acceptable.

skuruppu:
Just to close out the discussion, yes, we think it's reasonable for the `CommitResponse` to be logged in the ADO.NET layer. But the values should only be logged if the user sets the ReturnCommitStats property (we might want to call it LogCommitStats) and either of the values in the commit stats is not null.

Hope that makes sense.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
olavloite: It makes it a lot easier for `EphemeralTransaction`s, as those transactions are not exposed to the client application.

PR #5490, opened 2020-10-26
GitHub actions

jskeet:
Okay, ready for review now.

Two commits:

- First commit just adds the GitHub actions
- Second commit fixes a long-standing issue where any change that doesn't affect `apis/` would build everything.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -1,0 +1,49 @@
+name: Build and test PR (changed APIs only)
+
+on: [pull_request]	
+
+jobs:	
+
+  build:	
+    runs-on: ubuntu-18.04	
+    env:	
+      DOTNET_NOLOGO: true	
+    strategy:	
+      matrix:	
+        regex: ["'Google\\.Cloud\\.[A-L].*'", "'Google\\.Cloud\\.[M-Z].*'", "'!Google\\.Cloud'"]	
+    steps:	
+    - uses: actions/checkout@v2	
+      with:	
+        submodules: true	
+        fetch-depth: 100	
+
+    # We use the .NET Core 2.1 runtime for testing most packages
+    - name: Setup .NET Core 2.1	
+      uses: actions/setup-dotnet@v1	
+      with:	
+        dotnet-version: 2.1.x	
+
+    # We build with .NET Core 3.1 SDK	
+    - name: Setup .NET Core 3.1	
+      uses: actions/setup-dotnet@v1	
+      with:	
+        dotnet-version: 3.1.x
+
+    - name: Setup Python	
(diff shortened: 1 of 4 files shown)


.github/workflows/build-pr.yml
```diff
@@ -0,0 +1,53 @@
...
+    strategy:	
+      matrix:	
+        regex: ["'Google\\.Cloud\\.[A-L].*'", "'Google\\.Cloud\\.[M-Z].*'", "'!Google\\.Cloud'"]	
+    steps:	
+    - uses: actions/checkout@v2	
+      with:	
+        submodules: true	
+        fetch-depth: 100	
+
+    # We need just the .NET Core 2.1 runtime for testing
```
amanda-tarafa, line 20: We need .Net Core 3 for Diagnostics.AspNetCore3

.github/workflows/build-pr.yml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Setup Python	
+      uses: actions/setup-python@v2	
+      with:	
+        python-version: 3.x	
+
+    # The GitHub checkout action leaves the repo in a slightly awkward
+    # state. This tidies it up.
+    - name: Set up git branches
+      run: |
+    # The GitHub checkout action leaves the repo in a slightly awkward
```
amanda-tarafa, line 41: This comment and the next instruction is repeated starting on line 37

.github/workflows/build-push.yml
```diff
@@ -0,0 +1,43 @@
...
+    env:	
+      DOTNET_NOLOGO: true	
+
+    steps:	
+    - uses: actions/checkout@v2	
+      with:	
+        submodules: true	
+        fetch-depth: 100	
+
+    # We need just the .NET Core 2.1 runtime for testing
```
amanda-tarafa, line 21: Same comment re .NET Core 3

amanda-tarafa (approved):
LGTM with a couple of comments

.github/workflows/build-pr.yml
jskeet: The comment is meant to indicate that we don't need .NET Core 2.1 for building - only for testing. (It's "just the runtime".)
Will leave now rather than making a more complex comment.

.github/workflows/build-pr.yml
jskeet: Whoops, thanks, will remove.

.github/workflows/build-pr.yml
jskeet: Updated the comment in the end :)

jskeet:
Thanks, will merge on green. I suggest we wait about a week before actually disabling Travis.

PR #5498, opened 2020-10-28
Clean output directories for desktop frameworks when running on Linux

jskeet:
Fixes #5497.

Note that this will still block us from using GitHub actions on
Windows, unless we can get more disk space or switch to "build+test"
one API at time, instead of building everything then testing
everything.


amanda-tarafa (approved):
LGTM

PR #5528, opened 2020-11-06
feat: Enable server side flow control by default with the option to turn it off

fayssalmartanigcp:
This change enables sending flow control settings automatically to the server. If FlowControlSettings.MaxOutstandingElementCount > 0 or FlowControlSettings.MaxOutstandingByteCount > 0, flow control will be enforced at the server side (in addition to the client side).

This behavior is enabled by default and users who would like to opt-out of this feature --in case they encounter issues with server side flow control-- can set Settings.UseLegacyFlowControl=True in SubscriberClient.CreateAsync().

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
@@ -89,6 +89,12 @@ namespace Google.Cloud.PubSub.V1
             public FlowControlSettings FlowControlSettings { get; set; }
 
             /// <summary>
+            /// If set to true, disables enforcing flow control settings at the Cloud PubSub server
+            /// and uses the less accurate method of only enforcing flow control at the client side.
+            /// </summary>
+            public bool UseLegacyFlowControl { get; set; } = false;
+
+            /// <summary>
             /// The lease time before which a message must either be ACKed
             /// or have its lease extended. This is truncated to the nearest second.
             /// If <c>null</c>, uses the default of <see cref="DefaultAckDeadline"/>.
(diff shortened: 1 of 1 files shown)


jskeet:
I'll have a look at this on Monday (I'm on vacation today) but I'm a bit nervous of "it's okay, users can turn it off if it breaks something" as an approach to compatibility. Will probably email internally for more details.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -992,8 +1002,8 @@ private void HandleStartStreamingPullWithoutBackoff()
                 {
                     SubscriptionAsSubscriptionName = _subscriptionName,
                     StreamAckDeadlineSeconds = _modifyDeadlineSeconds,
-                    MaxOutstandingMessages = _flow.MaxOutstandingElementCount,
-                    MaxOutstandingBytes = _flow.MaxOutstandingByteCount
+                    MaxOutstandingMessages = _useLegacyFlowControl ? 0 : _flow.MaxOutstandingElementCount,
```
jskeet, line 1005: Are you sure this is the right way round? Given the description of the PR, I'd expected UseLegacyFlowControl=true to be the "do what we did before" setting, whereas this change makes the *default* be "as it is now".

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -88,6 +88,12 @@ internal Settings(Settings other)
             /// </summary>
             public FlowControlSettings FlowControlSettings { get; set; }
 
+            /// <summary>
+            /// If set to true, disables enforcing flow control settings at the Cloud PubSub server
+            /// and uses the less accurate method of only enforcing flow control at the client side.
+            /// </summary>
+            public bool UseLegacyFlowControl { get; set; } = false;
```
jskeet, line 95: If this property effectively makes the FlowControlSettings property redundant/ignored, it would probably be worth stating that in this property *and* in the FlowControlSettings property above. It's possible I've misunderstood though.

jskeet (commented):
I still need to understand how likely it is that this will break something, but even aside from this it *looks* like the logic may be the wrong way round - I may well be missing something though.

jskeet (approved):
Okay, I now understand. We'll want careful release notes, but I can merge this now.

PR #5536, opened 2020-11-09
API Index prototype

jskeet:


jskeet:
Here's the JSON index created by the prototype.

[api-index-v1.zip](https://github.com/googleapis/google-cloud-dotnet/files/5510621/api-index-v1.zip)

ApiIndex/index.proto
```diff
@@ -0,0 +1,183 @@
...
+// The overall API index. The index file
+// is a JSON representation of this message.
+// See the glossary for more details about the
+// terminology used in this schema.
+message Index {
+  // All the known APIs, ordered lexicographically by ID.
+  repeated Api apis = 1;
+  
+  // Metadata about the current API index.
+  IndexMetadata metadata = 2;
```
amanda-tarafa, line 29: What is the relationship between this metadata and the package version `google.tools.apiindex.v1`. Am I right in thinking that they will move at the same pace and IndexMetadata is just there so that it's programatically easy to know the version of the generated index?

ApiIndex/ApiModel.cs
```diff
@@ -0,0 +1,218 @@
...
+                    // Note: we assume there are no "nested" APIs
+                    foreach (var subdir in System.IO.Directory.GetDirectories(directory))
+                    {
+                        AddServiceConfigsRecursively(subdir);
+                    }
+                }
+            }
+        }
+
+        public V1.Api ToV1Api() =>
```
amanda-tarafa, line 142: Might be overkill, but can this be moved to ApiIndex.V1 as an extension method or by creating a new class there that inherits from `ApiIndex.ApiModel`.
If anything the name `ToV1Api` is very confusing, at least to me, I read the version as if it were the API version and not the ApiIndex version.
Also, if different ApiIndex versions get to coexist, does that mean that there'll be several `ToV*Api` methods in this class?
And as it stands, if we want to get rid of ApiIndex.V1 (in favour of ApiIndex.V2 for instance), this would still have to take a breaking change to remove this public method, so the apparent independece of `Google.Cloud.Tools.ApiIndex` from the index protos, is really not there if using this pattern for the conversion methods.

ApiIndex/Index.cs
```diff
@@ -0,0 +1,1786 @@
...
+//     source: index.proto
+// </auto-generated>
+#pragma warning disable 1591, 0612, 3021
+#region Designer generated code
+
+using pb = global::Google.Protobuf;
+using pbc = global::Google.Protobuf.Collections;
+using pbr = global::Google.Protobuf.Reflection;
+using scg = global::System.Collections.Generic;
+namespace Google.Cloud.Tools.ApiIndex.V1 {
```
amanda-tarafa, line 12: nit: Should this file be generated in an `ApiIndex/V1` folder?

ApiIndex/Program.cs
```diff
@@ -0,0 +1,71 @@
...
+            FileDescriptorSet descriptorSet = FileDescriptorSet.Parser.ParseFrom(descriptorSetInput);
+
+            var apiModels = ApiModel.Load(descriptorSet, args[1]);
+
+            WriteV1Index(args[2], apiModels);
+
+            return 0;
+        }
+
+        private static void WriteV1Index(string outputDirectory, IEnumerable<ApiModel> apiModels)
```
amanda-tarafa, line 46: Maybe this whole project needs to be ApiIndex.V1, and if at some point several index versions coexist, the common code can be extracted to a library or something?

ApiIndex/ServiceModel.cs
```diff
@@ -0,0 +1,51 @@
...
+    public class ServiceModel
+    {
+        private ServiceDescriptor service;
+
+        public ServiceModel(ServiceDescriptor service)
+        {
+            this.service = service;
+        }
+
+        public V1.Service ToV1Service() => new V1.Service
```
amanda-tarafa, line 33: Here, same comment as for ApiModel.

ApiIndex/ServiceModel.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            FullName = service.FullName,
+            ShortName = service.Name,
+            Methods = { service.Methods.OrderBy(method => method.Name).Select(ConvertMethod) }
+        };
+
+        private static V1.Method ConvertMethod(MethodDescriptor method) => new V1.Method
```
amanda-tarafa, line 40: And if things are staying as they are, even though this method is private, it'd be good if it was called `ConvertToV1Method` for consistency.

amanda-tarafa (commented):
I know this is a draft still, but I've left a couple of comments.

jskeet:
(Will respond to comments tomorrow, probably.)

ApiIndex/ApiModel.cs
jskeet: We can't make it an extension method, but we could create a partial method of `FromModel` in `ApiIndex.V1.Api` potentially. I tend to find "To" conversions more discoverable than "From" conversions though. How about `ToProtoV1` as a name to use everywhere?

> Also, if different ApiIndex versions get to coexist, does that mean that there'll be several ToV*Api methods in this class?

Yes.

> And as it stands, if we want to get rid of ApiIndex.V1 (in favour of ApiIndex.V2 for instance), this would still have to take a breaking change to remove this public method

This isn't a library, so "breaking change" isn't really a problem.

The intention is that these model classes should always contain everything required to generate "all current schema versions", and have that information in the most convenient form. That just makes it easier to support multiple versions without too much duplication.

ApiIndex/Program.cs
jskeet: I'd definitely prefer to anticipate multiple versions from the start, separating out "everything we can find out" from "the representation in the schema at the moment".

ApiIndex/ServiceModel.cs
jskeet: Or `ToProtoV1Method` in this case, perhaps. But yes, completely agree. (Or I could make it a private method within the `ToV1Proto` method, in which case it doesn't need the version part.)

ApiIndex/index.proto
jskeet: There's one difference, in terms of compatibility: it's possible that `google.tools.apiindex.v2.IndexMetadata` could have more information, but we *mustn't* make breaking changes within the `IndexMetadata` message. (We could *potentially* deprecate something in v2 and remove it in v3, but I hope we don't need to.)

Tools should be able to read any version of the schema by ignoring JSON fields they don't understand, but they should always be able to understand "enough" of the metadata to find out the version and report if it's deprecated.

ApiIndex/index.proto
```diff
@@ -0,0 +1,183 @@
...
+  // the same directory, that directory only appears once.
+  repeated string import_directories = 8;
+
+  // The values for a specific option from across all protos contributing
+  // to this API. The option name itself is the key in the `Api.options`
+  // field.
+  message OptionValues {
+    // Map from "a value for the option" to "the number of protos in
+    // which that value occurred". Where an option is omitted entirely in
+    // one or more protos (but does occur in other ones), the key is an
```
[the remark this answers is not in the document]
jskeet: Thanks, will address once we've decided on the conversion method names.

ApiIndex/ServiceModel.cs
```diff
@@ -0,0 +1,51 @@
...
+        public ServiceModel(ServiceDescriptor service)
+        {
+            this.service = service;
+        }
+
+        public V1.Service ToV1Service() => new V1.Service
+        {
+            FullName = service.FullName,
+            ShortName = service.Name,
+            Methods = { service.Methods.OrderBy(method => method.Name).Select(ConvertMethod) }
```
martinvarelaj, line 37: As discussed by email, it would be nice to add here the google.api.http (https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L29) message included in the custom options field of the rpc.

For our use-case a simplified JSON representation could be a flat list of repeated HttpRule messages with only the "oneof pattern" and body. e.g:

https://github.com/googleapis/googleapis/blob/master/google/cloud/iot/v1/device_manager.proto#L107

```
...
"httpRules": [{
  "patch": "/v1/{device.name=projects/*/locations/*/registries/*/devices/*}",
  "body": "device"
},{
  "patch": "/v1/{device.name=projects/*/locations/*/registries/*/groups/*/devices/*}",
  "body": "device"
}]
...
```

jskeet: Having discussed this internally, we do indeed need something where different bindings can have different HTTP methods. We don't need to represent the body, however.

I'll create another commit that uses a representation of:

```json
"httpRules": [{
  "method": "patch",
  "path": "/v1/{device.name=projects/*/locations/*/registries/*/devices/*}"
},{
  "method": "patch",
  "path": "/v1/{device.name=projects/*/locations/*/registries/*/groups/*/devices/*}"
}]
```

(I'd expect that having the method as the value of a field rather than as a key in the object should be easier to consume.)

jskeet:
Now implemented method bindings. Sample index attached.
[api-index-v1.zip](https://github.com/googleapis/google-cloud-dotnet/files/5702533/api-index-v1.zip)

jskeet:
(This now has its own repo, so the prototype isn't needed here as a PR or branch.)

PR #5575, opened 2020-11-14
feat: add support for request and transaction tagging to Spanner

olavloite:
Adds support for request and transaction tagging. Request tags are set on `SpannerCommand` and `SpannerBatchCommand`. Transaction tags are set on `SpannerTransactions` and are automatically included for all requests that are executed on the transaction, including the commit request. Commit requests cannot include a separate request tag.

Ephemeral transactions and read-only transactions do not support transaction tags. Commands that are executed using an ephemeral or read-only transaction can include a request tag.

Example usage:

```cs
var requestTag = "request-tag-1";
var transactionTag = "transaction-tag-1";
SpannerConnection connection = ...;
SpannerTransaction transaction = connection.BeginTransaction();
transaction.TransactionTag = transactionTag;

var command = connection.CreateSelectCommand("SELECT * FROM FOO");
command.Transaction = transaction;
command.RequestTag = requestTag;
using (var reader = command.ExecuteReader())
{
  // do something with the data
}
transaction.Commit();
```


olavloite:
cc @skuruppu

jskeet:
We'll look on Monday

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -157,7 +157,7 @@ public void CommandHasOptimizerVersionSetOnCommand()
                 // Optimizer version set at a command level has higher precedence
                 // than version set through the connection or the environment
                 // variable.
-                Mock<SpannerClient> spannerClientMock = SetupExecuteStreamingSql(cmdOptimizerVersion);
+                Mock<SpannerClient> spannerClientMock = SetupExecuteStreamingSql("foo"); //, cmdOptimizerVersion);
```
jskeet, line 160: This line looks like it probably needs to be reverted

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -173,6 +173,135 @@ public void CommandHasOptimizerVersionSetOnCommand()
...
+            var command1 = connection.CreateSelectCommand("SELECT * FROM FOO");
+            command1.Transaction = transaction;
+            using (var reader = command1.ExecuteReader())
+            {
+                Assert.True(reader.HasRows);
+            }
+            try
+            {
+                // Try to set the transaction tag.
+                transaction.TransactionTag = transactionTag;
```
jskeet, line 290: Instead of all of this, just use:

```csharp
Assert.Throws<InvalidOperationException>(() => transaction.TransactionTag = transactionTag);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RequestOptions.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    /// <summary>
+    /// Immutable class representing request options.
+    /// </summary>
+    internal class RequestOptions : IEquatable<RequestOptions>
+    {
+        /// <summary>
+        /// The request tag configured in the options.
+        /// </summary>
+        internal string RequestTag { get => Proto.RequestTag; }
```
jskeet, line 28: This can be simplified to

```csharp
internal string RequestTag => Proto.RequestTag;
```

(Ditto for TransactionTag.)

I'm not sure why GitHub isn't offering me the "suggestion" option...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RequestOptions.cs
```diff
@@ -0,0 +1,105 @@
...
+            var protoCopy = Proto.Clone();
+            protoCopy.TransactionTag = tag ?? "";
+            return new RequestOptions(protoCopy);
+        }
+
+        /// <summary>
+        /// The proto representation of the request options. Must not be mutated
+        /// or exposed publicly.
+        /// </summary>
+        internal V1.RequestOptions Proto { get; }
```
jskeet, line 67: Does this need to be internal rather than private, given that we have the ToProto method?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RequestOptions.cs
```diff
@@ -0,0 +1,105 @@
...
+            return new RequestOptions(proto.Clone());
+        }
+
+        /// <summary>
+        /// Get proto version of the request options.
+        /// </summary>
+        internal V1.RequestOptions ToProto() => Proto.Clone();
+
+        /// <inheritdoc />
+        public bool Equals(RequestOptions other) => other is object
```
jskeet, line 95: This is inconsistent with GetHashCode() at the moment - I'd either specify individual fields that you're interested in in both, or just use proto equality, e.g.

```csharp
public bool Equals(RequestOptions other) => Proto.Equals(other?.Proto);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -38,6 +38,7 @@ public sealed class SpannerTransaction : DbTransaction, ISpannerTransaction
         private readonly List<Mutation> _mutations = new List<Mutation>();
         private DisposeBehavior _disposeBehavior = DisposeBehavior.ReleaseToPool;
         private bool _disposed = false;
+        private bool _used = false;
```
jskeet, line 41: Could you either give this a more specific name, or add a comment to clarify its exact meaning?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -126,6 +127,24 @@ internal bool HasMutations
...
+        /// <summary>
+        /// The transaction tag to use for this transaction. This can only be set for read/write transactions,
+        /// and must be set before any statements are executed on the transaction.
+        /// </summary>
+        public string TransactionTag
+        {
+            get => _commitRequestOptions.TransactionTag;
+            set
+            {
+                GaxPreconditions.CheckState(Mode != TransactionMode.ReadOnly, "Transaction tag cannot be set on a read-only transaction");
```
jskeet, line 142: I don't think there's a test for this check... it would be useful as documentation.

jskeet (commented):
Almost entirely fine - just a few nits. Amanda will review as well - I assume we wouldn't want to merge until the protos have gone public (removing the need for the first commit) anyway.

amanda-tarafa (commented):
I've left some comments in code, but I do have a general question/comment.

Why do we need the wrapper class? Wouldn't it be clearer to just add a `string Tag` property in `SpannerCommand`, `SpannerBatchCommand` and `Transaction`, and then use that to create the `RequestOptions` proto when building the `ExecuteSqlRequest` message?

If the aim is to mimick `QueryOptions`, there's a huge difference between that and `RequestOptions` as it stands, and that is that there's a public settable `QueryOptions` property in `SpannerCommand` that user code can change at will. With `RequestOptions` we are wrapping the proto to then hide the `RequestOptions` field so we can only expose what's needed. I can think of a way of better mimicking `QueryOptions` and that is to have `CommandOptions` and `TransactionOptions` classes, that both wrap the `RequestOptions` proto (or not) and only expose what's relevant for commands or trasanctions. We could then have public properties of those options in `SpannerCommand` and `SpannerTransaction` respectively and when building the `ExecuteSqlRequest` we could merge those properties to obtain the `RequestOptions` proto that we needed. This could also be a solution that would help dealing with the mixed nature of the `RequestOptions` proto without having to add individual properties to `SpannerCommand` and `SpannerTransaction` whenever a new option is available for either or both.

Just my two cents, and I might be missing something.

skuruppu:
I think the @amanda-tarafa's idea of separating the wrappers for each kind of option is neat and may help prevent users from setting options based on the context. I think the C++ team had done a similar thing for timestamp bounds which has a similar issue.

amanda-tarafa:
Thanks @olavloite I'll take a look tomorrow.

jskeet:
I'm on vacation tomorrow, but I'll look again on Monday - again, I'm assuming that as this depends on proto changes that haven't been published yet, it's not super urgent.

olavloite:
@amanda-tarafa @jskeet Thanks both for the valuable input. And yes, Monday is certainly no problem.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -378,10 +380,12 @@ private ExecuteSqlRequest GetExecuteSqlRequest()
                     return Partition.ExecuteSqlRequest;
                 }
 
+                RequestOptions.TransactionTag = Transaction?.TransactionTag ?? "";
```
amanda-tarafa, line 383: nit: Maybe encapsulate this in a GetEffectiveRequestOptions? Although can also be done later if the options get more complicated.

amanda-tarafa (commented):
Yep, this is way cleaner now. Thanks.

A couple more tests I would like to add:

- Mostly for documentation purposes, tests that check when the tags are set to `null`, it's really set to `string.Empty`. This is somewhat of an "unexpected" behaviour due only to how the underlying protos work so it'll be nice that we have tests for this so we don't inadvertently modify it in the future. Or, implement the separate `CommandOptions` and `TransactionOptions` classes that could easily hide this behaviour from client code.

- A unit test for `RunWithRetryableTransaction` in which the user provided lambda sets the transaction tag before executing any command. Make the code retry at least once so that we can check that the lambda can always set the transaction tag, i.e. that we are not passing the old transaction instance with the `_hasExecutedCommands` set to `true` because it was used on the previous run.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
olavloite: Let's save that one for when `RequestOptions` get somewhat more complicated.

olavloite:
> Yep, this is way cleaner now. Thanks.
> 
> A couple more tests I would like to add:
> 
> * Mostly for documentation purposes, tests that check when the tags are set to `null`, it's really set to `string.Empty`. This is somewhat of an "unexpected" behaviour due only to how the underlying protos work so it'll be nice that we have tests for this so we don't inadvertently modify it in the future. Or, implement the separate `CommandOptions` and `TransactionOptions` classes that could easily hide this behaviour from client code.
> * A unit test for `RunWithRetryableTransaction` in which the user provided lambda sets the transaction tag before executing any command. Make the code retry at least once so that we can check that the lambda can always set the transaction tag, i.e. that we are not passing the old transaction instance with the `_hasExecutedCommands` set to `true` because it was used on the previous run.

Good point, those would certainly be useful. I've added tests for both.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -224,6 +224,18 @@ public override UpdateRowSource UpdatedRowSource
         /// </summary>
         public QueryOptions QueryOptions { get; set; }
 
+        // Request options for this command. Only the relevant properties are publicly exposed.
+        internal RequestOptions _requestOptions = RequestOptions.Empty;
+
+        /// <summary>
+        /// The statement tag to send to Cloud Spanner for this command.
+        /// </summary>
+        public string RequestTag
```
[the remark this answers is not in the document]
skuruppu: Yeah, I understand what you mean @amanda-tarafa. But I think @olavloite is right here in the sense that there will be UI that shows the stats and it would then be confusing for customers when they can't filter on a "Command Tag".

skuruppu (approved):

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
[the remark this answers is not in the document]
skuruppu: @syeduguri would you be able to give some information on how users will be able to see this data? Would it be too confusing if we called it a CommandTag?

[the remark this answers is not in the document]
syeduguri: Yes. CommandTag will be confusing. I think we can rename it as just "Tag".

amanda-tarafa (approved):
LGTM.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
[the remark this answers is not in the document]
olavloite: I've renamed both `RequestTag` and `TransactionTag` to `Tag`.

olavloite:
I've updated this PR to use the same approach as #5657, i.e. only create `RequestOptions` when building the request, and use `string`s to hold the tags in `SpannerCommand`, `SpannerBatchCommand` and `SpannerTransaction`.

jskeet:
Proto options now merged in https://github.com/googleapis/google-cloud-dotnet/pull/6285

olavloite:
> Proto options now merged in #6285

The PR has been merged with the latest changes from master and the manually generated code has been removed from the PR.

Any last thoughts / questions / comments on this PR from anyone?

amanda-tarafa:
Still LGTM :)

PR #5577, opened 2020-11-14
test: optimizer version tests did not actually verify the input

### LINKED ISSUE #5576: [Spanner]: CommandTests for QueryOptions don't actually verify the optimizer version that is used
The tests for checking that the correct optimizer version is used for queries do not actually verify this. Changing the value for the optimizer version [here](https://github.com/googleapis/google-cloud-dotnet/blob/23bc6204765e6a9aba0f8295362b0059b91694ff/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs#L116) to any random value will not make the test fail. This is also the case for the other tests verifying the optimizer version.

The problem is only related to the actual verification of the optimizer version. The correct version __is__ used by the client.

olavloite:
The optimizer version tests for `SpannerCommand` did not actually verify the optimizer version that was used. This PR fixes that by adding verifications for invocations on the `SpannerClient` mock.

Fixes #5576


jskeet:
Will look on Monday.

jskeet (approved):

PR #5580, opened 2020-11-16
Recreate the common storage test bucket

### LINKED ISSUE #5548: [Storage]: Unskip system tests when cross language bucket is restored.
For some system tests we use a cross language bucket named `storage-library-test-bucket`. This bucket has either become inaccessible or has been deleted, and that makes several of our tests fail.

I will now skip tests until the bucket is restored or an alternate bucket is found.

PHP has run into the same, see googleapis/google-cloud-php#3475.

@frankyn as FYI, I will start an internal thread for this.

jskeet:
Fixes #5548.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.CreateCommonTestBucket/Program.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.CreateCommonTestBucket/Program.cs
@@ -1,0 +1,126 @@
+﻿// Copyright 2020 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.CreateCommonTestBucket/Program.cs
```diff
@@ -0,0 +1,126 @@
...
+            catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.NotFound)
+            {
+                // Need to create the bucket...
+                // (But let's do so outside the catch.)
+            }
+            Console.WriteLine($"Creating bucket '{_bucket}'");
+            var serviceAccount = GoogleCredential.GetApplicationDefault().UnderlyingCredential as ServiceAccountCredential;
+            if (serviceAccount is null)
+            {
+                throw new InvalidOperationException("Default credentials are not for a service account, so we can't determine the project ID");
```
amanda-tarafa, line 76: nit: Or take the projectId as an arg.
But really totally fine as is.

amanda-tarafa (approved):
LGTM

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.CreateCommonTestBucket/Program.cs
jskeet: I did wonder about that - it's basically an option if we find it's a problem :)

PR #5588, opened 2020-11-17
Simplify the "add new API" experience

jskeet:
We always use "Initial release." as the message for the first
release, because there are no changes. This tooling change makes
that automatic.

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/History/HistoryFile.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/History/HistoryFile.cs
@@ -97,6 +97,8 @@ namespace Google.Cloud.Tools.ReleaseManager.History
 
         public sealed class Section
         {
+            private static readonly StructuredVersion s_expectedInitialReleaseVersion = StructuredVersion.FromString("1.0.0-beta01");
+
             public StructuredVersion Version { get; }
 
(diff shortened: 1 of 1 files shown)


tools/Google.Cloud.Tools.ReleaseManager/History/HistoryFile.cs
```diff
@@ -97,6 +97,8 @@ internal List<Section> MergeReleases(List<Release> releases)
 
         public sealed class Section
         {
+            private static readonly StructuredVersion ExpectedInitialReleaseVersion = StructuredVersion.FromString("1.0.0-beta01");
```
chrisdunelm, line 100: nit: Is this the right casing/naming convention for a static field??

chrisdunelm (approved):
LGTM - one nitty question.

amanda-tarafa (approved):
LGTM

tools/Google.Cloud.Tools.ReleaseManager/History/HistoryFile.cs
jskeet: Well, it's immutable, so it's sort of a constant... we go back and forth on this, to be honest.
Will change it to s_ though.

PR #5622, opened 2020-11-19
Use "skip" overrides in show-stale tool

jskeet:
Also adds some more skip commits for Ruby/PHP namespaces


amanda-tarafa (approved):
LGTM. Yep, this is good!

Issue #5325, opened 2020-09-15
Add labels on a per-log basis to Stackdriver upload via Log4net


williecohen-google:
**Is your feature request related to a problem? Please describe.**
I am using Log4net with the GoogleStackdriverAppender in a .NET framework project. I want to be able to add labels to our [log entries](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) for Cloud Monitoring.  I want to be able to group logs easily using these labels (which is what their purpose seems to be). However, the GoogleStackdriverAppender only allows static custom labels to be added that apply to all logs flowing through the Appender.  

 **Describe the solution you'd like**
Ideally there would be a way to attach labels to certain logs, and not just statically for all logs generated. I'm assuming this decision was made to try and "fit" Log4net nicely into the Stackdriver data model, but I would think there would be a way to accommodate per log labels.

 **Describe alternatives you've considered**
* Creating multiple appenders. (Seems like a terrible idea?)
* Creating structured logs to create our own labels at the proto payload level. (Seems like a band aid fix)

amanda-tarafa:
I'll take a look at whether we can do this and how, but just to set expectations, I won't get to it for at least a couple of weeks.

williecohen-google:
Any updates? We are running into log query latency issues and using log labels would speed things up dramatically (because they are indexed).

jskeet:
@williecohen-google: Sorry, not yet - we'll try to have a look next week.

amanda-tarafa:
I'm working on this, will have an update early next week.

amanda-tarafa:
I believe you can already achieve this using properties, custom labels and patterns for custom labels.

- Configure the `GoogleStackdriverAppender` to use patterns for custom labels by setting the `UsePatternWithinCustomLabels` property to true.
- Add costume labels that use patterns to reference properties, like such:
  `{ "userName", "%property{userName}" }`
- Use any of the log4net [`GlobaContext`](https://logging.apache.org/log4net/release/sdk/html/T_log4net_GlobalContext.htm), [`ThreadContext`](https://logging.apache.org/log4net/release/sdk/html/T_log4net_ThreadContext.htm) and [`LogicalThreadContext`](https://logging.apache.org/log4net/release/sdk/html/T_log4net_LogicalThreadContext.htm) to set properties.

[...]

Issue #1451, opened 2017-09-14
Users want client-side encryption for Google Cloud Storage.


SurferJeffAtGoogle:

jskeet:
Do we have more information about this - a precise set of requirements? It'll be hard to do any work on it otherwise.

jskeet:
Closing for now having added this to the [backlog](https://github.com/GoogleCloudPlatform/google-cloud-dotnet/blob/master/BACKLOG.md)

SurferJeffAtGoogle:
The client needs to do envelope encryption and store the KEK, IV, and KML key url in the file metadata.  The backlog seems like a good home for this feature request.

amanda-tarafa:
This is being evaluated across client libraries.
Removing from backlog.

Issue #1038, opened 2017-05-12
Write correlation-id with log entry metadata


JustinBeckwith:
By writing a correlation-id into the logging metadata, users can get request-correlated logs in pantheon. See https://gist.github.com/JustinBeckwith/43dd04cfbc4551ddd8996a5b803da0b0 for details.

iantalarico:
cc: @csells

jskeet:
Ian: I've set the priority of this as P2, but feel free to adjust up...

iantalarico:
@JustinBeckwith The link you have here 404s now.  Do you have updated context on this?

amanda-tarafa:
This has been addressed for several use cases already in #2120 and #5388. 
We can expand for different use cases if there's demand, but this issue can be considered addressed.
Removing from backlog.

PR #5628, opened 2020-11-19
chore: Backlog update.

amanda-tarafa:
- Removes #110, #803, #1038 and #1451
 - Adds #5325

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -7,12 +7,7 @@ corresponding issues open, but we'll reopen the issue if we revisit
 the request and want to put work into it.
 
 
-## Write correlation-id with log entry metadata
-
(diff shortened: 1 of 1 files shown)


BACKLOG.md
```diff
@@ -24,7 +19,7 @@ the request and want to put work into it.
 - Status: Further investigation required. We already have
   some analysis in the form of custom Roslyn analyzers.
   
-## Storage client-side encryption
+## Storage: Client-side encryption
```
jskeet, line 22: Perhaps we should remove this one as well, given internal work?

jskeet (approved):
One suggestion for another entry to remove, but I'm fine for you to either do that and then merge, or merge as-is.

BACKLOG.md
amanda-tarafa: Ah yes, makes sense.

Issue #5634, opened 2020-11-24
[Google.Cloud.Logging.NLog] Nlog logging structure


orchowski:
Hi!
 **Problem description**
I have the following structure at stack driver
![image](https://user-images.githubusercontent.com/17548765/100084798-8264c880-2e4b-11eb-8574-97d357f6e47d.png)
using the following config:
![image](https://user-images.githubusercontent.com/17548765/100084979-c1931980-2e4b-11eb-9453-96dfbfeebf14.png)
and it's pretty understandable but I have to meet some standards where properties are in a flat structure.
Something like that is expected:
```
  "jsonPayload": {    
    "someProperty": "propertyValue",
    "message": "2020-11-24 11:50:32.192 | INFO | InfoTrace | traceId:  | Hawk Server-1.0.0: started ",
   ...
  },
```
where **someProperty** is near the message in structure. Isn't it better to require a full JSON layout from API clients to define **jsonPayload** exactly as they want?

 **Describe the solution you'd like**
I have two solutions actually:

1. **Accept JsonLayout as user desire:**
``` 
<layout xsi:type="JsonLayout" includeAllProperties="true">
                    <attribute name="message" layout="${defaultMessageFormat}" />
                    <attribute name="log_time" layout="${var:timeStampFormat}" />
                    <attribute name="team" layout="${var:team:default=dotnet}" />
                    <attribute name="service" layout="${var:serviceName:default=anonymous}" />
  </layout>
```
but it can be a breaking change, for that

2. By default all properties will be placed at properties object but with some prefixes like `.someProperty` properties are going to be at the root object

**Additional context**
It's something I really need now, but maybe there is a possibility to do that somehow?
I can implement that, but I think greenlight is needed before. Thanks in advance for your help and time.

amanda-tarafa:
I'll take a look as soon as I can, but just to set expectations, it might not be this week.

orchowski:
@amanda-tarafa ok! Thanks <3

jskeet:
Actually, I believe we already have the necessary flexibility available.

You can specify the `JsonConverterTypeName` and `JsonConverterMethodName` options to make the specify your own converter. They should identify a method (either an instance method in a type that can be instantiated using `Activator.CreateInstance` or a static method) that accepts `object` and returns `Google.Protobuf.WellKnownTypes.Value` (which is effectively a JSON value).

That method will be used to provide the JSON payload. Please could you try that and get back to us if you have any problems?

[...]

PR #5635, opened 2020-11-26
Google.Cloud.Logging.Nlog Adds possibility to use custom jsonLayout

orchowski:
As we discussed here: https://github.com/googleapis/google-cloud-dotnet/issues/5634

Now there is a possibility to use a custom JSON payload structure as the user desire.
Eg.
image

It's breaking nothing for that library users don't have to change config files and so on.

@jskeet I called flag to enable this feature **UseNLogCustomJsonLayout**. It informs users about what they should do if the current JSON layout doesn't meet their needs. I mean JsonLayout type from NLog's library.

### DIFF
--- a/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/nlog-jsonTemplate.xml
+++ b/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/nlog-jsonTemplate.xml
@@ -1,0 +1,36 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+  <extensions>
+    <add assembly="Google.Cloud.Logging.NLog" />
+  </extensions>
+  
+  <targets async="true">
+    <target type="GoogleStackdriver" projectId="PROJECT_ID" logId="LOG_ID" name="stackdriver" resourceType="container" sendJsonPayload="true" enableJsonLayout="true" includeCallSite ="true" includeCallSiteStackTrace="true" IncludeEventProperties="false" IncludeMdlc="true">
+      <contextproperty name="contextProp" layout="propContextLayout" />
+      <layout xsi:type="JsonLayout" includeAllProperties="true">
+        <attribute name="someProp" layout="prop" />
+        <attribute name="nestedProp" encode="false">
+          <layout type='JsonLayout'>
+            <attribute name='insideNested1' layout='hello nested1!' />
+            <attribute name='insideNested2' layout='hello nested2!' />
+            <attribute name="nestedNestedProp" encode="false">
+              <layout type='JsonLayout'>
+                <attribute name='insideNestedNested1' layout='hello NestedNested1!' />
+                <attribute name='insideNestedNested2' layout='hello NestedNested2!' />
+              </layout>
+            </attribute>
+          </layout>
+        </attribute>
+      </layout>
+    </target>
+    <target name="console" xsi:type="Console" />
+  </targets>
+
+  <rules>
+    <logger name="*" minlevel="Info" writeTo="stackdriver" />
+    <logger name="*" minlevel="Info" writeTo="console" />
+  </rules>
+  
+</nlog>
--- a/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
+++ b/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
@@ -491,43 +491,55 @@ namespace Google.Cloud.Logging.NLog
             if (SendJsonPayload)
             {
                 var jsonStruct = new Struct();
-                jsonStruct.Fields.Add("message", Value.ForString(RenderLogEvent(Layout, loggingEvent)));
-                if (ServiceContextName != null)
+                if (EnableJsonLayout)
                 {
-                    var serviceName = RenderLogEvent(ServiceContextName, loggingEvent);
-                    if (!string.IsNullOrEmpty(serviceName))
+                    foreach (var jToken in JObject.Parse(RenderLogEvent(Layout, loggingEvent)))
                     {
-                        // Include ServiceContext to allow errors to be automatically forwarded
-                        var serviceVersion = RenderLogEvent(ServiceContextVersion, loggingEvent);
-                        if (string.IsNullOrEmpty(serviceVersion))
-                        {
-                            serviceVersion = "0.0.0.0";
-                        }
-
-                        var serviceContext = new Struct();
-                        jsonStruct.Fields.Add("serviceContext", Value.ForStruct(serviceContext));
-                        serviceContext.Fields.Add("service", Value.ForString(serviceName));
-                        serviceContext.Fields.Add("version", Value.ForString(serviceVersion));
+                        jsonStruct.Fields.Add(jToken.Key, _jsonConvertFunction(jToken.Value));
                     }
                 }
-
-                var propertiesStruct = new Struct();
-                jsonStruct.Fields.Add("properties", Value.ForStruct(propertiesStruct));
-
-                foreach (var combinedProperty in GetAllProperties(loggingEvent).Where(x => !string.IsNullOrEmpty(x.Key)))
+                else
                 {
-                    Value jsonValue;
-                    try
+                    jsonStruct.Fields.Add("message", Value.ForString(RenderLogEvent(Layout, loggingEvent)));
+                    
+                    if (ServiceContextName != null)
                     {
-                        jsonValue = _jsonConvertFunction(combinedProperty.Value);
+                        var serviceName = RenderLogEvent(ServiceContextName, loggingEvent);
+                        if (!string.IsNullOrEmpty(serviceName))
+                        {
+                            // Include ServiceContext to allow errors to be automatically forwarded
+                            var serviceVersion = RenderLogEvent(ServiceContextVersion, loggingEvent);
+                            if (string.IsNullOrEmpty(serviceVersion))
+                            {
+                                serviceVersion = "0.0.0.0";
+                            }
+
+                            var serviceContext = new Struct();
+                            jsonStruct.Fields.Add("serviceContext", Value.ForStruct(serviceContext));
+                            serviceContext.Fields.Add("service", Value.ForString(serviceName));
+                            serviceContext.Fields.Add("version", Value.ForString(serviceVersion));
+                        }
                     }
-                    catch (Exception ex)
+
+                    var propertiesStruct = new Struct();
+                    jsonStruct.Fields.Add("properties", Value.ForStruct(propertiesStruct));
+
+                    foreach (var combinedProperty in GetAllProperties(loggingEvent).Where(x => !string.IsNullOrEmpty(x.Key)))
                     {
-                        InternalLogger.Warn(ex,
-                            "GoogleStackdriver(Name={0}): Exception at BuildLogEntry with Key={1}", Name, combinedProperty.Key);
-                        jsonValue = Value.ForString($"<Exception: '{ex.Message}'>");
+                        Value jsonValue;
+                        try
+                        {
+                            jsonValue = _jsonConvertFunction(combinedProperty.Value);
+                        }
+                        catch (Exception ex)
+                        {
+                            InternalLogger.Warn(ex,
+                                "GoogleStackdriver(Name={0}): Exception at BuildLogEntry with Key={1}", Name, combinedProperty.Key);
+                            jsonValue = Value.ForString($"<Exception: '{ex.Message}'>");
+                        }
+
+                        propertiesStruct.Fields.Add(combinedProperty.Key, jsonValue);
                     }
-                    propertiesStruct.Fields.Add(combinedProperty.Key, jsonValue);
                 }
                 logEntry.JsonPayload = jsonStruct;
             }
--- a/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget_Configuration.cs
+++ b/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget_Configuration.cs
@@ -117,10 +117,17 @@ namespace Google.Cloud.Logging.NLog
 
         /// <summary>
         /// Fills <see cref="LogEntry.JsonPayload"/> instead of <see cref="LogEntry.TextPayload"/>.
-        /// JSON serialization can be customed using <see cref="JsonConverterTypeName"/> and <see cref="JsonConverterMethodName"/>,
+        /// JSON values serialization can be customized using <see cref="JsonConverterTypeName"/> and <see cref="JsonConverterMethodName"/>,
         /// or <see cref="JsonConverter"/>.
         /// </summary>
         public bool SendJsonPayload { get; set; }
+
+        /// <summary>
+        /// Fills <see cref="LogEntry.JsonPayload"/> in specified by user way with <see cref="T:NLog.Layouts.JsonLayout"/> layout type.
+        /// JSON values serialization can be customized using <see cref="JsonConverterTypeName"/> and <see cref="JsonConverterMethodName"/>,
+        /// or <see cref="JsonConverter"/>.
+        /// </summary>
+        public bool EnableJsonLayout { get; set; }
 
         /// <summary>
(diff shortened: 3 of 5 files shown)


apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget_Configuration.cs
```diff
@@ -117,11 +117,18 @@ public partial class GoogleStackdriverTarget
...
         /// or <see cref="JsonConverter"/>.
         /// </summary>
         public bool SendJsonPayload { get; set; }
 
+        /// <summary>
+        /// Fills <see cref="LogEntry.JsonPayload"/> in specified by user way with <see cref="T:NLog.Layouts.JsonLayout"/> layout type.
+        /// JSON values serialization can be customized using <see cref="JsonConverterTypeName"/> and <see cref="JsonConverterMethodName"/>,
+        /// or <see cref="JsonConverter"/>.
+        /// </summary>
+        public bool UseNLogCustomJsonLayout { get; set; }
```
snakefoot, line 130: Could this be called `EnabledJsonLayout` ? It would match other NLog targets doing similar stuff.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
```diff
@@ -491,7 +491,18 @@ private LogEntry BuildLogEntry(LogEventInfo loggingEvent)
             if (SendJsonPayload)
             {
                 var jsonStruct = new Struct();
-                jsonStruct.Fields.Add("message", Value.ForString(RenderLogEvent(Layout, loggingEvent)));
+                if (UseNLogCustomJsonLayout)
```
snakefoot, line 494: Instead of "just" replacing the "message", then I think it should replace the entire JsonPayload.

So the output from the JsonLayout becomes the JsonPayload. This gives total freedom, which you are chasing.

jskeet:
I'll look at this on Monday - I'm on vacation until then.

orchowski (commented):
> I'll look at this on Monday - I'm on vacation until then.

Okay! Have a nice time then :)

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
snakefoot: Nah I prefer that it is all or nothing when deciding to use JsonLayout for customizing JsonPayload. In one month another person will show up and say that "properties"-field should placed in a completely different way. Now this person will look at your code, and recognize that even more strange options has to implemented to control with it should be `default` - `special mix`- `new special mix` - etc.

Just because some random Java-library have decide to lock itself into its own corner, then it doesn't mean all others have to do the same. But no-matter it is not my decision to make, since I'm not the owner of this library.

orchowski: Not a random library, but google one. I actually also don't have any idea what is the approach we should follow. For me, there is no problem to add service context properties by myself, but should we ignore context properties totally? 
But I get your point of view, it's reasonable.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/nlog-jsonTemplate.xml
```diff
@@ -0,0 +1,72 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
```
jskeet, line 1: This file looks excessively large to me, for a snippet. The intention would be for it to appear in a page like this: https://googleapis.dev/dotnet/Google.Cloud.Logging.NLog/latest/index.html

I very much doubt that it's worth having a 9-level nested layout just to show how to use this.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/nlog-jsonTemplate.xml
```diff
@@ -0,0 +1,72 @@
...
+  <targets async="true">
+    <target type="GoogleStackdriver" projectId="PROJECT_ID" logId="LOG_ID" name="stackdriver" resourceType="container" sendJsonPayload="true" enableJsonLayout="true" includeCallSite ="true" includeCallSiteStackTrace="true" IncludeEventProperties="false" IncludeMdlc="true">
+      <contextproperty name="contextProp" layout="propContextLayout" />
+      <layout xsi:type="JsonLayout" includeAllProperties="true">
+        <attribute name="someProp" layout="prop" />
+        <attribute name="nestedProp" encode="false" >
+          <layout type='JsonLayout'>
+            <attribute name='insideNested1' layout='hello nested1!' />
+            <attribute name='insideNested2' layout='hello nested2!' />
+            <attribute name="nestedNestedProp" encode="false" >
```
jskeet, line 18: Please remove the space before the `>` in all of these elements. (It's fine to have one before the `/>` though.)

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget_Configuration.cs
```diff
@@ -117,11 +117,18 @@ public partial class GoogleStackdriverTarget
...
         /// <summary>
         /// Fills <see cref="LogEntry.JsonPayload"/> instead of <see cref="LogEntry.TextPayload"/>.
-        /// JSON serialization can be customed using <see cref="JsonConverterTypeName"/> and <see cref="JsonConverterMethodName"/>,
+        /// JSON values serialization can be customized using <see cref="JsonConverterTypeName"/> and <see cref="JsonConverterMethodName"/>,
         /// or <see cref="JsonConverter"/>.
         /// </summary>
         public bool SendJsonPayload { get; set; }
 
+        /// <summary>
+        /// Fills <see cref="LogEntry.JsonPayload"/> in specified by user way with <see cref="T:NLog.Layouts.JsonLayout"/> layout type.
```
jskeet, line 126:
```suggestion
        /// Fills <see cref="LogEntry.JsonPayload"/> in a manner specified by the <see cref="NLog.Layouts.JsonLayout"/> layout type.
```

(The "T:" can confuse VS and is recommended against by FxCop. The other change is just to make the English make more sense - please let me know if I've misinterpreted the intention.)

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
```diff
@@ -491,7 +491,18 @@ private LogEntry BuildLogEntry(LogEventInfo loggingEvent)
             if (SendJsonPayload)
             {
                 var jsonStruct = new Struct();
-                jsonStruct.Fields.Add("message", Value.ForString(RenderLogEvent(Layout, loggingEvent)));
+                if (EnableJsonLayout)
```
jskeet, line 494: Just to check my understanding: the idea is that most of the struct (properties, service etc) will be populated as normal this allows additional fields to be set in the JSON struct, and the message field is not automatically emitted.

I think I've understood correctly but I wanted to check - partly as I expect to write up documentation for this at some point.

jskeet (commented):
Would appreciate @amanda-tarafa looking at this too, but I think I'm happy enough with the overall approach. Just a few suggestions/requests.

amanda-tarafa:
Looking now...

amanda-tarafa (commented):
I have to agree with @snakefoot. I think the whole `JsonPayload` should be replaced by what's configured in the custom layout. If you need it the same as what the Java library is writing then you can recreate that when you define your custom layout using placeholders.

Imagine that in two weeks, we get someone not wanting `ServiceContext` to be included on their `JsonPayload`. With custom layouts, they should just be able to not include it in their custom layout and that's it, yet, the code will include it nonetheless. Now we have to implement a new feature that allows for a new config value `IncludeServiceContext`, and then two weeks after, someone comes and says that they don't want the property collection, or just some properties in the property collection and some at the upper level, ....

I get that it will be somewhat harder for you config wise to mimic what the Java library is doing, but it's still "easily" doable. And we can provide examples on how to include what is included when using our current "layout". (Unless I'm missing something here, I'm also not that familiar with NLog).

And it still won't be a breaking change, because for using the custom layout people will still have to set `EnableJsonLayout` to true.

It's also easier to document and understand someting like: 

  * _if you set `EnableJsonLayout` to `true`, then the `JsonPayload` will be fully defined by the custom json layout_ 

than something like 

  * _if you set `EnableJsonLayout` to true, then this part of the `JsonPayload` will be defined by the custom json layout but these other parts won't_ .

jskeet:
What @amanda-tarafa works for me as well, btw.

orchowski:
@amanda-tarafa @jskeet I get it. I'll commit changes tomorrow.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget_Configuration.cs
orchowski: I'm using IDEA Rider and it 'sees' reference after adding T:.

jskeet: Well, VS doesn't - so please change it. (I happened to make the same change in NodaTime this weekend.)

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
orchowski: Yes, but like @snakefoot and @amanda-tarafa said it'd be better to change that to overriding the whole structure

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget_Configuration.cs
orchowski: Hmm. It didn't work.
https://github.com/googleapis/google-cloud-dotnet/pull/5635/checks?check_run_id=1479470031#step:7:14
It breaks build, so I'm going to change that back

orchowski:
What 'detect differences' step actually do, and what's the problem. I can't get it

jskeet:
The problem is that you haven't rebased back to the `master` branch, which has a release PR for Vision merged. If you run the following commands, it should be fine (assuming you've got a remote of "upstream"):

```sh
git checkout master
git pull upstream master
git checkout nlog-logging-structure-feature
git rebase master
git push -f
```

jskeet:
(I haven't reviewed the changes yet though.)

jskeet:
Ah - I've just worked out, due to an extra complication, this probably *won't* work. (The Vision release PR has been created, but it hasn't actually been released. Don't worry about that for now.)

orchowski:
> The problem is that you haven't rebased back to the `master` branch, which has a release PR for Vision merged. If you run the following commands, it should be fine (assuming you've got a remote of "upstream"):


I'm pretty sure that I did it. Hmm, I think my fork is not synchronized

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
```diff
@@ -491,7 +491,18 @@ private LogEntry BuildLogEntry(LogEventInfo loggingEvent)
...
             {
                 var jsonStruct = new Struct();
+                if (EnableJsonLayout)
+                {
+                    foreach (var jToken in JObject.Parse(RenderLogEvent(Layout, loggingEvent)))
+                    {
+                        jsonStruct.Fields.Add(jToken.Key, ProtoConverter.Convert(jToken.Value));
+                    }
+                    logEntry.JsonPayload = jsonStruct;
+                    return FillLogEntryCommonData(loggingEvent, logEntry);
```
jskeet, line 501: I think it would be clearer just to add an "else" for the normal case here, and let `FillLogEntryCommonData` occur at the end of the method. The current structure confused me for a bit.

jskeet (commented):
This now looks fine to me, other than one requested change about if/else structure. Once that's done, I'll wait until the Vision release has happened (later today, I hope) and rerun the diff check, which should go green. I can then merge.

Thanks for this!

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget.cs
orchowski: Hmm... If someone wants to trace the `EnableJsonLayout` case then he can finish reading the method at this place instead of looking further. In this case, the 'else' statement is redundant. So... maybe it's a little bit better then?

If I understood you correctly you also prefer to check this condition near the properties and service context structure filling?

My goal was to avoid double-checking (or even triple) the same scenario. But maybe I got you wrong and you want to nest everything in else statement?

jskeet: >  In this case, the 'else' statement is redundant. 

Yes, but it involves the duplication of the `FillLogEntryCommonData` return statement. So in order to check the differences between EnableJsonLayout being set or not, you have to look quite hard IMO. What I'm suggesting is removing this `return` statement, and putting everything that happens when EnableJsonLayout is *disabled* in an `else` block. That keeps the two sets of either/or statements at the same level of nesting, making it clearer. It's easy to miss the return statement when skimming, and assume that the EnableJsonLayout code is *in addition* rather than *instead of* the rest.

orchowski:
> This now looks fine to me, other than one requested change about if/else structure. Once that's done, I'll wait until the Vision release has happened (later today, I hope) and rerun the diff check, which should go green. I can then merge.
> 
> Thanks for this!

Okay, great to hear! It was a pleasure for me too. I synced this branch with master and everything should go well.
Thanks for your fast responses and advice!

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget_Configuration.cs
```diff
@@ -117,11 +117,18 @@ public partial class GoogleStackdriverTarget
...
+        /// JSON values serialization can be customized using <see cref="JsonConverterTypeName"/> and <see cref="JsonConverterMethodName"/>,
```
amanda-tarafa, line 127: Could I trouble you to add a test that uses the JsonLayout and at least one of the converters? Just to see that it all comes together nicely. Thanks.

amanda-tarafa (commented):
Yes, looking good! I mostly just have a request for a new test.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget_Configuration.cs
orchowski: @amanda-tarafa I was sure that my link will refer to history, but I can't find it now. Build failed because of that. I can turn it back but it'll fail again.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog/GoogleStackdriverTarget_Configuration.cs
orchowski: okay

orchowski: Thanks for this request! It'd be a mistake because I used ProtoConverter directly. Now I fixed that.

amanda-tarafa (approved):
LGTM

amanda-tarafa:
Thanks a lot for this.

orchowski:
> Thanks a lot for this.

Yes. I thank you to for being vigilant. Cheers

jskeet:
FYI: I was wrong about release timetables, so it'll be tomorrow before Vision is released, at which point I'll rerun the diff check, which should go green.

jskeet:
Merged at least - thanks for your patience. I want to update the docs a bit before we release it though. I *hope* we'll be able to do that next week.

orchowski:
Great, thanks

snakefoot:
@orchowski Maybe to the initial post for this pull-request with an example how to setup this feature (Using `EnalbleJsonLayout`) along with actual output being generated.

PR #5650, opened 2020-12-02
Update Log4Net dependency to fix issue on Linux

jskeet:
Log4net 2.0.11 (which
Microsoft.Extensions.Logging.Log4Net.AspNetCore version 3.1.5
depends on) fails on Linux due to depending on something not
implemented in .NET Core. 2.0.12 fixes that.

I've had 2.0.12 working successfully in production.


amanda-tarafa (approved):
LGTM

PR #5651, opened 2020-12-02
docs: Recommend using CreateAsync also for creating generated clients.

amanda-tarafa:
Should we propose this better?


jskeet (commented):
Not sure about this. It depends on the context. For example, if you're creating a singleton client for dependency injection in ASP.NET Core, async is tricky but sync would be fine because you're only doing it once.

Maybe specify both methods here?

amanda-tarafa:
Yes, will do both then.

jskeet (approved):

PR #5656, opened 2020-12-03
docs: Fix a link in the Firestore docs

jskeet:
It's tricky to link to the GAX enum within Markdown (and it resolves
differently based on the docs final destination) so I figured I'd
just link to the property instead - users can click on the property
type from there if they want to, and it should be handled correctly
everywhere.


amanda-tarafa (approved):
LGTM

PR #5657, opened 2020-12-03
feat(spanner): add support for RPC priority

olavloite:
Adds support for RPC priority to Cloud Spanner.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
@@ -14,10 +14,11 @@
 
 using Google.Api.Gax;
 using Google.Api.Gax.Grpc;
+using Google.Api.Gax.Testing;
 using Google.Cloud.Spanner.Common.V1;
 using Google.Cloud.Spanner.V1;
-using Google.Cloud.Spanner.V1.Tests;
 using Google.Cloud.Spanner.V1.Internal.Logging;
+using Google.Cloud.Spanner.V1.Tests;
 using Grpc.Core;
 using Moq;
 using System;
@@ -25,7 +26,6 @@ using System.Data;
 using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
-using Google.Api.Gax.Testing;
 
 namespace Google.Cloud.Spanner.Data.Tests
 {
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
@@ -399,7 +399,7 @@ namespace Google.Cloud.Spanner.Data.Tests
             spannerClientMock
                 .SetupBatchCreateSessionsAsync()
                 .SetupBeginTransactionAsync()
-                .SetupExecuteStreamingSqlForPdml();
+                .SetupExecuteStreamingSqlForDml(ResultSetStats.RowCountOneofCase.RowCountLowerBound);
             SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
 
             var command = connection.CreateDmlCommand("DELETE FROM Users WHERE Active=False");
@@ -408,6 +408,31 @@ namespace Google.Cloud.Spanner.Data.Tests
 
             spannerClientMock.Verify(client => client.ExecuteStreamingSql(
                 It.Is<ExecuteSqlRequest>(request => request.RequestOptions.Priority == RequestOptions.Types.Priority.Low),
+                It.IsAny<CallSettings>()), Times.Once());
+        }
+
+        [Fact]
+        public void EphemeralTransactionIncludesPriorityOnDmlCommandAndCommit()
+        {
+            var priority = Priority.Medium;
+            Mock<SpannerClient> spannerClientMock = SpannerClientHelpers
+                .CreateMockClient(Logger.DefaultLogger, MockBehavior.Strict);
+            spannerClientMock
+                .SetupBatchCreateSessionsAsync()
+                .SetupBeginTransactionAsync()
+                .SetupExecuteStreamingSqlForDml(ResultSetStats.RowCountOneofCase.RowCountExact)
+                .SetupCommitAsync();
+            SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+
+            var command = connection.CreateDmlCommand("UPDATE FOO SET BAR=1 WHERE ID=1");
+            command.Priority = priority;
+            command.ExecuteNonQuery();
+
+            spannerClientMock.Verify(client => client.ExecuteStreamingSql(
+                It.Is<ExecuteSqlRequest>(request => request.RequestOptions.Priority == PriorityConverter.ToProto(priority)),
+                It.IsAny<CallSettings>()), Times.Once());
+            spannerClientMock.Verify(client => client.CommitAsync(
+                It.Is<CommitRequest>(request => request.RequestOptions.Priority == PriorityConverter.ToProto(priority)),
                 It.IsAny<CallSettings>()), Times.Once());
         }
 
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
@@ -210,6 +210,27 @@ namespace Google.Cloud.Spanner.Data.Tests
         }
 
         [Fact]
+        public void CommandPriorityDefaultsToUnspecified()
+        {
+            var connection = new SpannerConnection("Data Source=projects/p/instances/i/databases/d");
+            var command = connection.CreateSelectCommand("SELECT * FROM FOO");
+            Assert.Equal(Priority.Unspecified, command.Priority);
+        }
+
+        [Fact]
+        public void CommitPriorityDefaultsToUnspecified()
+        {
+            Mock<SpannerClient> spannerClientMock = SpannerClientHelpers
+                .CreateMockClient(Logger.DefaultLogger, MockBehavior.Strict);
+            spannerClientMock
+                .SetupBatchCreateSessionsAsync()
+                .SetupBeginTransactionAsync();
+            SpannerConnection connection = BuildSpannerConnection(spannerClientMock);
+            SpannerTransaction transaction = connection.BeginTransaction();
+            Assert.Equal(Priority.Unspecified, transaction.CommitPriority);
+        }
+
+        [Fact]
         public void CommandIncludesPriority()
         {
             var priority = Priority.High;
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
@@ -25,6 +25,7 @@ using System.Data;
 using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
+using Google.Api.Gax.Testing;
 
 namespace Google.Cloud.Spanner.Data.Tests
 {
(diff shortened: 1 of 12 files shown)


jskeet:
I can look at this on Monday, but it won't be before that

olavloite:
cc @skuruppu

olavloite:
> I can look at this on Monday, but it won't be before that

No rush, the proto changes haven't been merged yet.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
```diff
@@ -108,8 +108,9 @@ async Task<IEnumerable<long>> Impl()
         /// <param name="mutations">The list of changes to apply.</param>
         /// <param name="cancellationToken">A cancellation token used for this task.</param>
         /// <param name="timeoutSeconds">The timeout which will apply to the commit part of this method.</param>
+        /// <param name="priority">The priority that will be used for the commit request.</param>
         /// <returns>The number of rows modified.</returns>
-        public Task<int> ExecuteMutationsAsync(List<Mutation> mutations, CancellationToken cancellationToken, int timeoutSeconds)
+        public Task<int> ExecuteMutationsAsync(List<Mutation> mutations, CancellationToken cancellationToken, int timeoutSeconds, Priority priority)
```
jskeet, line 113: Note to self: this is an internal class, so it's not a breaking change despite looking like one. Ditto the ISpannerTransaction change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Priority.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// Executes the request with medium priority.
+        /// </summary>
+        Medium,
+        /// <summary>
+        /// Executes the request with high priority.
+        /// </summary>
+        High
+    }
+
+    internal class PriorityConverter
```
jskeet, line 40: This makes me nervous. We need to consider what will happen if a new value is added in the proto vs our enum here.
I would be tempted to keep things in terms of the Spanner.Data enum everywhere we can, so we only ever need to convert *to* the proto form. That would mean keeping it separately from the RequestOptions, admittedly.

Alternatively, we could potentially just convert numerically:

```csharp
internal static Priority FromProto(V1.RequestOptions.Types.Priority proto) => (Priority) (int) proto;
```

We'd need to keep the numeric values in sync, but I think that's fine. At that point, we really get transparency.

jskeet (commented):
I *think* I'm generally fine with this. Admittedly with 3 PRs queued up, it's hard to work out exactly how they'll all interact with each other, which makes me nervous.

amanda-tarafa:
I've added the `do not merge` label and will be adding it to the other PRs. There's now a check that won't allow merging if the label is set.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Priority.cs
amanda-tarafa: +1 to keeping things separately from the RequestOptions. For this PR and the Tags one. And only construct the RequestOptions when they are going to be used.

amanda-tarafa (commented):
I've left a suggestion for an alternative implementation for Mutations, which I think will be cleaner.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -126,6 +126,24 @@ internal bool HasMutations
             }
         }
 
+        // Request options for the commit request of this transaction. Only the relevant properties are exposed publicly.
+        private V1.RequestOptions CommitRequestOptions { get; } = new V1.RequestOptions();
```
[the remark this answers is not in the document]
olavloite: Correct, and it's not yet clear exactly which of the two will be ready for merge first.

olavloite:
> I've added the `do not merge` label and will be adding it to the other PRs. There's now a check that won't allow merging if the label is set.

Thanks. I don't have permission to add labels to PRs in this repo.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Priority.cs
olavloite: I've removed `RequestOptions` from the commands and transactions and introduced `Priority` as a separate property. Once we have agreement on this in this PR, I'll look into changing #5575 so they are in sync with each other.

I also changed the conversion of `Priority` to a pure numeric conversion.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -289,7 +291,7 @@ private async Task<int> ExecuteMutationsAsync(CancellationToken cancellationToke
                 var mutations = GetMutations();
                 var transaction = Transaction ?? Connection.AmbientTransaction ?? new EphemeralTransaction(Connection, s_readWriteOptions);
```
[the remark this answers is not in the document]
olavloite: I've applied this in the latest version. This also has one additional benefit: It makes it very natural to include the `Priority` on the commit operation for other update statements in ephemeral transactions as well. The specs don't mention whether this should be done or not, but that is (probably) because ephemeral transactions are only present in .NET and JDBC.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -289,7 +291,7 @@ private async Task<int> ExecuteMutationsAsync(CancellationToken cancellationToke
...
                 // Make the request. This will commit immediately or not depending on whether a transaction was explicitly created.
-                await transaction.ExecuteMutationsAsync(mutations, cancellationToken, CommandTimeout).ConfigureAwait(false);
+                await transaction.ExecuteMutationsAsync(mutations, cancellationToken, CommandTimeout, PriorityConverter.FromProto(RequestOptions.Priority)).ConfigureAwait(false);
```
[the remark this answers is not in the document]
olavloite: Done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
olavloite: (Not relevant anymore, as with the latest changes there are no changes to the public methods of `EphemeralTransaction` or `ISpannerTransaction`)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -335,6 +353,11 @@ public Task<DateTime> CommitAsync(CancellationToken cancellationToken = default)
             "SpannerTransaction.Commit", SpannerConnection.Logger);
         }
 
+        private RequestOptions BuildCommitRequestOptions()
+        {
+            return new RequestOptions { Priority = PriorityConverter.ToProto(CommitPriority) };
+        }
+
```
amanda-tarafa, line 360:
```suggestion
        private RequestOptions BuildCommitRequestOptions() =>
            new RequestOptions { Priority = PriorityConverter.ToProto(CommitPriority) };
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -371,6 +373,11 @@ private V1.ExecuteSqlRequest.Types.QueryOptions GetEffectiveQueryOptions()
                 return queryOptionsProto;
             }
 
+            private RequestOptions BuildRequestOptions()
+            {
+                return new RequestOptions { Priority = PriorityConverter.ToProto(Priority) };
+            }
```
amanda-tarafa, line 379:
```suggestion
            private RequestOptions BuildRequestOptions() =>
                new RequestOptions { Priority = PriorityConverter.ToProto(Priority) };
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -126,6 +126,23 @@ internal bool HasMutations
             }
         }
 
+        private Priority _commitPriority;
+
+        /// <summary>
+        /// The RPC priority to use for the commit RPC of this transaction. This can only be set for read/write transactions.
```
amanda-tarafa, line 132: Consider adding here what it deafaults to? Or making it nullable, I think this is not a required field, right? Make sure to add a test to check for its initial value (whtether a default or `null`) so that we don't inadvertenly change it over time. Here and in commands as well.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
```diff
@@ -98,6 +103,11 @@ private ExecuteBatchDmlRequest GetExecuteBatchDmlRequest()
                 return request;
             }
 
+            private RequestOptions BuildRequestOptions()
+            {
+                return new RequestOptions { Priority = PriorityConverter.ToProto(Priority) };
+            }
```
amanda-tarafa, line 109:
```suggestion
            private RequestOptions BuildRequestOptions() =>
                RequestOptions { Priority = PriorityConverter.ToProto(Priority) };
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Priority.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    /// <summary>
+    /// The relative priority for requests.
+    /// </summary>
+    public enum Priority
+    {
+        /// <summary>
+        /// Executes the request with unspecified priority.
+        /// </summary>
+        Unspecified,
```
amanda-tarafa, line 25: Should this value be surfaced here? I think I would prefer making the properties on transactions and commands nullable, and convert the null to the `UNSPECIFIED` in the proto. If you look at the proto generated code, `UNSPECIFIED` is being ignored (i.e. treated as a `null`) for all `RequestOptions` operations (`WriteTo`s, `Merge`, `CalculateSize`, even `GetHashCode` ...).
An alternative to making the properties null, and given that it says in the proto docs that `UNSPECIFIED` means `HIGH`, is to default `Priority` to `High` always. I still prefer making the priority properties nullable.
@jskeet What do you think?

amanda-tarafa (commented):
Yep, I like this way better, it's overall cleaner I believe.
Mostly nits, execpt for Priority related comments.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Priority.cs
jskeet: I can see pros and cons to this. There's something appealing about keeping it mirroring the proto enum, but equally Amanda's point is fairly convincing.

I don't much mind either way - I'd just point out that if this is removed, `Low` needs to be given the explicit value of 1 to keep parity with the proto enum.

jskeet (commented):
I have nothing particularly to add beyond Amanda's existing suggestions.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
olavloite: I've added comments and tests for the default. The value will default to `Unspecified`, which is also one of the possible values in the proto enum. I don't think we should make this nullable, as it would mean that we would either have two ways of setting unspecified, or that we would not have a one-to-one mapping between the Spanner.Data.Priority enum and the proto enum. The property is not nullable in the `RequestOptions` proto.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Priority.cs
olavloite: My preference would be to keep this as closely aligned with the proto definition as possible, and keep the `Unspecified` value accessible for users. That way, anyone reading the Spanner gRPC documentation will see the same values here as in that part of the documentation.
Making it nullable would also be an option, although it would not be my first choice for the reasons above.

I don't think we should use `High` as the default. The current documentation states that `Unspecified` will default to `High`. However, if the backend team decides to change that in the future, it would require a manual intervention in the client as well.

jskeet:
https://github.com/googleapis/google-cloud-dotnet/pull/6265 has now been merged, so I suspect this PR can be rebased and potentially merged.

olavloite:
> #6265 has now been merged, so I suspect this PR can be rebased and potentially merged.

It's been rebased and the tests are succeeding. It's still awaiting a final approval from the backend team that it can be merged.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -218,7 +235,8 @@ public DisposeBehavior DisposeBehavior
         Task<int> ISpannerTransaction.ExecuteMutationsAsync(
             List<Mutation> mutations,
             CancellationToken cancellationToken,
-            int timeoutSeconds)
+            int timeoutSeconds // ignored
+            )
```
amanda-tarafa, line 239: tiny nit: 
```suggestion
            int timeoutSeconds // ignored)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
amanda-tarafa: OK

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -25,6 +25,7 @@
 using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
+using Google.Api.Gax.Testing;
```
amanda-tarafa, line 28: nit: alphabetically

amanda-tarafa (approved):
LGTM, which was missing from our side as well.

skuruppu:
@larkee please confirm whether this PR can be merged now.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
olavloite: done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
olavloite: done

larkee:
The PR should be okay to merge two days from now. I'll let you know if that changes :+1:

PR #5661, opened 2020-12-07
ci: Pass "APIs to check" into the NuGet compatibility check

jskeet:
This will prevent PRs from being held up by other releases that haven't completed yet.


amanda-tarafa (approved):
LGTM

PR #5662, opened 2020-12-07
chore: Remove generation tweaking scripts for Google.Cloud.Container.V1

jskeet:
I've regenerated locally without these scripts present, and there are no differences.


amanda-tarafa (approved):
LGTM

PR #5663, opened 2020-12-07
docs: Hack to make sure we generate ASP.NET Core 3 diagnostics snippets

jskeet:
This is ugly, but it's not worth trying to generalize.

This commit also removes support for ".Experimental" packages - we decided that was a bad idea a while ago.


amanda-tarafa (approved):
LGTM

PR #5673, opened 2020-12-09
Support the Pub/Sub emulator in a more standard way

### LINKED ISSUE #5261: Support the PubSub emulator in the same way as other APIs
Currently we have a few APIs (Datastore, Firestore and Spanner) that use EmulatorDetection to make it easy to connect to the emulator.

We should apply the same pattern to PubSub (which is *slightly* complex due to having a wrapper layer) and then update the documentation around this.

jskeet:
I haven't yet run the integration tests with this turned on - I will do so shortly. (There's no urgency on this review.)

It's a little convoluted due to the way PublisherClient and SubscriberClient work, but I've managed to keep the emulator awareness within the builders.

Fixes #5261.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
@@ -38,6 +38,10 @@ namespace Google.Cloud.PubSub.V1
     /// </summary>
     public abstract class PublisherClient
     {
+        private static readonly GrpcChannelOptions s_unlimitedSendReceiveChannelOptions = GrpcChannelOptions.Empty
+            .WithMaxReceiveMessageSize(-1)
+            .WithMaxSendMessageSize(-1);
+
         // TODO: Logging
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


jskeet:
I've just run the integration tests. Out of 19 tests, 15 passed, 3 failed and 1 was skipped (as normal). The three failing tests are:

- StopStartSubscriber(10000, 1000, 2) - "No message has been received for too long; failing test"
- DeadLetterQueueAndDeliveryAttempt (gRPC Unimplemented error)
- OversizedMessage (expected an exception; it wasn't thrown)

Of these, I suspect that DLQ is just a feature not implemented in the emulator, and the oversized message is a *limit* not implemented in the emulator. I'll look at StopStartSubscriber again when I get a chance.

chrisdunelm (approved):
Slightly convoluted as you say.
But LGTM

jskeet:
Thanks Chris - Amanda, now that we know Chris is happy, I'd love a third pair of eyes on it, just because it's fiddly. I still need to look at the integration tests that were failing, too... in particular the first of those failures.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
```diff
@@ -204,15 +239,23 @@ public static async Task<PublisherClient> CreateAsync(TopicName topicName, Clien
...
             for (int i = 0; i < clientCount; i++)
             {
-                var channelOptions = new[]
+                // Use a random arg to prevent sub-channel re-use in gRPC, so each channel uses its own connection.
+                var grpcChannelOptions = s_unlimitedSendReceiveChannelOptions
+                    .WithCustomOption("sub-channel-separator", Guid.NewGuid().ToString());
+
+                // First builder to handle any endpoint detection etc. We build a gRPC channel
+                // with this.
+                var builder = new PublisherServiceApiClientBuilder
```
amanda-tarafa, line 248: Just wondering if writing an `internal class ChannelBuilder : ClientBuilderBase<ChannelBase>` that `PublisherClient`, `PublisherServiceApiClientBuilder`, `SubscriberClient` and `SubscriberServiceApiClientBuilder` could rely on would simplify things.
The CallInvoker bits on this channel builder could be overwritten to throw, and this would be an internal class after all.
Not sure though.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -158,11 +162,42 @@ public sealed class ClientCreationSettings
...
             {
                 ClientCount = clientCount;
                 SubscriberServiceApiSettings = subscriberServiceApiSettings;
                 Credentials = credentials;
                 ServiceEndpoint = serviceEndpoint;
+                EmulatorDetection = emulatorDetection;
+            }
+
+            /// <summary>
+            /// Specifies how the builder responds to the presence of emulator environment variables.
```
amanda-tarafa, line 184:
```suggestion
            /// Specifies how to respond to the presence of emulator environment variables.
```

amanda-tarafa (approved):
LGTM.

jskeet:
Hmm... StopStartSubscriber is still failing - it only receives 100 messages out of 10000.

amanda-tarafa (approved):
LGTM

PR #5676, opened 2020-12-09
Require a client to be specified in all smoke tests

jskeet:
While not needing one is convenient when there's only a single
client, it makes the act of adding a new client a breaking change.


amanda-tarafa (approved):
LGTM

PR #5681, opened 2020-12-09
Updated NLog config snippets to be simpler and cater for sendJsonPayload

snakefoot:
When using `enableJsonLayout="true"` then it doesn't make sense to specify `<contextproperty>`.

NLog callsite capture has a performance-hit, so think it is a bad idea to activate them in example snippets (`includeCallSite ="true" includeCallSiteStackTrace="true"`)

### DIFF
--- a/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/GoogleStackdriverTargetSnippets.cs
+++ b/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/GoogleStackdriverTargetSnippets.cs
@@ -34,6 +34,7 @@ namespace Google.Cloud.Logging.NLog.Snippets
 
         public GoogleStackdriverTargetSnippets(NLogSnippetFixture fixture) => _fixture = fixture;
 
+        // Resource: nlog-jsonPayload.xml nlog_jsonPayload
         // Resource: nlog-jsonTemplate.xml nlog_jsonTemplate
 
         [Fact]
--- a/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/nlog-jsonPayload.xml
+++ b/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/nlog-jsonPayload.xml
@@ -1,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<nlog xmlns="http://www.nlog-project.org/schemas/NLog.xsd"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+
+  <extensions>
+    <add assembly="Google.Cloud.Logging.NLog" />
+  </extensions>
+  
+  <targets async="true">
+    <target name="stackdriver" xsi:type="GoogleStackdriver" projectId="PROJECT_ID" logId="LOG_ID" includeEventProperties="true" sendJsonPayload="true">
+      <contextproperty name="exception" layout="${exception:format=tostring}" />  <!-- Optional and repeatable -->
+    </target>
+    <target name="console" xsi:type="Console" />
+  </targets>
+
+  <rules>
+    <logger name="*" minlevel="Info" writeTo="stackdriver" />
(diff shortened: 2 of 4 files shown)


apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/nlog-template.xml
```diff
@@ -7,8 +7,8 @@
   </extensions>
   
   <targets async="true">
-    <target name="stackdriver" xsi:type="GoogleStackdriver" projectId="PROJECT_ID" logId="LOG_ID">
-      <!-- Potentially additional configuration -->
+    <target name="stackdriver" xsi:type="GoogleStackdriver" projectId="PROJECT_ID" logId="LOG_ID" includeEventProperties="true" sendJsonPayload="true">
```
jskeet, line 10: The intention is for this to be a pretty minimal example - I'd rather not include those optional configuration values in there. (While the log ID and project ID are optional in some cases, they're often the difference between "working" and "not working at all" whereas the format of the log entry is a bit different.)

I wouldn't mind having an extra section in here talking about the pros and cons of using a JSON payload, but I don't want it right at the start of the docs.

snakefoot: I have now reverted the change of `nlog-template.xml` and instead introduced `nlog-jsonPayload.xml`.

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/nlog-template.xml
```diff
@@ -5,7 +5,7 @@
   <extensions>
     <add assembly="Google.Cloud.Logging.NLog" />
   </extensions>
-  
+
```
jskeet, line 8: Given that the only changes to this file are now whitespace, please revert it.

jskeet (commented):
Please add a resource reference here as well: https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/GoogleStackdriverTargetSnippets.cs#L37

apis/Google.Cloud.Logging.NLog/Google.Cloud.Logging.NLog.Snippets/nlog-template.xml
snakefoot: Tried but Visual Studio is very persistent about making this change. Will try again.

snakefoot: Success

snakefoot:
> @jskeet  Please add a resource reference here as well

I'm guessing the test verifies if the nlog-xml-examples are working, but I can only see it loads `nlog-template.xml`:

``string` resourceName = $"{typeof(GoogleStackdriverTargetSnippets).Namespace}.nlog-template.xml";`

You want me to add 2 more tests with these files:

- nlog-jsonTemplate.xml
- nlog-jsonPayload.xml

Or do you just want to update the comment line?

- `// Resource: nlog-jsonTemplate.xml nlog_jsonTemplate`

I'm not sure I understand the format or reason of the above comment. Can you give me the verbatim line you want instead?

jskeet:
No, it's not about tests. It's just a comment line that tells the snippet extractor to include the resource. Just copy the line I referred to and make an equivalent one. Can't say more easily now - on a mobile with a cat in my lap...

snakefoot:
> It's just a comment line that tells the snippet extractor to include the resource. Just copy the line I referred to and make an equivalent one. 

Added line with `nlog_jsonPayload`

jskeet:
Thanks, will pull and build the docs to check them tomorrow, although I'm officially on vacation until Monday so I won't be spending huge amounts of time on it.

jskeet (approved):
I've now tested this locally and can confirm that the XML file shows up in the docs. Merging.

snakefoot:
Thank you for the review and merge.

PR #5688, opened 2020-12-11
Add Google.Cloud.Retail.V2

busunkim96:
I didn't notice until I had already gotten to the 'create a smoke test' stage that I was supposed to do Domains 🙈 . I thought I might as well open the PR for a review.


jskeet (approved):
Not a problem - let's stick with Retail then :)

jskeet:
I assume you'll send a second PR for the release?

PR #5692, opened 2020-12-14
Add smoke test for AccessApproval

### LINKED ISSUE #5683: Investigate Access Approval failures
For some reason, ListApprovalRequests is failing with an unknown error. It works using the REST .NET library, and it works using the gRPC API from Java, so we need to work out what's going on here. Although my suspicion is that it's to do with the renaming, I did attempt to effectively test it without the renaming, and it still failed. Need to look into it.

jskeet:
This appeared not to work before, for unknown reasons... seems okay now.

Closes #5683.


amanda-tarafa (approved):
LGTM

PR #5694, opened 2020-12-14
Update the default versions of Grpc.Core and Google.Protobuf

jskeet:
We don't need to manually update apis.json, as versions are
now automatically bumped for minor or major releases when the
package is released.


amanda-tarafa (approved):
LGTM

jskeet:
Whoops, I'll need to rerun generateprojects.sh in another commit.

jskeet:
Ouch... it looks like Grpc.Core.Api 2.34.0 may have a different signing key or something like that - I'm looking into it now.

jskeet:
Yup, this is tracked here: https://github.com/grpc/grpc/issues/24935

jskeet:
I'm going to close this PR as we definitely don't want to update to 2.34.0. When there's a new Grpc.Core release with this issue fixed, we can do an equivalent PR then.

PR #5697, opened 2020-12-15
Add Google.Cloud.Domains.V1Beta1

busunkim96:
Playbook exercise take 2. :) 

Release will be opened in a separate PR.


jskeet (approved):
Could you rebase to remove the key from the commit in which it was added, then force push? We don't want it in a commit anywhere in the history... I could do a squash and merge if necessary, but it's nice to have the history with three separate commits if you're happy to fix up the history.

busunkim96:
I believe the key has been properly removed from the history now, PTAL.

jskeet (approved):
Yay, LGTM - will merge

PR #5721, opened 2020-12-29
fix: local date could lead to wrong date in database

olavloite:
Using a local `DateTime` to write a value to a `DATE` column in Spanner could lead to an unexpected date to be written in the database if the local timezone has an offset, as the `DateTime` would first be serialized as a UTC timestamp, and then have the time information stripped away. This meant that a date like `2020-12-29T00:00:00+01:00` would be written as `2020-12-28` to the database. This also means that writing the value of `DateTime.Today` to the database could lead to an unexpected date being stored in the database.


jskeet:
We'll look at this on Tuesday - both myself and Amanda are on vacation until then.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DateTimestampReadTests.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public DateTimestampReadTests(DateTimestampTableFixture fixture) =>
+            _fixture = fixture;
+
+        [Fact]
+        public async void WriteDateThenRead_ShouldBeEqual()
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                foreach (var date in new DateTime[] {
```
jskeet, line 34: I think it would be easier to read if you declared a variable with these values in rather than declaring the array inline within the foreach statement itself.

Alternatively, remove the `DateTime.Today` value and use a parameterized test, which makes it easier to see which ones work. (I don't tend to like using `Today` or `Now` in tests anyway. They can lead to flaky tests if something only fails if the second is 0 for example.)

jskeet: (Same comment for the other test.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
```diff
@@ -173,6 +173,13 @@ internal Value ToProtobufValue(object value, SpannerConversionOptions options)
                         StringValue = XmlConvert.ToString(Convert.ToDateTime(value, InvariantCulture), XmlDateTimeSerializationMode.Utc)
                     };
                 case TypeCode.Date:
+                    if (value is DateTime date && date.Kind == DateTimeKind.Local)
+                    {
+                        // If the DateTime is local the value should be changed to unspecified
+                        // before serializing it as UTC to prevent accidentally changing to a
+                        // different date.
+                        value = new DateTime(date.Year, date.Month, date.Day);
```
jskeet, line 181: It *might* be clearer to have:

```csharp
value = DateTime.SpecifyKind(date.Date, DateTimeKind.Unspecified);
```

See what you think.

jskeet (commented):
I'm a *little* nervous of this, as it could break code that already works around this... but any *sensible* workarounds would be converting it into an unspecified kind anyway, so wouldn't be broken, so I'm less worried.

Would like to wait for Amanda's review as well though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DateTimestampReadTests.cs
```diff
@@ -0,0 +1,99 @@
...
+                    // Write the date value and read it back.
+                    await connection.CreateInsertOrUpdateCommand(_fixture.TableName,
+                        new SpannerParameterCollection
+                        {
+                            new SpannerParameter("DateValue", SpannerDbType.Date, date),
+                        }
+                    ).ExecuteNonQueryAsync();
+                    using (var reader = await connection.CreateSelectCommand($"SELECT DateValue FROM {_fixture.TableName}").ExecuteReaderAsync())
+                    {
+                        Assert.True(await reader.ReadAsync());
```
amanda-tarafa, line 52: The reader and commands are extensively tested elsewhere, maybe remove the Asserts around the reader reading or not a row.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DateTimestampReadTests.cs
```diff
@@ -0,0 +1,99 @@
...
+                            new SpannerParameter("TimestampValue", SpannerDbType.Timestamp, timestamp),
+                        }
+                    ).ExecuteNonQueryAsync();
+                    using (var reader = await connection.CreateSelectCommand($"SELECT TimestampValue FROM {_fixture.TableName}").ExecuteReaderAsync())
+                    {
+                        Assert.True(await reader.ReadAsync());
+                        var dbTimestamp = reader.GetFieldValue<DateTime>("TimestampValue");
+                        // Cloud Spanner always stores the timestamp in UTC, so if the original input value was
+                        // a local time, we need to convert the value that we read in the database back to local time.
+                        if (timestamp.Kind == DateTimeKind.Local)
```
amanda-tarafa, line 88: Better if there are no iffs, just convert the timestamp to UTC always before the comparison.

amanda-tarafa (changes requested):
> This meant that a date like 2020-12-29T00:00:00+01:00 would be written as 2020-12-28

Honestly I wouldn't agree with this being a problem. The dates are read back in UTC always so it's fine that they are converted to UTC and then stripped of time. As the code is now, if you store DateTime.Now at the same time, in all timezones, the same value (date part od UTC Now) is stored. With this PR, doing that would result in different values being stored for different timezones. .NET doesn't have a type for date only, DateTime.Today is just a shortcut, so even if you are using DateTime.Today, it shouldn't be terribly surprising that you get your date converted to UTC first.
The fact that for testing this change you compare local dates with UTC dates, local being possibly different than UTC, and expect them to be equal, is also counterintuitive to Microsoft's own recommendations [here](https://docs.microsoft.com/en-us/dotnet/api/system.datetime.compare?view=netcore-3.1#remarks).

I'll defer back to @jskeet who definetely knows more about dates than I do. I've left some specific comments on the PR in case it is merged.

jskeet:
> With this PR, doing that would result in different values being stored for different timezones.

Yes - but I believe that's the correct thing to do. If someone has a `DateTime` with a value of 2021-01-05T00:00:00 and they store it as a date, I think it's reasonable to expect that when they retrieve it they still have 2021-01-5T00:00:00 (albeit not with a kind of Local any more).

One alternative is to just throw an exception, of course, but that's not terribly pleasant :)

At the heart of this is .NET's nasty date/time handling of course... but I do think this is the right way forward.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DateTimestampReadTests.cs
olavloite: I've changed the tests to a `Theory` and removed the `Now` and `Today` values. The important test values here are the different kinds, and not the actual date values.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
olavloite: Yes, that does seem more appropriate for what the intention is here. Changed.

olavloite:
> Honestly I wouldn't agree with this being a problem. The dates are read back in UTC always so it's fine that they are converted to UTC and then stripped of time. As the code is now, if you store DateTime.Now at the same time, in all timezones, the same value (date part od UTC Now) is stored. With this PR, doing that would result in different values being stored for different timezones. .NET doesn't have a type for date only, DateTime.Today is just a shortcut, so even if you are using DateTime.Today, it shouldn't be terribly surprising that you get your date converted to UTC first.

I think the biggest challenge here is that there does not seem to be a very good mapping available for a Cloud Spanner `DATE` to any (standard) data type in .NET. A Cloud Spanner `DATE` column is by definition [timezone independent](https://cloud.google.com/spanner/docs/data-types#date_type), and is not intended to specify a specific point in time (i.e. also not midnight UTC of the given date). When a Cloud Spanner `DATE` value is used in a context where a timezone is needed, the default timezone [that is used is not UTC but America/Los Angeles](https://cloud.google.com/spanner/docs/date_functions#date). So at least I was somewhat surprised when using `DateTime.Today` did not result in my local date being inserted in the database.

I would even argue that it might have been more appropriate to read dates back specifically as `Unspecified` instead of UTC for `DATE` columns, but I don't think that is a change we should make now as it would probably break things.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DateTimestampReadTests.cs
olavloite: I've changed test test cases to `Theory` and also changed the code reading the values to using `ExecuteScalarAsync` without any unnecessary asserts for stuff that is not really tested here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DateTimestampReadTests.cs
olavloite: Actually, that does not work (assuming that you mean calling `timestamp.ToUniversalTime()` on the expected timestamp). The reason is that:
1. If I write a `DateTime` instance with `DateTimeKind.Unspecified` to the database, that is interpreted by the client library as a `DateTime` that is already in UTC. So writing `new DateTime(2020, 9, 29, 18, 51, 10)` to the database inserts `20200929T18:51:10Z`.
2. Calling `DateTime.ToUniversalTime()` on a `DateTime` instance with `DateTimeKind.Unspecified` will make .NET treat the `DateTime` as if it was in the local timestamp.

So there is a discrepancy between how the Cloud Spanner client library interprets an unspecified `DateTime` and how .NET interprets such as a `DateTime`. I think the current behavior of the Spanner client library is right in interpreting an unspecified `DateTime` as already being in UTC, as Cloud Spanner makes it clear that `TIMESTAMP` values are stored in UTC.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DateTimestampReadTests.cs
```diff
@@ -0,0 +1,93 @@
...
+    [Collection(nameof(DateTimestampTableFixture))]
+    [CommonTestDiagnostics]
+    public class DateTimestampReadTests
+    {
+        private readonly DateTimestampTableFixture _fixture;
+
+        public DateTimestampReadTests(DateTimestampTableFixture fixture) =>
+            _fixture = fixture;
+
+        public static IEnumerable<object[]> TestDates =>
```
jskeet, line 30: I think this can be made simpler with `TheoryData<DateTime>` but I can do that in a follow-up PR. (This is much better than before already.)

jskeet (approved):

jskeet:
Looks like Amanda has to approve this before we can merge, as it was her review that requested changes.

amanda-tarafa (approved):
LGTM

PR #5817, opened 2021-01-01
Regenerate all APIs to update copyright year

jskeet:
The following command shows no changes:

```text
git diff --cached --unified=0 | \
  grep -v '"sha"' | \
  grep -v Copyright | \
  grep -v '\-\-\-' | \
  grep -v '+++' | \
  grep -v @@ | \
  grep -v "diff --git" | \
  grep -v -E '^index'
```

That basically accounts for:

- SHA differences in synth.metadata files
- Copyright differences
- All the additional context that git diff provides


jskeet:
(I'll self-merge this on green, then rerun autosynth to check that it's really caught up.)

jskeet:
(Ignoring the diff failure, which is due to a "missing" version number for AspNetCore3... although I'd expect the diff to only note APIs with changed version numbers anyway.)

PR #5836, opened 2021-01-13
fix: NUMERIC was missing from default mappings

olavloite:
Adds `NUMERIC` to the default mappings of `DbType => SpannerDbType`, `SpannerDbType => DbType` and `ClrType => SpannerDbType`.

Two considerations:
1. The Spanner `NUMERIC` data type is mapped to `DbType.VarNumeric` and not `DbType.Decimal`, as the latter has a lower range and precision than the Spanner `NUMERIC` type.
2. The .NET `decimal` ClrType [is currently mapped](https://github.com/olavloite/google-cloud-dotnet/blob/243c9fb4f32cd9e1cd48af07a48962c38f8e831e/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs#L281) to the Spanner `FLOAT64` data type. That is correct in the sense that both are floating point types. It does however not correspond with the default mapping for other database systems:
2.1. [SQL Server](https://docs.microsoft.com/en-us/dotnet/framework/data/adonet/sql/linq/sql-clr-type-mapping#numeric-mapping) maps it to `DECIMAL(29,4)`
2.2. [PostgreSQL](https://www.npgsql.org/doc/types/basic.html) maps it to `numeric` (which is a fixed point exact type)
2.3. [MySQL ](https://www.devart.com/dotconnect/mysql/docs/datatypemapping.html)maps it to `decimal` (which is a fixed point exact type).

Point 2 might cause some surprises for users moving from other DBMS'es to Spanner, and I would prefer to change that mapping to `NUMERIC` as well. WDYT?

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Cloud.Spanner.V1;
 using System;
 using System.Collections.Generic;
 using System.Data;
@@ -29,13 +30,14 @@ namespace Google.Cloud.Spanner.Data.Tests
             yield return new object[] { SpannerDbType.Timestamp, DbType.DateTime };
             yield return new object[] { SpannerDbType.Float64, DbType.Double };
             yield return new object[] { SpannerDbType.Int64, DbType.Int64 };
+            yield return new object[] { SpannerDbType.Numeric, DbType.VarNumeric };
             yield return new object[] { SpannerDbType.Unspecified, DbType.Object };
             yield return new object[] { SpannerDbType.String, DbType.String };
         }
 
         [Theory]
         [MemberData(nameof(GetDbTypeConversions))]
-        public void DbTypeMappgings(SpannerDbType spannerType, DbType adoType)
+        public void DbTypeMappings(SpannerDbType spannerType, DbType adoType)
         {
             var parameter = new SpannerParameter { SpannerDbType = spannerType };
             Assert.Equal(adoType, parameter.DbType);
@@ -44,6 +46,29 @@ namespace Google.Cloud.Spanner.Data.Tests
 
             parameter.DbType = adoType;
             Assert.Equal(spannerType, parameter.SpannerDbType);
+        }
+
+        // TODO: There is no value that will default to Spanner type DATE.
+        public static IEnumerable<object[]> GetValueConversions()
+        {
+            yield return new object[] { new byte[] { 1 }, SpannerDbType.Bytes, DbType.Binary };
+            yield return new object[] { true, SpannerDbType.Bool, DbType.Boolean };
+            yield return new object[] { new DateTime(2021, 1, 13, 12, 3, 10), SpannerDbType.Timestamp, DbType.DateTime };
+            yield return new object[] { 3.14D, SpannerDbType.Float64, DbType.Double };
+            yield return new object[] { 1L, SpannerDbType.Int64, DbType.Int64 };
+            yield return new object[] { (SpannerNumeric)3.14m, SpannerDbType.Numeric, DbType.VarNumeric };
+            yield return new object[] { "test", SpannerDbType.String, DbType.String };
+        }
+
+        [Theory]
+        [MemberData(nameof(GetValueConversions))]
+        public void ValueMappings(object value, SpannerDbType spannerType, DbType adoType)
+        {
+            var parameter = new SpannerParameter { Value = value };
+
+            Assert.Equal(value.GetType(), spannerType.DefaultClrType);
+            Assert.Equal(spannerType, parameter.SpannerDbType);
+            Assert.Equal(adoType, parameter.DbType);
         }
 
         public static IEnumerable<object[]> GetDbTypeSizes()
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
@@ -51,22 +51,33 @@ namespace Google.Cloud.Spanner.Data.Tests
         // TODO: There is no value that will default to Spanner type DATE.
         public static IEnumerable<object[]> GetValueConversions()
         {
-            yield return new object[] { new byte[] { 1 }, SpannerDbType.Bytes, DbType.Binary };
-            yield return new object[] { true, SpannerDbType.Bool, DbType.Boolean };
-            yield return new object[] { new DateTime(2021, 1, 13, 12, 3, 10), SpannerDbType.Timestamp, DbType.DateTime };
-            yield return new object[] { 3.14D, SpannerDbType.Float64, DbType.Double };
-            yield return new object[] { 1L, SpannerDbType.Int64, DbType.Int64 };
-            yield return new object[] { (SpannerNumeric)3.14m, SpannerDbType.Numeric, DbType.VarNumeric };
-            yield return new object[] { "test", SpannerDbType.String, DbType.String };
+            yield return new object[] { new byte[] { 1 }, SpannerDbType.Bytes, DbType.Binary, typeof(byte[]) };
+            yield return new object[] { new List<byte> { 1 }, SpannerDbType.Bytes, DbType.Binary, typeof(byte[]) };
+            yield return new object[] { true, SpannerDbType.Bool, DbType.Boolean, typeof(bool) };
+            yield return new object[] { new DateTime(2021, 1, 13, 12, 3, 10), SpannerDbType.Timestamp, DbType.DateTime, typeof(DateTime) };
+
+            yield return new object[] { 3.14f, SpannerDbType.Float64, DbType.Double, typeof(double) };
+            yield return new object[] { 3.14d, SpannerDbType.Float64, DbType.Double, typeof(double) };
+            yield return new object[] { 3.14m, SpannerDbType.Float64, DbType.Double, typeof(double) };
+
+            yield return new object[] { (short)1, SpannerDbType.Int64, DbType.Int64, typeof(long) };
+            yield return new object[] { (ushort)1, SpannerDbType.Int64, DbType.Int64, typeof(long) };
+            yield return new object[] { 1, SpannerDbType.Int64, DbType.Int64, typeof(long) };
+            yield return new object[] { 1u, SpannerDbType.Int64, DbType.Int64, typeof(long) };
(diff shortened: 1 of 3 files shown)


jskeet:
Will have a look at the PR later today - I think it would be best to file a new issue for point 2 for further discussion. (My initial guess is that we'd want to make that an option somewhere - just changing it immediately would be potentially breaking.)

I don't currently see any tests in the PR... it would be great to see a test that currently fails, that's then fixed by the production change.

olavloite:
> I don't currently see any tests in the PR... it would be great to see a test that currently fails, that's then fixed by the production change.

I added test cases for the change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
```diff
@@ -59,6 +59,9 @@ public override DbType DbType
                     case DbType.Int64:
                         SpannerDbType = SpannerDbType.Int64;
                         break;
+                    case DbType.VarNumeric:
```
amanda-tarafa, line 62: Would it be worth it to add a method in SpannerDbType.FromDbType and move all the code in this set there? Just for the sake of having type conversions in one place?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
```diff
@@ -46,6 +48,29 @@ public void DbTypeMappgings(SpannerDbType spannerType, DbType adoType)
...
 
+        // TODO: There is no value that will default to Spanner type DATE.
+        public static IEnumerable<object[]> GetValueConversions()
+        {
+            yield return new object[] { new byte[] { 1 }, SpannerDbType.Bytes, DbType.Binary };
+            yield return new object[] { true, SpannerDbType.Bool, DbType.Boolean };
+            yield return new object[] { new DateTime(2021, 1, 13, 12, 3, 10), SpannerDbType.Timestamp, DbType.DateTime };
+            yield return new object[] { 3.14D, SpannerDbType.Float64, DbType.Double };
+            yield return new object[] { 1L, SpannerDbType.Int64, DbType.Int64 };
+            yield return new object[] { (SpannerNumeric)3.14m, SpannerDbType.Numeric, DbType.VarNumeric };
```
amanda-tarafa, line 59: The literal suffixes are case insensitive, but you've used capital letters on the ones above, so, for consistency:
```suggestion
            yield return new object[] { (SpannerNumeric)3.14M, SpannerDbType.Numeric, DbType.VarNumeric };
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
```diff
@@ -46,6 +48,29 @@ public void DbTypeMappgings(SpannerDbType spannerType, DbType adoType)
             Assert.Equal(spannerType, parameter.SpannerDbType);
         }
 
+        // TODO: There is no value that will default to Spanner type DATE.
+        public static IEnumerable<object[]> GetValueConversions()
+        {
+            yield return new object[] { new byte[] { 1 }, SpannerDbType.Bytes, DbType.Binary };
+            yield return new object[] { true, SpannerDbType.Bool, DbType.Boolean };
+            yield return new object[] { new DateTime(2021, 1, 13, 12, 3, 10), SpannerDbType.Timestamp, DbType.DateTime };
+            yield return new object[] { 3.14D, SpannerDbType.Float64, DbType.Double };
```
amanda-tarafa, line 57: Can you add a case for float and decimal as well which currently map to Float64? Actually, for all of these, be exhaustive.

amanda-tarafa (changes requested):
A few nits.
But, I suppose this isn't being used anywhere?? Or how did we manage to be without these mappings for SpannerNumeric before?
If it's being used somewhere, can you please at a test for that?
Or are we just using the conversions? If these mappings are really not being used, maybe we could consider depracating them?

olavloite:
Thanks for the review. I'll try to look into your points later today.

> A few nits.
> But, I suppose this isn't being used anywhere?? Or how did we manage to be without these mappings for SpannerNumeric before?

These are only being used if the client application _only_ sets a _value_ and not a type. This is not very common when working specifically with the Spanner client library, but is more common when working with it through a framework that treats it like a generic `DbConnection`. So something like this would cause trouble:

```csharp
SpannerConnection connection = ...
var cmd = connection.CreateDbCommand();
var param = cmd.CreateDbParameter();
param.Value = (SpannerNumeric)3.14M;
```

When executing this command the Spanner client would try to infer the type from the value [here](https://github.com/googleapis/google-cloud-dotnet/blob/7adf800d20142a1ecd2e88d27fde290daf2e15b7/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs#L121). In all other cases it would just use the type that the client application had supplied.

> If it's being used somewhere, can you please at a test for that?
I'll try to add some tests for using `SpannerCommand` with only a value and no specific type.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
jskeet: I'd suggest the opposite, actually - I suggest we fix the suffixes above to be lower-case, as IMO that's more idiomatic C# :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
jskeet: +1 to this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
amanda-tarafa: I actually debated which was best, but got "confused" (is it lower case that I normally use, upper case?) and then I found [this](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types#integer-literals) (see the note there). Maybe lower case for anything except L? But I'm fine with the most idiomatic/consistent we can find ;).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -251,13 +271,6 @@ internal V1.Type ToProtobufType()
         internal static SpannerDbType ForStruct(SpannerStruct spannerStruct) =>
             new SpannerDbType(TypeCode.Struct, spannerStruct.Select(f => new StructField(f.Name, f.Type)).ToList());
 
-        /// <summary>
```
olavloite, line 252: While going through this class I also stumbled across this one. It is not in use anywhere and does not really make any sense.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
olavloite: Done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
olavloite: Done (everything lower case except for long). I couldn't really find anything authoritative convention for this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
olavloite: I've added additional test cases for the currently existing mappings.

olavloite:
> I'll try to add some tests for using `SpannerCommand` with only a value and no specific type.

I just realized that any test like that would just be a repetition of the `ValueMappings` test, which is literally doing this already.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -189,6 +191,26 @@ public System.Type DefaultClrType
...
+                DbType.Date => Date,
+                DbType.DateTime => Timestamp,
+                DbType.Double => Float64,
+                DbType.Int64 => Int64,
+                DbType.VarNumeric => Numeric,
+                DbType.Object => Unspecified,
+                DbType.String => String,
+                _ => throw new ArgumentOutOfRangeException(nameof(DbType), dbType, null),
+            };
+        }
```
jskeet, line 212:
```suggestion
        internal static SpannerDbType FromDbType(DbType dbType) => dbType switch
        {
            DbType.Binary => Bytes,
            DbType.Boolean => Bool,
            DbType.Date => Date,
            DbType.DateTime => Timestamp,
            DbType.Double => Float64,
            DbType.Int64 => Int64,
            DbType.VarNumeric => Numeric,
            DbType.Object => Unspecified,
            DbType.String => String,
            _ => throw new ArgumentOutOfRangeException(nameof(DbType), dbType, null),
        };
```

(We could do the same for the DbType property, but I'm not suggesting that change in this PR.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
jskeet: +1

(I suspect this is an inherent problem in the "let's introduce a feature latently, but not make it public yet" - when it *does* go public, it can be hard to make sure we catch everything like this.)

jskeet (approved):
Okay, this now looks good to me - worth just waiting for a final approval from Amanda as well, in case I've missed any compatibility issues, but I think this one *should* be non-breaking, given that we're only *adding* a mapping rather than changing one, yes?

olavloite:
> Okay, this now looks good to me - worth just waiting for a final approval from Amanda as well, in case I've missed any compatibility issues, but I think this one _should_ be non-breaking, given that we're only _adding_ a mapping rather than changing one, yes?

AFAICT the only 'breaking' change in this case is that if you currently try to only set a `SpannerNumeric` value on a parameter without also setting a type, you will get an error. After this change it will work. We are not changing the behavior of any of the other currently working (or failing) cases.

jskeet:
Forced the CLA bit: for some reason, when a suggestion from me is accepted, the CLA bot doesn't like it :(

amanda-tarafa (approved):
LGTM. I can't think of any compatibility issues here and this looks good.

olavloite:
@amanda-tarafa Would you mind merging the PR? I don't have permission to merge anything in this repo.

amanda-tarafa:
Ah yes, done!

PR #5845, opened 2021-01-15
Add Google.Cloud.Memcache.V1 to the API catalog

jskeet:
We'll release this once it's live, but I've just tried a smoke test which suggests it's not available yet.


amanda-tarafa (approved):
LGTM

PR #5858, opened 2021-01-19
fix: parse WITH command as SELECT

### LINKED ISSUE #5857: [Spanner] `WITH` clause is not recognized as query command type
The Cloud Spanner SQL dialect allows queries to start with either `SELECT` or `WITH`. The latter is currently not recognized as a keyword indicating the start of a `SELECT` command in the client library.

See https://cloud.google.com/spanner/docs/query-syntax for full query syntax reference.

Example query that is a valid Cloud Spanner query, but that is not recognized:

```sql
WITH AlbumsView AS (SELECT * FROM Albums)
select Title
from AlbumsView
```

> I suspect this is what needs changing: https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs

olavloite:
Parses `WITH` statements as queries and allows other whitespace characters than only ' '.

Fixes #5857

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
@@ -601,12 +601,12 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
                 // A statement that starts with a WITH clause should be treated as a query.
                 var cmd = connection.CreateCommand();
                 cmd.CommandText =
-                    $"WITH LastValue AS (" +
-                    $"  SELECT Key, StringValue" +
-                    $"  FROM {_fixture.TableName}" +
-                    $"  ORDER BY StringValue" +
-                    $"  LIMIT {int.MaxValue} OFFSET {_fixture.RowCount - 1}) " +
-                    $"SELECT * FROM LastValue";
+                    $@"WITH LastValue AS (
+                      SELECT Key, StringValue
+                      FROM {_fixture.TableName}
+                      ORDER BY StringValue
+                      LIMIT {int.MaxValue} OFFSET {_fixture.RowCount - 1})
+                    SELECT * FROM LastValue";
                 using (var reader = await cmd.ExecuteReaderAsync())
                 {
                     Assert.True(await reader.ReadAsync());
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
@@ -591,5 +591,28 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
                 }
             }
         }
+
+        [Fact]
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -217,7 +219,8 @@ public static SpannerCommandTextBuilder FromCommandText(string commandText)
         {
             GaxPreconditions.CheckNotNullOrEmpty(commandText, nameof(commandText));
             commandText = commandText.Trim();
-            var commandSections = commandText.Split(' ');
+            // Split(new char[0]) splits the string using all whitespace characters.
+            var commandSections = commandText.Split(new char[0], StringSplitOptions.RemoveEmptyEntries);
```
jskeet, line 223: The documentation suggests that we could use `(char[]) null` instead of `new char[0]`, which would avoid creating an empty array on every call.

jskeet (approved):
One suggested change, but otherwise looks good.

amanda-tarafa (commented):
LGTM. But can you add an integration test that uses WITH? Just to make extra certain.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: Good point, done.

olavloite:
> LGTM. But can you add an integration test that uses WITH? Just to make extra certain.

I've added an integration test case to the existing `ReadTests` class.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/ReadTests.cs
```diff
@@ -591,5 +591,28 @@ public async Task GetOrdinalAsync()
...
         }
+
+        [Fact]
+        public async Task WithCommand()
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                // Create a generic command and let the type be determined based on the command text.
+                // A statement that starts with a WITH clause should be treated as a query.
+                var cmd = connection.CreateCommand();
```
jskeet, line 602: Not sure why GitHub isn't showing me a suggestion option, but assuming Spanner doesn't mind line breaks, this can be simplified to:

                cmd.CommandText = $@"
                     WITH LastValue AS (
                        SELECT Key, StringValue
                        FROM {_fixture.TableName}
                        ORDER BY StringValue
                        LIMIT {int.MaxValue} OFFSET {_fixture.RowCount - 1})
                      SELECT * FROM LastValue";

jskeet (approved):
One suggestion which could be fixed later if it's inconvenient now, but other than that, I'm happy.

jskeet (approved):
Great, thanks. All looks good to me - let's just wait for @amanda-tarafa to give a final LGTM, then I'll merge.

amanda-tarafa (approved):
Thanks for the test! LGTM

PR #5884, opened 2021-01-26
Prepare for the next stage of Cloud Rad

jskeet:
This involves a bit more work *after* metadata generation, but *before* uploading to DevSite.
With this in place, we need to wait for:

- docuploader to handle JSON files
- doc-pipeline to handle tar files with separate api/ and examples/ directories

At that point, we'll just need to update uploaddocs.sh to tweak the stage where it currently uses bareapi.

### DIFF
--- a/tools/Google.Cloud.Tools.PostProcessDevSite/Program.cs
+++ b/tools/Google.Cloud.Tools.PostProcessDevSite/Program.cs
@@ -1,0 +1,262 @@
+﻿// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


tools/Google.Cloud.Tools.PostProcessDevSite/Program.cs
```diff
@@ -0,0 +1,262 @@
...
+            {
+                ["version"] = apiMetadata.Version,
+                ["language"] = "dotnet",
+                ["name"] = apiMetadata.Id,
+                ["xrefs"] = new JArray(GetXrefUrls(apiMetadata)),
+                ["xrefservices"] = new JArray("https://xref.docs.microsoft.com/query?uid={uid}")
+                // TODO: xrefs
+            };
+            string json = docUploaderMetadata.ToString(Formatting.Indented);
+            File.WriteAllText(Path.Combine(_devSiteRoot, "docs-metadata.json"), json);
```
jskeet, line 213: Note: this should be docs.metadata.json

amanda-tarafa (approved):
LGTM

PR #5894, opened 2021-02-01
Add script to generate utility package metadata

jskeet:
This includes a script to be run from a new Kokoro job, controlled
by environment variables - we don't need to run this on a regular
basis, so it doesn't matter that it's a little clunky.

### DIFF
--- a/docs/generate-devsite-utilities.sh
+++ b/docs/generate-devsite-utilities.sh
@@ -1,0 +1,181 @@
+#!/bin/bash
+
+source ../toolversions.sh
+
+install_docfx
+
+if [[ -z "$1" || -z "$2" ]]
+then
+  echo 'Please specify utility and version'
+  echo 'Known types: google-apis protobuf common-proto grpc gax'
+  exit 1
+fi
+
+if [[ "$3" != "" ]]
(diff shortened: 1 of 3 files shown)


docs/generate-devsite-utilities.sh
```diff
@@ -0,0 +1,181 @@
...
+
+set -eu -o pipefail
+
+declare -r DEVSITE_STAGING_BUCKET=docs-staging-v2
+declare -r VERSION=$2
+
+# These are listed in the order of dependencies:
+# each type only depends on types listed earlier
+case $1 in
+"protobuf")
```
amanda-tarafa, line 29: Is there a `(` missing here? And all the rest of the cases.

amanda-tarafa (approved):
LGTM.

docs/generate-devsite-utilities.sh
jskeet: Nope, that's just the syntax for bash cases

jskeet: Having said that, the indentation is a bit skew. Will fix that.

Issue #5425, opened 2020-10-12
[CI-Diagnostics]: Investigate and fix issue with outgoing traces and SSL certificate.


amanda-tarafa:
```
Error Message:
   System.Net.Http.HttpRequestException : An error occurred while sending the request.
---- System.Net.WebException : The request was aborted: Could not create SSL/TLS secure channel.
```
In tests:

```
Google.Cloud.Diagnostics.AspNetCore.Snippets.TraceSnippetsTests.Traces_OutgoingClientFactory
Google.Cloud.Diagnostics.AspNetCore.Snippets.TraceSnippetsTests.Traces_Outgoing
```

These only fail on Windows CI. They are green on local and on Linux CI.

I'm skipping these tests on Windows until I've taken a look at this.

amanda-tarafa:
Tests have been skipped in #5426

jskeet:
Closing as this will be handled internally.

PR #5904, opened 2021-02-03
[Diagnostics]: Un-skips outgoing trace tests.

amanda-tarafa:
They were being skipped on Windows only because of an issue with the testing environment's SSL certificate that should be fixed now.
See #5425


amanda-tarafa:
Manual run failed with the same error "Could not establish SSL connection". But we don't need to make a secured request for testing trace propagation, I've started making requests to `http://weather.com` instead of `https://weather.com`. Will run again once the currently running schedulled run is done.

amanda-tarafa:
Running manually again to test new changes. Will report back when done.

amanda-tarafa:
This ran OK after the latest changes. @jskeet PTAL when you have a chance.

jskeet (approved):
LGTM. Still weird that we can't get through to weather.com, but hey ho.

PR #5916, opened 2021-02-08
Updates gRPC default version to 2.35.0

amanda-tarafa:
(Releasing will gradually pick up the new version)


jskeet (commented):
Possibly change the generator version we use in toolversions.sh too?
(I suggest making that change and then seeing if anything actually changes - if it does, we'll want to regenerate everything in the same PR to avoid an autosynth avalanche.)

amanda-tarafa:
Ah yes, on both...

jskeet (approved):
As discussed offline, generation can come later.

amanda-tarafa:
AppVeyor is failing because projects need to be regenerated for non GAs.
I'm looking into fully regenerating though.

jskeet:
Just running generateprojects.sh should be enough - we can get this PR in, then look at generation.

amanda-tarafa:
Ran `generateproject.sh`, only non-GA projects have been updated as expected.
Waiting for green before merging...

PR #5924, opened 2021-02-09
Improve clean-up of StandaloneSnippets

jskeet:
(Not every API starts with Google.)

Also remove other clean-up which is now redundant due to fixes in the generator.


amanda-tarafa (approved):
LGTM.

PR #5935, opened 2021-02-10
Improve ordering of guides in DevSite docs

jskeet:
(We could attempt to have fuller control, but this is sufficient for
now.)


amanda-tarafa (approved):
LGTM

PR #5942, opened 2021-02-12
Add Google.Apps.Script.Type

jskeet:
This isn't an API, it's a "common" package - which will then let us generate Google.Cloud.GSuiteAddOns.V1 next week.
(We won't release for a little while though.)


amanda-tarafa (approved):
LGTM

Issue #5948, opened 2021-02-18
Skip the CommitStats integration tests against the emulator


skuruppu:
Not sure why it wasn't picked up in #5506 presubmits but the CommitStats integration test is failing against the emulator. We need to skip it until the support is added to the emulator.

skuruppu:
@olavloite if you could please take this on, that would be great. It is apparently blocking the emulator release.

skuruppu:
For some reason I can't assign this to anyone other than myself so you will have to assign this to yourself @olavloite.

olavloite:
> Not sure why it wasn't picked up in #5506 presubmits but the CommitStats integration test is failing against the emulator. We need to skip it until the support is added to the emulator.

I'll look into it right away. The reason that it wasn't picked up by the presubmits, is that as far as I know we are not running any of the integration tests against the emulator during presubmits for the C# client. They are only executed against a real Cloud Spanner instance.

amanda-tarafa:
> The reason that it wasn't picked up by the presubmits, is that as far as I know we are not running any of the integration tests against the emulator during presubmits for the C# client. They are only executed against a real Cloud Spanner instance.

On presubmits we don't run integration tests against a real instance. I think we run Spanner integration tests against the emulator if the PR makes changes to Spanner libraries, #5480 actually failed against the emulator but #5506 didn't.  
Also, CI has been green since #5506 was merged and I believe that on CI we run the tests against the emulator always. So something else is going on, either the emulator version we are using does support the feature or the tests are not being even run against the emulator or the test is wrong (which doesn't seem to be the case).  
I'll take a look at our CI scripts now to confirm that we are really running the tests against the emulator.

amanda-tarafa:
Part of the mistery solved, we moved from Travis to GitHub Actions in between #5480 and #5506 and we accidentally dropped running the tests against the emulator when there were changes in Spanner. I'm adding that back in now. Sorry!

skuruppu:
Oh nice work figuring out the root cause. I didn't know that we had moved to GH Actions.

amanda-tarafa:
And as far as I can see we have never ran the integration tests against the emulator on our nightly CI run, so that's all mysteries solved.

PR #5950, opened 2021-02-18
test: skip CommitStats test on emulator

olavloite:
Skips the CommitStats integration test when running on the emulator, as the emulator does not yet support this feature.

This change also adds automatic creation of a test instance on the emulator when needed. This makes it a lot easier to run integration tests on the emulator, as you don't have to manually create an instance on the emulator first before running the integration tests.

Fixes #5948

PR #5952, opened 2021-02-18
ci: Runs Spanner tests in emulator on presubmits.

amanda-tarafa:
Only when there are changes to Spanner libraries. It is the shell script that checks for changes in Spanner before running.

Towards #5948.

@olavloite FYI.

### DIFF
--- a/.github/workflows/spanner-emulator-pr-push.yml
+++ b/.github/workflows/spanner-emulator-pr-push.yml
@@ -1,0 +1,36 @@
+name: Spanner tests against emulator (Spanner changes only)
+
+on:
+  pull_request:
+    paths:
+    - '**[Ss]panner**'
+  push:
+    branches:
+      - master
+    paths:
+    - '**[Ss]panner**'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    env:
+      DOTNET_NOLOGO: true
+    services:
+      emulator:
+        image: gcr.io/cloud-spanner-emulator/emulator:latest
+        ports:
+          - 9010:9010
+          - 9020:9020
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+
+    - name: Spanner Emulator
+      env:
+          SPANNER_EMULATOR_HOST: localhost:9010
+          TEST_PROJECT: emulator-test-project
+      run: |
+        dotnet test -c Release ./apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests
+        dotnet test -c Release ./apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets
--- a/apis/Google.Cloud.Spanner.Data/run-emulator-tests.sh
+++ b/apis/Google.Cloud.Spanner.Data/run-emulator-tests.sh
@@ -17,51 +17,18 @@
 # Fail on any error.
 set -e
 
-# Check if we need to run
-if ! git diff master --name-only | grep -iq spanner
-then
-  echo "Skipping Spanner integration tests; no Spanner-related changes."
-  exit 0
-fi
-
-declare -r REPO_ROOT=$(readlink -f $(dirname ${BASH_SOURCE})/../..)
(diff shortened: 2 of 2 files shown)


amanda-tarafa:
Checking failures..

skuruppu (approved):

skuruppu (commented):
Hmmm I don't really know why there's that failure when running the tests.

I'm wondering now that it's possible to run these through GH Actions, then is it possible to have a simpler setup like in https://github.com/googleapis/nodejs-spanner/blob/master/.github/workflows/system-tests-against-emulator.yaml.

amanda-tarafa:
Ah yes, that looks better. And actually the failure was because we were explicitly using an old Cloud SDK version that didn't support the python version that we are installing. I'll try something like Node's and see what happens.

amanda-tarafa:
The failure is now the expected commit stats missing feature.
I'll review #5950 now so we can merge that one before looking at improving the actions setup.

amanda-tarafa:
The Spanner emulator action is passing now. Thanks! @olavloite for adding the instance creation bit on the tests, it simplified these .yml a lot.
I'll clean up the PR and resubmit for review.

amanda-tarafa:
Ready for review, @jskeet , @skuruppu , @olavloite PTAL. Thanks

.github/workflows/spanner-emulator-pr-push.yml
```diff
@@ -0,0 +1,46 @@
...
+    - '**[Ss]panner**'
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    env:
+      DOTNET_NOLOGO: true
+    services:
+      emulator:
+        image: gcr.io/cloud-spanner-emulator/emulator:latest
```
jskeet, line 20: Wow - that's simpler than I'd expected!

.github/workflows/spanner-emulator-pr-push.yml
```diff
@@ -0,0 +1,46 @@
...
+        ports:
+          - 9010:9010
+          - 9020:9020
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
+        fetch-depth: 100
```
jskeet, line 29: I suspect you don't need the fetch-depth here - it was really for diff purposes

.github/workflows/spanner-emulator-pr-push.yml
```diff
@@ -0,0 +1,46 @@
...
+
+    # The GitHub checkout action leaves the repo in a slightly awkward
+    # state. This tidies it up.
+    - name: Set up git branches
```
jskeet, line 33: I don't think you need this - again, it was just so that diffs would work.

.github/workflows/spanner-emulator-pr-push.yml
```diff
@@ -0,0 +1,46 @@
...
+        git checkout -b pr-head
+        git checkout -b master ${{ github.event.pull_request.base.sha }}
+        git checkout pr-head
+
+    - name: Spanner Emulator
+      env:
+          SPANNER_EMULATOR_HOST: localhost:9010
+          TEST_PROJECT: emulator-test-project
+      run: |
+        dotnet test ./apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests
```
jskeet, line 45: I think this will use the debug version - not sure whether it's worth specifying `-c Release`

apis/Google.Cloud.Spanner.Data/run-emulator-tests.sh
```diff
@@ -1,67 +0,0 @@
-#!/bin/bash
```
jskeet, line 1: Is it useful to keep this (possibly slimmed down a bit, now that we create the instance ourselves) for the sake of running locally?

amanda-tarafa (commented):
Comments addressed, PTAL.

skuruppu (approved):
This looks great. Thanks for working on this :)

jskeet (approved):
Lovely - thanks!

PR #5953, opened 2021-02-18
ci: Cleans up GitHub actions configs.

amanda-tarafa:
@jskeet  this is a proposal, not sure if you want to be explicit about Ubuntu, .NET and Python versions, but looking at the [GitHub-hosted runners spec](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners) it seems that we should be OK.

Also, making some changes here based on some of your comments in #5952.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -1,38 +1,22 @@
 name: Build and test PR (changed APIs only)
 
-on: [pull_request]	
+on: [pull_request]
 
-jobs:	
+jobs:
 
-  build:	
-    runs-on: ubuntu-18.04	
-    env:	
-      DOTNET_NOLOGO: true	
-    strategy:	
-      matrix:	
-        regex: ["'Google\\.Cloud\\.[A-L].*'", "'Google\\.Cloud\\.[M-Z].*'", "'!Google\\.Cloud'"]	
-    steps:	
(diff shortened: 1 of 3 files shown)


jskeet:
Will have a look when I've got the mental space to do this justice - may well be tomorrow.

amanda-tarafa:
Yep, no rush at all.

.github/workflows/build-pr.yml
```diff
@@ -1,49 +1,24 @@
...
+    env:
+      DOTNET_NOLOGO: true
+    strategy:
+      matrix:
+        regex: ["'Google\\.Cloud\\.[A-L].*'", "'Google\\.Cloud\\.[M-Z].*'", "'!Google\\.Cloud'"]
+    
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        submodules: true
```
jskeet, line 18: Please reinstate the fetch-depth here, and setting up the git branches - they're needed for `./build.sh --diff`

jskeet (approved):
Just a couple of changes for build-pr, but otherwise this looks very nice.

.github/workflows/build-pr.yml
amanda-tarafa: Ah yes, will do.

PR #5962, opened 2021-02-22
test(Storage): Loosens public access prevention tests.

amanda-tarafa:
This is to match what we are getting from the server.


amanda-tarafa:
@frankyn and @tritone I've had to make these changes in tests because we are either getting (from the backend) a `null` or an `unspecified` value in `Bucket.IamConfiguration.PublicAccessPrevention` when it is not configured. Up until now we had been consistently getting a `null` value.
This is effectively a breaking change on the API.

jskeet (approved):

tritone:
Discussed offline with @amanda-tarafa , this is a bug fix on the backend. The feature is not yet public so we don't have to worry about a breaking change.

PR #5971, opened 2021-02-24
Improve TOC handling for DevSite

jskeet:
We now group by assembly rather than by namespace; if an assembly
contains multiple namespaces, those are listed within the assembly
node.


jskeet:
Note: all the changes here *only* affect DevSite, so we shouldn't run into any compatibility issues etc.

Once this is in and we can try it out, it should be easy to test things like tweaking handling of nested types.

amanda-tarafa (approved):
LGTM.

PR #5984, opened 2021-02-26
Release Google.Cloud.Metastore.V1Alpha version 1.0.0-alpha01

jskeet:
Changes in this release:

Initial release.


jskeet:
Note: although this generates a NetworkName resource that we might want to call ComputeNetworkName, we can do that later with a breaking change once we've got suitable tooling; this is only an alpha API so breaking changes are okay.

amanda-tarafa (approved):
LGTM

PR #5992, opened 2021-03-02
Fix diff PR check

jskeet:


jskeet:
(It took a couple of tries, and it turns out we need the latest preview, but it does fix the problem. I'll keep an eye on the dependency and update it again when there's a fuller release containing the fix.)

amanda-tarafa (approved):
LGTM

PR #5999, opened 2021-03-03
Include history-less dependencies within parent version history

jskeet:
This isn't as much of a change as it looks. The biggest "new code" is the private GitHelpers.GetHistoryApiIds method. The rest is mostly changes due to needing the ApiCatalog instead of just the API.


jskeet:
Tested via Spanner - I made a change in Google.Cloud.Spanner.Common.V1, and another in Google.Cloud.Spanner.Admin.Database.V1, then faked a release of all the Spanner libraries - both changes (as well as the CMEK one) were picked up.

amanda-tarafa (approved):
LGTM. This is cool!

PR #6006, opened 2021-03-05
Separate out the Devsite docfx.json from googleapis.dev

jskeet:
Building the metadata for both in the same docfx run creates
duplicate inheritance lines. This is no doubt a bug in docfx, but
unlikely to get any attention as this is an unusual use of docfx.
It's simpler just to run docfx twice.


amanda-tarafa (approved):
LGTM

PR #6013, opened 2021-03-08
fix: retry errors in stream until timeout

### LINKED ISSUE #5977: Spanner: We may not be retrying requests when WithSessionExpiryChecking() fails with UNAVAILABLE
We got a customer issue reported where they are using the `SpannerDataReader` and they see an error in their logs:

```
Exception details: google.Cloud.Spanner.Data.SpannerException: The service is currently unavailable. This is a most likely a transient condition.

---> Grpc.Core.RpcException: Status(StatusCode="Unavailable", Detail="GOAWAY received", DebugException="Grpc.Core.Internal.CoreErrorDetailException:
```

There doesn't seem to be any impact on the workload from this error other than it taking a long time when this happens. Looking at the code, an UNAVAILABLE error is [caught](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs#L182) and [retried](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs#L244). But that's only if [`hasNext`](https://github.com/googleapis/google-cloud-dotnet/blob/13a903abdc395fcdb8c22dbe0434a1fa0288cf8e/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs#L140) is computed and [`_safeToRetry`](https://github.com/googleapis/google-cloud-dotnet/blob/13a903abdc395fcdb8c22dbe0434a1fa0288cf8e/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs#L154) is set.

I don't think it's getting this far based on the below stack trace because I see [`WithExpirationChecking()`](https://github.com/googleapis/google-cloud-dotnet/blob/master/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs#L142).

BSPLuxw3GsCLTMD

My question is will the RPC be retried elsewhere even if the `UNAVAILBLE` error is not caught in this loop?

CC @olavloite

olavloite:
Retry consecutive errors that are retryable until either the call succeeds or until the deadline of the call has been exceeded.

Fixes #5977


jskeet:
Hoping to get to this today or tomorrow.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
```diff
@@ -111,7 +111,12 @@ public async Task<bool> MoveNext(CancellationToken cancellationToken)
         private async Task<PartialResultSet> ComputeNextAsync(CancellationToken cancellationToken)
         {
             // The retry state is local to the method as we're not trying to handle callers retrying.
-            RetryState retryState = new RetryState(_client.Settings.Scheduler ?? SystemScheduler.Instance, _retrySettings);
+            // Consecutive retryable errors will be retried until the call succeeds, or until the call times out or is cancelled.
+            RetryState retryState = new RetryState(_client.Settings.Scheduler ?? SystemScheduler.Instance, _retrySettings, maxConsecutiveErrors: int.MaxValue);
+            var clock = _client.Settings.Clock ?? SystemClock.Instance;
+            var startTime = clock.GetCurrentDateTimeUtc();
+            var originalTimeout = _callSettings.Expiration?.Timeout ?? _client.Settings.ExecuteStreamingSqlSettings.Expiration?.Timeout ?? TimeSpan.FromMilliseconds(3600000);
```
amanda-tarafa, line 118: I think that if the Expiration is a deadline, it should be respected as well.

amanda-tarafa, line 118: nit: can the ultimate default value (the 1hr) be set somewhere in a  static variable or similar?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
```diff
@@ -111,7 +111,12 @@ public async Task<bool> MoveNext(CancellationToken cancellationToken)
         private async Task<PartialResultSet> ComputeNextAsync(CancellationToken cancellationToken)
         {
             // The retry state is local to the method as we're not trying to handle callers retrying.
-            RetryState retryState = new RetryState(_client.Settings.Scheduler ?? SystemScheduler.Instance, _retrySettings);
+            // Consecutive retryable errors will be retried until the call succeeds, or until the call times out or is cancelled.
+            RetryState retryState = new RetryState(_client.Settings.Scheduler ?? SystemScheduler.Instance, _retrySettings, maxConsecutiveErrors: int.MaxValue);
```
amanda-tarafa, line 115: RetryState is an internal class. If `maxConsecutiveErrors` is not meaningful anymore maybe we can just remove it.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
```diff
@@ -188,6 +194,14 @@ private async Task<PartialResultSet> ComputeNextAsync(CancellationToken cancella
                     _buffer.Clear();
                     _grpcCall.Dispose();
                     _grpcCall = null;
+
+                    // Calculate the remaining timeout.
```
amanda-tarafa, line 198: Can't all of this be moved to RetryState so we can just keep asking CanRetry as we are doing now?

amanda-tarafa: Maybe we can do something similar to what we have in [Gax](https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax.Grpc/ApiCallRetryExtensions.cs).

amanda-tarafa (commented):
I've left a few comments.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
jskeet: (It's slightly fiddly though, in terms of whether CanRetry should check for timing. I'd be open to a bit of restructuring here - we don't want to duplicate checking in CanRetry and RecordErrorAndWaitAsync. Maybe CanRetry should work out how long the retry should be for as an out parameter, and then just call the scheduler delay directly in the finally block?)

jskeet (commented):
Some initial thoughts. I'm still not thinking entirely clearly, which doesn't put me into a great state for retry work, which is always fiddly.

I think it might make sense to do this in multiple commits though. Start with a commit that *just* removes the maximum consecutive failures, then another commit that observes the deadline/timeout more accurately.

We may well want to refactor RetryState significantly here... but I'd like to keep as much retry-related complexity as possible out of the main loop. (It probably won't *all* be movable, but we should do what we can.)

olavloite:
> I think it might make sense to do this in multiple commits though. Start with a commit that _just_ removes the maximum consecutive failures, then another commit that observes the deadline/timeout more accurately.

I've created a [separate PR](https://github.com/googleapis/google-cloud-dotnet/pull/6015) for that, as that will fix the current problems that clients are experiencing. I've added a somewhat randomly chosen max retry count of 115 to prevent the theoretical possibility of an eternal loop.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
olavloite: I've updated the implementation to respect a deadline as well. There is however one potential issue with using a deadline here:
1. The `CallSettings` are passed in to constructor of `SqlResultStream`, for example when [ExecuteQueryAsync](https://github.com/googleapis/google-cloud-dotnet/blob/34b71e8b87efdbc3043906d9b2dd7d1b8e9bb43d/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs#L141) is called by a client application.
2. That means that the same `CallSettings` are used for all calls to `MoveNext()`, which means that a deadline is applied to the entire stream and not to a single RPC.

Currently it's not a real-life problem as a user can't pass in custom `CallSettings` and the default `CallSettings` currently use a timeout. Each timeout will only apply to a single RPC invocation.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
olavloite: Done.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
olavloite: I've changed the implementation to using a deadline instead of calculating a remaining timeout. I'm not sure we should add the timing check to `CanRetry`:
1. If the deadline is exceeded, a `DeadlineExceeded` exception should be thrown and not the otherwise retryable error that occurred.
2. It does not seem logical to let `CanRetry` throw a `DeadlineExceeded` error if the error cannot be retried because it would exceed the deadline.
3. Letting `CanRetry` return false and a deadline that is in the past would not make this construct any simpler, as it would mean that the catch block would have to take that condition into account as well.

Another reason I would prefer not to have `CanRetry` return the delay as an output parameter is that it would not be deterministic because of the jittering of the retry delay (or we would have to delay the jittering until the actual wait is executed).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SqlResultStreamTests.cs
```diff
@@ -244,5 +331,18 @@ public override ExecuteStreamingSqlStream ExecuteStreamingSql(ExecuteSqlRequest
                 return new ExecuteStreamingSqlStreamImpl(call);
             }
         }
+
+        private sealed class FakeScheduler : IScheduler
```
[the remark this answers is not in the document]
olavloite: I've updated the name. The standard `FakeScheduler` would probably also work, albeit with a little extra code, as it requires the `FakeClock` to be advanced manually by the test code. This local implementation automatically advances the `FakeClock` when the `Delay` method is called.

olavloite:
@jskeet @amanda-tarafa PTAL, I've updated the implementation to use a deadline instead of calculating a remaining timeout.

jskeet:
I'll try to have a look tomorrow.

skuruppu (approved):

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
```diff
@@ -256,9 +247,13 @@ internal bool CanRetry(RpcException exception)
             /// </summary>
             internal async Task RecordErrorAndWaitAsync(RpcException exception, CancellationToken cancellationToken)
             {
-                _consecutiveErrors++;
                 TimeSpan? delayFromException = GetRetryDelay(exception);
-                await _scheduler.Delay(_retrySettings.BackoffJitter.GetDelay(delayFromException ?? _retrySettingsBackoffs.Current), cancellationToken).ConfigureAwait(false);
+                TimeSpan delay = _retrySettings.BackoffJitter.GetDelay(delayFromException ?? _retrySettingsBackoffs.Current);
+                if (_clock.GetCurrentDateTimeUtc() + delay >= _currentDeadline)
+                {
+                    throw new RpcException(new Grpc.Core.Status(StatusCode.DeadlineExceeded, "Streaming call exceeded timeout"));
```
amanda-tarafa, line 254: nit: should this be a SpannerException? It's our combined operation that has timed out, not a specific RPC? If a specific RPC times out, then CanRetry should be false and the RpcException for the deadline should be thrown.
Not sure though if it makes sense to make that distinction.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
amanda-tarafa: > That means that the same CallSettings are used for all calls to MoveNext(), which means that a deadline is applied to the entire stream and not to a single RPC

I might be missing something here, but I don't see how this is a problem. A deadline is a deadline, and applying it to the whole operation means that it is effectively applied to each single RPC. There's no way that the whole operation can finish before the deadline if the single RPCs didn't all finish before the deadline?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
olavloite: I think it's better to throw an `RpcException` here to keep it as similar to other `DeadlineExceeded` exceptions that might occur. This `RpcException` is [automatically wrapped](https://github.com/googleapis/google-cloud-dotnet/blob/34b71e8b87efdbc3043906d9b2dd7d1b8e9bb43d/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs#L309) in a `SpannerException`, and anyone catching that exception and then unwrapping the underlying `RpcException` would then get the same behavior regardless whether the exception is thrown here or whether it actually occurs while the RPC is being executed (which in this case could also happen if for example the backoff time is less than the remaining time to the deadline, but the deadline is reached while retrying the RPC after the backoff).

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
```diff
@@ -256,9 +247,13 @@ internal bool CanRetry(RpcException exception)
             /// </summary>
             internal async Task RecordErrorAndWaitAsync(RpcException exception, CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
olavloite: Done.

olavloite:
@amanda-tarafa 
> I might be missing something here, but I don't see how this is a problem. A deadline is a deadline, and applying it to the whole operation means that it is effectively applied to each single RPC. There's no way that the whole operation can finish before the deadline if the single RPCs didn't all finish before the deadline?

It's not really a big problem, but what I meant is the following:
1. If the returned stream never breaks and we never have to retry any calls, there's no problem at all. The deadline will only be applied to the initial RPC and the client application can consume the stream without any issues, regardless how much time the application uses to consume the entire stream.
2. If the returned stream however breaks somewhere halfway with a retryable error, a new RPC invocation will be executed using the initial deadline and the last seen resume token. If the client application has been slow to consume the data so far, it might be that the deadline has now been exceeded, and that the new RPC invocation will fail with a `DeadlineExceeded` error. That error would not have occurred if the stream did not break halfway.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SqlResultStream.cs
amanda-tarafa: OK

amanda-tarafa (approved):
LGTM

amanda-tarafa:
> 2\. If the returned stream however breaks somewhere halfway with a retryable error, a new RPC invocation will be executed using the initial deadline and the last seen resume token. If the client application has been slow to consume the data so far, it might be that the deadline has now been exceeded, and that the new RPC invocation will fail with a `DeadlineExceeded` error. That error would not have occurred if the stream did not break halfway.

Ah yes, I see what you mean. I think it's fine though, in my view the alternatives are even less intuitive:

1. Ignore the deadline.
2. Stop recalculating the deadline from the timeout on the RetryState.Reset, so that the timeout applies from the beginning of the operation and not for each possible RPC.
3. On the first RPC calculate a timeout based on the deadline (from this moment on, how much till we get to the deadline) and apply that to all possible RPCs as a timeout.

So yes, I think that it's fine how it is and it's not a big problem.

amanda-tarafa:
@olavloite there are some conflicts with the main branch apparently. I'm happy to merge once those are resolved. Thanks.

olavloite:
> @olavloite there are some conflicts with the main branch apparently. I'm happy to merge once those are resolved. Thanks.

I think those conflicts were fixed with the last commit, so please go ahead and merge if this is otherwise OK.

amanda-tarafa:
Merging is disabled because of the conflicts.

olavloite:
> Merging is disabled because of the conflicts.

Apparently I now also have merge permission, so I went ahead and merged the PR. I'm not sure why it was marked as blocked for merging for you because of merge conflicts.

PR #6015, opened 2021-03-09
fix: retry consecutive retryable errors in sql stream without timeout

### LINKED ISSUE #5977: Spanner: We may not be retrying requests when WithSessionExpiryChecking() fails with UNAVAILABLE

olavloite:
Retry consecutive retryable errors for the same resume token in a stream of `PartialResultSet`s. This will prevent `SpannerDataReader` from throwing `Unavailable` exceptions while reading data. The `ExecuteSqlStreaming` RPC will be retried at most 115 times for the same resume token. This value is chosen as it will cause retries to be attempted for roughly the same amount of time as the default timeout of ExecuteStreamingSql (1 hour), and prevents an eternal loop in the unlikely case that the backend continues to return `Unavailable`.

Fixes #5977

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SqlResultStreamTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SqlResultStreamTests.cs
@@ -108,10 +108,23 @@ namespace Google.Cloud.Spanner.V1.Tests
             // We expect calls of:
             // - Start: no token, MoveNext(yes), MoveNext(throw)
             // - Start: token1, MoveNext(throw)
-            // - No further calls, due to second failure
-            // Only the result with token1 is returned
+            // - Start: token1, succeeds until the end of the stream.
+            await AssertResultsAsync(stream, results);
+            Assert.Equal(3, client.Calls);
+        }
+
+        [Fact]
+        public async Task MultipleErrors_ForSameResumeToken_DoesNotRetryForever()
+        {
+            var results = CreateResultSets("token1", "token2");
+            var client = new FakeSpannerClient(results, CreateExceptionFilter("token2", RetriableStatusCode, 120));
+            var stream = CreateStream(client);
+
+            // Expected calls:
+            // - Start: no token, MoveNext(yes), MoveNext(throw)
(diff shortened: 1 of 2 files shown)


jskeet (approved):
I definitely like this in terms of being a quick fix that will unblock things while we try to get a clean implementation that respects timeouts more appropriately.

skuruppu (approved):

amanda-tarafa (approved):
LGTM

PR #6029, opened 2021-03-15
[apis/Google.Cloud.PubSub.V1] docs: Improve PubSub coding guidance

### LINKED ISSUE #5889: PubSub: High TCP Connection usage
I recently upgraded my dev PC (running ubuntu based pop_os) to Ryzen 5900x and found I was regularly seeing the following two exceptions while running a service that is subscribed to 36 subscriptions using SubscriberClient's and also publishing to the same number using PublisherClient's.
```
Grpc.Core.RpcException: Status(StatusCode="Unavailable", Detail="failed to connect to all addresses", DebugException="Grpc.Core.Internal.CoreErrorDetailException: {"created":"@1611699451.269672651","description":"Failed to pick subchannel","file":"/var/local/git/grpc/src/core/ext/filters/client_channel/client_channel.cc","file_line":4090,"referenced_errors":[{"created":"@1611699451.269671371","description":"failed to connect to all addresses","file":"/var/local/git/grpc/src/core/ext/filters/client_channel/lb_policy/pick_first/pick_first.cc","file_line":394,"grpc_status":14}]}")
```
```
Grpc.Core.RpcException: Status(StatusCode="DeadlineExceeded", Detail="Deadline Exceeded", DebugException="Grpc.Core.Internal.CoreErrorDetailException: {"created":"@1611699451.222505294","description":"Deadline Exceeded","file":"/var/local/git/grpc/src/core/ext/filters/deadline/deadline_filter.cc","file_line":69,"grpc_status":4}")
```
We maintain a single instance of PublisherClient per topic.

After some debugging, I noticed that my process was holding over 900 TCP connections open and so after manually setting the client count on creation to 1 for both Subscriber and Publisher I found that all of these exceptions went away, and my process now holds a fewer TCP connections open. The overall throughput of my service does not seem greatly diminished (although I will need to run more tests to check this). I was under the assumption that gRPC should multiplex connections to the same host, but, looking through the [code](https://github.com/googleapis/google-cloud-dotnet/blame/cf77dacdc306fc4eaff0ad6eb3ab47eaf34ef748/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs#L241) I can see that it specifically prevents this, whilst also creating a channel per CPU core for both publishers and subscribers. 

Should I be worried that I was seeing the above exceptions when using the default settings? 
Do you have any recommendations to avoid/diagnose these sorts of problems quickly?
Do you have any information on why the code avoids channel multiplexing?

### LINKED ISSUE #5893: PubSub: PublisherClient.CreateAsync and ShutdownAsync
I have a quick question and some minor feedback on the PubSub library.

[According to the documentation](https://googleapis.dev/dotnet/Google.Cloud.PubSub.V1/2.3.0/index.html) a `PublisherClient` must be created via the `PublisherClient.CreateAsync` method and later shut down via `publisher.ShutdownAsync`.

My quarrel or misunderstanding with these methods is that they feel very unidiomatic and difficult to use in .NET. [In issue #3447](https://github.com/googleapis/google-cloud-dotnet/issues/3447) @chrisdunelm suggests that the `PublisherClient` should be re-used as much as possible:

> Reuse the publisher and subscribers as much as possible

This suggests to me to initialise a `PublisherClient` in the constructor of a singleton service, however due to the async nature of the factory it feels wrong and pushes me (and probably many other users) to believe that it should be a short lived transient dependency initialised at the time of publishing a message.

Which one should it be?

My even bigger confusion is around the `ShutdownAsync` method:

```csharp
// PublisherClient instance should be shutdown after use.
// The TimeSpan specifies for how long to attempt to publish locally queued messages.
await publisher.ShutdownAsync(TimeSpan.FromSeconds(15));
```

What is this method doing and what does it mean to pass in a 15 second timespan? Shall I await this method and will it then block me for 15 seconds or shall I fire and forget it without an await?

Assuming I have a singleton instance of `PublisherClient` does the `ShutdownAsync` method cause an issue if I don't invoke it until the application ends? Or perhaps the `ShutdownAsync` method is a bit like a flush with loggers and should get invoked more frequently? If it should get called more frequently then how frequent? At the end of a series of published messages or can it also get invoked at the end of every message published? Remember that most people will probably wrap the logic to publish a single message in some sort of service, and therefore such class design could cause the `ShutdownAsync` method to be invoked way too often, which is why I want to really understand what it's supposed to do.

The documentation is extremely thin on the proper usage and the meaning of those methods. This is very unidiomatic .NET code making it very hard even for very experienced developers to defer the meaning/purpose of the methods by looking at their signature. Furthermore the `ShutdownAsync` method converts the `TimeSpan` into a `CancellationToken` which is even weirder, but doesn't let me to pass in a cancellation token myself.

If `ShutdownAsync` is supposed to do some clean up work at the end of its lifecycle then would it have made more sense to implement the `IDisposable` or `IDisposableAsync` interface to steer .NET developers to do the correct thing?

Overall I'm really not clear how to manage a `PublisherClient` in order to avoid unnecessary long work, blocking code and to be safe. What is the best practice for this class?

chrisdunelm:
Fixes #5893
Fixes #5889

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/docs/index.md
+++ b/apis/Google.Cloud.PubSub.V1/docs/index.md
@@ -48,18 +48,18 @@ the CPU processor count as returned by `Environment.ProcessorCount`. This is to 
 than a single gRPC channel can support; the processor count is is a pragmatic choice to approximately
 scale maximum throughput performance by potential machine workload.
 
-If using multiple clients on a machine with a high processor count, then this may cause problems
+When using multiple clients on a machine with a high processor count, this may cause problems
 with TCP connection exhaustion. Set the relevant `ClientCreationSettings.ClientCount` to a low value
 (`1` is suitable for low or moderate throughput requirements) to mitigate this.
 
 # Coding considerations
 
-`PublisherClient` and `SubscriberClient` are expensive to create, so if regularly publishing or
+`PublisherClient` and `SubscriberClient` are expensive to create, so when regularly publishing or
 subscribing to the same topic or subscription then a singleton client instance should be created and
 used for the lifetime of the application.
 
 Both synchronous `Create(...)` and asynchronous `CreateAsync(...)` methods are provided, but note that
-if using default credentials on Google Compute Engine (GCE) then a network request may need to be made
+when using default credentials on Google Compute Engine (GCE) then a network request may need to be made
 to retrieve credentials from the GCE Metadata Server.
 
 When publishing, the `Task` returned by the various `Publish(...)` methods will complete only
--- a/apis/Google.Cloud.PubSub.V1/docs/index.md
+++ b/apis/Google.Cloud.PubSub.V1/docs/index.md
@@ -4,17 +4,6 @@
 
 {{version}}
 
-# Class renaming in [v1.0.0-beta16](https://www.nuget.org/packages/Google.Cloud.PubSub.V1/1.0.0-beta16)
-
-The following classes have been renamed when moving from version 1.0.0-beta15 to 1.0.0-beta16:
-
-* `SimplePublisher` is now named [PublisherClient](obj/api/Google.Cloud.PubSub.V1.PublisherClient.yml)
-* `SimpleSubscriber` is now named [SubscriberClient](obj/api/Google.Cloud.PubSub.V1.SubscriberClient.yml)
-* `PublisherClient` is now named [PublisherServiceApiClient](obj/api/Google.Cloud.PubSub.V1.PublisherServiceApiClient.yml)
-  * `PublisherSettings` is now named `PublisherServiceApiSettings`
-* `SubscriberClient` is now named [SubscriberServiceApiClient](obj/api/Google.Cloud.PubSub.V1.SubscriberServiceApiClient.yml)
-  * `SubscriberSettings` is now named `SubscriberServiceApiSettings`
-
 {{installation}}
 
 {{auth}}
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -47,6 +36,44 @@ Using [PublisherServiceApiClient](obj/api/Google.Cloud.PubSub.V1.PublisherServic
...
+and default settings have been chosen with this in mind;
+however, note that these classes are also well suited to use cases where performance is not a major
+consideration.
+
+By default multiple gRPC channels are created on client startup, with the channel count defaulting to
+the CPU processor count as returned by `Environment.ProcessorCount`. This is to allow greater bandwidth
+than a single gRPC channel can support; the processor count is is a pragmatic choice to approximately
+scale maximum throughput performance by potential machine workload.
+
+If using multiple clients on a machine with a high processor count, then this may cause problems
```
jskeet, line 51:
```suggestion
When using multiple clients on a machine with a high processor count, this may cause problems
```

("If using" doesn't sound quite right to me. "If you are using" would be okay, as an alternative)

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -47,6 +36,44 @@ Using [PublisherServiceApiClient](obj/api/Google.Cloud.PubSub.V1.PublisherServic
...
+with TCP connection exhaustion. Set the relevant `ClientCreationSettings.ClientCount` to a low value
+(`1` is suitable for low or moderate throughput requirements) to mitigate this.
+
+# Coding considerations
+
+`PublisherClient` and `SubscriberClient` are expensive to create, so if regularly publishing or
```
jskeet, line 57: Again, I'd prefer "when regularly publishing" or "if you are regularly publishing"

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -47,6 +36,44 @@ Using [PublisherServiceApiClient](obj/api/Google.Cloud.PubSub.V1.PublisherServic
...
+subscribing to the same topic or subscription then a singleton client instance should be created and
+used for the lifetime of the application.
+
+Both synchronous `Create(...)` and asynchronous `CreateAsync(...)` methods are provided, but note that
+if using default credentials on Google Compute Engine (GCE) then a network request may need to be made
```
jskeet, line 62: And again :)

jskeet (approved):
A few wording nits, but otherwise great, thanks.

apis/Google.Cloud.PubSub.V1/docs/index.md
chrisdunelm: Done (not using your suggestion due to CLA issue)

apis/Google.Cloud.PubSub.V1/docs/index.md
chrisdunelm: Done

apis/Google.Cloud.PubSub.V1/docs/index.md
chrisdunelm: Done

PR #6149, opened 2021-03-18
Provide a cleaner method for configuring API-specific resources

jskeet:
This introduces a new field into the API catalog:
`commonResourcesConfig`. When this field is present, it must refer
to a file which is used to augment the root common resources config.

The generation script no longer detects the presence of a file in
the API being generated. Rationale: usually the reason for having an
API-specific config is so that multiple APIs can use the same
config. Specifying which file should be used means we get less
duplication (we duplicate the file name in apis/apis.json instead of
the file contents).

The rest of this commit refactors the existing configuration into
the new scheme. Note that the Bigtable and Spanner resources are now
not present in the root config.


generateapis.sh
```diff
@@ -88,12 +88,16 @@ generate_microgenerator() {
   fi
 
   # Defalt to "all resources are common" but allow a per-API config file too.
```
amanda-tarafa, line 90:
```suggestion
  # Default to "all resources are common" but allow a per-API config file too.
```

amanda-tarafa (approved):
LGTM

generateapis.sh
jskeet: Doh, thanks!

PR #6259, opened 2021-03-20
feat: Add CCAI API

jskeet:


jskeet:
Self-merging to unblock autosynth

PR #6286, opened 2021-03-29
Generate GAPIC metadata files

jskeet:
This depends on https://github.com/googleapis/gapic-generator-csharp/pull/320

Once merged, API generation should be rerun manually to add all the
metadata files in a single pass rather than autosynth creating a
bunch of PRs.

PR #6290, opened 2021-03-30
Initial GAPIC metadata run

jskeet:
(Done in a single PR to avoid a slew of autosynth PRs.)

This is just the result of regenerating all APIs using the new
generator, with the option enabled in #6286.


amanda-tarafa (approved):
LGTM

PR #6328, opened 2021-04-09
Resource manager v3

jskeet:
Notes:

- The first commit has more in than normal, as ResourceManager is special in terms of resource names
- There are possible smoke tests (although the generator suggests DeleteProject, somewhat scarily!) but they don't currently work with all projects due to that service not being GA yet; will add that smoke test when it goes GA
- No release yet for the same reason


amanda-tarafa (approved):
LGTM

PR #6329, opened 2021-04-09
Populate library_type in repo metadata files

jskeet:
(Some of the categorization may be a little arbitrary, but I'm asking for clarification internally.)

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -342,6 +342,7 @@
       "version": "2.1.1",
       "commonResourcesConfig": "apis/Google.Cloud.Bigtable.Common.V2/CommonResourcesConfig.json",
       "type": "grpc",
+      "metadataType": "GAPIC_COMBO",
       "description": "Recommended Google client library to access the Google Cloud Bigtable API.",
       "tags": [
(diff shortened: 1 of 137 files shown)


apis/apis.json
```diff
@@ -901,6 +905,7 @@
       "listingDescription": "Firestore high-level library",
       "version": "2.3.1",
       "type": "other",
+      "metadataType": "VENEER_GRPC",
```
amanda-tarafa, line 908: If I remember correctly this was going to be replaced by GAPIC_COMBO.

amanda-tarafa (approved):
LGTM

apis/apis.json
jskeet: It's unclear - because for something like Firestore, it's *not* a GAPIC-like surface. I've asked for more clarity on this in the doc.

I'll merge this first and then update the ResourceManager PR to include the metadata type for that as well.

jskeet: No need to update ResourceManager v3 - we don't create a repo metadata file for that yet, as there's no release.

PR #6330, opened 2021-04-09
Remove legacy empty resource name files

jskeet:
These were generated by the monolithic generator, and we hadn't spotted them because they're empty. They're not doing any harm, but it would be better not to have them.

(There's still one ResourceNames.cs file in IAM, but that's non-empty.)


amanda-tarafa (approved):
LGTM

PR #6331, opened 2021-04-12
Use .g.cs extension for all generated files

jskeet:
(There will be some that we haven't picked up yet around conformance tests etc, but these are the most important ones anyay.)

If we can make it trivial to identify generated files, that should make it easier to migrate to Owl Bot.

(Definitely review the first commit first...)


amanda-tarafa (approved):
LGTM

PR #6333, opened 2021-04-13
Generate Google.Cloud.Security.PrivateCA.V1

jskeet:
This is not ready for release yet, but I wanted to check it generated.

The change to Google.Protobuf version is due to PrivateCA v1 using optional fields, which are no longer experimental.
We'll want to make a corresponding update in GAX.

I've regenerated all APIs with the new generator, and there are no changes.


amanda-tarafa (approved):
LGTM

PR #6342, opened 2021-04-14
Fix service directory generation around "import public"

jskeet:


jskeet:
This will fix the generation of #6341 - we'll just rerun autosynth.

amanda-tarafa (approved):
LGTM

PR #6343, opened 2021-04-14
More changes to use .g.cs for proto generators

jskeet:
Basically we should use .g.cs everywhere we use protoc. I think with this change in place, we're most of the way there.


amanda-tarafa (approved):
LGTM

PR #6355, opened 2021-04-19
REGAPIC and Compute support

jskeet:
This adds support for "regapic" as both a generator and package type, and then moves the Compute.V1 API from the regapic branch of gax-dotnet.


jskeet:
cc @viacheslav-rostovtsev

apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1.IntegrationTests/AcceptanceTest.cs
```diff
@@ -0,0 +1,125 @@
...
+    public sealed class AcceptanceTest
+    {
+        private readonly ITestOutputHelper _output;
+        private readonly ComputeFixture _fixture;
+
+        public AcceptanceTest(ComputeFixture fixture, ITestOutputHelper output) =>
+            (_fixture, _output) = (fixture, output);
+
+        [Fact]
+        public void CreateDeleteIPAddress()
```
amanda-tarafa, line 36: I would write individual tests with proper resuource initialization and cleanup after.
We wouldn't need to write so much to the output then.
Basically, this is a big departure from how we write integration tests in this repo and I would prefrer if we stick to one way of writing tests. I think these were going to be worked on more, so maybe something to take into account.

apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1.IntegrationTests/AcceptanceTest.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                _output.WriteLine($"Deleting address with the name: {addressName}");
+                Operation deleteOp = addressesClient.Delete(projectId, region, addressName);
+                _output.WriteLine($"Operation to delete address: {deleteOp.Name} status {deleteOp.Status}; start time {deleteOp.StartTime}");
+                deleteOp = PollForCompletion(deleteOp, "delete");
+                _output.WriteLine($"Operation to delete address completed: status {deleteOp.Status}; start time {deleteOp.StartTime}; end time {deleteOp.EndTime}");
+            }
+        }
+
+        private Operation PollForCompletion(Operation operation, string alias)
```
amanda-tarafa, line 90: Should this be a public extension method in the library, or is this something that will be generated better later?

amanda-tarafa (approved):
LGTM

apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1.IntegrationTests/AcceptanceTest.cs
jskeet: The latter, I believe. At least, that's what I'm hoping.

apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1.IntegrationTests/AcceptanceTest.cs
jskeet: We've got a *few* tests that are like this - particularly in Storage, for one access control feature. (Can't remember the name now.)

I did consider doing resource clean up in the disposal part - listing all IP addresses either specifying a filter (if there's such a thing as a "name starts with" filter) or deleting everything starting with "testaddr". Or possibly generating a per-test prefix. Thoughts on that?

amanda-tarafa: You'd probably need a per-test prefix, there's a chance that Windows and Ubuntu are running at the same time and one tramples on the other's IPs.

jskeet: Yup - it'll be part of the test fixture.

jskeet:
Merging, and will then create a follow-up PR for cleanup.

PR #6357, opened 2021-04-21
Simplify deleting generated files

jskeet:
Now that everything is generated as .g.cs, it's trivial to delete everything that's generated.

Tested by regenerating all APIs: no changes. (This would pick up cases where we'd named non-generated files as .g.cs.)


amanda-tarafa (approved):
LGTM

PR #6358, opened 2021-04-22
feat(spanner): add option to return read timestamp

skuruppu:
For snapshot read-only transactions, it's possible to get the backend to return the [`read_timestamp`](https://cloud.google.com/spanner/docs/reference/rpc/google.spanner.v1#google.spanner.v1.Transaction). This PR introduces a `TimestampBound.ReturnReadTimestamp` property so that users can opt into returning the timestamp.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
@@ -19,6 +19,7 @@ using Google.Cloud.Spanner.V1.Internal.Logging;
 using System;
 using System.Threading.Tasks;
 using Xunit;
+using wkt = Google.Protobuf.WellKnownTypes;
 
 namespace Google.Cloud.Spanner.Data.IntegrationTests
 {
@@ -446,6 +447,98 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
                     if (await reader.ReadAsync())
                     {
                         Assert.Equal(_newestEntry.Value, reader.GetFieldValue<string>(reader.GetOrdinal("StringValue")));
+                    }
+                }
+            }
+        }
+
+        [Fact]
+        public async Task ExactReadSinglueUseWithoutReturnReadTimestamp()
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                await connection.OpenAsync();
+                var cmd = CreateSelectAllCommandForKey(connection);
+                var bound = TimestampBound.OfReadTimestamp(_oldestEntry.Timestamp);
+                using (var reader = (SpannerDataReader)(await cmd.ExecuteReaderAsync(bound)))
+                {
+                    Assert.Null(reader.GetReadTimestamp());
+                    if (await reader.ReadAsync())
+                    {
+                        Assert.Equal(_oldestEntry.Value, reader.GetFieldValue<string>(reader.GetOrdinal("StringValue")));
+                    }
+                }
+            }
+        }
+
+        [Fact]
+        public async Task ExactReadSinglueUseWithReturnReadTimestamp()
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                await connection.OpenAsync();
+                var cmd = CreateSelectAllCommandForKey(connection);
+                var bound = TimestampBound.OfReadTimestamp(_oldestEntry.Timestamp).WithReturnReadTimestamp(true);
+                using (var reader = (SpannerDataReader)(await cmd.ExecuteReaderAsync(bound)))
+                {
+                    Assert.Equal(wkt::Timestamp.FromDateTime(_oldestEntry.Timestamp), reader.GetReadTimestamp());
+                    if (await reader.ReadAsync())
+                    {
+                        Assert.Equal(_oldestEntry.Value, reader.GetFieldValue<string>(reader.GetOrdinal("StringValue")));
+                    }
+                }
+            }
+        }
+
+        [Fact]
+        public async Task ExactReadMultiUseWithoutReturnReadTimestamp()
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                await connection.OpenAsync();
+                var targetReadTimestamp = _fixture.TimestampBeforeEntries;
+                var bound = TimestampBound.OfReadTimestamp(targetReadTimestamp);
+                using (var tx = await connection.BeginReadOnlyTransactionAsync(bound))
+                {
+                    Assert.Equal(TransactionMode.ReadOnly, tx.Mode);
+                    Assert.Equal(targetReadTimestamp, tx.TimestampBound.Timestamp);
+                    Assert.False(tx.TimestampBound.ReturnReadTimestamp);
+
+                    var cmd = CreateSelectAllCommandForKey(connection);
+                    cmd.Transaction = tx;
+                    using (var reader = await cmd.ExecuteReaderAsync())
+                    {
+                        Assert.Null(reader.GetReadTimestamp());
+                        Assert.False(
+                            await reader.ReadAsync(),
+                            "no data should be here from yesterday!");
+                    }
+                }
+            }
+        }
+
+        [Fact]
+        public async Task ExactReadMultiUseWithReturnReadTimestamp()
+        {
+            using (var connection = _fixture.GetConnection())
+            {
+                await connection.OpenAsync();
+                var targetReadTimestamp = _fixture.TimestampBeforeEntries;
+                var bound = TimestampBound.OfReadTimestamp(targetReadTimestamp).WithReturnReadTimestamp(true);
+                using (var tx = await connection.BeginReadOnlyTransactionAsync(bound))
+                {
+                    Assert.Equal(TransactionMode.ReadOnly, tx.Mode);
+                    Assert.Equal(targetReadTimestamp, tx.TimestampBound.Timestamp);
+                    Assert.True(tx.TimestampBound.ReturnReadTimestamp);
+
+                    var cmd = CreateSelectAllCommandForKey(connection);
+                    cmd.Transaction = tx;
+                    using (var reader = await cmd.ExecuteReaderAsync())
+                    {
+                        Assert.Equal(wkt::Timestamp.FromDateTime(targetReadTimestamp), reader.GetReadTimestamp());
+                        Assert.False(
+                            await reader.ReadAsync(),
+                            "no data should be here from yesterday!");
                     }
                 }
             }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
@@ -45,6 +45,7 @@ namespace Google.Cloud.Spanner.Data
         private readonly ReliableStreamReader _resultSet;
         private ConcurrentDictionary<string, int> _fieldIndex;
         private ResultSetMetadata _metadata;
+        private Timestamp _readTimestamp;
         private readonly IDisposable _resourceToClose;
         private readonly SpannerConversionOptions _conversionOptions;
         private readonly bool _provideSchemaTable;
@@ -57,6 +58,7 @@ namespace Google.Cloud.Spanner.Data
         internal SpannerDataReader(
             Logger logger,
             ReliableStreamReader resultSet,
+            Timestamp readTimestamp,
             IDisposable resourceToClose,
             SpannerConversionOptions conversionOptions,
             bool provideSchemaTable,
@@ -68,6 +70,7 @@ namespace Google.Cloud.Spanner.Data
                 "SpannerDataReader.ActiveCount",
                 () => Interlocked.Increment(ref s_readerCount));
             _resultSet = resultSet;
+            _readTimestamp = readTimestamp;
             _resourceToClose = resourceToClose;
             _conversionOptions = conversionOptions;
             _provideSchemaTable = provideSchemaTable;
@@ -252,6 +255,49 @@ namespace Google.Cloud.Spanner.Data
                 InitializeFieldIndexFromMetadata();
             }
             return _fieldIndex[name];
+        }
+
+        /// <summary>
+        /// Gets the read timestamp if <see cref="TimestampBound.ReturnReadTimestamp" /> was true
+        /// else returns <c>null</c>.
+        /// </summary>
+        /// <remarks>
+        /// The read timestamp can be read before <see cref="SpannerDataReader.Read"/> or
+        /// <see cref="SpannerDataReader.ReadAsync"/> is called.
+        /// </remarks>
+        /// <remarks>
+        /// The read timestamp is available even if no data is returned.
+        /// </remarks>
+        /// <returns>The value converted to a <see cref="Timestamp"/>, or <c>null</c> if
+        /// <see cref="TimestampBound.ReturnReadTimestamp" /> was false.</returns>
+        public Timestamp GetReadTimestamp()
+        {
+            if (_readTimestamp != null)
+            {
+                return _readTimestamp;
+            }
+            return Task.Run(() => GetReadTimestampAsync(CancellationToken.None)).ResultWithUnwrappedExceptions();
+        }
+
+        /// <summary>
+        /// Gets the read timestamp if <see cref="TimestampBound.ReturnReadTimestamp" /> was true
+        /// else returns <c>null</c>.
+        /// </summary>
+        /// <returns>The value converted to a <see cref="Timestamp"/>, or <c>null</c> if
+        /// <see cref="TimestampBound.ReturnReadTimestamp" /> was false.</returns>
+        public Task<Timestamp> GetReadTimestampAsync(CancellationToken cancellationToken)
+        {
+            if (_readTimestamp != null)
+            {
+                return Task.FromResult(_readTimestamp);
+            }
+            return ExecuteHelper.WithErrorTranslationAndProfiling(
+                async () =>
+                {
+                    var metadata = await PopulateMetadataAsync(cancellationToken).ConfigureAwait(false);
+                    _readTimestamp = metadata.Transaction?.ReadTimestamp;
+		    return _readTimestamp;
+                }, "SpannerDataReader.GetReadTimestamp", Logger);
         }
 
         // Initializes _fieldIndex, but should only be called after PopulateMetadata or PopulateMetadataAsync
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
@@ -417,6 +417,12 @@ namespace Google.Cloud.Spanner.Data
             _session.TransactionId.ToBase64(),
             TimestampBound);
 
+        /// <summary>
+        /// The read timestamp of the read-only transaction if
+        /// <see cref="TimestampBound.ReturnReadTimestamp" /> is true, else <c>null</c>.
+        /// </summary>
+        public Timestamp ReadTimestamp => _session.ReadTimestamp;
+
         /// <inheritdoc />
         protected override void Dispose(bool disposing)
         {
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
@@ -28,26 +28,22 @@ namespace Google.Cloud.Spanner.Data
     /// </summary>
     public sealed class TimestampBound : IEquatable<TimestampBound>
     {
-        private TimestampBound(TimestampBoundMode mode)
-        {
-            Mode = mode;
-        }
+        private static readonly DateTime s_utcMinValue = new DateTime(1, 1, 1, 0, 0, 0, DateTimeKind.Utc);
 
-        private TimestampBound(TimestampBoundMode mode, TimeSpan staleness)
+        private TimestampBound(TimestampBoundMode mode, TimeSpan staleness,
+                               DateTime timestamp, bool returnReadTimestamp = false)
         {
             if (staleness.Ticks < 0)
             {
                 throw new ArgumentOutOfRangeException(nameof(staleness), "Staleness must not be negative");
             }
+
+            GaxPreconditions.CheckArgument(timestamp.Kind == DateTimeKind.Utc, nameof(timestamp), "Timestamps must be expressed in UTC");
+
             Mode = mode;
             Staleness = staleness;
-        }
-
-        private TimestampBound(TimestampBoundMode mode, DateTime timestamp)
-        {
-            GaxPreconditions.CheckArgument(timestamp.Kind == DateTimeKind.Utc, nameof(timestamp), "Timestamps must be expressed in UTC");
-            Mode = mode;
             Timestamp = timestamp;
+            ReturnReadTimestamp = returnReadTimestamp;
         }
 
         /// <summary>
@@ -75,10 +71,22 @@ namespace Google.Cloud.Spanner.Data
         public DateTime Timestamp { get; }
 
         /// <summary>
+        /// If true, the read timestamp is included in the
+        /// <see cref="Transaction" /> message that describes the transaction.
+        /// <remarks>
+        /// The read timestamp is Cloud Spanner-selected unless it's an exact
+        /// read, in which case it will be set to the exact read timestamp set
+        /// in the request.
+        /// </remarks>
+        /// </summary>
+        public bool ReturnReadTimestamp { get; }
+
+        /// <summary>
         /// Read at a timestamp where all previously committed transactions
         /// are visible.
         /// </summary>
-        public static TimestampBound Strong { get; } = new TimestampBound(TimestampBoundMode.Strong);
+        public static TimestampBound Strong { get; } =
+            new TimestampBound(TimestampBoundMode.Strong, TimeSpan.Zero, s_utcMinValue);
 
         /// <summary>
         /// Executes all reads at a timestamp that is <paramref name="duration"/>
@@ -96,7 +104,7 @@ namespace Google.Cloud.Spanner.Data
         /// <param name="duration">The exact staleness to use. Must not be negative.</param>
         /// <returns>A created <see cref="TimestampBound"/>.</returns>
         public static TimestampBound OfExactStaleness(TimeSpan duration) =>
-            new TimestampBound(TimestampBoundMode.ExactStaleness, duration);
+            new TimestampBound(TimestampBoundMode.ExactStaleness, duration, s_utcMinValue);
 
         /// <summary>
         /// Read data at a timestamp >= `NOW - <paramref name="duration"/>`. Guarantees that all
@@ -116,7 +124,7 @@ namespace Google.Cloud.Spanner.Data
         /// <param name="duration">The maximum duration of staleness to use. Must not be negative.</param>
         /// <returns>A created <see cref="TimestampBound"/>.</returns>
         public static TimestampBound OfMaxStaleness(TimeSpan duration) =>
-            new TimestampBound(TimestampBoundMode.MaxStaleness, duration);
+            new TimestampBound(TimestampBoundMode.MaxStaleness, duration, s_utcMinValue);
 
         /// <summary>
         /// Executes all reads at a timestamp >= <paramref name="minReadTimestamp"/>.
@@ -130,7 +138,7 @@ namespace Google.Cloud.Spanner.Data
         /// <param name="minReadTimestamp">The earliest timestamp to read from. Must be in UTC.</param>
         /// <returns>A created <see cref="TimestampBound"/>.</returns>
         public static TimestampBound OfMinReadTimestamp(DateTime minReadTimestamp) =>
-            new TimestampBound(TimestampBoundMode.MinReadTimestamp, minReadTimestamp);
+            new TimestampBound(TimestampBoundMode.MinReadTimestamp, TimeSpan.Zero, minReadTimestamp);
 
         /// <summary>
         /// Executes all reads at the given timestamp. Unlike other modes,
@@ -146,7 +154,22 @@ namespace Google.Cloud.Spanner.Data
         /// <param name="timestamp">The timestamp to read from. Must be in UTC.</param>
         /// <returns>A created <see cref="TimestampBound"/>.</returns>
         public static TimestampBound OfReadTimestamp(DateTime timestamp) =>
-            new TimestampBound(TimestampBoundMode.ReadTimestamp, timestamp);
+            new TimestampBound(TimestampBoundMode.ReadTimestamp, TimeSpan.Zero, timestamp);
+
+        /// <summary>
+        /// Creates a <see cref="TimestampBound"/> with the given value on
+        /// whether to return a read timestamp.
+        /// </summary>
+        /// <param name="returnReadTimestamp">Whether to return a read timestamp.</param>
+        /// <returns>A created <see cref="TimestampBound"/>.</returns>
+        public TimestampBound WithReturnReadTimestamp(bool returnReadTimestamp)
+        {
+            if (this.ReturnReadTimestamp == returnReadTimestamp)
+            {
+                return this;
+            }
+            return new TimestampBound(Mode, Staleness, Timestamp, returnReadTimestamp);
+        }
 
         /// <inheritdoc />
         public bool Equals(TimestampBound other)
@@ -159,7 +182,7 @@ namespace Google.Cloud.Spanner.Data
             {
                 return true;
             }
-            return Mode == other.Mode && Staleness.Equals(other.Staleness) && Timestamp.Equals(other.Timestamp);
+            return Mode == other.Mode && Staleness.Equals(other.Staleness) && Timestamp.Equals(other.Timestamp) && ReturnReadTimestamp == other.ReturnReadTimestamp;
         }
 
         /// <inheritdoc />
@@ -173,6 +196,7 @@ namespace Google.Cloud.Spanner.Data
                 var hashCode = (int) Mode;
                 hashCode = (hashCode * 397) ^ Staleness.GetHashCode();
                 hashCode = (hashCode * 397) ^ Timestamp.GetHashCode();
+                hashCode = (hashCode * 397) ^ ReturnReadTimestamp.GetHashCode();
                 return hashCode;
             }
         }
@@ -190,7 +214,8 @@ namespace Google.Cloud.Spanner.Data
                 {
                     [nameof(Mode)] = Value.ForString(Mode.ToString()),
                     [nameof(Timestamp)] = Value.ForString(Timestamp.ToString("O")),
-                    [nameof(Staleness)] = Value.ForString(Staleness.Ticks.ToString(CultureInfo.InvariantCulture))
+                    [nameof(Staleness)] = Value.ForString(Staleness.Ticks.ToString(CultureInfo.InvariantCulture)),
+                    [nameof(ReturnReadTimestamp)] = Value.ForBool(ReturnReadTimestamp)
                 }
             };
             return structValue.ToByteString().ToBase64();
@@ -207,6 +232,7 @@ namespace Google.Cloud.Spanner.Data
             TimestampBoundMode? mode = null;
             DateTime? timestamp = null;
             TimeSpan? staleness = null;
+            bool returnReadTimestamp = false;
 
             if (structValue.Fields.TryGetValue(nameof(Mode), out Value timestampBoundModeValue))
             {
@@ -224,19 +250,22 @@ namespace Google.Cloud.Spanner.Data
             {
                 throw new InvalidOperationException($"Unable to properly deserialize {nameof(TimestampBound)}.{nameof(Mode)}");
             }
+            if (structValue.Fields.TryGetValue(nameof(ReturnReadTimestamp), out Value returnReadTimestampValue))
+            {
+                returnReadTimestamp = returnReadTimestampValue.BoolValue;
+            }
             switch (mode.Value)
             {
                 case TimestampBoundMode.ExactStaleness:
                 case TimestampBoundMode.MaxStaleness:
                     GaxPreconditions.CheckArgument(staleness.HasValue, nameof(base64String), "Unable to properly deserialize {0}.{1}", nameof(TimestampBound), nameof(Staleness));
-                    return new TimestampBound(mode.Value, staleness.Value);
+                    return new TimestampBound(mode.Value, staleness.Value, s_utcMinValue, returnReadTimestamp);
                 case TimestampBoundMode.MinReadTimestamp:
                 case TimestampBoundMode.ReadTimestamp:
                     GaxPreconditions.CheckArgument(timestamp.HasValue, nameof(base64String), "Unable to properly deserialize {0}.{1}", nameof(TimestampBound), nameof(Timestamp));
-                    return new TimestampBound(mode.Value, timestamp.Value);
+                    return new TimestampBound(mode.Value, TimeSpan.Zero, timestamp.Value, returnReadTimestamp);
                 case TimestampBoundMode.Strong:
-                    // No need to create a new instance
-                    return Strong;
+                    return Strong.WithReturnReadTimestamp(returnReadTimestamp);
                 default:
                     throw new ArgumentException($"Invalid mode for timestamp bound: {mode.Value}");
             }
@@ -266,6 +295,7 @@ namespace Google.Cloud.Spanner.Data
                 default:
                     throw new ArgumentOutOfRangeException();
             }
+            innerOptions.ReturnReadTimestamp = ReturnReadTimestamp;
             return new TransactionOptions { ReadOnly = innerOptions };
         }
     }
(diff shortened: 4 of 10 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
```diff
@@ -475,6 +476,33 @@ public async Task ReadStrongSingle()
...
+                    await connection.BeginReadOnlyTransactionAsync(bound))
+                {
+                    Assert.Equal(TransactionMode.ReadOnly, tx.Mode);
+                    Assert.Equal(targetReadTimestamp, tx.TimestampBound.Timestamp);
+
+                    var cmd = CreateSelectAllCommandForKey(connection);
+                    cmd.Transaction = tx;
+                    using (var reader = await cmd.ExecuteReaderAsync())
+                    {
+                        Assert.Equal(reader.GetReadTimestamp(), wkt::Timestamp.FromDateTime(targetReadTimestamp));
```
skuruppu, line 499: Note that this line throws a null reference error because `PopulateMetadata().Transaction` doesn't exist. It might be because there's no data being returned by this reader or for some other reason. I need to debug.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
```diff
@@ -74,6 +74,12 @@ private TimestampBound(TimestampBoundMode mode, DateTime timestamp)
...
 
+        /// <summary>
+        /// If true, the Cloud Spanner-selected read timestamp is included in the
+        /// <see cref="Transaction" /> message that describes the transaction.
+        /// </summary>
+        public bool ReturnReadTimestamp { get; set; }
+
         /// <summary>
         /// Read at a timestamp where all previously committed transactions
         /// are visible.
```
skuruppu, line 85: `Strong` is a singleton so if we set the `ReturnReadTimestamp` property once, then all strong reads will return this value. But changing it to `Strong()` will be a breaking change. What's the best way to fix this?

jskeet: TimestampBound is intended to be immutable. We definitely don't want to introduce a setter here.

Instead, I would suggest adding a `WithReturnReadTimestamp(bool)` method that returns a new instance with the given value. (I'd also suggest probably refactoring to have a single constructor that accepts all the property values.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
jskeet: Happy to help debug if you'd like me to. Just give me a shout.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
```diff
@@ -224,22 +233,32 @@ public static TimestampBound FromBase64String(string base64String)
             {
                 throw new InvalidOperationException($"Unable to properly deserialize {nameof(TimestampBound)}.{nameof(Mode)}");
             }
+            if (structValue.Fields.TryGetValue(nameof(ReturnReadTimestamp), out Value returnReadTimestampValue))
+            {
+                returnReadTimestamp = Boolean.Parse(returnReadTimestampValue.StringValue);
+            }
+            TimestampBound timestampBound;
```
jskeet, line 240: All of this will be simplified back to return statements if we make this immutable again.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
olavloite: You need to call `reader.Read()` or `reader.ReadAsync()` at least once before any (meta)data is returned.

skuruppu: Ah that old problem! Nice catch, thanks @olavloite :)

Turns out the timestamp is also unavailable if your read doesn't return any results so I had to change the test a bit.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
skuruppu: Good idea. Done, including the refactoring.

The downside of the refactoring is that you can't have non-static default values (for TimeSpan and DateTime) so I had to pass these in explicitly. A bit tedious but not a huge issue given that it's all private.

skuruppu:
@jskeet and @olavloite this is ready for another look.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
jskeet: Eek - is that "also unavailable" part of this implementation somewhere lower level, or is it general Spanner-wide? It's a surprise to me.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +254,24 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+        /// Gets the read timstamp if <see cref="TimestampBound.ReturnReadTimestamp" /> was specified.
+        /// </summary>
+        /// <remarks>
+        /// The read timestamp is only available after at least one result is
+        /// read. If the returned results are empty, then a read timestamp
+        /// won't be available.
+        /// </remarks>
+        /// <returns>The value converted to a <see cref="Timestamp"/>.</returns>
+        /// <exception cref="InvalidOperationException">A read timestamp is unavailable,
+        /// possibly becuase <see cref="TimestampBound.ReturnReadTimestamp" /> was not set.</exception>
```
jskeet, line 267: becuase => because

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +254,24 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
             return _fieldIndex[name];
         }
 
+        /// <summary>
+        /// Gets the read timstamp if <see cref="TimestampBound.ReturnReadTimestamp" /> was specified.
```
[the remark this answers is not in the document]
jskeet: And possibly "was true"? (It can be specified as false, after all.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
jskeet, line 267: Again, possibly "was not set" => "was false"

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +254,24 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+        /// The read timestamp is only available after at least one result is
+        /// read. If the returned results are empty, then a read timestamp
+        /// won't be available.
+        /// </remarks>
+        /// <returns>The value converted to a <see cref="Timestamp"/>.</returns>
+        /// <exception cref="InvalidOperationException">A read timestamp is unavailable,
+        /// possibly becuase <see cref="TimestampBound.ReturnReadTimestamp" /> was not set.</exception>
+        public Timestamp GetReadTimestamp()
+        {
+            GaxPreconditions.CheckState(PopulateMetadata().Transaction != null, "Transaction unavailable so read timestamp can't be returned.");
```
jskeet, line 270: Can we call PopulateMetadata once and keep the result in a local variable? (I know it's cached, but it feels cleaner to me.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +254,24 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+        /// </summary>
+        /// <remarks>
+        /// The read timestamp is only available after at least one result is
+        /// read. If the returned results are empty, then a read timestamp
+        /// won't be available.
+        /// </remarks>
+        /// <returns>The value converted to a <see cref="Timestamp"/>.</returns>
+        /// <exception cref="InvalidOperationException">A read timestamp is unavailable,
+        /// possibly becuase <see cref="TimestampBound.ReturnReadTimestamp" /> was not set.</exception>
+        public Timestamp GetReadTimestamp()
```
jskeet, line 268: Given that `PopulateMetadata()` runs `PopulateMetadataAsync()` if it's not already set, perhaps we need an async version of this call as well? We could do that later, of course.

jskeet: Thinking about this more, if the timestamp is only actually provided when you call Read or ReadAsync, maybe we shouldn't call PopulateMetadata at all, and just use the existing cached version? Under what circumstances would PopulateMetadata provide something here without it having been cached before? (I may have misunderstood something.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +254,24 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+        {
+            GaxPreconditions.CheckState(PopulateMetadata().Transaction != null, "Transaction unavailable so read timestamp can't be returned.");
+            GaxPreconditions.CheckState(PopulateMetadata().Transaction.ReadTimestamp != null, "ReadTimestamp is not returned in the transaction.");
```
jskeet, line 271: I'm not sure about this. If the timestamp can be unavailable because there was no data, I wonder whether returning a null timestamp would be better if this transaction just doesn't have a timestamp. Otherwise I can see it being tricky for users to know whether or not they can call this - I don't want every user to have to put this in a try/catch statement.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
```diff
@@ -28,26 +28,26 @@ namespace Google.Cloud.Spanner.Data
...
-        }
+        // Determines whether a read-only transaction will return the read timestamp.
+        private bool _returnReadTimestamp = false;
 
-        private TimestampBound(TimestampBoundMode mode, TimeSpan staleness)
+        private TimestampBound(TimestampBoundMode mode, TimeSpan staleness,
+                               DateTime timestamp, bool returnReadTimestamp = false)
         {
-            if (staleness.Ticks < 0)
+            if (staleness != TimeSpan.Zero && staleness.Ticks < 0)
```
jskeet, line 37: No need for the extra check here - TimeSpan.Zero.Ticks is 0 anyway, so it won't throw.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
```diff
@@ -28,26 +28,26 @@ namespace Google.Cloud.Spanner.Data
...
                 throw new ArgumentOutOfRangeException(nameof(staleness), "Staleness must not be negative");
             }
-            Mode = mode;
-            Staleness = staleness;
-        }
 
-        private TimestampBound(TimestampBoundMode mode, DateTime timestamp)
-        {
-            GaxPreconditions.CheckArgument(timestamp.Kind == DateTimeKind.Utc, nameof(timestamp), "Timestamps must be expressed in UTC");
+            if (timestamp != DateTime.MinValue)
```
jskeet, line 42: I think it would be simpler to have a min-value-in-utc private static variable:

```csharp
private static readonly s_utcMinValue = new DateTime(1, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```

Then use that everywhere we're otherwise using `DateTime.MinValue`. That way the timestamp will *genuinely* always be in UTC, which is nicer IMO.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
```diff
@@ -74,11 +74,18 @@ private TimestampBound(TimestampBoundMode mode, DateTime timestamp)
         /// </summary>
         public DateTime Timestamp { get; }
 
+        /// <summary>
+        /// If true, the Cloud Spanner-selected read timestamp is included in the
+        /// <see cref="Transaction" /> message that describes the transaction.
+        /// </summary>
+        public bool ReturnReadTimestamp { get => _returnReadTimestamp; }
```
jskeet, line 81: This can just be `{ get; }` and remove the separately-declared field.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
```diff
@@ -146,7 +153,18 @@ private TimestampBound(TimestampBoundMode mode, DateTime timestamp)
...
-            new TimestampBound(TimestampBoundMode.ReadTimestamp, timestamp);
+            new TimestampBound(TimestampBoundMode.ReadTimestamp, TimeSpan.Zero, timestamp);
+
+        /// <summary>
+        /// Creates a <see cref="TimestampBound"/> with the given value on
+        /// whether to return a read timestamp.
+        /// </summary>
+        /// <param name="returnReadTimestamp">Whether to return a read timestamp.</param>
+        /// <returns>A created <see cref="TimestampBound"/>.</returns>
+        public TimestampBound WithReturnReadTimestamp(bool returnReadTimestamp)
```
jskeet, line 164: Nit: use an expression bodied member:

```csharp
public TimestampBound WithReturnReadTimestamp(bool returnReadTimestamp) =>
    new TimestampBound(Mode, Staleness, Timestamp, returnReadTimestamp);
````

**Review by jskeet — COMMENTED** [review]
Plenty more comments I'm afraid - but much happy with the TimestampBound part. I think we'll be able to get there soon.

File: apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs (hunk 1 above)
Comment (olavloite) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs]: > Turns out the timestamp is also unavailable if your read doesn't return any results so I had to change the test a bit.

No, that should not be the case. Cloud Spanner always returns a read timestamp if your read-only transaction requests one, regardless whether the query returns any rows or not. So I think that this is a problem in the C# client in that case. If you execute a query on Cloud Spanner that requests a timestamp and the query does not return any results, Cloud Spanner will always return a `PartialResultSet` containing the metadata of the query. That metadata will include the columns in the query (also when there are no results), and the transaction metadata including the read timestamp.
I think the problem might be that the metadata are not read at all by the client if there are no rows in the result set.

(Just tested the above with the Java client to be absolutely sure, and I can verify that Cloud Spanner returns the read timestamp also when the query returns 0 rows)

Comment (jskeet) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs]: That's encouraging - I'm far happier for this to be a problem in the C# code, as we can fix that :) @skuruppu let me know if you'd like me to take a look diagnostically; otherwise, I'm happy to let you figure it out :)

File: apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs (hunk 7 above)
Comment (skuruppu) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs:268]: I think the reason we're calling it is because maybe the metadata is not already populated. It seems that all of the code that relies on a metadata field calls this method as a precaution just in case no one else has called it before (the value is cached after the first call so we're using a cached value if it's already been called before).

Comment (skuruppu) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs:268]: I added an async version.

File: apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs (hunk 8 above)
Comment (skuruppu) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs:271]: Since it may be a bug in the client that we don't return a timestamp if there's no data (I still need to investigate and fix it), do you think we should still return null for the case where Read() or ReadAsync() has not yet been called?

File: apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs (hunk 1 above)
Comment (skuruppu) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs]: @olavloite thanks for checking this. I'll do a dig around the code to see why we're not getting the metadata.

Comment (skuruppu) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs]: I realize this had nothing to do with whether data is returned. It's because of the type of transaction.

If you try to do an exact read (with a BeginReadOnlyTransactionAsync call as [here](https://github.com/googleapis/google-cloud-dotnet/blob/13ed74b0e70f5b9932c9f679a3378493586c2fcc/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs#L302), then you can't pass the bound to `ExecuteReaderAsync()` ([link](https://github.com/googleapis/google-cloud-dotnet/blob/13ed74b0e70f5b9932c9f679a3378493586c2fcc/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs#L308)). Without the bound, the transaction doesn't know that you want to return the timestamp. If you do try to pass it into `ExecuteReaderAsync()`, the client returns https://github.com/googleapis/google-cloud-dotnet/blob/63f54b5ebb8c2f7eec26f0fd7fabfc77e1237596/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs#L122

I don't know if this is a case where a user might expect to be able to return the timestamp.

**Comment by skuruppu** [comment]
PTAL, I've addressed all the comments.

**Comment by jskeet** [comment]
I will have a look at this when I get the chance, but it may not be before Tuesday (due to vacation and a public holiday). Please shout if there's more urgency than I'm aware of.

File: apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs (hunk 1 above)
Comment (olavloite) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs]: (Maybe it's already working, but I noticed that there's no sample or test case for it in the current PR)

There are two ways that a user should be able to request and get a read timestamp:
1. Using a single-use read-only transaction. This is done by passing in a `TimestampBound` to the `SpannerCommand.ExecuteReaderAsync(..)` method.
2. Using a multi-use read-only transaction. This is done by passing in a `TimestampBound` to the `SpannerConnection.BeginReadOnlyTransaction(..)`. The first read operation on the transaction (i.e. the first call to `SpannerCommand.ExecuteReaderAsync()` with a `SpannerCommand` that uses the transaction) should return the read timestamp.

So it should not be necessary to pass in any `TimestampBound` to the `SpannerCommand.ExecuteReaderAsync()` when you are using a multi-use read-only transaction.

Comment (skuruppu) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs]: Ah thanks for the context. I added back the test that wasn't working, which is the multi-use case. I also added a debug print so when we get the test failure, we can see in the logs what is returned in the metadata. You can see it in the emulator test failure in the PR. I haven't yet tried to debug why this is.

Comment (olavloite) [reply apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs]: Sorry, what I meant was that with multi-use read-only transactions, the read timestamp is __not__ returned as part of the metadata of the ResultSet. Instead you have to get it from the transaction that is returned when `BeginTransaction` is called. That is hidden away somewhat in the session pool, which makes it a little harder to get to.
I took the liberty of adding a commit to this PR that shows how you could register and get it in the `DataReader`, but feel free to change the design of it.

(I also have a couple of worries regarding the current implementation of `GetReadTimestamp` on `DataReader`, but I'll add separate comments for those)

File: apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs [hunk 13]
```diff
@@ -254,6 +257,53 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+        /// <returns>The value converted to a <see cref="Timestamp"/>.</returns>
+        /// <exception cref="InvalidOperationException">A read timestamp is unavailable,
+        /// possibly because <see cref="TimestampBound.ReturnReadTimestamp" /> was false.</exception>
+        public Timestamp GetReadTimestamp()
+        {
+            if (_readTimestamp != null)
+            {
+                return _readTimestamp;
+            }
+            return Task.Run(() => GetReadTimestampAsync(CancellationToken.None)).ResultWithUnwrappedExceptions();
```
olavloite, line 277: I worry a little about this. If I understand this correctly, anyone who calls this method __before__ having called `Read` or `ReadAsync`, will block until some other thread calls `Read` or `ReadAsync`. I don't think that is a good idea, as it could cause cause a thread to be blocked indefinitely, which can be very hard to debug, if someone accidentally calls this without having called `Read` first. I would rather just document clearly that:
1. `GetReadTimestamp` will only return a timestamp __after__ `Read` or `ReadAsync` has been called at least once for single-use transactions.
2. For multi-use transactions it will return it directly, as it is known from the read-only transaction that has already been started.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
skuruppu: Thanks @olavloite. I think your design works well. The only thing I changed was for `PooledSession.WithTransaction()` to take a null read timestamp by default as it didn't make much sense for RW transactions to accept a timestamp. But let me know if I misunderstood something there.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
skuruppu: I ran the integration test for single-use transactions to test this and it was happy to not have called `Read` or `ReadAsync`. Before I added the async variant, we were calling `PopulateMetadata()` which was already internally calling `PopulateMetadataAsync()` so the behavior didn't change much by making this method async.

Let me know if I misunderstood something.

skuruppu: Interestingly, it even had the read timestamp without me calling `Read` or `ReadAsync` which makes me think that internally `ResultSet.GetMetadataAsync()` is somehow able to get the metadata anyway.

skuruppu:
@googlebot I consent.

skuruppu:
PTAL @olavloite @jskeet 

@jskeet I don't understand the "Detect differences" failure. `SessionPool.CreateDetachedSession()` has an additional parameter but with a default value so it shouldn't affect any existing call sites. What am I missing?

jskeet:
@skuruppu:

> I don't understand the "Detect differences" failure. SessionPool.CreateDetachedSession() has an additional parameter but with a default value so it shouldn't affect any existing call sites. What am I missing?

That's still a breaking *binary* change. Optional parameters provide *compile-time* compatibility (mostly - not completely, in fact) but it's still a breaking change for previously-compiled code, because the CLR would still look for the method without the extra parameter.

To add a parameter in a compatible way, you need to overload the method instead, making the method with the current signature call the method with the additional parameter. (I would then *not* make that parameter optional - let the overload do the defaulting, effectively.)

Will look at the rest of the PR later today.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
olavloite: Ah, yes, that seems to be the case for this client. The `ResultSet.GetMetaDataAsync()` method will call `ReliableStreamReader.GetMetaDataAsync()`, which again will call `ReliableStreamReader.EnsureInitializedAsync()`. The latter method automatically calls `MoveNextAsync()` (which is basically the same as `Read` or `ReadAsync`):
https://github.com/googleapis/google-cloud-dotnet/blob/13ed74b0e70f5b9932c9f679a3378493586c2fcc/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReliableStreamReader.cs#L170-L180

olavloite (approved):
LGTM (with a couple of small nits)

skuruppu:
@olavloite thanks for the approval. If you could also do a "@googlebot I consent." commit, I think the CLA check will pass.

@jskeet if you could take another look, that would be awesome.

olavloite:
@googlebot I consent.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -58,6 +59,11 @@ public sealed class PooledSession : IDisposable
         /// </summary>
         internal ModeOneofCase TransactionMode { get; }
 
+        /// <summary>
+        /// The read timestamp of the transaction. (Always empty iff TransactionMode != ReadOnly.)
```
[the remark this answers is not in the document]
jskeet: Okay, I've now fetched the code and I understand how it's used - I'd only been looking at the SpannerDataReader, not the SpannerTransaction. Resolving.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +257,50 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+            if (_readTimestamp != null)
+            {
+                return Task.FromResult(_readTimestamp);
+            }
+            return ExecuteHelper.WithErrorTranslationAndProfiling(
+                async () =>
+                {
+                    var metadata = await PopulateMetadataAsync(cancellationToken).ConfigureAwait(false);
+                    GaxPreconditions.CheckState(metadata.Transaction != null, "Transaction unavailable so read timestamp can't be returned.");
+                    GaxPreconditions.CheckState(metadata.Transaction.ReadTimestamp != null, "ReadTimestamp is not returned in the transaction.");
```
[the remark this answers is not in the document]
amanda-tarafa: Also, for consitency, we return null for the schema if it's not available instead of failing.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
```diff
@@ -475,6 +476,56 @@ public async Task ReadStrongSingle()
         }
         // [END spanner_test_strong_read_concurrent_updates]
 
+        // [START spanner_test_exact_read_single_use_with_return_read_timestamp]
```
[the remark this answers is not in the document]
amanda-tarafa: Hmm, I see the region tag is `spanner_test` so maybe they should be shown as tests?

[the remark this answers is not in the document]
amanda-tarafa: I'm not sure these are meant to be samples honestly. What are the region tags here being used for?

amanda-tarafa (commented):
I've left a few comments, and also it would be good to have a couple of tests where we don't specify the read timestamp we want but request it to be returned, just to check that path works.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
```diff
@@ -170,10 +171,30 @@ public Task<PooledSession> AcquireSessionAsync(DatabaseName databaseName, Transa
...
+        /// <remarks>
+        /// This is typically used for partitioned queries, where the same session is used across multiple machines, so should
+        /// not be reused by the pool.
+        /// </remarks>
+        /// <param name="sessionName">The name of the transaction. Must not be null.</param>
+        /// <param name="transactionId">The ID of the transaction. Must not be null.</param>
+        /// <param name="transactionMode">The mode of the transaction.</param>
+        /// <param name="readTimestamp">The read timestamp of the transaction.</param>
+        /// <returns>A <see cref="PooledSession"/> for the given session and transaction.</returns>
+        public PooledSession CreateDetachedSession(SessionName sessionName, ByteString transactionId, ModeOneofCase transactionMode, Timestamp readTimestamp = null)
```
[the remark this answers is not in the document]
jskeet: That would surprise me - the tie-breaking rules of C# mean that a call which doesn't require the compiler to supply default values for optional parameters is "better than" a call which does require it to. But whichever of us is right, it proves that it's a potential source of confusion :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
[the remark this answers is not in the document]
skuruppu: Oh I don't think the region tags are used for anything. These are not pulled into docs. I think we have them because someone started doing it this way initially and we just continued with it...

I can leave it as is for consistency and we can get rid of all them in one cleanup PR. Or I can remove it from here now. Let me know what you think might be better.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +257,50 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+        /// <returns>The value converted to a <see cref="Timestamp"/>.</returns>
+        /// <exception cref="InvalidOperationException">A read timestamp is unavailable,
+        /// possibly because <see cref="TimestampBound.ReturnReadTimestamp" /> was false.</exception>
+        public Task<Timestamp> GetReadTimestampAsync(CancellationToken cancellationToken)
+        {
+            if (_readTimestamp != null)
+            {
+                return Task.FromResult(_readTimestamp);
+            }
+            return ExecuteHelper.WithErrorTranslationAndProfiling(
```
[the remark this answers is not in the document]
skuruppu: Ah ok, I guess that's reasonable. I didn't do it before because this is the single-use transaction case so there was unlikely to be multiple calls to this. But maybe someone will do that and it's better to have it cached.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +257,50 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+                async () =>
+                {
+                    var metadata = await PopulateMetadataAsync(cancellationToken).ConfigureAwait(false);
+                    GaxPreconditions.CheckState(metadata.Transaction != null, "Transaction unavailable so read timestamp can't be returned.");
```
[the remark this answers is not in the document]
skuruppu: That's a good question. Previously, when I wasn't handling the multi-use transaction case correctly, this was null. Now that that's handled, maybe I can remove this check.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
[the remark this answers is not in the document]
skuruppu: I see, I think that's fair. I wasn't sure what users would expect. I guess either they have to handle an exception or a null, the user has to know that this value may not exist.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
```diff
@@ -190,7 +208,8 @@ public string ToBase64String()
                 {
                     [nameof(Mode)] = Value.ForString(Mode.ToString()),
                     [nameof(Timestamp)] = Value.ForString(Timestamp.ToString("O")),
-                    [nameof(Staleness)] = Value.ForString(Staleness.Ticks.ToString(CultureInfo.InvariantCulture))
+                    [nameof(Staleness)] = Value.ForString(Staleness.Ticks.ToString(CultureInfo.InvariantCulture)),
+                    [nameof(ReturnReadTimestamp)] = Value.ForString(ReturnReadTimestamp ? Boolean.TrueString : Boolean.FalseString)
```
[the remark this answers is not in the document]
skuruppu: Compiler says:
```
TimestampBound.cs(218,59): error CS0117: 'Value' does not contain a definition for 'ForBoolean' [/google-cloud-dotnet/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.csproj]
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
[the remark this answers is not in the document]
skuruppu: Hehe ok that's fair. I removed the default value.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -58,6 +59,11 @@ public sealed class PooledSession : IDisposable
...
+        /// </summary>
+        public Timestamp ReadTimestamp { get; }
```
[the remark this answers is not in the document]
skuruppu: Ah yes, that's meant to be null.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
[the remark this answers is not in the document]
amanda-tarafa: I'ts fine to leave these in and then do a PR removing them.
But I think we should remove them, there's the SnippetTracker bot which is pulling these into the devrel services sample DB which means someone could include these on docs.
I'll create an issue so that we don't forget. Thanks!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +257,48 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+        /// </summary>
+        /// <remarks>
+        /// The read timestamp can be read before <see cref="SpannerDataReader.Read"/> or
+        /// <see cref="SpannerDataReader.ReadAsync"/> is called.
+        /// </remarks>
+        /// <remarks>
+        /// The read timestamp is available even if no data is returned.
+        /// </remarks>
+        /// <returns>The value converted to a <see cref="Timestamp"/>.</returns>
+        /// <exception cref="InvalidOperationException">A read timestamp is unavailable,
```
amanda-tarafa, line 271: I think this is obsolete now.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +257,48 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
             return _fieldIndex[name];
         }
 
+        /// <summary>
+        /// Gets the read timestamp if <see cref="TimestampBound.ReturnReadTimestamp" /> was true.
```
amanda-tarafa, line 261:
```suggestion
        /// Gets the read timestamp if <see cref="TimestampBound.ReturnReadTimestamp" /> was true else returns null.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -254,6 +257,48 @@ public async Task<int> GetOrdinalAsync(string name, CancellationToken cancellati
...
+        {
+            if (_readTimestamp != null)
+            {
+                return _readTimestamp;
+            }
+            return Task.Run(() => GetReadTimestampAsync(CancellationToken.None)).ResultWithUnwrappedExceptions();
+        }
+
+        /// <summary>
+        /// Gets the read timestamp if <see cref="TimestampBound.ReturnReadTimestamp" /> was true.
```
amanda-tarafa, line 283:
```suggestion
        /// Gets the read timestamp if <see cref="TimestampBound.ReturnReadTimestamp" /> was true else returns null.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -417,6 +417,13 @@ public Task RollbackAsync(CancellationToken cancellationToken = default)
             _session.TransactionId.ToBase64(),
             TimestampBound);
 
+        /// <summary>
+        /// The read timestamp of the read-only transaction if
+        /// <see cref="TimestampBound.ReturnReadTimestamp" /> is true. Always <c>null</c>
```
amanda-tarafa, line 422:
```suggestion
        /// <see cref="TimestampBound.ReturnReadTimestamp" /> is true, else null. Always <c>null</c>
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -58,6 +59,12 @@ public sealed class PooledSession : IDisposable
         /// </summary>
         internal ModeOneofCase TransactionMode { get; }
 
+        /// <summary>
+        /// The read timestamp of the transaction. (Always <c>null</c> if and
```
amanda-tarafa, line 63: Is this really and `if and only if`? I think it may also be null if ReturnReadTimestamp was set to false.

amanda-tarafa (commented):
A few minor comments.
And also, for getting rid of the diff checker failure, I think you need to rebase your PR on master.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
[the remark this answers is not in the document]
skuruppu: Thanks, I removed the tags from this PR since Lalji is already working on removing the tags from the rest of the tests.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
skuruppu: You're totally right. I updated the text.

jskeet:
I'm on vacation today - I'll have a last look on Monday, or Amanda can merge if she's happy. (I only had a couple of tiny points last time, if I remember rightly.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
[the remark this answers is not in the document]
jskeet: My mistake - it's "ForBool", but that should be fine: https://github.com/protocolbuffers/protobuf/blob/master/csharp/src/Google.Protobuf/WellKnownTypes/ValuePartial.cs

I don't much mind whether change that or not, but it's worth deciding once and for all *now*, as changing it later would be a breaking change.

amanda-tarafa (approved):
LGTM.
If you could rebase all the commits into one before merging that'd be great, to keep the repo history clean.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/TimestampBound.cs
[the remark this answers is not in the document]
skuruppu: Ah works now. Thanks.

skuruppu:
Thanks for your patience in reviewing this everyone. When you're ready, please merge it in. @jskeet if you could also please cut a release after this is merged, that would be fantastic.

PR #6362, opened 2021-04-22
Simplify how we run smoke tests

jskeet:
We already have the TEST_PROJECT environment variable, so there's no
need to specify the project ID every time we want to run smoke tests.


amanda-tarafa (approved):
LGTM

PR #6364, opened 2021-04-23
Fix Compute documentation generation

jskeet:
(In general, we need to be careful when adding an ApiType value.)


amanda-tarafa (approved):
LGTM

PR #6374, opened 2021-04-27
Implement OwlBot for "simple" APIs

jskeet:
Simple = no pre/mid/post-processing, no extra resource configuration, GAPIC only

The OwlBot YAML files aren't really "ready" yet (need a different
Docker image) but they show the output

Running generateprojects.sh will regenerate .OwlBot.yaml and also
delete existing synth.py and synth.metadata files.


tools/owlbot-postprocess.sh
```diff
@@ -0,0 +1,31 @@
...
+# Note: we could do various parts of this globally from the top level,
+# but doing it on a per-API level means we don't need to worry about
+# overly long command lines etc
+
+cd owlbot-staging
+for api in *
+do
+  echo "Processing $api"  
+  # Remove existing generated code from target directory
+  rm -rf ../apis/*/*.g.cs
```
amanda-tarafa, line 14: Should this remove the code for just the current api?
Also, it's only removing the .g.cs, but what about the .cs only that you have to rename later?
There might be something I'm missing here though.

tools/owlbot-postprocess.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+set -e
+
+# Note: we could do various parts of this globally from the top level,
```
amanda-tarafa, line 5: Who calls this script?

amanda-tarafa (commented):
Looks good. Just a couple of comments.

tools/owlbot-postprocess.sh
jskeet: Eesh! Yes it should. That first `/*` should be `/$api`. Doh!

tools/owlbot-postprocess.sh
jskeet: OwlBot itself. Probably easiest to explain in a VC :)

jskeet:
Having talked about this with Jeff, we're *probably* going to move to a model where each API has its own .OwlBot.yaml file, just like synthtool. But it's very much TBD.

jskeet:
Closing this for now (but keeping the branch) to avoid it being a distraction. I believe the Owl Bot team have a bit of work to do in order to fully support mono repos.

PR #6382, opened 2021-04-28
Use CopySettingsForEmulator in all emulator clients

### LINKED ISSUE #5671: Emulator-based builders should use CopyCommonSettings
Currently with any of the APIs that support EmulatorDetection, if you configure the builder to use the emulator *and* you set a custom GrpcAdapter, the adapter property is ignored. Ditto for UserAgent, QuotaProject etc. Usually for an emulator most of these aren't important, but we should honour them anyway.

We'd want to copy the common settings, then set the emulator detection to None (and probably some other settings) to avoid infinite recursion.

jskeet:
Fixes #5671


amanda-tarafa (approved):
LGTM

PR #6424, opened 2021-05-04
fix: 'Transaction outcome unknown' errors should be retried

### LINKED ISSUE #6421: Spanner: Retry Transaction Outcome Unknown errors
Cloud Spanner can return the error `Transaction Outcome Unknown` if a transaction fails with an unknown transient reason. It is safe to retry the entire transaction when this error is returned, even though the error message indicates that the outcome is inconclusive. The specific error should therefore be added to the conditions for retrying a transaction. Currently, only `Aborted` errors cause an entire transaction to be retried.

cc @skuruppu

olavloite:
Transactions that return `Transaction outcome unknown` with error code `Unknown` may be retried automatically. These errors are returned by the backend if it cannot guarantee the correctness of a transaction, and a retry of the same transaction will probably succeed.

Do not merge: There's still some discussion going on whether this should automatically be retried or not.

Fixes #6421


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerExceptionTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+using Grpc.Core;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    public class SpannerExceptionTests
+    {
+        [Fact]
+        public void IsRetryable()
```
jskeet, line 23: I'd suggest this would be clearer using some Theory tests with InlineTheoryData - you probably need two tests, one taking an ErrorCode, a string and an expected result, and the other taking an ErrorCode, a StatusCode, a string and an expected result.

jskeet (commented):
I'm fine with the production code - I just think one of the tests could be simpler using a data-driven test.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerExceptionTests.cs
olavloite: I've updated the test to a data-driven test, and that does indeed look tidier.

Please hold off on merging this one, as there's some discussion going on whether we should really retry this.

jskeet: > Please hold off on merging this one, as there's some discussion going on whether we should really retry this.

Ack.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerExceptionTests.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Google.Cloud.Spanner.Data.Tests
+{
+    public class SpannerExceptionTests
+    {
+        [InlineData(ErrorCode.InvalidArgument, StatusCode.InvalidArgument, "Invalid argument", false)]
+        [InlineData(ErrorCode.Aborted, StatusCode.Aborted, "Transaction aborted", true)]
+        [InlineData(ErrorCode.Unknown, StatusCode.Unknown, "An unknown error occurred", false)]
+        [InlineData(ErrorCode.Unknown, StatusCode.Unknown, "Transaction outcome unknown", true)]
+        [Theory]
+        public void IsRetryableWithErrorCodeAndInnerException(ErrorCode errorCode, StatusCode statusCode, string message, bool isRetryable)
```
amanda-tarafa, line 27: Not a blocker, but I would prefer if there were three distinct tests for these, one of each way of building the exception, each receiving the appropiate data.

olavloite: I've changed it into three separate test cases.

jskeet (approved):
I'm now happy with this, but I understand we don't want to merge it until there's further confirmation.

amanda-tarafa (approved):
LGTM

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerExceptionTests.cs
jskeet: Any more news on whether we should actually retry?

olavloite: Yes and no... We should retry in most cases, but we're awaiting confirmation for exactly what cases we should not retry (or when we need to do additional checks to determine whether it is safe to retry)

olavloite:
After an internal discussion, the decision has been made not to automatically retry these errors.

PR #6443, opened 2021-05-07
Remove codecov reporting

jskeet:
We still want to run the tests with coverage enabled in one CI
build, but for the moment the results won't be reported to anywhere
public.

Temporarily, the simplest way of configuring a CI build to perform
coverage testing is still to detect the presence of a codecov token,
but that can be improved (e.g. to use an environment variable) later.


amanda-tarafa (approved):
LGTM

PR #6455, opened 2021-05-12
Remove AppVeyor build

jskeet:
It's been a long time since it's proved valuable, and it's almost never done in time for merging a PR.

(If you'd like me to create a GitHub action that runs the project generator to check that doesn't change anything, I could definitely do that. It's about the only thing that's still useful in AppVeyor.)


amanda-tarafa (approved):
LGTM and +100 to this.

I think the action with the project generator would be good as a failsafe for when we forget to regenarate the projects ourselves (it has happened to me with Diagnostics). But no rush at all.

jskeet:
(Will do the GH action next week, with any luck.)

PR #6462, opened 2021-05-13
Improvements to Compute integration test polling

jskeet:
This is in two commits as I'm slightly less convinced by the second one, and it's simpler to see the changes in stages too.


amanda-tarafa (approved):
LGTM

PR #6465, opened 2021-05-14
test(spanner): remove region tags from tests

### LINKED ISSUE #6459: Spanner: Remove region tags from tests.
Test code shouldn't be surfaced in docs, and with the region tags in place there's a chance that may happen.

laljikanjareeya:
Fixes #6459


amanda-tarafa (approved):
LGTM!
And the snippet bot confirms that none of these were being used.

PR #6474, opened 2021-05-18
First proper pass at link canonicalization

jskeet:
This is ludicrously complicated, but that's because the packages in
googleapis.dev just try to include everything that's needed... and
because some of our packages (e.g. Google.Apis and Google.Apis.Core)
aren't brilliant in terms of consistency.

There may still be corner cases that I'm missing. Ideally we'll run
a long-running test that checks the validity of these before doing
anything irreversible.

### DIFF
--- a/tools/Google.Cloud.Tools.GenerateCanonicalLinks.Tests/CanonicalizerTest.cs
+++ b/tools/Google.Cloud.Tools.GenerateCanonicalLinks.Tests/CanonicalizerTest.cs
@@ -1,0 +1,53 @@
+﻿// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Xunit;
+
+namespace Google.Cloud.Tools.GenerateCanonicalLinks.Tests
+{
+    public class CanonicalizerTest
+    {
+        [Theory]
+        [InlineData("Google.Cloud.CloudBuild.V1", "api/Google.Cloud.CloudBuild.V1.Artifacts.Types.ArtifactObjects.html", "Google.Cloud.CloudBuild.V1/latest/Google.Cloud.CloudBuild.V1.Artifacts.Types.ArtifactObjects")]
+        [InlineData("Google.Cloud.CloudBuild.V1", "history.html", "Google.Cloud.CloudBuild.V1/latest/history")]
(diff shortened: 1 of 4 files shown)


tools/Google.Cloud.Tools.GenerateCanonicalLinks.Tests/CanonicalizerTest.cs
```diff
@@ -0,0 +1,50 @@
...
+        [InlineData("Google.Apis", "api/Google.Apis.ETagAction.html", "Google.Apis/latest/Google.Apis.ETagAction")]
+        [InlineData("Google.Apis", "api/Google.Apis.Download.DownloadStatus.html", "Google.Apis/latest/Google.Apis.Download.DownloadStatus")]
+        [InlineData("Google.Cloud.Asset.V1", "api/Google.Cloud.Iam.V1.AuditConfigDelta.html", "Google.Cloud.Iam.V1/latest/Google.Cloud.Iam.V1.AuditConfigDelta")]
+        [InlineData("Google.Cloud.Logging.V2", "api/Google.Cloud.Logging.Type.HttpRequest.html", "Google.Cloud.Logging.Type/latest/Google.Cloud.Logging.Type.HttpRequest")]
+        [InlineData("Google.Cloud.Firestore", "api/Google.Cloud.Firestore.V1.Firestore.html", "Google.Cloud.Firestore.V1/latest/Google.Cloud.Firestore.V1.Firestore")]
+        [InlineData("Google.Cloud.Asset.V1", "api/Google.Cloud.OrgPolicy.V1.Policy.html", "Google.Cloud.OrgPolicy.V1/latest/Google.Cloud.OrgPolicy.V1.Policy")]
+        [InlineData("Google.Cloud.Asset.V1", "api/Google.Identity.AccessContextManager.Type.DeviceEncryptionStatus.html", "Google.Identity.AccessContextManager.Type/latest/Google.Identity.AccessContextManager.Type.DeviceEncryptionStatus")]
```
amanda-tarafa, line 41: Would it make sense to try with something Diagnostics? I'm thinking of two cases (which might already work):
- Something from Diagnostics.Common being used in Diagnostics.AspNetCore.
- And how we have one solution only for Diagnostics.AspNetCore[3] with pragma if for setting one namespace or the other. So maybe try the same type in both Diagnostics.AspNetCore and Diagnostics.AspNetCore3?

amanda-tarafa (commented):
LGTM. Just a suggestion for more tests cases maybe.

amanda-tarafa (approved):

tools/Google.Cloud.Tools.GenerateCanonicalLinks.Tests/CanonicalizerTest.cs
jskeet: Ooh, I'd definitely missed Diagnostics.Common, yes. Okay, will add tests for those and then fix them. Thanks!

jskeet: Added Common tests and fixed them... AspNetCore and AspNetCore3 should be okay, but the published docs are currently broken - I need to investigate that separately.

PR #6477, opened 2021-05-19
Regenerate all APIs with new generator

jskeet:
The new generator generates more obsolete methods and properties (to accurately reflect the API).


jskeet:
Looks like that's actually a bit that I missed in the generator... need to fix it there.

amanda-tarafa (approved):
LGTM

PR #6488, opened 2021-05-24
Diagnostics: Minor refactors and cleanups.

amanda-tarafa:
These are all things that I've found while working on another PR.
Commits are independent from one another.


jskeet (approved):
Nice :)

PR #5291, opened 2020-09-07
feat(spanner): add `query_optimizer_statistics_package` support

skuruppu:
The optimizer statistics package can be set through `QueryOptions`, which can be configured through the following mechanisms.
1. At the `SpannerConnection` level.
1. Through the `SPANNER_OPTIMIZER_STATISTICS_PACKAGE` environment variable.
1. At a query level.
    
If the options are configured through multiple mechanisms then:
1. Options set at an environment variable level will override options configured at the `SpannerConnection` level.
1. Options set at a query-level will override options set at either the `SpannerConnection` or environment variable level.
    
If no options are set, the optimizer statistics package will default to:
1. The package the database is pinned to.
1. If the database is not pinned to a specific package, then the Cloud Spanner backend will use the "latest" version.

PR #6584, opened 2021-06-04
feat(spanner): add query_optimizer_statistics_package support

skuruppu:
The optimizer statistics package can be set through `QueryOptions`, which can be configured through the following mechanisms.
1. At the `SpannerConnection` level.
1. Through the `SPANNER_OPTIMIZER_STATISTICS_PACKAGE` environment variable.
1. At a query level.
    
1. Options set at an environment variable level will override options configured at the `SpannerConnection` level.
1. Options set at a query-level will override options set at either the `SpannerConnection` or environment variable level.
    
If no options are set, the optimizer statistics package will default to:
1. The package the database is pinned to.
1. If the database is not pinned to a specific package, then the Cloud Spanner backend will use the "latest" version.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/QueryOptionsTests.cs
@@ -38,7 +38,9 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         {
             using (var connection = _fixture.GetConnection())
             {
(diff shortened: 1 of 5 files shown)


skuruppu:
@jskeet this is exactly where we left off #5291 but I wasn't able to re-open the PR so creating a new one here.

This change can now be merged as soon as it's approved.

jskeet:
The first commit doesn't seem to have any changes in... I'm a bit confused. Given the overall size of the change, would it be reasonable to rebase this to just a single commit?

(And has this been rebased to the current head yet? If not, I think that would be worth doing.)

skuruppu:
@jskeet all done. Commits are squashed into one.

jskeet (approved):

PR #6598, opened 2021-06-08
feat: Allows per log entry labels

amanda-tarafa:
And some refactoring needed for that.

As always one commit at a time is best.
Also, this does not include the moving of GoogleLogger to Common because it was becoming a very big PR.


amanda-tarafa:
I'll look at the failures, but all tests are passing locally. This is ready for review.

amanda-tarafa:
All checks are now green, I think there's a language version missmatch. I had to change

```
if (State.ToStructValue() is not Value newValue)
```
to
```
if (!(State.ToStructValue() is Value newValue))
```

jskeet:
Yup - `is not` is in C# 9, which isn't supported by the .NET Core 3.1 SDK we're still using to build. Once .NET 6 is released, we can update the SDK and start using *some* of the newer features...

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
```diff
@@ -268,5 +246,42 @@ internal void WriteDiagnostics(TextWriter writer)
...
+            private readonly GoogleLoggerScope _permanentParent;
+            private readonly IServiceProvider _serviceProvider;
+
+            internal AmbientScopeManager(LoggerOptions options, IServiceProvider serviceProvider)
+            {
+                _permanentParent = options?.Labels is null ? null : GoogleLoggerScope.CreateScope(new LabellingScopeState(options.Labels), null);
+                _serviceProvider = serviceProvider;
+            }
+
+            public GoogleLoggerScope GetCurrentScope()
```
jskeet, line 269: Presumably most of the time this will return null really quickly? Otherwise I can see the inefficiencies could be a little alarming, given that it's invoked for every log entry.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Logging/ILoggerExtensionsTest.cs
```diff
@@ -0,0 +1,211 @@
...
+        public void WithAddedLabels_NoWithLabels()
+        {
+            var scopeMock = GetScopeMock();
+
+            var loggerMock = new Mock<ILogger>(MockBehavior.Strict)
+                .SetupLog("Cutting hair")
+                .SetupBeginScope(scopeMock.Object, new KeyValuePair<string, string>("job_title", "hairdresser"));
+
+            var originalLogger = loggerMock.Object;
+            var augmentedLogger = originalLogger.WithAddedLabels(new KeyValuePair<string, string>("job_title", "hairdresser"));
```
jskeet, line 125: Nit: use `KeyValuePair.Create` instead of `new KeyValuePair<string, string>` everywhere in this file, just for brevity?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/KeyValuePairEnumerableExtensions.cs
```diff
@@ -55,5 +57,21 @@ public static Value ToStructValue(this IEnumerable<KeyValuePair<string, object>>
             }
             return hasValues ? Value.ForStruct(fieldsStruct) : null;
         }
+
+
```
jskeet, line 61: Nit: remove blank line?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/KeyValuePairEnumerableExtensions.cs
```diff
@@ -55,5 +57,21 @@ public static Value ToStructValue(this IEnumerable<KeyValuePair<string, object>>
...
+        /// <summary>
+        /// Returns a string representing the given fields with the format:
+        /// ['key1'='value1']['key2'='value2']...
+        /// </summary>
+        internal static string ToDefaultStringRepresentation<TKey, TValue>(this IEnumerable<KeyValuePair<TKey, TValue>> fields)
```
jskeet, line 66: This one's okay to be an extension method if that's helpful, given that it's internal.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/ILoggerExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// Adds labels to the returned logger that may be included with every subsequent log.
+        /// How this information is included will depend on the actual <paramref name="logger"/>.
+        /// GoogleLogger(TODO: change this to a cref when I've moved it to this package) will include this
+        /// information in <see cref="LogEntry.Labels"/>.
+        /// </summary>
+        /// <remarks>
+        /// If <paramref name="logger"/> had already been augmented with labels, the old labels will be replaced
+        /// by the new ones on the returned <see cref="ILogger"/>.
+        /// </remarks>
+        public static ILogger WithLabels(this ILogger logger, params KeyValuePair<string, string>[] labels) =>
```
jskeet, line 43: Are we happy for this to be on `ILogger`? I suspect so, but it's just worth thinking about it a bit more.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/ILoggerExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+            GaxPreconditions.CheckNotNull(logger, nameof(logger)) switch
+            {
+                AugmentedLogger augmented => augmented.WithLabels(labels),
+                _ => AugmentedLogger.ForLabels(logger, labels)
+            };
+
+        /// <summary>
+        /// Adds labels to the returned logger that may be included with every subsequent log.
```
jskeet, line 51: I think the difference between this and `WithLabels` should be in the summary of both of them - VS only shows the summary in IntelliSense for example (I think) and so it would be odd for users to not see any difference.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/ILoggerExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+            public AugmentedLogger WithAddedLabels(IEnumerable<KeyValuePair<string, string>> extraLabels) =>
+                new AugmentedLogger(_innerLogger, _labels.Concat(GaxPreconditions.CheckNotNull(extraLabels, nameof(extraLabels))));
+
+            public IDisposable BeginScope<TState>(TState state) => _innerLogger.BeginScope(state);
+
+            public bool IsEnabled(LogLevel logLevel) => _innerLogger.IsEnabled(logLevel);
+
+            public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception exception, Func<TState, Exception, string> formatter)
+            {
+                using(_innerLogger.BeginScope(new LabellingScopeState(_labels)))
```
jskeet, line 91: Nit: space after (

jskeet: This will make the labels the *innermost* scope for logging... whereas if I've added them at the start of the program, I think I'd expect them to be the *outermost* scope. Is that basically infeasible? Or have I misunderstood?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/ILoggerExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// <summary>
+    /// Extensions for <see cref="ILogger"/> to allow augmenting the logged information with contextual data.
+    /// GoogleLogger(TODO: change this to a cref when I've moved it to this package)
+    /// will add the information to the <see cref="LogEntry"/> that is sent to Google Cloud Logging.
+    /// For the consumption of other loggers, the information will be included in newly created scopes
+    /// at the moment of logging, in a format that all loggers should understand.
+    /// </summary>
+    public static class ILoggerExtensions
+    {
+        /// <summary>
```
jskeet, line 33: Should we have overloads for these methods accepting `IEnumerable<KeyValuePair<string, string>>` as well? That way users can supply a `Dictionary<string, string>` for example.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/LabellingScope.cs
```diff
@@ -0,0 +1,95 @@
...
+using System.Linq;
+
+namespace Google.Cloud.Diagnostics.Common
+{
+    /// <summary>
+    /// Represents the data for the labelling scope, basically, labels.
+    /// </summary>
+    // Note: We implement IEnumerable for the benefit of other ILogger implementations,
+    // so they can represent this information as they see fit.
+    public class LabellingScopeState : IEnumerable<KeyValuePair<string, object>>
```
jskeet, line 28: Does this need to be public?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/LabellingScope.cs
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    // Let's do this to allow overriding
+                    dictionary[pair.Key] = pair.Value;
+                }
+            }
+            Labels = new ReadOnlyDictionary<string, string>(dictionary);
+        }
+
+        internal LabellingScopeState(params KeyValuePair<string, string>[] labels)
+            :this((IEnumerable<KeyValuePair<string, string>>)labels)
```
jskeet, line 48: Nit: space after colon. Might be simpler as `: this(labels.AsEnumerable())` too...

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/LabellingScope.cs
```diff
@@ -0,0 +1,95 @@
...
+        protected internal override void ApplyThisScope(LogEntry entry)
+        {
+            // We very explicitly don't call the base implementation of this
+            // method which adds the state information to the scope field of the
+            // LogEntry.JsonPayload. We know these to be labels so we add them
+            // on the LogEntry.Labels collection.
+
+            foreach(var pair in State.Labels)
+            {
+                // It's OK to override, these we'll get executed in the correct order.
```
jskeet, line 90: we'll => will

jskeet (commented):
Fine with the general approach, now that I've got my head round it, but there's a mixture of nits and a very few more important things I'd like to either see changed or discuss further.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/ILoggerExtensions.cs
amanda-tarafa: Will do.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/ILoggerExtensions.cs
amanda-tarafa: For the `GoogleLogger`, that this is implemented with scopes, is just that, an implementation detail (basically the only way I could think of doing this). Labels are not treated as any other scopes, their values are just added to `LogEntry.Labels` and for that the only relevant thing is the order in which the labels are added, and that is respected whether we begin the scope at logging or elsewhere.

For other loggers, I would say the same, users are not adding a scope , so they shouldn't expect labels to behave as a scope. We (I) have decided to communicate the labelling information to other loggers "in the form of a scope **on logging**" (basically because there's nothing else I can do) and loggers will treat that as they treat other scopes. I think I can be more explicit about what happens with other loggers in docs, but I think the decision can be regarded as a design decision.

Beginning the scope when the labels are added is not infeasible but would add many other consecuenses as:
- The inner logger will have an open scope for everyone else using it from the moment you add labels. That would be unexpected I think
- I'd need to return an `IDisposable` somehow (or something wrapping that) so that user code can remove the labels from the inner logger.

The main feature this is trying to solve is being able to add labels per log entry. If you want to add labels that are applied to all log entries you can use `LoggerOptions.Labels` or `ILogEntryLabelProvider` which will only afect Google's LogEntry.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/LabellingScope.cs
amanda-tarafa: Won't be once I've moved the logger to Common.

amanda-tarafa (commented):
I've left a few comments (but ignored the nits for now). Let me know if you want to VC about some of this.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/ILoggerExtensions.cs
jskeet: Okay - sounds like we can regard it as unfortunate but basically infeasible to get what we might want.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/GoogleLoggerScope.cs
```diff
@@ -0,0 +1,187 @@
...
+        protected internal override void ApplyThisScope(LogEntry entry)
+        {
+            base.ApplyThisScope(entry);
+
+            if (!(State.ToStructValue() is Value newValue))
+            {
+                return;
+            }
+
+            if (entry.JsonPayload.Fields.TryGetValue("parent_scopes", out Value existingValues))
```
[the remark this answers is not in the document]
amanda-tarafa: Done.

apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore/Logging/GoogleLogger.cs
amanda-tarafa: Changed to use a single dictionary and labelling scope.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.Tests/Logging/ILoggerExtensionsTest.cs
amanda-tarafa: It's not available on .net461 :(

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/KeyValuePairEnumerableExtensions.cs
amanda-tarafa: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/ILoggerExtensions.cs
amanda-tarafa: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/ILoggerExtensions.cs
amanda-tarafa: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/Logging/LabellingScope.cs
amanda-tarafa: Done, both.

amanda-tarafa (commented):
Addressed all comments on the third commit. PTAL (LGTM is missing :))

I'm running tests on local

jskeet (approved):

PR #6604, opened 2021-06-09
Release Spanner libraries version 3.9.0

jskeet:
Changes in Google.Cloud.Spanner.Data version 3.9.0:

- [Commit 0fb438e](https://github.com/googleapis/google-cloud-dotnet/commit/0fb438e): feat(spanner): add `query_optimizer_statistics_package` support (see below)

The optimizer statistics package can be set through `QueryOptions`, which can be configured through the following mechanisms:

- At the `SpannerConnection` level.
- Through the `SPANNER_OPTIMIZER_STATISTICS_PACKAGE` environment variable.
- At a query level.

- Options set at an environment variable level will override options configured at the `SpannerConnection` level.
- Options set at a query-level will override options set at either the `SpannerConnection` or environment variable level.

If no options are set, the optimizer statistics package will default to the package the database is pinned to. If the database is not pinned to a specific package, then the Cloud Spanner backend will use the "latest" version.

Packages in this release:
- Release Google.Cloud.Spanner.Admin.Database.V1 version 3.9.0
- Release Google.Cloud.Spanner.Admin.Instance.V1 version 3.9.0
- Release Google.Cloud.Spanner.Common.V1 version 3.9.0
- Release Google.Cloud.Spanner.Data version 3.9.0
- Release Google.Cloud.Spanner.V1 version 3.9.0

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/docs/history.md
+++ b/apis/Google.Cloud.Spanner.Data/docs/history.md
@@ -1,5 +1,25 @@
 # Version history
 
+# Version 3.9.0, released 2021-06-09
+
+- [Commit 0fb438e](https://github.com/googleapis/google-cloud-dotnet/commit/0fb438e): feat(spanner): add `query_optimizer_statistics_package` support (see below)
+
+The optimizer statistics package can be set through `QueryOptions`, which can be configured through the following mechanisms:
+
+- At the `SpannerConnection` level.
+- Through the `SPANNER_OPTIMIZER_STATISTICS_PACKAGE` environment variable.
+- At a query level.
+
+If the options are configured through multiple mechanisms then:
+
+- Options set at an environment variable level will override options configured at the `SpannerConnection` level.
+- Options set at a query-level will override options set at either the `SpannerConnection` or environment variable level.
+
+If no options are set, the optimizer statistics package will default
+to the package the database is pinned to. If the database is not
+pinned to a specific package, then the Cloud Spanner backend will
+use the "latest" version.
+
 # Version 3.8.0, released 2021-05-18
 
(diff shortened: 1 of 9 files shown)


apis/Google.Cloud.Spanner.Data/docs/history.md
```diff
@@ -1,5 +1,25 @@
...
+
+The optimizer statistics package can be set through `QueryOptions`, which can be configured through the following mechanisms:
+
+- At the `SpannerConnection` level.
+- Through the `SPANNER_OPTIMIZER_STATISTICS_PACKAGE` environment variable.
+- At a query level.
+
+If the options are configured through multiple mechanisms then:
+
+- Options set at an environment variable level will override options configured at the `SpannerConnection` level.
```
amanda-tarafa, line 15: Oh is this backwards. It's a little unexpected if it's right. Specially given that the one at the query level overrides both environment and connection, so this is neither a top-down nor bottom-up approach.

amanda-tarafa (approved):
LGTM, but I have a comment.

apis/Google.Cloud.Spanner.Data/docs/history.md
jskeet: Hmm... indeed. Okay, will get back to Shanika to check that.

skuruppu: Yup, I see why it might be confusing. But this is what the feature team insisted should be the behavior. It works the same way for the other optimizer version field that we added early last year and this one needs to be consistent with that.

jskeet: Okay, thanks for the confirmation - will merge on that basis.

PR #6616, opened 2021-06-15
fix: synchronize access to the underlying transaction for ambient transactions

olavloite:
Synchronizes access to the underlying Spanner transaction to prevent multiple transactions from being created when parallel commands are executed on the ambient transaction.

See error and discussion in https://github.com/GoogleCloudPlatform/dotnet-docs-samples/pull/1320 for background.


olavloite:
cc @skuruppu

amanda-tarafa (commented):
Great catch! I just have a suggestion for using [Lazy](https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1?view=net-5.0) instead of a normal lock. I think we could avoid calling Task.Result etc. and in any case it would be cleaner I think?

jskeet:
I'd definitely be happier using `Lazy<T>` if we can... but it's not really clear to me how it's expected to work, in that if one thread has started acquiring a transaction, we *have* to block if another thread asks for the transaction. I think I'll need to look a little closer at where VolatileResourceManager is used...

olavloite:
Using `Lazy` would be an option, but as far as I understand, it is by design synchronous. That would mean that we would have to:
1. Change `GetTransactionAsync` to be a synchronous method.
2. Add some custom code to make an `AsyncLazy` implementation.
3. Add a custom library that supports some kind of `AsyncLazy`.
4. Or enlighten me on what other options there might be ;-)

amanda-tarafa:
You can have `Lazy<Task<Transaction>>` and then you would normally do `await Lazy.Value`. That would take care of not blocking on our Execute...Async methods. The methods that we are implementing from the `ISinglePhaseNotification` interface are still sync so we would have to block on those. I'm trying to confirm from MS docs that blocking there shouldn't be a problem since those methods shouldn't be called from multiple threads.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
```diff
@@ -24,46 +24,49 @@ namespace Google.Cloud.Spanner.Data
...
+        private SpannerTransaction SpannerTransaction { get => SpannerTransactionTask.Result; }
+
+        private Task<SpannerTransaction> SpannerTransactionTask
+        {
+            get => _transaction == null ? throw new ObjectDisposedException("The transaction has already been disposed") : _transaction.Value;
+        }
+
+        private bool IsTransactionCreated { get => _transaction != null && _transaction.IsValueCreated; }
+
+        private bool HasExecutedDmlOrMutations { get => IsTransactionCreated && (_hasExecutedDml || SpannerTransaction.HasMutations); }
```
jskeet, line 50: No need for all those braces :)

```suggestion
        private SpannerTransaction SpannerTransaction => SpannerTransactionTask.Result;

        private Task<SpannerTransaction> SpannerTransactionTask =>
            _transaction == null ? throw new ObjectDisposedException("The transaction has already been disposed") : _transaction.Value;

        private bool IsTransactionCreated => _transaction != null && _transaction.IsValueCreated;

        private bool HasExecutedDmlOrMutations => IsTransactionCreated && (_hasExecutedDml || SpannerTransaction.HasMutations);
```

jskeet (commented):
I *think* I follow this and like it, but I'd like Amanda to check as well :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
```diff
@@ -27,28 +27,49 @@ internal sealed class VolatileResourceManager : ISinglePhaseNotification, ISpann
...
+
         private Logger Logger => _spannerConnection.Logger;
 
+        private Task<SpannerTransaction> CreateTransaction()
```
amanda-tarafa, line 54:
```suggestion
        private Task<SpannerTransaction> GetCreateTransactionTask()
```
or even

```suggestion
        private Task<SpannerTransaction> CreateTransactionAsync()
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
```diff
@@ -27,28 +27,49 @@ internal sealed class VolatileResourceManager : ISinglePhaseNotification, ISpann
...
+                : _spannerConnection.BeginTransactionAsync();
+        }
+
         public void Dispose()
         {
             try
             {
-                _transaction?.Dispose();
-                // Protect against multiple disposals
+                if (_transaction != null && _transaction.IsValueCreated)
```
amanda-tarafa, line 65: You added a property for this:
```suggestion
                if (IsTransactionCreated)
```

amanda-tarafa (approved):
LGTM with a couple of suggestions.

amanda-tarafa:
Added do not merge label just while we finish discussing internally.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
olavloite: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
olavloite: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
olavloite: Done

olavloite:
I removed the `Do not merge` label as we have decided to go ahead with these changes including the DML check.

olavloite:
@amanda-tarafa @jskeet Do you have any last comments/questions before merging this?

jskeet:
Will have a look in a bit - I have a couple of other things on this morning. (If Amanda is happy, she's welcome to merge it - she's typically a better reviewer than I am anyway!)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
```diff
@@ -27,28 +27,48 @@ internal sealed class VolatileResourceManager : ISinglePhaseNotification, ISpann
...
+                : _spannerConnection.BeginTransactionAsync();
+        }
+
         public void Dispose()
         {
             try
             {
-                _transaction?.Dispose();
-                // Protect against multiple disposals
+                if (IsTransactionCreated)
```
amanda-tarafa, line 64: Actually I think you want here `_transaction is null` only, without caring about the value having been created or not.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionScopeTests.cs
```diff
@@ -365,6 +365,36 @@ public void MultiTableWrite()
...
+
+            // Read the inserted values.
+            using var connection = _fixture.GetConnection();
+            using var command = connection.CreateSelectCommand($"SELECT COUNT(*) AS C FROM {_fixture.TableName} WHERE K IN UNNEST(@Keys)");
+            command.Parameters.Add("Keys", SpannerDbType.ArrayOf(SpannerDbType.String)).Value = keys;
+            using var reader = command.ExecuteReader();
+
+            Assert.True(reader.Read());
+            Assert.Equal(keys.Length, reader.GetInt32(0));
+            Assert.False(reader.Read());
```
amanda-tarafa, line 396: nit: I would remove this one, your select is a COUNT and if that returns more than one row, that's a problem with the reader, not with this test.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -760,6 +761,36 @@ public void PdmlRetriedOnEosError()
...
+                var tables = new string[] { "TAB1", "TAB2", "TAB3" };
+                await Task.WhenAll(tables.Select(table =>
+                {
+                    using var cmd = connection.CreateInsertCommand(table);
+                    return cmd.ExecuteNonQueryAsync();
+                }));
+                scope.Complete();
+            }
+
+            spannerClientMock.Verify(client => client.CommitAsync(
```
amanda-tarafa, line 787: The verification in the next line superseeds this one. I think this one can be removed.

amanda-tarafa (approved):
LGTM with three small requests.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
olavloite: No, in this case we do care, because it could be that the `VolatileResourceManager` is being disposed without the `SpannerTransaction` ever having been initialized. `_transaction` is set to a [non-null value in the constructor](https://github.com/googleapis/google-cloud-dotnet/blob/5a20380ab74dbe7570cb2d3cb6480d32c05bad40/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs#L37). The underlying `SpannerTransaction` is however only created if it is actually needed. So if `Dispose()` would be called before the transaction has been used, it would unnecessarily first create a new `SpannerTransaction` and then dispose it right afterwards.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionScopeTests.cs
olavloite: Removed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
olavloite: I think it's better to have both, as they do check two different things:
1. The first verification checks that there is __only one__ `Commit` call, regardless of the number of mutations in the request.
2. The second verification checks that the only `Commit` call actually contains 3 mutations.

Removing the first verification would mean that if two commit requests were sent from the client, one without any mutations and the other with 3 mutations, the test would still pass. That is one of the scenarios that we want to verify is not happening. (Checking the number of `BeginTransaction` calls is not a good measure, as multiple transactions could be prepared by the session pool.)

amanda-tarafa: Ah yes, you are right.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/VolatileResourceManager.cs
amanda-tarafa: Yep, you are right.

amanda-tarafa (approved):
LGTM

PR #6712, opened 2021-07-05
Another attempt at using autosynth for Compute

jskeet:
We're now resilient to looking up repos in the preconfig file that
don't exist. We do assume that any one synthtool run only requires
*either* googleapis *or* googleapis-discovery, but that's a
reasonable expectation for the moment.

I'm not 100% confident this will work first time, but it's hard to
test without merging.


generateapis.sh
```diff
@@ -38,26 +39,46 @@ else
   declare -r GOOGLEAPIS="$PWD/googleapis"
 fi
 
-# TODO: Work out whether we need synthtool handling etc.
-declare -r GOOGLEAPIS_DISCOVERY="$PWD/googleapis-discovery"
+# Handle either a locally-fetched googleapis-discovery, or
+# one specified in a preconfig file.
+if [[ "$SYNTHTOOL_PRECONFIG_FILE" != "" ]]
+then
+  declare -r GOOGLEAPIS=$(python - <<END
```
amanda-tarafa, line 46: Should this one be GOOGLEAPIS_DISCOVERY? Not sure though.

amanda-tarafa (approved):
LGTM, but with a comment.

generateapis.sh
jskeet: Yes it should! Thanks and we'll spotted

PR #6724, opened 2021-07-08
Remove use of obsolete methods in manually-written snippets

### LINKED ISSUE #6476: PubSub API obsolete signatures
In the underlying Pub/Sub API, we have a signature for SubscriberServiceApiClient.Pull accepting a subscriber name, number of messages, and a returnImmediately flag. The returnImmediately field is now obsolete.

We should:

- Add a new signature to the proto which doesn't accept returnImmediately
- Regenerate
- Modify our snippets to use the new signature

jskeet:
Fixes #6476.


amanda-tarafa (approved):
LGTM

PR #6725, opened 2021-07-08
feat: allow adding an additional version header

olavloite:
Allows adding an additional version header to the connection string. This will be added to the `x-goog-api-client` header that is used by the underlying Spanner client. Only a fixed set of values may be set for the header (currently only 'efcore' is allowed), and the property is not intended for public use.


jskeet:
Will have a look tomorrow.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -333,6 +335,40 @@ public EmulatorDetection EmulatorDetection
...
 
+        /// <summary>
+        /// Other known clients may set an additional value for the x-goog-api-client version header. This
+        /// array contains the only valid values.
+        /// </summary>
+        private static readonly string[] s_allowedVersionHeaders = { "efcore" };
+
+        private string ValidateVersionHeader(string header)
+        {
+            GaxPreconditions.CheckNotNullOrEmpty(header, nameof(header));
```
jskeet, line 346: We should allow it to be set to null IMO, given that's the default value. (Setting a property to its current value shouldn't be an error.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
```diff
@@ -45,6 +45,13 @@ private static async Task<ChannelCredentials> CreatedScopedDefaultCredentials()
         /// </summary>
         internal string Endpoint { get; }
 
+        /// <summary>
+        /// Any additional version header to add to the Spanner client.
+        /// This should only be set by frameworks that are developed by Google and not by normal end user applications.
+        /// </summary>
+        internal string AdditionalVersionHeaderName { get; }
```
jskeet, line 52: This sounds like it's the name of the header, not the name of the product *in* the header. How about `AdditionalVersionHeaderProduct`?

Two other things to consider within this class:

- What if we want multiple additional headers?
- Do we actually need to separate out the product name from the product version here?

I'm wondering whether we could just have a single string which can contain multiple "name/value" pairs, space-separated. We can validate it early on, and then in SessionPoolManager do the actual splitting into multiple names and values, calling AppendVersion multiple times where appropriate.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
olavloite: Sounds reasonable to me, I'll update that.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
olavloite: > How about AdditionalVersionHeaderProduct?

Sounds good, I'll update that.

> What if we want multiple additional headers?

I'm not sure that is something that will happen very often (if at all), and it would complicate things quite a bit. It must be possible to specify the header in the connection string for it to be easy to pass in from generic frameworks like Entity Framework. Having a property in the connection string that could have multiple values (or something like `header1=x;header2=y`) would make it complicated, while I doubt that it would ever be used.

> Do we actually need to separate out the product name from the product version here?

Not necessarily, but we need to do the parsing and splitting somewhere. The `ConnectionStringBuilder` and connection string just contain one string with both the header name (product) and version. `SpannerSettings` expect the header and version to be specified as two separate fields. My feeling is that this is a better place than in the `SessionPoolManager`, which now does not need to know what the format is of the string that is set in the connection string, and can just get the values that it needs from `SpannerClientCreationOptions`.

jskeet: It feels like it's *simpler* to me - it's a single string value which is passed along everywhere until we reach the VersionHeaderBuilder. It's validated along the way (in terms of splitting), but it's just a single string, which is really flexible. If we've got two separate strings, then if we ever want that to be multiple name/value pairs we end up having to change the whole API.

Even when we're not using the flexibility of multiple versions, it means having a single property consistently, instead of it being one value in the connection string which becomes exactly two strings elsewhere.

Let's see what @amanda-tarafa thinks though.

amanda-tarafa (commented):
I think I would do this differently (sorry!).
I think it's a weird thing to expose through the connection string, tbh. I would overload the SessionPoolManager.Create method with one that takes SpannerSettings. This method would add the gccl header to the settings (which is the one we currently add to default settings). Then on the EF Core library, I would use this SessionPoolManager.Create (instead of SessionPoolManager.Default) method to pass some custom SpannerSettings containing the "fixed" efcore values for the header, and use that SessionPoolManager across the EF Core library instead.

This seems to me more general (being able to set custom SpannerSettings is not a bad thing) but also less prone to abuse.

olavloite:
@amanda-tarafa I considered doing (something like) what you are suggesting, but I decided against it, because EF Core itself will normally take either a [connection string](https://github.com/dotnet/efcore/blob/ac2bb48b10ecf1289b568a94b7a35e8075c6d787/src/EFCore.Relational/Infrastructure/RelationalOptionsExtension.cs#L94) or a [connection](https://github.com/dotnet/efcore/blob/ac2bb48b10ecf1289b568a94b7a35e8075c6d787/src/EFCore.Relational/Infrastructure/RelationalOptionsExtension.cs#L118) when being configured. The Spanner EF Core provider exposes (roughly, there are some additional overloads) two different public `UseSpanner` methods for initializing EF Core with Spanner:
- One that takes a connection string: This one does some specific Spanner configuration, adds the header to the connection string, and then calls the `RelationalOptionsExtension.WithConnectionString(string)` method.
- One that takes a connection. This one passes the connection into `RelationalOptionsExtension.WithConnection(Connection)`.

If it would not be possible to add the header to the connection string, we would need to change the code in the method that takes a connection string to create the `SpannerConnection` using a custom `SessionPoolManager` in that method and then call `RelationalOptionsExtension.WithConnection` instead. I would prefer to avoid having to do that whenever possible.

amanda-tarafa:
> * One that takes a connection string: This one does some specific Spanner configuration, adds the header to the connection string, and then calls the `RelationalOptionsExtension.WithConnectionString(string)` method.
> * One that takes a connection. This one passes the connection into `RelationalOptionsExtension.WithConnection(Connection)`.

Does this means that the header is only added if the first method is used? Just curious.

> If it would not be possible to add the header to the connection string,

I don't think it is impossible, but I do find it really odd, these headers sort of make no sense in the connection string.

> we would need to change the code in the method that takes a connection string to create the `SpannerConnection` using a custom `SessionPoolManager` in that method and then call `RelationalOptionsExtension.WithConnection` instead. I would prefer to avoid having to do that whenever possible.

Why would you prefer to not have to change these methods. Curious again.

Also, I'm thinking that we should add changes that work well for all possible clients of the library, not just for a wraper implementation of EF Core, and this is looking a little like that. I just don't want to add something public to Spanner.Data that is very restricted in scope and then we have to live with it forever, I rather we do something that makes sense for a wider set of use cases.

I'm leaving a couple more comments on the PR, I think the current implementation will fail.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -307,6 +307,12 @@ private static async Task<SpannerClient> CreateClientAsync(SpannerClientCreation
                 new ChannelOption(GcpCallInvoker.ApiConfigChannelArg, apiConfig.ToString())
             };
 
+            if (!string.IsNullOrEmpty(channelOptions.AdditionalVersionHeaderName))
+            {
+                spannerSettings.VersionHeaderBuilder.AppendVersion(channelOptions.AdditionalVersionHeaderName,
```
amanda-tarafa, line 312: This will [fail](https://github.com/googleapis/gax-dotnet/blob/ab0c5a8895567066541c5567c113ac189288dd46/Google.Api.Gax/VersionHeaderBuilder.cs#L48) if more than one client is created with the same SessionPoolManager.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -333,6 +335,40 @@ public EmulatorDetection EmulatorDetection
             }
         }
 
+        /// <summary>
+        /// Other known clients may set an additional value for the x-goog-api-client version header. This
+        /// array contains the only valid values.
+        /// </summary>
+        private static readonly string[] s_allowedVersionHeaders = { "efcore" };
```
amanda-tarafa, line 342: I don't think we should check this. If values here are relevant, then it should be the API the one to check not us.
And anyway, anyone using Spanner.Data can now set "efcore" themselves (without actually using the EF Core library) and it would muddy whatever is being tracked with this.
The header is in the end a good faith kind of identifier, for Spanner and the rest of the libraries.

amanda-tarafa (changes requested):
I think the implementation as it stands will fail.

olavloite:
> > * One that takes a connection string: This one does some specific Spanner configuration, adds the header to the connection string, and then calls the `RelationalOptionsExtension.WithConnectionString(string)` method.
> > * One that takes a connection. This one passes the connection into `RelationalOptionsExtension.WithConnection(Connection)`.
> 
> Does this means that the header is only added if the first method is used? Just curious.

Correct. There was a discussion offline whether it would be preferable to only use the connection as a 'template' and open a new connection with the same properties + the headers, but it was decided against that, as the user expects the connection that is passed in to actually be used. Setting the header after a connection has been created is not possible.

> > we would need to change the code in the method that takes a connection string to create the `SpannerConnection` using a custom `SessionPoolManager` in that method and then call `RelationalOptionsExtension.WithConnection` instead. I would prefer to avoid having to do that whenever possible.
> 
> Why would you prefer to not have to change these methods. Curious again.

What I want to avoid is to create any unnecessary discrepancies between how the user initializes the Spanner provider and how the Spanner provider initializes EF Core. So:
1. When the user initializes the Spanner provider with a `ConnectionString`, the provider should also initialize (the `RelationalOptionsExtension` of) EF Core with a `ConnectionString`.
2. When the user initializes the Spanner provider with a `DbConnection`, the provider should also initialize (the `RelationalOptionsExtension` of) EF Core with a `DbConnection`.

The difference in behavior between 1 and 2 is very small as long as the connection that is used in 2 is not yet opened when it is passed in to EF Core. The main difference is that option 1 will only lazily create the connection when it is actually needed. It will do so by calling the [`RelationalConnection.CreateDbConnection` virtual method](https://github.com/cloudspannerecosystem/dotnet-spanner-entity-framework/blob/c520460874ac78edd68d3c071cb74e80575ac9ae/Google.Cloud.EntityFrameworkCore.Spanner/Storage/Internal/SpannerRelationalConnection.cs#L51) in the Spanner EF Core provider. 

> 
> Also, I'm thinking that we should add changes that work well for all possible clients of the library, not just for a wraper implementation of EF Core, and this is looking a little like that. I just don't want to add something public to Spanner.Data that is very restricted in scope and then we have to live with it forever, I rather we do something that makes sense for a wider set of use cases.
> 

This is actually one of the other reasons that I think we should add it to the `ConnectionString` and not only allow it to be set by creating a custom `SessionPoolManager`. The fact that EF Core offers the possibility of actually creating and setting a custom `SessionPoolManager` might not mean that all other frameworks will allow us to do the same.

For example for JDBC we also added this feature to the connection string, and it is currently used by the [Hibernate provider](https://github.com/GoogleCloudPlatform/google-cloud-spanner-hibernate/blob/f8e678a5969408949bc6ed5df98fe599531a5ece/google-cloud-spanner-hibernate-dialect/src/main/java/com/google/cloud/spanner/hibernate/SpannerServiceContributor.java#L69) and the [Liquibase extension](https://github.com/cloudspannerecosystem/liquibase-spanner/blob/b4372f3f96c0adf4cb31e4edd7b99d97f105c4cf/src/main/java/liquibase/ext/spanner/CloudSpannerConnection.java#L72). Both frameworks only allow us to configure the database through the connection string.

jskeet:
Hmm... I'm with Amanda in that this really doesn't feel like something that's inherently part of the connection string. I can see the appeal of making it consistent with Hibernate and Liquibase, but if we're effectively forced to do something odd in those cases, I'm not sure we should propagate that constraint to the .NET code.

If we can reasonably design this to avoid it being in the connection string, I'd prefer that. If it's a fact of life due other aspects of EF Core, that's a somewhat different situation.

(I do also agree with Amanda that we shouldn't validate the expected values within the library. It's reasonable to validate the *syntax* - e.g. entirely letters, digits and "." for example - but if someone wanted to build (say) an NHiberate provider, they shouldn't need a change to the core Spanner libraries to do so.

olavloite:
> Hmm... I'm with Amanda in that this really doesn't feel like something that's inherently part of the connection string. I can see the appeal of making it consistent with Hibernate and Liquibase, but if we're effectively forced to do something odd in those cases, I'm not sure we should propagate that constraint to the .NET code.

My argument in this case is not about being consistent with Hibernate and Liquibase, but the fact that some of these frameworks will create the connection internally based only on the connection string. EF Core supports passing in both a connection string and an existing connection during initialization. If the latter had not been supported, it would not have been possible to use a custom `SessionPoolManager` with the connection. Such a solution would therefore not work for any framework that only allows a connection string to be passed in during initialization. Hibernate and Liquibase are two examples of that. If that is not a worry for .NET, or if we don't consider it a problem that it can't be used in that case, then we can implement it using a setting on the `SessionPoolManager`.

> 
> If we can reasonably design this to avoid it being in the connection string, I'd prefer that. If it's a fact of life due other aspects of EF Core, that's a somewhat different situation.

For EF Core, it is not a requirement. It will cause an EF Core `DbContext` to be initialized with an (unopened) connection, even when the client application initializes it with a connection string. Normally, EF Core would create the connection lazily when one is needed here: https://github.com/dotnet/efcore/blob/4507cc38283ebc1c981c224d76612540ce30aee6/src/EFCore.Relational/Storage/RelationalConnection.cs#L107

> 
> (I do also agree with Amanda that we shouldn't validate the expected values within the library. It's reasonable to validate the _syntax_ - e.g. entirely letters, digits and "." for example - but if someone wanted to build (say) an NHiberate provider, they shouldn't need a change to the core Spanner libraries to do so.

The validation is primarily about discouraging (un)intentional use of this property. The header is used internally to measure the usage of the different connection options to Spanner. By only allowing a fixed set of values, we prevent anyone who might not know what it's for from setting a value for this property to a random value. And anyone implementing for example an NHibernate provider would only have use for this property if it were done in coordination with the backend team, as that is where this header is used to monitor the usage.
(Preferably, the entire property should be private, but that is unfortunately not doable, as the EF Core provider and other similar frameworks need to use it.)

amanda-tarafa:
> My argument in this case is not about being consistent with Hibernate and Liquibase, but the fact that some of these frameworks will create the connection internally based only on the connection string.
> ...
> For EF Core, it is not a requirement. It will cause an EF Core DbContext to be initialized with an (unopened) connection, even when the client application initializes it with a connection string. Normally, EF Core would create the connection lazily when one is needed here: [CreateDbConnection](https://github.com/dotnet/efcore/blob/4507cc38283ebc1c981c224d76612540ce30aee6/src/EFCore.Relational/Storage/RelationalConnection.cs#L107)

I'm probably missing something here, but since you gave these arguments a few comments above, in my head they make total sense if users of these platforms needed to change this setting (the header), and then it'd make total sense if it were a setting of the connection string. But that's not the case, this is not a setting to be changed by final users of data frameworks but by provider implementors. The method you linked [CreateDbConnection](https://github.com/dotnet/efcore/blob/4507cc38283ebc1c981c224d76612540ce30aee6/src/EFCore.Relational/Storage/RelationalConnection.cs#L107) is rightly abstract because provider implementors, and only them, know how to create a connection to the underlying data system (and I suppose you have that same control as an implementor of providers for other platforms like NHibernate, etc.). So I still don't see the problem with implementing that class/method with something like this (pardon some of the bits I'm ignoring) if we were to add the appropiate methods in SessionPoolManager, etc. Still again, I might be missing something here.

```
class SpannerRelationConnection : RelationalConnection
{
    internal static SessionPoolManager EFCorePoolManager { get; }

    static SpannerRelationConnection()
    {
        var settings = new SpannerSettings();
        settings.VersionHeaderBuilder.AppendVersion("efcore", "1.0.0");
        // We'd need to add this Create method overload that receives a SpannerSettings.
        EFCorePoolManager = SessionPoolManager.Create(new SessionPoolOptions(), settings);
    }

    protected override DbConnection CreateDbConnection()
    {
        var connectionStringBuilder = new SpannerConnectionStringBuilder(_connectionString, null, EFCorePoolManager);
        return new SpannerConnection(connectionStringBuilder);
    }
}
```

As for the validation, I really think it doesn't add much protection tbh. I'd rather just document what the feature is supposed to be used for, and not show it on any samples. Why I think I adds little:
- Anyone using the low level client libraries can still set the header to wathever they want.
- Anyone using the Spanner.Data library, but not the EFCore provider, can still set it to "efcore" (or future valid values) and it would be valid.
- Users of EFCore passing in their own connection won't count anyway.

olavloite:
@amanda-tarafa I've removed the header from the connection string and added an overload to allow the `SessionPoolManager` to be created with custom `SpannerSettings`. I've verified that this will work to append the additional header for EF Core. PTAL.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -107,7 +111,17 @@ private static SpannerSettings CreateSpannerSettingsWithVersionHeader()
...
+
+        /// <summary>
+        /// Creates a <see cref="SessionPoolManager"/> with the specified SpannerSettings and options.
+        /// </summary>
+        /// <param name="spannerSettings">The SpannerSettings to use. May be null, in which case the default settings are used.</param>
+        /// <param name="options">The options to use. Must not be null.</param>
+        /// <param name="logger">The logger to use. May be null, in which case the default logger is used.</param>
+        /// <returns>A <see cref="SessionPoolManager"/> with the given options.</returns>
+        public static SessionPoolManager Create(SessionPoolOptions options, SpannerSettings spannerSettings, Logger logger = null) =>
+            new SessionPoolManager(options, spannerSettings ?? DefaultSpannerSettings(), logger ?? Logger.DefaultLogger, CreateClientAsync);
```
jskeet, line 124: SpannerSettings already has a logger... I'm wondering whether this overload should *not* take a logger, but use the one from there. (Basically, if you're customizing the settings already and you want a custom logger, put it in the settings.)

If we do still want the logger parameter, I'd suggest using `logger ?? spannerSettings.Logger ?? Logger.DefaultLogger`.

jskeet: (But having it separately in the internal constructor still makes sense, for the case where the settings *aren't* specified explicitly.)

jskeet (commented):
One suggestion around the overload, but generally this looks *much* simpler to me. Thanks so much for bearing with us!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
olavloite: That does make sense, __however__: It could possibly break existing applications that are currently calling `SessionPoolManager.Create(options, null)`, as that would then become an ambiguous invocation. So I guess using `logger ?? spannerSettings.Logger ?? Logger.DefaultLogger` would be the safer option.

jskeet: It won't be ambiguous - it will unambiguously use the new overload (which won't have any optional parameters). It's a little odd that a version change would change the selected overload - but the effect (default settings, default logger) would be the same.

It doesn't change binary compatibility as any code *previously* compiled (where it picks the original method) will have the method signature in the IL already, and that method still exists.

olavloite: I think it will be ambiguous, but __only__ for anyone that is calling the current method with an unnecessary `null` for the second argument (the logger):

```csharp
var mgr = SessionPoolManager.Create(new SessionPoolOptions(), null);
```

If we were to remove the `Logger` argument from the new overload, we would have the following method definitions:

```csharp
public static SessionPoolManager Create(SessionPoolOptions options, Logger logger = null) => ...

public static SessionPoolManager Create(SessionPoolOptions options, SpannerSettings spannerSettings) => ...
```

So I don't think there will be very many (if any) who does, but in theory it could cause a compilation error in existing code.

Or did I misunderstand how you wanted to define the new overload without the logger argument?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -74,28 +74,32 @@ static SessionPoolManager()
         /// Currently the default settings are used in all cases, but with the "gccl" version header added to specify the version of Google.Cloud.Spanner.Data
```
amanda-tarafa, line 74: nit: This doc is outdated.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -74,28 +74,32 @@ static SessionPoolManager()
...
         /// being used.
         /// </summary>
-        internal SpannerSettings SpannerSettings { get; } = CreateSpannerSettingsWithVersionHeader();
+        internal SpannerSettings SpannerSettings { get; }
 
-        private static SpannerSettings CreateSpannerSettingsWithVersionHeader()
+        private static SpannerSettings AppendVersionHeader(SpannerSettings settings)
```
amanda-tarafa, line 79: nit:
```suggestion
        private static SpannerSettings AppendAssemblyVersionHeader(SpannerSettings settings)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -107,7 +111,18 @@ private static SpannerSettings CreateSpannerSettingsWithVersionHeader()
...
+
+        /// <summary>
+        /// Creates a <see cref="SessionPoolManager"/> with the specified SpannerSettings and options.
+        /// </summary>
+        /// <param name="spannerSettings">The SpannerSettings to use. May be null, in which case the default settings are used.</param>
+        /// <param name="options">The options to use. Must not be null.</param>
+        /// <param name="logger">The logger to use. May be null, in which case the logger returned by spannerSettings.Logger will be used.
+        /// If that is also null, the default logger is used.</param>
+        /// <returns>A <see cref="SessionPoolManager"/> with the given options.</returns>
+        public static SessionPoolManager Create(SessionPoolOptions options, SpannerSettings spannerSettings, Logger logger = null) =>
```
amanda-tarafa, line 124: +1 to removing the logger, and I think it's not ambiguos even if someone was specifying `null` for the logger before. From now on the compiler will pick the one that has no optional parameters, which is what Jon mentioned that was weird, but has no practical effects, because both methods end up doing the same.

Actually, I just tried it:

```
public void Method(int a, object o = null) { }
public void Method(int a, string s) { }
...
Method(5, null); // This binds to the one with the string parameter.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -107,7 +111,18 @@ private static SpannerSettings CreateSpannerSettingsWithVersionHeader()
...
+            new SessionPoolManager(options, spannerSettings ?? DefaultSpannerSettings(), logger ?? spannerSettings?.Logger ?? Logger.DefaultLogger, CreateClientAsync);
```
amanda-tarafa, line 125: Consider creating a new instance of SpannerSettings here `new SpannerSettings(spannerSettings)` to protect ourselves from changes being made to it after the fact.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -74,28 +74,32 @@ static SessionPoolManager()
...
 
-        private static SpannerSettings CreateSpannerSettingsWithVersionHeader()
+        private static SpannerSettings AppendVersionHeader(SpannerSettings settings)
         {
-            var settings = new SpannerSettings();
             settings.VersionHeaderBuilder.AppendAssemblyVersion("gccl", typeof(SessionPoolManager));
             return settings;
         }
 
+        internal static SpannerSettings DefaultSpannerSettings() => new SpannerSettings();
```
amanda-tarafa, line 85: nit:
```suggestion
        internal static SpannerSettings CreateDefaultSpannerSettings() => new SpannerSettings();
```

amanda-tarafa (commented):
Thanks Knut, I think this is clearer.

LGTM mostly, except for a couple of minor comments.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
jskeet: I promise you, it really won't be ambiguous. Overload resolution has a tie-break that means a candidate method that doesn't need the compiler to fill in any optional parameters with default values is "better than" a method that doesn't.
(But don't just take my word for it - try it :)

Those overloads are indeed what I'd be expecting.

olavloite: > (But don't just take my word for it - try it :)

I know that I'm communicating with people with a lot more C# experience than myself here, so to be a little bit more confident I did try it before I stated my initial "I think it will break..." :-)

But apparently not all compilers (or IDEs) agree in this case:

Screenshot from 2021-07-21 13-50-16

jskeet: Interesting. Will try it in VS by pulling from your branch first...
(I'm very happy to be proved wrong... it's going to be more awkward to decide what to do if this turns out to be a Rider bug.)

jskeet: VS agrees with you. Off for lunch now, but will investigate further afterwards.

For reference, this is what I expect to apply: https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#12643-better-function-member

In particular:

> If neither function member was found to be better, and all parameters of Mᵥ have a corresponding argument whereas default arguments need to be substituted for at least one optional parameter in Mₓ, then Mᵥ is better than Mₓ. Otherwise, no function member is better.

I need to work out why that doesn't apply here...
(And obviously we need to work out what to do in terms of the methods... having an extra unnatural parameter doesn't feel ideal. I wonder whether we should rename the one that accepts settings instead?)

jskeet: Oh, doh - I'm thinking of the rule the wrong way round ignore me, and I'm very glad you tested it. That rule is what makes the logger version "win" if you had an overload of 

```csharp
public static SessionPoolManager Create(SessionPoolOptions options, Logger logger = null, SpannerSettings spannerSettings = null)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
olavloite: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
olavloite: Good point, done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
olavloite: Done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
olavloite: Good catch, I've edited the comment.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
olavloite: I've removed the `Logger` argument from the overload and renamed the method to `CreateWithSettings` (name open for discussion). I agree that the additional `Logger` argument only makes it more complex.

jskeet: Fab, thanks. Will look after lunch. Sorry again about the incorrect assertion around ambiguity...

amanda-tarafa: Ah, and the one I tested wasn't ambiguous because string is a subtype of object, and that's why the method with the string was chosen.

jskeet (approved):
That's great - thanks very much!

amanda-tarafa (approved):
LGTM!

PR #6739, opened 2021-07-14
Start generating canonical links

jskeet:
https://cloud.google.com/dotnet/docs/reference is now live, so let's
use it.

(Later we'll convert all GitHub Pages files into redirects, which
will be handy.)


amanda-tarafa (approved):
LGTM

PR #6745, opened 2021-07-15
Explicitly use Google.Protobuf.WellKnownTypes.Value in ValueConverter

jskeet:
This will fix the post-generation build problem described in #6743. (We'll need to rerun autosynth to clear it.)


amanda-tarafa (approved):
LGTM

PR #6746, opened 2021-07-15
Fix links between APIs

jskeet:
We no longer need to worry about the difference between GitHub pages and googleapis.dev, as we don't publish to GitHub pages any more. But currently the links are broken on cloud.google.com.

In some cases (e.g. Language and Error Reporting) the links have been removed as they're out of date.


amanda-tarafa (approved):
LGTM

PR #6748, opened 2021-07-15
Tool to rewrite GitHub pages to redirects

jskeet:
This will be removed in a subsequent PR, but it's good to have in
the history. (Aside from anything else, we don't depend on the
content of the pages, so if we want to change what the redirect page
looks like, we can do that and just rerun the tool.)


jskeet:
(I'll create a PR with the results when I can... but it's taking a while to create the commit.)

amanda-tarafa (approved):
LGTM

jskeet:
Great - will push the change straight to the gh-pages branch as GitHub won't be able to cope with the diffs otherwise...

PR #6749, opened 2021-07-15
Remove GitHub pages rewriting tool

jskeet:
(It's in the history so we can revive it if necessary, but we don't
expect to need to.)


amanda-tarafa (approved):
LGTM

PR #6750, opened 2021-07-15
Fix link canonicalizer for Google.Cloud.Common package

jskeet:
(That package doesn't quite follow the normal conventions, so needs
special-casing.)


amanda-tarafa (approved):
LGTM

PR #6754, opened 2021-07-16
Release Google.Cloud.Security.PrivateCA.V1 version 2.0.0

jskeet:
Changes in this release:

- [Commit f09fdff](https://github.com/googleapis/google-cloud-dotnet/commit/f09fdff): fix!: mark some bools as optional, correct response type of DeleteCaPool

This is a new major version as the change in response type for `DeleteCaPool` is a breaking change. Customers using the 1.x libraries are encouraged to upgrade, as attempting to obtain the result of the operation created by the `DeleteCaPool` RPC will throw an exception at execution time.


amanda-tarafa (approved):
LGTM

PR #6764, opened 2021-07-20
tests(spanner): fix optimizer stats package IT

skuruppu:


amanda-tarafa:
What about using latest as you did on the sample repo, which was green there?

amanda-tarafa:
I just checked on the console, a newly created Database has no optimization packages (the query returns no rows). And that actually makes sense, since probably for building an automatic optimization package the system needs some queries to have been ran and some time.

Maybe if you specify latest as the options and there are no packages, the backend will simply ignore the option? I'm not familiar with the full feature specs.

skuruppu:
Thanks for verifying it @amanda-tarafa. That makes sense. I think the ITs actually passed on CI. Maybe we wait long enough between creating the database and running the test compared to when I'm running a filtered test locally.

But in any case, it may end up being flaky if I go with my approach so I changed it to use "latest". This is what folks working on the other repos are doing anyway.

jskeet (approved):

PR #6769, opened 2021-07-21
Release Google.Cloud.GkeHub.V1 version 1.0.0-beta01

jskeet:
Changes in this release:

Initial release.


jskeet:
Please review the second commit particularly carefully. Happy to chat about it over a VC.

amanda-tarafa (approved):
LGTM. And yes, this is better, it would have look like different APIs with the version at the end.

I'm wondering, doesn't this happen to the beta API as well, and would it make sense to change the library for the beta API, so that clients switching only have to change the version part of the namespaces?

jskeet:
The beta API doesn't have the extra protos, which is why we didn't see it before: https://github.com/googleapis/googleapis/tree/master/google/cloud/gkehub/v1beta1

(Without the proto patching, this code doesn't even generate using our script.)

PR #6780, opened 2021-07-23
Remove root docs now we don't have GitHub Pages

jskeet:
This change may well be incomplete, both in terms of what it can remove, and whether it leaves us in a broken state.

I've tried to think of everything, but I may well have missed something. If something's broken, we'll just fix it - and if we've missed other opportunities, we can take advantage of those over time.


amanda-tarafa (approved):
LGTM, as far as I understand it.

PR #6792, opened 2021-07-28
Fix docs build for cloud.google.com

jskeet:
(I was too enthusiastic when deleting code from builddocs.sh previously.)


amanda-tarafa (approved):
LGTM

PR #6795, opened 2021-07-28
Add Renovate to this repository

jskeet:
This only affects test/tooling projects, not production libraries. (We manage the dependencies manually for those.)

Given that the test projects are generated, we'll probably *actually* want to close each Renovate PR affecting those, but then update the project generator and rerun it - Renovate will be effectively just alerting us, but that's okay.


amanda-tarafa (approved):
LGTM

PR #6796, opened 2021-07-28
Adjust README generation for cloud.google.com

jskeet:
Most of the code is just moved, but the URL generation is adjusted to take into account cloud vs non-cloud


jskeet:
(This will conflict with #6795, but I'll rebase whichever is approved later.)

amanda-tarafa (approved):
LGTM

PR #6816, opened 2021-07-29
Remove tools we never use any more

jskeet:
This code hasn't been updated for nearly 4 years, and I doubt that
it's been run in that period. It will always attract dependency
updates, and we don't really need them.


amanda-tarafa (approved):
LGTM

PR #6818, opened 2021-07-29
Remove the old release note generator

jskeet:
This had already been removed from the tools solution file, but the
code was still present.


amanda-tarafa (approved):
LGTM

PR #6848, opened 2021-07-31
feat(spanner): support comments and statement hints in untyped commands

### LINKED ISSUE #6847: [Spanner]: Statements that start with a comment or statement hint are not recognized
Statements that are specified as only text and that start with a comment or a statement hint is not recognized as one of the valid Spanner command types. This means that it is not possible to use for example the tagging feature in Linq when working with Spanner through Entity Framework Core.

Also mentioned in #5857

olavloite:
SpannerCommandTextBuilder will try to determine the type of command based on the first word of the command text if no type is specified. This works well as long as the command text does not contain any comments or statement hints at the beginning of the command text.
This change adds a simple parser that will remove any comments and statement hints before checking the type of command. The command that is stripped of comments and hints is only used for determining the type of command. The original command will be passed on to Cloud Spanner.

Fixes #6847

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
@@ -230,9 +230,9 @@ namespace Google.Cloud.Spanner.Data.Tests
         [InlineData("SELECT * FROM Foo@{FORCE_INDEX=`INDEX FOR SELECT`}", "SELECT * FROM Foo@{FORCE_INDEX=`INDEX FOR SELECT`}")]
         [InlineData("@{OPTIMIZER_VERSION=1 SELECT 1", "@{OPTIMIZER_VERSION=1 SELECT 1")]
         [InlineData("{OPTIMIZER_VERSION=1} SELECT 1", "{OPTIMIZER_VERSION=1} SELECT 1")]
-        public void RemoveStatementHintAndComments(string sql, string sqlWithoutHintsAndComments)
+        public void RemoveStatementHints(string sql, string sqlWithoutHints)
         {
-            Assert.Equal(sqlWithoutHintsAndComments, SpannerCommandTextBuilder.RemoveCommentsAndStatementHint(sql));
+            Assert.Equal(sqlWithoutHints, SpannerCommandTextBuilder.RemoveCommentsAndStatementHints(sql));
         }
     }
 }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using Xunit;
 
 namespace Google.Cloud.Spanner.Data.Tests
@@ -77,6 +79,20 @@ namespace Google.Cloud.Spanner.Data.Tests
         [InlineData("\t\nWITH AlbumsView AS (SELECT * FROM Albums) select Title from AlbumsView")]
         [InlineData(" WITH   AlbumsView AS (SELECT * FROM Albums) select Title from AlbumsView")]
         [InlineData("WITH\t\nAlbumsView AS (SELECT * FROM Albums) select Title from AlbumsView")]
+        [InlineData("-- Single line comment\nSELECT * FROM Albums")]
+        [InlineData("# Single line comment\nSELECT * FROM Albums")]
+        [InlineData("/* Multi\nline\ncomment\n */\nSELECT * FROM Albums")]
+        [InlineData("/* Multi\n * line\n * comment\n */\nSELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} SELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} \n-- Single line comment\nSELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} \n# Single line comment\nSELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} /* Multi\nline\ncomment\n */\nSELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} /* Multi\n * line\n * comment\n */\nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("# Single line comment\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("/* Multi\nline\ncomment\n */\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("/* Multi\n * line\n * comment\n */\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE, OPTIMIZER_VERSION=1} \n\t \nSELECT * FROM Albums")]
         public void SelectCommand(string commandText)
         {
             var builder = SpannerCommandTextBuilder.FromCommandText(commandText);
@@ -95,6 +111,10 @@ namespace Google.Cloud.Spanner.Data.Tests
         [InlineData("\t\nINSERT\t\nAlbums\t\n(AlbumId)t\nVALUESt\n(@id)")]
         [InlineData("INSERT  Albums  (AlbumId) VALUES (@id)")]
         [InlineData("INSERT\t\nAlbums\t\n(AlbumId)t\nVALUESt\n(@id)")]
+        [InlineData("-- Single line comment\nINSERT INTO Albums (AlbumId) VALUES (@id)")]
+        [InlineData("# Single line comment\nINSERT INTO Albums (AlbumId) VALUES (@id)")]
+        [InlineData("/* Multi line comment */ INSERT INTO Albums (AlbumId) VALUES (@id)")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} INSERT INTO Albums (AlbumId) VALUES (@id)")]
         public void DmlInsertCommand(string commandText)
         {
             var builder = SpannerCommandTextBuilder.FromCommandText(commandText);
@@ -108,6 +128,10 @@ namespace Google.Cloud.Spanner.Data.Tests
         [InlineData("\t\nUPDATE\t\nAlbumst\nSETt\nTitle=@titlet\nWHEREt\nTRUE")]
         [InlineData("UPDATE  Albums  SET Title=@title WHERE TRUE")]
         [InlineData("UPDATE\t\nAlbums\t\nSETt\nTitle=@titlet\nWHEREt\nTRUE")]
+        [InlineData("-- Single line comment\nUPDATE Albums SET Title=@title WHERE TRUE")]
+        [InlineData("# Single line comment\nUPDATE Albums SET Title=@title WHERE TRUE")]
+        [InlineData("/* Multi line comment */ UPDATE Albums SET Title=@title WHERE TRUE")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} UPDATE Albums SET Title=@title WHERE TRUE")]
         public void DmlUpdateCommand(string commandText)
         {
             var builder = SpannerCommandTextBuilder.FromCommandText(commandText);
@@ -126,6 +150,10 @@ namespace Google.Cloud.Spanner.Data.Tests
         [InlineData("\t\nDELETE\t\nAlbumst\nWHEREt\nTRUE")]
         [InlineData("DELETE  Albums  WHERE TRUE")]
         [InlineData("DELETE\t\nAlbumst\nWHEREt\nTRUE")]
+        [InlineData("-- Single line comment\nDELETE FROM Albums WHERE TRUE")]
+        [InlineData("# Single line comment\nDELETE FROM Albums WHERE TRUE")]
+        [InlineData("/* Multi line comment */ DELETE FROM Albums WHERE TRUE")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} DELETE FROM Albums WHERE TRUE")]
         public void DmlDeleteCommand(string commandText)
         {
             var builder = SpannerCommandTextBuilder.FromCommandText(commandText);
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -218,9 +220,9 @@ internal string DatabaseToDrop
         public static SpannerCommandTextBuilder FromCommandText(string commandText)
         {
             GaxPreconditions.CheckNotNullOrEmpty(commandText, nameof(commandText));
-            commandText = commandText.Trim();
+            var trimmedCommandText = RemoveCommentsAndStatementHint(commandText).Trim();
```
amanda-tarafa, line 223: `commandText` is still being used in many places, let's keep it trimmed.
```suggestion
            var trimmedCommandText = RemoveCommentsAndStatementHint(commandText).Trim();
            commandText = commandText.Trim();
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -218,9 +220,9 @@ internal string DatabaseToDrop
...
             // Split(new char[0]) splits the string using all whitespace characters.
-            var commandSections = commandText.Split((char[]) null, StringSplitOptions.RemoveEmptyEntries);
+            var commandSections = trimmedCommandText.Split((char[]) null, StringSplitOptions.RemoveEmptyEntries);
             if (commandSections.Length < 2)
             {
                 throw new ArgumentException($"'{commandText}' is not a recognized Spanner command.", nameof(commandText));
```
amanda-tarafa, line 228: Shoul this be:
```suggestion
                throw new ArgumentException($"'{trimmedCommandText}' is not a recognized Spanner command.", nameof(commandText));
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -218,9 +220,9 @@ internal string DatabaseToDrop
         public static SpannerCommandTextBuilder FromCommandText(string commandText)
         {
             GaxPreconditions.CheckNotNullOrEmpty(commandText, nameof(commandText));
-            commandText = commandText.Trim();
+            var trimmedCommandText = RemoveCommentsAndStatementHint(commandText).Trim();
             // Split(new char[0]) splits the string using all whitespace characters.
-            var commandSections = commandText.Split((char[]) null, StringSplitOptions.RemoveEmptyEntries);
+            var commandSections = trimmedCommandText.Split((char[]) null, StringSplitOptions.RemoveEmptyEntries);
```
amanda-tarafa, line 225: In line 262 here (I can't leave a comment there), the code is doing `commandText = $"{commandName} {targetTable}";` and then a `SpannerCommandTextBuilder` is returned based on `commandText`, which means you loose any comments or hints, etc.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -276,5 +278,190 @@ private static string ValidateTableName(string table, string parameterName)
             }
             return table;
         }
+
+        private static readonly char s_singleQuote = '\'';
+        private static readonly char s_doubleQuote = '"';
+        private static readonly char s_backtickQuote = '`';
+        private static readonly char s_hyphen = '-';
+        private static readonly char s_dash = '#';
```
amanda-tarafa, line 286: I think this is sharp?
```suggestion
        private static readonly char s_sharp = '#';
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -276,5 +278,190 @@ private static string ValidateTableName(string table, string parameterName)
...
+            StringBuilder res = new StringBuilder(sql.Length);
+            int index = 0;
+            while (index < sql.Length)
+            {
+                char c = sql[index];
+                if (isInQuoted)
+                {
+                    if ((c == '\n' || c == '\r') && !isTripleQuoted)
+                    {
+                        throw new SpannerException(
```
amanda-tarafa, line 317: I think this being a parsing issue, it should throw the same exception as the `FromCommandText` method, which is `InvalidOperationException`. And applies to the rest of the exceptions here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -276,5 +278,190 @@ private static string ValidateTableName(string table, string parameterName)
...
+            {
+                res.Remove(res.Length - 1, 1);
+            }
+            return res.ToString();
+        }
+
+        /// <summary>
+        /// First removes all comments from the entire sql string, and then removes any statement hints
+        /// at the start of the sql statement. The returned statement should only be used to determine the
+        /// type of statement. The original statement should always be passed on to Cloud Spanner.
```
amanda-tarafa, line 427: This is what is not happening in all cases because of line 262.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -276,5 +278,190 @@ private static string ValidateTableName(string table, string parameterName)
...
+            sql = RemoveComments(sql);
+
+            // Valid statement hints at the beginning of a query statement can only contain a fixed set of
+            // possible values. Although it is possible to add a @{FORCE_INDEX=...} as a statement hint, the
+            // only allowed value is _BASE_TABLE. This means that we can safely assume that the statement
+            // hint will not contain any special characters, for example a closing curly brace or one of the
+            // keywords SELECT, UPDATE, DELETE, WITH, and that we can keep the check simple by just
+            // searching for the first occurrence of a keyword that should be preceded by a closing curly
+            // brace at the end of the statement hint.
+            int startStatementHintIndex = sql.IndexOf('{');
```
amanda-tarafa, line 444: What if the curly bracket is in a string literal?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -276,5 +278,190 @@ private static string ValidateTableName(string table, string parameterName)
...
+            string upperCaseSql = sql.ToUpperInvariant();
+            foreach (string keyword in QueryAndDmlStatements) {
+                startQueryIndex = upperCaseSql.IndexOf(keyword);
+                if (startQueryIndex > -1) {
+                    break;
+                }
+            }
+            if (startQueryIndex > -1) {
+                int endStatementHintIndex = sql.Substring(0, startQueryIndex).LastIndexOf('}');
+                if (startStatementHintIndex == -1 || startStatementHintIndex > endStatementHintIndex) {
```
amanda-tarafa, line 456: This is not if they are in string literals.

amanda-tarafa (changes requested):
I've left some code specific comments on the PR.
But I'll leave my general view on this change on the issue #6847 .

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: That is true, but that is not a command that supports comments or hints anyways. That is a pseudo command text that indicates that the command should use a mutation and not a DML statement, and mutations do not support comments or hints.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: TBH, I'm open to calling it anything =) But formally this is the number sign, _hash_ (so certainly not dash...) or pound sign, while the sharp sign is slightly different: https://en.wikipedia.org/wiki/Number_sign and https://en.wikipedia.org/wiki/Sharp_(music)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: I assume you mean `ArgumentException` (which is thrown by `FromCommandText`), so I've changed it to that.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: See also the note above: Those commands are converted to mutations, so the command text is never sent to Spanner.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: See also above. If the opening curly brace is in a string literal, it is guaranteed to be after the first keyword, which means that the substring from line 455 will only contain the statement keyword and never a closing curly brace. That means that the original sql string will be returned, but with the wrong explanation in the comment why that is.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -276,5 +279,194 @@ private static string ValidateTableName(string table, string parameterName)
...
+        /// <summary>
+        /// Removes any comments from the given sql string.
+        /// </summary>
+        /// <param name="sql">The sql string to strip for comments</param>
+        /// <returns>The sql string without any comments</returns>
+        /// <exception cref="ArgumentException">
+        /// Throws ArgumentException if the sql string contains an unclosed
+        /// literal or unclosed quoted identifier
+        /// </exception>
+        internal static string RemoveComments(string sql)
```
amanda-tarafa, line 300: OK, I think we can make this simpler as well:
- We have 2 kinds of comments, line comments, starting with `--` for instance, and block comments `/* */`. And we only care about those at the beginning of the command.

In a loop:
- If the command string starts with any of the line comment starters `--`, `#`, or others, remove the entire line and loop back.
- If the commnd string starts with any of the block comment starterts, `/*` (others?) just find the first comment block closer that's not escaped `*/`.
  -  If found, remove everything up to and including the closer `*/` and loop back again.
  - Else, return the original string and let the caller or the backend fail.
- Return what you have when it doesn't start with a line or block comment.

WDTY? Or am I missing something?

amanda-tarafa (commented):
OK, I'm proposing a few simplifications that I think are possible, given that we really just want to know the type of the command and that we can let the backend fail for ill formatted commands.

Hope it makes sense. Else I can try a draft PR next week.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: Good question. In this case the backend will not fail it, as the command text is only used to determine the type of command. The generated mutation will not contain the hint or comment, and will therefore be accepted by the backend. For comments, I would say that it's not really a problem. It basically means that we are introducing support for comments in these mutation command texts. For hints it might be better to fail in the client, as they will otherwise just silently be stripped away without any use.
Note: These mutation command texts is a .NET specific feature. Cloud Spanner (and also the client libraries for the other programming languages) do not support a command text like `INSERT <table_name>` or `UPDATE <table_name>`.

> Is EF Core generating mutations with comments?
No.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: I think this will break because line comments can also start in the middle of a line, which would not be stripped by such a simplified approach. And I worry that those 'inline line comments' that would be left in the command string could break the rest of the parsing. So the following kind of statements would probably not be parsed correctly:

```sql
@{
  OPTIMIZER_VERSION=1 -- Use the first optimizer for now. Change this to @{OPTIMIZER_VERSION=2} once XYZ is done.
}
SELECT *
FROM Singers

@{OPTIMIZER_VERSION=1} -- This is a comment /* this is not a block comment
SELECT *
FROM Singers
/* This is a block comment */
```

The current statement hint parsing is simplified to just look for the first and next curly brace, but that can only be safely done because we at that moment know that:
1. String literals cannot occur in that part of a valid SQL string.
2. All comments in that part of the SQL string have been removed.

> `/*` (others?)

No, `/*` is the only valid start of a block comment.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: You're right, this can be done a lot simpler, especially when we assume that the statement is valid and well-formed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
```diff
@@ -146,10 +174,65 @@ public void DmlDeleteCommand(string commandText)
...
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE}\nSELECT 1", "SELECT 1")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE}\tSELECT 1", "SELECT 1")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE}SELECT 1", "SELECT 1")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE, OPTIMIZER_VERSION=2} SELECT 1", "SELECT 1")]
+        [InlineData("SELECT '@{OPTIMIZER_VERSION=1}'", "SELECT '@{OPTIMIZER_VERSION=1}'")]
+        [InlineData("SELECT '@{OPTIMIZER_VERSION=1} SELECT 1'", "SELECT '@{OPTIMIZER_VERSION=1} SELECT 1'")]
+        [InlineData("SELECT * FROM Foo@{FORCE_INDEX=`INDEX FOR SELECT`}", "SELECT * FROM Foo@{FORCE_INDEX=`INDEX FOR SELECT`}")]
+        [InlineData("@{OPTIMIZER_VERSION=1 SELECT 1", "@{OPTIMIZER_VERSION=1 SELECT 1")]
+        [InlineData("{OPTIMIZER_VERSION=1} SELECT 1", "{OPTIMIZER_VERSION=1} SELECT 1")]
+        public void RemoveStatementHintAndComments(string sql, string sqlWithoutHintsAndComments)
```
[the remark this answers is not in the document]
olavloite: You're right. I've updated the test name, as we are already testing the stripping of comments above. I also changed the name of the method that removes comments and statement __hints__, as there could be more than one.

jskeet:
Presumably this PR can be closed now that #6947 has been merged?

olavloite:
> Presumably this PR can be closed now that #6947 has been merged?

Yes

PR #6876, opened 2021-08-03
feat: regenerate Compute to take advantage of DiREGapic Pagination

viacheslav-rostovtsev:


jskeet:
I'll have a look at this on Thursday. Of course, if we're happy with it, we'll merge the generator and let autosynth make the actual change in this repo.

apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1/NodeTypesClient.g.cs
```diff
@@ -571,23 +575,57 @@ public override NodeType Get(GetNodeTypeRequest request, gaxgrpc::CallSettings c
...
             Modify_ListNodeTypesRequest(ref request, ref callSettings);
-            return _callList.Async(request, callSettings);
+            return new gaxgrpc::GrpcPagedAsyncEnumerable<ListNodeTypesRequest, NodeTypeList, NodeType>(_callList, request, callSettings);
         }
     }
+
+    public partial class AggregatedListNodeTypesRequest : gaxgrpc::IPageRequest
+    {
+        /// <inheritdoc/>
+        public int PageSize
```
jskeet, line 601: As noted on the generator PR, I think this would be better using explicit interface implementation:

```csharp
gaxgrpc::IPageRequest int PageSize
```

That way it doesn't show up as publicly - it's odd to have a public set-only property. If we want to actually have a public property, it's probably worth including a getter as well.

jskeet (commented):
We won't be merging this PR anyway, but just one comment on the PageSize property.

apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1/NodeTypesClient.g.cs
viacheslav-rostovtsev: I've updated the corresponding comment on the generator PR (https://github.com/googleapis/gapic-generator-csharp/pull/344) to keep the discussion in one place

PR #6877, opened 2021-08-03
feat(compat): Support custom attributes in generic parameters.

amanda-tarafa:
For nullability support.

Mono.Cecil introduced a breaking change in https://github.com/jbevain/cecil/pull/594 because of https://github.com/dotnet/roslyn/issues/29997


amanda-tarafa:
@jskeet This should fix the errors we've been seing on the Diff Checker.

There's one thing I don't understand though, and it's the value of the `#nullable` directive. If I set it to `enable` the test that I added does not pass, I've debugged and and with `#nullable enable` the nullable attribute is added to both generic parameters.

jskeet (approved):
I don't fully understand the nullable part either - my *guess* is that in both cases the NullableAttribute is applied, but with different values. I can look at it another time if it ever becomes the most useful thing to spend time on, but I can't see that happening imminently.

PR #6879, opened 2021-08-04
build: Diagnostics tests dependencies

amanda-tarafa:
Closes #6878

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/GenerateProjectsCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/GenerateProjectsCommand.cs
@@ -249,6 +249,21 @@ namespace Google.Cloud.Tools.ReleaseManager
             {
                 ["extends"] = new JArray { "config:base" },
                 ["ignorePaths"] = new JArray(catalog.Apis.Select(api => $"apis/{api.Id}/{api.Id}/**").ToArray()),
+                ["packageRules"] = new JArray
+                {
+                    new JObject
+                    {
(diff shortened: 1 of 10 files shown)


tools/Google.Cloud.Tools.ReleaseManager/GenerateProjectsCommand.cs
```diff
@@ -249,6 +249,21 @@ public static void RewriteRenovate(ApiCatalog catalog)
             {
                 ["extends"] = new JArray { "config:base" },
                 ["ignorePaths"] = new JArray(catalog.Apis.Select(api => $"apis/{api.Id}/{api.Id}/**").ToArray()),
+                ["packageRules"] = new JArray
```
jskeet, line 252: Hmm... I can see this getting messy. Let's keep it for now, but maybe we should make the update just load the existing JSON and update the ignorePaths property of it, leaving everything as it is. If that sounds good to you, I can do that later on.

jskeet (approved):
We could quite possibly just exclude the ASP.NET Core dependencies in general, but this works too.

tools/Google.Cloud.Tools.ReleaseManager/GenerateProjectsCommand.cs
amanda-tarafa: Yes agreed, I mostly want to check that this will stop the renovate PRs from generating. Then, or when this part gets messier, we can do differently.

amanda-tarafa (commented):
I'm looking at the falures now.

jskeet (approved):
Still looking good.

PR #6881, opened 2021-08-04
chore: Fix the analyzer

jskeet:
Renovate didn't take the target framework into account when proposing changes :(

This commit updates everything and fixes a couple of Roslyn
warnings; it disables one warning suggesting a release tracker (we
don't release this as a package) and another which would take
significant extra work to fix and doesn't seem to be causing a
problem.


amanda-tarafa (approved):
LGTM

PR #6883, opened 2021-08-04
Remove the third_party breaking change detector

jskeet:
We don't use this code, and keeping it up-to-date would take effort.

(We can always revert this commit if we want to get it back, of course.)


amanda-tarafa (approved):
LGTM

PR #6898, opened 2021-08-06
tests(compute): Test surfacing meaningful error messages.

amanda-tarafa:
To be merged after regenerating the library with changes included in https://github.com/googleapis/gax-dotnet/pull/488. Else this test will fail.


amanda-tarafa:
I've set the `do not merge` label so we don' accidentally merge before the Compute library is regenerated with the appropiate changes.

jskeet (approved):

PR #6899, opened 2021-08-06
DO NOT MERGE: prototype of Compute LRO handling

jskeet:
cc @viacheslav-rostovtsev

Explanatory email internally.

PR #6903, opened 2021-08-09
DO NOT MERGE: Second prototype of Compute LROs

jskeet:
This is broadly equivalent to #6899 (which it depends on) in terms of public API, but with a lower-level adaptation layer.


apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1/InstancesClient.g.cs
```diff
@@ -6147,4 +6228,20 @@ public override Operation UpdateShieldedInstanceConfig(UpdateShieldedInstanceCon
             return _callUpdateShieldedInstanceConfig.Async(request, callSettings);
         }
     }
+
+    public static partial class Instances
+    {
+        public partial class InstancesClient
+        {
+            // TODO: This design assumes that every client is exactly one of zonal, regional or global.
```
amanda-tarafa, line 6236: But aint there a way to determine the type of the operation "dynamically", and by that I mean to have the operation inspected at runtime, instead of generating fixed clients for each?
So maybe we can have an implementation of `lro::Operations.OperationsClient` which internally really has three `OperationsClient` and who determines to which of the three to send it to. Something like an `OperationsClientDispatcher`.
If I'm not explaining myself well, I can give that a try on another PR.

amanda-tarafa (commented):
Just one comment.

apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1/InstancesClient.g.cs
jskeet: I did wonder about that - but each method really does "know" what kind of operation it's dealing with, so we don't actually need to do anything dynamic.

We could reasonably easily create a single OperationsClient with a single CallInvoker, where the CallInvoker "knows" how to create different operation types into a Compute operation. I even considered creating an interface for specific request types, so we'd just need to make all of the requests implement the appropriate interface (via partial classes). For the moment, I think this is simpler (less new stuff to generate) but we could change from one to the other without any breaking changes later.

amanda-tarafa: Hmmm, I like the interface for each type of operation. But agreed that any of it can be done without breaking later, so I'm fine with this as is.

jskeet:
Closing this in favor of the more complete #6952.

PR #6920, opened 2021-08-10
Workaround for autosynth squashing lines in commit messages

jskeet:
This fetches the original commit from googleapis instead.

Note that this does mean if the original commit message had been
tailored when squashing, we won't spot that. We can work out a way
around that when it next comes up.


amanda-tarafa (approved):
LGTM

PR #6930, opened 2021-08-10
Stop using labeled/unlabeled as a trigger for diff-pr

jskeet:
This doesn't play nicely with the CLA bot.

After applying the label, we'll just need to manually run the diff
job again. We could potentially trigger it with an issue comment,
but that looks at least mildly tricky.


amanda-tarafa (approved):
LGTM

PR #6936, opened 2021-08-10
Really don't run diff-pr on label changes

jskeet:
We need to be able to filter out changes other than "allow breaking
changes" in the trigger - doing it in the "if" is too late, as:

- it comes after permissions, so the CLA bot still triggers it
- it ends up causing the check to show as skipped, rather than just not triggering it


amanda-tarafa (approved):
LGTM

PR #6947, opened 2021-08-11
feat(Spanner): Support comments and statement hints in untyped commands.

### LINKED ISSUE #6847: [Spanner]: Statements that start with a comment or statement hint are not recognized

amanda-tarafa:
Alternative to #6848
Closes #6847

My approach here is to do the minimum necessary to be able to determine the command type, which is what we really need. Proper command correctness will be done by the backend. This is in contrast to #6848 which is more about completeness and correctness when it comes to comment and hint parsing. I prefer my approach as I believe is less prone to corner cases and easier to maintain, but I'm happy to go with #6848 if this does not seem enough.

My approach/assumptions:

- I assume the command is well formed. If it's not, the backend (or in some cases us) will fail.
- I'm only removing leading comments and hints, I don't look at what's after the start of the command. The advantage this has is that we don't care about literal strings.
- For a mutation, if it had leading comments or hints, I throw. Neither are supported by the backend and the alternative we have is to silently remove them, which I think will lead to unexpected results for users that attempted to use them, hints in particular.
  - Note that if a "mutation" has comments or hints after the start of the command (also not supported by the backend) it was being and will continue to be recognized as DML, which will fail in the backend.
- This passes the same tests as #6848, and I've added a few more on the SELECT theory to check for alternating comments and hints.

@olavloite there's a TODO there about escaping special characters in comments, I don't know if it's possible and which is the escape character, but that's an easy check anyway.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
@@ -77,6 +77,25 @@ namespace Google.Cloud.Spanner.Data.Tests
         [InlineData("\t\nWITH AlbumsView AS (SELECT * FROM Albums) select Title from AlbumsView")]
         [InlineData(" WITH   AlbumsView AS (SELECT * FROM Albums) select Title from AlbumsView")]
         [InlineData("WITH\t\nAlbumsView AS (SELECT * FROM Albums) select Title from AlbumsView")]
+        [InlineData("-- Single line comment\nSELECT * FROM Albums")]
+        [InlineData("# Single line comment\nSELECT * FROM Albums")]
+        [InlineData("/* Multi\nline\ncomment\n */\nSELECT * FROM Albums")]
+        [InlineData("/* Multi\n * line\n * comment\n */\nSELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} SELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} \n-- Single line comment\nSELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} \n# Single line comment\nSELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} /* Multi\nline\ncomment\n */\nSELECT * FROM Albums")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} /* Multi\n * line\n * comment\n */\nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("# Single line comment\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("/* Multi\nline\ncomment\n */\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("/* Multi\n * line\n * comment\n */\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE, OPTIMIZER_VERSION=1} \n\t \nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE, OPTIMIZER_VERSION=1} -- Another single line comment \n\t @{FORCE_INDEX=_BASE_TABLE, OPTIMIZER_VERSION=1} /* Multi\n * line\n * comment\n */\nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE, /* Multi\n * line\n * comment\n */ OPTIMIZER_VERSION=1} \n\t \nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE, -- Inner single line comment\n OPTIMIZER_VERSION=1} \n\t \nSELECT * FROM Albums")]
+        [InlineData("SELECT/*comment*/ * FROM Albums")]
+        [InlineData("SELECT--comment\n * FROM Albums")]
         public void SelectCommand(string commandText)
         {
             var builder = SpannerCommandTextBuilder.FromCommandText(commandText);
@@ -95,6 +114,10 @@ namespace Google.Cloud.Spanner.Data.Tests
         [InlineData("\t\nINSERT\t\nAlbums\t\n(AlbumId)t\nVALUESt\n(@id)")]
         [InlineData("INSERT  Albums  (AlbumId) VALUES (@id)")]
         [InlineData("INSERT\t\nAlbums\t\n(AlbumId)t\nVALUESt\n(@id)")]
+        [InlineData("-- Single line comment\nINSERT INTO Albums (AlbumId) VALUES (@id)")]
+        [InlineData("# Single line comment\nINSERT INTO Albums (AlbumId) VALUES (@id)")]
+        [InlineData("/* Multi line comment */ INSERT INTO Albums (AlbumId) VALUES (@id)")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} INSERT INTO Albums (AlbumId) VALUES (@id)")]
         public void DmlInsertCommand(string commandText)
         {
             var builder = SpannerCommandTextBuilder.FromCommandText(commandText);
@@ -108,6 +131,10 @@ namespace Google.Cloud.Spanner.Data.Tests
         [InlineData("\t\nUPDATE\t\nAlbumst\nSETt\nTitle=@titlet\nWHEREt\nTRUE")]
         [InlineData("UPDATE  Albums  SET Title=@title WHERE TRUE")]
         [InlineData("UPDATE\t\nAlbums\t\nSETt\nTitle=@titlet\nWHEREt\nTRUE")]
+        [InlineData("-- Single line comment\nUPDATE Albums SET Title=@title WHERE TRUE")]
+        [InlineData("# Single line comment\nUPDATE Albums SET Title=@title WHERE TRUE")]
+        [InlineData("/* Multi line comment */ UPDATE Albums SET Title=@title WHERE TRUE")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} UPDATE Albums SET Title=@title WHERE TRUE")]
         public void DmlUpdateCommand(string commandText)
         {
             var builder = SpannerCommandTextBuilder.FromCommandText(commandText);
@@ -126,6 +153,10 @@ namespace Google.Cloud.Spanner.Data.Tests
         [InlineData("\t\nDELETE\t\nAlbumst\nWHEREt\nTRUE")]
         [InlineData("DELETE  Albums  WHERE TRUE")]
         [InlineData("DELETE\t\nAlbumst\nWHEREt\nTRUE")]
+        [InlineData("-- Single line comment\nDELETE FROM Albums WHERE TRUE")]
+        [InlineData("# Single line comment\nDELETE FROM Albums WHERE TRUE")]
+        [InlineData("/* Multi line comment */ DELETE FROM Albums WHERE TRUE")]
+        [InlineData("@{FORCE_INDEX=_BASE_TABLE} DELETE FROM Albums WHERE TRUE")]
         public void DmlDeleteCommand(string commandText)
         {
             var builder = SpannerCommandTextBuilder.FromCommandText(commandText);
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -255,13 +255,117 @@ public static SpannerCommandTextBuilder FromCommandText(string commandText)
                     }
                     else
                     {
+                        if (removedCommentOrHint)
+                        {
+                            throw new ArgumentException(
+                                $"Spanner {commandName} commands may be specified as '{commandName} <table>' with " +
+                                "parameters added to customize the command with filtering or updated values. " +
+                                "In that case, the commands are sent to Spanner as mutations which don't support comments" +
```
jskeet, line 263: Nit: add space at the end of "comments".

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -255,13 +255,117 @@ public static SpannerCommandTextBuilder FromCommandText(string commandText)
...
+            return new SpannerCommandTextBuilder(commandText.Trim(), commandType, targetTable, extraStatements: null);
+        }
+
+        private static string RemoveLeadingCommentsAndHints(string commandText, out bool removedCommentOrHint)
+        {
+            try
+            {
+                return commandText.Substring(IndexOfCommandStart(commandText, out removedCommentOrHint));
+            }
+            catch(Exception ex) when (ex is ArgumentOutOfRangeException || ex is IndexOutOfRangeException)
```
jskeet, line 283: Mega-nit: add space after `catch`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -255,13 +255,117 @@ public static SpannerCommandTextBuilder FromCommandText(string commandText)
...
+
+            bool SkipLineComment()
+            {
+                if (command.IndexOf('#', currentIndex, 1) >= 0 ||
+                    command.IndexOf("--", currentIndex, 2) >= 0)
+                {
+                    // TODO: If this can be escaped, let's check that it's not
+                    // or continue searching.
+                    currentIndex = command.IndexOf('\n', currentIndex);
+                    if (currentIndex >= 0) // If we didn't find it, we don't fail here.
```
jskeet, line 319: I don't understand the comment here. I'd expect something like "If there are more lines, move to the start of the next line and continue".

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -255,13 +255,117 @@ public static SpannerCommandTextBuilder FromCommandText(string commandText)
...
+            }
+
+            bool SkipBlockComment()
+            {
+                if (command.IndexOf("/*", currentIndex, 2) >= 0 )
+                {
+                    // TODO: If this can be escaped, let's check that it's not
+                    // or continue searching.
+                    currentIndex = command.IndexOf("*/", currentIndex + 2);
+                    if (currentIndex >= 0) // If we didn't find it, we don't fail here.
```
jskeet, line 335: Again, presumably this is "Continue the search from the end of the comment"?

If we *don't* find the end of the comment, presumably now currentIndex is -1 and we'll throw an exception accidentally?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -255,13 +255,117 @@ public static SpannerCommandTextBuilder FromCommandText(string commandText)
...
+                        currentIndex += 2;
+                    }
+                    return true;
+                }
+
+                return false;
+            }
+            bool SkipHint()
+            {
+                if (command.IndexOf('@', currentIndex, 1) >= 0)
```
jskeet, line 346: This is just `if (command[currentIndex] == '@')` isn't it?

jskeet (commented):
I think it's fine to fail if the command isn't well-formed, but it should be with a clear exception. For example, I *suspect* that `"/* this comment is unterminated"` will fail with an unclear exception.

Other than that, I believe I understand this. Now to look at #6848.

jskeet:
Ooh, just one more thought. I believe this wouldn't cope with "UPDATE /* comment */ SomeTable" would it?

I suspect we could do that quite easily though:

- Remove leading comments
- Take the first word (letters and digits only?)
- Remove leading comments from the rest
- Split

Or I may have missed something.

amanda-tarafa:
> I believe this wouldn't cope with "UPDATE /* comment */ SomeTable" would it?

It wouldn't cope with it if it's a mutation. Comments and hints are not supported for mutations, which are not written or sent to the backend as SQL commands and "do not support SQL Syntax" as shown here https://cloud.google.com/spanner/docs/dml-versus-mutations. I think it's best if we don't silently strip comments and most importantly hints. In this case what will happen is that the command will be assumed to be DML and will be sent as such and the backend will fail, which is the same thing that happens without the changes in this PR.

This PR strips all leading hints, not sure if multiple hints are valid or not, but the backend will fail if multiple hints are not supported, we send the original comment.

My inclination is to not actively check validity of the command. There are some cases in which we will unavoidably fail, and that's fine, but I think we shouldn't try to validate "a few more cases". In a way adding support for comments and hints on the library just means that we need to be able to determine a command type even if it has leading comments and hints.

jskeet:
> My inclination is to not actively check validity of the command.

Agreed. But we don't want to end up treating invalid stuff as valid. I need to look through this code again, but there seems to be code that uses the second word in the SQL as a table name. I'm concerned we might get things wrong that way if we don't strip comments after the first word. I'll have a look tomorrow and try to come up with a more concrete example.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
```diff
@@ -77,6 +77,23 @@ public void DropDatabaseUnrecognized(string ddlString)
...
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("# Single line comment\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("/* Multi\nline\ncomment\n */\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("/* Multi\n * line\n * comment\n */\n@{FORCE_INDEX=_BASE_TABLE} \nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE, OPTIMIZER_VERSION=1} \n\t \nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE, OPTIMIZER_VERSION=1} -- Another single line comment \n\t @{FORCE_INDEX=_BASE_TABLE, OPTIMIZER_VERSION=1} /* Multi\n * line\n * comment\n */\nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE, /* Multi\n * line\n * comment\n */ OPTIMIZER_VERSION=1} \n\t \nSELECT * FROM Albums")]
+        [InlineData("-- Single line comment\n@{FORCE_INDEX=_BASE_TABLE, -- Inner single line comment\n OPTIMIZER_VERSION=1} \n\t \nSELECT * FROM Albums")]
         public void SelectCommand(string commandText)
```
olavloite, line 97: The following valid statements will fail, but it should be easily fixable in the check that searches for the keyword.

```suggestion
        [InlineData("SELECT/*comment*/ * FROM Albums")]
        [InlineData("SELECT--comment\n * FROM Albums")]
        public void SelectCommand(string commandText)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -255,13 +255,117 @@ public static SpannerCommandTextBuilder FromCommandText(string commandText)
...
+                    currentIndex++;
+                }
+            }
+
+            bool SkipLineComment()
+            {
+                if (command.IndexOf('#', currentIndex, 1) >= 0 ||
+                    command.IndexOf("--", currentIndex, 2) >= 0)
+                {
+                    // TODO: If this can be escaped, let's check that it's not
```
olavloite, line 316: It cannot be escaped, so we don't have to worry about that.

The only situation that we might need to worry about is that there is a statement hint that supports string literals (currently undocumented):

```sql
@{OPTIMIZER_STATISTICS_PACKAGE='auto_20191128_14_47_22UTC'}
SELECT *
FROM AlbumsWithVersion 
where AlbumId = 1
```

The above example is an auto-generated package name which will never contain any characters that will clash with comment or statement hint start or end characters, but if it would also support user-defined package names, we would need to skip string literals in the hints section as well.
(Note: #6848 also currently does not take this fully into account, and would break if a user-defined package name would contain a `}`)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
```diff
@@ -255,13 +255,117 @@ public static SpannerCommandTextBuilder FromCommandText(string commandText)
...
+                    return true;
+                }
+                return false;
+            }
+
+            bool SkipBlockComment()
+            {
+                if (command.IndexOf("/*", currentIndex, 2) >= 0 )
+                {
+                    // TODO: If this can be escaped, let's check that it's not
```
olavloite, line 332: The end of a block comment can also not be escaped, so we don't have to worry about that.

olavloite:
> > My inclination is to not actively check validity of the command.
> 
> Agreed. But we don't want to end up treating invalid stuff as valid. I need to look through this code again, but there seems to be code that uses the second word in the SQL as a table name. I'm concerned we might get things wrong that way if we don't strip comments after the first word. I'll have a look tomorrow and try to come up with a more concrete example.

The mutation commands treat the second word in the SQL as the table name, but those commands are .NET specific and not part of the Spanner SQL syntax. Those commands therefore also never supported comments anywhere in the command string before this change, and I don't think it's necessary to add support for that now.

amanda-tarafa:
Notice that since this PR is only looking at leading comments/hints, but not in the middle of the command, it is less likely to break existing behavior. Before this PR, untyped commands with comments/hints after the command type were being accepted by the library and validated by the backend and that will continue to be the case.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
amanda-tarafa: Yikes with those being valid! Funnily enough, without this PR, something like `SELECT/* comment */ * FROM myTable` will fail, but `SELECT /* comment */ * FROM myTable` will not.

I think the easiest way to account for this is to split the stripped command on `/` and `-` and `#` as well as blanks. WDYT?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
amanda-tarafa: This is just: we didn't find an end of line, so the whole thing is a comment, as we are skipping a line comment. currentIndex will be -1 and the next time we try to use it it will throw ArgumentOutOfRange or IndexOutOfRange which are caught by the caller to throw ArgumentException("unrecognized command").

I'm fine with throwing here something more explicit though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
amanda-tarafa: Same as above, This is just: we didn't find an end of block comment, so the whole thing is a comment. currentIndex will be -1 and the next time we try to use it it will throw ArgumentOutOfRange or IndexOutOfRange which are caught by the caller to throw ArgumentException("unrecognized command").

I'm fine with throwing here something more explicit though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
amanda-tarafa: @olavloite would you be OK to deferring that check to if and when that is supported? Presumable those package names, even if user defined, might never even accept a }?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
jskeet: Ah - I'd missed that we're already catching those exceptions. I think I'd *personally* prefer to remove that catch block and instead throw dedicated exceptions... I'm slightly worried that if we've got a bug in this code, we'll end up hiding it. We should be able to work out every case where we *would* end up throwing an exception, and doing so explicitly instead. (I don't mind if it throws an ArgumentException with "unrecognized command" as the message - it's more about doing it explicitly.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTextBuilderTests.cs
olavloite: > I think the easiest way to account for this is to split the stripped command on / and - and # as well as blanks. WDYT?

That, or just checking whether the command string starts with one of the known keywords. That would mean that a statement that starts with `SELECTINSERT` would be seen as a query (instead of being rejected by the client library), and then be handled by the backend.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommandTextBuilder.cs
olavloite: Yes, I'm fine with that.

amanda-tarafa (commented):
Addressed all comments, PTAL, thanks!

amanda-tarafa:
BTW, review addressed on second commit, which I'll squash before merging.

olavloite:
> Yes, I like this :) Let's check that @olavloite is happy, but then we can merge.

I'm happy with the latest version.

olavloite (approved):

amanda-tarafa:
Thanks!, I'll rebase both commits in one and merge after green.

PR #6950, opened 2021-08-12
Prototype of library changes for self-signed JWTs

jskeet:
This assumes:

- ChannelPool has a constructor with an additional Boolean parameter
- ClientBuilderBase has a protected Boolean property of UseJwtAccessWithScopes

The generated classes for both Speech and Shell have been modified
in the same way (indicating what we'd need to do in the generator).

Additionally, a manual partial class in SpeechClient shows what we'd
need to do in order to disable self-signed JWTs for a specific API.
(I'm assuming we *don't* have this information in the generator. If
we ever *do* know it in the generator, we can just generate the
right thing and remove the partial method.)

### DIFF
--- a/apis/Google.Cloud.Shell.V1/Google.Cloud.Shell.V1/CloudShellServiceClient.g.cs
+++ b/apis/Google.Cloud.Shell.V1/Google.Cloud.Shell.V1/CloudShellServiceClient.g.cs
@@ -212,6 +212,12 @@ namespace Google.Cloud.Shell.V1
 
         partial void InterceptBuildAsync(st::CancellationToken cancellationToken, ref stt::Task<CloudShellServiceClient> task);
 
+        /// <summary>Creates a new builder with default settings.</summary>
+        public CloudShellServiceClientBuilder()
+        {
+            UseJwtAccessWithScopes = CloudShellServiceClient.UseJwtAccessWithScopes;
+        }
+
         /// <summary>Builds the resulting client.</summary>
         public override CloudShellServiceClient Build()
         {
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.Shell.V1/Google.Cloud.Shell.V1/CloudShellServiceClient.g.cs
```diff
@@ -287,7 +293,19 @@ public abstract partial class CloudShellServiceClient
...
+        {
+            get
+            {
+                bool useJwtAccessWithScopes = true;
+                MaybeUseJwtAccessWithScopes(ref useJwtAccessWithScopes);
+                return useJwtAccessWithScopes;
+            }
+        }
+
+        static partial void MaybeUseJwtAccessWithScopes(ref bool useJwtAccessWithScopes);
```
amanda-tarafa, line 308: +1 to this instead of an interceptor for initializing another ChannelPool.

apis/Google.Cloud.Shell.V1/Google.Cloud.Shell.V1/CloudShellServiceClient.g.cs
```diff
@@ -287,7 +293,19 @@ public abstract partial class CloudShellServiceClient
             "https://www.googleapis.com/auth/cloud-platform",
         });
 
-        internal static gaxgrpc::ChannelPool ChannelPool { get; } = new gaxgrpc::ChannelPool(DefaultScopes);
+        internal static gaxgrpc::ChannelPool ChannelPool { get; } = new gaxgrpc::ChannelPool(DefaultScopes, UseJwtAccessWithScopes);
+
+        internal static bool UseJwtAccessWithScopes
```
amanda-tarafa, line 298: Maybe do this a Lazy, since we use it for ChannelPool initialization once, and then several times on the ClientBuilder, we don't want the interceptor to change its mind about using JWTs.

amanda-tarafa (commented):
Yep, this is great. I just left one comment.

apis/Google.Cloud.Shell.V1/Google.Cloud.Shell.V1/CloudShellServiceClient.g.cs
jskeet: It never would, unless someone wrote a really odd partial method. (This is all internal or private.)

jskeet:
cc @TimurSadykov - I'm now working on the generator change for this.

jskeet:
Closing now this is implemented in #6965

TimurSadykov:
@jskeet Many thanks!

PR #6965, opened 2021-08-17
Update and regenerate all APIs to use self-signed JWTs

jskeet:
Currently this leaves Compute using self-signed JWTs. That appears to work, but we need more confirmation before the next release.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
@@ -177,13 +177,6 @@ namespace Google.Cloud.Bigtable.V2
 
     public sealed partial class BigtableServiceApiClientBuilder : ClientBuilderBase<BigtableServiceApiClient>
     {
-        /// <summary>
-        /// Creates a new instance with no settings.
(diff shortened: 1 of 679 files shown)


tools/Google.Cloud.Tools.ReleaseManager/UpdateDependencies.cs
```diff
@@ -0,0 +1,45 @@
+﻿// Copyright 2020 Google LLC
```
amanda-tarafa, line 1: tiny nit: 2021 (not sure why I can add that as a suggestion)

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -234,17 +227,18 @@ protected override async Task<CallInvoker> CreateCallInvokerAsync(CancellationTo
             }
         }
 
-        /// <inheritdoc />
-        protected override bool CanUseChannelPool => false;
+        // Note: GetChannelPool() will still
```
amanda-tarafa, line 230: nit: incomplete note

amanda-tarafa (approved):
LGTM

jskeet:
Okay, with the fix in #6967 the diff should pass...

PR #6968, opened 2021-08-17
fix: Add missing return on compat checker.

amanda-tarafa:


jskeet:
Whoops, hadn't seen you were doing this as well. It actually needed to be continue rather than return. Have self-merged - will close this one.

PR #6984, opened 2021-08-18
Add a comment row between "real" rows in the README table

jskeet:
This should avoid conflicts when multiple APIs are released concurrently.


jskeet:
Humbug - this looks awful. Closing for now...

PR #7020, opened 2021-08-19
Tidy up output for batch releasing

jskeet:
The change to the push format is primarily to make it all fit on one line in a default-width terminal


amanda-tarafa (approved):
LGTM

PR #7025, opened 2021-08-19
Update Timed Buffer guidance to include mention of new --no-cpu-throttling flag

EatonZ:
[Cloud Run is getting unthrottled CPU support.](https://twitter.com/kelseyhightower/status/1428485756423602179)

This effectively solves https://github.com/googleapis/google-cloud-dotnet/issues/5275. This PR adds a small note to the documentation.

This flag is currently in alpha. It may be some time before GA. Feel free to keep this PR open if you want to merge it when the feature is GA.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/docs/index.md
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/docs/index.md
@@ -25,7 +25,7 @@ the deployed code.
 # Note
 When running on environments that limit or disable CPU usage for background activities, for instance
 [Google Cloud Run](https://cloud.google.com/run/docs/tips/general#avoiding_background_activities), take care
-not to use the timed buffer options for any of Logging, Tracing or Error Reporting. Take into account
+not to use the timed buffer options for any of Logging, Tracing or Error Reporting unless you deploy your container using Cloud Run's `--no-cpu-throttling` flag. Take into account
 that the timed buffer is used for all of these components by default so you will need to explicitly
 configure the buffers by using the `Google.Cloud.Diagnostics.AspNetCore.LoggerOptions`,
 `Google.Cloud.Diagnostics.Common.TraceOptions` and `Google.Cloud.Diagnostics.Common.ErrorReportingOptions` classes.


apis/Google.Cloud.Diagnostics.AspNetCore/docs/index.md
```diff
@@ -25,7 +25,7 @@ the deployed code.
 # Note
 When running on environments that limit or disable CPU usage for background activities, for instance
 [Google Cloud Run](https://cloud.google.com/run/docs/tips/general#avoiding_background_activities), take care
-not to use the timed buffer options for any of Logging, Tracing or Error Reporting. Take into account
+not to use the timed buffer options for any of Logging, Tracing or Error Reporting unless you deploy your container using Cloud Run's `--no-cpu-throttling` flag. Take into account
```
amanda-tarafa, line 28: What about something like:
```suggestion
not to use the timed buffer options for any of Logging, Tracing or Error Reporting. Also note that some of these environmets allow for background CPU usage through configuration, which might allow you to use time buffer. For Google Cloud Run you will find configuration options here.... Take into account
```

amanda-tarafa (commented):
@EatonZ thanks for flagging this.

We don't usually explictlly add docs from other produts in library docs, just so that we don't have to play catch-up. But we can add a link to Cloud Runf config docs and a note similar to the one I'm proposing.

Would that work you?

apis/Google.Cloud.Diagnostics.AspNetCore/docs/index.md
EatonZ: We can't add a link right now because the only source of documentation is on the alpha page, and it will need to be changed again later when it reaches GA. Your suggestion sounds good, but I think it's better to mention the option by name. Something like: _For Google Cloud Run you can use `--no-cpu-throttling`._

Alternatively, I believe the public preview is about to come out (since that tweet basically revealed the feature). Maybe we should wait a little so a link can added that goes into the real Cloud Run product docs. It'll still be listed as preview, but it'll be a more permanent link describing the feature, and not just a link to the gcloud cli option.

amanda-tarafa: OK, let's wait for a public preview link before merging.
And I'll chat internally with the rest of the team about mentioning the config option by name. I'm still hesitant, tbh, I wouldn't like us to explicitly recommend an option that may have other implications for user's applications, or that may interact with other current, or most importantly, future config options. I don't want us to have to either become "experts" on Cloud Run config or have to be playing catchup with them, their config features and their docs.

Marking this PR as `do not merge` for the time being.

EatonZ: Since you mention Cloud Run by name I figured it wouldn't hurt to also mention a flag to make it works better with this library.

Just trying to save someone else the hassle of figuring out the problems I ran into in https://github.com/googleapis/google-cloud-dotnet/issues/5275 (-:

EatonZ: @amanda-tarafa The feature went into public preview today. [Announcement](https://cloud.google.com/blog/products/serverless/cloud-run-gets-always-on-cpu-allocation)

There is also now a docs location that can be linked to: https://cloud.google.com/run/docs/configuring/cpu-allocation

amanda-tarafa: Thanks for the heads up. I'm working on rewriting some of the Diagnostics docs and I'll include this. I'll leave this PR open until then as a reminder.

PR #7043, opened 2021-08-27
tests(Diagnostics): Improvements on entry polling

amanda-tarafa:
This is to try and avoid some flakiness.

These changes have a twofold effect, on one side we are waiting more if needed, on the other we attempt to find what we are looking for sooner so we wait less if we find it.
There's still one test that might need tweaking (increase waiting times), that is Trace_QPS, but I'll let it run as is for a while and see what happens.

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/BaseEntryPolling.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/BaseEntryPolling.cs
@@ -25,10 +25,10 @@ namespace Google.Cloud.Diagnostics.Common.IntegrationTests
     internal abstract class BaseEntryPolling<T>
     {
         /// <summary>Default total time to spend sleeping when looking for entries.</summary>
-        private static readonly TimeSpan _defaultTimeout = TimeSpan.FromSeconds(10);
+        private static readonly TimeSpan _defaultTimeout = TimeSpan.FromMinutes(10);
 
         /// <summary>Default time to sleep between checks for entries.</summary>
-        private static readonly TimeSpan _defaultSleepInterval = TimeSpan.FromSeconds(5);
+        private static readonly TimeSpan _defaultSleepInterval = TimeSpan.FromSeconds(30);
 
         /// <summary>Total time to spend sleeping when looking for entries.</summary>
         private readonly TimeSpan _timeout;
@@ -36,14 +36,17 @@ namespace Google.Cloud.Diagnostics.Common.IntegrationTests
         /// <summary>Time to sleep between checks for entries.</summary>
         private readonly TimeSpan _sleepInterval;
 
+        /// <summary>
+        /// Construct a new BaseEntryPolling instance.
+        /// </summary>
         /// <param name="timeout">Optional, total time to spend sleeping when looking for entries.
-        ///     Defaults to <see cref="_defaultTimeout"/>.</param>
+        /// Defaults to <see cref="_defaultTimeout"/>.</param>
         /// <param name="sleepInterval">Optional, time to sleep between checks for entries.
-        ///     Defaults to <see cref="_defaultSleepInterval"/></param>
-        protected BaseEntryPolling(TimeSpan timeout = default, TimeSpan sleepInterval = default)
+        /// Defaults to <see cref="_defaultSleepInterval"/></param>
+        protected BaseEntryPolling(TimeSpan? timeout = null, TimeSpan? sleepInterval = null)
         {
-            _timeout = timeout == default ? _defaultTimeout : timeout;
-            _sleepInterval = sleepInterval == default ?  _defaultSleepInterval : sleepInterval;
+            _timeout = timeout ?? _defaultTimeout;
+            _sleepInterval = sleepInterval ?? _defaultSleepInterval;
         }
 
         /// <summary>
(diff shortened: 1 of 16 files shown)


apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/BaseEntryPolling.cs
```diff
@@ -55,22 +58,30 @@ protected BaseEntryPolling(TimeSpan timeout = default, TimeSpan sleepInterval =
...
+                }
 
-                lastFoundEntries = getEntries().ToList();
-                if (lastFoundEntries.Count >= minEntries)
+                if ((sleptTime += _sleepInterval) < _timeout)
                 {
-                    return lastFoundEntries;
+                    Thread.Sleep(_sleepInterval);
+                }
+                else if (minEntries == 0)
```
jskeet, line 75: If minEntries is 0 and there are no entries, we'll just sleep and then return the previously-fetched list anyway. Do we need that sleep? If not, we can simplify the check earlier, and remove this else block.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/Logging/LogEntryPolling.cs
```diff
@@ -59,12 +48,14 @@ internal LogEntryPolling(TimeSpan timeout = default)
...
         ///     entries.</param>
         /// <param name="minSeverity">The minimum severity a log can be.</param>
-        public IEnumerable<LogEntry> GetEntries(DateTime startTime, string testId, int minEntries, LogSeverity minSeverity)
+        public IEnumerable<LogEntry> GetEntries(DateTimeOffset startTime, string testId, int minEntries, LogSeverity minSeverity)
         {
             return GetEntries(minEntries, () =>
             {
                 // Convert the time to RFC3339 UTC format.
-                string time = XmlConvert.ToString(startTime, XmlDateTimeSerializationMode.Utc);
+                // We substract 5 minutes because on ocasion we were polling too fast after startTime
```
jskeet, line 56: Nit: ocasion => occasion

jskeet (approved):
Fine except for one query with minEntries==0.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/BaseEntryPolling.cs
amanda-tarafa: In many cases we use minEntries == 0 to test that no entries were written (because of QPS for instance). Since there's an SLA for entry freshness we really should sleep some to check that we really didn't write any, instead of just not seeing any because they are not available yet.
Last I checked SLA was 10 seconds for 95% of entries and 5 minutes for 99%. We are waiting for 2 minutes in these cases

jskeet: But we're waiting for 2 minutes *and then returning the list from 2 minutes earlier* unless I've missed something.

I do wonder whether for the "no entries" polling we shouldn't just use a different abstraction - it feels like this doesn't really fit.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/Logging/LogEntryPolling.cs
amanda-tarafa: Done

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/BaseEntryPolling.cs
amanda-tarafa: I don't think we are doing that:

In general, how it works for minEntries == 0, we poll at intervals, and as soon as we find at least one entry, we return that. Else, when we reach timeout, instead of throwing, we return an empty list.

Implemented on a do while:
- Fetch the entries
- If we can sleep withing the timeout, sleep -> loop again
- If we couldn't sleep, and we were expecting no entries, we return the list that we just fetch (we couldn't sleep so that list is fresh), which is empty anyways, else, we would have returned it earlier.
- Else we throw as we dind't find as many entries as we were expecting on the timeout.

Or is it me who is missing something?

I did think about separating the minEntries out but it seemed a little bit of a hassle. But I'm happy to reconsider.

jskeet: Ah... I'd ignored that the previous block would be executed for the retry situation. Doh.

Yes, this works - even though it's a little ugly IMO. Fundamentally "poll until a condition is met, with a timeout" and "poll until a timeout, then check a condition" aren't the same thing, and could do with being separated IMO. But that can definitely be done later. Maybe add a TODO at this point?

jskeet (approved):

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common.IntegrationTests/BaseEntryPolling.cs
amanda-tarafa: Yep, I agree it's ugly. Added a TODO for consideration when we revisit these.

amanda-tarafa (commented):
Thanks! I'll merge on green.

PR #7111, opened 2021-09-02
feat: add smoke tests for Language

viacheslav-rostovtsev:


apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.IntegrationTests/SmokeTest.cs
```diff
@@ -0,0 +1,54 @@
...
+using Google.Cloud.ClientTesting;
+using Grpc.Core;
+using Xunit;
+
+namespace Google.Cloud.Language.V1.IntegrationTests
+{
+    public sealed class SmokeTest : CloudProjectFixtureBase
+    {
+        [Fact]
+        public void AnalyzeSentiment_Language()
```
jskeet, line 24: It's hard to tell what the `_Language` means here - I'd change it to `_DefaultLanguage` to make it clearer.

apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.IntegrationTests/SmokeTest.cs
```diff
@@ -0,0 +1,54 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.ClientTesting;
+using Grpc.Core;
+using Xunit;
+
+namespace Google.Cloud.Language.V1.IntegrationTests
+{
+    public sealed class SmokeTest : CloudProjectFixtureBase
```
jskeet, line 21: Rather than just `SmokeTest` (which are usually handled in a SmokeTests project or via a `smoketests.json` file), could you specify what it is you're trying to test with this? We already have a `smoketests.json` file that does what I'd normally consider a smoke test: does the library work at all, can it make a successful request.

It looks like you're really trying to test the error handling and defaulting - in which case I think it would be clearer to say so.

jskeet:
If this PR is aimed at extended error retrieval, could it potentially be closed in favour of #7153?

jskeet:
Closing this now that we have extended error retrieval (unreleased) in GAX. After the next GAX release, I'll create an integration test using Language (so a bit like this).

PR #7145, opened 2021-09-07
Fix docs for releases

jskeet:
Once this PR is merged, we can re-release everything in #7144 (without retagging).


amanda-tarafa (approved):
LGTM

PR #7153, opened 2021-09-10
Demo of retrieving error details from RpcException

jskeet:
The exact set of methods we provide is definitely up for discussion.

We *could* create a class that contains the exception's code, message, reason, domain and metadata in one bundle if we wanted.

### DIFF
--- a/apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/LanguageServiceClientSnippets.cs
+++ b/apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/LanguageServiceClientSnippets.cs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Api.Gax.Grpc;
 using Google.Cloud.ClientTesting;
+using Google.Rpc;
+using Grpc.Core;
 using System;
 using Xunit;
 using static Google.Cloud.Language.V1.AnnotateTextRequest.Types;
@@ -125,6 +128,32 @@ namespace Google.Cloud.Language.V1.Snippets
                 Console.WriteLine($"  Sentiment score: {entity.Sentiment.Score}; magnitude: {entity.Sentiment.Magnitude}");
             }
             // End sample
+        }
+
+        [Fact]
+        public void ErrorDetails()
+        {
+            LanguageServiceClient client = LanguageServiceClient.Create();
+
+            var document = new Document
+            {
+                Content = "Hello, world!",
+                Type = Document.Types.Type.PlainText,
+                Language = "zz"
+            };
+
+            var ex = Assert.Throws<RpcException>(() => client.AnalyzeSentiment(document));
+            Assert.Equal(StatusCode.InvalidArgument, ex.StatusCode);
+            Assert.Matches("document.language is not valid", ex.Status.Detail);
+
+            // An example of getting the bad request details. GetBadRequest will return null
+            // if there are no suitable details.
+            var badRequest = ex.GetBadRequest();
+            Console.WriteLine(badRequest?.FieldViolations);
+
+            // And likewise for ErrorInfo
+            var errorInfo = ex.GetErrorInfo();
+            Console.WriteLine(errorInfo?.Domain);
         }
     }
 }
--- a/apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/RpcExceptionExtensions.cs
+++ b/apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/RpcExceptionExtensions.cs
@@ -1,0 +1,81 @@
+﻿// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Protobuf;
+using Google.Protobuf.WellKnownTypes;
+using Google.Rpc;
+using Grpc.Core;
+using System.Linq;
+
+using Status = Google.Rpc.Status;
+
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// Utility extension methods to make it easier to retrieve extended error information from an <see cref="RpcException"/>.
+    /// </summary>
+    public static class RpcExceptionExtensions
+    {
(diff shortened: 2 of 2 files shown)


jskeet:
Note: RpcExceptionExtensions would be added to GAX, not to the language project - just to be clear :)

amanda-tarafa (approved):
LGTM

apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/RpcExceptionExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+using Status = Google.Rpc.Status;
+
+namespace Google.Api.Gax.Grpc
+{
+    /// <summary>
+    /// Utility extension methods to make it easier to retrieve extended error information from an <see cref="RpcException"/>.
+    /// </summary>
+    public static class RpcExceptionExtensions
+    {
+        public static Status GetRpcStatus(this RpcException ex) =>
```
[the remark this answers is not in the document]
jskeet: Hmm... not sure. GetRpcStatus? GetGoogleRpcStatus?

Definitely something to bike-shed on.

apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/RpcExceptionExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+            DecodeTrailer(GetTrailer(ex, "grpc-status-details-bin"), Status.Parser);
+
+        // TODO: This attempts to find the BadRequest in two different trailers. Is
+        // that appropriate?
```
summer-ji-eng, line 34: "The error_details will be sent on the wire as gRPC metadata with key grpc-status-details-bin" [link](https://g3doc.corp.google.com/net/grpc/g3doc/cpp/status_mapping.md?cl=head#details-on-conversion)
In Typescript and go is based on `grpc-status-details-bin` since it is documented in previous.  However, I tried ask a [YAQS question](https://yaqs.corp.google.com/eng/q/1466717725127606272) about this for gRPC team. But haven't received any answer yet.

apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/RpcExceptionExtensions.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                return null;
+            }
+            try
+            {
+                return parser.ParseFrom(entry.ValueBytes);
+            }
+            catch
+            {
+                // TODO: Is this appropriate?
```
summer-ji-eng, line 76: Yes. Similar thing did in Typescript and Go. Because we don't want to break the user if the type is unknown.

apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/LanguageServiceClientSnippets.cs
```diff
@@ -126,5 +129,31 @@ public void AnalyzeEntitySentiment()
...
+            Assert.Matches("document.language is not valid", ex.Status.Detail);
+
+            // An example of getting the bad request details. GetBadRequest will return null
+            // if there are no suitable details.
+            var badRequest = ex.GetBadRequest();
+            Console.WriteLine(badRequest?.FieldViolations);
+
+            // And likewise for ErrorInfo
+            var errorInfo = ex.GetErrorInfo();
+            Console.WriteLine(errorInfo?.Domain);
```
summer-ji-eng, line 156: LGTM

summer-ji-eng (commented):
Thanks @jskeet make it happen. 
One comment is that the way we are get the details, I suggest we rely on `grpc-status-details-bin`, because I am not pretty sure if the `google.rpc.badrequest-bin` is the source of truth.

Another request is that if we could consider http error as well. Because details in http error is the same. So in Go and Typescript Http error could reused to parse the details. How about in C#? For reference: [go-gax](https://github.com/googleapis/gax-go/blob/master/v2/apierror/apierror.go#L154), [gax-nodejs](https://github.com/googleapis/gax-nodejs/blob/85af5280ee1db160445e64d1e6aa86089c2bdfee/src/fallbackProto.ts#L63)

apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/RpcExceptionExtensions.cs
jskeet: If we're absolutely confident that if it's available anywhere, it'll *always* be in grpc-status-details-bin, then that would definitely make things simpler.

jskeet:
I'm not sure what you mean by "http error" here. We only have the [`RpcException`](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/RpcException.cs) by the time there's a problem. Anything that's not already been parsed by the gRPC underlying code isn't available to us.

apis/Google.Cloud.Language.V1/Google.Cloud.Language.V1.Snippets/RpcExceptionExtensions.cs
summer-ji-eng: Loop around with gRPC team, and confirmed that the `grpc-status-details-bin` is the source of truth. Update the answer in YAQS question.

jskeet: Righto. Will update this on Monday. (I'm on vacation tomorrow.)

summer-ji-eng:
> I'm not sure what you mean by "http error" here. We only have the [`RpcException`](https://github.com/grpc/grpc/blob/master/src/csharp/Grpc.Core.Api/RpcException.cs) by the time there's a problem. Anything that's not already been parsed by the gRPC underlying code isn't available to us.

Sorry for the confusion. The "http error" refers to the error from REGAPIC(HTTP/JSON transport). Another view from end user is that they don't know GAPIC or REGAPIC, they expected the same debugging experience. So I wonder if `RpcException` handle the http error, it throw other `Exception` when encounter http error.

jskeet:
> Another view from end user is that they don't know GAPIC or REGAPIC, they expected the same debugging experience.

That's what I've been arguing in terms of making sure we get the HTTP -> gRPC status code mapping correct and consistent for REGAPIC :)

If this is purely about REGAPIC, we can definitely include additional error information in the RpcException when we construct it here: https://github.com/googleapis/gax-dotnet/blob/master/Google.Api.Gax.Grpc/Rest/RestMethod.cs#L190
(It already uses whatever trailers are present in the HTTP response, but we could add our own.)

Effectively, this is taking a direction of "gRPC is the canonical representation, let's make REGAPIC present all the information it has in a gRPC-appropriate fashion" rather than "a transport can be gRPC or HTTP, so we'll try to fetch data from both".

jskeet:
Closing now that we have a productionized version in https://github.com/googleapis/gax-dotnet/pull/500

PR #7166, opened 2021-09-15
chore: Remove additional Asset import

jskeet:
This is no longer necessary, as the generator has been made more lenient.


amanda-tarafa (approved):
LGTM

PR #7167, opened 2021-09-15
chore: Remove obsolete compatibility check in generateapis.sh

jskeet:
We don't need this now that we have a GitHub action to check compatibility.


amanda-tarafa (approved):
LGTM

PR #7168, opened 2021-09-15
Fix deletion of generated code

jskeet:
I was alerted to the problem by the GrpcCtorCompatibility.g.cs files *not* being deleted. We don't *want* them to be deleted, so we need to rename them first...


amanda-tarafa (approved):
LGTM

PR #7175, opened 2021-09-16
docs: Fix the documentation for BigtableServiceApiSettings

jskeet:
The XML doc comments had `list` elements without a type, and docfx
doesn't render those at all.


amanda-tarafa (approved):
LGTM

Issue #7021, opened 2021-08-19
Make multi-API release groups more concrete


jskeet:
Both batch release and other release tools could be more powerful if we made the release groups for multiple APIs (e.g. Diagnostics, Spanner, OSLogin) clearer in the API catalog. This shouldn't be particularly hard to do.

jskeet:
This was at least much improved by #7181 and #7182. We can probably live with this state until we've become more used to it, at which point we'll want to implement batch support.

PR #7181, opened 2021-09-20
First step towards package groups

jskeet:
This introduces package groups into the API catalog, and validates that all project references occur within package groups.

Next steps will be to update commit commands for release manager, and then batch releasing.

Towards #7021


amanda-tarafa (approved):
LGTM

PR #7188, opened 2021-09-20
chore: add owl bot post processor

SurferJeffAtGoogle:
The post processor is not actually invoked yet.


jskeet (approved):
I assume the post-processor isn't actually invoked just because there aren't any OwlBot configs yet, rather than due to something we'll need to enable.

But this basically looks fine, and I'll now work on the other tooling changes to selectively turn it on.

owl-bot-post-processor/main.sh
```diff
@@ -0,0 +1,100 @@
...
+
+TMP=$(mktemp -d)
+
+# Copies one api from the /owl-bot-staging directory into the the /apis dir.
+copy_one_api() {
+  PACKAGE=$1
+  STAGING_DIR=owl-bot-staging/$1
+  PACKAGE_DIR=apis/$1
+
+  # All source files need the extension .g.cs.  Some will only be generated with the
```
jskeet, line 33: Forgot to note that this should be removable really imminently. I'll keep an eye on it though - I suspect I'm competent to remove it when we can, and it won't do any harm anyway.

PR #7190, opened 2021-09-21
chore: Add support for OwlBot

jskeet:
Initially, the "autogenerator" type can be specified in JSON, and
never defaults to OwlBot. This will allow us to test OwlBot.
After this is complete, we can default to using OwlBot for any cases
where there are no scripts that tinker with the generation process.

cc @SurferJeffAtGoogle

### DIFF
--- a/generateapis.sh
+++ b/generateapis.sh
@@ -328,11 +328,13 @@ generate_api() {
     fi
   fi
 
-  # Record the commit in synth.metadata, using either googleapis or googleapis-discovery
-  # depending on the generator.
-  if [[ "$GENERATOR" == "regapic" ]]
+  if [[ -f $PACKAGE_DIR/synth.ph ]]
   then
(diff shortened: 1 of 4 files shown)


amanda-tarafa (approved):
LGTM

generateapis.sh
```diff
@@ -328,11 +328,13 @@ generate_api() {
     fi
   fi
 
-  # Record the commit in synth.metadata, using either googleapis or googleapis-discovery
-  # depending on the generator.
-  if [[ "$GENERATOR" == "regapic" ]]
+  if [[ -f $PACKAGE_DIR/synth.ph ]]
```
SurferJeffAtGoogle, line 331: Should `synth.ph` be `synth.py`?

jskeet: Doh - yes indeed. Will fix in the morning.

PR #7202, opened 2021-09-22
fix: Update the OwlBot Docker image hash to use

jskeet:
I believe this would normally be updated automatically, but a timing
issue in terms of creating this file left it in a broken state.


amanda-tarafa (approved):
LGTM

PR #7234, opened 2021-09-24
Add a "release all" criterion to batch release all changed packages.

jskeet:
This is useful for general catchup.

(Also a tiny bit of extra tidy-up in the batch release command.)

Sample config file:

```json
{
  "DryRun": true,
  "ShowHistory": false,
  "ReleaseAll": {}
}
```


amanda-tarafa (approved):
LGTM

PR #7250, opened 2021-09-27
chore: Add GAPIC generator version selection to generator scripts

jskeet:
This will allow us to keep OwlBot and autosynth in sync in terms of
generator versions.


amanda-tarafa (approved):
LGTM

PR #7258, opened 2021-09-29
fix(Diagnostics): Improve performance of buffers.

### LINKED ISSUE #7177: Thread starvation on asp.net core from logging
#### Environment details

  - OS: ubuntu on gke
  - .NET version: 5
  - Package name and version: Google.Cloud.Diagnostics.AspNetCore3 4.2

#### Steps to reproduce

  1. use the official way to hookup an asp.net 5 app to google diag
  2. log intensively

Hi fine folks.
Using the official asp.net core diag package i'm getting thread starvation due to the way logging is using synch semaphores to protect the buffer.

The simple proof is the next screenshot from VS for a dump i collected from prod.

image

The diagnostics was setup like this
```c#
  webBuilder.UseGoogleDiagnostics(
            projectId: "",
            serviceName: "",
            serviceVersion: "",
            loggerOptions: LoggerOptions.Create(
                logLevel: LogLevel.Information,
                loggerDiagnosticsOutput: Console.Out,
                logName: logName,                         
                monitoredResource: new Google.Api.MonitoredResource
                {
                    Type = "k8s_container",
                    Labels =
                    {
                            { "project_id", ""},
                            { "location", "us-east1-b"},
                            { "cluster_name", ""},
                            { "namespace_name", "default"},
                            { "pod_name", Environment.MachineName }
                    }
                }));
```

I think the bug is related to the semaphore slim used to protect the List that acts as a buffer. This is used on the Receive method, which is invoked from the common GoogleLogger.

I know it's protecting a small operation over an in-memory data structure, but it does mess up my process. 
Why aren't the .net concurrency-friendly collections used instead of manual sync over non-concurrency friendly collections?

lmk if you need more info or in case you believe there's smth wrong with how I use the gcp libs.

Thanks for looking into this!

amanda-tarafa:
Closes #7177

### DIFF
--- a/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
+++ b/apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
@@ -14,8 +14,10 @@
 
 using Google.Api.Gax;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Google.Cloud.Diagnostics.Common
 {
@@ -23,19 +25,21 @@ namespace Google.Cloud.Diagnostics.Common
     /// A <see cref="IFlushableConsumer{T}"/> that will automatically flush the buffer after a 
     /// given amount of time.
     /// </summary>
-    internal class TimedBufferingConsumer<T> : FlushableConsumerBase<T>
+    internal class TimedBufferingConsumer<T> : IFlushableConsumer<T>
     {
-        /// <summary>
-        /// The buffered items. This is not readonly as it is replaced when the buffer is flushed.
-        /// </summary>
-        private List<T> _items = new List<T>();
+        private const int RetryAttempts = 10;
+
+        /// <summary>The consumer to flush to.</summary>
+        private readonly IConsumer<T> _consumer;
+
+        private BlockingCollection<T> _receivedItems = new BlockingCollection<T>();
 
         /// <summary>The timer to automatically flush the buffer.</summary>
         private IThreadingTimer _timer;
 
         internal TimedBufferingConsumer(IConsumer<T> consumer, TimeSpan waitTime, IThreadingTimer timer)
-            : base(consumer)
         {
+            _consumer = GaxPreconditions.CheckNotNull(consumer, nameof(consumer));
             _timer = GaxPreconditions.CheckNotNull(timer, nameof(timer));
             // Initialize the timer to flush ever wait time interval.
             _timer.Initialize((e) => {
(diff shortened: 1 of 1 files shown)


amanda-tarafa:
@jskeet draft for now as I still need to run the stress tests against these changes, but take a look anyway and see if this seems like a better approach than locking.

Also, I'd still need to do something similar for the sized buffer.

amanda-tarafa:
Also, ignore the tests failing. It's just a missing validation. I'll fix it.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
```diff
@@ -61,26 +64,77 @@ public static TimedBufferingConsumer<T> Create(IConsumer<T> consumer, TimeSpan w
             => new TimedBufferingConsumer<T>(consumer, waitTime, new SimpleThreadingTimer());
 
         /// <inheritdoc />
-        public override void Dispose()
+        public void Receive(IEnumerable<T> items)
         {
-            _timer.Dispose();
-            Flush();
+            GaxPreconditions.CheckNotNull(items, nameof(items));
+            var enumerator = items.GetEnumerator();
```
jskeet, line 70: This should have a `using` statement, as `IEnumerator<T>` is disposable (and may be relevant for BlockingCollection)

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
```diff
@@ -61,26 +64,77 @@ public static TimedBufferingConsumer<T> Create(IConsumer<T> consumer, TimeSpan w
...
+            {
+                do
+                {
+                    // We very explicitly use _items here instead of a local
+                    // reference. On Flush, _items will be assigned a new
+                    // collection and we want to use the new collection as soon
+                    // as possible.
+                    // Note that it's still possible that
+                    // 1. We read _items
+                    // 2. We loose execution and a Flush starts.
```
jskeet, line 98:
```suggestion
                    // 2. We lose execution and a Flush starts.
```

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
```diff
@@ -14,28 +14,31 @@
...
+    internal class TimedBufferingConsumer<T> : IFlushableConsumer<T>
     {
-        /// <summary>
-        /// The buffered items. This is not readonly as it is replaced when the buffer is flushed.
-        /// </summary>
-        private List<T> _items = new List<T>();
+        /// <summary>The consumer to flush to.</summary>
+        private readonly IConsumer<T> _consumer;
+
+        private BlockingCollection<T> _items = new BlockingCollection<T>();
```
jskeet, line 34: Can we call this something else? I've been really confused when reviewing this because we're getting an enumerator from a parameter called `items` and adding values to `_items`.

Maybe `_bufferedItems` or `_receivedItems`?

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
```diff
@@ -61,26 +64,77 @@ public static TimedBufferingConsumer<T> Create(IConsumer<T> consumer, TimeSpan w
...
+                    // Also note that if we regain execution in between 3 and 4, that's fine.
+                    // Our call to Add will happen before the call to CompleteAdding, so it will
+                    // succeed before the collection is marked as complete.
+                    _items.Add(remainingItems.Current);
+                }
+                while (remainingItems.MoveNext());
+            }
+            catch (InvalidOperationException)
+            {
+                Receive(remainingItems);
```
jskeet, line 113: I'd like us to have some guard on the recursion here. If we end up recursing (say) 10 levels deep, I think we should fail with an exception or similar - rather than going potentially forever until a StackOverflowException brings down the whole process.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
```diff
@@ -61,26 +64,77 @@ public static TimedBufferingConsumer<T> Create(IConsumer<T> consumer, TimeSpan w
...
+                    // 2. We loose execution and a Flush starts.
+                    // 3. Flush swaps _items.
+                    // 4. Flush marks the old items (the one we are referencing) as CompleteAdding.
+                    // 5. We gain execution and execute Add on a completed collection, which fails.
+                    // That's why we catch InvalidOperationException and simply reattempt just
+                    // where we left off.
+                    // Also note that if we regain execution in between 3 and 4, that's fine.
+                    // Our call to Add will happen before the call to CompleteAdding, so it will
+                    // succeed before the collection is marked as complete.
+                    _items.Add(remainingItems.Current);
```
jskeet, line 107: This won't necessarily read the absolute latest value of `_items`... perhaps we should use `Interlocked.CompareExchange` here? (Or some other way of achieving a memory barrier, maybe within the catch block?)

jskeet (commented):
Mostly fine, but a few nits. Will be interested to hear how much difference it makes.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
amanda-tarafa: Yep, I thought of making it volatile, but that's probably too much.
I've added Interlock.MemoryBarrier() on the catch, which should be enough.

apis/Google.Cloud.Diagnostics.Common/Google.Cloud.Diagnostics.Common/TimedBufferingConsumer.cs
amanda-tarafa: OK, I'll reset it on each success and let's see what happens.

amanda-tarafa (commented):
All comments addressed, I'll now run the stress tests.

amanda-tarafa:
Closing as this is not necessary. See https://github.com/googleapis/google-cloud-dotnet/issues/7177#issuecomment-931499286 for details.

PR #7259, opened 2021-09-29
fix: Fix diff checker when we're building on a stable release

jskeet:
This will sort out the breaking change detection in #7157


amanda-tarafa (approved):
Ah yes! LGTM

PR #7284, opened 2021-10-05
chore: remove .cs -> .g.cs naming

jskeet:
The generator has been updated to generate the right files to start with.

PR #7285, opened 2021-10-05
OwlBot regeneration using generateapis.sh

jskeet:
First commit is from #7284.

This is very drafty at the moment - I'll want to do some local testing first to make sure that it's at least plausible, before I merge.

However, if you can spot anything hugely wrong with the overall approach, that would be really useful.


jskeet:
(Removed reference to first commit from the description, as I've now rebased this PR.)

owl-bot-post-processor/main.sh
```diff
@@ -30,6 +33,42 @@ copy_one_api() {
   STAGING_DIR=owl-bot-staging/$1
   PACKAGE_DIR=apis/$1
 
+  # We don't expect googleapis-gen to contain the right (or potentially
```
amanda-tarafa, line 36: Just a question: Are we sure that, whatever had happened in googleapis-gen, we still get to execute this? Or could that fail in such a way that this is not called?

jskeet: This function is only called if there's something in the owl-bot-staging directory for an API, which suggests it's managed to copy the API over from googleapis-gen already. If an API fails to generate in googleapis-gen itself, that's a whole other issue, and one which I assume Bazel Bot handles.

PR #7290, opened 2021-10-07
docs: Change notes from being headings to quoted text

jskeet:
The double space after `**Note**` in each case is required to force
a line break.

Screenshot of result is below. I'd prefer it if we had a blank line to separate the two notes, but the docfx Markdown processor doesn't do that, and I don't think it's worth going to extraordinary measures to introduce it.

Screenshot of documentation


jskeet:
Yup, that's absolutely fine.

PR #7415, opened 2021-11-02
fix: Adjust the common Bigtable Table resource service

jskeet:
This was changed in protos, and needs to be changed manually here to
allow both the Bigtable and Bigtable Admin APIs to pick it up when
being regenerated.


amanda-tarafa (approved):
LGTM

PR #7432, opened 2021-11-05
chore: Update GAPIC generator version

jskeet:
This now always emits gapic_metadata.json, so we need to remove the
flag in our generation script.


amanda-tarafa (approved):
LGTM

PR #7445, opened 2021-11-08
chore: remove facility to generate a "protogrpc" package

jskeet:
The "protogrpc" generator type ran protoc with the standard C# and
C# gRPC plugins, but not GAPIC. Google.Cloud.Iam.V1 was the only
package to use this, and it's now gone full-GAPIC. We don't expect
to ever use this facility again, so let's remove it.


amanda-tarafa (approved):
LGTM

PR #7476, opened 2021-11-10
Remove all traces of googleapis-discovery

jskeet:
This is a big win in terms of simplifying generation, and also unblocks moving Compute to OwlBot


amanda-tarafa (approved):
LGTM

PR #7479, opened 2021-11-10
fix: When generating history, get commits in the right version

jskeet:
If a commit is both "eligible for history" *and* a
version-tagging-commit, it should be part of the history for that
version, not the subsequent one. This shouldn't really happen, as we
should create release commits that only do the version changes, but
it happens every so often.


amanda-tarafa (approved):
LGTM

PR #7505, opened 2021-11-11
chore: Migrate Google.Cloud.Compute.V1 to OwlBot

jskeet:
With this change in place, we don't use synthtool for anything.
We'll wait a few weeks before removing support for it, however, just
in case something comes up.


amanda-tarafa (approved):
LGTM

PR #7513, opened 2021-11-12
docs: Convert h1 to h2 in all docs

jskeet:


amanda-tarafa (approved):
LGTM

Thanks for doing this!

jskeet:
(Unfortunately all APIs have "changed" here as our change detector isn't smart enough to spot that docs aren't code. And there's a parameter rename in Spanner that counts as a breaking change...)

PR #7515, opened 2021-11-12
pubsub: update emulator sample

anguillanneuf:
See internal b/206034017.


jskeet:
I don't believe this is what we want to do - see internal bug.

PR #7520, opened 2021-11-15
tests: Added two additional integration tests for dml insert

AlexandrTrf:
Added 2 tests for inserting data using DML API in Cloud Spanner

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AllTypesTableFixture.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AllTypesTableFixture.cs
@@ -28,12 +28,58 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         {
         }
 
+        /// <summary>
+        /// Creates INSERT command for the table created after executing the <see cref="CreateTable"/> method execution.
+        /// </summary>
+        /// <returns>The DML command to insert data into a table.</returns>
+        public string CreateInsertCommand()
+        {
+            // The emulator doesn't yet support the JSON type.
+            return $@"INSERT {TableName} (
+                              K,
+                              BoolValue,           
+                              Int64Value,          
+                              Float64Value,        
+                              StringValue,       
+                              NumericValue,        
+                              BytesValue,       
+                              TimestampValue,      
+                              DateValue,           
+                              BoolArrayValue,      
+                              Int64ArrayValue,     
+                              Float64ArrayValue,   
+                              NumericArrayValue,  
+                              StringArrayValue,    
+                              Base64ArrayValue,
+                              BytesArrayValue,     
+                              TimestampArrayValue,
+                              {(RunningOnEmulator ? "" : "JsonArrayValue,") } 
+                              DateArrayValue) VALUES(
+                              @K,
+                              @BoolValue,           
+                              @Int64Value,          
+                              @Float64Value,        
+                              @StringValue,         
+                              @NumericValue,        
+                              @BytesValue,          
+                              @TimestampValue,      
+                              @DateValue,           
+                              @BoolArrayValue,      
+                              @Int64ArrayValue,     
+                              @Float64ArrayValue,   
+                              @NumericArrayValue,   
+                              @StringArrayValue,
+                              @Base64ArrayValue,
+                              @BytesArrayValue,     
+                              @TimestampArrayValue,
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AllTypesTableFixture.cs
```diff
@@ -28,6 +28,91 @@ public AllTypesTableFixture() : base("TypesTable")
         {
         }
 
+        /// <summary>
+        /// Creates INSERT command for the table created after executing the <see cref="CreateTable"/> method execution.
+        /// </summary>
+        /// <returns>The DML command to insert data into a table.</returns>
+        public string CreateInsertCommand()
+        {
+            // The emulator doesn't yet support the JSON type.
```
jskeet, line 37: Couldn't this be rather simpler with an interpolated verbatim string literal which contains something like:

```csharp
{ (RunningOnEmulator ? "" : "JsonArrayValue,") }
```

and

```csharp
{ (RunningOnEmulator ? "" : "@JsonArrayValue,") }
```

(You'd want to reorder it so that the JSON array value isn't the last element, unless Spanner is happy with trailing commas.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -81,11 +81,11 @@ public async Task WriteValues()
             SpannerNumeric?[] nArray = { SpannerNumeric.Parse("0.0"), null, SpannerNumeric.Parse("2.0") };
             string[] jsonArray = { "{\"f1\":\"v1\"}", "{}", "[]", null };
             string[] sArray = { "abc", null, "123" };
-            string[] bArrayArray =
+            byte[][] bArrayArray =
```
jskeet, line 84: I'd suggest adding a second `BytesArray` field, so we can test that both base64-string and real-byte-array representations work. (Basically, I want to keep testing what we're already testing, rather than removing an existing test and replacing it with another.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AllTypesTableFixture.cs
jskeet: (There may be other places we could do that, too.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -150,6 +148,96 @@ public async Task WriteValues()
             });
         }
 
+        [Fact]
+        public async Task WriteValuesDml()
+        {
+            var testTimestamp = new DateTime(2017, 3, 17, 15, 30, 0);
```
jskeet, line 154: Rather than copying the data from WriteValues, can we just have some common code?

jskeet:
Is this now ready for another review? Please let me know. (If you could rebase from the default branch instead of merging, that would be handy too.)

AlexandrTrf:
> Is this now ready for another review? Please let me know. (If you could rebase from the default branch instead of merging, that would be handy too.)

Yes, it`s ready. I will do rebase a little bit later.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AllTypesTableFixture.cs
jskeet: Note that now this is the only statement, the method can be turned into an expression-bodied one.

jskeet (approved):
This is much better, thanks.

amanda-tarafa (approved):
LGTM

AlexandrTrf:
I just did a rebase. Could you run a workflows?

amanda-tarafa:
Just as a note, although we may keep the commit history during review, so it's easier to know what changes between commits, we usually squash all commits into one, or maybe two meaningful commits before merging, so that the repo commit history is cleaner.
I think you can squash your last two commits into the firt one. Thanks.

AlexandrTrf:
Thanks! Done!

jskeet (approved):

jskeet:
Thanks - will merge on green.

PR #7539, opened 2021-11-17
Change how Owl Bot decides to regenerate locally

jskeet:
This is rather more flexible, and puts logic in C# instead of in the shell script


amanda-tarafa (approved):
LGTM

+1 in doing this in C#

PR #7553, opened 2021-11-18
fix: clone each SpannerParameter when cloning a parameter collection

olavloite:
Cloning a `SpannerParameterCollection` would not clone the individual `SpannerParameter` instances
in the collection. This would mean that if you cloned a `SpannerCommand` and then changed the
value of a parameter on the original command, the value would also be changed on the cloned
command.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
@@ -280,6 +280,7 @@ namespace Google.Cloud.Spanner.Data
         /// Clones the collection.
         /// </summary>
         /// <returns>A cloned copy of this instance.</returns>
-        public SpannerParameterCollection Clone() => new SpannerParameterCollection(_innerList);
+        public SpannerParameterCollection Clone() =>
+            new SpannerParameterCollection(_innerList.Select(p => p.Clone() as SpannerParameter));
     }
 }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
@@ -281,6 +281,6 @@ namespace Google.Cloud.Spanner.Data
         /// </summary>
         /// <returns>A cloned copy of this instance.</returns>
         public SpannerParameterCollection Clone() =>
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
```diff
@@ -280,6 +280,7 @@ private void FillSpannerInternalTypes(MapField<string, V1.Type> typeDictionary)
         /// Clones the collection.
         /// </summary>
         /// <returns>A cloned copy of this instance.</returns>
-        public SpannerParameterCollection Clone() => new SpannerParameterCollection(_innerList);
+        public SpannerParameterCollection Clone() =>
+            new SpannerParameterCollection(_innerList.Select(p => p.Clone() as SpannerParameter));
```
jskeet, line 284: I'd suggest casting rather than using `as`, on general principle: while I'd always expect it to actually give the same result, any time it *didn't* I'd prefer a useful exception immediately instead of an ArgumentNullException or NullReferenceException later on.

jskeet (commented):
I'm fine with the fix (other than the nitpick) - but I'd suggest adding a test (that would have failed before) too.

olavloite:
@jskeet PTAL
I've added a test case that would fail before this change, and now succeeds.

In addition it might be worth noting that the `CloneUpdateWithParameters()` test failed after introducing this change, as it checked whether the parameters of a cloned command were equal to the original command. However, as `SpannerCommand` does not implement `Equals(..)` it in reality checked for reference equality of the underlying parameters of the original and cloned command.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
olavloite: Done.

jskeet:
Will review on Monday - I'm on vacation today.

jskeet (approved):

jskeet:
Closed and reopened with "allow breaking changes" as the parameter reordering in an earlier PR is strictly-speaking breaking (but in a way we're okay with). This is probably a sign that we should do a minor release soon :)

olavloite:
> This is probably a sign that we should do a minor release soon :)

+1 to that. This change and the one for `GetBytes` are needed for the NHibernate driver and dialect that we are currently working on.

jskeet:
Cool. We're in a release freeze right now, but we can do a release on December 1st if that would be useful.

olavloite:
> Cool. We're in a release freeze right now, but we can do a release on December 1st if that would be useful.

Yes, that would certainly be useful. Thanks!

PR #7563, opened 2021-11-22
chore: Make the generation script more selective about gRPC service configs

jskeet:
This will fix local Compute generation, which otherwise doesn't
apply the service config due to the other JSON file in the same
directory.


amanda-tarafa (approved):
LGTM

PR #7575, opened 2021-12-02
chore: Remove synthtool support

jskeet:
A few notes on the shell script changes:

- The environment variable SYNTHTOOL_GOOGLEAPIS used to specify an
  existing googleapis directory (and indicating that it shouldn't be
  fetched/updated) is now GOOGLEAPIS_DIR
- We don't expect to ever have to run against a googleapis directory
  without google/cloud/common now, so remove that conditionality
- We don't need to cache the generator any more. Running locally we
  generate all APIs in a single run anyway; running in OwlBot would
  only generate a single API (or a small set) so there's no/little
  benefit to caching.


amanda-tarafa (approved):
LGTM as far as I can tell

PR #7622, opened 2021-12-10
feat: First pass at prettier release notes

jskeet:
These are used in both the version histories and the GitHub release
pages.

This currently assumes that every line in a production code commit
message is a feature/fix/etc. In reality, we'll need to handle
comments like the one on this commit, which are longer - but that
should be reasonably rare, and easily done by hand.

"Chore" commits are discarded, but any other commits with unknown
(or missing) conventional commit types end up in a section of the
history document which prevents ReleaseManager from committing until
it's cleaned up.

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/History/HistoryFile.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/History/HistoryFile.cs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 using Google.Cloud.Tools.Common;
+using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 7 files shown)


jskeet:
cc @bcoe

tools/Google.Cloud.Tools.ReleaseManager/History/HistoryFile.cs
```diff
@@ -132,18 +134,44 @@ internal Section(Release release)
...
+            internal IEnumerable<string> GetNotesFromElement(IGrouping<ReleaseNoteElementType, ReleaseNoteElement> elements)
+            {
+                string description = elements.Key switch
+                {
+                    ReleaseNoteElementType.Fix => "Bug fixes",
+                    ReleaseNoteElementType.Feature => "New features",
+                    ReleaseNoteElementType.Chore => "Chores",
+                    ReleaseNoteElementType.Docs => "Documentation improvements",
+                    ReleaseNoteElementType.Unknown => FixmeBlockingRelease,
+                    _ => throw new InvalidOperationException("Unknown element type")
```
amanda-tarafa, line 165: This should signal a bug in the tool, right? Unknown elements sould really be written out on the Fixme section?

amanda-tarafa (approved):
LGTM

tools/Google.Cloud.Tools.ReleaseManager/History/HistoryFile.cs
jskeet: Yes, it's a bug. This is where "we don't know what the element type is" rather than "we know that the element type is 'unknown', because we parsed it to that". Really, really shouldn't happen :)

PR #7642, opened 2022-01-04
fix: Adds AddGoogleDiagnosticsForAspNetCore empty overload.

### LINKED ISSUE #7633: The call is ambiguous between the following methods
#### Environment details

  - OS: macOS 12
  - .NET version: 6.0
  - Package name and version: Google.Cloud.Diagnostics.AspNetCore 4.3.1

#### Steps to reproduce

  1. Create a basic dotnet 6 ASP.NET app
  2. Follow [this Google documentation] (https://cloud.google.com/dotnet/docs/reference/Google.Cloud.Diagnostics.AspNetCore3/latest#configuration-1) up to and including Getting Started.
  3. Assume that your code will run on GCP Cloud Run and omit `ProjectId`, `Service` and `Version` as documented.
  4. Observe that your code will not compile.

`services.AddGoogleDiagnosticsForAspNetCore();`

`The call is ambiguous between the following methods or properties: 'AspNetCoreGoogleDiagnosticsExtensions.AddGoogleDiagnosticsForAspNetCore(IServiceCollection, AspNetCoreTraceOptions, LoggingServiceOptions, ErrorReportingServiceOptions)' and 'AspNetCoreGoogleDiagnosticsExtensions.AddGoogleDiagnosticsForAspNetCore(IServiceCollection, string, string, string, TraceOptions, LoggingOptions, ErrorReportingOptions)' [Evoq.Timesheets.Web.Application]csharp(CS0121)`

I think it's an API bug to have two overloaded signatures that cannot be resolved in this way. I'm not entirely sure there's a solution without breaking it or introducing a second method or third overload with a unique signature, I've not given it much thought, but perhaps the documentation needs changing.

amanda-tarafa:
To avoid ambiguity, which results in a compiler error, when not specifying any of the optional parameters for the other overloads.
Fixes #7633

### DIFF
--- a/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsTests.cs
+++ b/apis/Google.Cloud.Diagnostics.AspNetCore/Google.Cloud.Diagnostics.AspNetCore.IntegrationTests/DiagnosticsTests.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Api.Gax;
 using Google.Cloud.ClientTesting;
 using Google.Cloud.Diagnostics.Common;
 using Google.Cloud.Diagnostics.Common.IntegrationTests;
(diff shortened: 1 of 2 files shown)


jskeet (approved):
Excellent - so using null literals (with no types) would still lead to ambiguities, but that's presumably a much less important use case.

amanda-tarafa:
> Excellent - so using null literals (with no types) would still lead to ambiguities, but that's presumably a much less important use case.

Yes, that would be the case, but as you said, the use case would be less common and meaningful. And named parameters can also be used for disambiguation in some cases. not just the types.
And, I would say that this kind of ambiguity (null literals) is more intrinsic of overloads + optional paramters than of any particular API surface.

PR #7655, opened 2022-01-06
Fix .repo-metadata.json issues

jskeet:
I don't believe this will actually remove *all* the issues from #7651, but it will reduce it to a simpler set to reason about. (I think there are discrepancies in library_type between what we believe is valid and what the tool believes is valid...)

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
@@ -68,7 +68,9 @@ namespace Google.Cloud.Tools.ReleaseManager
                 Type = ApiType.Grpc,
                 Generator = GeneratorType.Micro,
                 // Let's not include test dependencies, which are rarely useful.
-                TestDependencies = null
+                TestDependencies = null,
+                // Translate the host name into the "short name", e.g. bigquery.googleapis.com => bigquery
+                ShortName = targetApi.HostName.Split('.').First()
             };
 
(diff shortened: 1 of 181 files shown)


tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
```diff
@@ -68,7 +68,9 @@ protected override void ExecuteImpl(string[] args)
                 Type = ApiType.Grpc,
                 Generator = GeneratorType.Micro,
                 // Let's not include test dependencies, which are rarely useful.
-                TestDependencies = null
+                TestDependencies = null,
+                // Translate the host name into the "short name", e.g. bigquery.googleapis.com => bigquery
+                ShortName = targetApi.HostName.Split('.').First()
```
amanda-tarafa, line 73: You need to account for the regional default endpoints as well, which are supposed to become more common in the short term.
See https://github.com/googleapis/gapic-generator-csharp/blob/543fe7042d6de918f9bab841256400b1b97b238e/Google.Api.Generator/Generation/ServiceDetails.cs#L60

amanda-tarafa (approved):
LGTM

tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
jskeet: I don't think so - that's based on the default host in a .proto service; this is the "name" from the service config.

PR #7656, opened 2022-01-06
More repo metadata fixes

jskeet:
These should really take #7651 down to a very small number of issues (if any)


amanda-tarafa (approved):
LGTM

PR #7860, opened 2022-01-19
"Upgrade" Identity Access Context Manager to a full API

jskeet:
After this, regenerating #7717 should be a no-op.


amanda-tarafa (approved):
LGTM

PR #7861, opened 2022-01-19
chore: Fix Google.LongRunning postgeneration

jskeet:
When running in OwlBot, we need to make sure protoc is actually installed.


amanda-tarafa (approved):
LGTM

PR #7894, opened 2022-01-26
docs: Update help pages with respect to gRPC

jskeet:
Ideally we'd have more comprehensive documentation about choosing
the gRPC adapter (rather than linking to an issue) - but all that
would need to change substantially in a few months anyway. We should
revisit all of this documentation when the next major version ships.


amanda-tarafa (approved):
LGTM

PR #7898, opened 2022-01-27
Retry handling

anshuldavid13:
Retry handler sample

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
@@ -28,8 +28,15 @@ namespace Google.Cloud.Storage.V1
     /// </summary>
     internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
     {
-        // Retriable error codes: 408—Request Timeout, 429-Too Many Requests, 500-Internal Server Error, 502—Bad Gateway, 503—Service Unavailable, 504—Gateway Timeout
-        private static readonly int[] s_retriableErrorCodes = { 408, 429, 500, 502, 503, 504 };
+        private static readonly int[] s_retriableErrorCodes =
+        {
+                408, // Request timeout
+                429, // Too many requests
+                500, // Internal server error
+                502, // Bad gateway
+                503, // Service unavailable
+                504 // Gateway timeout
+        };
         private static RetryHandler s_instance = new RetryHandler();
 
         private RetryHandler() { }
@@ -38,7 +45,8 @@ namespace Google.Cloud.Storage.V1
             request.AddUnsuccessfulResponseHandler(s_instance);
 
         // This function is designed to support asynchrony in case we need to examine the response content, but for now we only need the status code
-        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response) => Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
+        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response) => 
+            Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
 
         public async Task<bool> HandleResponseAsync(HandleUnsuccessfulResponseArgs args)
         {
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
@@ -1,0 +1,63 @@
+﻿// Copyright 2022 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Apis.Http;
+using Google.Apis.Storage.v1;
+using System;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Storage.V1
+{
+    /// <summary>
+    /// Handler for exponential backoff for request messages which permit retries.
+    /// This is similar to BackoffHandler in Google.Apis.Http, but that doesn't give us enough flexibility.
+    /// This handler does not enforce a maximum number of retries; that's configured via <see cref="ConfigurableMessageHandler.NumTries"/>.
+    /// </summary>
+    internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
+    {
+        private static readonly int[] s_retriableErrorCodes = { 408, 429, 500, 502, 503, 504 };
+        private static RetryHandler s_instance = new RetryHandler();
+
+        private RetryHandler() { }
+
+        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request) =>
+            request.AddUnsuccessfulResponseHandler(s_instance);
+
+        //This function is designed to be asynchronous in case we need to examine the response content, but for now we only need the status code
+        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response)
+        {
+            return Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
+        }
+
+        public async Task<bool> HandleResponseAsync(HandleUnsuccessfulResponseArgs args)
+        {
+            var retry = args.SupportsRetry && await IsRetriableResponse(args.Response).ConfigureAwait(false);
+            if (!retry)
+            {
+                return false;
+            }
+            // The first failure will have args.CurrentFailedTry set to 1,
+            // whereas we want the first delay to be 1 second. We use Math.Min on the power
+            // rather than on the result to obtain a max retry of 32 seconds without risking
+            // calling Math.Pow with a huge number.
+            int power = Math.Min(args.CurrentFailedTry - 1, 5);
+            double seconds = Math.Pow(2.0, power);
+            var delay = TimeSpan.FromSeconds(seconds);
+            await Task.Delay(delay, args.CancellationToken).ConfigureAwait(false);
+            return true;
+        }
+    }
+}
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
@@ -28,6 +28,7 @@ namespace Google.Cloud.Storage.V1
     /// </summary>
     internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
     {
+        // Retriable error codes: 408—Request Timeout, 429-Too Many Requests, 500-Internal Server Error, 502—Bad Gateway, 503—Service Unavailable, 504—Gateway Timeout
         private static readonly int[] s_retriableErrorCodes = { 408, 429, 500, 502, 503, 504 };
         private static RetryHandler s_instance = new RetryHandler();
 
@@ -36,11 +37,8 @@ namespace Google.Cloud.Storage.V1
         internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request) =>
             request.AddUnsuccessfulResponseHandler(s_instance);
 
-        //This function is designed to be asynchronous in case we need to examine the response content, but for now we only need the status code
-        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response)
-        {
-            return Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
-        }
+        // This function is designed to support asynchrony in case we need to examine the response content, but for now we only need the status code
+        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response) => Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
 
         public async Task<bool> HandleResponseAsync(HandleUnsuccessfulResponseArgs args)
         {
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryHandlerTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryHandlerTest.cs
@@ -1,0 +1,53 @@
+﻿// Copyright 2022 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Net;
+using System.Net.Http;
+using Xunit;
+
+namespace Google.Cloud.Storage.V1.Tests
+{
+    public class RetryHandlerTest
+    {
+        [Fact]
+        public void IsRetriableResponse_BadGateway()
+        {
+            var httpResponse = new HttpResponseMessage
+            {
+                StatusCode = HttpStatusCode.BadGateway
+            };
+            var actual = RetryHandler.IsRetriableResponse(httpResponse);
+            Assert.True(actual.Result);
+        }
+
+        [Fact]
+        public void NonRetriableResponse_RequestUriTooLong()
+        {
+            var httpResponse = new HttpResponseMessage
+            {
+                StatusCode = HttpStatusCode.RequestUriTooLong
+            };
+            var actual = RetryHandler.IsRetriableResponse(httpResponse);
+            Assert.False(actual.Result);
+        }
+
+        [Fact]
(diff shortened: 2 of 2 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+            try
+            {
+                // Always retry 502 errors.
+                if (response.StatusCode == HttpStatusCode.BadGateway)
+                {
+                    return true;
+                }
+
+                // For other errors, check the response content.
+                var str = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
```
jskeet, line 55: Given that we expect this to be JSON, I'd suggest renaming the variable to `json`.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+                errorResponse = NewtonsoftJsonSerializer.Instance.Deserialize<StandardResponse<object>>(str);
+                var errors = errorResponse.Error?.Errors;
+                // If we couldn't deserialize properly, or the result doesn't contain any reasons, we won't retry.
```
jskeet, line 58: If we couldn't deserialize properly, we'll be in the catch block, right?
Basically the comment doesn't seem to match the condition - it looks like this is really "Only retry if we know that all errors specified in the response are retriable"

jskeet: And I note this is a problem in the BigQuery one as well - let's improve it in both places :)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+
+        internal static void MarkAsRetriable<TResponse>( StorageBaseServiceRequest<TResponse> request) =>
+            request.AddUnsuccessfulResponseHandler(s_instance);
+
+        internal static async Task<bool> IsRetriableResponse(HttpResponseMessage response)
+        {
+            StandardResponse<object> errorResponse = null;
+            try
+            {
+                // Always retry 502 errors.
```
jskeet, line 48: Is this correct for Storage as well as BigQuery?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,86 @@
...
+namespace Google.Cloud.Storage.V1
+{
+    /// <summary>
+    /// Handler for exponential backoff for request messages which permit retries.
+    /// This is similar to BackoffHandler in Google.Apis.Http, but that doesn't give us enough flexibility.
+    /// This handler does not enforce a maximum number of retries; that's configured via <see cref="ConfigurableMessageHandler.NumTries"/>.
+    /// </summary>
+    internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
+    {
+        private static readonly string[] s_retriableErrors = { "rateLimitExceeded", "backendError" };
```
jskeet, line 34: This is retrying based on the reason, which is what we were advised for BigQuery. Is that the same for Storage, or should we use the code instead?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.GetBucket.cs
```diff
@@ -36,6 +36,7 @@ private BucketsResource.GetRequest CreateGetBucketRequest(string bucket, GetBuck
             var bucketEntity = new Bucket { Name = bucket };
             var request = Service.Buckets.Get(bucket);
             options?.ModifyRequest(request);
+            RetryHandler.MarkAsRetriable(request);
```
jskeet, line 39: Does the implementation document provide any more conditions? I suspect we'll want the retry handler to be more fine-grained for Storage than for BigQuery.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -52,11 +52,14 @@ internal static async Task<bool> IsRetriableResponse(HttpResponseMessage respons
...
-                var str = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
-                errorResponse = NewtonsoftJsonSerializer.Instance.Deserialize<StandardResponse<object>>(str);
-                var errors = errorResponse.Error?.Errors;
-                // If we couldn't deserialize properly, or the result doesn't contain any reasons, we won't retry.
-                return errors?.All(singleError => s_retriableErrors.Contains(singleError.Reason)) ?? false;
+                var json = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                errorResponse = NewtonsoftJsonSerializer.Instance.Deserialize<StandardResponse<object>>(json);
+                return s_retriableErrorCodes.Contains(errorResponse.Error.Code);
+            }
+            catch (WebSocketException wsex)
```
jskeet, line 59: I think this needs a comment, at least. Are we convinced this is the exception thrown on all platforms?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -52,11 +52,14 @@ internal static async Task<bool> IsRetriableResponse(HttpResponseMessage respons
...
+                when (wsex.WebSocketErrorCode == WebSocketError.ConnectionClosedPrematurely)
```
jskeet, line 60: I'd unwrap this into the previous line. It'll be a long line, but more obvious than at the moment IMO.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
jskeet: Ping for this comment.

jskeet: And in fact, if we're going on HTTP status code, this is an example of how it can be retrieved, without deserializing the error at all. I think this can be simplified.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.GetBucket.cs
anshuldavid13: It doesnt have any condition for this implementation. Will add them according to the documentation

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+        private static RetryHandler s_instance = new RetryHandler();
+
+        private RetryHandler() { }
+
+        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request) =>
+            request.AddUnsuccessfulResponseHandler(s_instance);
+
+        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response)
+        {
+            return Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
```
jskeet, line 45: It might be worth noting in a comment that this is designed to be asynchronous in case we need to examine the response content, but for now we only need the status code.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public async Task<bool> HandleResponseAsync(HandleUnsuccessfulResponseArgs args)
+        {
+            var retry = args.SupportsRetry &&
```
jskeet, line 50: Nit: I'd unwrap this to a single line.

jskeet (approved):
Looking good. Is the plan to merge as-is and then create a follow-up PR with other operations, or modify this one to be complete?

jskeet (commented):
Something seems to have gone wrong with the rebase here - let's talk about the git side of things offline.

jskeet (commented):
"Merge branch 'retryHandling' of" - this is part of the problem. If you rebase and force push, you shouldn't end up with any merge commits.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        private static readonly int[] s_retriableErrorCodes = { 408, 429, 500, 502, 503, 504 };
+        private static RetryHandler s_instance = new RetryHandler();
+
+        private RetryHandler() { }
+
+        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request) =>
+            request.AddUnsuccessfulResponseHandler(s_instance);
+
+        //This function is designed to be asynchronous in case we need to examine the response content, but for now we only need the status code
```
jskeet, line 39: Nit: add a space after `//` - and probably change "to be asynchronous" to "to support asynchrony" (as it's *currently* synchronous)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+namespace Google.Cloud.Storage.V1
+{
+    /// <summary>
+    /// Handler for exponential backoff for request messages which permit retries.
+    /// This is similar to BackoffHandler in Google.Apis.Http, but that doesn't give us enough flexibility.
+    /// This handler does not enforce a maximum number of retries; that's configured via <see cref="ConfigurableMessageHandler.NumTries"/>.
+    /// </summary>
+    internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
+    {
+        private static readonly int[] s_retriableErrorCodes = { 408, 429, 500, 502, 503, 504 };
```
jskeet, line 31: It would probably be worth listing what these error codes mean, textually. (Just in a comment.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+        private static RetryHandler s_instance = new RetryHandler();
+
+        private RetryHandler() { }
+
+        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request) =>
+            request.AddUnsuccessfulResponseHandler(s_instance);
+
+        //This function is designed to be asynchronous in case we need to examine the response content, but for now we only need the status code
+        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response)
```
jskeet, line 40: I'd personally use an expression-bodied member here, i.e. `=>` at the end of this line, get rid of the braces, and remove the `return` token, but it's fine as-is.

jskeet (approved):
Basically fine - a few comments that can be updated, then you're fine to merge.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+            var retry = args.SupportsRetry && await IsRetriableResponse(args.Response).ConfigureAwait(false);
+            if (!retry)
+            {
+                return false;
+            }
+            // The first failure will have args.CurrentFailedTry set to 1,
+            // whereas we want the first delay to be 1 second. We use Math.Min on the power
+            // rather than on the result to obtain a max retry of 32 seconds without risking
+            // calling Math.Pow with a huge number.
+            int power = Math.Min(args.CurrentFailedTry - 1, 5);
```
tritone, line 56: Are there currently retried RPCs in any of the dotnet libraries? If so is it possible to share the backoff timing code with them? Probably don't need a one-off for storage. In Go for example we use gax across products.  https://pkg.go.dev/github.com/googleapis/gax-go/v2

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// This handler does not enforce a maximum number of retries; that's configured via <see cref="ConfigurableMessageHandler.NumTries"/>.
+    /// </summary>
+    internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
+    {
+        private static readonly int[] s_retriableErrorCodes = { 408, 429, 500, 502, 503, 504 };
+        private static RetryHandler s_instance = new RetryHandler();
+
+        private RetryHandler() { }
+
+        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request) =>
```
tritone, line 36: We'd also want to mark connection reset and broken stream errors as retryable. Have you been able to see what these will look like in dotnet? If not the testbench is a good way to test this out. https://github.com/googleapis/storage-testbench

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        //This function is designed to be asynchronous in case we need to examine the response content, but for now we only need the status code
+        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response)
+        {
+            return Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
+        }
+
+        public async Task<bool> HandleResponseAsync(HandleUnsuccessfulResponseArgs args)
+        {
+            var retry = args.SupportsRetry && await IsRetriableResponse(args.Response).ConfigureAwait(false);
```
tritone, line 47: Will `SupportsRetry` cover the idempotency piece? Curious how this will work.

tritone (commented):
A couple thoughts from the storage perspective, happy to discuss more.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -36,11 +37,8 @@ internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
         internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request) =>
             request.AddUnsuccessfulResponseHandler(s_instance);
 
-        //This function is designed to be asynchronous in case we need to examine the response content, but for now we only need the status code
-        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response)
-        {
-            return Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
-        }
+        // This function is designed to support asynchrony in case we need to examine the response content, but for now we only need the status code
+        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response) => Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
```
jskeet, line 41: This is fine - but I'd have put the actual implementation on a new line to avoid the line being quite so long. Apologies if I wasn't clear in my previous comment. We don't have hard-and-fast line length rules, but this is currently 165 characters which is longer than I'd normally like. (I'm sure you can find counterexamples though!)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -28,6 +28,7 @@ namespace Google.Cloud.Storage.V1
     /// </summary>
     internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
     {
+        // Retriable error codes: 408—Request Timeout, 429-Too Many Requests, 500-Internal Server Error, 502—Bad Gateway, 503—Service Unavailable, 504—Gateway Timeout
```
jskeet, line 31: I'd probably put these into multiple lines:

```csharp
// Retriable error codes:
// - 408: Request Timeout
// - 429: Too Many Requests
// - 500: Internal Server Error
// - 502: Bad Gateway
// - 503: Service Unavailable
// - 504: Gateway Timeout
```

An alternative formatting would be to comment inline in the array declaration:

```csharp
private static readonly int[] s_retriableErrorCodes = { 408, 429, 500, 502, 503, 504 };
{
    408, // Request timeout
    429, // Too many requests
    500, // Internal server error
    502, // Bad gateway
    503, // Service unavailable
    504, // Gateway timeout
};
```

I *think* that's probably better, as it makes it harder for the comment and array content to get out of sync... but see what you think.

jskeet (approved):
Semantically fine - a couple more formatting suggestions, but I'm fine with those being implemented in another PR.

I'm also okay with handling connection reset and broken stream errors in a separate PR, once you've got the testbench working - it might be worth a TODO for that though.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
jskeet: We do have back-off code in GAX, but it's not easy to integrate into this particular code, where the only state we have is the `CurrentFailedTry`. (GAX code tends to assume that the retry sequence can be maintained between calls.)

I suspect it's simpler to stick to this effective duplication but simple integration than to try to somehow use the GAX code.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
anshuldavid13: Hi Chris, yes I had discussed this with Jon as well and we decided to wait till we can test it with testbench to see if its needed for dotnet and not handled by default. I have documented this for now in the documentation : https://docs.google.com/document/d/14EjYUGrEkize4rZQ9j4PuMIrrPOyJNwYqhhWkrP8Cbw/edit?usp=sharing&resourcekey=0-4ks2t_XktfR0vL8Ak39REQ

PR #7899, opened 2022-01-27
Pass the service config file into the generator

jskeet:
A few notes:

- This requires the latest (unreleased at the time of writing) version of the generator
- It currently fails for Google.Cloud.Iap.V1 as there's an additional API in the service config file; we're working out what to do about that
- I don't know why quite a few APIs didn't have a service config file in apis.json, but I've fixed them. A couple genuinely don't have service config files - I've used "none" to make it clear that's deliberate. (Generation will fail if the value doesn't exist at all.)
- The APIs which change the service names mid-generation need to change the service config file too

With a manual fix for IAP, and using the latest (unreleased generator, generation itself succeeds with this change. I have not validated that the results build or are correct, however.


amanda-tarafa (approved):
LGTM

jskeet:
This is replaced by #8147

Issue #7897, opened 2022-01-27
Google.Cloud.Stoage.V1 DownloadObjectAsync doesn't provide content version information


lukepuplett:
**Is your feature request related to a problem? Please describe.**

I'm learning Google Cloud from a C#.NET world. The basic Cloud Storage for blobs has an unusual design. I'm using **Google.Cloud.Stoage.V1** package.

When I download an object's contents, I don't seem to be furnished with versioning information to go with it. This is the norm with web resources and I understand that the JSON API returns HTTP/1.1 ETags and supports standard HTTP conditional headers.

The library is either an awful design (perhaps it's not used much by C# devs and no one has ever told you), or I'm missing something :)

In order to get hold of version information to use in subsequent requests, I need to get it in a second request, by which time the object may have changed.

I can only assume the design is to force two requests, i.e. to get the "object" first (metadata), then use the version information on that to conditionally download its contents using if-match semantics. But I'm not sure what happens when the condition fails, since `DownloadObjectAsync()` has no return type to communicate anything with me.

Specifically I am talking about this method.

```csharp
    // Google.Cloud.Storage.V1.dll / StorageClient.cs
    public virtual Task DownloadObjectAsync(string bucket, string objectName, Stream destination, DownloadObjectOptions options = null, CancellationToken cancellationToken = default(CancellationToken), IProgress<IDownloadProgress> progress = null);
```

 **Describe the solution you'd like**

The `DownloadObjectAsync()` needs to return something to communicate a) whether request preconditions were met or not, and b) the ETag and generation number and any other salient information for the contents at the instant they were read at source.

 **Describe alternatives you've considered**

Make two requests, one for metadata and another conditional request for the contents, and retry the whole thing if the second request fails conditions. If the object is changing fast, this would cause twice the traffic on retries. It's weird.

 **Additional context**

I am new to Google Cloud. I am very old to C#. I haven't downloaded or even created a bucket yet, I'm reading and coding my first try with Google Storage and have come across this odd design.

lukepuplett:
In the meantime, can someone let me know what is the behaviour when a condition is not met? (and even better, update the docs). It would save me and others from spending an hour to test it so we know what code to write.

jskeet:
[...]

PR #7904, opened 2022-01-28
Prototype of recording content metadata while downloading

jskeet:
We won't do this until the next major version (as it's a breaking change) but this would address #7897.

This is really just a prototype to:

- Check that it's feasible (looks like it is)
- Provide a straw man of the general design to discuss

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/DownloadObjectTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/DownloadObjectTest.cs
@@ -45,7 +45,10 @@ namespace Google.Cloud.Storage.V1.IntegrationTests
         {
             using (var stream = new MemoryStream())
             {
-                await _fixture.Client.DownloadObjectAsync(_fixture.ReadBucket, _fixture.SmallObject, stream);
+                var metadata = await _fixture.Client.DownloadObjectAsync(_fixture.ReadBucket, _fixture.SmallObject, stream);
+                Assert.NotNull(metadata.Hash);
+                Assert.NotNull(metadata.Generation);
+                Assert.NotNull(metadata.Metageneration);
                 Assert.Equal(_fixture.SmallContent, stream.ToArray());
             }
         }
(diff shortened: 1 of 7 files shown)


lukepuplett:
Hi Jon, thanks for this. It looks good, and the return type provides an opportunity to expand if ever needed. The expansions that come to mind would be other information that the API/HTTP thinks is useful, e.g. Content-Length, Content-Type, Content-Encoding, Etag, and Last-Modified.

Content-Encoding may be a great or terrible choice. If the downloading of a gzipped blob automatically unzips it, then that information would become useless, incorrect to the programmer. However, if no such decompression occurs then the programmer will need this in order to deal with the stream appropriately.

Etag may also be useless if it's not possible to subsequently use the value in a future conditional request.

The only other thing that came to mind is the property name 'Hash'. Maybe it should be the same as it is in the `Google.Apis.Storage.v1.Data.Object` type, i.e. `Md5Hash` or kept as you have it, which is the same as the source header.

I guess the fear would be if it was `Md5Hash` the JSON API must never switch to another algorithm, or if it did, it'd need to also send the MD5 in another header. As it stands, the JSON API being equivocal on what exactly its "hash" header is, affords the potential to change. This may be extremely unlikely in practice.

I've no experience of maintaining a public C# API and the considerations for breaking changes, so this could be a daft idea, but I wonder if this could be a new method like `DownloadContentAsync` and the other one can be hidden from Intellisense (I don't know what tooling support is like for `EditorBrowsable`) and marked obsolete.

Thanks for your time and sharing early :)

lukepuplett:
I should add that these other HTTP response data may also place too strong a link between the C# API and its underlying transport. I don't know; just thoughts.

jskeet:
Thanks for the comments. It looks like Go already has this: https://pkg.go.dev/cloud.google.com/go/storage#ReaderObjectAttrs

... so we'll align with Go, possibly adding to (or clarifying) what they're returning.

In terms of the hash, this *isn't* the MD5 hash - it's the CRC32 hash, but the value includes the hash type ("crc32=..."). We could potentially make the property return a sequence of hash values, so that if the server ever starts returning multiple values (adding md5 as well as crc32 for example) clients could continue to work. But we'll get to that level of detail when we're closer to being able to merge this.

jskeet:
Closing this PR for now, but leaving the branch. We will almost certainly revive this in March/April as we approach the new major version.

PR #7911, opened 2022-02-01
docs: Add Google.Apis.Auth.AspNetCore3 for Cloud docs

jskeet:
As far as I can tell, it looks like using netcoreapp3.1 should be okay for generating docs. (It appears to work...)


amanda-tarafa (approved):
LGTM

PR #7882, opened 2022-01-24
feat: Add BigQueryBigNumeric type.

amanda-tarafa:
This commit only adds the type but it doesn't add support for it on the different BigQuery operations, that will come in a following commit.

PR #7921, opened 2022-02-04
perf: Improves performance of BigQuery numerics to decimal conversions.

amanda-tarafa:
Only the last commit is relevant here as the others are from #7882 .
Benchmarks are still running but I think this should show some improvement.


amanda-tarafa:
This is now rebased.

jskeet:
Closing for now as we don't have a clear benefit yet. Will leave the branch for further experimentation.

PR #7947, opened 2022-02-09
misc: test BigTable updates

viacheslav-rostovtsev:


jskeet:
Marking as "do not merge" as we really shouldn't merge this - aside from anything else, OwlBot will create a PR reverting it. I'm fine with it sticking around for testing purposes though.

jskeet:
I've run the integration tests with this PR, and they still pass.

jskeet:
I'm actually going to close this PR, as I'm expecting to create a single big PR with the new generator once it's released - I don't think we need to have this one for just Bigtable.

PR #7951, opened 2022-02-09
Experimental: Support routing headers in bidi streaming calls.

amanda-tarafa:
Uses https://github.com/googleapis/gax-dotnet/pull/527

Note that we lock alrady on the buffer, so an extra lock here for initialization shouldn't be a big deal.


jskeet:
Closing this PR for now; we will revisit when we know more about bidi/client-streaming routing requirements.

PR #8147, opened 2022-02-16
Regenerate with mixin support

jskeet:
Review one commit at a time, as normal


amanda-tarafa (approved):
LGTM! It looks really good!

PR #8190, opened 2022-02-22
chore: Install .NET Core 2.1 for testing

jskeet:
This used to be included by default, but isn't any more. We will
remove this in the next major version, when we can start testing
against .NET Core 3.1 as the lower bound.


amanda-tarafa (approved):
LGTM

PR #8198, opened 2022-02-23
Retry implementation

anshuldavid13:
[go/gcs-api-retry-guidance-dotnet](http://gcs-api-retry-guidance-dotnet/)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.CopyObject.cs
```diff
@@ -76,6 +76,10 @@ public sealed partial class StorageClientImpl : StorageClient
             options?.ModifyRequest(request);
             ApplyEncryptionKey(options?.EncryptionKey, options?.KmsKeyName, request);
             request.ModifyRequest += (options?.SourceEncryptionKey ?? EncryptionKey).ModifyRequestForRewriteSource;
+            if(options.IfGenerationMatch != null)
```
jskeet, line 79: Nit: space after `if`.

jskeet: Currently this will throw if `options` is null. I suggest you move the logic to the `ModifyRequest` method.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.DeleteObject.cs
```diff
@@ -64,6 +64,10 @@ private ObjectsResource.DeleteRequest CreateDeleteObjectRequest(string bucket, s
             GaxPreconditions.CheckNotNull(name, nameof(name));
             var request = Service.Objects.Delete(bucket, name);
             options?.ModifyRequest(request);
+            if(options.IfGenerationMatch != null || options.Generation != null)
```
jskeet, line 67: Again, I'd put this in the ModifyRequest method.

jskeet: (Same for all the conditional ones, basically.)

jskeet (commented):
First thoughts attached.

We also need tests - it would be great to be able to break them into four tests:

- Unconditional retry
- Conditional retry (condition not met)
- Conditional retry (condition met)
- No retry

That would make it easy to compare the tests with the implementation doc.

jskeet (commented):
I think it might be best to change this to a set of non-parameterized tests for now, each one self-contained, and then we can pair on how to parameterize it.

anshuldavid13:
Yet to do formatting and cleanup of the code.. Few test cases error to be fixed

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplTest.cs
```diff
@@ -69,5 +76,418 @@ public void GetEffectiveEncryptionKey_Invalid(EncryptionKey clientkey, Encryptio
...
+         * Also not sure where to add  IfGenerationMatch condition to test
+         * Error:
+         Assert.Equal() Failure
+        Expected: https://storage.googleapis.com/storage/v1/b/sourceBucket/o/sourceObject/copyTo/b/destinationBucket/o/destinationObject
+        Actual:   https://storage.googleapis.com/storage/v1/b/sourceBucket/o/sourceObject/rewriteTo/b/destinationBucket/o/destinationObject
+         */
+
+        [Fact]
+        public void CopyObject_NoRetry() =>
+      NoRetryHelper(service => service.Objects.Copy(new Object() { Id = "id", Name = "name" }, "sourceBucket", "sourceObject", "destinationBucket", "destinationObject"), client => client.CopyObject("sourceBucket", "sourceObject", "destinationBucket", "destinationObject"));
```
jskeet, line 271: We're calling RewriteObject the API to implement object copying, so this should be `service.Objects.RewriteObject(...)`

Not sure what you mean about "not sure where to add IfGenerationMatch condition to test" - that should be basically in the same way as other cases. (To specify it in the request you'll need `new ObjectsResource.RewriteRequest` instead of `service.Objects.RewriteObject`, so that you can use an object initializer.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplTest.cs
```diff
@@ -69,5 +76,418 @@ public void GetEffectiveEncryptionKey_Invalid(EncryptionKey clientkey, Encryptio
...
+        public void ListObject_RetryThenFail() =>
+    RetryThenFailHelper(service => service.Objects.List("bucket"), client => client.ListObjects("bucket").Count());
+
+        [Fact]
+        public void DeleteObject_NoRetry() =>
+    NoRetryHelper(service => service.Objects.Delete("bucket", "objectName"), client => client.DeleteObject("bucket", "objectName"));
+
+        // Need to create overloaded helper method for below operations to run:
+
+        /*
```
jskeet, line 299: Right. I'm suggesting that you have:

```csharp
private static void RetryOnceHelper<T>(Func<StorageService, ClientServiceRequest<T>> requestProvider, Action<StorageClient> clientAction,
   HttpStatusCode firstStatusCode,
   T response)
{
    var service = new FakeStorageService();
    var client = new StorageClientImpl(service);
    var request = requestProvider(service);

    service.ExpectRequest(request, firstStatusCode);
    service.ExpectRequest(request, new T());
    clientAction(client);
    service.Verify();
}

private static void RetryOnceHelper<T>(Func<StorageService, ClientServiceRequest<T>> requestProvider, Action<StorageClient> clientAction,
   HttpStatusCode firstStatusCode = HttpStatusCode.BadGateway)
   where T : new() => RetryOnce(requestProvider, clientAction, firstStatusCode, new T());
```

You can then call the version which accepts the response as the last parameter for `DeleteObject` where the response is just a string.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplTest.cs
```diff
@@ -69,5 +76,418 @@ public void GetEffectiveEncryptionKey_Invalid(EncryptionKey clientkey, Encryptio
...
+           NoRetryHelper(
+               service => new ObjectsResource.PatchRequest(service, new Object { Bucket = "bucket", Name = "object" }, "bucket", "object"),
+               client => client.PatchObject(new Object { Bucket = "bucket", Name = "object" }, new PatchObjectOptions()));
+
+        [Fact]
+        public void PatchObject_RetryThenFail() =>
+RetryThenFailHelper(service => new ObjectsResource.PatchRequest(service, new Object { Bucket = "bucket", Name = "object" }, "bucket", "object") { IfMetagenerationMatch = 70 },
+        client => client.PatchObject(new Object { Bucket = "bucket", Name = "object" }, new PatchObjectOptions { IfMetagenerationMatch = 70 }));
+
+        /*
```
jskeet, line 349: The tests you've got should be removed anyway. They're not testing the behaviour of `StorageClient`, which is what we're really trying to test.

The only StorageClient operation we've got that uses the Rewrite API is Copy, which you've got tests for above.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/FakeStorageService.cs
```diff
@@ -0,0 +1,76 @@
...
+using Google.Apis.Util;
+using Google.Cloud.ClientTesting;
+using System.Net;
+using System.Net.Http;
+
+namespace Google.Cloud.Storage.V1.Tests
+{
+    class FakeStorageService : StorageService
+    {
+        // TODO: Vary this content based on the actual error code?
```
jskeet, line 15: I think it would be worth turning this into a GetErrorContent method that changed the "code" part (and had a general message) so that it's not confusing when it shows up in a failure for a non-401 status code.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/FakeStorageService.cs
```diff
@@ -0,0 +1,76 @@
...
+            //MaybeDisablePrettyPrint(request);
+            var httpRequest = request.CreateRequest();
+            var responseMessage = new HttpResponseMessage(statusCode)
+            {
+                Content = new StringContent(SampleErrorContent)
+            };
+            handler.ExpectRequest(httpRequest.RequestUri, httpRequest.Content?.ReadAsStringAsync()?.Result, responseMessage);
+        }
+
+        private void MaybeDisablePrettyPrint(dynamic request)
```
jskeet, line 66: I don't think this is actually the case for storage, so let's remove the method (and the commented-out calls to it above).

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.HmacKeys.cs
```diff
@@ -106,6 +107,7 @@ private HmacKeysResource.UpdateRequest CreateUpdateHmacKeyRequest(HmacKeyMetadat
             GaxPreconditions.CheckArgument(key.AccessId != null, nameof(key), "Key must contain an access ID");
             var request = Service.Projects.HmacKeys.Update(key, key.ProjectId, key.AccessId);
             options?.ModifyRequest(request);
+            RetryHandler.MarkAsRetriable(request);
```
jskeet, line 110: I believe UpdateHmac is expected to be conditionally retriable.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplTest.cs
```diff
@@ -69,5 +76,402 @@ public void GetEffectiveEncryptionKey_Invalid(EncryptionKey clientkey, Encryptio
             var exception = Assert.Throws<ArgumentException>(() => client.GetEffectiveEncryptionKey(operationKey, kmsKey));
             Assert.Contains(expectedMessageSubstring, exception.Message);
         }
+
+        // Tests for GCS Retry :
```
jskeet, line 80: These comments were just off-the-cuff - I'd suggest either rewriting them more thoroughly, or deleting them.
(We could potentially make this StorageClientImplRetryTest.cs as a name as well, to separate it out from any other StorageClientImpl tests we write.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplTest.cs
```diff
@@ -69,5 +76,402 @@ public void GetEffectiveEncryptionKey_Invalid(EncryptionKey clientkey, Encryptio
...
+        // - No retry applied
+        // - Retry then success (sometimes with conditions)
+        // - Retry with eventual failure
+
+
+        private static void NoRetryHelper<T>(
```
jskeet, line 86: I would suggest moving all the helper methods to the end of the class, and adding an XML comment for each one.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplTest.cs
```diff
@@ -69,5 +76,402 @@ public void GetEffectiveEncryptionKey_Invalid(EncryptionKey clientkey, Encryptio
...
+            }
+            Assert.Throws<GoogleApiException>(() => clientAction(client));
+            service.Verify();
+        }
+
+        private static void RetryThenFailHelper<T>(Func<StorageService, ClientServiceRequest<T>> requestProvider, Action<StorageClient> clientAction,
+            HttpStatusCode errorCode = HttpStatusCode.BadGateway, int retryCount = 3)
+            where T : new() => RetryThenFail(requestProvider, clientAction, new T(), errorCode, retryCount);
+
+        // Bucket test cases BEGIN
```
jskeet, line 135: While I'm not a massive fan of regions, I think they'd be reasonable here - and allow someone to really easily collapse all the bucket test cases, etc.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplTest.cs
```diff
@@ -69,5 +76,402 @@ public void GetEffectiveEncryptionKey_Invalid(EncryptionKey clientkey, Encryptio
...
+        public void InsertBucket_NoRetry() =>
+            NoRetryHelper(service => service.Buckets.Insert(new Bucket { Name = "bucket" }, "project"), client => client.CreateBucket("project", "bucket"));
+
+        [Fact]
+        public void UpdateBucket_WithMetageneration_RetryOnce() =>
+            RetryOnceHelper(service => new BucketsResource.UpdateRequest(service, new Bucket { Name = "bucket" }, "bucket") { IfMetagenerationMatch = 70 }, 
+                client => client.UpdateBucket(new Bucket { Name = "bucket" }, new UpdateBucketOptions() { IfMetagenerationMatch = 70 }));
+
+        [Fact]
+        public void UpdateBucket_withoutMetgeneration_NoRetry() =>
```
jskeet, line 191: Nit: without => Without and Metgeneration => Metageneration (ditto for PatchBucket test below)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplTest.cs
```diff
@@ -69,5 +76,402 @@ public void GetEffectiveEncryptionKey_Invalid(EncryptionKey clientkey, Encryptio
...
+        public void GetObject_RetryOnce() =>
+            RetryOnceHelper(service => service.Objects.Get("bucket", "object"), client => client.GetObject("bucket", "object"));
+
+        [Fact]
+        public void GetObject_RetryThenFail() =>
+            RetryThenFailHelper(service => service.Objects.Get("bucket", "object"), client => client.GetObject("bucket", "object"));
+
+        [Fact]
+        public void CopyObject_NoRetry() =>
+            NoRetryHelper(service => service.Objects.Rewrite(new Object(), "sourceBucket", "sourceObject", "destinationBucket", "destinationObject"), client => client.CopyObject("sourceBucket", "sourceObject", "destinationBucket", "destinationObject"));
```
jskeet, line 288: I'd suggest splitting this line before the `client => ...` argument

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplTest.cs
```diff
@@ -69,5 +76,402 @@ public void GetEffectiveEncryptionKey_Invalid(EncryptionKey clientkey, Encryptio
...
+                client => client.UpdateObject(new Object { Bucket = "bucket", Name = "object" }));
+
+        [Fact]
+        public void UpdateObject_RetryThenFail() =>
+            RetryThenFailHelper(service => new ObjectsResource.UpdateRequest(service, new Object { Bucket = "bucket", Name = "object" }, "bucket", "object") { IfMetagenerationMatch = 70 },
+                client => client.UpdateObject(new Object { Bucket = "bucket", Name = "object" }, new UpdateObjectOptions { IfMetagenerationMatch = 70 }));
+
+        [Fact]
+        public void PatchObject_WithMetageneration_RetryOnce() =>
+            RetryOnceHelper(
```
jskeet, line 351: This format looks very readable to me - so let's apply it to all the tests above as well, to be consistent. (At least the ones where it doesn't all just fit on one line easily.

Another option would be to keep the method name on the same line as the test:

```csharp
[Fact]
public void PatchObject_WithMetageneration_RetryOnce() => RetryOnceHelper(
    service => new ObjectsResource.PatchRequest(service, new Object { Bucket = "bucket", Name = "object" }, "bucket", "object") { IfMetagenerationMatch = 70 },
    client => client.PatchObject(new Object { Bucket = "bucket", Name = "object" }, new PatchObjectOptions { IfMetagenerationMatch = 70 }));
```

jskeet (commented):
Mostly just nits now - getting there!

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/FakeStorageService.cs
```diff
@@ -0,0 +1,76 @@
+﻿using Google.Apis.Storage.v1;
```
jskeet, line 1: Forgot to mention before - this needs a copyright header.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/FakeStorageService.cs
```diff
@@ -44,33 +36,18 @@ class FakeStorageService : StorageService
...
         {
-            // TODO: Validate that this isn't a success status code (given that we're returning an error response).
-
-            //MaybeDisablePrettyPrint(request);
             var httpRequest = request.CreateRequest();
-            var responseMessage = new HttpResponseMessage(statusCode)
-            {
-                Content = new StringContent(SampleErrorContent)
-            };
+            var responseMessage = new HttpResponseMessage(statusCode);
```
jskeet, line 47: This now appears to be back to the original "not having any content at all" - which made it hard to understand the error. It's okay while the tests are passing, but it will be confusing on some failures. I'd create a method to return an appropriate StringContent for the given status code - but that could come after this PR.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplRetryTest.cs
```diff
@@ -0,0 +1,459 @@
...
+            service => service.Buckets.Insert(new Bucket { Name = "bucket" }, "project"), 
+            client => client.CreateBucket("project", "bucket"));
+
+        [Fact]
+        public void UpdateBucket_WithMetageneration_RetryOnce() => RetryOnceHelper(
+            service => new BucketsResource.UpdateRequest(service, new Bucket { Name = "bucket" }, "bucket") { IfMetagenerationMatch = 70 }, 
+            client => client.UpdateBucket(new Bucket { Name = "bucket" }, new UpdateBucketOptions() { IfMetagenerationMatch = 70 }));
+
+        [Fact]
+        public void UpdateBucket_WithoutMetgeneration_NoRetry() => NoRetryHelper(
```
jskeet: This is another "Metgeneration" - it looks like there are still three of them...

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/StorageClientImplRetryTest.cs
```diff
@@ -0,0 +1,459 @@
...
+            RetryThenFail(service => service.Notifications.Delete("bucket", "notification"), client => client.DeleteNotification("bucket", "notification"), "check");
+
+        #endregion
+
+        #region Helper Methods
+
+        /// <summary>
+        /// Helper method to check in case of non retriable error codes
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
```
jskeet, line 369: Please either add useful comments here or remove the elements - likewise for the other XML comments below.

jskeet (approved):
Just a few formatting nits - and a suggestion around FakeStorageService, but that could be done later.

Before merging, we'll want to get this down to a useful set of commits. Do you want it to be a single commit, or split it up? (I think a single commit would probably be fine here.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/FakeStorageService.cs
anshuldavid13: Sure we can put it in the next PR

tritone (commented):
Thanks for your work on this, and sorry for the belated response to the PR (I was OOO last week). Just wanted to confirm-- it looks like this covers marking requests as retryable or not based on idempotency for most requests, though not for media operations yet. Is that an accurate summary of where things stand at this point?

jskeet:
@tritone: Yes, that's correct. Media operations may well be rather more involved given that almost all of the media handling happens in the support library (Google.Apis and Google.Apis.Core).

Issue #8201, opened 2022-02-24
Consider dropping internal Spanner logging and using normal MS logging


jskeet:
When the Spanner libraries first came out, the Microsoft.Extensions.Logging libraries were relatively new, and it wasn't clear how well adopted they'd be. It's likely that we'll depend on them in the next major version of GAX though.

I think it would be reasonable to drop the internal logging library from Spanner now.

I'll look into this in terms of feasibility.

amanda-tarafa:
+1 to this.

jskeet:
We've decided not to take the breaking change in the end, for the next major version - as the change didn't feel clean enough. We didn't have a good story on how to inject suitable loggers etc, and we didn't want to block the next version on it.

Instead, we'll write an implementation of the Spanner internal logger that delegates to an `ILogger`, and include that in the next major version as a way of at least allowing logging from Spanner. We'll then look at removing the internal logger in the future.

jskeet:
The delegating code was in #8448. We should add some documentation for this when we're doing a documentation overhaul for the new major version.

PR #8220, opened 2022-03-03
Prototype of moving to ILogger in Spanner

jskeet:
Notes:

The explicit package reference to
Microsoft.Extensions.Logging.Abstractions in Spanner.V1 will be
removed for GAX v4.

Further work to move ExecuteHelper out of the Internal namespace
would be beneficial, and really shouldn't be too hard.

Currently this makes it hard to set a default logger. It's global
state which is convenient but ugly. It wouldn't be too hard to
introduce another place for the default SessionPoolManager's logger
to be set, if we wanted to.

Currently the logging is all based on SessionPoolManager - that's
effectively the root of the logging tree. If we wanted to allow
different SpannerConnections to use different loggers for example,
that might be tricky.

Specifying a SessionPoolManager which creating a new
SpannerConnection is tricky, requiring a
SpannerConnectionStringBuilder... if we decide that it's fine to
require users to create a new SessionPoolManager for custom logging,
we might want to add a convenience overload for SpannerConnection
that accepts a string and a SessionPoolManager.

We now unconditionally log SQL at a Debug level, where previously we
had an option for whether or not to log sensitive information. We
could put that option into SpannerConnection potentially.

This change removes the performance counters we had - but there were
very few of them, and I'd argue that tying them to logging probably
isn't ideal anyway.

There's also one test for commit stats that I haven't rewritten yet
- it would be fairly fiddly to do, but feasible.


jskeet:
This would address #8201

amanda-tarafa (commented):
Yep, this looks good.

I think, that we can add DI extension methods for letting users configure the session pool managar in general and in particular logging. And in that case we can have SpannerConnection  receive a specific logger on the constructor even if they are all using the same session pool manager.
We'll have to write some extra constructors or factory methods that we can make public to expose the configuration functionality to users not using DI or using a different container than MS's?

jskeet:
A few notes as I look over this again...

- We could potentially use an `ILoggerFactory` instead of an `ILogger` in `SessionPoolManager` and `SessionPool`
- Acquiring a session could specify the logger for that session to use as an optional override of the one in the `SessionPool` - that would allow per-connection logging

jskeet:
Closing this as we're somewhat nervous about it - we'd prefer to get more information about how customers use the new `ILogger` support, and do a "better" version of this in the future.

PR #8249, opened 2022-03-15
feat: Prototype PR for adding PgNumeric support

Rishabh-V:
This is the prototype PR to get early feedback on the approach for adding PgNumeric support in Spanner.


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+            }
+
+            // if scale is same, we can compare the integer value
+            if (_scale == other._scale)
+            {
+                return _integer.CompareTo(other._integer);
+            }
+
+            // return comparision of normalized value.
+            return Normalize(this).NormalizedValue.CompareTo(Normalize(other).NormalizedValue);
```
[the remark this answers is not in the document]
amanda-tarafa: Actually we might need to "normalize" to MaxScale always, so that we have a consistent way of calculating the HashCode, else Equals and HashCode won't match.

We chatted yesterday and we said that maybe storing the numbers in MaxScale always might hit performance as we would, for the most part, carry a lot of zeros, around order of 10^4. I still think that's a bad idea. But if for HashCode we need to normalize to MaxScale then not storing the number fully scaled in the first place solves very little.

An alternative is to represent the number slightly different:
- BigInteger always without trailing zeros
- Scale that accepts negative values

Then we don't carry extra zeros around. HashCode is trivial, i.e. the combination of the parts, and comparison can still be done by temporarily scaling to the biggest scale of the two operands as my code above. @jskeet what do you think?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+        }
+
+        /// <summary>
+        /// Removes trailing zeros on the value and returns normalized value and scale.
+        /// Normalized value of 1.000, 1.00, 1.0, 1 is 1 and Normalized scale is 3, 2, 1, 0 respectively.
+        /// Noemalized value of 10.000, 10.00, 10.0, 10 is 10 and Normalized scale is 3, 2, 1, 0 respectively.
+        /// </summary>
+        /// <param name="pgNumeric">The PgNumeric value to be normalized.</param>
+        /// <returns>The normalized value and scale.</returns>
+        private static (BigInteger NormalizedValue, int NormalizedScale) Normalize(PgNumeric pgNumeric)
```
amanda-tarafa, line 185: If this remains useful after review, it should be called TrimTrailingZeros or something similar.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+                text = text.Remove(pointIndex, 1);
+                scale = decimalPlaces;
+                rawValue = BigInteger.Parse(text, CultureInfo.InvariantCulture);
+            }
+            else
+            {
+                // We don't have '.'
+                rawValue = BigInteger.Parse(text, CultureInfo.InvariantCulture);
+            }
+            if (!IsRawValueInRange(rawValue))
```
amanda-tarafa, line 299: Again, scale is relevant here as we are not scaling the integer to the max scale possible.

amanda-tarafa (changes requested):
I've left several comments. I haven't reviewed conversions of PgNumeric because I think we'll have to change the internal represantation of the number and that will affect those.

I think also that this PR can be split in two or three commits:
- The first commit, let's forget about PgNumeric and only make changes around the new TypeAnnotation in the Type proto, as there are some implications there for the ADO.NET interface.
- The second commit, adds PgNumeric type only, i.e. the file PgNumeric.cs
- The third commit makes all the remaining changes so that PgNumeric is supported.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+        /// </summary>
+        /// <param name="other">The value to compare with this one</param>
+        /// <returns>A negative integer if this value is less than <paramref name="other"/>; a positive integer
+        /// if it's greater than <paramref name="other"/>; zero if they're equal.</returns>
+        public int CompareTo(PgNumeric other)
+        {
+            // Check for NaN
+            if (_notANumber)
+            {
+                return other._notANumber ? 0 : 1;
```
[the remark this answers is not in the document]
Rishabh-V: In order to allow numeric values to be sorted and used in tree-based indexes, PostgreSQL treats NaN values as equal, and greater than all non-NaN values. [Source](https://www.postgresql.org/docs/14/datatype-numeric.html#:~:text=In%20order%20to%20allow%20numeric%20values%20to%20be%20sorted%20and%20used%20in%20tree%2Dbased%20indexes%2C%20PostgreSQL%20treats%20NaN%20values%20as%20equal%2C%20and%20greater%20than%20all%20non%2DNaN%20values.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
```diff
@@ -563,6 +592,29 @@ private object ConvertToClrTypeImpl(Value wireValue, System.Type targetClrType,
...
+                if (TypeCode != TypeCode.Numeric)
+                {
+                    throw new ArgumentException($"{targetClrType.FullName} can only be used for numeric results");
+                }
+                switch (wireValue.KindCase)
+                {
+                    case Value.KindOneofCase.NullValue:
+                        return null;
+                    case Value.KindOneofCase.StringValue:
+                        if (string.Compare(wireValue.StringValue, "NaN", StringComparison.OrdinalIgnoreCase) == 0)
```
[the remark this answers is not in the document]
Rishabh-V: "NaN" is used to represent "NaN". It is case insensitive in input [Documentation](https://www.postgresql.org/docs/8.1/datatype.html#:~:text=On%20input%2C%20the%20string%20NaN%20is%20recognized%20in%20a%20case%2Dinsensitive%20manner.)

Rishabh-V (commented):
Removed breaking changes and updated scaling, normalized value and GetHashCode methods

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -125,7 +137,7 @@ private SpannerDbType(TypeCode typeCode, int? size = null)
         public int? Size { get; }
 
         private SpannerDbType(TypeCode typeCode, SpannerDbType arrayElementType)
-            : this(typeCode) => ArrayElementType = arrayElementType;
+            : this(typeCode, arrayElementType.TypeAnnotationCode) => ArrayElementType = arrayElementType;
```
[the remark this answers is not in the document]
Rishabh-V: Apart from PgNumeric, TypeAnnotationCode of all other types would be Unspecified. Only in case of array of PgNumeric being constructed that TypeAnotationCode.PgNumeric would be passed. In all other cases, Unspecified annotation would be passed.

Rishabh-V (commented):
Made few of the changes and reverted the breaking changes. Will discuss about handling the breaking change in our sync up today. Thanks.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+        /// The maximum precision of numeric type.
+        /// </summary>
+        private const int MaxPrecision = MaxDigitsBeforeDecimalPoint + MaxScale + 1;
+
+        /// <summary>
+        /// The string constant for not a number (NaN).
+        /// </summary>
+        private const string NotANumber = "NaN";
+
+        /// For a numeric(precision, scale) maximum and minimum values are defined as +- ((10^precision)-1)/(10^scale), so maximum possible numeric would be numeric
```
jskeet, line 50: Either make these `//`, or add the XML document tags.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+            {
+                throw new ArgumentException("Invalid value and scale provided to PgNumeric for NaN");
+            }
+
+            _scale = scale;
+            _notANumber = isNotANumber;
+            _scaledValue = integer;
+        }
+
+        private static bool IsRawValueInRange(BigInteger integer, int scale)
```
jskeet, line 115: It's not clear to me that this is correct. I think we need to talk about it more.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+        {
+            // Check for NaN.
+            // In order to allow numeric values to be sorted and used in tree-based indexes,
+            // PostgreSQL treats NaN values as equal, and greater than all non-NaN values.
+            if (_notANumber)
+            {
+                return other._notANumber ? 0 : 1;
+            }
+
+            // if scale is same, we can compare the integer value.
```
jskeet, line 138: Nit: capital letter to start the sentence, and on the line four lines down.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+        /// Read here: https://stackoverflow.com/questions/26396856/determine-postgres-numeric-max-min-values       
+        private static readonly BigInteger s_maxValueWithoutScale = BigInteger.Pow(10, MaxPrecision) - 1;
+
+        private static readonly BigInteger s_maxValueWithScale = ((BigInteger.Pow(10, MaxDigitsBeforeDecimalPoint) - 1) / (BigInteger.Pow(10, MaxScale)));
+
+        private static readonly BigInteger s_minValueWithScale = -s_maxValueWithScale;
+
+        private static readonly BigInteger s_minValueWithoutScale = -s_maxValueWithoutScale;
+
+        private static readonly Dictionary<int, BigInteger> s_powersOf10 = new Dictionary<int, BigInteger>();
```
jskeet, line 61: This should be a ConcurrentDictionary.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+        public bool Equals(PgNumeric other) => CompareTo(other) == 0;
+
+        /// <summary>
+        /// Gets the normalized value of the PgNumeric instance.
+        /// </summary>
+        /// <returns>The normalized value as big integer.</returns>
+        public BigInteger GetNormalizedValue()
+        {
+            int differenceToMaxScale = MaxScale - _scale;
+            if (!s_powersOf10.TryGetValue(differenceToMaxScale, out BigInteger powerOf10))
```
jskeet, line 179: Once s_powersOf10 is a ConcurrentDictionary, use GetOrAdd.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+            }
+
+            return _scaledValue * powerOf10;
+        }
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj) => obj is PgNumeric other && Equals(other);
+
+        /// <inheritdoc/>
+        public override int GetHashCode() => GetNormalizedValue().GetHashCode() + _notANumber.GetHashCode();
```
jskeet, line 192: I'd suggest using a fixed "large and negative" value for the hash code of `_notANumber`, otherwise it'll end up being 1 which is more likely to cause hash collisions.

It's a bit of a shame that we need to call GetNormalizedValue here, when hashing should be cheap.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+        /// <inheritdoc/>
+        public bool Equals(PgNumeric other) => CompareTo(other) == 0;
+
+        /// <summary>
+        /// Gets the normalized value of the PgNumeric instance.
+        /// </summary>
+        /// <returns>The normalized value as big integer.</returns>
+        public BigInteger GetNormalizedValue()
+        {
+            int differenceToMaxScale = MaxScale - _scale;
```
jskeet, line 178: Shortcut to return `_scaledValue` if `_scale` is 0?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+            if (sign == 0)
+            {
+                return "0";
+            }
+
+            var integerText = _scaledValue.ToString(CultureInfo.InvariantCulture);
+            int signLength = sign > 0 ? 0 : 1;
+
+            // We can have a scaled value or a non-scaled value, check if its scaled. 
+            if (_scale != 0)
```
jskeet, line 330: I'd suggest changing this to:

```csharp
if (_scale == 0)
{
    return integerText;
}
```

... then the rest of the code doesn't need to be indented, and it makes it clearer that that's basically an optimization path.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+        }
+
+        // Conversions from CLR types to PgNumeric
+
+        /// <summary>
+        /// Explicit conversion from <see cref="Decimal"/> to <see cref="PgNumeric"/>.
+        /// </summary>
+        /// <param name="value">The decimal value to convert.</param>
+        /// <returns>The result of the conversion.</returns>
+        public static explicit operator PgNumeric(decimal value) =>
```
jskeet, line 402: Unwrap, here and below? Or wrap the `int` conversion, for consistency.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+    /// <summary>
+    /// Representation of PostgreSQL numeric type which has arbritrary precision and scale. 
+    /// The PostgreSQL numeric type has max precision of 147,455 and a max scale of 16383. 
+    /// </summary>
+    public struct PgNumeric : IEquatable<PgNumeric>, IComparable<PgNumeric>, IComparable
+    {
+        /// <summary>
+        /// The maximum number of digits before decimal point.
+        /// </summary>
+        private const int MaxDigitsBeforeDecimalPoint = 131072;
```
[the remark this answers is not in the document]
jskeet: Ah - no, I now see the requirement in a chat message. Urgh.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+        /// The maximum valid value for <see cref="PgNumeric"/>, equal to BigInteger.Pow(10, MaxPrecision) - 1
+        /// </summary>
+        public static PgNumeric MaxValue { get; } = new PgNumeric(s_maxValueWithoutScale, 0, false);
+
+        /// <summary>
+        /// The minimum valid value for <see cref="PgNumeric"/>, equal to -(BigInteger.Pow(10, MaxPrecision) - 1)
+        /// </summary>
+        public static PgNumeric MinValue { get; } = new PgNumeric(s_minValueWithoutScale, 0, false);
+
+        // The scaled value of PgNumeric would be stored in this variable.
```
jskeet, line 89: I think the "would be" in these comments is a bit confusing. That *is* the value.

However, more clarity around what "scaled value" means would definitely be useful. Which direction is this scaling in?

For example, if _scaledValue is 123 and _scale is 1, does that mean the value is 12.3? I suspect so, but just putting that simple example here would really help.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
jskeet: Even now I've seen the chat comment about a maximum number of decimal digits, I'm not sure this is actually correct. It would be quite odd for the maximum values to be a step function like this rather than something that changes by a factor of 10 per change of 1 in scale.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,487 @@
...
+
+        /// <summary>
+        /// The string constant for not a number (NaN).
+        /// </summary>
+        private const string NotANumber = "NaN";
+
+        /// For a numeric(precision, scale) maximum and minimum values are defined as +- ((10^precision)-1)/(10^scale), so maximum possible numeric would be numeric
+        /// with max precision and zero scale, i.e. numeric(147455, 0), and its max and min values would be ((BigInteger.Pow(10, 147455)-1)/(BigInteger.Pow(10, 0)))
+        /// Read here: https://stackoverflow.com/questions/26396856/determine-postgres-numeric-max-min-values       
+        private static readonly BigInteger s_maxValueWithoutScale = BigInteger.Pow(10, MaxPrecision) - 1;
```
jskeet, line 53: Unless there's some special-casing of a scale of 0, I'd actually expect this to be `BigInteger.Pow(10, MaxDigitsBeforeDecimalPoint) - 1` - the maximum integer value has that many digits.

Rishabh-V (commented):
Hey Jon, Amanda, I have fixed most of the review comments. Only refactoring ones are pending for which I will raise a new PR after PgNumeric. I have also raised a new PR with just PgNumeric and its tests.  I am still uncertain about Max and min values and solicit your feedback.

amanda-tarafa:
@Rishabh-V I think it's best to wait for PGNumeric to be ready before re-reviewing this one, and also so you have refactored this PR in two separate commits, one for the changes concerning the introduction of TypeAnnotation, and one for actually adding support for PgNumeric across the Spanner lib.

Let me know if this sounds OK.

Rishabh-V:
> @Rishabh-V I think it's best to wait for PGNumeric to be ready before re-reviewing this one, and also so you have refactored this PR in two separate commits, one for the changes concerning the introduction of TypeAnnotation, and one for actually adding support for PgNumeric across the Spanner lib.
> 
> Let me know if this sounds OK.

Sure Amanda. I have fixed the review comments for PgNumeric. Should I raise a new PR for TypeAnnotation and PgNumeric support commits or can I add 2 commits to the PgNumeric PR itself? Please let me know and I will make the changes.  Thanks.

amanda-tarafa:
I think we can now close this PR in favour of #8261

Rishabh-V:
Closing this PR as all the changes were done in PR #8261 and its merged now.

PR #8261, opened 2022-03-18
feat:PgNumeric and its unit tests.

Rishabh-V:
Just PgNumeric type with its unit tests. I am not certain about about Max and Min values, so solicit your feedback on max and min values.


amanda-tarafa (changes requested):
Note that:
- We are now trimming trailing zeros after decimal so we know that we are not counting "extra" scales.
- This means that for equality (i.e. hashcode) we don't need to scale anymore.
- For comparison we need only scale to the biggest scale of the two operands.

I'm leaving specific comments that I believe simplify things taking into account the points above.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,497 @@
...
+        private const int MaxPrecision = MaxDigitsBeforeDecimalPoint + MaxScale;
+
+        /// <summary>
+        /// The string constant for not a number (NaN).
+        /// </summary>
+        private const string NotANumber = "NaN";
+
+        // For a numeric(precision, scale) maximum and minimum values are defined as +- ((10^precision)-1)/(10^scale), so maximum possible numeric would be numeric
+        // with max precision and zero scale, i.e. numeric(147455, 0), and its max and min values would be ((BigInteger.Pow(10, 147455)-1)/(BigInteger.Pow(10, 0)))
+        // Read here: https://stackoverflow.com/questions/26396856/determine-postgres-numeric-max-min-values
```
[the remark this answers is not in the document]
Rishabh-V: I kept it for now so that the rational of max value is clear few months down the line. Please let me know if its not needed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+using System;
+using System.Collections.Generic;
+using System.Numerics;
+using Xunit;
+
+namespace Google.Cloud.Spanner.V1.Tests
+{
+    public class PgNumericTests
+    {
+        public static IEnumerable<object[]> GetRawValueAndScales()
```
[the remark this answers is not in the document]
amanda-tarafa: Also, if I suggested on the PgNumeric class, we stop testing RawValueInRange directly, don't get rid of these, instead use them for testing Parsing etc.
You can probably write the big strings using char enumerables, something like Enumerable.Repeat('9', 147455).ToArray().....

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+            yield return new object[] { -((BigInteger.Pow(10, 131072) - 1)/BigInteger.Pow(10, 16383)), 16383, true };
+
+            yield return new object[] { ((BigInteger.Pow(10, 131072) - 1) / BigInteger.Pow(10, 16383)), 16384, false };
+            yield return new object[] { -((BigInteger.Pow(10, 131072) - 1) / BigInteger.Pow(10, 16383)), 16384, false };
+
+            yield return new object[] { BigInteger.Pow(10, 147455) - 1, 16383, true };
+            yield return new object[] { -BigInteger.Pow(10, 147455) + 1, 16383, true };
+        }
+
+        public static IEnumerable<object[]> GetStringInputs()
```
amanda-tarafa, line 44: Split the data and the tests in two, one for success and one for failure. On the failure test you can use Assert.ThrowsException...

Also, avoid using PgNumeric.ToString() to build the data as we could potentially have the same bug on ToString and Parse. Use the Enumerable approach I descibed above to build the strings.

amanda-tarafa: And als add more general type cases, i.e. normal numbers 1 or 2 only, but don't test only the extremes.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        [Theory]
+        [InlineData("NaN", "NaN")]
+        [InlineData("Nan", "nan")]
+        [InlineData("0.00", "000")]
+        [InlineData("-1.0", "-1.000")]
+        [InlineData("-1", "-1.0")]
+        [InlineData("1", "1.0")]
+        [InlineData("0.00000009", "0.00000009")]
```
amanda-tarafa, line 62:
```suggestion
        [InlineData("0.00000009", "0.00000009")]
        [InlineData("0.0000000900000", "0.00000009")]
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        [InlineData("-10.0", "0.00000009000000000")]
+        [InlineData("0", "1.00000000")]
+        [InlineData("1", "1.01")]
+        [InlineData("0.00000009", "0.0000008")]
+        public void CompareToGreaterValueTest(string input1, string input2)
+        {
+            var number1 = PgNumeric.Parse(input1);
+            var number2 = PgNumeric.Parse(input2);
+            var result = number1.CompareTo(number2);
+            Assert.Equal(-1, result);
```
amanda-tarafa, line 85: Let's not rely on it being -1, just negative which is what the spec says.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+
+        [Theory]
+        [InlineData("9999999999999", "NaN")]
+        [InlineData("0.00", "001")]
+        [InlineData("-1.0", "1.000")]
+        [InlineData("-10.0", "0.00000009000000000")]
+        [InlineData("0", "1.00000000")]
+        [InlineData("1", "1.01")]
+        [InlineData("0.00000009", "0.0000008")]
+        public void CompareToGreaterValueTest(string input1, string input2)
```
amanda-tarafa, line 80:
```suggestion
        public void CompareToLessThanTest(string input1, string input2)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        }
+
+        [Theory]
+        [InlineData("NaN", "99999999999999999999999999999")]
+        [InlineData("99.99", "99.90")]
+        [InlineData("1.0", "-1.000")]
+        [InlineData("-1", "-55.0")]
+        [InlineData("0", "-0.0009")]
+        [InlineData("0.00009", "0.00000009")]
+        public void CompareToSmallerValueTest(string input1, string input2)
```
amanda-tarafa, line 95:
```suggestion
        public void CompareToGreaterThanTest(string input1, string input2)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        public void CompareToEqualValueTest(string input1, string input2)
+        {
+            var number1 = PgNumeric.Parse(input1);
+            var number2 = PgNumeric.Parse(input2);
+            var result = number1.CompareTo(number2);
+            Assert.Equal(0, result);
+        }
+
+        [Theory]
+        [InlineData("9999999999999", "NaN")]
```
amanda-tarafa, line 73: Nan should be compared aginst maxValue and minValue, and zero, and a negative value as well.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        [InlineData("0.00", "001")]
+        [InlineData("-1.0", "1.000")]
```
amanda-tarafa, line 75:
```suggestion
        [InlineData("-1.0", "1.000")]
                [InlineData("-1.0", "0")]
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+            Assert.Equal(true, result);
+        }
+
+        [Theory]
+        [InlineData("1.00", "1")]
+        [InlineData("1", "1.00")]
+        [InlineData("NaN", "NaN")]
+        [InlineData("1.0000", "1.000")]
+        [InlineData("999999999999999999999999999999999999999999999999.999999999999999999999999999999999999999", "999999999999999999999999999999999999999999999999.999999999999999999999999999999999999999")]
+        [InlineData("-999999999999999999999999999999999999999999999999.999999999999999999999999999999999999999", "-999999999999999999999999999999999999999999999999.999999999999999999999999999999999999999")]
```
amanda-tarafa, line 126: These would benefit from instead of using nines, using 12345678912345.... it's easier to count the number blocks than a lot of nines.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+                Assert.Equal(true, expectedParsingResult);
+            }
+            catch (FormatException)
+            {
+                Assert.Equal(false, expectedParsingResult);
+            }
+        }
+
+        [Theory]
+        [InlineData(null)]
```
amanda-tarafa, line 173: This is not a theory

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        [InlineData("-Infinity")]
+        [InlineData("-inf")]
+        [InlineData("ABCDEFGH")]
+        [InlineData("1.@#$%^&*")]
+        [InlineData("1.0!")]
+        [InlineData("1.00D")]
+        [InlineData("1.00M")]
+        [InlineData("1.00F")]
+        // Space before and/or after valid number. We can call .Trim() on input, but should we? Its an invalid input.
+        // Also, SpannerNumeric doesn't do that, so its consistent. We should call Trim() on both or on neither.
```
amanda-tarafa, line 194: No we don't, you can remove this.
Also test for ".0", ".123456", "-.123456"

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        [InlineData("1.00")]
+        [InlineData("1.0")]
+        [InlineData("NaN")]
+        [InlineData("1.0000")]
+        [InlineData("-1.00")]
+        [InlineData("-0.00")]
+        [InlineData("0.0000000000000009")]
+        [InlineData("-0.0000000000000009")]
+        [InlineData("9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999990.99999999999999999999999999999999999999999999")]
+        [InlineData("-9999999999999999999999999999999999999999999999999999999999999999999999999999999999999999999990.99999999999999999999999999999999999999999999")]
```
amanda-tarafa, line 211: 1234567891.......

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        public void TryParseValidTest(string input)
+        {
+            var success = PgNumeric.TryParse(input, out PgNumeric number);
+            Assert.IsType<PgNumeric>(number);
+            Assert.NotNull(number.ToString());
+            Assert.True(success);
+        }
+
+        [Theory]
+        [InlineData(null)]
```
amanda-tarafa, line 221: not a theory

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,325 @@
...
+        [InlineData("-Infinity")]
+        [InlineData("-inf")]
+        [InlineData("ABCDEFGH")]
+        [InlineData("1.@#$%^&*")]
+        [InlineData("1.0!")]
+        [InlineData("1.00D")]
+        [InlineData("1.00M")]
+        [InlineData("1.00F")]
+        [InlineData(" 1.00  ")]
+        public void TryParseInValidTest(string input)
```
amanda-tarafa, line 243: The equivalent tests for parse and try parse should use the same data

amanda-tarafa (changes requested):
Looking better.
Also please check that with testing you are covering all cases that are covered in [BigQuery](https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryBigNumericTest.cs)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
Rishabh-V: > Split the data and the tests in two, one for success and one for failure. On the failure test you can use Assert.ThrowsException...
>   There is already a Valid and InValid test for Parse and TryParse that does that.
 
> Also, avoid using PgNumeric.ToString() to build the data as we could potentially have the same bug on ToString and Parse. Use the Enumerable approach I descibed above to build the strings.

PgNumeric.ToString() is just one of the input data. The corresponding input is also provided using BigInteger and have also added Enumerable approach as well as suggested by you.
Also added general test cases.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,473 @@
...
+            int signLength = sign > 0 ? 0 : 1;
+
+            // Always have something before the period, even if it's just a 0.
+            int originalIntegerPlaces = Math.Max(integerText.Length - signLength - _scale, 0);
+            int resultIntegerPlaces = Math.Max(originalIntegerPlaces, 1);
+
+            // Sign, then integer part, then period (maybe), then significant decimals.
+            int resultLength = signLength + resultIntegerPlaces + (decimalPlaces > 0 ? 1 : 0) + decimalPlaces;
+            var chars = new char[resultLength];
+            int position = 0;
```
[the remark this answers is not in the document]
Rishabh-V: Have done few of the changes which keeps the test passing.

Rishabh-V (commented):
Fixed review comments

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,514 @@
...
+            var number1 = PgNumeric.Parse(input1);
+            var number2 = PgNumeric.Parse(input2);
+            var result = number1.CompareTo(number2);
+            Assert.Equal(expectedResult, Math.Sign(result));
+        }
+
+        [Fact]
+        public void CompareTo_NegativeZero()
+        {
+            // These really are the same value... we don't differentiate.
```
amanda-tarafa, line 157: No one differentiates :)
```suggestion
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,514 @@
...
+            Assert.Equal(0, (IComparable)regularZero.CompareTo(negativeZero));
+            Assert.Equal(0, negativeZero.CompareTo(regularZero));
+            Assert.Equal(0, (IComparable)negativeZero.CompareTo(regularZero));
+        }
+
+        [Fact]
+        public void CompareTo_Null()
+        {
+            IComparable minValue = PgNumeric.MinValue;
+            Assert.InRange(minValue.CompareTo(null), 1, int.MaxValue);
```
amanda-tarafa, line 170: Use Math.Sign is for consistency here.
```suggestion
            Assert.Equal(1, Math.Sign(minValue.CompareTo(null)));
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,514 @@
...
+        {
+            var number1 = PgNumeric.Parse(input1);
+            var number2 = PgNumeric.Parse(input2);
+            var result = number1.CompareTo(number2);
+            Assert.Equal(0, result);
+        }
+
+        [Theory]
+        [MemberData(nameof(GetComparisionData))]
+        public void CompareToTest(string input1, string input2, int expectedResult)
```
amanda-tarafa, line 146: If this is just for comparing NaN, the the method should be called CompareWithNan and the data member should be called GetNaNComparisonData or similar.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,514 @@
...
+        public void CompareToLessThanTest(string input1, string input2)
+        {
+            var number1 = PgNumeric.Parse(input1);
+            var number2 = PgNumeric.Parse(input2);
+            var result = number1.CompareTo(number2);
+            Assert.True(result < 0);
+        }
+
+        [Theory]
+        [InlineData("NaN", "99999999999999999999999999999")]
```
amanda-tarafa, line 194: There should be as many cases here as in the method above.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,514 @@
...
+        [InlineData("-0.00")]
+        [InlineData("0.0000000000000009")]
+        [InlineData("-0.0000000000000009")]
+        [InlineData("1234567891234567891234567891234567891234567891234567891234567891234567891234567891234567891230.12345678912345678912345678912345678912345678")]
+        [InlineData("-1234567891234567891234567891234567891234567891234567891234567891234567891234567891234567891230.12345678912345678912345678912345678912345678")]
+        public void ParseValidTest(string input)
+        {
+            var number = PgNumeric.Parse(input);
+            Assert.IsType<PgNumeric>(number);
+            Assert.NotNull(number.ToString());
```
amanda-tarafa, line 242: We don't need asserts here, Parse will throw if it cannot parse the string. In particular:

- If the first Assert executes it will always pass as PgNumeric is a struct.
- If the second Assert executes it will always pass as we never return null in ToString.

That is, you can remove the asserts.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,514 @@
...
+        }
+
+        [Theory]
+        [MemberData(nameof(GetStringInputs))]
+        public void ParseTest(string input, bool expectedParsingResult)
```
amanda-tarafa, line 247: Remove this one, you now have tests for parsing success and tests for parsing failures.
Remove the data member as well, if you want to keep some cases from the data member you can add them to the specific tests, maybe creating a data member for each instead of using inline data.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,514 @@
...
+            bool success = PgNumeric.TryParse(input, out PgNumeric number);
+            Assert.Equal(PgNumeric.Zero, number);
+            Assert.False(success);
+        }
+
+        [Theory]
+        [InlineData("1234,567", "es-ES")]
+        [InlineData("1.234,567", "es-ES")]
+        [InlineData("1234,567", "fr-FR")]
+        public void TryParse_UnsupportedCultures(string text, string culture)
```
amanda-tarafa, line 376: Hmmm, I don't like this tests, as they may run in parallel and this will affect the culture of all tests.
I think you can add these cases to the invalid parsing test without specifying the culture?

amanda-tarafa (changes requested):
I only reviewed PgNumeric and its tests. There are many other things on this PR now in commits that have no good explanatory names.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerDbTypeTests.ValueConversions.cs
```diff
@@ -255,6 +270,25 @@ public static IEnumerable<object[]> GetValidValueConversions()
             yield return new object[] {"1", SpannerDbType.Numeric, "\"1\""};
             yield return new object[] {"1.5", SpannerDbType.Numeric, "\"1.5\""};
             yield return new object[] {DBNull.Value, SpannerDbType.Numeric, "null"};
+            // PgNumeric tests.
+            yield return new object[] { PgNumeric.NaN, SpannerDbType.PgNumeric, "\"NaN\"" };
```
amanda-tarafa, line 274: There;s a Quote method that I think will escape the quotations, see a few lines up.

amanda-tarafa (changes requested):
PGNumeric is just a few nits and a tiny case that we missed, I've left a comment on the test.

A few comments on the support commits but looking good.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,456 @@
...
+        [InlineData("100", "100")]
+        [InlineData("1.33333333333333333", "1.33333333333333333")]
+        public void ToDecimal(string input, string expectedValue)
+        {
+            var numeric = PgNumeric.Parse(input);
+            var decimalValue = numeric.ToDecimal(LossOfPrecisionHandling.Truncate);
+            Assert.Equal(decimal.Parse(expectedValue), decimalValue);
+        }
+
+        [Theory]
```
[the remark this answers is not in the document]
Rishabh-V: Please correct me but method will overflow only for specified set of values which are greater than a threshold, so it needs parameterized test cases and hence a theory. Am I missing something?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
```diff
@@ -34,6 +34,7 @@ public sealed class SpannerParameter : DbParameter, ICloneable
         private SpannerDbType _spannerDbType;
 
         /// <inheritdoc />
+        // Using DbType of VarNumeric will always default to SpannerDbType of SpannerNumeric and never to PgNumeric.
```
amanda-tarafa, line 37: So that this makes it into public docs, I think you have to stop inheriting the docs, rewrite them here, and add this on the Remarks section.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -80,26 +80,36 @@ public sealed partial class SpannerDbType
...
 
-        internal static SpannerDbType FromTypeCode(TypeCode code) 
-            => s_simpleTypes.TryGetValue(code, out SpannerDbType type) ? type : null;
+        internal static SpannerDbType FromType(V1.Type type) =>
+            s_simpleTypes.TryGetValue(type, out SpannerDbType dbType) ?
+            dbType : null;
 
         internal TypeCode TypeCode { get; }
 
+        internal TypeAnnotationCode TypeAnnotationCode { get; }
```
amanda-tarafa, line 111: At this point, it might even make sense to keep Type here. It's fine as it is though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -151,6 +159,7 @@ public DbType DbType
                     case TypeCode.Float64:
                         return DbType.Double;
                     case TypeCode.Numeric:
+                        // This handles PG numeric as well.
```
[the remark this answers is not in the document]
amanda-tarafa: Unresolving just for the sake of keeping this top of mind while reviewing.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.StringParsing.cs
```diff
@@ -33,7 +33,7 @@ public sealed partial class SpannerDbType
         public static bool TryParse(string spannerType, out SpannerDbType spannerDbType)
         {
             spannerDbType = null;
-            if (!TryParsePartial(spannerType, out TypeCode code, out int? size, out string remainder))
+            if (!TryParsePartial(spannerType, out TypeCode code, out int? size, out string remainder, out TypeAnnotationCode typeAnnotationCode))
```
amanda-tarafa, line 36: Here maybe also change the pair TypeCode, TypeAnnotation by just Type

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.StringParsing.cs
```diff
@@ -199,6 +208,11 @@ public override string ToString()
                 s.Append(">");
                 return s.ToString();
             }
+            if (TypeAnnotationCode == TypeAnnotationCode.PgNumeric)
+            {
+                return "PGNUMERIC";
```
amanda-tarafa, line 213: Given that there's no type named PGNumeric, I wonder if it'd make more sense to do "NUMERIC{PG}", but that's just a very tiny nit.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerDbTypeTests.ValueConversions.cs
```diff
@@ -147,130 +162,149 @@ public static IEnumerable<object[]> GetValidValueConversions()
...
-            yield return new object[] {1.5D, SpannerDbType.String, Quote("1.5")};
-            yield return new object[] {1.5F, SpannerDbType.String, Quote("1.5")};
-            yield return new object[] {1, SpannerDbType.String, Quote("1")};
-            yield return new object[] {1U, SpannerDbType.String, Quote("1")};
-            yield return new object[] {1L, SpannerDbType.String, Quote("1")};
-            yield return new object[] {(ulong) 1, SpannerDbType.String, Quote("1")};
-            yield return new object[] {(short) 1, SpannerDbType.String, Quote("1")};
-            yield return new object[] {(ushort) 1, SpannerDbType.String, Quote("1")};
-            yield return new object[] {s_testDate, SpannerDbType.String, Quote("2017-01-31T03:15:30.5Z")};
+            yield return new object[] { true, SpannerDbType.String, Quote("True") };
```
amanda-tarafa, line 242: Most of these changes are white space only. We usually, specially if it is this much, don't include them on the same commit as actual changes, because it makes the actual changes harder to find a review.

amanda-tarafa (changes requested):
A few nits and minor changes but I think this looks good.

@jskeet I think you can take a look now to catch anything I might have missed.

@Rishabh-V maybe wait till Jon has reviewed so you can push all changes together.

jskeet:
Note: you'll want to make this not-a-draft PR.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,481 @@
...
+using System.Linq;
+using System.Numerics;
+using System.Text.RegularExpressions;
+
+namespace Google.Cloud.Spanner.V1
+{
+    /// <summary>
+    /// Representation of PostgreSQL numeric type which has max precision of 147455 and a max scale of 16383. 
+    /// </summary>
+    public struct PgNumeric : IEquatable<PgNumeric>, IComparable<PgNumeric>, IComparable
```
jskeet, line 28: I believe we can use `public readonly struct` which can help to optimize some cases.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,481 @@
...
+            if (string.Equals(text, NotANumber, StringComparison.InvariantCultureIgnoreCase))
+            {
+                value = new PgNumeric(default, 0, 0, true);
+                return null;
+            }
+
+            if (!s_validation.IsMatch(text))
+            {
+                value = default;
+                return "Text representation must be digits, containing an optional decimal point, and an optional leading '-' sign.";
```
jskeet, line 237: Or NaN

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,481 @@
...
+            // The raw value which will end up in the result, on success.
+            BigInteger rawValue;
+            int scale = 0;
+            int trailingZeros = 0;
+            // If we have '.', then we need to extract the value and scale.
+            if (pointIndex != -1)
+            {
+                int originalLength = text.Length;
+                // Remove trailing Zeros after decimal, so that correct scale is calculated.
+                // So, 1.000 will become 1, 1.001000 will become 1.001 and so on, while 100 will remain 100.
```
jskeet, line 251: Not quite - "1.000" will become "1."

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -0,0 +1,481 @@
...
+            {
+                value = default;
+                return "Text representation must be digits, containing an optional decimal point, and an optional leading '-' sign.";
+            }
+
+            int pointIndex = text.IndexOf('.');
+
+            // The raw value which will end up in the result, on success.
+            BigInteger rawValue;
+            int scale = 0;
```
jskeet, line 244: I'd declare this without a value, and assign a value of 0 in the `else` block, where you assign `rawValue`. That way it's clearer that the scale is zero when there's no decimal point.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,453 @@
...
+            yield return new object[] { "1234,567" };
+            yield return new object[] { "1.234,567" };
+            yield return new object[] { (BigInteger.Pow(10, 147455)).ToString() };
+            yield return new object[] { new string('9', 147456) };
+            yield return new object[] { "-" + new string('9', 147456) };
+        }
+
+        public static IEnumerable<object[]> NaNComparisionData()
+        {
+            // The values are: string input1, string input2, and expected sign of comparision result.
```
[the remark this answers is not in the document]
Rishabh-V: NaN is greater than all the other values. Added `y.CompareTo(x)` in addition to `x.CompareTo(y)`. We have not implemented <> operators in PgNumeric. Please advise if further changes are needed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,453 @@
...
+        [InlineData("1.00000000", "0")]
+        [InlineData("1.01", "1")]
+        [InlineData("0.0000008", "0.00000009")]
+        [InlineData("2", "1.999999999999999999999999")]
+        [InlineData("99.99", "99.90")]
+        [InlineData("1.0", "-1.000")]
+        [InlineData("-1", "-123.0")]
+        [InlineData("0", "-0.0009")]
+        [InlineData("0.00009", "0.00000009")]
+        public void CompareTo_GreaterThan(string input1, string input2)
```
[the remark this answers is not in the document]
Rishabh-V: All the scenarios mentioned above are covered in existing test cases except for explicit `<` , `>` tests as PgNumeric doesn't implement `<`  and `>` operators but they are covered via `CompareTo`. Please advise if individual tests aren't fine. I believe it would be easier to test and isolate issues if we have specific tests for equality, less than and greater than comparison. Please let me know you thoughts and I will do as you suggest.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,453 @@
...
+
+        [Theory]
+        [InlineData("NaN", "NaN")]
+        [InlineData("1.00", "1.00")]
+        [InlineData("1.000000", "1.000000")]
+        [InlineData("99999.999999000", "99999.999999000")]
+        [InlineData("-0.000001", "-0.000001")]
+        [InlineData("100", "100")]
+        [InlineData("1.33333333333333333", "1.33333333333333333")]
+        public void To_String(string input, string expected)
```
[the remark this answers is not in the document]
Rishabh-V: I thought so as well. But xUnit doesn't let us do that. Please see [xUnit1024](https://xunit.net/xunit.analyzers/rules/xUnit1024.html)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,453 @@
...
+        [InlineData("-12345678912.12345678912", "-12345678912.12345678912")]
+        [InlineData("123456789123456789.12345678912", "123456789123456789.12345678912")]
+        [InlineData("12345678912345678912345678.1234567891234567891234", "12345678912345678912345678.1234567891234567891234")]
+        [InlineData("9999999999999999999999999999.999999999999999999999999", "9999999999999999999999999999.999999999999999999999999")]
+        [InlineData("100", "100")]
+        [InlineData("1.33333333333333333", "1.33333333333333333")]
+        public void ToDecimal(string input, string expectedValue)
+        {
+            var numeric = PgNumeric.Parse(input);
+            var decimalValue = numeric.ToDecimal(LossOfPrecisionHandling.Truncate);
```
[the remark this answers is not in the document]
Rishabh-V: Yes. Now they do.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,453 @@
...
+            yield return new object[] { "1e6" };
+            yield return new object[] { "1e-6" };
+            yield return new object[] { "1234,567" };
+            yield return new object[] { "1.234,567" };
+            yield return new object[] { (BigInteger.Pow(10, 147455)).ToString() };
+            yield return new object[] { new string('9', 147456) };
+            yield return new object[] { "-" + new string('9', 147456) };
+        }
+
+        public static IEnumerable<object[]> NaNComparisionData()
```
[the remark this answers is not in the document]
Rishabh-V: Right Jon. We have MemberData because the data cannot be specified inline.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerDbTypeTests.ValueConversions.cs
Rishabh-V: Sure Amanda, will keep this in mind going forward.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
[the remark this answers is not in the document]
jskeet: It doesn't need to be done before we merge, but I think it's simpler to have *either* a single test that covers all non-equal comparisons, *or* multiple test methods but with a single set of test data.

Currently if we found some corner case, we'd need to add that test data in multiple places, and that's not great.

While I'm aware that some testing philosophies suggest a single assertion per test, my experience is that it's far more pragmatic to test multiple very-closely-related operations (such as "all comparisons") in a single test. We're unlikely to have a problem in greater-than but not in less-than, for example. (I can see inequality being potentially separated out.)

jskeet (commented):
These are mostly just changes to the layout of test data, but I think the truncation one needs further discussion.

(I'd also suggest that it's better to let the reviewer who started a comment thread do the resolving - otherwise they have to go through every resolved comment, expand it, then work out whether it's actually resolved.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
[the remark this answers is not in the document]
Rishabh-V: Hi Jon, Please see if the updated test makes better sense of if any further change is desired.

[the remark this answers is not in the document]
jskeet: I'm still not seeing anything that tests actual truncation in an obvious way - that 9999999999999999999999999999.999999999999999999999999 ends up with truncated digits.

I can sort of see how the test is doing it, due to using decimal.Parse, but I think it would be clearer if the test case included the expected result of `decimal.ToString` - so we could really see what the result is.

I'm also confused as to how the test worked before, in the outdated version. That's what really worried me. Unfortunately I can't see the old version now, so I can't see *exactly* what was going on. Basically, I don't know whether a production code change is required or not.

(I'd also prefer the method names to give more information about what's going on - for example, "ToDecimal_LossOfPrecision" doesn't tell us what value of LossOfPrecision is being supplied, rather than just that it *is* being supplied.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -0,0 +1,445 @@
...
+            decimal expected = decimal.Parse(input, CultureInfo.InvariantCulture);
+            decimal actual = numeric.ToDecimal(LossOfPrecisionHandling.Truncate);
+            Assert.Equal(expected, actual);
+            // Conversion via the explicit conversion should do the same thing.
+            decimal actual2 = (decimal)numeric;
+            Assert.Equal(expected, actual2);
+        }
+
+        [Theory]
+        [InlineData("123456789012345678900.123456789", "123456789012345678900.12345679")]
```
jskeet, line 358: Hmm... this example still doesn't look like truncation. It should be converting to "123456789012345678900.12345678". Let me know if you'd like me to pull the code and debug this.

jskeet (commented):
With the new test, it looks likely that there's an issue in the production code here... although IIRC, decimal has "28 or 29" significant digits, so it may be that ambiguity that's causing issues.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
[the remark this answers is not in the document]
Rishabh-V: Please see if these new test cases and test data makes sense. 
My understanding about tests working earlier is that when we parse 999[...].999[...] as decimal, since it is more than 29 digits, it rounds up to 10[...] and that is what was put as expected value as well. 
I have also changed the name of methods by appending _Truncate and _Throw to the test methods. Thanks

PR #8278, opened 2022-03-25
First pass at mix-in libraries for GAX v4

jskeet:
This provides help with bootstrapping: the generator depends on the
mix-ins, which need to be generated...


amanda-tarafa (commented):
Looks good as well

PR #8287, opened 2022-03-30
fix(PubSub): Invalid Arguments error should not be thrown to the client

Rishabh-V:
This fixes the error in Subscriber client caused by toggling Exactly Once delivery on the subscription.  
Status(StatusCode="InvalidArgument", Detail="Some acknowledgement ids in the request were invalid. This could be because the acknowledgement ids have expired or the acknowledgement ids were malformed.")

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
@@ -17,6 +17,7 @@ using Google.Api.Gax.Grpc;
 using Google.Cloud.PubSub.V1.Tasks;
 using Google.Cloud.PubSub.V1.Tests.Tasks;
 using Google.Protobuf;
+using Google.Protobuf.WellKnownTypes;
 using Grpc.Core;
 using System;
 using System.Collections.Generic;
@@ -64,6 +65,54 @@ namespace Google.Cloud.PubSub.V1.Tests
             public RpcException MoveNextEx { get; }
             public RpcException CurrentEx { get; }
             public int? DeliveryAttempt { get; }
+        }
+
+        // Class to simulate exceptions in Acknowledgement and/or ModifyAcknowledgementDeadline calls.
+        private class AckModifyAckDeadlineAction
+        {
+            public static AckModifyAckDeadlineAction Data(Exception ex, int numberOfFailures, bool onAck, bool onModifyAckDeadline) => new AckModifyAckDeadlineAction(ex, numberOfFailures, onAck, onModifyAckDeadline);
+            public static AckModifyAckDeadlineAction BadAck(Exception ex, int numberOfFailures) => new AckModifyAckDeadlineAction(ex, numberOfFailures, true, false);
+            public static AckModifyAckDeadlineAction BadModifyAckDeadline(Exception ex, int numberOfFailures) => new AckModifyAckDeadlineAction(ex, numberOfFailures, false, true);
+
+            /// <summary>
+            /// The exception to be thrown in Ack/ModifyAckDeadline call.
+            /// </summary>
+            /// <value>
+            /// The exception.
+            /// </value>
+            public Exception Exception { get; }
+
+            /// <summary>
+            /// The number of failures to be simulated in Ack/ModifyAckDeadline call.
+            /// </summary>
+            /// <value>
+            /// The number of failures.
+            /// </value>
+            public int NumberOfFailures { get; }
+
+            /// <summary>
+            /// Gets a value indicating whether the exception should be thrown in Acknowledge call.
+            /// </summary>
+            /// <value>
+            ///   <c>true</c> if [on ack]; otherwise, <c>false</c>.
+            /// </value>
+            public bool OnAck { get; }
+
+            /// <summary>
+            /// Gets a value indicating whether the exception should be thrown in ModifyAcknowledgementDeadline call.
+            /// </summary>
+            /// <value>
+            ///   <c>true</c> if [on modify ack deadline]; otherwise, <c>false</c>.
+            /// </value>
+            public bool OnModifyAckDeadline { get; }
+
+            private AckModifyAckDeadlineAction(Exception exception, int numberOfFailures, bool onAck, bool onModifyAckDeadline)
+            {
+                Exception = exception;
+                NumberOfFailures = numberOfFailures;
+                OnAck = onAck;
+                OnModifyAckDeadline = onModifyAckDeadline;
+            }
         }
 
         private class FakeSubscriberServiceApiClient : SubscriberServiceApiClient
@@ -206,7 +255,7 @@ namespace Google.Cloud.PubSub.V1.Tests
 
             public FakeSubscriberServiceApiClient(
                 IEnumerator<IEnumerable<ServerAction>> msgsEn, IScheduler scheduler, IClock clock,
-                TaskHelper taskHelper, TimeSpan writeAsyncPreDelay, bool useMsgAsId)
+                TaskHelper taskHelper, TimeSpan writeAsyncPreDelay, bool useMsgAsId, AckModifyAckDeadlineAction ackModifyAckDeadlineAction)
             {
                 _msgsEn = msgsEn;
                 _scheduler = scheduler;
@@ -214,6 +263,8 @@ namespace Google.Cloud.PubSub.V1.Tests
                 _taskHelper = taskHelper;
                 _writeAsyncPreDelay = writeAsyncPreDelay;
                 _useMsgAsId = useMsgAsId;
+                _ackModifyAckDeadlineAction = ackModifyAckDeadlineAction;
+                _numberOfAckModifyAckDeadlineFailures = 0;
             }
 
             private readonly object _lock = new object();
@@ -223,6 +274,8 @@ namespace Google.Cloud.PubSub.V1.Tests
             private readonly TaskHelper _taskHelper;
             private readonly TimeSpan _writeAsyncPreDelay; // Simulates slow network or server
             private readonly bool _useMsgAsId;
+            private readonly AckModifyAckDeadlineAction _ackModifyAckDeadlineAction; // Simulates Ack ModifyAckDeadline errors.
+            private int _numberOfAckModifyAckDeadlineFailures;
 
             private readonly List<TimedId> _extends = new List<TimedId>();
             private readonly List<TimedId> _acks = new List<TimedId>();
@@ -257,6 +310,19 @@ namespace Google.Cloud.PubSub.V1.Tests
                 {
                     await _taskHelper.ConfigureAwait(_scheduler.Delay(_writeAsyncPreDelay, CancellationToken.None));
                 }
+
+                // Simulate Ack failure if _ackModifyAckDeadlineAction is specified.
+                // All gRPC exceptions from this method are fire and forget, so no exception should propagate to the caller.
+                // Non-gRPC exceptions will propagate to the caller.
+                if (_ackModifyAckDeadlineAction?.Exception != null && _ackModifyAckDeadlineAction.OnAck)
+                {
+                    if (Interlocked.CompareExchange(ref _numberOfAckModifyAckDeadlineFailures, _ackModifyAckDeadlineAction.NumberOfFailures, _ackModifyAckDeadlineAction.NumberOfFailures) < _ackModifyAckDeadlineAction.NumberOfFailures)
+                    {
+                        Interlocked.Increment(ref _numberOfAckModifyAckDeadlineFailures);
+                        throw _ackModifyAckDeadlineAction.Exception;
+                    }
+                }
+
                 lock (_lock)
                 {
                     _acks.AddRange(ackIds.Select(id => new TimedId(_clock.GetCurrentDateTimeUtc(), id)));
@@ -269,6 +335,19 @@ namespace Google.Cloud.PubSub.V1.Tests
                 {
                     await _taskHelper.ConfigureAwait(_scheduler.Delay(_writeAsyncPreDelay, CancellationToken.None));
                 }
+
+                // Simulate ModifyAckDeadline failure if _ackModifyAckDeadlineAction is specified.
+                // All gRPC exceptions from this method are fire and forget, so no exceptions should propagate to the caller.
+                // Non-gRPC exceptions will propagate to the caller.
+                if (_ackModifyAckDeadlineAction?.Exception != null && _ackModifyAckDeadlineAction.OnModifyAckDeadline)
+                {
+                    if (Interlocked.CompareExchange(ref _numberOfAckModifyAckDeadlineFailures, _ackModifyAckDeadlineAction.NumberOfFailures, _ackModifyAckDeadlineAction.NumberOfFailures) < _ackModifyAckDeadlineAction.NumberOfFailures)
+                    {
+                        Interlocked.Increment(ref _numberOfAckModifyAckDeadlineFailures);
+                        throw _ackModifyAckDeadlineAction.Exception;
+                    }
+                }
+
                 lock (_lock)
                 {
                     var ids = ackDeadlineSeconds == 0 ? _nacks : _extends;
@@ -290,14 +369,14 @@ namespace Google.Cloud.PubSub.V1.Tests
                 TimeSpan? ackDeadline = null, TimeSpan? ackExtendWindow = null,
                 int? flowMaxElements = null, int? flowMaxBytes = null,
                 int clientCount = 1, int threadCount = 1, TimeSpan? writeAsyncPreDelay = null,
-                bool useMsgAsId = false)
+                bool useMsgAsId = false, AckModifyAckDeadlineAction ackModifyAckDeadlineAction = null)
             {
                 var scheduler = new TestScheduler(threadCount: threadCount);
                 TaskHelper taskHelper = scheduler.TaskHelper;
                 List<DateTime> clientShutdowns = new List<DateTime>();
                 var msgEn = msgs.GetEnumerator();
                 var clients = Enumerable.Range(0, clientCount)
-                    .Select(_ => new FakeSubscriberServiceApiClient(msgEn, scheduler, scheduler.Clock, taskHelper, writeAsyncPreDelay ?? TimeSpan.Zero, useMsgAsId))
+                    .Select(_ => new FakeSubscriberServiceApiClient(msgEn, scheduler, scheduler.Clock, taskHelper, writeAsyncPreDelay ?? TimeSpan.Zero, useMsgAsId, ackModifyAckDeadlineAction))
                     .ToList();
                 var settings = new SubscriberClient.Settings
                 {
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1402,8 +1402,17 @@ private void HandleAckResponse(Task writeTask, List<string> ackIds, List<string>
                     }
                     else
                     {
-                        // Unrecoverable error; throw exception.
-                        throw writeTask.Exception.FlattenIfPossible();
+                        // Check if its RpcException with StatusCode of InvalidArguments and Reason as EXACTLY_ONCE_ACKID_FAILURE.
```
jskeet, line 1405:
```suggestion
                        // Check if it's an RpcException with StatusCode of InvalidArgument and Reason of EXACTLY_ONCE_ACKID_FAILURE.
```

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1402,8 +1402,17 @@ private void HandleAckResponse(Task writeTask, List<string> ackIds, List<string>
...
+                        // If it is, then ignore it and continue as they are fire and forget. We may want to log them later. 
+                        // Other non-recoverable errors will continue to be thrown.
+                        var rpcException = writeTask.Exception.As<RpcException>();
+                        bool ignoreError = rpcException?.StatusCode == StatusCode.InvalidArgument &&
+                            string.Equals(rpcException?.GetErrorInfo()?.Reason, "EXACTLY_ONCE_ACKID_FAILURE", StringComparison.InvariantCulture);
```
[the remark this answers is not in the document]
jskeet: Also if we've got here, then `rpcException` isn't null, so the first `?` in `rpcException?.GetErrorInfo()?.Reason` can be removed. (The second one is still needed.)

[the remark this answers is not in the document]
jskeet: We should also check the domain, really. That's not *vital* as it's very unlikely that anything else would use a reason of EXACTLY_ONCE_ACKID_FAILURE, but it's really the domain+reason that's expected to be unique.

jskeet (commented):
Nit on the commit comment - I believe it needs to be "fix" rather than "Fix"; I think conventional commits are case-sensitive. (Certainly we normally use lower-case categories.)

The fix itself looks fine other than the comments I've left, but it would be good to see four tests:

- A non-RpcException (still thrown)
- An RpcException with a different status code (still thrown)
- An RpcException for InvalidArgument with an ErrorInfo specifying a different reason (still thrown)
- An RpcException for InvalidArgument with the right reason/domain (not thrown)

We could add even more tests than that, but let's be pragmatic. However, I do think we should have at least one rethrown and one not-thrown case.

See https://github.com/googleapis/gax-dotnet/blob/main/Google.Api.Gax.Grpc.Tests/RpcExceptionExtensionsTest.cs for sample code that builds an RpcException with an error info.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1402,8 +1402,17 @@ private void HandleAckResponse(Task writeTask, List<string> ackIds, List<string>
                     }
                     else
                     {
-                        // Unrecoverable error; throw exception.
-                        throw writeTask.Exception.FlattenIfPossible();
+                        // Check if its RpcException with StatusCode of InvalidArguments and Reason as EXACTLY_ONCE_ACKID_FAILURE.
+                        // If it is, then ignore it and continue as they are fire and forget. We may want to log them later. 
+                        // Other non-recoverable errors will continue to be thrown.
+                        var rpcException = writeTask.Exception.As<RpcException>();
+                        bool ignoreError = rpcException?.StatusCode == StatusCode.InvalidArgument &&
```
pradn, line 1409: So the ack/modack behavior should be:
1) Retry Unavailable errors for ack/modack (this already happens at the GAPIC/GRPC level, see [here](https://github.com/googleapis/googleapis/blob/480e3fbafbe4eb61c2118e3cb8d9e0d8eb9a9e36/google/pubsub/v1/pubsub_grpc_service_config.json#L221))
2) Ignore all GRPC errors for ack/modack

Please change the code to ignore all GRPC errors for ack/modack in this PR.

The reason we ignore GRPC errors for ack/modack is because they are best-effort RPCs. The user shouldn't see an exception when these happen. There's a subtlety - what if the subscription is not accessible (PermissionDenied) or doesn't exist (NotFound)? Our thinking around this is that it's not that important that the user sees these errors with ack/modack because the StreamingPull stream will fail if these errors occur and then the whole subscriber object will fail. So no need to do extra work here. We may revisit this choice for these two special error codes though. For now, we'll stick to this in the name of uniformity (the Java client has the behavior described here.)

Once we do the full exactly-once delivery work, the whole behavior will be:
1) Retry Unavailable errors for ack/modack at the GAPIC/GRPC level (no change to this behavior)
2) If exactly-once is disabled, ignore all GRPC errors for ack/modack
3) If exactly-once is enabled, bubble up extra info via futures.

pradn (changes requested):

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
```diff
@@ -40,7 +40,9 @@ public abstract class PublisherClient
     {
         private static readonly GrpcChannelOptions s_unlimitedSendReceiveChannelOptions = GrpcChannelOptions.Empty
             .WithMaxReceiveMessageSize(-1)
-            .WithMaxSendMessageSize(-1);
+            .WithMaxSendMessageSize(-1)
+            // Set max metadata size to 4 MB i.e., 4194304 bytes.
+            .WithCustomOption("grpc.max_metadata_size", 4194304);
```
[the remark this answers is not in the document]
pradn: You've already fixed this here: https://github.com/googleapis/google-cloud-dotnet/commit/e78afdf41474e38a266a46cb6ae0489ec56a5266

I think you need to rebase the branch on head.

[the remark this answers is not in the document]
jskeet: No, that's for SubscriberClient. This is PublisherClient.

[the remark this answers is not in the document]
Rishabh-V: Thanks Jon. I have excluded PublisherClient from this PR and will raise a new PR for it.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Swallowed all the gRPC errors in Ack/ModAck call. Have tested thoroughly with a client app for an hour. Created new issue #8291 for writing unit tests as it may require substantial time.

jskeet (approved):
This looks fine to me. @pradn are you happy with non-gRPC exceptions still being propagated?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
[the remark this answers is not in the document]
pradn: Ah, gotcha! Thanks both!

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
pradn: Thank you Rishabh, especially for the thorough testing. Please write the unit tests in this PR itself. We usually prefer it that way unless there's a strong reason otherwise. Yes, this issue is high priority, but we can wait for unit tests.

pradn:
I think its fine for non-gRPC exceptions to be propagated, for now at least. We need a wider discussion around the exactly error reporting behavior for acks/modacks. I'll bring it up at the CPS client libs meeting.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -66,6 +67,32 @@ private ServerAction(TimeSpan preInterval, IEnumerable<string> msgs, RpcExceptio
...
+            
+            private AckModifyAckDeadlineAction(Exception exception, int numberOfFailures, bool onAck, bool onModifyAckDeadline)
+            {
+                Exception = exception;
+                NumberOfFailures = numberOfFailures;
+                OnAck = onAck;
+                OnModifyAckDeadline = onModifyAckDeadline;
+            }
+
+            public Exception Exception { get; }
```
jskeet, line 87: Nit: please move the properties to above the constructor. I'd also remove the blank lines between the properties, and between the three factory methods. (But keep a line between the three factory methods and the properties.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -66,6 +67,32 @@ private ServerAction(TimeSpan preInterval, IEnumerable<string> msgs, RpcExceptio
...
+
+            public int NumberOfFailures { get; }
+
+            public bool OnAck { get; }
```
jskeet, line 91: I don't understand what this or `OnModifyAckDeadline` actually mean. I'd suggest either renaming, or adding an XML summary comment. (If you add a comment, then ignore the bit above about removing blank lines.)

jskeet: Reading a bit more, I see what it's doing, but I suspect there may be a cleaner design. We can revisit once this is in though.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -257,6 +288,19 @@ public override async Task AcknowledgeAsync(SubscriptionName subscription, IEnum
                 {
                     await _taskHelper.ConfigureAwait(_scheduler.Delay(_writeAsyncPreDelay, CancellationToken.None));
                 }
+
+                // Simulate Ack failure if _ackModifyAckDeadlineAction is specified.
+                // All gRPC exceptions from this method are fire and forget, so no exception should propogate to the caller.
```
jskeet, line 293: Nit: propogate => propagate

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -257,6 +288,19 @@ public override async Task AcknowledgeAsync(SubscriptionName subscription, IEnum
...
+                // Non-gRPC exceptions will propogate to the caller.
+                if (_ackModifyAckDeadlineAction?.Exception != null && _ackModifyAckDeadlineAction.OnAck)
+                {
+                    if (Interlocked.CompareExchange(ref _numberOfAckModifyAckDeadlineFailures, _ackModifyAckDeadlineAction.NumberOfFailures, _ackModifyAckDeadlineAction.NumberOfFailures) < _ackModifyAckDeadlineAction.NumberOfFailures)
+                    {
+                        Interlocked.Increment(ref _numberOfAckModifyAckDeadlineFailures);
+                        await Task.FromException(_ackModifyAckDeadlineAction.Exception);
```
jskeet, line 300: Why are you awaiting something you know will throw, rather than just using `throw _ackModifyAckDeadlineAction.Exception`?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -953,5 +1010,77 @@ public void DeliveryAttempt()
                 });
             }
         }
+
+        // Acknowledge / ModifyAcknowledgeDeadline calls may throw RpcException. RpcExceptions should not be thrown to the client.
+        [Theory, CombinatorialData]
+        public void AckModifyAckDeadlineFault_NotThrown(
+            [CombinatorialValues(true, false, null)] bool? ackOrModifyAck)
```
jskeet, line 1017: Unwrap onto the declaration line

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -953,5 +1010,77 @@ public void DeliveryAttempt()
...
+                ServerAction.Data(TimeSpan.Zero, new[] { "4" }),
+                ServerAction.Inf()
+            };
+
+            // Any RpcException thrown from Acknowledge and/ or ModifyAcknowledgementDeadline should not be thrown to the client.
+            var rpcException = new RpcException(new Status(StatusCode.InvalidArgument, ""), "");
+            // If ackOrModifyAck is true, then Acknowledge (Ack) will throw the supplied RpcException 2 times.
+            // If ackOrModifyAck is false, then ModifyAcknowledgementDeadline (ModifyAck) will throw the supplied RpcException 2 times.
+            // If ackOrModifyAck is null, then both Ack and ModifyAck will throw the supplied RpcException, total 2 times.
+            var ackModifyAckDeadlineAction = ackOrModifyAck == null ? AckModifyAckDeadlineAction.Data(rpcException, 2, true, true) : ackOrModifyAck.Value ? AckModifyAckDeadlineAction.BadAck(rpcException, 2) : AckModifyAckDeadlineAction.BadModifyAckDeadline(rpcException, 2);
```
jskeet, line 1032:
```suggestion
            var ackModifyAckDeadlineAction =
                ackOrModifyAck == null ? AckModifyAckDeadlineAction.Data(rpcException, 2, true, true)
                : ackOrModifyAck.Value ? AckModifyAckDeadlineAction.BadAck(rpcException, 2)
                : AckModifyAckDeadlineAction.BadModifyAckDeadline(rpcException, 2);
```

(That's the way we normally format this sort of multiple-conditional operator.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -953,5 +1010,77 @@ public void DeliveryAttempt()
...
+                ServerAction.Data(TimeSpan.Zero, new[] { "3" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "4" }),
+                ServerAction.Inf()
+            };
+
+            // Any RpcException thrown from Acknowledge and/ or ModifyAcknowledgementDeadline should not be thrown to the client.
+            var rpcException = new RpcException(new Status(StatusCode.InvalidArgument, ""), "");
+            // If ackOrModifyAck is true, then Acknowledge (Ack) will throw the supplied RpcException 2 times.
+            // If ackOrModifyAck is false, then ModifyAcknowledgementDeadline (ModifyAck) will throw the supplied RpcException 2 times.
+            // If ackOrModifyAck is null, then both Ack and ModifyAck will throw the supplied RpcException, total 2 times.
```
jskeet, line 1031: Please reorder these comments to match the order in the code

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -953,5 +1010,77 @@ public void DeliveryAttempt()
                 });
             }
         }
+
+        // Acknowledge / ModifyAcknowledgeDeadline calls may throw RpcException. RpcExceptions should not be thrown to the client.
+        [Theory, CombinatorialData]
+        public void AckModifyAckDeadlineFault_NotThrown(
+            [CombinatorialValues(true, false, null)] bool? ackOrModifyAck)
+        {
+            var msgs = new[] {
```
jskeet, line 1019: Nit: I'd typically put the `{` on a line on its own here.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -953,5 +1010,77 @@ public void DeliveryAttempt()
...
+            // If ackOrModifyAck is true, then Acknowledge (Ack) will throw the supplied RpcException 2 times.
+            // If ackOrModifyAck is false, then ModifyAcknowledgementDeadline (ModifyAck) will throw the supplied RpcException 2 times.
+            // If ackOrModifyAck is null, then both Ack and ModifyAck will throw the supplied RpcException, total 2 times.
+            var ackModifyAckDeadlineAction = ackOrModifyAck == null ? AckModifyAckDeadlineAction.Data(rpcException, 2, true, true) : ackOrModifyAck.Value ? AckModifyAckDeadlineAction.BadAck(rpcException, 2) : AckModifyAckDeadlineAction.BadModifyAckDeadline(rpcException, 2);
+
+            using var fake = Fake.Create(new[] { msgs }, ackDeadline: TimeSpan.FromSeconds(30), ackExtendWindow: TimeSpan.FromSeconds(10), ackModifyAckDeadlineAction: ackModifyAckDeadlineAction);
+            fake.Scheduler.Run(async () =>
+            {
+                var handledMsgs = new List<string>();
+                var doneTask = fake.Subscriber.StartAsync(async (msg, ct) =>
```
jskeet, line 1038: Is this code ever actually going to call ModifyAckDeadline?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -953,5 +1010,77 @@ public void DeliveryAttempt()
...
+                await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(100), CancellationToken.None));
+                await fake.TaskHelper.ConfigureAwait(fake.Subscriber.StopAsync(CancellationToken.None));
+                // Despite RpcException being thrown, all 4 messages should be handled.
+                Assert.Equal(new[] { "1", "2", "3", "4" }, handledMsgs);
+            });
+        }
+
+        // If Acknowledge / ModifyAcknowledgeDeadline calls throw exceptions other than RpcExceptions, they should be thrown to the client.
+        [Theory, CombinatorialData]
+        public void AckModifyAckDeadlineFault_Thrown(
```
jskeet, line 1053: Again, unwrap...

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
jskeet: (Here and elsewhere.)

jskeet (approved):
Mostly formatting nits. It feels a little odd to have the combinatorial tests for different RPCs behaving differently... that may just be due to the way the calls are made though.

I think it's okay to merge this after fixing the formatting, even though it feels odd.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
Rishabh-V: Done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
Rishabh-V: > I don't understand what this or `OnModifyAckDeadline` actually mean. I'd suggest either renaming, or adding an XML summary comment. (If you add a comment, then ignore the bit above about removing blank lines.)
Added the XML comments.

> Reading a bit more, I see what it's doing, but I suspect there may be a cleaner design. We can revisit once this is in though.
Sure Jon.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
Rishabh-V: Yes Jon. I confirmed again by debugging. We have explicitly added 70 second delay in handler and ackDeadline is set to lower value to simulate the call.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
Rishabh-V: I was somehow so keen to use `Task.FromException` that `throw` didn't come to my mind. I have now changed it to throw.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Unit Tests are now added. Please have a look.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
jskeet: Aha, right. It might be worth adding a comment to explain that.

jskeet (approved):
One comment suggest, but other than that, let's ship it.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
Rishabh-V: Added comment in both the tests.

pradn (approved):
I'm approving the behavior/implementation change. I'll leave it to Jon to review the tests, too.

PR #8292, opened 2022-03-31
fix(PubSub): Increased grpc.max_metadata_size value to 4 M.B.

Rishabh-V:
Contains changes in PublisherClient to increase grpc.max_metadata_size value to 4 M.B. SubscriberClient changes were done in  #8269  This change is done to maintain consistency with other language libraries.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
@@ -40,7 +40,9 @@ namespace Google.Cloud.PubSub.V1
     {
         private static readonly GrpcChannelOptions s_unlimitedSendReceiveChannelOptions = GrpcChannelOptions.Empty
             .WithMaxReceiveMessageSize(-1)
-            .WithMaxSendMessageSize(-1);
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
```diff
@@ -40,7 +40,9 @@ public abstract class PublisherClient
     {
         private static readonly GrpcChannelOptions s_unlimitedSendReceiveChannelOptions = GrpcChannelOptions.Empty
             .WithMaxReceiveMessageSize(-1)
-            .WithMaxSendMessageSize(-1);
+            .WithMaxSendMessageSize(-1)
+            // Set max metadata size to 4 MB i.e., 4194304 bytes.
+            .WithCustomOption("grpc.max_metadata_size", 4194304);
```
maheshgattani, line 45: This is only publisher client. We need this change for the subscriber client.

Rishabh-V: Changes for Subscriber client are already done. Please see #8269

maheshgattani (approved):

PR #8314, opened 2022-04-05
Regenerate APIs with updated IAM messages

jskeet:
The final commit here was manually separated from the IAM change.


amanda-tarafa (approved):
LGTM

PR #8327, opened 2022-04-07
Modify and release mixin libraries compatible with GAX v4.0.0-alpha01

jskeet:


jskeet:
Ah... the breaking change detector itself has failed to run.

Rather than fixing it now in a "coping with all kinds of targets" way, let's continue without it (given that we're expecting breaking changes anyway) and just merge after review.

amanda-tarafa (approved):
LGTM

Owlbot only runs on main, right?

jskeet:
> Owlbot only runs on main, right?

Yes, I believe so.

PR #8328, opened 2022-04-07
chore: Fix enum constants generator for Compute

jskeet:
This only tackles the one problem we've seen so far, but is easy to extend. (We don't need to try to anticipate everything.)

Once this is in, we can regenerate #8323 and it should build.


amanda-tarafa (approved):
LGTM

PR #8332, opened 2022-04-08
feat:SpannerDate struct

### LINKED ISSUE #5839: [Spanner] feat: add `SpannerDate` struct
The Spanner client library currently uses `datetime` as the default type for both `TIMESTAMP` and `DATE` columns. That produces some challenges:
1. Setting a `datetime` value on a `SpannerParameter` without explicitly also setting the `DbType` or `SpannerDbType`, will always create a `SpannerParameter` with type `TIMESTAMP`. This makes it difficult to make a good default mapping for generic frameworks that only set a value and not a specific type for a parameter (e.g. Entity Framework).
2. `datetime` values always contain a time component and a kind (local, UTC, unspecified). This indicates that the `datetime` represents a specific moment in time, while a `DATE` value in Cloud Spanner [explicitly does not do so](https://cloud.google.com/spanner/docs/data-types#date_type).

Having a `SpannerDate` struct similar to the `SpannerNumeric` struct would mitigate the above problems. We should consider adding that to the client library and use that as the default type for `DATE` columns.

Rishabh-V:
Prototype PR to discuss implementation of SpannerDate. Closes #5839 
This is very much a WIP and created in lieu of design document for better understanding and discussion.

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
@@ -1,0 +1,280 @@
+﻿// Copyright 2022 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using System;
+using System.Globalization;
+
+namespace Google.Cloud.Spanner.V1
+{
+    /// <summary>
+    /// The SpannerDate type represents a logical calendar date, independent of time zone.
+    /// Represents dates with values ranging from January 1, 0001 A.D through December 31, 9999 A.D. in the Gregorian calendar.
+    /// </summary>
+    public readonly struct SpannerDate : IEquatable<SpannerDate>, IComparable<SpannerDate>, IComparable
+    {
+        // TODO : Once upgraded to .NET 6, have an operator for converting DateOnly to SpannerDate and vice versa.
+
+        // The latest possible date is 9999/12/31 (DateTime.MaxValue) and earliest possible date is 0001/01/01 (DateTime.MinValue).
+        // The days between maximum and minimum date is MaxNumberOfDays.
+        private const int MaxNumberOfDays = 3652058;
+
+        /// <summary>
+        /// Gets the latest possible date that can be created.
+        /// </summary>
+        /// <value>
+        /// The maximum date as SpannerDate.
+        /// </value>
+        public static SpannerDate MaxDate => new SpannerDate(DateTime.MaxValue);
+
+        /// <summary>
+        /// Gets the earliest possible date that can be created.
+        /// </summary>
+        /// <value>
+        /// The minimum date as SpannerDate.
+        /// </value>
+        public static SpannerDate MinDate => new SpannerDate(DateTime.MinValue);
+
+        // The value of SpannerDate stored as the number of days since the DateTime.MinValue day i.e., 0001/01/01.
+        private readonly int _days;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SpannerDate"/> structure to the specified DateTime.
+        /// </summary>
+        /// <param name="dateTime">The dateTime as <see cref="DateTime"/>.</param>
+        public SpannerDate(DateTime dateTime) : this(dateTime.Subtract(DateTime.MinValue).Days)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SpannerDate"/> structure to the specified year, month, and day in the Gregorian calendar.
+        /// </summary>
+        /// <param name="year">The year (1 through the number of years in calendar).</param>
+        /// <param name="month">The month (1 through the number of months in calendar).</param>
+        /// <param name="day">The day (1 through the number of days in <paramref name="month"/>).</param>
+        public SpannerDate(int year, int month, int day) : this(new DateTime(year, month, day))
+        {
+        }
+
+        private SpannerDate(int days)
+        {
+            GaxPreconditions.CheckArgumentRange(days, nameof(days), 0, MaxNumberOfDays);
+            _days = days;
+        }
+
+        ///<inheritdoc/>
+        public int CompareTo(SpannerDate other) => _days.CompareTo(other._days);
+
+        /// <summary>
+        /// Implementation of <see cref="IComparable.CompareTo"/> to compare two <see cref="SpannerDate"/> values.
+        /// </summary>
+        /// <remarks>
+        /// This uses explicit interface implementation to avoid it being called accidentally. The generic implementation should usually be preferred.
+        /// </remarks>
+        /// <exception cref="ArgumentException"><paramref name="obj"/> is non-null but does not refer to an instance of <see cref="SpannerDate"/>.</exception>
+        /// <param name="obj">The object to compare this value with.</param>
+        /// <returns>The result of comparing this value with <paramref name="obj"/>. <paramref name="obj"/> is null, this method returns a value greater than 0.
+        /// </returns>
+        int IComparable.CompareTo(object obj)
+        {
+            if (obj == null)
+            {
+                return 1;
+            }
+            return obj is SpannerDate date
+                ? CompareTo(date)
+                : throw new ArgumentException($"Object must be of type {nameof(SpannerDate)}", nameof(obj));
+        }
+
+        /// <inheritdoc/>
+        public bool Equals(SpannerDate other) => CompareTo(other) == 0;
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj) => obj is SpannerDate other && Equals(other);
+
+        /// <inheritdoc/>
+        public override int GetHashCode() => _days.GetHashCode();
+
+        /// <summary>
+        /// Parses a textual representation as a <see cref="SpannerDate"/>.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// <paramref name="text"/> must be a representation of a date value which can be represented by <see cref="SpannerDate"/>.
+        /// </para>
+        /// </remarks>
+        /// <param name="text">The text to parse. Must not be null.</param>
+        /// <returns>The parsed value.</returns>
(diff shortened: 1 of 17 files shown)


Rishabh-V (commented):
Created a new branch yesterday and did git fetch upstream before pushing, but not sure how PubSub commit came along. Please ignore it for now. I will fix this.

amanda-tarafa:
I'll take a look on Monday

jskeet:
> I'll take a look on Monday

Likewise - or *possibly* this afternoon (UK time).

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// <paramref name="text"/> must be a representation of a date value which can be represented by <see cref="SpannerDate"/>
+        /// </para>
+        /// </remarks>
+        /// <param name="text">The text to parse. Must not be null.</param>
+        /// <returns>The parsed value.</returns>
+        /// <exception cref="FormatException">The value could not be parsed as a <see cref="SpannerDate"/>.</exception>
+        public static SpannerDate Parse(string text)
```
jskeet, line 145: Nit: I'd use an expression-bodied method here.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        public bool Equals(SpannerDate other) => CompareTo(other) == 0;
+
+        /// <inheritdoc/>
+        public override bool Equals(object obj) => obj is SpannerDate other && Equals(other);
+
+        /// <inheritdoc/>
+        public override int GetHashCode() => _days.GetHashCode();
+
+        /// <summary>
+        /// Parses a textual representation as a <see cref="SpannerDate"/>.
```
jskeet, line 135: Does Spanner itself allow multiple cultural representations? I'm tempted to make this an "invariant-only" parse...

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// <paramref name="text"/> must be a representation of a date value which can be represented by <see cref="SpannerDate"/>
+        /// </para>
+        /// </remarks>
+        /// <param name="text">The text to parse. Must not be null.</param>
+        /// <param name="formatProvider">The format provider.</param>
```
jskeet, line 159: It's worth noting that this can be null, and that that means the current culture is used.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// Returns the error message for a FormatException if parsing fails, or null for success.
+        /// </summary>
+        private static string TryParseImpl(string text, IFormatProvider formatProvider, out SpannerDate value)
+        {
+            // TODO: Do we implement the 3 overloads of Parse just like DateTime does? One taking just a string input,
+            // other taking string input and IFormatProvider and third one taking string input,
+            // IFormatProvider and DateTimeStyles enum, which is mostly related to time.
+            GaxPreconditions.CheckNotNull(text, nameof(text));
+
+            // TODO: For now, have passed current culture
```
jskeet, line 195: Not sure what this TODO means.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+            string message = TryParseImpl(text, formatProvider, out value);
+            return message == null;
+        }
+
+        /// <summary>
+        /// Returns the error message for a FormatException if parsing fails, or null for success.
+        /// </summary>
+        private static string TryParseImpl(string text, IFormatProvider formatProvider, out SpannerDate value)
+        {
+            // TODO: Do we implement the 3 overloads of Parse just like DateTime does? One taking just a string input,
```
jskeet, line 190: I don't think we should have a DateTimeStyles, but I'm somewhat concerned that we can parse whole date/time values rather than just dates. Let's discuss.

(We obviously don't want to implement full date/time parsing ourselves, but we could potentially restrict this to "only ISO-8601" completely. If folks want to parse anything else, they can go via DateTime...)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+
+        /// <summary>
+        /// Creates a new instance of the SpannerDate structure to the specified number of days.
+        /// </summary>
+        /// <param name="dayNumber">The day number.</param>
+        /// <returns></returns>
+        public static SpannerDate FromDayNumber(int dayNumber) => new SpannerDate(dayNumber);
+
+        /// <summary>
+        /// Converts value of SpannerDate to The DateTime.
```
jskeet, line 263: Nit: capital T here and below.

I think we also need to document the Kind returned (which I assume is Unspecified)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// </summary>
+        /// <returns>The DateTime value.</returns>
+        public DateTime ToDateTime() => DateTime.MinValue.AddDays(_days);
+
+        /// <summary>
+        /// Converts the value of SpannerDate to The DateTimeOffset.
+        /// </summary>
+        /// <value>
```
jskeet, line 271: I'd remove the `<value>` documentation here and below, but we probably need to specify that it's a DateTimeOffset with an offset of 0. (We should probably check that that's what the DateTimeOffset constructor used here does, too...)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// To date time offset.
+        /// </value>
+        public DateTimeOffset ToDateTimeOffset => new DateTimeOffset(ToDateTime());
+
+        // DateTime Properties
+
+        /// <summary>
+        /// Gets the day component of the date represented by this instance.
```
jskeet, line 279: day => day-of-month (for clarity - MS docs often aren't clear on this).

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// <summary>
+        /// Gets the day of week represented by this instance.
+        /// </summary>
+        /// <value>
+        /// The day of week.
+        /// </value>
+        public DayOfWeek DayOfWeek => ToDateTime().DayOfWeek;
+
+        /// <summary>
+        /// Gets the number of days since January 1, 0001 represented by this instance.
```
jskeet, line 319: Again, I'd remove this until we find that anyone really wants it.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// <summary>
+        /// Compares two values for inequality.
+        /// </summary>
+        /// <param name="lhs">The first value to compare.</param>
+        /// <param name="rhs">The second value to compare.</param>
+        /// <returns><c>false</c> if the two arguments are equal; <c>true</c> otherwise.</returns>
+        public static bool operator !=(SpannerDate lhs, SpannerDate rhs) => !lhs.Equals(rhs);
+
+        /// <summary>Compares two values to determine which is less.</summary>
+        /// <param name="lhs">The value to compare with <paramref name="lhs" />.</param>
```
jskeet, line 345: The paramrefs here and below are incorrect - they're all saying they compare with the same parameter they're defining.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// <returns><c>true</c> if <paramref name="lhs" /> is greater than <paramref name="rhs" />; otherwise, <c>false</c>.</returns>
+        public static bool operator >(SpannerDate lhs, SpannerDate rhs) => lhs.CompareTo(rhs) > 0;
+
+        /// <summary>Compares two values to determine which is greater or equal.</summary>
+        /// <param name="lhs">The value to compare with <paramref name="lhs" />.</param>
+        /// <param name="rhs">The value to compare with <paramref name="rhs" />.</param>
+        /// <returns><c>true</c> if <paramref name="lhs" /> is greater than or equal to <paramref name="rhs" />; otherwise, <c>false</c>.</returns>
+        public static bool operator >=(SpannerDate lhs, SpannerDate rhs) => lhs.CompareTo(rhs) >= 0;
+
+        /// <summary>Adds days to the SpannerDate to compute the new SpannerDate.</summary>
```
jskeet, line 368: I'd remove these three operators. I don't think adding an int to a date is a clear operation.

We can always have a PlusDays or AddDays method instead. (Personally I prefer PlusDays for clarity, but AddDays would be consistent with DateTime)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerDate.cs
```diff
@@ -0,0 +1,408 @@
...
+        /// Performs an implicit conversion from <see cref="DateTime"/> to <see cref="SpannerDate"/>.
+        /// </summary>
+        /// <param name="value">The value.</param>
+        /// <returns>
+        /// The result of the conversion.
+        /// </returns>
+        public static implicit operator SpannerDate(DateTime value) => new SpannerDate(value);
+
+        /// <summary>
+        /// Performs an explicit conversion from <see cref="SpannerDate"/> to <see cref="DateTime"/>.
```
jskeet, line 396: Any reason for this being explicit while the reverse is implicit? I'd argue that if anything, they should be the other way round, given that the conversion to SpannerDate loses information (time) whereas the conversion to DateTime doesn't. I think it's probably clearer if both are explicit though.

jskeet (commented):
Apologies for not getting to this earlier. We should discuss the text aspects.

jskeet:
Btw, I haven't reviewed the tests at all yet. Plenty to talk about before we get there :)

Rishabh-V:
Closing for #8367

PR #8340, opened 2022-04-12
[Bugfix] Allow to read data from big query table with connected google sheet

dpozimski:
Hi everyone,

I would like to propose the code change to include "Drive" scope in BigQueryClient.
Currently when big query table is connected with google sheet the client library is not able to read data from it.
The library throws "Access Denied: BigQuery BigQuery: Permission denied while getting Drive credentials." even when service account has all required privileges (access to the sheet, access to the dataset/table etc...).

By adding these 2 lines of code everything starts to work without any problems.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClient.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClient.cs
@@ -49,11 +49,17 @@ namespace Google.Cloud.BigQuery.V2
     /// </remarks>
     public abstract partial class BigQueryClient : IDisposable
     {
+        /// <summary>
+        /// Google Drive Scope is required when Big Query Table has Google Sheet as a data source.
+        /// </summary>
+        private const string GoogleDriveScope = "https://www.googleapis.com/auth/drive";
+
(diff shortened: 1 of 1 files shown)


amanda-tarafa:
We wouldn't add a Drive scope as a default scope for BigQuery because [Drive scopes may add extra verfication requirements on apps](https://support.google.com/cloud/answer/9110914?hl=en#zippy=%2Cdrive-apis), which might be very surprising for both developers and users, specially if they are not using BigQuery connected to Sheets.

What you can already do is use [`BigQueryClientBuilder`](https://cloud.google.com/dotnet/docs/reference/Google.Cloud.BigQuery.V2/latest/Google.Cloud.BigQuery.V2.BigQueryClientBuilder) to set a [`GoogleCredential`](https://cloud.google.com/dotnet/docs/reference/Google.Apis/latest/Google.Apis.Auth.OAuth2.GoogleCredential) that has been scoped according to your needs, something like (pardon possible typos):

```csharp
var credential = await GoogleCredential.GetApplicationDefaultAsync();
credential = credential.CreateScoped(...); // scopes you need here
var builder = new BigQueryClientBuilder
{
    Credential = credential
};
BigQueryClient client = builder.Build();
```

I'll be closing this PR now, but do feel free to create an issue if you want to discuss further or need help using the client builder.

PR #8371, opened 2022-04-25
Fix all libraries to work with GAX v4.0.0-alpha03

jskeet:
Other than Diagnostics (which @amanda-tarafa is working on) this should now all work with unit, smoke and integration tests.

Once this is in, we can rebase gax-v4 to main, removing the current final regeneration commit and regenerating from scratch. (We can also release mixin packages targeting alpha03.)

Breaking change detection is still expected to fail.


jskeet:
I'll have a look at what's wrong with the other bits of the build.

amanda-tarafa (approved):
LGTM

PR #8372, opened 2022-04-25
tools: Fix version compatibility detection across TFMs

jskeet:
- In the detect-pr-changes.sh script, we expect new code to build with a netstandard2.1 target (at least)
- In the compatibility checker, we check against multiple TFMs if no TFM is specified


jskeet:
The second commit is just for testing purposes, to make sure we do actually detect changes... I'll remove it once it's working.

amanda-tarafa (approved):
LGTM

PR #8414, opened 2022-04-28
chore: Regenerate all APIs

jskeet:
For some reason this final commit wasn't pushed to the gax-v4 branch. Will merge this on green.


jskeet:
Diff is failing because the previous commit basically would fail to build due to not regeneration. Will merge.

PR #8425, opened 2022-04-29
fix: Fix emulator tests running on Grpc.Net.Client on .NET Core 3.1

jskeet:
We need to document this as something that users will need to do
when using the emulator.


amanda-tarafa (approved):
LGTM
Could we automatically add this in the library when we detect the emulator, or even if we detect an insecure connection and are on .netcore3.1? Else we should document this somewhere.

jskeet:
> Could we automatically add this in the library when we detect the emulator, or even if we detect an insecure connection and are on .netcore3.1? Else we should document this somewhere.

I prefer documentation over doing this implicitly - it's an application-wide setting which could have unexpected consequences - best to make the code explicit in user-space IMO.

PR #8430, opened 2022-05-03
fix: Default to using netstandard2.1 for docs

jskeet:
Hoping this will fix the release...


jskeet:
(Self-merging as this is trivial.)

PR #8432, opened 2022-05-03
fix: Specify switch for netcoreapp3.1 to fix testing against emulators

jskeet:
The tests still don't all pass, but that's expected due to
differences between production and the emulators, and happens with
Grpc.Core as well. We need to document the switch in the emulators
page.


amanda-tarafa (approved):
LGTM

PR #8437, opened 2022-05-04
feat(Spanner):decimal tests

Rishabh-V:
Testing Spanner decimal write and read. Should handle decimal.MaxValue and decimal.MinValue

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DecimalReadTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DecimalReadTests.cs
@@ -1,0 +1,56 @@
+﻿// Copyright 2022 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DecimalReadTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            using var connection = _fixture.GetConnection();
+            // Write the decimal value and read it back.
+            await connection.CreateInsertOrUpdateCommand(_fixture.TableName,
+                new SpannerParameterCollection
+                {
+                        new SpannerParameter("DecimalValue", SpannerDbType.Float64, expectedValue),
+                }
+            ).ExecuteNonQueryAsync();
+            var decimalValue = await connection.CreateSelectCommand($"SELECT DecimalValue FROM {_fixture.TableName}").ExecuteScalarAsync<decimal>();
+            Assert.Equal(expectedValue, decimalValue);
```
jskeet, line 53: I wouldn't be surprised by this code failing - I'd expect it to lose precision, given that we're trying to store more than 64 bits of significant data in a FLOAT64 type.

I'll pull the PR to see what actually happens though...

jskeet: Okay, this isn't actually a Spanner issue really - it's just lossy conversions that then overflow. This throws an exception too:

```csharp
using System;

decimal x = decimal.MaxValue;
double d = (double) x;
decimal y = (decimal) d;
```

Basically the conversion to `double` is rounding up, at which point we're out of the range of `decimal`.

jskeet:
Closing this PR as we understand what's going on.

PR #8439, opened 2022-05-04
feat: Provide metadata for the object that has been downloaded

### LINKED ISSUE #7897: Google.Cloud.Stoage.V1 DownloadObjectAsync doesn't provide content version information

jskeet:
Fixes #7897.

This feels almost too easy, and that I may be missing things. We're not guaranteeing which properties *are* populated - partly in case the headers provided by the server change.


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/HashValidatingDownloader.cs
```diff
@@ -27,13 +27,13 @@ namespace Google.Cloud.Storage.V1
     /// Subclass of <see cref="MediaDownloader"/> which validates the data it receives
     /// against a CRC32c hash set in the header.
     /// </summary>
-    internal sealed class HashValidatingDownloader : MediaDownloader
+    internal sealed class HashValidatingDownloader : ContentMetadataRecordingMediaDownloader
```
amanda-tarafa, line 30: I probably would have gone with composition here, but the classes are internal anyway so not important.

amanda-tarafa (approved):
LGTM

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/HashValidatingDownloader.cs
jskeet: Yes, I was wondering about that. I don't *like* the inheritance approach, but it's probably simpler than over-architecting. I definitely wouldn't have done it if it had been public :) (We can refactor if necessary in the future.)

PR #8449, opened 2022-05-05
tools: Adds the PatchVersionCommand

amanda-tarafa:
- `increment-version` only increments minor or prerelease.
- `set-version` doesn't take API groups into account, and can't do so easily, because in a couple of groups the versions are not the same for all APIs (`Google.Cloud.OsLogin`)
- I thought of adding a `--force-patch` flag to `increment-version` but optional flags are currently unsupported, so adding a specific patch command seemed easiest

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/PatchVersionCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/PatchVersionCommand.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2020 Google LLC
+﻿// Copyright 2022 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tools/Google.Cloud.Tools.ReleaseManager/PatchVersionCommand.cs
```diff
@@ -0,0 +1,51 @@
+﻿// Copyright 2020 Google LLC
```
jskeet, line 1: 2022

jskeet (approved):

tools/Google.Cloud.Tools.ReleaseManager/PatchVersionCommand.cs
amanda-tarafa: Done

PR #8459, opened 2022-05-06
chore: Use "transports" generator option

jskeet:
This allows us to remove one piece of special-casing for Compute

We could potentially remove the "regapic" type at some point, but that can come later.

### DIFF
--- a/apis/Google.Cloud.Compute.V1/postgeneration.sh
+++ b/apis/Google.Cloud.Compute.V1/postgeneration.sh
@@ -2,9 +2,6 @@
 
 set -e
 
-# Until we have transports in the generator itself, manually fix this up.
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Compute.V1/postgeneration.sh
```diff
@@ -2,9 +2,6 @@
 
 set -e
 
-# Until we have transports in the generator itself, manually fix this up.
-sed -i 's/ApiTransports.Grpc/ApiTransports.Rest/g' Google.Cloud.Compute.V1/*.g.cs
```
amanda-tarafa, line 6: Nice! And +1 to trying to remove regapic later on

amanda-tarafa (approved):
LGTM

PR #8463, opened 2022-05-10
feat: SpannerConversionOptions infrastructure changes, decimal handling with options, Date handling with options.

Rishabh-V:
Contains 3 commits:

1. Commit 1: 
    - New options in `SpannerConnectionBuilder` and its unit tests. 
    - Changes to ensure that `SpannerConversionOptions` are passed to APIs that require it. Currently, `null` is being passed in few call sites. There is still a place where `SpannerConnection` is not available and null is passed. Marked it as TODO to check if we should pass `SpannerConversionOptions.Default` or do anything else.
2. Commit 2:
    - Default handling for decimal using `SpannerConversionOptions` along with its unit tests and integration tests in GSQL and PostgreSQL dialects.   
3. Commit 3:
   - Default handling for date using `SpannerConversionOptions` along with its unit tests and integration tests in GSQL and PostgreSQL dialects.

Added TODO at couple of places to seek your input.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerFixtureBasePostgre.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerFixtureBasePostgre.cs
@@ -1,0 +1,41 @@
+﻿// Copyright 2022 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.ClientTesting;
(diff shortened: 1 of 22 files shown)


jskeet:
+1 to skipping the tests on the emulator. I'll try to get to this review tomorrow.

jskeet:
Just starting to look at this now - given the size of the PR, please could we split it into three PRs? (Or rather, one PR for the first commit, and then once that's done we can create a second for the second commit etc.) I think that's much more likely to be easy to review and end up with a clean commit history.

Rishabh-V:
> Just starting to look at this now - given the size of the PR, please could we split it into three PRs? (Or rather, one PR for the first commit, and then once that's done we can create a second for the second commit etc.) I think that's much more likely to be easy to review and end up with a clean commit history.

Sure Jon.

jskeet (commented):
Just looking at the first commit, I think I must have not explained my expectations clearly enough... I'd expect the first commit to be *just* infrastructure without reference to any actual options. I'd like to see it *just* making sure that `SpannerConversionOptions` is available anywhere that we want it to be. The actual options would then be each in a separate commit (and ideally PR) entirely.

Rishabh-V:
> Just looking at the first commit, I think I must have not explained my expectations clearly enough... I'd expect the first commit to be _just_ infrastructure without reference to any actual options. I'd like to see it _just_ making sure that `SpannerConversionOptions` is available anywhere that we want it to be. The actual options would then be each in a separate commit (and ideally PR) entirely.

Sure Jon. I think I understood it incorrectly. Will make the changes and raise a new PR.

Rishabh-V:
Closing for #8534

PR #8466, opened 2022-05-11
feat: Passing SpannerConversionOptions to the APIs requiring it.

Rishabh-V:
Passing `SpannerConversionOptions` to the APIs that require it. Option will be used in subsequent PRs where default type handling will be done based on options.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
@@ -41,6 +41,7 @@ namespace Google.Cloud.Spanner.Data
             internal SpannerBatchCommandType CommandType { get; }
             internal Priority Priority { get; }
             internal string Tag { get; }
+            internal SpannerConversionOptions ConversionOptions => SpannerConversionOptions.ForConnection(Connection);
 
             public ExecutableCommand(SpannerBatchCommand command)
             {
@@ -79,7 +80,6 @@ namespace Google.Cloud.Spanner.Data
             private async Task<IReadOnlyList<long>> ExecuteBatchDmlAsync(CancellationToken cancellationToken)
             {
                 await Connection.EnsureIsOpenAsync(cancellationToken).ConfigureAwait(false);
-
                 var transaction = Transaction ?? Connection.AmbientTransaction ?? new EphemeralTransaction(Connection, Priority);
                 ExecuteBatchDmlRequest request = GetExecuteBatchDmlRequest();
                 IEnumerable<long> result = await transaction.ExecuteBatchDmlAsync(request, cancellationToken, CommandTimeout).ConfigureAwait(false);
@@ -92,10 +92,12 @@ namespace Google.Cloud.Spanner.Data
                 {
                     RequestOptions = BuildRequestOptions()
                 };
+                // Avoid calling method multiple times in the loop.
+                var conversionOptions = ConversionOptions;
                 foreach (var command in Commands)
                 {
                     var statement = new Statement { Sql = command.CommandText };
-                    command.Parameters.FillSpannerCommandParams(out var parameters, statement.ParamTypes, null);
+                    command.Parameters.FillSpannerCommandParams(out var parameters, statement.ParamTypes, conversionOptions);
                     statement.Params = parameters;
                     request.Statements.Add(statement);
                 }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
@@ -62,6 +62,7 @@ namespace Google.Cloud.Spanner.Data
             internal QueryOptions QueryOptions { get; }
             internal Priority Priority { get; }
             internal string Tag { get; }
+            internal SpannerConversionOptions ConversionOptions => SpannerConversionOptions.ForConnection(Connection);
 
             public ExecutableCommand(SpannerCommand command)
             {
@@ -126,12 +127,11 @@ namespace Google.Cloud.Spanner.Data
                 var resultSet = await ExecuteReadOrQueryRequestAsync(singleUseReadSettings, effectiveTransaction, cancellationToken)
                         .ConfigureAwait(false);
 
-                var conversionOptions = SpannerConversionOptions.ForConnection(Connection);
                 var enableGetSchemaTable = Connection.Builder.EnableGetSchemaTable;
                 // When the data reader is closed, we may need to dispose of the connection.
                 IDisposable resourceToClose = (behavior & CommandBehavior.CloseConnection) == CommandBehavior.CloseConnection ? Connection : null;
 
-                return new SpannerDataReader(Connection.Logger, resultSet, Transaction?.ReadTimestamp, resourceToClose, conversionOptions, enableGetSchemaTable, CommandTimeout);
+                return new SpannerDataReader(Connection.Logger, resultSet, Transaction?.ReadTimestamp, resourceToClose, ConversionOptions, enableGetSchemaTable, CommandTimeout);
             }
 
             private Task<ReliableStreamReader> ExecuteReadOrQueryRequestAsync(TimestampBound singleUseReadSettings, ISpannerTransaction effectiveTransaction, CancellationToken cancellationToken)
@@ -311,17 +311,12 @@ namespace Google.Cloud.Spanner.Data
 
             private List<Mutation> GetMutations()
             {
-                // Currently, ToProtobufValue doesn't use the options it's provided. They're only
-                // required to prevent us from accidentally adding call sites that wouldn't be able to obtain
-                // valid options. For efficiency, we just pass in null for now. If we ever need real options
-                // from the connection string, uncomment the following line to initialize the options from the connection.
-                // SpannerConversionOptions options = SpannerConversionOptions.ForConnection(SpannerConnection);
-                SpannerConversionOptions conversionOptions = null;
-
+                // Avoid calling method multiple times in the loop.
+                var conversionOptions = ConversionOptions;
                 // Whatever we do with the parameters, we'll need them in a ListValue.
                 var listValue = new ListValue
                 {
-                    Values = { Parameters.Select(x => x.SpannerDbType.ToProtobufValue(x.GetValidatedValue(), conversionOptions)) }
+                    Values = { Parameters.Select(x => x.GetConfiguredSpannerDbType(conversionOptions).ToProtobufValue(x.GetValidatedValue())) }
                 };
 
                 if (CommandTextBuilder.SpannerCommandType != SpannerCommandType.Delete)
@@ -399,9 +394,7 @@ namespace Google.Cloud.Spanner.Data
                     RequestOptions = BuildRequestOptions()
                 };
 
-                // See comment at the start of GetMutations.
-                SpannerConversionOptions options = null;
-                Parameters.FillSpannerCommandParams(out var parameters, request.ParamTypes, options);
+                Parameters.FillSpannerCommandParams(out var parameters, request.ParamTypes, ConversionOptions);
                 request.Params = parameters;
 
                 return request;
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
@@ -178,45 +178,50 @@ namespace Google.Cloud.Spanner.Data
         }
 
         /// <summary>
-        /// The default <see cref="System.Type"/> for this Cloud Spanner type.
+        /// Gets the configured <see cref="System.Type"/> for this SpannerDbType based on <see cref="SpannerConversionOptions"/>.
+        /// If the configuration for this type doesn't exist in options, the default CLR type is returned.
         /// </summary>
-        public System.Type DefaultClrType
+        /// <param name="options">Options to configure the CLR Type for this SpannerDbType.</param>
+        /// <returns>The configured <see cref="System.Type"/>.</returns>
+        internal System.Type GetConfiguredClrType(SpannerConversionOptions options)
         {
-            get
+            switch (TypeCode)
             {
-                switch (TypeCode)
-                {
-                    case TypeCode.Bool:
-                        return typeof(bool);
-                    case TypeCode.Int64:
-                        return typeof(long);
-                    case TypeCode.Float64:
-                        return typeof(double);
-                    case TypeCode.Timestamp:
-                    case TypeCode.Date:
-                        return typeof(DateTime);
-                    case TypeCode.String:
-                        return typeof(string);
-                    case TypeCode.Bytes:
-                        return typeof(byte[]);
-                    case TypeCode.Array:
-                        return typeof(List<>).MakeGenericType(ArrayElementType.DefaultClrType);
-                    case TypeCode.Struct:
-                        return typeof(SpannerStruct);
-                    case TypeCode.Numeric:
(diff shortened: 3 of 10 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -62,6 +62,7 @@ private static DatabaseAdminSettings CreateDatabaseAdminSettings()
             internal QueryOptions QueryOptions { get; }
             internal Priority Priority { get; }
             internal string Tag { get; }
+            internal SpannerConversionOptions ConversionOptions { get; }
```
jskeet, line 65: Given that we keep the connection anyway, could we maybe change this to:

```csharp
internal SpannerConversionOptions ConversionOptions => SpannerConversionOptions.ForConnection(Connection);
```

... and avoid the implicit field?

jskeet: (That can definitely be done later though.)

jskeet (approved):
One potential nit, but otherwise fine. I thought we need to change a DefaultClrType property somewhere as well though? (Making the property delegate to a new method with a parameter.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
Rishabh-V: Done.

Rishabh-V:
> One potential nit, but otherwise fine. I thought we need to change a DefaultClrType property somewhere as well though? (Making the property delegate to a new method with a parameter.)

Right Jon. There are couple of methods which will do the configuration of the default type, one in _write_ path and one in _read_ path. I was planning to bring them in next PR with corresponding options. 
Without option, they would just return the `DefaultClrType` and  default `SpannerDbType`. Please let me know if they should be part of this infrastructure PR and I will do that. Thank you.

jskeet:
I'd try to get all of the non-option-specific infrastructure into the one PR, then deal with the options one by one (which shouldn't require any infrastructure changes).

Rishabh-V:
> I'd try to get all of the non-option-specific infrastructure into the one PR, then deal with the options one by one (which shouldn't require any infrastructure changes).

Sure Jon. Including them in this PR.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
```diff
@@ -41,6 +41,7 @@ private class ExecutableCommand
             internal SpannerBatchCommandType CommandType { get; }
             internal Priority Priority { get; }
             internal string Tag { get; }
+            internal SpannerConversionOptions ConversionOptions => SpannerConversionOptions.ForConnection(Connection);
```
jskeet, line 44: Are we okay to evaluate this one lazily as well? The comment of "The SpannerConnection can be modified, but other objects should be fine" made me somewhat nervous. It's been a while since I looked at SpannerConnection - maybe the bits that can be modified don't include the builder where we get the conversion options? (Or maybe we can just assert that they *shouldn't* change, and behavior isn't guaranteed if they do?)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -311,17 +311,10 @@ private async Task<int> ExecuteMutationsAsync(CancellationToken cancellationToke
...
-                // valid options. For efficiency, we just pass in null for now. If we ever need real options
-                // from the connection string, uncomment the following line to initialize the options from the connection.
-                // SpannerConversionOptions options = SpannerConversionOptions.ForConnection(SpannerConnection);
-                SpannerConversionOptions conversionOptions = null;
-
                 // Whatever we do with the parameters, we'll need them in a ListValue.
                 var listValue = new ListValue
                 {
-                    Values = { Parameters.Select(x => x.SpannerDbType.ToProtobufValue(x.GetValidatedValue(), conversionOptions)) }
+                    Values = { Parameters.Select(x => x.GetConfiguredSpannerDbType(ConversionOptions).ToProtobufValue(x.GetValidatedValue(), ConversionOptions)) }
```
jskeet, line 317: Is it worth evaluating ConversionOptions once before the implicit loop here? Not sure. (We can definitely optimize later if we really need to.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -177,6 +177,14 @@ public DbType DbType
             }
         }
 
+        /// <summary>
+        /// Gets the configured <see cref="System.Type"/> for this SpannerDbType based on <see cref="SpannerConversionOptions"/>.
+        /// If the configuration for this type doesn't exist in options, the <see cref="DefaultClrType"/> is returned.
+        /// </summary>
+        /// <param name="options">Options to configure the CLR Type for this SpannerDbType.</param>
+        /// <returns>The configured <see cref="System.Type"/>.</returns>
+        internal System.Type GetConfiguredClrType(SpannerConversionOptions options) => DefaultClrType;
```
jskeet, line 186: I'd move the implementation to this method now, and have `DefaultClrType` call this - that way when we implement an option, there's less code to change.

We might want to clarify the docs for DefaultClrType, too.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
```diff
@@ -254,16 +254,16 @@ private void FillSpannerInternalValues(MapField<string, Value> valueDictionary,
             foreach (var parameter in _innerList)
             {
                 valueDictionary[GetCorrectedParameterName(parameter.ParameterName)]
-                    = parameter.SpannerDbType.ToProtobufValue(parameter.GetValidatedValue(), options);
+                    = parameter.GetConfiguredSpannerDbType(options).ToProtobufValue(parameter.GetValidatedValue(), options);
```
jskeet, line 257: Do we definitely need ToProtobufValue to still accept an options if it's part of GetConfiguredSpannerDbType?

The answer may well be yes - but now is the time to check.

jskeet (commented):
Looking really close now - just a few things to look at.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
Rishabh-V: `ToProtobufValue` doesn't need options. I found that out the hard way that once we reach this method, its already late to make any type configuration changes based on options. Most of the tests and few call sites pass null for options.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
Rishabh-V: Sure. Will do so.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
Rishabh-V: Good catch. I will make this change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
Rishabh-V: Yes, its internal and we can remove it . Contrary to YAGNI, I was just wondering if there may be a scenario in future where we may need to do some processing in this method :) 
Would you like this change to be part of this PR?

jskeet: Yes please - again, it's a matter of "get all the infrastructure to where we want it".

I think if we've got a fair amount of flexibility within SpannerDbType, and that's what we call ToProtobufValue on, I think it should be fine to remove it.

Rishabh-V: Sure Jon.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
Rishabh-V: Okay. Sure. I will make it lazy in SpannerCommand.ExecutableCommand as well. 

Just for my understanding - In this class, there is a foreach loop so `SpannerConversionOptions` will be evaluated multiple times. Given that this class has just one internal method and we will always require option in that, is it better to make it lazy or just initialize options in the constructor?

>It's been a while since I looked at SpannerConnection - maybe the bits that can be modified don't include the builder where we get the conversion options? (Or maybe we can just assert that they shouldn't change, and behavior isn't guaranteed if they do?)

I think the builder is immutable. But I didn't understand the action required here. Could you please help.

jskeet: Will look tomorrow - I'm signed out for the day now I'm afraid.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
Rishabh-V: Now that we have made it lazy, so options construction should happen only once. We should be good now?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
Rishabh-V: Have moved the implementation to method. Could you please help with what I need to update in docs for `DefaultClrType`?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
Rishabh-V: Done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
Rishabh-V: I had a look at it again and now I think I understand it a little better.
We have used Lazy initialization now. So, options will be evaluated the first time they are needed. Post that if `ConnectionString` is updated, it will change the builder, but won't be honored as options are already constructed.
How would you like it to be handled?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
```diff
@@ -51,6 +54,8 @@ public ExecutableCommand(SpannerBatchCommand command)
                 CommandType = command.CommandType;
                 Priority = command.Priority;
                 Tag = command.Tag;
+                _lazyConversionOptions = new Lazy<SpannerConversionOptions>(
```
jskeet, line 57: Sorry, I wasn't suggesting using `Lazy<T>` here - I was only suggesting that we evaluate the ConversionOptions property as a local variable before the foreach loop around line 102. Even that is probably premature optimization, to be honest. I think we should probably just go back to making the property call SpannerConversionOptions.ForConnection.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -62,6 +62,9 @@ private static DatabaseAdminSettings CreateDatabaseAdminSettings()
             internal QueryOptions QueryOptions { get; }
             internal Priority Priority { get; }
             internal string Tag { get; }
+            internal SpannerConversionOptions ConversionOptions => _lazyConversionOptions.Value;
+
+            private readonly Lazy<SpannerConversionOptions> _lazyConversionOptions;
```
jskeet, line 67: Ditto here, I don't think a `Lazy<T>` is called for.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -311,17 +315,10 @@ private async Task<int> ExecuteMutationsAsync(CancellationToken cancellationToke
...
-                // valid options. For efficiency, we just pass in null for now. If we ever need real options
-                // from the connection string, uncomment the following line to initialize the options from the connection.
-                // SpannerConversionOptions options = SpannerConversionOptions.ForConnection(SpannerConnection);
-                SpannerConversionOptions conversionOptions = null;
-
                 // Whatever we do with the parameters, we'll need them in a ListValue.
                 var listValue = new ListValue
                 {
-                    Values = { Parameters.Select(x => x.SpannerDbType.ToProtobufValue(x.GetValidatedValue(), conversionOptions)) }
+                    Values = { Parameters.Select(x => x.GetConfiguredSpannerDbType(ConversionOptions).ToProtobufValue(x.GetValidatedValue())) }
```
jskeet, line 321: This is another place we *could* extract ConversionOptions to a local variable (just above the declaration of `listValue`) to avoid calling evaluating it more than once - but again, that may well be premature optimization anyway.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -178,46 +178,51 @@ public DbType DbType
...
+                case TypeCode.Numeric:
+                    if (TypeAnnotationCode == TypeAnnotationCode.PgNumeric)
+                    {
+                        return typeof(PgNumeric);
+                    }
+                    return typeof(SpannerNumeric);
+                case TypeCode.Json:
+                    return typeof(string);
+                default:
+                    //if we don't recognize it (or its a struct), we use the google native wellknown type.
```
jskeet, line 216: Let's fix this comment while we're here:

```suggestion
                    // If we don't recognize it (or it's a struct), we use the protobuf Value well-known type.
```

jskeet (approved):
Okay, just one final thing in terms of not using `Lazy<T>`. (I didn't go into enough detail earlier, but basically I'm trying to avoid adding a new field if we can, especially as most of the time I expect we'll be using the defaults.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
Rishabh-V: My apologies for interpreting it incorrectly. I have now reverted the `Lazy<T>` change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
Rishabh-V: Done.

jskeet (approved):
Looks good, thanks!

PR #8472, opened 2022-05-12
feat:Default handling of decimal value based on SpannerConversionOptions.

Rishabh-V:


jskeet (commented):
I'll need to wait until a bit later to look through this properly, but there are lots of things ending with "Postgre" which I think should actually be "Postgres" or "PostgreSql".

Rishabh-V:
> I'll need to wait until a bit later to look through this properly, but there are lots of things ending with "Postgre" which I think should actually be "Postgres" or "PostgreSql".

Sure Jon. I will fix it right away.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerFixtureBasePostgreSql.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// <summary>
+    /// Base classes for test fixtures for Spanner PostgreSQL database.
+    /// </summary>
+    public abstract class SpannerFixtureBasePostgreSql : CloudProjectFixtureBase
+    {
+        public SpannerTestDatabasePostgreSql Database { get; }
+
+        public SpannerFixtureBasePostgreSql()
+        {
+            GrpcInfo.EnableSubchannelCounting();
```
jskeet, line 30: I would suggest removing this line, unless we really need subchannel counting in Postgres tests.
(This line would need to be conditionalized for GAX v4, as GrpcInfo only works with Grpc.Core.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabasePostgreSql.cs
```diff
@@ -0,0 +1,197 @@
...
+        private static readonly object s_lock = new object();
+        private static SpannerTestDatabasePostgreSql s_instance = null;
+
+        /// <summary>
+        /// Fetches the PostgreSQL database, creating it if necessary.
+        /// </summary>
+        /// <param name="projectId">The project ID to use, typically from a fixture.</param>
+        public static SpannerTestDatabasePostgreSql GetInstance(string projectId)
+        {
+            if (s_instance == null)
```
jskeet, line 42: This isn't performance sensitive - I'd suggest dropping double-checked locking for the sake of simplicity.
(Note that in its current form, the check for the project ID being incorrect would almost *never* fail, even if it should.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabasePostgreSql.cs
```diff
@@ -0,0 +1,197 @@
...
+            SpannerClientCreationOptions = new SpannerClientCreationOptions(builder);
+            var databaseBuilder = builder.WithDatabase(SpannerDatabase);
+            ConnectionString = databaseBuilder.ConnectionString;
+            DatabaseName = databaseBuilder.DatabaseName;
+
+            MaybeCreateInstanceOnEmulator(projectId);
+            if (Fresh)
+            {
+                // Skip creating PostgreSQL database in Emulator, as it would fail.
+                if (!SpannerClientCreationOptions.UsesEmulator)
```
jskeet, line 109: Rather than skipping the database creation, I think it would be better to just fail hard - otherwise we're effectively pretending that we've made sure a database exists when we haven't.

Basically, we shouldn't run any tests that require Postgres if we're trying to run on the emulator, IMO. Quite how those are skipped is up for debate, but I don't think we should just continue here as if it's all fine.

jskeet (commented):
Just reviewed the first three files - I'd really like to see that infrastructure (which has nothing to do with handling of decimal values) in a separate PR. I'll keep reviewing the actual decimal code here for now though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DecimalReadTests.cs
```diff
@@ -0,0 +1,211 @@
...
+            Assert.Equal(expectedValue, decimalValue);
+        }
+
+        // Test Case: Using decimal with Float64 column in database. SpannerConversionOptions set to use Numeric for decimal.
+        // SpannerDbType is explicitly set for the parameter.
+        // The configuration should be ignored in .NET to Spanner path, as SpannerDbType of parameter is specified.
+        // The configuration should be honored in Spanner to .NET path if ClrType isn't specified while reading.
+        [Theory]
+        [MemberData(nameof(TestDecimals))]
+        public async Task WriteThenRead_ShouldBeEqual_UseOptions(decimal expectedValue)
```
jskeet, line 72: If this is meant to be for *read* tests, I'd expect to just insert rows into the table to start with.
The parameter type used to insert the data won't change how the data from that table is read.

It would be fine to have tests that use a parameter and basically return that parameter value, without reading from any actual table. (I can't remember the syntax for that right now, but it's there somewhere.)

Basically I think we need a bit more clarity about what each test is meant to achieve. I see it as:

- When writing, how do we send a parameter to Spanner
- When reading, how do we receive a value from Spanner

... in each case of "explicitly use Float64, explicitly use Numeric, implicitly use Float64, implicitly use Numeric due to connection options".

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DecimalReadTests.cs
```diff
@@ -0,0 +1,211 @@
...
+            var decimalValue = await connection.CreateSelectCommand($"SELECT DecimalValue FROM {_fixture.TableName}").ExecuteScalarAsync<decimal>();
+            Assert.Equal(expectedValue, decimalValue);
+        }
+
+        // Test Case: Using decimal with Float64 column in database without passing SpannerDbType of the parameter.
+        // SpannerConversionOptions is set to use Numeric for decimal.
+        // This should throw SpannerException as Float64 column expects Float64 value but Numeric value is passed.
+        [Theory]
+        [MemberData(nameof(TestDecimals))]
+        public async Task WriteWithoutSpannerDbType_UseOptions(decimal expectedValue)
```
jskeet, line 117: I don't think this should be in DecimalReadTests.

(We can rename the test of course...)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DecimalReadTestsPostgreSql.cs
```diff
@@ -0,0 +1,227 @@
...
+using Google.Cloud.Spanner.V1;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.IntegrationTests
+{
+    [Collection(nameof(DecimalTableFixturePostgreSql))]
+    public class DecimalReadTestsPostgreSql
```
jskeet, line 24: I suggest we get to the stage where we've got agreement on what the tests should look like for regular Spanner, then implement similar tests for Postgres. I don't mind if that's in the same PR or a different one.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
```diff
@@ -186,6 +186,32 @@ public void LogCommitStats()
             Assert.False(connectionStringBuilder.LogCommitStats);
         }
 
+        [Fact]
```
jskeet, line 189: I'm fine with both of these tests being in the same PR.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -19,21 +19,32 @@ namespace Google.Cloud.Spanner.Data
...
 
         /// <summary>
         /// True to return DBNull.Value for null values; false to return a null reference.
         /// </summary>
         internal bool UseDBNull { get; }
 
-        private SpannerConversionOptions(bool useDBNull)
+        /// <summary>
+        /// True to use <see cref="SpannerDbType.Numeric"/> as default SpannerDbType for decimal values; false to use <see cref="SpannerDbType.Float64"/>.
+        /// This should be used, only when working with Spanner Google Standard SQL (GSQL) dialect database.
```
jskeet, line 31:
```suggestion
        /// This should only be used when working with Spanner Google Standard SQL (GSQL) dialect database.
```

(And same below.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -19,21 +19,32 @@ namespace Google.Cloud.Spanner.Data
...
+        /// </summary>
+        internal bool UseSpannerNumericForDecimal { get; }
+
+        /// <summary>
+        /// True to use <see cref="SpannerDbType.PgNumeric"/> as default SpannerDbType for decimal values; false to use <see cref="SpannerDbType.Float64"/>.
+        /// This should be used, only when working with Spanner PostgreSQL dialect database.
+        /// </summary>
+        internal bool UsePgNumericForDecimal { get; }
```
jskeet, line 39: Is it ever useful to specify both? Would it perhaps make things simpler to have a single parameter of "DefaultTypeForDecimal=PgNumeric" or "DefaultTypeForDecimal=Numeric"?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -46,6 +57,6 @@ private SpannerConversionOptions(bool useDBNull)
         /// Determines the right conversion options to use based on the connection string of the given connection string builder.
         /// </summary>
         internal static SpannerConversionOptions ForConnectionStringBuilder(SpannerConnectionStringBuilder builder) =>
-            builder == null ? Default : builder.UseClrDefaultForNull ? s_useClrDefaultForNull : s_useDBNullForNull;
+            builder == null ? Default : new SpannerConversionOptions(!builder.UseClrDefaultForNull, builder.UseSpannerNumericForDecimal, builder.UsePgNumericForDecimal);
```
jskeet, line 60: Depending on how complex this gets (with multiple options) we *may* want to have a set of "common configurations" and reuse the same objects.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
```diff
@@ -603,14 +604,16 @@ private object ConvertToClrTypeImpl(Value wireValue, System.Type targetClrType,
                         return null;
                     case Value.KindOneofCase.StringValue:
                         return SpannerNumeric.Parse(wireValue.StringValue);
+                    case Value.KindOneofCase.NumberValue:
```
jskeet, line 607: Not specific to this line: this method is really getting out of hand now. Separate to this PR, we should try to refactor it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
```diff
@@ -603,14 +604,16 @@ private object ConvertToClrTypeImpl(Value wireValue, System.Type targetClrType,
...
+                        return SpannerNumeric.FromDecimal(Convert.ToDecimal(wireValue.NumberValue), LossOfPrecisionHandling.Truncate);
                     default:
                         throw new InvalidOperationException(
                             $"Invalid Type conversion from {wireValue.KindCase} to {targetClrType.FullName}");
                 }
             }
             if (targetClrType == typeof(PgNumeric))
             {
-                if (TypeCode != TypeCode.Numeric || TypeAnnotationCode != TypeAnnotationCode.PgNumeric)
+                if ((TypeCode != TypeCode.Numeric || TypeAnnotationCode != TypeAnnotationCode.PgNumeric) && TypeCode != TypeCode.Float64)
```
jskeet, line 616: I think this might be clearer as something like:

```csharp
bool isPgNumeric = TypeCode == TypeCode.Numeric && TypeAnnotationCode == TypeAnnotationCode.PgNumeric;
bool isFloat64 = TypeCode == TypeCode.Float64;

if (!isPgNumeric && !isFloat64)
{
    ...
}
```

Do we even need to check the type code, if the type annotation code is PgNumeric? (Do we ever end up with a PgNumeric annotation for a non-numeric type?)

I think we should also clarify the exception message to indicate that it can only be used for numeric results, and only when working with a Postgres dialect.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -46,7 +45,7 @@ public sealed partial class SpannerDbType
         public static SpannerDbType Int64 { get; } = new SpannerDbType(TypeCode.Int64);
 
         /// <summary>
-        /// 64 bit floating point number.
+        /// 64 bit floating point number. This is same as Float8 in PostgreSQL dialect.
```
jskeet, line 48:
```suggestion
        /// 64 bit floating point number. This is equivalent to Float8 in the PostgreSQL dialect.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -192,6 +191,14 @@ internal System.Type GetConfiguredClrType(SpannerConversionOptions options)
                 case TypeCode.Int64:
                     return typeof(long);
                 case TypeCode.Float64:
+                    if (options != null && options.UseSpannerNumericForDecimal)
```
jskeet, line 194: I don't think this is right, because we're not dealing with decimal - we're dealing with a double, inherently.

If someone wants a decimal -> numeric conversion by default, that doesn't mean they want float64 -> decimal.
I suspect we may need directional options, basically - which will be slightly more longwinded (but precise and predictable) for the customer to specify... it'll make this code simple though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
```diff
@@ -160,7 +161,28 @@ internal object GetValidatedValue()
...
+                    }
+                    else if (options != null && options.UsePgNumericForDecimal)
+                    {
+                        return SpannerDbType.PgNumeric;
+                    }
+                }
+            }
+
+            // If we are here, use defaults.
+            return SpannerDbType;
```
jskeet, line 184: We definitely need to comment on the SpannerDbType property that it doesn't take the connection string into account.

jskeet (commented):
I've *mostly* reviewed the production code, as changes there will affect how we do everything else.

Rishabh-V (commented):
Thanks Jon. Will discuss about tests, directional option with you and make the changes tomorrow.

Rishabh-V:
Closing as changes merged with #8534

PR #8484, opened 2022-05-17
fix!: Address changes in Spanner ADO API due to netstandard2.1 async methods

jskeet:
BREAKING CHANGE: CommitAsync has been renamed to avoid clashing with a standard method that differs only by return type.

Users who were not using the returned timestamp won't notice any difference, as the CommitAsync method returning a Task will return the same task as before - it just doesn't have the timestamp in (at least in terms of compile-time checking; the actual returned task is the same).

Amanda: no need to review this before you're back, but we do need to do so before we release the new major version.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
@@ -353,24 +353,29 @@ namespace Google.Cloud.Spanner.Data
         /// Commits the database transaction synchronously, returning the database timestamp for the commit via <paramref name="timestamp"/>.
         /// </summary>
         /// <param name="timestamp">Returns the UTC timestamp when the data was written to the database.</param>
-        public void Commit(out DateTime timestamp) => timestamp = Task.Run(() => CommitAsync(default)).ResultWithUnwrappedExceptions();
+        public void Commit(out DateTime timestamp) => timestamp = Task.Run(() => CommitAndReturnTimestampAsync(default)).ResultWithUnwrappedExceptions();
 
-        // FIXME: This is "new" because there's now CommitAsync in the BCL. That returns just Task, rather than Task<DateTime>,
-        // which is awkward...        
+        /// <inheritdoc />
+        public
+#if NETSTANDARD2_1_OR_GREATER
+            override
+#endif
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -353,24 +353,29 @@ Task<IEnumerable<long>> ISpannerTransaction.ExecuteBatchDmlAsync(ExecuteBatchDml
...
+        /// Commits the database transaction asynchronously, returning the UTC timestamp when the data was written to the database.
         /// </summary>
         /// <param name="cancellationToken">A cancellation token used for this task.</param>
         /// <returns>Returns the UTC timestamp when the data was written to the database.</returns>
-#if NET462
-        public Task<DateTime> CommitAsync(CancellationToken cancellationToken = default)
-#else
-        public new Task<DateTime> CommitAsync(CancellationToken cancellationToken = default)
-#endif
+        public Task<DateTime> CommitAndReturnTimestampAsync(CancellationToken cancellationToken = default)
```
amanda-tarafa, line 370: So that we don't have to add the breaking change nor this method I think we can add an intermediate internal class, `SpannerTransactionImpl` that overrides `CommitAsync` for asynchrony and then we can have the public `SpannerTransaction` hide the `CommitAsync` implementation to change the return type.

What do you think?

amanda-tarafa (commented):
I've left a note with an alternative I think.

Also, I only reviewed the last commit as there seems to be a lot of commits in this PR, possbily because of the base branch being gax-v4.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
jskeet: I did wonder about that. https://docs.microsoft.com/en-us/dotnet/core/compatibility/ states that it's a "requires judgement" change - we can probably be careful enough, and it's across a major boundary anyway.

Will try this in a new PR.

jskeet: Alternative approach is now in https://github.com/googleapis/google-cloud-dotnet/pull/8535 - let me know if that's not what you meant.

jskeet:
Closed in favour of #8535

PR #8493, opened 2022-05-20
Deliberately bad code

jskeet:


jskeet:
This is just to test whether the analyzers are used in CI.

jskeet:
Okay, these changes passed, which they shouldn't if the analyzers were running. Closing this PR.

PR #8494, opened 2022-05-20
Remove analyzers

jskeet:
(First commit is manual; second is autogenerated.)

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -119,11 +119,7 @@ then
   exit 0
 fi
 
-# First build the analyzers, for use in everything else.
 log_build_action "(Start) build.sh"
-log_build_action "Building analyzers"
-
-dotnet publish -nologo -clp:NoSummary -v quiet -c Release -f netstandard2.0 tools/Google.Cloud.Tools.Analyzers
 
 # Then build the requested APIs, working out the test projects as we go.
 > AllTests.txt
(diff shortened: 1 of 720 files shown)


Rishabh-V (approved):

jskeet:
Thanks @Rishabh-V - I'll wait until Amanda is back before merging. (This isn't urgent; if we need to rebase main into gax-v4 again before this is merged, I'll just rerun the project generation part.)

jskeet:
I'm very confused as to some of the changes shown in the second commit. Looking into it more now...

build.sh
```diff
@@ -119,11 +119,7 @@ then
   exit 0
 fi
 
-# First build the analyzers, for use in everything else.
 log_build_action "(Start) build.sh"
-log_build_action "Building analyzers"
-
-dotnet publish -nologo -clp:NoSummary -v quiet -c Release -f netstandard2.0 tools/Google.Cloud.Tools.Analyzers
```
amanda-tarafa, line 126: I think we were not using them at all, we are building for `netstandard2.0` here but from the next commit it seems that we were looking for them on the `netstandard1.3` path. Unless I'm missing something.

amanda-tarafa (approved):
LGTM. I'm fine with this. A few notes on an internal email though.

Also, flagging that you didn't delete the actual analyzers code, I'm fine with keeping it around for reference but not sure if that is intentional or just something that the commit didn't picked up.

jskeet:
I suspect that not removing the analyzer code itself was just a git mistake on my part - I'll do that in a separate PR.

PR #8495, opened 2022-05-20
Prototype of fixing numeric truncation

jskeet:
We'd want tests for BigQueryBigNumeric as well, and I strongly
suspect the implementation should be improved too... but it's a
straw man for discussion.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Numeric.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Numeric.cs
@@ -452,14 +452,38 @@ namespace Google.Cloud.BigQuery.V2
 
             // FIXME: Awful performance!
             string text = ToString();
-            // Handles overflow
-            decimal dec = decimal.Parse(text, CultureInfo.InvariantCulture);
-            if (lossOfPrecisionHandling == LossOfPrecisionHandling.Throw && this != Type.ValueFromDecimal(dec, LossOfPrecisionHandling.Truncate))
+
+            // For non-extreme values, we can just parse the text and we're fine.
+            int maxSafeLength = 28 + (_scaledValue < 0 ? 1 : 0) + (text.Contains('.') ? 1 : 0);
+            if (text.Length < maxSafeLength)
             {
-                // TODO: Is this the right exception to use?
+                return decimal.Parse(text, CultureInfo.InvariantCulture);
+            }
+
+            // Otherwise, we may need to care about a loss of precision. Try a round-trip first.
+            decimal dec = decimal.Parse(text, CultureInfo.InvariantCulture);
+            if (this == Type.ValueFromDecimal(dec, LossOfPrecisionHandling.Truncate))
+            {
+                return dec;
+            }
+            // If we've lost precision already and that's unacceptable, throw.
+            if (lossOfPrecisionHandling == LossOfPrecisionHandling.Throw)
+            {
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Numeric.cs
```diff
@@ -452,14 +452,38 @@ public decimal ToDecimal(LossOfPrecisionHandling lossOfPrecisionHandling)
...
+            }
+            // If we've lost precision already and that's unacceptable, throw.
+            if (lossOfPrecisionHandling == LossOfPrecisionHandling.Throw)
             {
-                // TODO: Is this the right exception to use?
                 throw new InvalidOperationException("Conversion would lose information");
             }
-            return dec;
+            // Otherwise, we need to truncate - which means truncating *before* parsing.
+            int length = Math.Min(text.Length - 1, maxSafeLength + 3);
```
Rishabh-V, line 475: I may be missing something obvious, but I couldn't understand the +3 to maxSafeLength. (given that sign and '.' are already included in maxSafeLength)

Rishabh-V (approved):
LGTM

Just one question for my understanding.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Numeric.cs
jskeet: It's very much a "this is just a prototype, my maths might be wrong - this is inefficient but safe" :)

But we would at least want `maxSafeLength + 1`, as there might be a digit which *can* be parsed, but isn't safe to *assume* will be parsed just based on the length.

Rishabh-V: Right Jon :) Thanks for helping me understand. 

One more thought - If we plan to use the same fix everywhere, PgNumeric (where we can have very large string due to large precision and scale) may suffer a little with substring in loop.

jskeet: We should definitely apply better than "Friday-afternoon thinking" for the loop - but even if the original string is very large, we'd start off with a relatively small substring due to the `Math.Min` call. We shouldn't need more than two or three iterations.

jskeet:
Urgh - I need to rebase this onto the new gax-v4....

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Numeric.cs
```diff
@@ -452,14 +452,38 @@ public decimal ToDecimal(LossOfPrecisionHandling lossOfPrecisionHandling)
 
             // FIXME: Awful performance!
             string text = ToString();
```
amanda-tarafa, line 454: I was wondering if instead of doing truncation on this method, we could have `ToString(int maxDigits, LossOfPrecisionHandling lossOfPrecisionHandling)` and we would call that here `ToString(28, lossOfPrecisionHandling)`. If we get the string already truncated or a failure then we only need to parse to decimal. And I think it's easier to truncate when forming the string? I'll try to push a PR with that later.

But I'm fine with this as well.

jskeet: Frankly I get confused every time I look at this code, so whatever you can do would be good :)

PR #8496, opened 2022-05-20
Prepare console log formatter for release

### LINKED ISSUE #5578: Update Google.Cloud.Logging.Console with GA Microsoft.Extensions.Logging.Console
Now Microsoft.Extensions.Logging.Console is GA, we can productionize our logging formatter.

jskeet:
(There's no rush on this PR.)

Fixes #5578, if we're happy to then release a beta.


apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
```diff
@@ -23,120 +23,195 @@
...
+        static bool ContainsFormatParameters(IEnumerable<KeyValuePair<string, object>> fields)
         {
-            string message = logEntry.Formatter(logEntry.State, logEntry.Exception);
-            if (logEntry.Exception == null && message == null)
-            {
-                return;
-            }
-            // TODO: Use PooledByteBufferWriter when it's public...
-            using var output = new MemoryStream();
+            using (var iterator = fields.GetEnumerator())
```
amanda-tarafa, line 128: Wondering if this could be done with someting like

`fields.Any(pair => pair.Key != "{OriginalFormat}")`

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/ConsoleLoggerExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+using System;
+
+namespace Google.Cloud.Logging.Console;
+
+/// <summary>
+/// Extension methods to make it easier to configure <see cref="GoogleCloudConsoleFormatter"/>.
+/// </summary>
+public static class ConsoleLoggerExtensions
+{
+    // TODO: Should these be called AddGoogleCloudConsole?
```
amanda-tarafa, line 25: Yep, I'd go with this.

apis/Google.Cloud.Logging.Console/SampleApp/Pages/Error.cshtml.cs
```diff
@@ -0,0 +1,27 @@
+using System.Diagnostics;
```
amanda-tarafa, line 1: No licensences on these files?
Are most of these just from the project template?

amanda-tarafa (approved):
LGTM. As for the questions...

- Property name for the state (currently format_parameters, which is also what the Functions Framework and Google.Cloud.Diagnostics.Common use)
  I think effectively we only get format_parameters there. So that's fine. And it matches everything else we are doing so less breakage.
- Should we just not log anything if the message and exception are null?
  The empty message is fine, there's more information possibly, including scopes, etc. that might be useful. If someone very intentionally logged an empty message we should propagate it.
- Should we log an empty scopes array when a scope provider is present but there aren't any scopes? (Currently we don't - before this commit we did.)
 No, I think it could interfere with search of log entries when filtering by having scopes or not. Not sure how the empty array is stored in the backend but in any case, it's better if we don't store anything in that case.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
jskeet: Yes, I *think* you're right. Will think about it more when editing for other things.

jskeet (commented):
Thanks, will add another commit with the simple changes, and discuss the sample project with you.

apis/Google.Cloud.Logging.Console/SampleApp/Program.cs
```diff
@@ -0,0 +1,42 @@
+// Import the namespace required for the AddGoogleCloudConsole extension method
```
amanda-tarafa, line 1: Maybe this one should have a license now?

amanda-tarafa (approved):
LGTM

PR #8524, opened 2022-05-24
chore: Fix line endings for Analytics and API catalog

jskeet:
(These were broken by the earlier release.)


jskeet:
(Will self-merge.)

PR #8531, opened 2022-05-26
Prototype of a sample that converts a SpannerParameter into a value

jskeet:
Do we actually need it this end-to-end without having an integration
test? Maybe not...


jskeet:
Closing this now as the code is in #8534

Issue #5837, opened 2021-01-13
[Spanner] ClrType `decimal` is currently mapped by default to `FLOAT64`


olavloite:
The .NET `decimal` ClrType is [currently mapped](https://github.com/googleapis/google-cloud-dotnet/blob/34b71e8b87efdbc3043906d9b2dd7d1b8e9bb43d/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs#L279) to the Spanner `FLOAT64` data type. That is correct in the sense that both are floating point types. It does however not correspond with the default mapping for other database systems:
1. SQL Server maps it to `DECIMAL(29,4)`
2. PostgreSQL maps it to `numeric` (which is a fixed point exact type)
3. MySQL maps it to `decimal` (which is a fixed point exact type).

This might cause some surprises for users moving from other DBMS'es to Spanner. We should consider changing the default mapping to `NUMERIC`.

jskeet:
It sounds like that could break users who are currently using FLOAT64. The FromClrType method looks like it's only called by SpannerParameter.SpannerDbType, if the type hasn't been set explicitly. So I can see that change causing a problem if someone were trying to update a FLOAT64 column, and only specifying the value (as a decimal). Does that sounds correct as a compatibility issue?

We could make this a [connection string option](https://googleapis.dev/dotnet/Google.Cloud.Spanner.Data/latest/connection_string.html), defaulting to the current behavior in the current major version, but changing the default to `NUMERIC` in a later major version.

jskeet:
(Assigned to @olavloite just so that it *has* an assignee, but we don't need to decide just yet who will make the change.)

olavloite:
> Does that sounds correct as a compatibility issue?

Yes, but I think that Spanner will [automatically coerce](https://cloud.google.com/spanner/docs/conversion_rules) that value to a `FLOAT64` server side. I'll try to see if I can test that to verify if my assumption is right.

amanda-tarafa:
But even if Spanner coerces the NUMERIC value to FLOAT64, the potential precission loss has already happened, just because of going through NUMERIC first? So, the surprise might not be a conversion error, but (arguiably worst) a silent loss of precission?
Or am I missing something?

olavloite:
[...]

PR #8534, opened 2022-05-27
feat: default handling of decimal and Float64

Rishabh-V:
Default handling of decimal and Float64 with unit tests.
 
The implementation uses string splitting and parsing and makes use of dictionary to keep strongly typed mappings. 

Commit 1: Decimal and Float64 default handling.
Commit 2: Builds on top of Commit 1 and adds support for default DateTime and Date handling.

Will raise a new PR for integration tests once this PR is good.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
@@ -30,6 +30,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using System.Transactions;
 using Xunit;
+using static Google.Cloud.Spanner.Data.SpannerConversionOptions;
 
 namespace Google.Cloud.Spanner.Data.Tests
 {
@@ -989,7 +990,134 @@ namespace Google.Cloud.Spanner.Data.Tests
                 It.IsAny<CallSettings>()), Times.Exactly(10));
         }
 
-        internal static SpannerConnection BuildSpannerConnection(Mock<SpannerClient> spannerClientMock)
+        public static IEnumerable<object[]> ConfiguredSpannerDbTypes()
+        {
+            // Format : ClrToSpannerTypeDefaultMappings value, Parameter value, expected SpannerDbType.
+            // Decimal mappings.
+            yield return new object[] { default, 3.14m, SpannerDbType.Float64 };
+            yield return new object[] { DecimalToFloat64, 3.14m, SpannerDbType.Float64 };
+            yield return new object[] { DecimalToNumeric, 3.14m, SpannerDbType.Numeric };
+            yield return new object[] { DecimalToPgNumeric, 3.14m, SpannerDbType.PgNumeric };
+
+            // DateTime mappings.
+            yield return new object[] { default, new DateTime(2022, 05, 26, 12, 12, 12), SpannerDbType.Timestamp };
+            yield return new object[] { DateTimeToTimestamp, new DateTime(2022, 05, 26, 12, 12, 12), SpannerDbType.Timestamp };
+            yield return new object[] { DateTimeToDate, new DateTime(2022, 05, 26, 12, 12, 12), SpannerDbType.Date };
+        }
+
+        [Theory]
+        [MemberData(nameof(ConfiguredSpannerDbTypes))]
+        public void ExecuteStreamingSql_ParameterDefaultsToConfiguredType(string clrToSpannerTypeMapping, object value, SpannerDbType spannerDbType)
+        {
+            var builder = new SpannerConnectionStringBuilder();
+            if (!string.IsNullOrWhiteSpace(clrToSpannerTypeMapping))
+            {
+                builder.ClrToSpannerTypeDefaultMappings = clrToSpannerTypeMapping;
+            }
+
+            var parameter = new SpannerParameter
+            {
+                ParameterName = "p",
+                Value = value
+            };
+            var actualValue = RunExecuteStreamingSqlWithParameter(builder, parameter).Fields["p"];
+            var expectedValue = spannerDbType.ToProtobufValue(parameter.Value);
+            Assert.Equal(expectedValue, actualValue);
+        }
+
+        public static IEnumerable<object[]> ConfiguredClrTypes()
+        {
+            // Format : SpannerToClrTypeDefaultMappings value, value, SpannerDbType, expected CLR type.
+            // Float64 mappings.
+            yield return new object[] { default, 3.14d, SpannerDbType.Float64, typeof(double) };
+            yield return new object[] { Float64ToSingle, 3.14d, SpannerDbType.Float64, typeof(float) };
+            yield return new object[] { Float64ToDouble, 3.14d, SpannerDbType.Float64, typeof(double) };
+            yield return new object[] { Float64ToDecimal, 3.14d, SpannerDbType.Float64, typeof(decimal) };
+            yield return new object[] { Float64ToSpannerNumeric, 3.14d, SpannerDbType.Float64, typeof(SpannerNumeric) };
+            yield return new object[] { Float64ToPgNumeric, 3.14d, SpannerDbType.Float64, typeof(PgNumeric) };
+
+            //// Date mappings.
+            yield return new object[] { default, "2022-05-26", SpannerDbType.Date, typeof(DateTime) };
+            yield return new object[] { DateToDateTime, "2022-05-26", SpannerDbType.Date, typeof(DateTime) };
+            yield return new object[] { DateToSpannerDate, "2022-05-26", SpannerDbType.Date, typeof(SpannerDate) };
+        }
+
+        [Theory]
+        [MemberData(nameof(ConfiguredClrTypes))]
+        public void ExecuteReadRequest_ColumnDefaultsToConfiguredClrType(string spannerToClrTypeMapping, object dbValue, SpannerDbType dbType, System.Type clrType)
+        {
+            var builder = new SpannerConnectionStringBuilder();
+            if (!string.IsNullOrWhiteSpace(spannerToClrTypeMapping))
+            {
+                builder.SpannerToClrTypeDefaultMappings = spannerToClrTypeMapping;
+            }
+            var options = SpannerConversionOptions.ForConnectionStringBuilder(builder);
+            Value protobufValue = RunReadRequest(builder, dbType, dbValue).Values[0];
+            // This is a hack, but simpler to do what DataReader does behind the scenes.
+            var clrValue = dbType.ConvertToClrType<object>(protobufValue, options);
+            // Check that CLR type of value is as expected.
+            Assert.Equal(clrType, clrValue.GetType());
+        }
+
+        private Struct RunExecuteStreamingSqlWithParameter(SpannerConnectionStringBuilder builder, SpannerParameter parameter)
+        {
+            var request = new ExecuteSqlRequest();
+            Mock<SpannerClient> spannerClientMock = SpannerClientHelpers
+                .CreateMockClient(Logger.DefaultLogger);
+            spannerClientMock
+                .SetupBatchCreateSessionsAsync()
+                .SetupExecuteStreamingSql(request);
+
+            var connection = BuildSpannerConnection(spannerClientMock, builder);
+
+            var command = connection.CreateSelectCommand("SELECT * FROM FOO");
+            command.Parameters.Add(parameter);
+            using var reader = command.ExecuteReader();
+            Assert.True(reader.HasRows);
+            spannerClientMock.Verify(client => client.ExecuteStreamingSql(
+                It.IsAny<ExecuteSqlRequest>(),
+                It.IsAny<CallSettings>()), Times.Once());
+            return request.Params;
+        }
+
+        // TODO: Ensure that builder is correctly used.
+        private ListValue RunReadRequest(SpannerConnectionStringBuilder builder, SpannerDbType dbType, object value)
+        {
+            var request = new ReadRequest();
+            Mock<SpannerClient> spannerClientMock = SpannerClientHelpers
+                .CreateMockClient(Logger.DefaultLogger);
+            spannerClientMock
+                .SetupBatchCreateSessionsAsync()
+                .SetupStreamingRead(request);
+
+            var connection = BuildSpannerConnection(spannerClientMock, builder);
+
+            var command = connection.CreateReadCommand(
+                 "Foo", ReadOptions.FromColumns(new List<string> { "Col1" }),
+                 KeySet.FromParameters(new SpannerParameterCollection
+                     {
+                        { "p", dbType, value } 
+                     }));
+            using var reader = command.ExecuteReader();
+            Assert.True(reader.HasRows);
+            // Handle the cases of Float64 and Date for now.
+            Value val = dbType == SpannerDbType.Float64 ? Value.ForNumber((double)value) : Value.ForString((string)value);
+            spannerClientMock.Verify(client => client.StreamingRead(It.Is<ReadRequest>(request => request.KeySet.Equals(new V1.KeySet
+            {
+                Keys = { new ListValue { Values = { val } } }
+            })),
+            It.IsAny<CallSettings>()));
+            return request.KeySet.Keys[0];
+        }
+
+        internal static SpannerConnection BuildSpannerConnection(Mock<SpannerClient> spannerClientMock) =>
+            BuildSpannerConnection(spannerClientMock, new SpannerConnectionStringBuilder());
+
+        /// <summary>
+        /// Builds a spanner connection based on the given mock and connection string builder.
+        /// The connection string builder will be mutated during the course of this method.
+        /// </summary>
+        internal static SpannerConnection BuildSpannerConnection(Mock<SpannerClient> spannerClientMock, SpannerConnectionStringBuilder builder)
         {
             var spannerClient = spannerClientMock.Object;
             var sessionPoolOptions = new SessionPoolOptions
@@ -1000,13 +1128,9 @@ namespace Google.Cloud.Spanner.Data.Tests
             var sessionPoolManager = new SessionPoolManager(sessionPoolOptions, spannerClient.Settings, spannerClient.Settings.Logger, (_o, _s, _l) => Task.FromResult(spannerClient));
             sessionPoolManager.SpannerSettings.Scheduler = spannerClient.Settings.Scheduler;
             sessionPoolManager.SpannerSettings.Clock = spannerClient.Settings.Clock;
-
-            SpannerConnectionStringBuilder builder = new SpannerConnectionStringBuilder
-            {
-                DataSource = DatabaseName.Format(SpannerClientHelpers.ProjectId, SpannerClientHelpers.Instance, SpannerClientHelpers.Database),
-                SessionPoolManager = sessionPoolManager
-            };
-
+            
+            builder.DataSource = DatabaseName.Format(SpannerClientHelpers.ProjectId, SpannerClientHelpers.Instance, SpannerClientHelpers.Database);
+            builder.SessionPoolManager = sessionPoolManager;
             return new SpannerConnection(builder);
         }
 
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
@@ -215,5 +215,67 @@ namespace Google.Cloud.Spanner.Data.Tests
             builder = builder.WithDatabase(null);
             Assert.Null(builder.SpannerDatabase);
         }
+
+        [Fact]
+        public void ClrToSpannerTypeDefaultMappings()
+        {
+            // Defaults - When property is not set.
+            var builder = new SpannerConnectionStringBuilder();
+            Assert.True(string.IsNullOrEmpty(builder.ClrToSpannerTypeDefaultMappings)); 
+            
+            // Explicitly set valid type mapping.
+            builder = new SpannerConnectionStringBuilder("ClrToSpannerTypeDefaultMappings=DecimalToNumeric,DateTimeToDate");
+            // DbConnectionStringBuilder lower-cases keywords, annoyingly.
+            Assert.Equal("clrtospannertypedefaultmappings=DecimalToNumeric,DateTimeToDate", builder.ToString());
+
+            // Explicitly set valid type mappings. Latest should be reflected.
+            // Also check that strings  are case insensitive.
+            builder.ClrToSpannerTypeDefaultMappings = "decimalTofloat64,dateTimetotimestamp";
+            // DbConnectionStringBuilder lower-cases keywords, annoyingly.
+            Assert.Equal("clrtospannertypedefaultmappings=decimalTofloat64,dateTimetotimestamp", builder.ToString());
+        }
+
+        [Fact]
+        public void SpannerToClrTypeDefaultMappings()
+        {
+            // Defaults - When property is not set.
+            var builder = new SpannerConnectionStringBuilder();
+            Assert.True(string.IsNullOrEmpty(builder.SpannerToClrTypeDefaultMappings));
+
+            // Explicitly set valid type mappings.
+            builder = new SpannerConnectionStringBuilder("SpannerToClrTypeDefaultMappings=Float64ToSpannerNumeric,DateToDateTime");
+            // DbConnectionStringBuilder lower-cases keywords, annoyingly.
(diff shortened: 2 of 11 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -989,7 +989,128 @@ public async Task CanExecuteReadPartitionedReadCommand()
...
+            {
+                builder.SpannerToClrTypeDefaultMappings = spannerToClrTypeMapping;
+            }
+            var options = SpannerConversionOptions.ForConnectionStringBuilder(builder);
+            Value protobufValue = RunReadRequest(dbType, dbValue).Values[0];
+            // This is a hack, but simpler to do what DataReader does behind the scenes.
+            var clrValue = dbType.ConvertToClrType<object>(protobufValue, options);
+            // Check that CLR type of value is as expected.
+            Assert.Equal(clrType, clrValue.GetType());
+            // Check the values.
```
jskeet, line 1059: This only checks the value for double. To be honest, I think it's probably simpler to just remove this bit and assume that if we've got the right type, the right conversion has taken place - that's tested elsewhere already.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -989,7 +989,128 @@ public async Task CanExecuteReadPartitionedReadCommand()
...
+
+            var command = connection.CreateReadCommand(
+                 "Foo", ReadOptions.FromColumns(new List<string> { "Col1" }),
+                 KeySet.FromParameters(new SpannerParameterCollection
+                     {
+                        { "p", dbType, value } 
+                     }));
+            using var reader = command.ExecuteReader();
+            Assert.True(reader.HasRows);
+            // Handle the case of Float64 for now.
```
jskeet: I think it would be simpler to just return the request instead of doing any validation here. Let the caller do the validation.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
```diff
@@ -215,5 +215,82 @@ public void WithDatabase()
...
+            builder.SpannerToClrTypeDefaultMappings = "Float64ToDecimal";
+            // DbConnectionStringBuilder lower-cases keywords, annoyingly.
+            Assert.Equal("spannertoclrtypedefaultmappings=Float64ToDecimal", builder.ToString());
+            // Check that internal mapping dictionary is populated with user provided configuration.
+            Assert.True(builder.SpannerToClrTypeMappings.Count == 1);
+            Assert.Equal(typeof(decimal), builder.SpannerToClrTypeMappings[SpannerDbType.Float64]);
+        }
+
+        [Theory]        
+        [InlineData("DecimalToFloat64,DecimalToNumeric;")] // Multiple mappings for a type.
```
jskeet: It's not clear to me from this test whether *just* additional whitespace is a problem, e.g. "DecimalToFloat64    ". Every test that includes whitespace is *also* broken in some other way.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -140,6 +181,160 @@ private string ValidatedDataSource(string dataSource)
...
+            }
+
+            // Check multiples for each type.
+            // Currently, we have Decimal to check.
+            if (mappings.Count(j => j.StartsWith("Decimal", StringComparison.OrdinalIgnoreCase)) > 1)
+            {
+                throw new ArgumentException($"'{input}' is not a valid value as multiple mappings from Decimal to SpannerDbType are provided for ${nameof(ClrToSpannerTypeDefaultMappings)}");
+            }
+
+            // If we reach here all is well. Return the input as validation passed.
```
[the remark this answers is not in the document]
jskeet: You could argue that that's putting "the connection string representation of the options" into SpannerConversionOptions, but I think that's *probably* okay.

jskeet (commented):
I've only reviewed the first commit, because I'm suggesting a significant change - basically moving quite a lot of the code out of SpannerConnectionStringBuilder to SpannerConverisonOptions. (I've also suggested a change in option name, which I should have caught earlier - sorry about that.) I haven't actually reviewed the changes later than SpannerConnectionStringBuilder as they'd probably change if we moved things anyway.

I'm not suggesting a change to the public API, so we *could* potentially merge this (after the other comments are addressed) and then refactor. See what you think.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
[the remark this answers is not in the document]
Rishabh-V: I was thinking, if I can do the same for "UseClrDefaultForNull" as that is the only other property in SpannerConnectionStringBuilder that impacts the SpannerConversionOptions. And post that I could just return the builder.ConversionOptions in the factory methods ForConnection and ForConnectionStringBuilder.  Please let me know if this sounds fine?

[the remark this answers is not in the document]
jskeet: Yes, that sounds fine to me - I do like the idea that everything is in one place and that obtaining the conversion options is trivial.

Rishabh-V:
> I've only reviewed the first commit, because I'm suggesting a significant change - basically moving quite a lot of the code out of SpannerConnectionStringBuilder to SpannerConverisonOptions. (I've also suggested a change in option name, which I should have caught earlier - sorry about that.) I haven't actually reviewed the changes later than SpannerConnectionStringBuilder as they'd probably change if we moved things anyway.
> 
> I'm not suggesting a change to the public API, so we _could_ potentially merge this (after the other comments are addressed) and then refactor. See what you think.

Hi Jon. I have tried to implement the suggested changes. Currently, it has both decimal and DateTime changes (unfortunately :( as I was working on the same branch).  Just so that the core implementation gets reviewed and I get the feedback, I have raised the PR containing both. 

Once the implementation gets ok, I will raise 2 different PRs for merging, one for decimal and Float64 and one for DateTime and Date.  Hope it is fine.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -989,7 +989,128 @@ public async Task CanExecuteReadPartitionedReadCommand()
...
+            command.Parameters.Add(parameter);
+            using var reader = command.ExecuteReader();
+            Assert.True(reader.HasRows);
+            spannerClientMock.Verify(client => client.ExecuteStreamingSql(
+                It.IsAny<ExecuteSqlRequest>(),
+                It.IsAny<CallSettings>()), Times.Once());
+            return request.Params;
+        }
+
+        private ListValue RunReadRequest(SpannerDbType dbType, object value)
```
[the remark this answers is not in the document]
Rishabh-V: Right Jon. I have used it in the caller as reader will use it. Please let me know if this needs to be changed.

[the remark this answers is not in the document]
jskeet: Ah, so this isn't for testing the format of the parameter being sent, but the value being received? That's a little confusing - I'd expected this to be testing "is the request created for a given parameter correct".

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -455,6 +568,21 @@ private string GetValueOrDefault(string key, string defaultValue = "")
                 {
                     value = ValidatedDataSource((string)value);
                 }
+                if (string.Equals(keyword, ClrToSpannerTypeDefaultMappingsKeyword, StringComparison.OrdinalIgnoreCase))
```
jskeet, line 571: All of these can be `else if`, I think. We could also probably change it just to set the releated property to `(string) value` (and then return), as a way of avoiding duplication.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -46,6 +108,196 @@ private SpannerConversionOptions(bool useDBNull)
...
+            {
+                throw new ArgumentException($"'{input}' is not a valid value as multiple mappings from Float64 to CLR type are provided for ${nameof(SpannerConnectionStringBuilder.SpannerToClrTypeDefaultMappings)}");
+            }
+
+            if (mappings.Count(j => j.StartsWith("Date", StringComparison.OrdinalIgnoreCase)) > 1)
+            {
+                throw new ArgumentException($"'{input}' is not a valid value as multiple mappings from Date to CLR type are provided for ${nameof(SpannerConnectionStringBuilder.SpannerToClrTypeDefaultMappings)}");
+            }
+
+            // We don't want to modify the current instance, so create a new dictionary.
```
jskeet, line 194: I suspect we can do this without a dictionary to be honest... but now the logic is in the "right" place, I'm happier to refactor later.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
```diff
@@ -603,23 +604,30 @@ private object ConvertToClrTypeImpl(Value wireValue, System.Type targetClrType,
...
                             $"Invalid Type conversion from {wireValue.KindCase} to {targetClrType.FullName}");
                 }
             }
             if (targetClrType == typeof(PgNumeric))
             {
-                if (TypeCode != TypeCode.Numeric || TypeAnnotationCode != TypeAnnotationCode.PgNumeric)
+                bool isPgNumeric = TypeAnnotationCode == TypeAnnotationCode.PgNumeric;
+                bool isFloat64 = TypeCode == TypeCode.Float64;
+
+                if (!isPgNumeric && !isFloat64)
```
jskeet, line 619: Should this be `||` instead of `&&`? (At the moment, we wouldn't fail if it's a string with a PgNumeric type annotation code for example. That might never be feasible, but it feels like it's logically `||`.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
```diff
@@ -603,23 +604,30 @@ private object ConvertToClrTypeImpl(Value wireValue, System.Type targetClrType,
...
+                    throw new ArgumentException($"{targetClrType.FullName} can only be used for numeric results and only when working with a PostgreSQL dialect.");
                 }
                 switch (wireValue.KindCase)
                 {
                     case Value.KindOneofCase.NullValue:
                         return null;
                     case Value.KindOneofCase.StringValue:
                         return V1.PgNumeric.Parse(wireValue.StringValue);
+                    case Value.KindOneofCase.NumberValue:
+                        return V1.PgNumeric.FromDecimal(Convert.ToDecimal(wireValue.NumberValue));
```
jskeet, line 630: Ick - I can see why we've got this, but it's unfortunate. Maybe we need to add a `FromDouble` as well at some point, if only internally.

jskeet (commented):
I don't have time for a full review before Monday, but the separation of concerns definitely looks a lot cleaner now, thanks! Will review properly on Monday.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
Rishabh-V: We need `&&` here as PgNumeric value by default comes as string and while reading a Float64 value as PgNumeric, we need to read the NumberValue :(

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -46,6 +116,159 @@ private SpannerConversionOptions(bool useDBNull)
...
+            // Empty string or null is valid to unset the values.
+            if (string.IsNullOrEmpty(input))
+            {
+                // Set defaults.
+                options.SetSpannerToClrTypeDefaults();
+                return;
+            }
+
+            // Get all type mappings.
+            var mappings = input.Split(new char[] { ',', ';' }, StringSplitOptions.RemoveEmptyEntries);
```
jskeet, line 161: I suspect we don't need `;` in the list here. (I don't *think* it should come through from the connection string builder.)
That would allow us to use `input.Split(',',StringSplitOptions.RemoveEmptyEntries)'`, avoiding allocating the extra array.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -46,6 +116,159 @@ private SpannerConversionOptions(bool useDBNull)
...
+
+            // Validate that mappings are valid.
+            if (mappings == null || !mappings.Any())
```
jskeet, line 164: `mappings` will never be null here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -46,6 +116,159 @@ private SpannerConversionOptions(bool useDBNull)
...
+            // Validate that mapping matches one of the pre-defined values.
+            var invalidMappings = mappings.Except(AllowedSpannerToClrTypeMappings, StringComparer.OrdinalIgnoreCase);
+            if (invalidMappings.Any())
+            {
+                throw new ArgumentException($"The following mappings are invalid:'{string.Join(",", invalidMappings)}' for ${nameof(SpannerConnectionStringBuilder.SpannerToClrTypeDefaultMappings)}");
+            }
+
+            // Check multiples for each type.
+            // Currently, we have Float64 and Date to check.
+            if (mappings.Count(j => j.StartsWith("Float64", StringComparison.OrdinalIgnoreCase)) > 1)
```
jskeet, line 178: I think we might want to have a more "data-oriented" approach to this, grouping options by affected type - but we don't need to do that immediately. (I'm happy to send a refactoring PR.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -46,6 +116,159 @@ private SpannerConversionOptions(bool useDBNull)
...
+            if (mappings.Count(j => j.StartsWith("Date", StringComparison.OrdinalIgnoreCase)) > 1)
+            {
+                throw new ArgumentException($"'{input}' is not a valid value as multiple mappings from Date to CLR type are provided for ${nameof(SpannerConnectionStringBuilder.SpannerToClrTypeDefaultMappings)}");
+            }
+
+            // If we reach here, all is well.
+            foreach (var mapping in mappings)
+            {
+                // Discard pattern, primarily to simplify code using switch expressions.
+                _ = mapping switch
```
jskeet, line 192: This will be case-sensitive, having done a case-insensitive comparison earlier.
We could either make the comparisons case-sensitive, or lower-case the input and expected values, or use if statements.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -46,6 +116,159 @@ private SpannerConversionOptions(bool useDBNull)
...
+            DateTimeToConfiguredSpannerType = SpannerDbType.Timestamp;
+        }
+
+        internal void SetSpannerToClrTypeDefaults()
+        {
+            DateToConfiguredClrType = typeof(DateTime);
+            Float64ToConfiguredClrType = typeof(double);
+        }
+
+        private void ValidateAndParseClrToSpannerTypeMappings(string input, SpannerConversionOptions options)
```
jskeet, line 219: Ideally, I'd like to reduce the duplication between this and the other method - although it's entirely possible that abstracting the infrastructure will be larger than the current code.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
jskeet: I'm afraid I don't understand - and in particular, this *is* a change to behavior.
Was there a bug before?

As a concrete example: we *won't* throw an exception here if the user has requested a `PgNumeric` conversion for a `TypeCode.Float64` value in a regular Spanner database... that seems wrong, given the error message. Surely we want to throw an exception if *either* we're not in Postgres *or* we don't have a numeric value.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -316,6 +317,10 @@ public static SpannerDbType FromClrType(System.Type type)
             {
                 return Bytes;
             }
+            if (type == typeof(SpannerDate))
```
jskeet, line 320: Was this just missing before? (It doesn't seem related to the custom conversion part.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
```diff
@@ -160,7 +161,25 @@ internal object GetValidatedValue()
         /// </summary>
         /// <param name="options">Options to configure the <c>SpannerDbType</c> for this <c>SpannerParameter</c>.</param>
         /// <returns>The configured <see cref="SpannerDbType"/>.</returns>
-        internal SpannerDbType GetConfiguredSpannerDbType(SpannerConversionOptions options) => SpannerDbType;
+        internal SpannerDbType GetConfiguredSpannerDbType(SpannerConversionOptions options)
+        {
+            // Only if SpannerDbType of parameter is not explicitly provided by user.
+            if (_spannerDbType == null)
+            {
+                if (Value is decimal && options != null)
```
jskeet, line 169: Will options ever be null here? How hard would it be to ensure it's never null?

jskeet (commented):
Nearly there, I think. A couple of actual behavioral changes I'd like to see (case handling in SpannerConversionOptions, and the argument exception around PgNumeric) but the rest are potential future changes.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
Rishabh-V: Yes Jon. It should have gone with SpannerDate changes.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameter.cs
Rishabh-V: I will make the change to ensure options are not null.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
jskeet: Okay, it sounds like what you *really* want is something saying "is the type code string or float64" - although I may have misunderstood the relationship between the type code and the Value kind here.

Fundamentally this code needs to be much clearer about what's valid to convert into PgNumeric...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
jskeet: I would probably do something like having a `Dictionary<string, (SpannerDbType, Type)>` for each option to "source and target types". Then you can find out any options that conflict by whether there multiple options with the same source. (That can also be a case-insensitive dictionary.) I can picture the code, but it's somewhat fiddly to actually describe. But basically we'd end up with the two dictionaries (one in each direction) having almost all the type-specific aspects, so adding a new conversion option should be a small change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -140,6 +145,114 @@ private string ValidatedDataSource(string dataSource)
             return dataSource;
         }
 
+        /// <summary>
+        /// Option to configure the default CLR type to SpannerDbType mapping. This option comes into picture
+        /// only if <see cref="SpannerDbType"/> and <see cref="System.Data.DbType"/> of the <see cref="SpannerParameter"/>
+        /// are not explicitly provided. Currently only <see cref="decimal"/> and <see cref="DateTime"/> CLR types are supported.
+        /// </summary>
+        /// <remarks>
```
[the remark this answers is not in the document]
Rishabh-V: Done.

Rishabh-V (commented):
Hi Jon. Have made most of the requested changes. Didn't get the string.Split overload in .netstandard2.0. Have now made case insensitive comparison and PgNumeric and Numeric handling is updated with options. Please take a look.
 
Have added TODO to the following items. 
(Will raise a new PR for them)
1) Validation methods code reuse.
2) Use Dictionary for validating multiple mappings.
3) RunReadStream should use builder correctly in Tests.
4) Docs should be updated in connection string. 

Have squashed the commits to make it easier to review. Once this is good, I can raise two different PRs. One for decimal and Float64 and one for DateTime and Date for merging. Thank you.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
```diff
@@ -603,23 +608,36 @@ private object ConvertToClrTypeImpl(Value wireValue, System.Type targetClrType,
...
+                bool isPgNumeric = TypeAnnotationCode == TypeAnnotationCode.PgNumeric;
+                bool isFloat64 = TypeCode == TypeCode.Float64;
+                // Check if Float64 is being read as PgNumeric due to user provided type mapping. 
+                bool configuredFloat64AsPgNumeric = isFloat64 && options.Float64ToConfiguredClrType == typeof(PgNumeric); 
+                // For PgNumeric, the value is contained in Value.StringValue.
+                // If the user configures that Float64 should be mapped to PgNumeric CLR type in connection string,
+                // we will need to read Float64 as PgNumeric. The Float64 value is populated in NumberValue of Value.
+                // If the TypeCode is Float64 due to user configured mapping or TypeAnnotationCode is PgNumeric, it is fine.
+                // In all other cases, throw an exception.
+                if (!isPgNumeric && !configuredFloat64AsPgNumeric)
```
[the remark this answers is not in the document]
jskeet: (I think it's slightly differently wrong now to how it was wrong before... it shouldn't matter whether the way that we've ended up wanting to convert a value to PgNumeric is via connection string options or a direct request.)

jskeet (approved):
I haven't rechecked the tests, although I've looked through the new SpannerConversionOptions tests. As noted, I'm still definitely uncomfortable about the validation condition - we need to get that sorted before we next release, but I'd like to do it in a standalone fashion.

Rishabh-V:
> I haven't rechecked the tests, although I've looked through the new SpannerConversionOptions tests. As noted, I'm still definitely uncomfortable about the validation condition - we need to get that sorted before we next release, but I'd like to do it in a standalone fashion.

Sure Jon.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.ValueConversion.cs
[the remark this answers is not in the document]
Rishabh-V: Sure Jon.

jskeet:
Can we now close #5837, #5839, #8463 and #8472? (Possibly wait until we've got docs before closing the issues, actually?)

PR #8535, opened 2022-05-30
chore: Address changes in Spanner ADO API due to netstandard2.1 async methods

jskeet:
This is slightly more invasive in code terms by introducing a new
base class for SpannerTransaction, but won't break existing uses of
SpannerTransaction.CommitAsync.


amanda-tarafa (approved):
LGTM.

I like this one better. Mostly so that we don't have to introduce the new method to keep the old functionality.

PR #8555, opened 2022-06-06
feat: Prototype PR for adding connection string options support to Keys

### LINKED ISSUE #8541: Add support for ConnectionString options in ReadCommand to configure the type mappings.
This was caught and pointed by @jskeet while reviewing the code for making the type conversions configurable via connection string.

After the introduction of ConnectionString options to configure the CLR to Spanner and Spanner to CLR types, user should be able to configure the type mappings for Decimal and DateTime CLR types. However, if the user uses the following code:

```
using var cmd = connection.CreateReadCommand(
     "MyTable",
      ReadOptions.FromColumns("Col1", "Col2"),
      KeySet.FromParameters(new SpannerParameterCollection
      {
           new SpannerParameter { ParameterName = "Col1", Value = 3.14m}, //decimal
           new SpannerParameter { ParameterName = "Col2", Value = new DateTime(2022, 05, 31) }  // DateTime
       }));
using var reader = await cmd.ExecuteReaderAsync();
```

the connection string options are not honored.

Rishabh-V:
Draft Prototype PR for adding support of connection string options to Keys.

Fixed existing tests with backward compatible changes in ToString implementation.

Will write new tests once the main code is ok. 

There are a couple of TODOs to implement IEqualityComparer<T> for KeyRanges and if implementation of IEqualityComparer<T> should be in a new file.

Fixes #8541

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
@@ -38,24 +38,32 @@ namespace Google.Cloud.Spanner.Data.Tests
                     { "", SpannerDbType.String, "test" },
                     { "", SpannerDbType.Timestamp, new DateTime(2021, 9, 10, 9, 37, 10, DateTimeKind.Utc) }
                 });
-            var index = 0;
-            Assert.True(key.KeyParts.Values[index++].BoolValue);
-            Assert.Equal(Convert.ToBase64String(new byte[] {1,2,3}), key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("2021-09-10", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal(3.14, key.KeyParts.Values[index++].NumberValue);
-            Assert.Equal("1", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("{\"key\": \"value\"}", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("10.1", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("20.1", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("test", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("2021-09-10T09:37:10Z", key.KeyParts.Values[index++].StringValue);
+
+            var actual = key.ToProtobuf(SpannerConversionOptions.Default);
+            var expected = new ListValue
+            {
+                Values =
+                {
+                    Value.ForBool(true),
+                    Value.ForString(Convert.ToBase64String(new byte[] {1,2,3})),
+                    Value.ForString("2021-09-10"),
+                    Value.ForNumber(3.14),
+                    Value.ForString("1"),
+                    Value.ForString("{\"key\": \"value\"}"),
+                    Value.ForString("10.1"),
+                    Value.ForString("20.1"),
+                    Value.ForString("test"),
+                    Value.ForString("2021-09-10T09:37:10Z")
+                }
+            };
+            Assert.Equal(expected, actual);
         }
 
         [Fact]
         public void CreateKeyFromEmptyParameters()
         {
             var key = new Key(new SpannerParameterCollection());
-            Assert.Empty(key.KeyParts.Values);
+            Assert.Empty(key.ToProtobuf(SpannerConversionOptions.Default).Values);
         }
 
         [Fact]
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -101,10 +98,39 @@ private KeySet(bool allKeys, IEnumerable<Key> keys, IEnumerable<KeyRange> ranges
 
         /// <inheritdoc/>
         public override bool Equals(object other)
```
amanda-tarafa, line 100: Given that these are now mutable, do we really need to implement Equality?
I would rather not I think, @jskeet thoughts?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -101,10 +98,39 @@ private KeySet(bool allKeys, IEnumerable<Key> keys, IEnumerable<KeyRange> ranges
...
+                // If one is null and other is not, they can't be equal.
+                if ((Ranges != null && keySet.Ranges == null) || (Ranges == null && keySet.Ranges != null))
+                {
+                    return false;
+                }
+
+                // Now, we will have either both as null or both as non-null.
+                // TODO: Implement IEqualityComparer<T> for KeyRange in actual implementation.
+                return AllKeys == keySet.AllKeys &&
+                   ((Keys == null && keySet.Keys == null) || Keys.SequenceEqual(keySet.Keys)) &&
```
amanda-tarafa, line 125: Also, I think `SequenceEquals` requieres elemts to be on the same position and compares element references only. Not sure that's what we would want now.

Again, I think I would prefer not to have equality at all, unless we need it for something?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -227,44 +252,46 @@ public sealed class KeyRange
         }
 
         /// <inheritdoc/>
-        public override bool Equals(object other)
-            => other is KeyRange range && Protobuf.Equals(range.Protobuf);
-
-        /// <inheritdoc/>
-        public override int GetHashCode() => Protobuf.GetHashCode();
+        public override bool Equals(object other) => other is KeyRange range && BeginAt.Equals(range.BeginAt) &&
```
amanda-tarafa, line 255: Same comment here, I don't think we need this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -227,44 +252,46 @@ public sealed class KeyRange
...
-            builder.Append(End.KeyParts);
+            builder.Append(End.ToString());
             builder.Append(EndInclusive ? ']' : ')');
 
             return builder.ToString();
         }
 
-        private V1.KeyRange ToProtobuf()
+        /// <inheritdoc/>
+        public override int GetHashCode() => GaxEqualityHelpers.CombineHashCodes(BeginAt.GetHashCode(), BeginInclusive.GetHashCode(), End.GetHashCode(), EndInclusive.GetHashCode());
```
amanda-tarafa, line 274: If we keep equality move GetHashCode write after Equals.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -179,8 +205,7 @@ public sealed class KeyRange
         /// </summary>
         public bool EndInclusive { get; }
 
-        private Lazy<V1.KeyRange> LazyProtobuf => new Lazy<V1.KeyRange>(ToProtobuf);
-        internal V1.KeyRange Protobuf => LazyProtobuf.Value;
+        internal V1.KeyRange GetProtobuf(SpannerConversionOptions options) => ToProtobuf(options);
```
amanda-tarafa, line 208: We don't need GetProtobuf and ToProtobuf.

Rename ToProtobuf => GetProtobuf.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +302,72 @@ private V1.KeyRange ToProtobuf()
...
+        internal SpannerParameterCollection Parameters { get; }
 
         /// <summary>
         /// Creates a key from the values in the parameter collection.
         /// </summary>
         /// <param name="keyParts">The values that define the key</param>
         internal Key(SpannerParameterCollection keyParts)
         {
-            KeyParts = ToListValue(keyParts);
+            Parameters = keyParts.Clone();
```
amanda-tarafa, line 313: Cloning here does not guarantee inmutability, the list of parameters is itself created new but the parameters are shallowed copied.

I prefer if we don't stand in middle ground, so let's not clone anything here.
@jskeet thoughts?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +302,72 @@ private V1.KeyRange ToProtobuf()
...
         }
 
+        // Note: We could have just used IEnumerable<SpannerParameter>, but this constructor overload accepting params object[]
```
amanda-tarafa, line 316: I don't understand this note. Where could you use IEnumerable...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +302,72 @@ private V1.KeyRange ToProtobuf()
...
 
+        /// <summary>
+        /// Gets the values of this Key.
+        /// </summary>
+        /// <param name="options">The options for conversion between protobuf and CLR types.</param>
+        /// <returns>The values of this Key.</returns>
+        internal ListValue GetKeyParts(SpannerConversionOptions options) => ToListValue(Parameters, options);
+
         /// <inheritdoc/>
         public override bool Equals(object other)
```
amanda-tarafa, line 335: Same comment as before, I don't think we need equality.

And if we don't need it here, we don't need it in Parameters either.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +302,72 @@ private V1.KeyRange ToProtobuf()
...
-        public override int GetHashCode() => KeyParts.GetHashCode();
+        public override int GetHashCode() => Parameters.GetHashCode();
 
+        // This is to make the implementation backward compatible.
         /// <inheritdoc/>
-        public override string ToString() => KeyParts.ToString();
+        public override string ToString() => ToListValue(Parameters, SpannerConversionOptions.Default).ToString();
 
-        private static ListValue ToListValue(SpannerParameterCollection parameters)
+        private static ListValue ToListValue(SpannerParameterCollection parameters, SpannerConversionOptions options)
```
amanda-tarafa, line 363:
```suggestion
        private ListValue ToProtobuf(SpannerConversionOptions options)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerParameterCollection.cs
```diff
@@ -282,5 +282,44 @@ private void FillSpannerInternalTypes(MapField<string, V1.Type> typeDictionary,
         /// <returns>A cloned copy of this instance.</returns>
         public SpannerParameterCollection Clone() =>
             new SpannerParameterCollection(_innerList.Select(p => (SpannerParameter) p.Clone()));
+
+        /// <inheritdoc />
+        public override int GetHashCode() => _innerList.GetHashCode();
```
amanda-tarafa, line 287: Hopefully we don't need any of this.

amanda-tarafa (changes requested):
A few change requests.
Also, @jskeet I've flagged you in a couple of my comments for thoughts.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
jskeet: We should probably find out why they implemented equality in the first place - probably just for convenience, knowing it was safe as they were immutable and based on protos.

If we want to make a breaking change by removing the equality, now is the time to do so.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
jskeet: SequenceEquals uses the specified-or-default equality comparer - but yes, it's ordered, which isn't great for a set. If we can get away with removing equality entirely, life becomes much simpler.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +302,72 @@ private V1.KeyRange ToProtobuf()
...
+            }
+
+            return false;
+        }
 
         /// <inheritdoc/>
-        public override int GetHashCode() => KeyParts.GetHashCode();
+        public override int GetHashCode() => Parameters.GetHashCode();
 
+        // This is to make the implementation backward compatible.
```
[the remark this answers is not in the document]
jskeet: I agree that we don't have to make it backwardly-compatible. I do think we should have *some* string representation though. We might want to take the opportunity to document that it's only for diagnostic purposes and may change over time...

amanda-tarafa:
@olavloite If you could take a look at this PR:

- @Rishabh-V is reimplementing the key related classes for the read operations so that they accept type conversion options.
- The easiest way is to lose inmutability on the key related classes.
- Whe you implemented these, you add equality to all key related classes and that was fine since they were inmutable.
- Is equality needed or just a nice to have/convenience. It's definetely not needed in Google.Cloud.Spanner.Data... but maybe in the ORM implementations. Although I wouldn't think those use the read operations as those are Spanner specific.

I've requested your review, let us know what you think. Thanks!

olavloite:
> @olavloite If you could take a look at this PR:
> 
> * @Rishabh-V is reimplementing the key related classes for the read operations so that they accept type conversion options.
> * The easiest way is to lose inmutability on the key related classes.
> * Whe you implemented these, you add equality to all key related classes and that was fine since they were inmutable.
> * Is equality needed or just a nice to have/convenience. It's definetely not needed in Google.Cloud.Spanner.Data... but maybe in the ORM implementations. Although I wouldn't think those use the read operations as those are Spanner specific.
> 
> I've requested your review, let us know what you think. Thanks!

I don't remember specifically, but as far as I can tell, the equality operations were just nice-to-have. ORMs certainly do not use the Read API, so we don't have to worry about that.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
olavloite: As far as I can tell, it was just for convenience and not really needed. The Read API is not used in any of the ORM implementations that we have.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +302,72 @@ private V1.KeyRange ToProtobuf()
...
         /// <inheritdoc/>
-        public override string ToString() => KeyParts.ToString();
+        public override string ToString() => ToListValue(Parameters, SpannerConversionOptions.Default).ToString();
```
olavloite, line 361: This will be a little confusing as the string representation will always use the default conversion options. Maybe you should consider adding a `ToString(SpannerConversionOptions options)` method as well.

Another option could be to include the `SpannerConversionOptions` as a property of this class, so you don't have to add it as an argument to all the methods in the class.

Rishabh-V: Unfortunately `SpannerConversionOptions` is `internal` so we can't add a public overload of `ToString` that takes `SpannerConversionOptions` as a parameter.

I can add a property, but it will end up having `SpannerConversionOptions.Default` as its value or am I missing something?

Rishabh-V:
> > @olavloite If you could take a look at this PR:
> > 
> > * @Rishabh-V is reimplementing the key related classes for the read operations so that they accept type conversion options.
> > * The easiest way is to lose inmutability on the key related classes.
> > * Whe you implemented these, you add equality to all key related classes and that was fine since they were inmutable.
> > * Is equality needed or just a nice to have/convenience. It's definetely not needed in Google.Cloud.Spanner.Data... but maybe in the ORM implementations. Although I wouldn't think those use the read operations as those are Spanner specific.
> > 
> > I've requested your review, let us know what you think. Thanks!
> 
> I don't remember specifically, but as far as I can tell, the equality operations were just nice-to-have. ORMs certainly do not use the Read API, so we don't have to worry about that.

Thank you Knut!

Rishabh-V:
> A few change requests. Also, @jskeet I've flagged you in a couple of my comments for thoughts.

I have addressed all the review comments. Removed equality and hashcode and streamlined the methods. Updated tests for the changes. Please take a look and let me know if further changes are needed. Thank you.

amanda-tarafa:
Thanks I'll review in a little while.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +255,56 @@ private V1.KeyRange ToProtobuf()
...
 
-        /// <inheritdoc/>
-        public override bool Equals(object other)
-            => other is Key key && KeyParts.Equals(key.KeyParts);
+        /// <summary>
+        /// Gets the values of this Key.
+        /// </summary>
+        /// <param name="options">The options for conversion between protobuf and CLR types.</param>
+        /// <returns>The values of this Key.</returns>
+        internal ListValue GetKeyParts(SpannerConversionOptions options) => ToProtobuf(options);
```
amanda-tarafa, line 282: We don't need this method that does the same as T`oProtobuf`. I think the better name is `ToProtobuf`, or better still change the `ToProtobuf` in all 3 classes to `GetProtobuf`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +255,56 @@ private V1.KeyRange ToProtobuf()
...
 
-        /// <inheritdoc/>
-        public override int GetHashCode() => KeyParts.GetHashCode();
+        /// <summary>
+        /// Returns the string representation of this key using the default type conversion mappings.
+        /// The string representation is only for diagnostic purposes and the implementation may change over time.
+        /// </summary>
+        /// <returns>A string that represents the current key.</returns>
+        public override string ToString() => ToProtobuf(SpannerConversionOptions.Default).ToString();
```
amanda-tarafa, line 289: Hmmm, using Default here won't match the actual values sent to Spanner and I wonder if that would hinder diagnostics which is what this method will most likely be used for.

It was actually @jskeet who requested the ToString initially (https://github.com/googleapis/google-cloud-dotnet/pull/7157#discussion_r729612411) so we know it's not a requierement. I say we remove it. What do you think @jskeet .

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +255,56 @@ private V1.KeyRange ToProtobuf()
...
-        /// <inheritdoc/>
-        public override string ToString() => KeyParts.ToString();
+        /// <summary>
+        /// Returns the string representation of this key using the type conversion options specified in 
+        /// <see cref="SpannerConnectionStringBuilder"/>. The string representation is only for diagnostic purposes
+        /// and the implementation may change over time.
+        /// </summary>
+        /// <param name="builder">The <see cref="SpannerConnectionStringBuilder"/> instance.</param>
+        /// <returns>A string that represents the current key.</returns>
+        public string ToString(SpannerConnectionStringBuilder builder) => ToProtobuf(builder.ConversionOptions).ToString();
```
amanda-tarafa, line 298: This one we can keep for diagnostic purposes. Could we should pass ConversionOptions instead?

amanda-tarafa (changes requested):
A few more change requests. Mostly for cleanup and tests.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
```diff
@@ -107,14 +112,6 @@ public void CreateKeyRangeClosedClosed()
             var closedClosed = KeyRange.ClosedClosed(new Key("begin"), new Key("end"));
             Assert.True(closedClosed.BeginInclusive);
             Assert.True(closedClosed.EndInclusive);
-            Assert.Equal(new Key("begin"), closedClosed.BeginAt);
```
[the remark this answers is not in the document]
Rishabh-V: Right Amanda.  I had used `ToString()` earlier. But, given we are discussing its removal, comparing ProtobufValue via `GetProtobuf` is fine or you expect something else. Please let me know and I shall do so.

Rishabh-V (commented):
Thanks Amanda. Please help me with couple of clarifications and I can make all the changes in the next commit.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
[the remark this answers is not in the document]
amanda-tarafa: Yes, I thinkg that should be the right approach, obtaing the protobuf value and make sure it is what you are expecting.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -99,13 +96,6 @@ private KeySet(bool allKeys, IEnumerable<Key> keys, IEnumerable<KeyRange> ranges
...
 
-        /// <inheritdoc/>
-        public override bool Equals(object other)
-            => other is KeySet keySet && Protobuf.Equals(keySet.Protobuf);
-
-        /// <inheritdoc/>
-        public override int GetHashCode() => Protobuf.GetHashCode();
-
         /// <inheritdoc/>
         public override string ToString()
```
jskeet, line 100: Another ToString to potentially remove, and definitely add an overload for.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,49 +255,56 @@ private V1.KeyRange ToProtobuf()
...
     public sealed class Key
     {
-        /// <summary>
-        /// The values of this key.
-        /// </summary>
-        internal ListValue KeyParts { get; }
+        internal SpannerParameterCollection Parameters { get; }
 
         /// <summary>
         /// Creates a key from the values in the parameter collection.
```
jskeet, line 261: Possibly expand the documentation to note that a reference is stored to this collection - changes to the collection will be visible when the key is used.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
```diff
@@ -39,23 +44,23 @@ public void CreateKeyFromParameterCollection()
...
-            Assert.Equal(Convert.ToBase64String(new byte[] {1,2,3}), key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("2021-09-10", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal(3.14, key.KeyParts.Values[index++].NumberValue);
-            Assert.Equal("1", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("{\"key\": \"value\"}", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("10.1", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("20.1", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("test", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("2021-09-10T09:37:10Z", key.KeyParts.Values[index++].StringValue);
+            Assert.True(key.GetKeyParts(s_conversionOptions).Values[index++].BoolValue);
```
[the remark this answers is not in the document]
amanda-tarafa: I'm fine with either really, GetKeyParts is the one I really really don't like.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
amanda-tarafa: Yep, I still think removing is the better option right now, and adding the overload with the parameter to all three classes.

Rishabh-V: Thanks Jon and Amanda . I will remove the override of ToString() and add ToString(SpannerConnectionStringBuilder),  update the tests and raise the commit by tomorrow morning.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
```diff
@@ -21,6 +21,14 @@ namespace Google.Cloud.Spanner.Data.Tests
 {
     public class KeysTests
     {
+        // Options will now be needed in all the tests.
```
jskeet, line 24: Given that these are reasonably short expressions anyway, are they actually worth having as fields? I suspect the tests might be clearer explicitly specifying `SpannerConversionOptions.Default` instead of `s_conversionOptions`. (Aside from anything else, it prompts the question "should we add a test for non-default options?")

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
```diff
@@ -39,23 +47,24 @@ public void CreateKeyFromParameterCollection()
...
-            Assert.Equal(Convert.ToBase64String(new byte[] {1,2,3}), key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("2021-09-10", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal(3.14, key.KeyParts.Values[index++].NumberValue);
-            Assert.Equal("1", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("{\"key\": \"value\"}", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("10.1", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("20.1", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("test", key.KeyParts.Values[index++].StringValue);
-            Assert.Equal("2021-09-10T09:37:10Z", key.KeyParts.Values[index++].StringValue);
+            var protobuf = key.ToProtobuf(s_conversionOptions);
```
jskeet, line 50: I would suggest just creating the expected value instead, e.g.

```csharp
var expected = new ListValue
{
    Values =
    {
        Value.ForBool(true),
        Value.ForString(Convert.ToBase64String(new byte[] { 1, 2, 3 })),
        Value.ForString("2021-09-10"),
        ...
    }
}
Assert.Equal(expected, actual);
```

jskeet: (That can come later though.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
```diff
@@ -187,42 +205,126 @@ public void CreateKeySetFromKeys()
...
             Assert.Collection(ranges.Ranges,
-                range => Assert.Equal(KeyRange.ClosedOpen(new Key("begin"), new Key("end")), range),
-                range => Assert.Equal(KeyRange.ClosedClosed(new Key(1L), new Key(2L)), range)
+                range => Assert.Equal(KeyRange.ClosedOpen(new Key("begin"), new Key("end")).ToProtobuf(s_conversionOptions), range.ToProtobuf(s_conversionOptions)),
+                range => Assert.Equal(KeyRange.ClosedClosed(new Key(1L), new Key(2L)).ToProtobuf(s_conversionOptions), range.ToProtobuf(s_conversionOptions))
             );
         }
 
+        [Fact]
+        public void ToStringKeyWithBuilder()
```
jskeet, line 250: Probably after this PR, I think it would be worth having one test class per production class. I don't mind whether they're in the same file or not though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -99,15 +96,14 @@ private KeySet(bool allKeys, IEnumerable<Key> keys, IEnumerable<KeyRange> ranges
...
-        public override int GetHashCode() => Protobuf.GetHashCode();
-
-        /// <inheritdoc/>
-        public override string ToString()
+        /// <summary>
+        /// Returns the string representation of this KeySet using the type conversion options specified in 
+        /// <see cref="SpannerConnectionStringBuilder"/>. The string representation is only for diagnostic purposes
+        /// and the implementation may change over time.
+        /// </summary>
+        /// <param name="builder">The <see cref="SpannerConnectionStringBuilder"/> instance.</param>
```
jskeet, line 104: I'd expand this description to explain the purpose of the builder - i.e. that it's used to derive default conversion options.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -226,45 +219,44 @@ public sealed class KeyRange
...
         {
-            var builder = new StringBuilder();
-            builder.Append(BeginInclusive ? '[' : '(');
-            builder.Append(BeginAt.KeyParts);
-            builder.Append(", ");
-            builder.Append(End.KeyParts);
-            builder.Append(EndInclusive ? ']' : ')');
-
-            return builder.ToString();
+            var stringBuilder = new StringBuilder();
```
jskeet, line 231: We could easily make this an expression-bodied member - again, that's later refactoring.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -226,45 +219,44 @@ public sealed class KeyRange
...
         }
 
-        private V1.KeyRange ToProtobuf()
+        internal V1.KeyRange ToProtobuf(SpannerConversionOptions options)
         {
             var range = new V1.KeyRange();
             if (BeginInclusive)
             {
-                range.StartClosed = BeginAt.KeyParts;
+                range.StartClosed = BeginAt.ToProtobuf(options);
```
jskeet, line 246: Another refactoring: I'd use an expression-bodied member here assigning all properties, just assigning null values where necessary. Basically remind me to have a look at this file later :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/Keys.cs
```diff
@@ -275,51 +267,49 @@ private V1.KeyRange ToProtobuf()
...
+        /// </summary>
+        /// <param name="builder">The <see cref="SpannerConnectionStringBuilder"/> instance.</param>
+        /// <returns>A string that represents the current key.</returns>
+        public string ToString(SpannerConnectionStringBuilder builder) => ToProtobuf(builder.ConversionOptions).ToString();
 
-        private static ListValue ToListValue(SpannerParameterCollection parameters)
-        {
-            return new ListValue
+        /// <summary>
+        /// Gets the Protobuf value of this Key.
```
jskeet, line 301: Suggest: `Converts this key to a protobuf value, with default type conversion options as supplied in <paramref name="options" />.`

jskeet (approved):
Looking good - various nits which are basically more around later refactoring than anything else.

(This has ended up being a relatively small change, which is encouraging!)

Rishabh-V:
> Looking good - various nits which are basically more around later refactoring than anything else.
> 
> (This has ended up being a relatively small change, which is encouraging!)

Thanks Jon. I will try to fix these in the next commit.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
amanda-tarafa: +1 to adding a test for non default options.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/KeysTests.cs
amanda-tarafa: Yes, this that Jon is saying is what I meant in my earlier comment "obtain the protobuf value and make sure it is what you are expecting". `actual` would be the protobuf value. Sorry if I wasn't clear enough.

Rishabh-V:
> > Looking good - various nits which are basically more around later refactoring than anything else.
> > (This has ended up being a relatively small change, which is encouraging!)
> 
> Thanks Jon. I will try to fix these in the next commit.

Hi Jon. I have done all the changes (except for one expression body member). Running into a strange issue while using expression bodied member in ToProtobuf

```
internal V1.KeyRange ToProtobuf(SpannerConversionOptions options) =>
            new V1.KeyRange
            {
                StartClosed = BeginInclusive ? BeginAt.ToProtobuf(options) : default,
                StartOpen = BeginInclusive ? default : BeginAt.ToProtobuf(options),
                EndClosed = EndInclusive ? End.ToProtobuf(options) : default,
                EndOpen = EndInclusive ? default : End.ToProtobuf(options)
            };
```
This isn't giving the same desired Protobuf value. Trying to figure out what is done incorrectly.

amanda-tarafa (approved):
Yep, this looks good!

jskeet (approved):

PR #8557, opened 2022-06-06
Merge gax-v4 branch into main

jskeet:


jskeet:
Difference detection breakage is fine. The OwlBot post processor has an older version of .NET - I'll fix that afterwards. (OwlBot will be using an old version of the generator for a short while anyway...)

amanda-tarafa (approved):
LGTM

PR #8558, opened 2022-06-06
First chunk of gax-v4 rebasing

jskeet:
(Rebasing all 117 commits was failing; this is just the first 63.)

We still expect OwlBot and breaking changes to fail.


jskeet:
Even though this is completely broken, I'm going to rebase this in - it'll then be fixed very soon by the next change.

PR #8560, opened 2022-06-06
docs: Minimal doc changes to support starting to release libraries

jskeet:
(There are larger docs changes coming, and we may well want to add to this set of breaking changes, but this allows us to start linking to the new doc as we release.)


amanda-tarafa (approved):
LGTM

PR #8561, opened 2022-06-06
chore: Update the OwlBot Dockerfile to use .NET 6

jskeet:


amanda-tarafa (approved):
LGTM

I have a vague recolection that the hash of this dockerfile needed to be updated somewhere? But I might be confusing things.

jskeet:
I *think* it all happens automatically - but we'll see :)

PR #8568, opened 2022-06-07
chore: Update Dockerfile for OwlBot to install .NET Core 3.1

jskeet:
This is required while the generator still uses .NET Core 3.1.
While we expect to update this soon, this temporary measure is the
smallest change that will get OwlBot working.


Rishabh-V (approved):
LGTM

PR #8569, opened 2022-06-07
docs: Add TOC entry for GAX v4 breaking changes

jskeet:
Without this, it looks like the file isn't actually published on
cloud.google.com.


Rishabh-V (approved):
LGTM

PR #8625, opened 2022-06-08
tools: Allow a default message to be specified in batch release

jskeet:
The "default default" message is just "No API surface changes; just
dependency updates" but for new major versions (or similar) that's
not as useful as a canned message.


amanda-tarafa (approved):
Ohh this is nice!

PR #8731, opened 2022-06-09
chore: fix PubSub build by removing multi-channel test

jskeet:
I don't think this was *working* before anyway as it wasn't enabling
subchannel counting. Since moving GrpcInfo in an earlier commit,
this wouldn't even have compiled; simplest to remove it now.

(The Grpc.Core-specific tests are definitely less useful now that we
expect the majority of calls to be via Grpc.Net.Client.)


amanda-tarafa (approved):
LGTM

PR #8732, opened 2022-06-09
tools: Improve show-dependency-levels command

jskeet:
We were already showing that Google.Cloud.BinaryAuthorization.V1
depends on a group - but we weren't excluding it from the earlier
display.

With the change in place, each API should be listed exactly once,
and in release order.


amanda-tarafa (approved):
LGTM

PR #8760, opened 2022-06-09
docs: Improve documentation for Google.Cloud.Logging.Console

jskeet:
Most importantly, this will include installation instructions.

### DIFF
--- a/apis/Google.Cloud.Logging.Console/docs/index.md
+++ b/apis/Google.Cloud.Logging.Console/docs/index.md
@@ -5,27 +5,7 @@ console
 formatter](https://docs.microsoft.com/en-us/dotnet/core/extensions/console-log-formatter)
 designed to be used with [Cloud Logging](https://cloud.google.com/logging/docs).
 
-The formatter writes JSON entries in the format expected by Cloud
-Logging. In particular, it uses the following JSON property names:
-
-- `message`: the formatted log message
-- `severity`: the log entry severity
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.Logging.Console/docs/index.md
```diff
@@ -50,3 +30,27 @@ using Google.Cloud.Logging.Console;
...
+  `IncludeScopes` option is enabled and there is scope information
+  available
+
+No timestamp is included in the JSON, as this is automatically
+picked up by the logging agent.
+
+Most log entry severity mappings are obvious, but the ASP.NET Core
+`LogLevel.Trace` level maps to `DEBUG` in Cloud Logging.
+
+When viewing the log entries in the Logs Explorer, the severity is
```
amanda-tarafa, line 54: Just a question to confirm, we now don't have any control on where the entries are shown as the Google Cloud log name is set by the agent, right?
This is fine, and not much we can do about it, but something to take into account as we used to get some questions around this before.

amanda-tarafa (approved):
LGTM

apis/Google.Cloud.Logging.Console/docs/index.md
jskeet: At the moment, that's right. I suspect if we start including extra information in the JSON, that would be used - and we've got the options for that when we know what we want to add.

PR #8764, opened 2022-06-10
Add (but don't release) Batch API

jskeet:


amanda-tarafa (approved):

amanda-tarafa:
I've added the label so that snippet bot does not require product prefixes and it's now all green.

PR #8765, opened 2022-06-10
Prototype of what CloudEvent integration might look like

jskeet:
Using extension methods, we could end up with much the same experience but with a separate library. That would avoid the "pure API" library from needing a CloudEvents dependency.

If we were to use a JSON representation instead of a protobuf representation, we could depend on one of the CloudEvents JSON format packages, allowing the data to be serialized automatically - but that would be opinionated in terms of which JSON library was being used.


jskeet:
Closing as a PR, but I'll leave the branch available.

PR #8767, opened 2022-06-10
Adding operators

anshuldavid13:


amanda-tarafa (approved):
LGTM

But, do we have integration tests for the operations, if so, we should add some as well for the new operators.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/FilterPartial.cs
```diff
@@ -62,6 +62,34 @@ public static Filter And(IEnumerable<Filter> filters)
...
+        /// a value with <see cref="Value.NullValue"/> set.</param>
+        /// <returns>The newly created filter.</returns>
+        public static Filter NotEqual(string propertyName, Value propertyValue) =>
+            Property(propertyName, propertyValue, Operator.NotEqual);
+
+        /// <summary>
+        /// Creates a filter to check that the specified property is in a given array of values.
+        /// </summary>
+        /// <param name="value">The name of the property. Must not be null.</param>
+        /// <param name="collection">The array of values to compare against. </param>
```
jskeet, line 79: Can this be null? If so, what's the meaning? If not, we should probably validate it. (Ditto for NotIn.)

jskeet:
> But, do we have integration tests for the operations, if so, we should add some as well for the new operators.

We have integration tests that *include* filters, but not really testing the filter parts. I think the unit tests are fine here.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/FilterPartial.cs
jskeet: (Apologies for not catching this first time round...)

PR #8779, opened 2022-06-14
chore: Fix Google.Cloud.AssuredWorkloads.V1Beta1 generation

jskeet:
The YAML file has been renamed; it's not entirely clear why, but it
needs to be reflected in apis.json.


jskeet:
(I'll self-merge this.)

PR #8781, opened 2022-06-14
feat:Editorconfig 

Rishabh-V:
First draft of editorconfig. Most of the items are configured as suggestion/silent.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,210 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
+# top-most EditorConfig file
+root = true
+###############################
+# Core EditorConfig Options   #
+###############################
+# All files
+[*]
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+# Generated code
+[*{_AssemblyInfo.cs,.g.cs}]
+generated_code = true
+
+# XML project files
+[*.{csproj}]
+indent_size = 2
+
+# Code files
+
+###############################
+# .NET Coding Conventions     #
+###############################
+[*.{cs}]
+
+# Organize usings
+dotnet_sort_system_directives_first = false
+
+# this. preferences
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:suggestion
+dotnet_style_qualification_for_method = false:suggestion
+dotnet_style_qualification_for_event = false:suggestion
+
+# Language keywords vs BCL types preferences
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+dotnet_style_predefined_type_for_member_access = true:suggestion
+
+# Parentheses preferences
+dotnet_style_parentheses_in_arithmetic_binary_operators = always_for_clarity:silent
+dotnet_style_parentheses_in_relational_binary_operators = always_for_clarity:silent
+dotnet_style_parentheses_in_other_binary_operators = always_for_clarity:silent
+dotnet_style_parentheses_in_other_operators = never_if_unnecessary:silent
+
+# Modifier preferences
+csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:warning
+dotnet_style_require_accessibility_modifiers = for_non_interface_members:warning
+dotnet_style_readonly_field = true:warning
+
+# Expression-level preferences
+dotnet_style_object_initializer = true:suggestion
+dotnet_style_collection_initializer = true:suggestion
+dotnet_style_explicit_tuple_names = true:suggestion
+dotnet_style_null_propagation = true:suggestion
+dotnet_style_coalesce_expression = true:suggestion
+dotnet_style_prefer_is_null_check_over_reference_equality_method = true:suggestion
+dotnet_style_prefer_inferred_tuple_names = true:suggestion
+dotnet_style_prefer_inferred_anonymous_type_member_names = true:suggestion
+dotnet_style_prefer_auto_properties = true:suggestion
+dotnet_style_prefer_conditional_expression_over_assignment = true:silent
+dotnet_style_prefer_conditional_expression_over_return = true:silent
+csharp_prefer_simple_default_expression = true:suggestion
+
+###############################
+# Naming Conventions          #
+###############################
+# Style Definitions
+dotnet_naming_style.pascal_case_style.capitalization             = pascal_case
+
+# Use PascalCase for constant fields  
+dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = suggestion
+dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols  = constant_fields
+dotnet_naming_rule.constant_fields_should_be_pascal_case.style = pascal_case_style
+dotnet_naming_symbols.constant_fields.applicable_kinds            = field
+dotnet_naming_symbols.constant_fields.applicable_accessibilities  = *
+dotnet_naming_symbols.constant_fields.required_modifiers          = const
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+
+# static fields should have s_ prefix
+dotnet_naming_rule.static_fields_should_have_prefix.severity = suggestion
+dotnet_naming_rule.static_fields_should_have_prefix.symbols  = static_fields
+dotnet_naming_rule.static_fields_should_have_prefix.style = static_prefix_style
+dotnet_naming_symbols.static_fields.applicable_kinds   = field
+dotnet_naming_symbols.static_fields.required_modifiers = static
+dotnet_naming_symbols.static_fields.applicable_accessibilities = private, internal, private_protected
+dotnet_naming_style.static_prefix_style.required_prefix = s_
+dotnet_naming_style.static_prefix_style.capitalization = camel_case
+
+# internal and private fields should be _camelCase
+dotnet_naming_rule.camel_case_for_private_internal_fields.severity = suggestion
+dotnet_naming_rule.camel_case_for_private_internal_fields.symbols  = private_internal_fields
+dotnet_naming_rule.camel_case_for_private_internal_fields.style = camel_case_underscore_style
+dotnet_naming_symbols.private_internal_fields.applicable_kinds = field
+dotnet_naming_symbols.private_internal_fields.applicable_accessibilities = private, internal
+dotnet_naming_style.camel_case_underscore_style.required_prefix = _
+dotnet_naming_style.camel_case_underscore_style.capitalization = camel_case
+
+###############################
+# C# Coding Conventions       #
+###############################
+[*.cs]
+# var preferences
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+# Core EditorConfig Options   #
+###############################
+# All files
+[*]
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+# Generated code
```
jskeet, line 14: Why do we need any suggestions at all for generated code? Any changes would have to be in the generator.

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+
+# XML project files
+[*.{csproj}]
+indent_size = 2
+
+# Code files
+[*.{cs}]
+indent_size = 4
+insert_final_newline = true
+charset = utf-8-bom
```
jskeet, line 57: Maybe we should change to UTF-8 *without* a BOM? (The BOM is a bit of a Windows-centric concept, as I understand it.)

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+###############################
+# .NET Coding Conventions     #
+###############################
+[*.{cs}]
+
+# Organize usings
+dotnet_sort_system_directives_first = true
```
jskeet, line 64: I believe we should sort using directives, but I don't agree with putting system ones first - that's not what we do at the moment, as far as I'm aware. (Example: https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs)

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+dotnet_style_predefined_type_for_member_access = true:suggestion
+
+# Parentheses preferences
+dotnet_style_parentheses_in_arithmetic_binary_operators = always_for_clarity:silent
+dotnet_style_parentheses_in_relational_binary_operators = always_for_clarity:silent
+dotnet_style_parentheses_in_other_binary_operators = always_for_clarity:silent
+dotnet_style_parentheses_in_other_operators = never_if_unnecessary:silent
+
+# Modifier preferences
+csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:suggestion
```
jskeet, line 83: I'd be happy for all three of these to be warnings, in fact. I know that may generate quite a few warnings to be fixed (is there something we can run to check these outside VS?) but I think it would basically be good to do anyway. I can't immediately think of times where we don't want to do it.

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+# Use PascalCase for constant fields  
+dotnet_naming_rule.constant_fields_should_be_pascal_case.severity = suggestion
+dotnet_naming_rule.constant_fields_should_be_pascal_case.symbols  = constant_fields
+dotnet_naming_rule.constant_fields_should_be_pascal_case.style = pascal_case_style
+dotnet_naming_symbols.constant_fields.applicable_kinds            = field
+dotnet_naming_symbols.constant_fields.applicable_accessibilities  = *
+dotnet_naming_symbols.constant_fields.required_modifiers          = const
+dotnet_naming_style.pascal_case_style.capitalization = pascal_case
+
+# static fields should have s_ prefix
```
jskeet, line 116: It's worth everyone being aware that for "effective constants" we sometimes use Pascal case, e.g.

```csharp
private static TimeSpan DefaultTimeout = TimeSpan.FromMinutes(1);
```

We're not massively consistent about it though, and *normally* use `s_` - so maybe we should just update everything to do that.

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+dotnet_naming_symbols.private_internal_fields.applicable_kinds = field
+dotnet_naming_symbols.private_internal_fields.applicable_accessibilities = private, internal
+dotnet_naming_style.camel_case_underscore_style.required_prefix = _
+dotnet_naming_style.camel_case_underscore_style.capitalization = camel_case
+
+###############################
+# C# Coding Conventions       #
+###############################
+[*.cs]
+# var preferences
```
jskeet, line 139: I would probably make all of these silent, to be honest. In particular, in handwritten snippets we *don't* use `var`, and elsewhere I'm happy for it to be personal preference.

I realise these are only suggestions, but if we're presented with thousands of suggestions, half of which we don't want to change, we're unlikely to get any benefit from the other ones.

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+csharp_style_var_for_built_in_types = true:silent
+csharp_style_var_when_type_is_apparent = true:suggestion
+csharp_style_var_elsewhere = true:suggestion
+dotnet_style_predefined_type_for_locals_parameters_members = true:suggestion
+dotnet_style_predefined_type_for_member_access = true:suggestion
+
+csharp_using_directive_placement = outside_namespace:suggestion
```
jskeet, line 146: I think this should probably be a warning. We only use them inside namespaces in generated code.

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+csharp_style_pattern_matching_over_is_with_cast_check = true:suggestion
+csharp_style_pattern_matching_over_as_with_null_check = true:suggestion
+csharp_style_inlined_variable_declaration = true:suggestion
+
+# Null-checking preferences
+csharp_style_throw_expression = true:suggestion
+csharp_style_conditional_delegate_call = true:suggestion
+
+# Modifier preferences
+csharp_preferred_modifier_order = public,private,protected,internal,static,extern,new,virtual,abstract,sealed,override,readonly,unsafe,volatile,async:suggestion
```
jskeet, line 174: We've got this earlier.

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+
+# Expression-level preferences
+csharp_prefer_braces = true:suggestion
```
jskeet, line 177: We've got this earlier. (IMO for any given file type we should only specify each option once.)

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+# Indentation preferences
+csharp_indent_block_contents = true
+csharp_indent_braces = true
+csharp_indent_case_contents = true
+csharp_indent_case_contents_when_block = true
+csharp_indent_switch_labels = true
+csharp_indent_labels = one_less_than_current
+
+# Space preferences
+csharp_space_after_cast = false
```
jskeet, line 209: I believe this should be true if we want to follow the convention of most of the code. I'm open to debate as to changing it, but personally I like the space there :)

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+csharp_space_between_method_call_empty_parameter_list_parentheses = false
+csharp_space_between_method_declaration_name_and_open_parenthesis = false
+csharp_space_between_square_brackets = false
+
+# Wrapping preferences
+csharp_preserve_single_line_statements = true
+csharp_preserve_single_line_blocks = true
+
+# License header
+file_header_template = Copyright 2022 Google LLC\n\nLicensed under the Apache License, Version 2.0 (the "License").\nyou may not use this file except in compliance with the License.\nYou may obtain a copy of the License at \n\nhttps://www.apache.org/licenses/LICENSE-2.0 \n\nUnless required by applicable law or agreed to in writing, software \ndistributed under the License is distributed on an "AS IS" BASIS,\nWITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\nSee the License for the specific language governing permissions and \nlimitations under the License.
```
jskeet, line 238: Ooh, does this get applied when creating a new file? That'd be great.

Hopefully it *won't* show as a warning if it's something other than this though? (In particular, we shouldn't be updating the copyright year for existing files.)

.editorconfig
```diff
@@ -0,0 +1,250 @@
...
+csharp_style_namespace_declarations = file_scoped:suggestion
+csharp_style_prefer_null_check_over_type_check = true:suggestion
+csharp_style_prefer_local_over_anonymous_function = true:suggestion
+csharp_style_prefer_tuple_swap = true:suggestion
+csharp_style_implicit_object_creation_when_type_is_apparent = true:suggestion
```
jskeet, line 243: I'm not sure on this one yet. Let's see. (It will also add thousands and thousands of suggestions...)

jskeet (commented):
A few comments - I think we should chat more about this in the team meeting tomorrow, in terms of how many suggestions we want.

I tend to think that if there's a huge list of suggestions, we're unlikely to use them, so even the best of them become useless.

Is there a way of saying "Only make suggestions on lines that are changed according to git"? If so, that would be a good way of gradually modernizing the codebase.

.editorconfig
Rishabh-V: Sure Jon. Switched it off now. (System directives at top was what I followed earlier)

Rishabh-V (commented):
Addressed all the comments. Please have a look. Thank you.

.editorconfig
jskeet: We've still got a lot of entries in this section - do we need them at all? (Perhaps you just haven't pushed your change?)

Rishabh-V: Removed the not required entries for generated code.

.editorconfig
```diff
@@ -0,0 +1,208 @@
...
+[*]
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
+
+# Generated code
+[*{_AssemblyInfo.cs,.g.cs}]
+generated_code = true
+# XML project files
```
jskeet, line 17: Nit: let's have a blank line here :)

.editorconfig
```diff
@@ -0,0 +1,208 @@
...
+csharp_style_conditional_delegate_call = true:suggestion
+
+# Other features
+csharp_style_prefer_index_operator = true:suggestion
+csharp_style_prefer_range_operator = true:suggestion
+csharp_style_pattern_local_over_anonymous_function = false:none
+
+###############################
+# C# Formatting Rules         #
+###############################
```
jskeet, line 148: I'd expect the earlier `indent_size` etc lines (the ones *before* ".NET conventions") to be in here.

jskeet (approved):
A couple of minor tweaks, but other than that I suggest we get this merged and see how it "feels". We can definitely change it as we go.

.editorconfig
Rishabh-V: Added the blank line.

.editorconfig
```diff
@@ -0,0 +1,208 @@
...
+# XML project files
+[*.{csproj}]
+indent_size = 2
+
+# Code files
+[*.{cs}]
+indent_size = 4
+insert_final_newline = true
+charset = utf-8
+###############################
```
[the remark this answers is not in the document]
Rishabh-V: Added blank line and moved the three settings below.

.editorconfig
Rishabh-V: Done.

Rishabh-V (commented):
Thanks Jon. Incorporated your feedback.

Rishabh-V:
Merging now.

PR #8784, opened 2022-06-15
docs: Add more TOC entries

jskeet:
We *don't* have TOC entries for major-version.md or
breaking-gax2.md, but I think that's okay.


amanda-tarafa (approved):
LGTM

PR #8789, opened 2022-06-16
Add operators

anshuldavid13:


apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/FilterPartial.cs
```diff
@@ -62,6 +62,34 @@ public static Filter And(IEnumerable<Filter> filters)
         public static Filter Equal(string propertyName, Value propertyValue) =>
             Property(propertyName, propertyValue, Operator.Equal);
 
+        /// <summary>
```
jskeet, line 65: Haven't we already added these? I'm a bit confused about what's going on.

jskeet (commented):
I'm travelling so won't be able to do a significant review until Monday (probably; maybe tomorrow).

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/FilterPartial.cs
jskeet: If we've already added them, why are they showing up as new in this PR? That's what's confusing me. I'd expect the PR only to show *changes* between this and what's already merged. Perhaps you just need to rebase this PR and force push?

PR #8791, opened 2022-06-16
WIP

anshuldavid13:


apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
```diff
@@ -120,6 +121,46 @@ public void RunQuery_NoResults()
...
+            {
+                PropertyFilter = new PropertyFilter
+                {
+                    Op = Operator.In,
+                    Property = new PropertyReference { Name = "1" },
+                    Value = Value.ForNull()
+                }
+            };
+            var expectedQuery = new Query("inFilter") { Filter = expected };
+            Assert.Equal(db.RunQuery(actualQuery), db.RunQuery(expectedQuery));
```
jskeet, line 141: This still doesn't tell us what the server actually does though. That's what I really want to know, so we can document it - or prohibit it if it's not useful.

We want to know what it will match, if anything. This is effectively checking that the expected query is the actual query, and we can do that in a unit test. It's the *impact* of doing that in terms of returned results that's important.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
```diff
@@ -120,6 +121,50 @@ public void RunQuery_NoResults()
             Assert.Equal(MoreResultsType.NoMoreResults, singleResponse.Batch.MoreResults);
         }
 
+        [Fact]
+        public void InFilterQuery()
```
jskeet, line 125: Please change the names to indicate that these are checking null values, e.g. `InFilterQuery_NullArrayValue`

jskeet:
I've tried running the tests, and they fail with errors of:

- IN requires an ArrayValue.
- NOT_IN requires an ArrayValue.

So I think we should instead change the implementation to prevent callers from providing a null argument to start with.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
```diff
@@ -120,6 +121,50 @@ public void RunQuery_NoResults()
             Assert.Equal(MoreResultsType.NoMoreResults, singleResponse.Batch.MoreResults);
         }
 
+        [Fact]
```
jskeet, line 124: These integration tests should be removed now - they'll fail before we get there anyway (because we're throwing an exception).

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/FilterPartial.cs
```diff
@@ -76,19 +76,25 @@ public static Filter And(IEnumerable<Filter> filters)
...
         /// </summary>
         /// <param name="value">The name of the property. Must not be null.</param>
-        /// <param name="collection">The array of values to compare against. </param>
+        /// <param name="collection">The array of values to compare against. Must not be null. </param>
         /// <returns>The newly created filter.</returns>
-        public static Filter In(string value, ArrayValue collection) =>
-            Property(value, collection, Operator.In);
+        public static Filter In(string value, ArrayValue collection)
+        {
+            GaxPreconditions.CheckArgument(collection != null, nameof(collection), "Filter collection must not be null");
```
jskeet, line 83: Please use `GaxPreconditions.CheckNotNull(collection, nameof(collection))` here and below.

PR #8793, opened 2022-06-17
fix(Spanner): Fixes truncation for numeric types.

### LINKED ISSUE #8285: ToDecimal(LossOfPrecisionHandling.Truncate) performs rounding and not truncation.
This was discovered and pointed out by @jskeet  while reviewing the [code for PgNumeric](https://github.com/googleapis/google-cloud-dotnet/pull/8261) but applies to `SpannerNumeric`, `BigQueryNumeric`, `BigQueryBigNumeric` as well.

Issue: `ToDecimal` API of `PgNumeric` called with the parameter `LossOfPrecisionHandling.Truncate`  performs rounding of numeric value rather than truncation as one would expect by passing the parameter as `LossOfPrecisionHandling.Truncate` 

E.g.
```
var numeric = PgNumeric.Parse("123456789012345678900.123456789");
var decimalValue = numeric.ToDecimal(LossOfPrecisionHandling.Truncate);
```
One would expect the `decimalValue` to be 123456789012345678900.12345678
However, actual value is rounded to 123456789012345678900.1234567*9*

The reason for this behavior is that the code of `ToDecimal()` first calls the `ToString()` method to convert numeric value to string and then parses the string into decimal by calling `decimal.Parse` which [performs rounding and not truncation](https://docs.microsoft.com/en-us/dotnet/api/system.decimal.parse?view=netstandard-2.1) 

The code fix needs to be made in PgNumeric, SpannerNumeric, BigQueryNumeric, BigQueryBigNumeric (and other numeric implementations if any)

amanda-tarafa:
Fixes #8285


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -329,56 +330,104 @@ public void FromDecimal(decimal input, string expectedValue)
...
-        [InlineData("12345.123456000")]
-        [InlineData("-0.000001")]
-        [InlineData("1234567891.123456789123")]
-        [InlineData("-12345678912.12345678912")]
-        [InlineData("123456789123456789.12345678912")]
-        [InlineData("12345678912345678912345678.1234567891234567891234")]
-        [InlineData("9999999999999999999999999999.999999999999999999999999")]
-        [InlineData("100")]
-        [InlineData("1.33333333333333333")]
-        public void ToDecimal(string input)
```
amanda-tarafa, line 345: I've removed this test entirely as this is precisely what we stopped doing, i.e. rounding as decimal does, instead we are now truncating.

Some of the relevant test cases from here I've moved to the other tests.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -329,56 +330,104 @@ public void FromDecimal(decimal input, string expectedValue)
...
+                Assert.Equal(actual, actual2);
+            }
         }
 
-        [Fact]
-        public void ToDecimal_Overflow()
+        [Theory]
+        [InlineData("0.00000000000000000000000000000000000000", "0.0000000000000000000000000000")]
+        [InlineData("-0.00000000000000000000000000000000000000", "0.0000000000000000000000000000")]
+        public void ToDecimal_LossOfPrecision_Truncate_Zero(string text, string expectedDecimalText)
```
amanda-tarafa, line 407: Zero on it's own as the negative sign is not included on the string representation.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -329,56 +330,104 @@ public void FromDecimal(decimal input, string expectedValue)
...
-        [Theory]
-        [InlineData("12345678.1234567891234567891234567891")]
-        [InlineData("123456789012345678900.123456789")]
-        public void ToDecimal_LossOfPrecision_Throw(string input)
+        [Theory, CombinatorialData]
+        public void ToDecimal_LossOfPrecision_Throw(
+            [CombinatorialValues(
+                "79228162514264337593543950335.00000000000000000000000000000000000001", // decimal.MaxValue + epsilon; doesn't count as overflow
+                "123456789012345678900.123456789", // Simpler example of more significant digits than decimal can handle
+                "0.00000000000000000000000000000000000000" // Even for zero, this loses information as there are more significant digits than decimal can handle
```
amanda-tarafa, line 422: Please note this test case, as I believe this is right, i.e. that we are losing information here even if for zero, because this number is not fixed point.

amanda-tarafa (commented):
@jskeet I've left a few comments to highlight things to look at carefully.
Thanks!

jskeet:
Will have a look on Monday. Definitely too hot and tired to do this justice. Thanks so much for tackling it though!

amanda-tarafa:
Yep, no rush at all!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -329,56 +330,104 @@ public void FromDecimal(decimal input, string expectedValue)
...
-        public void ToDecimal_LossOfPrecision_Truncate(string input, string expectedValue)
+        [InlineData("0.1234567891234567891234567899999", "0.1234567891234567891234567899")]
+        [InlineData("1.1234567891234567891234567899999", "1.1234567891234567891234567899")]
+        [InlineData("123456789012345678900.123456780", "123456789012345678900.12345678")]
+        [InlineData("123456789012345678900.123456785", "123456789012345678900.12345678")]
+        [InlineData("123456789012345678900.123456789", "123456789012345678900.12345678")]
+        [InlineData("123456789012345678900.123456770", "123456789012345678900.12345677")]
+        [InlineData("123456789012345678900.123456775", "123456789012345678900.12345677")]
+        [InlineData("123456789012345678900.123456779", "123456789012345678900.12345677")]
+        [InlineData("7.2345678901234567890123456789", "7.2345678901234567890123456789")]
```
jskeet, line 383: Possibly add a comment explaining the difference between this test case and the next one?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
```diff
@@ -329,56 +330,104 @@ public void FromDecimal(decimal input, string expectedValue)
...
+            PerformTest(text, expectedDecimalText);
+            PerformTest("-" + text, "-" + expectedDecimalText);
+
+            static void PerformTest(string text, string expectedDecimalText)
+            {
+                PgNumeric numeric = PgNumeric.Parse(text);
+                decimal actual = numeric.ToDecimal(LossOfPrecisionHandling.Truncate);
+                Assert.Equal(expectedDecimalText, actual.ToString(CultureInfo.InvariantCulture));
+                // Conversion via the explicit conversion should do the same thing
+                decimal actual2 = (decimal)numeric;
```
jskeet, line 399: Nit: space after cast (given that I've asked Rishabh to include that in the editorconfig :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
jskeet: Maybe add that as a comment for posterity?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
jskeet: (In some other tests too.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PgNumericTests.cs
jskeet: Hmm... I would disagree here. There are more *specified* digits, but they're not *significant* in the normal mathematical sense. Basically I think we should be okay to lose trailing zeroes after a decimal point. On the other hand, as we do preserve them normally, we're definitely losing *information* - just not significant digits :)

I think it's okay to keep the behavior, but let's document it and change the comment.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/SpannerNumericTest.cs
```diff
@@ -105,7 +124,11 @@ public void ConversionFromUInt64(ulong input)
             [CombinatorialValues(
                 "79228162514264337593543950335", // decimal.MaxValue
                 "0.000000001", // Epsilon for numeric
-                "12345678901234567890.123456789" // Maximum significant digits with 9dps
+                "12345678901234567890.123456789", // Maximum significant digits with 9dps
+                "72345678901234567890.123456789",
+                "8234567890123456789.012345678",
```
jskeet, line 129: Again, I think a comment explaining this difference would be useful.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -275,7 +275,38 @@ private static string TryParseImpl(string text, out PgNumeric value)
...
+        /// <para>
+        /// A function to calculate how many decimal places should the
+        /// representation be truncated to. This number should be greater or equal to zero and less than or equal to
+        /// the total decimal places in the number.
+        /// </para>
+        /// <para>
+        /// The parameters of this function are:
+        /// <list type="bullet">
+        /// <item>
+        /// The text representation of the the scaled value, scaled <see cref="_scale"/>.
```
jskeet, line 297: We're actually calling it with an empty string for zero - call that out here?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -431,15 +465,70 @@ public static PgNumeric FromDecimal(decimal value)
...
         {
-            // FIXME: Awful performance!
-            string text = ToString();
-            // Handles overflow.
-            decimal dec = Decimal.Parse(text, CultureInfo.InvariantCulture);
-            if (lossOfPrecisionHandling == LossOfPrecisionHandling.Throw && this != dec)
+            if (_notANumber)
             {
-                throw new InvalidOperationException("Conversion would lose information");
+                throw new OverflowException($"NaN cannot be represented as {typeof(decimal).FullName}");
```
jskeet, line 470: Is this existing behavior for NaN? I'm not sure whether OverflowException is suitable here - I think I'd use InvalidOperationException. But if this is the existing behavior, obviously we should keep it.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
```diff
@@ -431,15 +465,70 @@ public static PgNumeric FromDecimal(decimal value)
...
+                if (decimalPlaces == 0 || integerPlaces >= 29)
+                {
+                    decimalPrecision = 0;
+                }
+                // If the number has at most 28 significant digits we can represent it fully.
+                else if (integerPlaces + decimalPlaces <= 28)
+                {
+                    decimalPrecision = decimalPlaces;
+                }
+                // If there are no integer places we can represent up to 28 decimal places.
```
jskeet, line 493: Can we not represent 29 decimal places in some cases? or does the structure of decimal mean that 29 significant digits always includes one before the decimal point?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs
```diff
@@ -379,27 +414,76 @@ public static SpannerNumeric FromDecimal(decimal value, LossOfPrecisionHandling
...
+                }
+                // integerPlaces + decimalPlaces >= 29 and
+                // 1 <= integerPlaces <= 28 which means we may have room for some of the decimal places
+                // but we need to find out if the number can have 28 or 29 significant digits.
+                else
+                {
+                    int significantPlaces = 29;
+                    string maxDecimalText = decimal.MaxValue.ToString(CultureInfo.InvariantCulture);
+                    // Let's compare digit by digit to check if our unscaled number is greater than decimal max,
+                    // in which case we can only have 28 significan places.
```
jskeet, line 461: significan => significant

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs
```diff
@@ -379,27 +414,76 @@ public static SpannerNumeric FromDecimal(decimal value, LossOfPrecisionHandling
...
+                }
+                // If the number has at most 28 significant digits we can represent it fully.
+                // Note that this takes of the case where there are no integer places as well,
+                // since decimal places for SpannerNumeric are at most 9.
+                else if (integerPlaces + decimalPlaces <= 28)
+                {
+                    decimalPrecision = decimalPlaces;
+                }
+                // integerPlaces + decimalPlaces >= 29 and
+                // 1 <= integerPlaces <= 28 which means we may have room for some of the decimal places
```
jskeet, line 454: integerPlaces could be zero here - we don't have the same check that's in PgNumeric. That may be a correct code difference, but I think it means the comment is incorrect.

jskeet (commented):
A few things to discuss, but I think it'll be fine in general. I'd love to get rid of some of the duplication, but I'd rather get it right first and refactor later :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
amanda-tarafa: This was a slight surprise for me as well (when I did BigQuery), but apparently "there's always one digit before the decimal point" as demonstrated by the following two snippets:

``` csharp
decimal d = decimal.Parse("0.12345678901234567890123456789");
Console.WriteLine(d);
```
writes (decimal parsing rounds instead of truncating but still proves the point):
```
0.1234567890123456789012345679
```

but:

```
decimal o = decimal.Parse("12345678901234567890123456789");
Console.WriteLine(o);
```
doesn't overflow, which is expected, but makes the above case surprising, instead it just prints the whole number:
```
12345678901234567890123456789
```

It really is unexpected as a decimal is supposed to be stored as a 96 bit integer + 32 bits of flags for sign and scale, so I really don't understand why that 0 before the decimal point is counting towards the amount of digits.

If I didn't know better I'd think there was a bug in `decimal.Parse` :) that's counting that zero as relevant.

Or I'm really missunderstanding something here, but the fact that it jumped at you as well makes me think I'm not.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs
amanda-tarafa: Nope, integerPlaces being zero would have been solved on the `else if` above, as the maximum number of decimal places on `SpannerNumeric` is 9, so:
```
integerPlaces == 0 => integerPlaces + decimalPlaces <= 9 <= 28
```

There's a note on the `else if` above but let me know if that's not clear enough.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
jskeet: Remind me to have another look at this later in the week. (It needn't block merging.) It would be interesting to try doing this from the raw integers...

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerNumeric.cs
jskeet: Ah yes - I hadn't read it carefully enough.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
amanda-tarafa: Yep, I continued to tried things out, and if there's a bug (unlikely) is not in `decimal.Parse`:

```
decimal e = 0.12345678901234567890123456789m;
Console.WriteLine(e);
```

also prints the 28 significant digit rounded number:

```
0.1234567890123456789012345679
```

This is really surprising. I'll fidle with the bits more actually and see what I find.

jskeet:
I can't reply to your comment about whether 0.000 has 3 significant decimal digits or not for some reason. Having just looked into the Wikipedia definition, you're right - so it's fine to leave as-is.

jskeet (approved):

amanda-tarafa:
Thanks! Squashing first and then mergin on green.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PgNumeric.cs
amanda-tarafa: Just for posterity, there's nothing surprising going on here, although `0.12345678901234567890123456789` scaled, i.e. `12345678901234567890123456789` would really fit on the decimal 96bits for the scaled integer, the exponent would need to be `29` which is greater than the max possible exponent `28`.

PR #8831, opened 2022-06-30
Prototype of PublisherClient client creation setting change

jskeet:
This is an initial prototype for comment before I add tests and documentation (and do the same for SubscriberClient).

The intention is that `ClientCreationSettings` can stop growing, and we'd prefer people to use the version that modifies the builder. For example:

```csharp
var settings = new PublisherClient.ClientCreationSettings(builder => builder.CredentialsPath = "...")
    .WithClientCount(5);
var client = PublisherClient.Create(settings);
```

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2018 Google LLC
+// Copyright 2018 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -120,7 +120,19 @@ namespace Google.Cloud.PubSub.V1
                 PublisherServiceApiSettings publisherServiceApiSettings = null,
                 ChannelCredentials credentials = null,
                 string serviceEndpoint = null)
-                : this(clientCount, publisherServiceApiSettings, credentials, serviceEndpoint, EmulatorDetection.None)
+                : this(clientCount, publisherServiceApiSettings, credentials, serviceEndpoint, EmulatorDetection.None, null)
+            {
+            }
+
+            // TODO: Allow null?
+
+            /// <summary>
+            /// Creates a settings object which executes the specified configuration action
+            /// on a client builder after applying other settings.
+            /// </summary>
+            /// <param name="configureAction">The configuration action. Must not be null.</param>
+            public ClientCreationSettings(Action<PublisherServiceApiClientBuilder> configureAction)
+                : this(null, null, null, null, EmulatorDetection.None, configureAction)
             {
             }
 
@@ -129,14 +141,22 @@ namespace Google.Cloud.PubSub.V1
                 PublisherServiceApiSettings publisherServiceApiSettings,
                 ChannelCredentials credentials,
                 string serviceEndpoint,
-                EmulatorDetection emulatorDetection)
+                EmulatorDetection emulatorDetection,
+                Action<PublisherServiceApiClientBuilder> configureAction)
             {
                 ClientCount = clientCount;
                 PublisherServiceApiSettings = publisherServiceApiSettings;
                 Credentials = credentials;
                 ServiceEndpoint = serviceEndpoint;
                 EmulatorDetection = emulatorDetection;
+                ConfigureAction = configureAction;
             }
+
+            /// <summary>
+            /// The configuration action to perform on the client builder after all other properties
+            /// in these settings have been applied.
+            /// </summary>
+            public Action<PublisherServiceApiClientBuilder> ConfigureAction { get; }
 
             /// <summary>
             /// Creates a new instance of this type with the specified emulator detection value.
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
```diff
@@ -120,7 +120,19 @@ public sealed class ClientCreationSettings
                 PublisherServiceApiSettings publisherServiceApiSettings = null,
                 ChannelCredentials credentials = null,
                 string serviceEndpoint = null)
-                : this(clientCount, publisherServiceApiSettings, credentials, serviceEndpoint, EmulatorDetection.None)
+                : this(clientCount, publisherServiceApiSettings, credentials, serviceEndpoint, EmulatorDetection.None, null)
+            {
+            }
+
+            // TODO: Allow null?
```
jskeet, line 127: And if not, validate it :)

amanda-tarafa: Yep, why not, everything else may be null.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
```diff
@@ -120,7 +120,19 @@ public sealed class ClientCreationSettings
...
+
+            /// <summary>
+            /// Creates a settings object which executes the specified configuration action
+            /// on a client builder after applying other settings.
+            /// </summary>
+            /// <param name="configureAction">The configuration action. Must not be null.</param>
+            public ClientCreationSettings(Action<PublisherServiceApiClientBuilder> configureAction)
```
amanda-tarafa, line 134: Could we at some point just deprecate `ClientCreationSettings` and rely entirely on the Builder?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
```diff
@@ -146,7 +166,18 @@ public sealed class ClientCreationSettings
...
-                return new ClientCreationSettings(ClientCount, PublisherServiceApiSettings, Credentials, ServiceEndpoint, emulatorDetection);
+                return new ClientCreationSettings(ClientCount, PublisherServiceApiSettings, Credentials, ServiceEndpoint, emulatorDetection, ConfigureAction);
+            }
+
+            /// <summary>
+            /// Creates a new instance of this type with the specified client count.
+            /// </summary>
+            /// <param name="clientCount">The new client count, or null to reset to the default</param>
+            /// <returns>The new instance</returns>
+            public ClientCreationSettings WithClientCount(int? clientCount)
```
amanda-tarafa, line 177: And WithConfigureAction? For completeness mostly.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
```diff
@@ -274,6 +305,8 @@ private static async Task<PublisherClient> CreateMaybeAsync(TopicName topicName,
                     Settings = clientCreationSettings?.PublisherServiceApiSettings,
                     ChannelOptions = grpcChannelOptions
                 };
+                clientCreationSettings?.ConfigureAction?.Invoke(builder);
+
                 var channel = isAsync ?
```
amanda-tarafa, line 310: What if what the ConfigureAction does is set a CallInvoker? Possibly a corner case but still...

amanda-tarafa: Also, I don't fully understand/remember why we need the two builders, well except for the shutdown of the channels. But if it's just for the shutdown of the channels then the comment in 298 is innacurate?
And also, for I think we don't need the second builder for shutdown now as there's the `LastCreateChannel` or similar property now?

amanda-tarafa (commented):
I've left a few comments, but to be taken with a grain of salt as I'm not that familiar with this code.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
amanda-tarafa: Actually, I meant, deprecate `ClientCreantionSettings` and have a `PublisherClientBuilder` with a `PublisherServiceApiClientBuilder` property.

amanda-tarafa (commented):
Updated one of my comments.

jskeet:
> Actually, I meant, deprecate ClientCreantionSettings and have a PublisherClientBuilder with a PublisherServiceApiClientBuilder property.

Well we could have a PublisherClientBuilder deriving from ClientBuilderBase, I expect. (And with a ClientCount property.) That would be neater, if we could get it to work. I'll have a think - this isn't urgent.

jskeet:
Closing in favour of https://github.com/googleapis/google-cloud-dotnet/pull/8849

PR #8849, opened 2022-07-08
feat: add PublisherClientBuilder and SubscriberClientBuilder

jskeet:
If this looks good, I'll implement it fully and do the same for
SubscriberClientBuilder.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2018 Google LLC
+// Copyright 2018 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -103,7 +103,10 @@ namespace Google.Cloud.PubSub.V1
 
         /// <summary>
         /// Settings for creating <see cref="PublisherServiceApiClient"/>s.
+        /// This type is now obsolete; please use <see cref="PublisherClientBuilder"/> which provides an
+        /// API surface consistent with other clients (as well as additional Pub/Sub-specific properties such as <see cref="PublisherClientBuilder.ClientCount"/>).
         /// </summary>
+        [Obsolete("Use PublisherClientBuilder to customize client settings.")]
         public sealed class ClientCreationSettings
         {
             /// <summary>
@@ -207,6 +210,22 @@ namespace Google.Cloud.PubSub.V1
         public static BatchingSettings ApiMaxBatchingSettings { get; } = new BatchingSettings(1000L, 10_000_000L, null);
 
         /// <summary>
+        /// Create a <see cref="PublisherClient"/> instance associated with the specified <see cref="TopicName"/>, using default settings.
+        /// </summary>
+        /// <param name="topicName">The <see cref="TopicName"/> to publish messages to. Must not be null.</param>
+        /// <returns>A <see cref="PublisherClient"/> instance associated with the specified <see cref="TopicName"/>.</returns>
+        public static PublisherClient Create(TopicName topicName) =>
+            CreateBuilder(topicName, null, null).Build();
+
+        /// <summary>
+        /// Creates a <see cref="PublisherClient"/> instance associated with the specified <see cref="TopicName"/>, using default settings.
+        /// </summary>
+        /// <param name="topicName">The <see cref="TopicName"/> to publish messages to. Must not be null.</param>
+        /// <returns>A <see cref="PublisherClient"/> instance associated with the specified <see cref="TopicName"/>.</returns>
+        public static Task<PublisherClient> CreateAsync(TopicName topicName) =>
+            CreateBuilder(topicName, null, null).BuildAsync();
+
+        /// <summary>
         /// Create a <see cref="PublisherClient"/> instance associated with the specified <see cref="TopicName"/>.
         /// The default <paramref name="settings"/> and <paramref name="clientCreationSettings"/> are suitable for machines with
         /// high network bandwidth (e.g. Google Compute Engine instances). If running with more limited network bandwidth, some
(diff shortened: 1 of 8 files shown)


Rishabh-V (approved):
The refactoring looks good to me. All the client builder stuff is now in its own class. Given this is a prototype, I assume that docs are deliberately not implemented. 

I am not well versed with gRPC (my homework to checkout on channels, ChannelPool and subchannels, so that I understand that better)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
```diff
@@ -0,0 +1,135 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.PubSub.V1;
+
+// TODO: Should the TopicName just be a regular property?
```
amanda-tarafa, line 26: Yes, and we can validate that it's present on build same as we do with the rest. Granted the rest has defaults, but still, this is a builder.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
```diff
@@ -220,10 +220,9 @@ internal void Validate()
         /// <see cref="PublisherServiceApiClient"/>s.</param>
         /// <param name="settings">Optional. <see cref="Settings"/> for creating a <see cref="PublisherClient"/>.</param>
         /// <returns>A <see cref="PublisherClient"/> instance associated with the specified <see cref="TopicName"/>.</returns>
+        [Obsolete("TODO - instructions for PublisherClientBuilder, or maybe don't make it obsolete")]
```
amanda-tarafa, line 223: I would even make ClientCreationSettings obsolete.

amanda-tarafa (commented):
I like this way better.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
jskeet: Mmm... it just feels odd for something that's required not to be specified in the constructor. I'm trying to think of similar situations we've got in other libraries - will check Storage and BigQuery.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClient.cs
jskeet: Maybe introduce the builder in the next release, and only make things obsolete a bit later? It's a shame there isn't a "suggestion-only" obsolete to allow us to ramp up even more gradually.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
jskeet: `BigQueryClientBuilder` has it as a normal property, validated - so yes, I'll switch to that pattern for consistency.

jskeet: Maybe we should introduce new client static Create/CreateAsync overloads without any settings for "I just want to create it for this topic, using defaults". Let's talk about this more in our meeting :)

jskeet:
Have now updated this:

- Docs everywhere
- New single-parameter Create/CreateAsync methods accepting TopicName (not obsolete)
- Both ClientCreationSettings and the Create methods which accept settings are obsolete
- TopicName is a regular property in the builder that has to be non-null before building

PTAL; if this looks good, I'll roll out the same changes for SubscriberClient.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
```diff
@@ -0,0 +1,153 @@
...
+                : builder.Build();
+            var channel = builder.LastCreatedChannel;
+            shutdowns[i] = channel is null ? () => Task.CompletedTask : channel.ShutdownAsync;
+        }
+        Func<Task> shutdown = () => Task.WhenAll(shutdowns.Select(x => x()));
+        return new PublisherClientImpl(TopicName, clients, settings, shutdown);
+    }
+
+    /// <inheritdoc />
+    protected override ChannelPool GetChannelPool() => throw new NotImplementedException();
```
amanda-tarafa, line 145: Can you override CanUseChannelPool to always return false as well?

amanda-tarafa (commented):
Yep, this looks really good!

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
jskeet: There's no need - we'll never end up calling that, because we'll never call CreateChannel on this class.

jskeet:
Okay, this is now ready for review - first commit is publisher changes, second commit is subscriber changes.

(The first commit probably still has a modifier reordering to subscriber code, but let's ignore that...)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
```diff
@@ -0,0 +1,153 @@
...
+    /// </summary>
+    public int? ClientCount
+    {
+        get => _clientCount;
+        set
+        {
+            if (value is int count)
+            {
+                // Fairly arbitrary upper limit.
+                _clientCount = GaxPreconditions.CheckArgumentRange(count, nameof(value), 1, 256);
```
amanda-tarafa, line 59: Maybe check this in Validate? Sorry for not saying anything before.

And same in SubscriberClientBuilder.

amanda-tarafa (approved):
LGTM

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
jskeet: Previously I had it in validate, I think :)

The expectation is that it's more useful in the property setter, as it's immediately wrong, and it's never useful to explicitly set it to an invalid value. (Compare that with TopicName, which is null to start with, and has to be non-null in the end, but if we made `builder.TopicName = builder.TopicName;` throw an exception, that would be odd.)

This is similar to what we do in Storage, e.g. https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UploadObjectOptions.cs

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
```diff
@@ -0,0 +1,153 @@
...
+    /// <summary>
+    /// Creates a new instance with default settings. <see cref="TopicName"/> must be set
+    /// before calling either <see cref="Build"/> or <see cref="BuildAsync"/>.
+    /// </summary>
+    public PublisherClientBuilder() : base(PublisherServiceApiClient.ServiceMetadata)
+    {
+    }
+
+    /// <summary>
+    ///
```
Rishabh-V, line 102: Do we need docs here and below?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientBuilder.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// before calling either <see cref="Build"/> or <see cref="BuildAsync"/>.
+    /// </summary>
+    public SubscriberClientBuilder() : base(SubscriberServiceApiClient.ServiceMetadata)
+    {
+    }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <returns></returns>
```
Rishabh-V, line 103: Here as well?

Rishabh-V (approved):
LGTM
(Maybe couple of places where we can update docs)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
jskeet: Yes indeed - thanks for catching this. I'll fix these and then merge.

PR #8884, opened 2022-07-14
Dotnet Storage Performance Benchmarking Tool

anuragsrivstv:


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Configuration.cs
```diff
@@ -0,0 +1,84 @@
...
+        private Configuration(int numOfSamples, int dataSizeInBytes, string outputFolder, int uploadChunkSizeInBytes, int downloadChunkSizeInBytes)
+        {
+            DataSizeInBytes = dataSizeInBytes;
+            LocalFile = Path.GetTempFileName();
+            OutputFolder = outputFolder;
+            UploadChunkSizeInBytes = uploadChunkSizeInBytes;
+            DownloadChunkSizeInBytes = downloadChunkSizeInBytes;
+            NumOfSamples = numOfSamples;
+        }
+        internal static Configuration FromCommandLineArguments(string[] args)
```
jskeet, line 44: Please add a blank line between methods.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Configuration.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            if (!int.TryParse(args[3], out int uploadChunkSize) || !int.TryParse(args[4], out int downloadChunkSize)) // args[2] and args[4] is the chunk size in which data can be upoaded or downloaded.
+            {
+                throw new ArgumentException("Error: Provided buffer size is not an integer.");
+            }
+
+            return new Configuration(numOfSamples, ConvertMBIntoBytes(fileSize), args[2], ConvertMBIntoBytes(uploadChunkSize), ConvertMBIntoBytes(downloadChunkSize));
+        }
+
+        private static int ConvertMBIntoBytes(int dataInMiB)
```
jskeet, line 79: Change "MB" to "MiB" in the name, given the parameter name? And maybe "Into" to just "To"?
I'd also implement this as an expression-bodied method.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Configuration.cs
```diff
@@ -0,0 +1,84 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.ClientTesting;
+using System;
+using System.IO;
+
+namespace Google.Cloud.Storage.V1.BenchmarkingTool
```
jskeet, line 19: Given that this is a new class, I'd use a file-scoped namespace. Hit semi-colon at the end of this line and VS will do the rest. (Ditto other classes.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/PerformanceLogger.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            _logFile = $"{logFolder}/BenchmarkResults_{DateTime.UtcNow.ToString("yyyy-MM-dd_HH-mm-ss-fff", CultureInfo.InvariantCulture)}.csv";
+        }
+
+        /// <summary>
+        /// Log collection of values in comma separated format.
+        /// </summary>
+        internal void Log(IEnumerable<String> values)
+        {
+            Log($"{string.Join(",", values)}");
```
jskeet, line 38: No need for an interpolated string - just `Log(string.Join(",", values));`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/PerformanceLogger.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            lock (_lockObj)
+            {
+                WriteMessage(message);
+            }
+        }
+
+        // Should be called within the lock
+        private void WriteMessage(string message) =>
+            File.AppendAllText(_logFile, $"{message}{Environment.NewLine}");
```
jskeet, line 54: Or potentially `File.AppendAllLines(_logFile, new[] { message })`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+using System.IO;
+using System.Threading;
+
+namespace Google.Cloud.Storage.V1.BenchmarkingTool
+{
+    public static class Program
+    {
+        private static readonly string s_projectId = TestEnvironment.GetTestProjectId(TestEnvironment.TestProjectEnvironmentVariable);
+
+        private static readonly string[] s_logHeaders = new string[] { "Op", "ObjectSize", "AppBufferSize", "LibBufferSize", "Crc32cEnabled", "MD5Enabled", "ApiName", "CpuTimeUs", "ElapsedTimeUs", "Status" };
```
jskeet, line 31: No need for `new string[]` here - just `private static readonly string[] s_logHeaders = { "Op", "ObjectSize", ... };`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+                var configuration = Configuration.FromCommandLineArguments(args);
+
+                var logger = new PerformanceLogger(configuration.OutputFolder);
+                logger.Log(s_logHeaders);
+
+                var client = StorageClient.Create();
+
+                s_stopwatch = Stopwatch.StartNew();
+
+                for (int i = 1; i <= configuration.NumOfSamples; ++i)
```
jskeet, line 50: I tend to think that `i++` is more idiomatic for loop iterations. (Obviously it doesn't matter which is used here, but it's good to be consistent across our codebase.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+                {
+                    Console.WriteLine($"Running W1R3 iteration number {i}..");
+                    RunW1R3Benchmark(configuration, client, logger);
+                }
+
+                s_stopwatch.Stop();
+            }
+            catch (ArgumentException e)
```
jskeet, line 58: I don't know how any of the console colour changes will work when running in Kokoro etc.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+
+        private static void DeleteBucket(StorageClient client, string bucketName)
+        {
+            try
+            {
+                client.DeleteBucket(bucketName, new DeleteBucketOptions { UserProject = null, DeleteObjects = true });
+            }
+            catch (GoogleApiException)
+            {
+                //Some tests fail to delete buckets due to object retention locks etc.
```
jskeet, line 103: Nit: add a space after `//` here and below.
(If we don't have a space in the test fixture code this was copied from, feel free to update that in the same PR...)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+                bytesLeft -= count;
+            }
+        }
+
+        private static void Write(Configuration configuration, StorageClient client, PerformanceLogger logger, string opName)
+        {
+            string elapsedTimeUs = "-";
+            bool isSucccess = false;
+
+            var logs = new string[10]; //logs array should be in same order as s_logHeaders.
```
jskeet, line 129: Nit: space after `//`

And I'd rename `logs` to something like `logEntryValues` or even `rowValues`. (They're not really logs themselves - they form a log entry.)

You might also want to change `10` to `s_logHeaders.Length` to avoid having to change things in multiple places.

Alterantively, simplify and clarify the code by creating the array with values to start with:

```csharp
// Values in the same order as s_logHeaders
string[] rowValues =
{
    opName,
    configuration.DataSizeInBytes.ToString(),
    "NA", // AppBufferSize,
    (options?.ChunkSize ?? ResumableUpload.DefaultChunkSize).ToString(),
    "TRUE", // Crc32cEnabled
    "NA", // MD5Enabled
    "JSON", // ApiName
    "-1", // CpuTimeUs,
    null, // ElapsedTimeUs (populated later)
    null // Status (populated later)
};
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+                logs[1] = configuration.DataSizeInBytes.ToString(CultureInfo.InvariantCulture);
+                logs[2] = "NA";
+                logs[3] = options != null ? options.ChunkSize.ToString() : ResumableUpload.DefaultChunkSize.ToString();
+                logs[4] = "TRUE";
+                logs[5] = "NA";
+                logs[6] = "JSON";
+                logs[7] = "-1";
+
+                Console.WriteLine("Uploading..");
+                using (var input = File.OpenRead(configuration.LocalFile))
```
jskeet, line 154: No need for a block here - just `using var input = File.OpenRead(configuration.LocalFile);`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+
+namespace Google.Cloud.Storage.V1.BenchmarkingTool
+{
+    public static class Program
+    {
+        private static readonly string s_projectId = TestEnvironment.GetTestProjectId(TestEnvironment.TestProjectEnvironmentVariable);
+
+        private static readonly string[] s_logHeaders = new string[] { "Op", "ObjectSize", "AppBufferSize", "LibBufferSize", "Crc32cEnabled", "MD5Enabled", "ApiName", "CpuTimeUs", "ElapsedTimeUs", "Status" };
+
+        private static Stopwatch s_stopwatch;
```
jskeet, line 33: I think the code would be simpler if we used a different Stopwatch per operation. They're cheap to create.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+                int count = (int) Math.Min(bytesLeft, buffer.Length);
+                output.Write(buffer, 0, count);
+                bytesLeft -= count;
+            }
+        }
+
+        private static void Write(Configuration configuration, StorageClient client, PerformanceLogger logger, string opName)
+        {
+            string elapsedTimeUs = "-";
+            bool isSucccess = false;
```
jskeet, line 127: Suggest renaming to just "success" (note two c's rather than three, as well as removing the "is").

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+            {
+                Console.WriteLine($"Error while upload : {e.Message}");
+            }
+            finally
+            {
+                logs[8] = elapsedTimeUs;
+                logs[9] = isSucccess ? "OK" : "FAIL";
+                logger.Log(logs);
+
+                if (configuration != null & File.Exists(configuration.LocalFile))
```
jskeet, line 177: I would just call `File.Delete(configuration.LocalFile)` unconditionally here. `configuration` should never be null, and it's fine to call `File.Delete` for a file that doesn't exist (IIRC).

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+                {
+                    File.Delete(configuration.LocalFile);
+                }
+            }
+        }
+
+        private static void Read(Configuration configuration, StorageClient client, PerformanceLogger logger, string opName)
```
jskeet, line 184: Same comments as before, basically - but let's get the write test into a good state, then make this one look the same.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,244 @@
...
+            while (bytesLeft > 0)
+            {
+                random.NextBytes(buffer);
+                int count = (int) Math.Min(bytesLeft, buffer.Length);
+                output.Write(buffer, 0, count);
+                bytesLeft -= count;
+            }
+        }
+
+        private static void Write(Configuration configuration, StorageClient client, PerformanceLogger logger, string opName)
```
jskeet, line 124: I think this would be clearer as something like `RunWriteTest`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
anuragsrivstv: Yes, maybe it'll not work in Kokoro. It's just for someone if they are running the tool on their desktop, but if it'll not be a very frequent case then this part of code can be removed. What do you suggest ?

jskeet: I'd just remove this catch block entirely. It'll be handled by the exception below, and the stack trace won't be terribly large anyway.

anuragsrivstv:
> > Merge branch 'googleapis:main' into perf-benchmarking
> 
> Please don't merge _into_ your feature branch - instead, fetch the main branch and then rebase your feature branch from it.
> 
> For example:
> 
> ```shell
> $ git checkout main
> $ git pull upstream main
> $ git checkout perf-benchmarking
> $ git rebase main
> $ git push -f
> ```

Done.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Configuration.cs
```diff
@@ -0,0 +1,85 @@
...
+    internal int NumOfSamples { get; }
+    internal int DataSizeInBytes { get; }
+    internal string LocalFile { get; }
+    internal string OutputFolder { get; }
+    internal int UploadChunkSizeInBytes { get; }
+    internal int DownloadChunkSizeInBytes { get; }
+    internal string ObjectName => "obj_random_data";
+    internal string Bucket => _bucketName;
+
+    private readonly string _bucketName = IdGenerator.FromDateTime(prefix: "benchmark-");
```
jskeet, line 35: There's no need for this separate private field. It could just be:

```csharp
internal string Bucket { get; } = IdGenerator.FromDateTime(prefix: "benchmark-");
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Configuration.cs
```diff
@@ -0,0 +1,85 @@
...
+        {
+            throw new ArgumentException("Error: File size should be positive number.");
+        }
+
+        if (!Directory.Exists(args[2])) // args[2] is the folder path where log will be generated.
+        {
+            throw new ArgumentException("Error: Please provide a valid output folder path.");
+        }
+
+        // args[2] and args[4] are the chunk sizes in which data can be uploaded or downloaded respectively.
```
jskeet, line 74: args[3] instead of args[2]

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,208 @@
...
+
+        try
+        {
+            CreateRandomFile(configuration);
+            long totalElapsedMicroseconds = 0;
+            Console.WriteLine("Uploading..");
+            using var input = File.OpenRead(configuration.LocalFile);
+            var stopwatch = Stopwatch.StartNew();
+            client.UploadObject(configuration.Bucket, configuration.ObjectName, "application/binary", input, options);
+            totalElapsedMicroseconds = stopwatch.Elapsed.Ticks / TicksPerMicrosecond;
```
jskeet, line 140: I don't think we need this local variable at all - or if we want it for clarity, we could at least declare and initialize it at the same place.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
```diff
@@ -0,0 +1,208 @@
...
+using System.IO;
+using System.Threading;
+
+namespace Google.Cloud.Storage.V1.BenchmarkingTool;
+
+public static class Program
+{
+    private static readonly string s_projectId = TestEnvironment.GetTestProjectId(TestEnvironment.TestProjectEnvironmentVariable);
+    private static readonly string[] s_logHeaders = { "Op", "ObjectSize", "AppBufferSize", "LibBufferSize", "Crc32cEnabled", "MD5Enabled", "ApiName", "CpuTimeUs", "ElapsedTimeUs", "Status" };
+    private const long TicksPerMicrosecond = 10;
```
jskeet, line 31: It might be worth noting that these are TimeSpan/DateTime ticks, not Stopwatch ticks. (Later you've got code that uses `stopwatch.Elapsed.Ticks`, which is fine - but it *looks* like you could just change it to `stopwatch.ElapsedTicks`, giving a totally different answer!

jskeet (approved):
A few small suggestions, but nothing major - feel free to merge after addressing the ones you want to address.

(I would suggest amending the existing commit rather than adding another commit.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Configuration.cs
anuragsrivstv: I believe it'll then generate different bucket names every time getter of property ''Bucket' is called because IdGenerator.FromDateTime() always generate new ID based on current date time and we'll need to keep bucket name consistent .

jskeet: No, it won't. There's a big difference between these two:

```csharp
// Auto-implemented property with a backing field, but initialized at construction time
internal string Bucket { get; } = IdGenerator.FromDateTime(prefix: "benchmark-");

// Expression-bodied property with no backing field
internal string Bucket => IdGenerator.FromDateTime(prefix: "benchmark-");
```

anuragsrivstv: Okay, Got it! Thanks for explaining.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Program.cs
anuragsrivstv: Hi Jon, Can you please elaborate what are you suggesting ? 

Later in the code should we de something like:-
_var elapsedTimeSpan = stopwatch.Elapsed;
totalElapsedMicroseconds = elapsedTimeSpan.Ticks / TicksPerMicrosecond;_

or are you suggesting something else ?

jskeet: I'm only suggesting that you add a comment to this `const` declaration - I'd probably call it `TimeSpanTicksPerMicrosecond` or something like that. Because it's not the number of *stopwatch* ticks per microsecond, and we're using it "near" a stopwatch, so it's worth being super clear.

PR #8890, opened 2022-07-15
feat: Add support for BigQuery job metadata deletion.

amanda-tarafa:


jskeet (approved):

Rishabh-V (approved):
Hi @amanda-tarafa. Thanks for looping me in so that I can get the understanding. Everything looks good. Just couple of questions for my understanding. 
1) I don't see any changes in DeleteDatasetOptions.cs (just the first line header update). Is it fine?
2) BigQueryClient.JobCrud has auto-generated code. Is this file generated from JobCrud.xml?

amanda-tarafa:
> 1. I don't see any changes in DeleteDatasetOptions.cs (just the first line header update). Is it fine?

Hmmm, that first line is being changed in almost all files, I'm guessing and EOL of similar character as there's no "visible" difference. I'll leave the change in all the files that have other changes but revert it in this one.

> 2. BigQueryClient.JobCrud has auto-generated code. Is this file generated from JobCrud.xml?

Yes, take a look at the tool in https://github.com/googleapis/google-cloud-dotnet/tree/main/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.GenerateOverloads, which generates the overload wrappers for the Apiary library (Google.Apis.Bigquery.v2) client methods. There's still some copy pasting in tests that we need to do, but this tool saves a lot of mechanical work.

PR #8891, opened 2022-07-15
feat:Add support for exactly once subscription

Rishabh-V:
Hi @jskeet,

This is the first draft PR to support exactly once subscription. It covers the following:

1. Detect if subscription is exactly once or not.
2.  New defaults for exactly once subscriptions for Ack and ModifyAckDeadline RPCs.
3. A new AckError class to keep temporary and permanent failures.
4. Retry the temporary failures for a duration of 10 minutes for Acks/Nacks using exponential backoff.
5. Retry the temporary failures for a duration of 10 seconds or 3 attempts using exponential backoff. 
6. Bubble up permanent failures for Acks/Nacks.
7.  TODO: Permanent errors for extends needs to be logged.

Seeks your special review, inputs and feedback on the following:

1. Thread safety of code (in case we need it and I have missed or I have added lock where we don't need :) )
2. Error propagation - Currently, I am just populating exception with error messages and creating AggregateException. Do we need a custom exception, something like AcknowledgementException?
3. Retry implementation - Currently, I am maintaining a concurrent dictionary with Id and RetryInfo which contains the first time of failure and current backoff. Only Ids that are present in this concurrent dictionary are requeued (and hence retried) after the calculated backoff. I keep the dictionary updated with backoff of Id and if Id should be removed (based on id becoming successful or permanent failure).

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2017, Google Inc. All rights reserved.
+// Copyright 2017, Google Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
+using static Google.Cloud.PubSub.V1.StreamingPullResponse.Types;
 
 namespace Google.Cloud.PubSub.V1.Tests
 {
@@ -134,7 +135,7 @@ namespace Google.Cloud.PubSub.V1.Tests
 
             private class En : IAsyncStreamReader<StreamingPullResponse>
             {
-                public En(IEnumerable<ServerAction> msgs, IScheduler scheduler, TaskHelper taskHelper, IClock clock, bool useMsgAsId, CancellationToken? ct)
+                public En(IEnumerable<ServerAction> msgs, IScheduler scheduler, TaskHelper taskHelper, IClock clock, bool useMsgAsId, CancellationToken? ct, bool isExactlyOnce)
                 {
                     _msgsEn = msgs.Select((x, i) => (i, x)).GetEnumerator();
                     _scheduler = scheduler;
@@ -142,12 +143,14 @@ namespace Google.Cloud.PubSub.V1.Tests
                     _clock = clock;
                     _useMsgAsId = useMsgAsId;
                     _ct = ct ?? CancellationToken.None;
+                    _isExactlyOnce = isExactlyOnce;
                 }
 
                 private readonly IScheduler _scheduler;
                 private readonly TaskHelper _taskHelper;
                 private readonly IClock _clock;
                 private readonly bool _useMsgAsId;
+                private readonly bool _isExactlyOnce;
                 private readonly CancellationToken _ct;
 
                 private readonly IEnumerator<(int Index, ServerAction Action)> _msgsEn;
@@ -189,7 +192,8 @@ namespace Google.Cloud.PubSub.V1.Tests
                             ReceivedMessages = {
                                 _msgsEn.Current.Action.Msgs.Select((s, i) =>
                                     MakeReceivedMessage(_useMsgAsId ? s : MakeMsgId(_msgsEn.Current.Index, i), s, _msgsEn.Current.Action.DeliveryAttempt))
-                            }
+                            },
+                            SubscriptionProperties = _isExactlyOnce ? new SubscriptionProperties { ExactlyOnceDeliveryEnabled = true} : null
                         };
                     }
                 }
@@ -203,16 +207,17 @@ namespace Google.Cloud.PubSub.V1.Tests
             {
                 public PullStream(TimeSpan writeAsyncPreDelay,
                     IEnumerable<ServerAction> msgs, List<DateTime> writeCompletes, List<DateTime> streamPings,
-                    IScheduler scheduler, IClock clock, TaskHelper taskHelper, bool useMsgAsId, CancellationToken? ct)
+                    IScheduler scheduler, IClock clock, TaskHelper taskHelper, bool useMsgAsId, CancellationToken? ct, bool isExactlyOnce)
                 {
                     _taskHelper = taskHelper;
                     _scheduler = scheduler;
                     _writeAsyncPreDelay = writeAsyncPreDelay; // delay within the WriteAsync() method. Simulating network or server slowness.
-                    var responseStream = new En(msgs, scheduler, taskHelper, clock, useMsgAsId, ct);
+                    var responseStream = new En(msgs, scheduler, taskHelper, clock, useMsgAsId, ct, isExactlyOnce);
                     _call = new AsyncDuplexStreamingCall<StreamingPullRequest, StreamingPullResponse>(null, responseStream, Task.FromResult(new Metadata()), null, null, null);
                     _clock = clock;
                     _writeCompletes = writeCompletes;
                     _streamPings = streamPings;
+                    _isExactlyOnce = isExactlyOnce;
                 }
 
                 private readonly object _lock = new object();
@@ -223,6 +228,7 @@ namespace Google.Cloud.PubSub.V1.Tests
                 private readonly IClock _clock;
                 private readonly List<DateTime> _writeCompletes;
                 private readonly List<DateTime> _streamPings;
+                private readonly bool _isExactlyOnce;
 
                 public override AsyncDuplexStreamingCall<StreamingPullRequest, StreamingPullResponse> GrpcCall => _call;
 
@@ -255,7 +261,7 @@ namespace Google.Cloud.PubSub.V1.Tests
 
             public FakeSubscriberServiceApiClient(
                 IEnumerator<IEnumerable<ServerAction>> msgsEn, IScheduler scheduler, IClock clock,
-                TaskHelper taskHelper, TimeSpan writeAsyncPreDelay, bool useMsgAsId, AckModifyAckDeadlineAction ackModifyAckDeadlineAction)
+                TaskHelper taskHelper, TimeSpan writeAsyncPreDelay, bool useMsgAsId, AckModifyAckDeadlineAction ackModifyAckDeadlineAction, bool isExactlyOnce)
             {
                 _msgsEn = msgsEn;
                 _scheduler = scheduler;
@@ -265,6 +271,7 @@ namespace Google.Cloud.PubSub.V1.Tests
                 _useMsgAsId = useMsgAsId;
                 _ackModifyAckDeadlineAction = ackModifyAckDeadlineAction;
                 _numberOfAckModifyAckDeadlineFailures = 0;
+                _isExactlyOnce = isExactlyOnce;
             }
 
             private readonly object _lock = new object();
@@ -275,6 +282,7 @@ namespace Google.Cloud.PubSub.V1.Tests
             private readonly TimeSpan _writeAsyncPreDelay; // Simulates slow network or server
             private readonly bool _useMsgAsId;
             private readonly AckModifyAckDeadlineAction _ackModifyAckDeadlineAction; // Simulates Ack ModifyAckDeadline errors.
+            private readonly bool _isExactlyOnce;
             private int _numberOfAckModifyAckDeadlineFailures;
 
             private readonly List<TimedId> _extends = new List<TimedId>();
@@ -300,7 +308,7 @@ namespace Google.Cloud.PubSub.V1.Tests
                     }
                     var msgs = _msgsEn.Current;
                     return new PullStream(_writeAsyncPreDelay, msgs, _writeCompletes, _streamPings,
-                        _scheduler, _clock, _taskHelper, _useMsgAsId, callSettings?.CancellationToken);
+                        _scheduler, _clock, _taskHelper, _useMsgAsId, callSettings?.CancellationToken, _isExactlyOnce);
                 }
             }
 
@@ -369,14 +377,14 @@ namespace Google.Cloud.PubSub.V1.Tests
                 TimeSpan? ackDeadline = null, TimeSpan? ackExtendWindow = null,
                 int? flowMaxElements = null, int? flowMaxBytes = null,
                 int clientCount = 1, int threadCount = 1, TimeSpan? writeAsyncPreDelay = null,
-                bool useMsgAsId = false, AckModifyAckDeadlineAction ackModifyAckDeadlineAction = null)
+                bool useMsgAsId = false, AckModifyAckDeadlineAction ackModifyAckDeadlineAction = null, bool isExactlyOnce = false)
             {
                 var scheduler = new TestScheduler(threadCount: threadCount);
                 TaskHelper taskHelper = scheduler.TaskHelper;
                 List<DateTime> clientShutdowns = new List<DateTime>();
                 var msgEn = msgs.GetEnumerator();
                 var clients = Enumerable.Range(0, clientCount)
-                    .Select(_ => new FakeSubscriberServiceApiClient(msgEn, scheduler, scheduler.Clock, taskHelper, writeAsyncPreDelay ?? TimeSpan.Zero, useMsgAsId, ackModifyAckDeadlineAction))
+                    .Select(_ => new FakeSubscriberServiceApiClient(msgEn, scheduler, scheduler.Clock, taskHelper, writeAsyncPreDelay ?? TimeSpan.Zero, useMsgAsId, ackModifyAckDeadlineAction, isExactlyOnce))
                     .ToList();
                 var settings = new SubscriberClient.Settings
                 {
@@ -1109,6 +1117,118 @@ namespace Google.Cloud.PubSub.V1.Tests
                 Exception ex = await fake.TaskHelper.ConfigureAwaitHideErrors(() => fake.Subscriber.StopAsync(CancellationToken.None));
                 Assert.Equal(exception, ex.AllExceptions().FirstOrDefault());
                 Assert.Equal(1, fake.ClientShutdowns.Count);
+            });
+        }
+
+        [Theory, CombinatorialData]
+        public void ExactlyOnce_TemporaryFault([CombinatorialValues(true, false, null)] bool? ackNackOrExtends)
+        {
+            var msgs = new[]
+            {
+                ServerAction.Data(TimeSpan.Zero, new[] { "1" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "2" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "3" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "4" }),
+                ServerAction.Inf()
+            };
+
+            // Any temporary failure in ack/nack/extend request should not be thrown to the client. It should be retried.
+            var exception = new RpcException(new Status(StatusCode.DeadlineExceeded, ""), "");
+            // If ackNackOrExtends is null, then it is extend request. ModifyAcknowledgeDeadline RPC will throw the supplied exception.
+            // If ackNackOrExtends is true, then it is acknowledge request. Acknowledge RPC will throw the supplied exception.
+            // If ackNackOrExtends is false, then it is nack request. ModifyAcknowledgeDeadline RPC will throw the supplied exception.
+            var ackModifyAckDeadlineAction =
+                ackNackOrExtends == null ? AckModifyAckDeadlineAction.Data(exception, numberOfFailures: 10, onAck: false, onModifyAckDeadline: true)
+                : ackNackOrExtends.Value ? AckModifyAckDeadlineAction.BadAck(exception, numberOfFailures: 10)
+                : AckModifyAckDeadlineAction.BadModifyAckDeadline(exception, numberOfFailures: 10);
+
+            using var fake = Fake.Create(new[] { msgs }, ackModifyAckDeadlineAction: ackModifyAckDeadlineAction, isExactlyOnce: true);
+            fake.Scheduler.Run(async () =>
+            {
+                var handledMsgs = new List<string>();
+                var doneTask = fake.Subscriber.StartAsync(async (msg, ct) =>
+                {
+                    handledMsgs.Locked(() => handledMsgs.Add(msg.Data.ToStringUtf8()));
+                    if (ackNackOrExtends == null)
+                    {
+                        // Add delay greater than ackDeadline to simulate extends.
+                        await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(70), ct));
+                    }
+
+                    return await Task.FromResult(ackNackOrExtends == false ? SubscriberClient.Reply.Nack : SubscriberClient.Reply.Ack);
+                });
+                await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(100), CancellationToken.None));
+                await fake.TaskHelper.ConfigureAwait(fake.Subscriber.StopAsync(CancellationToken.None));
+                // Despite temporary failures, all 4 messages should be handled.
+                Assert.Equal(new[] { "1", "2", "3", "4" }, handledMsgs);
+            });
+        }
+
+        [Theory, CombinatorialData]
+        public void ExactlyOnce_AckNack_PermanentFault([CombinatorialValues(true, false)] bool ackOrNack)
+        {
+            var msgs = new[]
+            {
+                ServerAction.Data(TimeSpan.Zero, new[] { "1" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "2" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "3" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "4" }),
+                ServerAction.Inf()
+            };
+
+            // Any permanent failure in ack/nack request should be thrown to the client.
+            var exception = new RpcException(new Status(StatusCode.FailedPrecondition, ""), "");
+            // If ackNackOrExtends is true, then it is acknowledge request. Acknowledge RPC will throw the supplied exception.
+            // If ackNackOrExtends is false, then it is nack request. ModifyAcknowledgeDeadline RPC will throw the supplied exception.
+            // Since exception is thrown, the client will shutdown, so exception count is specified as 1.
+            var ackModifyAckDeadlineAction =
+                ackOrNack ? AckModifyAckDeadlineAction.BadAck(exception, numberOfFailures: 10)
+                : AckModifyAckDeadlineAction.BadModifyAckDeadline(exception, numberOfFailures: 10);
+
+            using var fake = Fake.Create(new[] { msgs }, ackModifyAckDeadlineAction: ackModifyAckDeadlineAction, isExactlyOnce: true);
+            fake.Scheduler.Run(async () =>
+            {
+                var doneTask = fake.Subscriber.StartAsync(async (msg, ct) =>
+                await Task.FromResult(ackOrNack ? SubscriberClient.Reply.Ack : SubscriberClient.Reply.Nack));
+                await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(100), CancellationToken.None));
+                Exception ex = await fake.TaskHelper.ConfigureAwaitHideErrors(() => fake.Subscriber.StopAsync(CancellationToken.None));
+                // For Ack/Nack, exception should be thrown.
+                Assert.True(ex is not null);
+                Assert.IsType<AcknowledgementException>(ex);
+                Assert.Equal(1, fake.ClientShutdowns.Count);
+            });
+        }
+
+        [Fact]
+        public void ExactlyOnce_Extends_PermanentFault()
+        {
+            var msgs = new[]
+            {
+                ServerAction.Data(TimeSpan.Zero, new[] { "1" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "2" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "3" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "4" }),
+                ServerAction.Inf()
+            };
+
+            // Any permanent failure in extend request should not be thrown to the client.
+            var exception = new RpcException(new Status(StatusCode.FailedPrecondition, ""), "");
+            // This is extend request. ModifyAcknowledgeDeadline RPC will throw the supplied exception.
+            var ackModifyAckDeadlineAction = AckModifyAckDeadlineAction.BadModifyAckDeadline(exception, numberOfFailures: 10);
+
+            using var fake = Fake.Create(new[] { msgs }, ackModifyAckDeadlineAction: ackModifyAckDeadlineAction, isExactlyOnce: true);
+            fake.Scheduler.Run(async () =>
+            {
+                var doneTask = fake.Subscriber.StartAsync(async (msg, ct) =>
+                {
+                    // Add delay greater than ackDeadline to simulate the extends.
+                    await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(70), ct));
+                    return await Task.FromResult(SubscriberClient.Reply.Ack);
+                });
+                await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(100), CancellationToken.None));
+                Exception ex = await fake.TaskHelper.ConfigureAwaitHideErrors(() => fake.Subscriber.StopAsync(CancellationToken.None));
+                // For extends, exception shouldn't be thrown.
+                Assert.True(ex is null);
             });
         }
     }
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
@@ -1,0 +1,152 @@
+// Copyright 2022, Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax;
+using Google.Api.Gax.Grpc;
+using Grpc.Core;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Google.Cloud.PubSub.V1;
+
+/// <summary>
+/// This class contains the temporary and/or permanent failures from Acknowledge and ModifyAckDeadline RPCs for exactly once subscriptions.
+/// </summary>
+internal sealed class AckError
+{
+    /// <summary>
+    /// True if there are any permanent failures else false if <see cref="PermanentFailureIds"/> is empty.
+    /// </summary>
+    internal bool HasPermanentFailures => PermanentFailureIds.Any();
+
+    /// <summary>
+    /// Gets the AckIds/ExtendIds/NackIds with temporary failures.
+    /// This collection will be populated with AckIds/ExtendIds/NackIds only if the failure is temporary,
+    /// else this collection will be empty (but not null).
+    /// </summary>
+    internal IEnumerable<string> TemporaryFailureIds => TemporaryFailures.Select(kvp => kvp.Key);
+
+    /// <summary>
+    /// Gets the temporary failures.
+    /// This collection will be populated with AckIds/NackIds/ExtendIds with their corresponding failures only if the failure is temporary,
+    /// else this collection will be empty (but not null).
+    /// </summary>
+    internal IEnumerable<KeyValuePair<string, string>> TemporaryFailures { get; }
+
+    /// <summary>
+    /// Gets the AckIds/ExtendIds/NackIds with permanent failures.
+    /// This collection will be populated with AckIds/ExtendIds/NackIds only if the failure is permanent,
+    /// else this collection will be empty (but not null).
+    /// </summary>
+    internal IEnumerable<string> PermanentFailureIds => PermanentFailures.Select(kvp => kvp.Key);
+
+    /// <summary>
+    /// Gets the permanent failures.
+    /// This collection will be populated with AckIds/NackIds/ExtendIds with their corresponding failures only if the failure is permanent,
+    /// else this collection will be empty (but not null).
+    /// </summary>
+    internal IEnumerable<KeyValuePair<string, string>> PermanentFailures { get; }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AckError"/> class.
+    /// </summary>
+    /// <param name="temporaryFailures">The collection of temporary failed AckId/NackId/ExtendId as key and their corresponding failure reason as value.</param>
+    /// <param name="permanentFailures">The collection of permanent failed AckId/NackId/ExtendId as key and their corresponding failure reason as value.</param>
+    internal AckError(IEnumerable<KeyValuePair<string, string>> temporaryFailures, IEnumerable<KeyValuePair<string, string>> permanentFailures)
+    {
+        TemporaryFailures = temporaryFailures;
+        PermanentFailures = permanentFailures;
+    }
+
+    /// <summary>
+    /// Gets the permanent errors as the sequence of <see cref="AcknowledgementException"/>
+    /// from the current instance of <seealso cref="AckError"/>.
+    /// </summary>
+    /// <remarks>
+    /// This method will always return a sequence of <see cref="AcknowledgementException"/>. This sequence will be empty (but not null) if there are no permanent failures
+    /// in the current instance of <seealso cref="AckError"/>.
+    /// </remarks>
+    /// <returns>The sequence of <see cref="AcknowledgementException"/> if there are permanent failures, empty otherwise.</returns>
+    internal IEnumerable<AcknowledgementException> GetPermanentExceptions() =>
+        HasPermanentFailures
+            ? PermanentFailures.Select(AcknowledgementException.FromIdErrorPair).ToList()
+            : Enumerable.Empty<AcknowledgementException>();
+
+    /// <summary>
+    /// Gets the <see cref="AckError"/> from the <see cref="RpcException"/> while calling Acknowledge or ModifyAckDeadline RPC.
+    /// </summary>
+    /// <param name="rpcException">The underlying <see cref="RpcException"/>.</param>
+    /// <param name="ids">The ids for which the Acknowledge/ModifyAckDeadline RPC was called.</param>
+    /// <returns>The populated instance of <see cref="AckError"/> containing temporary and permanent failures.</returns>
+    internal static AckError ForRpcException(RpcException rpcException, IEnumerable<string> ids)
+    {
+        GaxPreconditions.CheckNotNull(rpcException, nameof(rpcException));
+        GaxPreconditions.CheckNotNull(ids, nameof(ids));
+
+        // Check the gRPC StatusCode. For certain StatusCodes, we need to retry all ids.            
+        bool retryAll = ShouldRetryAll(rpcException);
+        if (retryAll)
+        {
+            // Return all ids in temporary failures and empty permanent failures, so that the caller can retry the whole temporary batch.
+            return new AckError(ids.ToDictionary(key => key, value => rpcException.Status.StatusCode.ToString()), Enumerable.Empty<KeyValuePair<string, string>>());
+        }
+
+        // We can have complete or partial failure. We classify the error as temporary or permanent
+        // based on the error message starting from TRANSIENT_FAILURE_ or PERMANENT_FAILURE_ (which is ugly but that's the way it is)
+        // and return:
+        // the temporary errors in temporaryErrors collection and
+        // the permanent errors in permanentErrors collection.
+        // For exactly once, the metadata of the rpcException contains the error details. A sample metadata is shown below:
+        // {
+        //    "ack_id_1": "PERMANENT_FAILURE_INVALID_ACK_ID",
+        //    "ack_id_2": "PERMANENT_FAILURE_INVALID_ACK_ID",
+        //    "ack_id_3": "TRANSIENT_FAILURE_UNORDERED_ACK_ID"
+        // }
+        var metadata = rpcException.GetErrorInfo()?.Metadata ?? Enumerable.Empty<KeyValuePair<string, string>>();
+        var temporaryErrors = metadata.Where(j => j.Value.StartsWith("TRANSIENT_FAILURE_", StringComparison.Ordinal));
+        var permanentErrors = metadata.Where(j => j.Value.StartsWith("PERMANENT_FAILURE_", StringComparison.Ordinal));
+
+        if (temporaryErrors.Any() || permanentErrors.Any())
+        {
+            // If temporary or permanent error messages are present, it means the entire request hasn't failed with a permanent error.
+            return new AckError(temporaryErrors, permanentErrors);
+        }
+        else
+        {
+            // Now, if temporary or permanent error messages are *not* present, it means the entire request has failed with a permanent error.
+            // It may be due to a gRPC error like a failed precondition or a permission denied error or other permanent error.
+            return new AckError(temporaryErrors, ids.ToDictionary(key => key, value => rpcException.Status.StatusCode.ToString()));
+        }
+    }
+
+    /// <summary>
+    /// Checks if the gRPC StatusCode of <see cref="RpcException"/> is retry-able and
+    /// the entire batch of AckIds/NackIds/ExtendIds should be retried for Acknowledge/ModifyAckDeadline RPC.
+    /// </summary>
+    /// <param name="e">The <see cref="RpcException"/>.</param>
+    /// <returns><c>True</c> if all AckIds/NackIds/ExtendIds should be retried, else <c>false</c>.</returns>
+    /// <remarks>
+    /// This API should be used for exactly-once subscriptions only.
+    /// </remarks>
+    internal static bool ShouldRetryAll(RpcException e) => e.Status.StatusCode switch
+    {
+        StatusCode.DeadlineExceeded => true,
+        StatusCode.ResourceExhausted => true,
+        StatusCode.Aborted => true,
+        StatusCode.Internal => true,
+        StatusCode.Unavailable => true,
+        _ => false,
+    };
+}
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Extensions.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Extensions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2017, Google Inc. All rights reserved.
+// Copyright 2017, Google Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
@@ -17,6 +17,7 @@ using Google.Api.Gax.Grpc;
 using Google.Cloud.PubSub.V1.Tasks;
 using Grpc.Core;
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
@@ -44,12 +45,6 @@ namespace Google.Cloud.PubSub.V1
     /// </remarks>
     public abstract class SubscriberClient
     {
-        private static readonly GrpcChannelOptions s_unlimitedSendReceiveChannelOptions = GrpcChannelOptions.Empty
-            .WithMaxReceiveMessageSize(int.MaxValue)
-            .WithMaxSendMessageSize(int.MaxValue)
-            // Set max metadata size to 4 MB i.e., 4194304 bytes.
-            .WithCustomOption("grpc.max_metadata_size", 4194304);
-
         /// <summary>
         /// Reply from a message handler; whether to <see cref="Ack"/>
         /// or <see cref="Nack"/> the message to the server. 
@@ -84,6 +79,7 @@ namespace Google.Cloud.PubSub.V1
                 AckDeadline = other.AckDeadline;
                 AckExtensionWindow = other.AckExtensionWindow;
                 Scheduler = other.Scheduler;
+                Clock = other.Clock;
                 MaxTotalAckExtension = other.MaxTotalAckExtension;
             }
 
@@ -125,6 +121,16 @@ namespace Google.Cloud.PubSub.V1
             /// This is usually only used for testing.
             /// </summary>
             public IScheduler Scheduler { get; set; }
+
+            /// <summary>
+            /// The <see cref="IClock"/> used for timeouts and retries.
+            /// If <c>null</c>, the default <see cref="SystemClock"/> is used.
+            /// This is usually only used for testing.
+            /// </summary>
+            /// <remarks>
+            /// This is used only in exactly once flow as we need to retry temporary failures.
+            /// </remarks>
+            public IClock Clock { get; set; }
 
             internal void Validate()
             {
@@ -262,6 +268,11 @@ namespace Google.Cloud.PubSub.V1
         /// The minimum message ACKnowledgement extension window of 50 milliseconds.
         /// </summary>
         public static TimeSpan MinimumAckExtensionWindow { get; } = TimeSpan.FromMilliseconds(50);
+
+        /// <summary>
+        /// The minimum message acknowledgement extension window of 60 seconds for exactly once subscriptions.
+        /// </summary>
+        public static TimeSpan MinimumAckExtensionWindowForExactlyOnce { get; } = TimeSpan.FromSeconds(60);
 
         /// <summary>
         /// The default message ACKnowledgement extension window of 15 seconds.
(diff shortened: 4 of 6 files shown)


jskeet:
Will review on Tuesday - I'm out on Monday, and between other meetings etc I don't have time to give this the dedicated time it'll need to review it properly today.

Rishabh-V:
> Will review on Tuesday - I'm out on Monday, and between other meetings etc I don't have time to give this the dedicated time it'll need to review it properly today.

Sure Jon. Thank you.

Rishabh-V:
Hi @jskeet - Just for your information. I was having a look today and identified couple of issues. I am making the fix and should have them checked-in in couple of hours.

Rishabh-V:
> Thanks for the heads-up - I'll wait until I hear back before reviewing. (I'm working slowly today anyway as it's so hot in the UK...)

Thanks Jon. I have now pushed the draft changes. 

Need your review, inputs and suggestions for:
1) Keeping the code thread-safe in correct way. I suspect I may have missed it in a few places in this PR.
2) Retry logic implementation.
3) Error propagation (need to do logging as well- have added a TODO)

Thank you.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
                     _pushInFlight += acks.Count;
                     _concurrentPushCount += 1;
                     Task ackTask = _client.AcknowledgeAsync(_subscriptionName, acks, _hardStopCts.Token);
-                    Add(ackTask, Next(false, () => HandleAckResponse(ackTask, acks, null, null)));
+                    Add(ackTask, Next(false, _exactlyOnceDeliveryEnabled ? () => HandleAckResponseForExactlyOnce(ackTask, acks, null, null) : () => HandleAckResponse(ackTask, acks, null, null)));
```
jskeet, line 1380: Two options for avoiding the redundancy here:

1. Declare `Action<Task, List<string>, List<string>, List<TimedId>> ackHandler = _exactlyOnceDeliveryEnabled ? HandleAckResponseForExactlyOnce : HandleAckResponse;` just above this (line 1375), and then use `ackHandler` here and in the two uses below.
2. Just change `HandleAckResponse` to split off into HandleAckResponseForExactlyOnce if _exactlyOnceDeliveryEnabled is true. Note that this would (I guess) potentially act differently as the choice about which path is taken would be made after the acknowledgement has been received rather than here. I don't know whether that matters.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
...
                 {
                     _pushInFlight += nacks.Count;
                     _concurrentPushCount += 1;
                     Task nackTask = _client.ModifyAckDeadlineAsync(_subscriptionName, nacks, 0, _hardStopCts.Token);
-                    Add(nackTask, Next(false, () => HandleAckResponse(nackTask, null, nacks, null)));
+                    Add(nackTask, Next(false, _exactlyOnceDeliveryEnabled ? () => HandleAckResponseForExactlyOnce(nackTask, null, nacks, null) : () => HandleAckResponse(nackTask, null, nacks, null)));
+                }
+            }
+
+            private void Retry(IEnumerable<string> temporaryFailures, Action<IEnumerable<string>> actionToRetry, int durationInSeconds = 600)
```
jskeet, line 1398: Please rename to clarify what this is going to retry - and add a method comment for even more detail.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
...
+                    // Get retryInfo for the first retryableError from the _retryableIds dictionary.
+                    if (!_retryableIds.TryGetValue(retryableErrors.First(), out var retryInfo))
+                    {
+                        // Nothing to retry.
+                        return;
+                    }
+
+                    var backoff = s_ackBackoff.NextBackoff(retryInfo.CurrentBackoff ?? TimeSpan.Zero);
+                    // If we have errors to retry and we are not past the specified duration, then retry.
+                    if (DateTime.UtcNow.Subtract(retryInfo.FirstTimeOfFailureInUtc).Add(backoff) <= TimeSpan.FromSeconds(durationInSeconds))
```
jskeet, line 1429: I would change this to:

```csharp
if (DateTime.UtcNow <= retryInfo.FirstTimeOfFailureInUtc + TimeSpan.FromSeconds(durationInSeconds) - backoff)
```

... assuming that's what is meant. Possibly refactor in terms of a deadline?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
...
+                    {
+                        // Update backoff of retryableErrors in _retryableIds dictionary.
+                        UpdateBackoff(retryableErrors, backoff);
```
jskeet, line 1432: This method name sounds like it's going to change backoff, but it's really updating the retryable errors, right? Maybe rename to UpdateAllRetryableErrorBackoff? Possibly make it a static method in RetryableError?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
...
+                }
+            }
+
+            private void Retry(IEnumerable<string> temporaryFailures, Action<IEnumerable<string>> actionToRetry, int durationInSeconds = 600)
+            {
+                // TemporaryFailures may be empty but will never be null, so null check is not needed.
+                // By design, only items in temporaryFailures that exist in _retyrableIds will be retried.
+                // If id is not in _retryableIds dictionary, it is either already succeeded or not retryable.
+                // Failure may be non-retryable due to exceeded time duration or that failure may have become permanent.
+                if (temporaryFailures.Any())
```
jskeet, line 1404: I'd consider inverting the logic: if there aren't any temporary failures, return.
Likewise later on, if we've reached the deadline, return. That keeps nesting to a minimum, and makes the "we're done" cases really clear.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
...
+                            // We don't need recursion as the handling of Ack response will take care of retry and _retryableIds
+                            // dictionary will ensure that we don't retry after 10 minutes.
+                            actionToRetry(retryableErrors);
+                            StartPush();
+                        }));
+                    }
+                }
+            }
+
+            // TODO: Possible refactor. This method is exact copy paste of Retry method except that, it uses TimedId instead of string.
```
jskeet, line 1447: Yes, I think it would be good to try to refactor this.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
...
+                {
+                    if (_retryableIds.TryGetValue(item, out var info))
+                    {
+                        // _retryableIds is a concurrent dictionary but we are updating the value.
+                        _retryableIds.Locked(() => _retryableIds.TryUpdate(item, new RetryInfo(info.FirstTimeOfFailureInUtc, backoff), info));
+                    }
+                }
+            }
+
+            private void HandleAckResponseForExactlyOnce(Task writeTask, List<string> ackIds, List<string> nackIds, List<TimedId> extendIds)
```
jskeet, line 1510: I'd like this to be the only new top-level method - if all the other methods are local methods within this, it makes it much clearer that they're *only* used within this.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
...
+                            // We don't need recursion as the handling of ack response will take care of retry and _retryableIds
+                            // dictionary will ensure that we don't retry after 10 seconds.
+                            actionToRetry(retryableErrors);
+                            StartPush();
+                        }));
+                    }
+                }
+            }
+
+            private void UpdateBackoff(IEnumerable<string> retryableErrors, TimeSpan? backoff)
```
jskeet, line 1498: These aren't the errors, they're the IDs, right?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
...
+            {
+                foreach (var item in retryableErrors)
+                {
+                    if (_retryableIds.TryGetValue(item, out var info))
+                    {
+                        // _retryableIds is a concurrent dictionary but we are updating the value.
```
jskeet, line 1504: Yes, but that's fine. I don't think Locked is necessary here - and the fact that we don't lock on the dictionary elsewhere means it probably doesn't help anyway.

Would we ever expect to see this change, really? Perhaps we should just use `_retryableIds[item] = info.WithBackoff(backoff);` where `WithBackoff` is a new method on `RetryableInfo`.

Note that if we *do* keep this current code, we should implement `IEquatable<RetryableInfo>` - I'd rather not do that if we can help it.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1342,21 +1377,265 @@ private void StartPush()
...
+                    {
+                        var ackError = rpcException.GetAckError(ids);
+                        // Only temporary failures should be retried. Mark them as retryable if not already.
+                        UpdateRetryableIds(ids, ackError);
+
+                        if (hasAckIds)
+                        {
+                            RetryAndHandleAcks(true, ackError);
+                        }
+                        else if (hasNackIds)
```
jskeet, line 1536: What happens if we have both acks and nacks? Do we not want to retry the nacks in that case?

jskeet (commented):
First set of comments.

In general, I think I'd like to emphasize making things as "local" as possible - change everything used only by HandleAckResponseForExactlyOnce to be a local method, etc. That means it'll be a huge method, but it can be completely ignored when we're *not* thinking about exactly once.

How feasible would it be to refactor HandleAckResponse to cover both exactly-once and non-exactly-once scenarios? There's clearly some overlap there, but it's not obvious to me at the moment how much we can refactor.

(Could we change the "split point" to be just error handling, for example? So there'd be a single HandleAckResponse method, but perhaps HandleAckErrorReponse and HandleAckErrorResponseForExactlyOnce as separate methods? Not sure.)

Rishabh-V:
Hi Jon! I have addressed almost all of your comments. Now, there is only one top level method for exactly once and all the other methods are local to it. This is still a work in progress and I am doing further refactoring and clean up. Will update here once I am done and the design document is updated with the changes :) Thanks.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Extensions.cs
```diff
@@ -72,6 +73,66 @@ internal static Exception FlattenIfPossible(this Exception e)
             return exs.Count == 1 ? exs[0] : new AggregateException(exs);
         }
 
+        // This API should be used for excatly once subscriptions only.
+
+        /// <summary>
+        /// Gets the error while calling ACKnowledge or ModifyAckDeadline API.
```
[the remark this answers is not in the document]
Rishabh-V: Yes Jon. I found it in the existing docs. I have now changed it as per your suggestion at other places.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Extensions.cs
```diff
@@ -72,6 +73,66 @@ internal static Exception FlattenIfPossible(this Exception e)
...
+            if (!retryAll)
+            {
+                // We can have complete or partial failure. We classify the error as temporary or permanent
+                // based on the error message starting from TRANSIENT_FAILURE_ or PERMANENT_FAILURE_ (which is ugly but that's the way it is)
+                // and return:
+                // the temporary errors in temporaryErrors collection and
+                // the permanent errors in permanentErrors collection.
+                // Error text is returned in upper case by server, so shouldn't need case insensitive comparison.
+                // We are just reading from an already populated collection, so we should be fine from thread safety perspective here.
+                var temporaryErrors = e.GetErrorInfo()?.Metadata.Where(j => j.Value.Contains("TRANSIENT_FAILURE_")) ?? Enumerable.Empty<KeyValuePair<string, string>>();
```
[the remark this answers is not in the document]
Rishabh-V: Thanks Jon. Done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/Extensions.cs
```diff
@@ -72,6 +73,66 @@ internal static Exception FlattenIfPossible(this Exception e)
...
+                var permanentErrors = e.GetErrorInfo()?.Metadata.Where(j => j.Value.Contains("PERMANENT_FAILURE_")) ?? Enumerable.Empty<KeyValuePair<string, string>>();
+
+                if (temporaryErrors.Any() || permanentErrors.Any())
+                {
+                    // If temporary or permanent errors are present, the entire batch hasn't permanently failed.
```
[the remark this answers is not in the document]
Rishabh-V: Right Jon. I have now updated the comments. PTAL. Happy to change that further if it still doesn't make sense.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -415,11 +415,17 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
             GaxPreconditions.CheckNotNull(settings, nameof(settings));
             settings.Validate();
             // These values are validated in Settings.Validate() above, so no need to re-validate here.
+            // Note: _modifyDeadlineSeconds will always have either the user provided value or the default value of 60 seconds.
```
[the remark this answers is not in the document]
Rishabh-V: Done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Implemented the second option.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: My understanding was that we should have one of acks or nacks or extends in one call. I have now refactored the method to mimic the existing HandleAckResponse method.

Rishabh-V:
Hi Jon. This is now ready for your second round of review. I have tried to incorporate all your feedback, but there are at-least couple of places where I suspect there could be further optimization (added TODO). Please have a look. I have also shared the draft design document (which will need few changes in terms of AckError class and its usage. I will update it once we are good with class here).

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
                 }
             }
 
+            private void HandleAckResponseForExactlyOnce(Task writeTask, List<string> ackIds, List<string> nackIds, List<TimedId> extendIds)
+            {
+                _concurrentPushCount -= 1;
+                _pushInFlight -= ackIds?.Count ?? 0 + nackIds?.Count ?? 0 + extendIds?.Count ?? 0;
```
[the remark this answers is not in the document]
jskeet: (I know it's this way in HandleAckResponse - I'd suggest doing the same there too.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+
+                bool hasAckIds = ackIds?.Count > 0;
+                bool hasNackIds = nackIds?.Count > 0;
+                bool hasExtendIds = extendIds?.Count > 0;
+                // In one call, we 'll have one of acks or nacks or extends, not all.
```
jskeet, line 1406: Nit: "we 'll" => we'll

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                    }
+                }
+                else
+                {
+                    // Everything succeeded. Update _retryableIds dictionary.
+                    UpdateRetryableIds(ids, default);
+                }
+
+                // Perform push so that other messages can be processed.
+                StartPush();
```
jskeet, line 1451: Maybe add an explicit return statement to clarify that everything below here is just local methods?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+
+                // This method schedules the retry of ackIds/nackIds/extendIds with temporary errors and
+                // handles the permanent errors. Permanent errors are propagated to the caller for Acknowledge RPC,
+                // while permanent errors are only logged for ModifyAckDeadline RPC as ModifyAckDeadline call is not user initiated.
+                void RetryAcksAndExtends(bool? ack, AckError ackError, IEnumerable<TimedId> extendIds)
+                {
+                    // ack = true implies acks. ack = false implies nacks. ack = null implies extends, in which case extendIds should be non-empty.
```
jskeet, line 1458: I'd just allow extendIds to be null for non-ack/nack. That's simpler to check for than using `Any` etc - it means it's easier to call, too. I *think* you can just call `RetryAcksAndExtends(hasAckIds, ackError, extendIds)`, if you use `extendIds` being non-null as a signal that it's definitely extend rather than ack/nack. (Or you may find you don't really need `hasAckIds` and `hasNackIds`, and can just use `RetryAcksAndExtends(ackIds is not null, ackError, extendIds)`

jskeet: Or we may find that this really should be two methods after all... anything where one parameter is only relevant half the time feels odd. I might try refactoring it as another commit myself, if you'd like me to. (That could potentially come later.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
```diff
@@ -0,0 +1,127 @@
...
+
+    /// <summary>
+    /// Gets the permanent failures.
+    /// This collection will be populated with AckIds/NackIds/ExtendIds with their corresponding failures only if the failure is permanent,
+    /// else this collection will be empty (but not null).
+    /// </summary>
+    internal IEnumerable<KeyValuePair<string, string>> PermanentFailures { get; }
+
+    /// <summary>
+    /// Attempts to get the exception from the current instance of <seealso cref="AckError"/>.
```
jskeet, line 69: Let's try to clarify this beyond "the exception"

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
```diff
@@ -0,0 +1,127 @@
...
+    /// </summary>
+    /// <remarks>
+    /// This method will return <c>true</c> if and only if there are permanent failures
+    /// in the current instance of <seealso cref="AckError"/>,
+    /// in which case <paramref name="exception"/> will be populated.
+    /// </remarks>
+    /// <param name="exception">The <see cref="AggregateException"/> to populate if there are permanent failures.</param>
+    /// <returns><c>true</c> if there are permanent failures and <paramref name="exception"/> is populated, <c>false</c> otherwise.</returns>
+    internal bool TryGetAcknowledgementException(out AggregateException exception)
```
jskeet, line 78: Possibly TryGetPermanentException? And possibly just return the exception itself or null, instead of using `out`?
(That's what I personally tend to do when I'm returning something nullable.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                        var idsToRetry = ackError.TemporaryFailureIds;
+                        var queue = ack.HasValue && ack.Value ? _ackQueue : _nackQueue;
+                        if (idsToRetry.Any())
+                        {
+                            // Retry for 600 seconds.
+                            RetryTemporaryFailures(idsToRetry, acksToRetry => queue.Locked(() => queue.Requeue(acksToRetry)), Enumerable.Empty<TimedId>(), null, 600);
+                        }
+                    }
+
+                    if (ackError.TryGetAcknowledgementException(out var exception))
```
jskeet, line 1482: If you change this method as suggested earlier, this would be

```csharp
if (ackError.TryGetPermanentException() is AggregateException permanentException)
{
}
```

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                        if (hasExtendIds)
+                        {
+                            // Don't bubble permanent failures for lease extends as they are not user initiated.
+                            // Just log them.
+                            // TODO: Log the permanent failures.
+                            return;
+                        }
+
+                        // For acks/nacks, bubble up the exception.
+                        throw exception.FlattenIfPossible();
```
jskeet, line 1493: If we're going to call `FlattenIfPossible`, maybe it would be simpler just to return a `List<AcknowledgementException>` from `TryGetPermanentException` (which would then be `GetPermanentExceptions`, and could return an empty list if there aren't any).

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                // For extends, extendActionToRetry must be non-null and specify the right queue,i.e, _extendQueue.
+                void RetryTemporaryFailures(
+                    IEnumerable<string> ackOrNackIds,
+                    Action<IEnumerable<string>> ackActionToRetry,
+                    IEnumerable<TimedId> extendIds,
+                    Action<IEnumerable<TimedId>> extendActionToRetry = default,
+                    int timeoutInSeconds = 600)
+                {
+                    // TemporaryFailures may be empty but will never be null, so null check is not needed.
+                    // By design, only items in temporaryFailures that exist in _retyrableIds will be retried.
```
jskeet, line 1511: _retyrableIds => _retryableIds

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                            return;
+                        }
+
+                        // For acks/nacks, bubble up the exception.
+                        throw exception.FlattenIfPossible();
+                    }
+                }
+
+                // Retries the ackIds/nackIds with temporary failures for Acknowledge RPC or extendIds with temporary failures for ModifyAckDeadline RPC.
+                // This method checks if the id exists in _retryableIds dictionary and has not been tried for more than the specified timeout.
```
jskeet, line 1498: Not been tried, or not been successful? (The timeout is since the first failure, right?)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                // For acks and nacks, ackActionToRetry must be non-null and specify the right queue (_ackQueue or _nackQueue).
+                // For extends, extendActionToRetry must be non-null and specify the right queue,i.e, _extendQueue.
+                void RetryTemporaryFailures(
+                    IEnumerable<string> ackOrNackIds,
+                    Action<IEnumerable<string>> ackActionToRetry,
+                    IEnumerable<TimedId> extendIds,
+                    Action<IEnumerable<TimedId>> extendActionToRetry = default,
+                    int timeoutInSeconds = 600)
+                {
+                    // TemporaryFailures may be empty but will never be null, so null check is not needed.
```
jskeet, line 1510: TemporaryFailures => Temporary failures here and elsewhere - I suspect there was a find/replace that went wrong somewhere.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                // If so, it retries those ids by adding them to the appropriate queue and calls the StartPush() method.
+                // For acks and nacks, ackOrnackIds must be non-null. For extends, extendIds must be non-null.
+                // For acks and nacks, ackActionToRetry must be non-null and specify the right queue (_ackQueue or _nackQueue).
+                // For extends, extendActionToRetry must be non-null and specify the right queue,i.e, _extendQueue.
+                void RetryTemporaryFailures(
+                    IEnumerable<string> ackOrNackIds,
+                    Action<IEnumerable<string>> ackActionToRetry,
+                    IEnumerable<TimedId> extendIds,
+                    Action<IEnumerable<TimedId>> extendActionToRetry = default,
+                    int timeoutInSeconds = 600)
```
jskeet, line 1508: Mmm... this is another "not really sure if this should be two methods" method at the moment. Again, it can wait for later refactoring.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+
+                    bool hasAcksOrNacks = ackOrNackIds.Any();
+                    bool hasExtends = extendIds.Any();
+                    if (!hasAcksOrNacks && !hasExtends)
+                    {
+                        return;
+                    }
+
+                    var idsToRetry = hasAcksOrNacks ? ackOrNackIds : extendIds.Select(j => j.Id);
+                    var retryableErrorIds = new ConcurrentBag<string>();
```
jskeet, line 1523: I don't think this needs to be a ConcurrentBag - I suspect it could just be a `List<string>`

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                        retryableErrorIds.Add(id);
+                    }
+
+                    if (retryableErrorIds.Count == 0)
+                    {
+                        // Nothing to retry.
+                        return;
+                    }
+
+                    // All ids in retryableErrorIds will have the same backoff.
```
jskeet, line 1537: I'm not sure why that would be the case. Might we not get into a situation where we're retrying various IDs, and some started a while ago (and have a long backoff) and some have only just started being ACKed?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                    if (!_retryableIds.TryGetValue(retryableErrorIds.First(), out var retryInfo))
+                    {
+                        // Nothing to retry.
+                        return;
+                    }
+
+                    var backoff = s_ackBackoff.NextBackoff(retryInfo.CurrentBackoff ?? TimeSpan.Zero);
+                    // We should retry only for specified timeout.
+                    // If current UTC time is greater than first UTC time of failure + timeout, we should not retry.
+                    if (DateTime.UtcNow > retryInfo.FirstTimeOfFailureInUtc + TimeSpan.FromSeconds(timeoutInSeconds) - backoff)
```
jskeet, line 1548: This isn't going to be testable - use an `IClock` instead, which you could add to `Settings` in the same way as `IScheduler`

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                    {
+                        // We are past the retry timeout.
```
jskeet, line 1550: So presumably these are now permanent failures... but how do they become permanent?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                    UpdateAllRetryableErrorBackoff(retryableErrorIds, backoff);
+                    // Retry errors that are still temporary errors. They will be retried after backOff.
+                    Task delayTask = _scheduler.Delay(backoff, _softStopCts.Token);
+                    // TODO: Check, if the below expression can be simplified or optimized.
+                    Add(delayTask, new NextAction(false, hasAcksOrNacks
+                    ? () => { ackActionToRetry(retryableErrorIds); StartPush(); }
+                    : () => { extendActionToRetry(extendIds.Where(j => retryableErrorIds.Contains(j.Id))); StartPush(); }));
+                }
+
+                // Updates the backoff of the retryableErrorIds in _retryableIds dictionary.
```
jskeet, line 1565: Presumably this is safe to do in the "try-get then update" because for any given ID, we won't be running this code while there's an in-flight request that could complete (and run it separately)?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                        {
+                            _retryableIds[id] = info.WithBackoff(backoff);
+                        }
+                    }
+                }
+
+                // Updates the _retryableIds dictionary. Ensures that the ids that should be retried exist in the dictionary.
+                // By design, only ids that are in _retryableIds dictionary will be retried with exponential backoff.
+                // Removes the ids that are either successful or become permanent failures as they should not be retried.
+                void UpdateRetryableIds(IEnumerable<string> allIds, AckError ackError)
```
jskeet, line 1580: Please add method parameter documentation - I don't know what `allIds` means here.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                {
+                    var temporaryFailureIds = ackError?.TemporaryFailureIds ?? Enumerable.Empty<string>();
+                    var permanentFailureIds = ackError?.PermanentFailureIds ?? Enumerable.Empty<string>();
+                    // Successful Ids = AllIds - (TemporaryFailures + Permanent failures)
+                    var successfulIds = allIds.Locked(() => allIds.Except(temporaryFailureIds.Union(permanentFailureIds)));
```
jskeet, line 1585: I don't think you need `Locked` here - and if you do, there's probably a bigger issue.

jskeet: I suspect it's also more efficient to write `allIds.Except(temporaryFailureIds).Except(permanentFailureIds);`

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                    // Some ids may have permanent failures and some may have succeeded. Those ids shouldn't be retried. 
+                    var nonRetryableIds = permanentFailureIds.Union(successfulIds);
+
+                    foreach (var item in temporaryFailureIds)
+                    {
+                        // If the id to be retried is already in _retryableIds dictionary we are good,
+                        // else add the id to _retryableIds dictionary with first time of failure as DateTime.UtcNow,
+                        // so that we can retry for specific duration.
+                        _ = _retryableIds.TryAdd(item, new RetryInfo(DateTime.UtcNow));
```
jskeet, line 1594: Again, let's use an IClock.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,8 +1395,221 @@ private void StartPush()
...
+                // By design, only ids that are in _retryableIds dictionary will be retried with exponential backoff.
+                // Removes the ids that are either successful or become permanent failures as they should not be retried.
+                void UpdateRetryableIds(IEnumerable<string> allIds, AckError ackError)
+                {
+                    var temporaryFailureIds = ackError?.TemporaryFailureIds ?? Enumerable.Empty<string>();
+                    var permanentFailureIds = ackError?.PermanentFailureIds ?? Enumerable.Empty<string>();
+                    // Successful Ids = AllIds - (TemporaryFailures + Permanent failures)
+                    var successfulIds = allIds.Locked(() => allIds.Except(temporaryFailureIds.Union(permanentFailureIds)));
+                    // Some ids may have permanent failures and some may have succeeded. Those ids shouldn't be retried. 
+                    var nonRetryableIds = permanentFailureIds.Union(successfulIds);
```
jskeet, line 1587: Possibly use Concat rather than Union, given that they don't intersect?

jskeet (commented):
I'm gradually understanding this more, I think. I still don't quite get how temporary errors become permanent after their deadline is exceeded though.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: If no successful ack or modack is made for a message id for 10 mins, server will treat that ack id as expired. That's because the max ack deadline can be 10 mins. After that, an ack/modack will fail with a permanent error since the previous ack is now invalid. So, the Pub/Sub team doesn't want us to retry after 10 minutes.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: We are only processing the response for the given ackIds/nackIds/extendIds and know their latest status before scheduling them again with a backoff, so we should be good.

Rishabh-V (commented):
Thanks Jon for the review. I will address these and push a new commit with fixes by tomorrow morning your time. I have answered the queries. Please see and let me know if any further changes are needed. Thank you.

Rishabh-V:
Hi Jon. I have pushed a new commit with the review comments addressed. Added couple of basic tests to check that the functionality is working. Will add few more tests in the later commit. Please take a look. Thank you.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
```diff
@@ -66,23 +60,31 @@ internal AckError(IEnumerable<KeyValuePair<string, string>> temporaryFailures, I
...
+    /// Gets the permanent errors as the <see cref="IEnumerable{T}">IEnumerable</see>&lt;<see cref="AcknowledgementException"/>&gt;
+    /// from the current instance of <seealso cref="AckError"/>.
+    /// </summary>
+    /// <remarks>
+    /// This method will return the populated <see cref="IEnumerable{T}">IEnumerable</see>&lt;<see cref="AcknowledgementException"/>&gt; if there are permanent failures
+    /// in the current instance of <seealso cref="AckError"/>, otherwise
+    /// <see cref="IEnumerable{T}">IEnumerable</see>&lt;<see cref="AcknowledgementException"/>&gt; will be empty (but not null).
+    /// </remarks>
+    /// <returns>The <see cref="IEnumerable{T}">IEnumerable</see>&lt;<see cref="AcknowledgementException"/>&gt; if there are permanent failures, empty otherwise.</returns>
+    internal IEnumerable<AcknowledgementException> GetPermanentException() =>
```
jskeet, line 83: Nit that's my fault - this should be GetPermanentExceptions given that it's returning a sequence.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
```diff
@@ -66,23 +60,31 @@ internal AckError(IEnumerable<KeyValuePair<string, string>> temporaryFailures, I
...
+        TemporaryFailures = temporaryFailures;
+        PermanentFailures = permanentFailures;
     }
 
+    /// <summary>
+    /// Gets the permanent errors as the <see cref="IEnumerable{T}">IEnumerable</see>&lt;<see cref="AcknowledgementException"/>&gt;
+    /// from the current instance of <seealso cref="AckError"/>.
+    /// </summary>
+    /// <remarks>
+    /// This method will return the populated <see cref="IEnumerable{T}">IEnumerable</see>&lt;<see cref="AcknowledgementException"/>&gt; if there are permanent failures
```
jskeet, line 78: I think this would be simpler as:

```csharp
This method will always return a sequence of <see cref="AcknowledgementException"/>; this sequence will be empty if there are no permanent errors.
```

(Whatever we do, let's get rid of all the `&lt;` etc. It's hard to read here, and it won't help for browsing in VS really.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1395,15 +1428,31 @@ private void StartPush()
                 }
             }
 
+            /// <summary>
+            /// This method is called when the acknowledgement request completes.
+            /// This method handles the response from an acknowledgement request for exactly once subscriptions only.
+            /// For non-exactly once subscriptions, <see cref="HandleAckResponse(Task, List{string}, List{string}, List{TimedId})"/> is called.
+            /// </summary>
```
jskeet, line 1435: Do we expect only one of ackIds, nackIds and extendIds to be non-null? If so, I think that's worth noting.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1520,78 +1568,70 @@ void RetryAcksAndExtends(bool? ack, AckError ackError, IEnumerable<TimedId> exte
...
                     var idsToRetry = hasAcksOrNacks ? ackOrNackIds : extendIds.Select(j => j.Id);
-                    var retryableErrorIds = new ConcurrentBag<string>();
+                    var retryableErrorIds = new List<string>();
                     // Not all temporaryFailures may need to be retried.
                     // See if the id exists in the _retryableIds dictionary and only then retry.
-                    foreach (var id in idsToRetry.Where(j => _retryableIds.ContainsKey(j)))
-                    {
-                        retryableErrorIds.Add(id);
-                    }
+                    retryableErrorIds.AddRange(idsToRetry.Where(j => _retryableIds.ContainsKey(j)));
```
jskeet, line 1574: Now this is a list, we can just have:

```csharp
var retryableErrorIds = idsToRetry.Where(j => _retryableIds.ContainsKey(j)).ToList();
```

Or potentially:

```csharp
var retryableErrorIds = idsToRetry.Intersect(_retryableIds.Keys).ToList();
```

I don't know whether the latter would create a snapshot though.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1520,78 +1568,70 @@ void RetryAcksAndExtends(bool? ack, AckError ackError, IEnumerable<TimedId> exte
...
 
-                    // All ids in retryableErrorIds will have the same backoff.
-                    // Get the retryInfo for the first retryableErrorId from the _retryableIds dictionary.
-                    if (!_retryableIds.TryGetValue(retryableErrorIds.First(), out var retryInfo))
+                    foreach (var id in retryableErrorIds)
                     {
-                        // Nothing to retry.
-                        return;
-                    }
+                        // Only ids that exist in _retryableIds dictionary are added to retryableErrorIds from idsToRetry.
```
jskeet, line 1584: And we won't ever update `_retriableIds` in a different thread, between the above code and this? I *think* that's the case, but please confirm if that's your understanding as well. (I believe the individual client is effectively single-threaded, just highly asynchronous. So the fact that there aren't any awaits involved means we should be fine. Is that right?)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1520,78 +1568,70 @@ void RetryAcksAndExtends(bool? ack, AckError ackError, IEnumerable<TimedId> exte
...
-                    Add(delayTask, new NextAction(false, hasAcksOrNacks
-                    ? () => { ackActionToRetry(retryableErrorIds); StartPush(); }
-                    : () => { extendActionToRetry(extendIds.Where(j => retryableErrorIds.Contains(j.Id))); StartPush(); }));
-                }
-
-                // Updates the backoff of the retryableErrorIds in _retryableIds dictionary.
-                void UpdateAllRetryableErrorBackoff(IEnumerable<string> ids, TimeSpan? backoff)
-                {
-                    foreach (var id in ids)
+                    // The ids that exist in both retryableErrorIds and _retryableIds now are the ones that need to be retried.
```
jskeet, line 1601: I've got an idea for a refactoring that I think would simplify this - but it can wait until this PR is in.
(Basically it would avoid creating the the first set of `retryableErrorIds`, instead looping through `idsToRetry` and only creating a list of IDs we *genuinely* want to retry. I won't go into more detail now, but I should probably try to do it very soon after this is in, before I forget what I meant...)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1520,78 +1568,70 @@ void RetryAcksAndExtends(bool? ack, AckError ackError, IEnumerable<TimedId> exte
...
+                        Task delayTask = _scheduler.Delay(backoff, _softStopCts.Token);
+                        Add(delayTask, new NextAction(false, hasAcksOrNacks
+                        ? () => { ackActionToRetry(retryIds); StartPush(); }
+                        : () => { extendActionToRetry(extendIds.Where(j => retryIds.Contains(j.Id))); StartPush(); }));
                     }
                 }
 
-                // Updates the _retryableIds dictionary. Ensures that the ids that should be retried exist in the dictionary.
-                // By design, only ids that are in _retryableIds dictionary will be retried with exponential backoff.
+                // Updates the _retryableIds dictionary. Ensures that the ids that should be re-tried exist in the dictionary.
```
jskeet, line 1615: I don't think there's any need for this change. (And retryable is fine as one word as well, without hyphenation - ditto backoff usually.)
If you find it easier to read this way though, that's fine. You're more likely to be maintaining it than I am, and it's not in documentation.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1520,78 +1568,70 @@ void RetryAcksAndExtends(bool? ack, AckError ackError, IEnumerable<TimedId> exte
...
                 // Removes the ids that are either successful or become permanent failures as they should not be retried.
+                // allIds - All the ids for which Acknowledge/ModifyAckDeadline request was sent.
+                // ackError - The AckError that occurred in the response.
                 void UpdateRetryableIds(IEnumerable<string> allIds, AckError ackError)
                 {
                     var temporaryFailureIds = ackError?.TemporaryFailureIds ?? Enumerable.Empty<string>();
                     var permanentFailureIds = ackError?.PermanentFailureIds ?? Enumerable.Empty<string>();
                     // Successful Ids = AllIds - (TemporaryFailures + Permanent failures)
-                    var successfulIds = allIds.Locked(() => allIds.Except(temporaryFailureIds.Union(permanentFailureIds)));
+                    var successfulIds = allIds.Except(temporaryFailureIds).Except(permanentFailureIds);
```
jskeet, line 1625: We should consider the impact of lazy evaluation at some point - I suspect we're fine, but it might be worth adding a TODO.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -1111,5 +1119,103 @@ public void AckModifyAckDeadlineFault_Thrown([CombinatorialValues(true, false)]
...
+                ServerAction.Inf()
+            };
+
+            // Any temporary failure in ack/nack/extend request should not be thrown to the client. It should be retried.
+            var exception = new RpcException(new Status(StatusCode.DeadlineExceeded, ""), "");
+            // If ackNackOrExtends is null, then it is extend request. ModifyAcknowledgeDeadline RPC will throw the supplied exception.
+            // If ackNackOrExtends is true, then it is acknowledge request. Acknowledge RPC will throw the supplied exception.
+            // If ackNackOrExtends is false, then it is nack request. ModifyAcknowledgeDeadline RPC will throw the supplied exception.
+            var ackModifyAckDeadlineAction =
+                ackNackOrExtends == null ? AckModifyAckDeadlineAction.Data(exception, 10, false, true)
```
jskeet, line 1141: Maybe use named arguments for 10, false, true so we have some idea what they mean?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -1111,5 +1119,103 @@ public void AckModifyAckDeadlineFault_Thrown([CombinatorialValues(true, false)]
...
+            {
+                ServerAction.Data(TimeSpan.Zero, new[] { "1" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "2" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "3" }),
+                ServerAction.Data(TimeSpan.Zero, new[] { "4" }),
+                ServerAction.Inf()
+            };
+
+            // Any temporary failure in ack/nack/extend request should not be thrown to the client. It should be retried.
+            var exception = new RpcException(new Status(StatusCode.DeadlineExceeded, ""), "");
```
jskeet, line 1136: So this is a *general* exception rather than one applying to specific messages, yes? I assume we will also end up with tests where some succeed, some fail permanently and some fail temporarily?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -1111,5 +1119,103 @@ public void AckModifyAckDeadlineFault_Thrown([CombinatorialValues(true, false)]
...
+                });
+                await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(100), CancellationToken.None));
+                await fake.TaskHelper.ConfigureAwait(fake.Subscriber.StopAsync(CancellationToken.None));
+                // Despite temporary failures, all 4 messages should be handled.
+                Assert.Equal(new[] { "1", "2", "3", "4" }, handledMsgs);
+            });
+        }
+
+        [Theory, CombinatorialData]
+        public void ExactlyOnce_PermanentFault([CombinatorialValues(true, false, null)] bool? ackNackOrExtends)
```
jskeet, line 1168: Given how much conditional code there is, possibly separate out extend from ack/nack?

jskeet (approved):
Looks good - I don't know whether you want to merge what there is so far before writing additional tests, or write the tests first.

Rishabh-V:
> Looks good - I don't know whether you want to merge what there is so far before writing additional tests, or write the tests first.

Thanks Jon for getting it to this state.  I will address your comments.  I was thinking of adding the new tests and then pushing. Please let me know if you have other suggestions. Thanks a ton for your help :)

jskeet:
Adding some more tests before merging sounds fine - I don't think it needs to be completely exhaustive though. Iterating on tests later (but before the first release!) would probably make it simpler.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
Rishabh-V: Right Jon. I will add them.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
Rishabh-V: I missed to spot. Thanks. Fixed it now.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Done. Used the version with Where.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Right Jon. This is my understanding as well. I was trying to be defensive just to be safe.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Sure Jon. Updated the comments. Removed hyphen.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Done.

Rishabh-V (commented):
Hi Jon. Addressed the review comments in third commit. Please take a look. If this is fine, I can squash and merge.

jskeet (approved):

Rishabh-V:
Squashed the changes in one commit. Will merge once tests are green.

Rishabh-V:
Based on discussion with Jon, there is one aspect that needs to be checked and validated - Its about IDs that had effectively timed out - Those should be propagated as errors somehow, but at the moment we're just stopping retrying.

Rishabh-V:
@pradn and @maheshgattani -Please take a look at this PR and share your feedback. Thank you.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
```diff
@@ -0,0 +1,152 @@
...
+        if (temporaryErrors.Any() || permanentErrors.Any())
+        {
+            // If temporary or permanent error messages are present, it means the entire request hasn't failed with a permanent error.
+            return new AckError(temporaryErrors, permanentErrors);
+        }
+        else
+        {
+            // Now, if temporary or permanent error messages are *not* present, it means the entire request has failed with a permanent error.
+            // It may be due to a gRPC error like a failed precondition or a permission denied error or other permanent error.
+            return new AckError(temporaryErrors, ids.ToDictionary(key => key, value => rpcException.Status.StatusCode.ToString()));
```
pradn, line 130: I know `temporaryErrors` will be empty at this point. But passing in a variable means we can to trace the logic of the variable and deduce this fact. Perhaps just pass in an empty container instead, just to make it easier to follow?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,10 +1428,231 @@ private void StartPush()
...
+            /// </remarks>
+            private void HandleAckResponseForExactlyOnce(Task writeTask, List<string> ackIds, List<string> nackIds, List<TimedId> extendIds)
+            {
+                _concurrentPushCount -= 1;
+                _pushInFlight -= (ackIds?.Count ?? 0) + (nackIds?.Count ?? 0) + (extendIds?.Count ?? 0);
+                
+                bool hasAckIds = ackIds?.Count > 0;
+                bool hasNackIds = nackIds?.Count > 0;
+                bool hasExtendIds = extendIds?.Count > 0;
+                // In one call, we'll have one of acks or nacks or extends, not all.
```
pradn, line 1456: Consider an assertion for this.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1125,6 +1192,7 @@ private void HandlePullMessageData(Task<bool> moveNextTask)
                     try
                     {
                         current = _pull.GrpcCall.ResponseStream.Current;
+                        _exactlyOnceDeliveryEnabled = current.SubscriptionProperties?.ExactlyOnceDeliveryEnabled ?? false;
```
pradn, line 1195: Do you implement the following in this PR?

> Send a new ack deadline via the streaming RPC’s stream_ack_deadline_seconds whenever the exactly_once setting changes.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1360,10 +1428,231 @@ private void StartPush()
...
+                    if (rpcException != null)
+                    {
+                        var ackError = AckError.ForRpcException(rpcException, ids);
+                        // Only temporary failures should be retried. Mark them as retryable if not already.
+                        UpdateRetryableIds(ids, ackError);
+
+                        if (hasAckIds)
+                        {
+                            // ack = true implies acks.
+                            RetryAcksAndNacks(true, ackError);
```
pradn, line 1473: It's best to have two methods `RetryAcks` and `RetryNacks` and then call into `RetryAcksAndNacks` in those methods. Then you'd not need to expose bools and explain them via a comment.

pradn (commented):
Generally looks good, Rishabh! Good work.

maheshgattani (commented):
Where can I see:

- The new APIs for ack and nack with response?
-  Changes to stream ack deadline?

I haven't looked at the full PR but I want to understand the layout of the changes first.

Rishabh-V (commented):
Hi @pradn and @maheshgattani - Please see my response inline. I will raise a new PR with the review comments addressed.
Hi @jskeet - Need your input on the assertion in the production code.

Thanks

Rishabh-V:
> Where can I see:
> 
> * The new APIs for ack and nack with response?
> * Changes to stream ack deadline?
> 
> I haven't looked at the full PR but I want to understand the layout of the changes first.

In C# Pub/Sub library, the acks, nacks and extends were not fire and forget. We were already handling and processing the exceptions for acks, nacks and extends. `HandleAckResponse` is used to handle the response of acks/nacks/extends. For exactly once delivery, `HandleAckResponseForExactlyOnce` (post new PR `HandleAckResponseForExactlyOnceDelivery`) is used. This is as described in the design doc.

_modifyDeadlineSeconds is passed to StreamAckDeadlineSeconds (line 1115). _modifyDeadlineSeconds always has either a user provided value or a default value of 60 seconds, so no change is needed.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
jskeet: We have plenty of precondition checks around the libraries - mostly for externally-provided arguments, but it would be fine to use here. I'd suggest using `GaxPrecondition.CheckArgument`. Admittedly I expect the expression will be slightly fiddly to write - something like:

```csharp
// Any of the parameter names would be okay here, I think...
GaxPreconditions.CheckArgument((hasAckIds ? 1 : 0) + (hasNackIds ? 1 : 0) + (hasExtendIds ? 1 : 0) <= 1, nameof(...), "At most one of AckIds, NackIds or ExtendIds should be non-empty.");
```

Rishabh-V: > We have plenty of precondition checks around the libraries - mostly for externally-provided arguments, but it would be fine to use here. I'd suggest using `GaxPrecondition.CheckArgument`. Admittedly I expect the expression will be slightly fiddly to write - something like:
> 
> ```cs
> // Any of the parameter names would be okay here, I think...
> GaxPreconditions.CheckArgument((hasAckIds ? 1 : 0) + (hasNackIds ? 1 : 0) + (hasExtendIds ? 1 : 0) <= 1, nameof(...), "At most one of AckIds, NackIds or ExtendIds should be non-empty.");
> ```

Thanks Jon. Somehow, immediately after hearing assertion, `Debug.Assert` , `Trace.Assert` came to my mind. I will now add your code suggestion.

Rishabh-V:
Hi @maheshgattani , @pradn , @jskeet  - Please see #8956 which addresses the review comments logged here. Thanks.

pradn:
SubscriberClient:1115 does send the modack deadline when a stream is *initialized*. and this does depend on the EOS setting. however:
1) the EOS setting is received on a stream response. so when sending the initial request, we're not going to send an EOS-based deadline.
2) sending only on initial request means we don't send if the EOS setting changes in the middle of the stream. remember that EOS can be turned on and off any time. so we need to send when there's a change in the EOS setting.

PR #8904, opened 2022-07-20
feat: Adds table ACLs support

amanda-tarafa:


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/TablesTest.cs
```diff
@@ -780,5 +780,44 @@ public void GetTable_PartialSchema_ComplexTypes()
...
+        {
+            var client = BigQueryClient.Create(_fixture.ProjectId);
+            var policy = client.GetTableIamPolicy(
+                _fixture.DatasetId, _fixture.HighScoreTableId, new GetTableIamPolicyOptions { PolicyVersion = 3 });
+
+            Assert.NotNull(policy.ETag);
+        }
+
+        [Fact]
+        public void SetTableIamPolicy()
```
jskeet, line 795: Given that this test is mutating the table, perhaps it should create a new table just for that purpose?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.GenerateOverloads/Methods/TableCrud.xml
```diff
@@ -90,4 +90,43 @@
       <returns>The updated table.</returns>
     </Comments>
   </Method>
+
+  <Method Name="GetTableIamPolicy"
+     TargetType="Table"
+     ReturnType="Policy">
+      <Comments>
+          <summary>
+              Gets the IAM Policy applied to {target}.
```
jskeet, line 99: Should policy really be capitalized, here and below?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryTable.cs
```diff
@@ -262,7 +262,40 @@ public BigQueryTable Patch(Table resource, bool matchETag, PatchTableOptions opt
...
+        /// <remarks>
+        /// This method delegates to <see cref="BigQueryClient.SetTableIamPolicy(TableReference, Policy, SetTableIamPolicyOptions)"/>.
+        /// </remarks>
+        /// <param name="policy">The policy to apply to this table.</param>
+        /// <param name="options">The options for the operation. May be null, in which case defaults will be supplied.</param>
+        /// <returns>The resulting IAM Policy after appying <paramref name="policy"/> to this table.</returns>
+        public Policy SetIamPolicy(Policy policy, SetTableIamPolicyOptions options = null) => _client.SetTableIamPolicy(Reference, policy, options);
+
+        /// <summary>
+        /// Tests whether the caller has the given permissions.
```
jskeet, line 288: "for this table"?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryTable.cs
```diff
@@ -480,6 +513,48 @@ public Task<BigQueryTable> PatchAsync(Table resource, bool matchETag, PatchTable
             return _client.PatchTableAsync(Reference, resource, options, cancellationToken);
         }
 
+        /// <summary>
+        /// Asynchronously gets the IAM Policy applied to this table
```
jskeet, line 517: Period and casing.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryTable.cs
```diff
@@ -480,6 +513,48 @@ public Task<BigQueryTable> PatchAsync(Table resource, bool matchETag, PatchTable
...
+        /// <param name="policy">The policy to apply to this table.</param>
+        /// <param name="options">The options for the operation. May be null, in which case defaults will be supplied.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <returns>A task representing the asynchronous operation. When complete, the result is
+        /// the resulting IAM Policy after appying <paramref name="policy"/> to this table.</returns>
+        public Task<Policy> SetIamPolicyAsync(Policy policy, SetTableIamPolicyOptions options = null, CancellationToken cancellationToken = default) =>
+            _client.SetTableIamPolicyAsync(Reference, policy, options, cancellationToken);
+
+        /// <summary>
+        /// Asynchronously tests whether the caller has the given permissions.
```
jskeet, line 544: "for this table"?

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetTableIamPolicyOptions.cs
```diff
@@ -0,0 +1,46 @@
...
+/// <summary>
+/// Options for <c>GetTableIamPolicy</c> operations.
+/// </summary>
+public sealed class GetTableIamPolicyOptions
+{
+    /// <summary>
+    /// Optional. The maximum policy version that will be used to format the policy.
+    /// </summary>
+    /// <remarks>
+    /// Valid values are 0, 1, and 3. Requests specifying an invalid value will be rejected.
```
jskeet, line 29: Potentially create our own enum here? This is really horrible...

amanda-tarafa: Yes, I thought about adding our enum, but I was worried that then we'd need to play catch up with new versions. Since we are exposing the Apiary types, we shouldn't need to make any changes in the future.

Unless we add the enum, but don't validate it's value, in which case we'd have the workaround of assigning the cast int to the enum property? What do you think?

jskeet: Mmm. Maybe it's best leaving it as it is.

amanda-tarafa: Agreed, at worst we can add an enum later on and sync the properties while we deprecate this one. Also maybe @shollyman has some context on how often this could change etc.

shollyman: This is the IAM meta API, so I have no confidence in anticipating changes.  The IAM folks may be better able to forecast here.

shollyman (approved):
Thanks for putting this together.  The IAM meta API is an odd bolt-on but valuable, so its good to get coverage here.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetTableIamPolicyOptions.cs
amanda-tarafa: OK, leaving as is for now, we can always reconsider an enum later on, but it seems like the raw int value will mean we'll be up to date easier.

PR #8906, opened 2022-07-20
fix(spanner/admin/database): switch service config

noahdietz:
This switches the generation config to use the new, more up-to-date service config file. The next generation of this client may include some differences as a result.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -3311,7 +3311,7 @@
       },
       "noVersionHistory": true,
       "shortName": "spanner",
-      "serviceConfigFile": "spanner_admin_database.yaml"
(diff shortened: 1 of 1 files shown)


noahdietz:
cc @rishabh-v

jskeet (approved):
Note for @Rishabh-V: this affects what happens if we run generate_apis.sh locally. It doesn't affect what Bazel Bot does. In *some* cases it affects the OwlBot PR, because there are some tweaks needed that involve regenerating. Not sure if that affects the Spanner Admin Database API offhand though.

Rishabh-V (approved):
I am unfamiliar with this area. The change looks fine to me and more importantly Jon has reviewed it.

PR #8933, opened 2022-07-27
Update gRPC generator

jskeet:


jskeet:
Looks like we need a generator change for at least one API, in terms of deprecation. That's surprising - I'll look into it.

jskeet:
I'm fixing OSConfig now...

amanda-tarafa (approved):
LGTM

PR #8935, opened 2022-07-27
test: Add more routes to the console logger test app

jskeet:
This helps to investigate how Error Reporting treats various kinds
of exception.


amanda-tarafa (approved):
LGTM.
Yep, this is good for helping with documenting behaviour, and also if some more work is done on the Error Reporting side of things to improve logs, we'll be able to check for them easier.

PR #8943, opened 2022-07-27
refactor: A small refactoring of exactly-once code

jskeet:
When we're retrying, we can avoid touching the dictionary most of
the time. We only need it for:

- Checking (once) whether an ID is already in the dictionary
- Removing an ID if we've gone beyond the deadline
- Updating the entry if we're backing off

As we go through the list of IDs a single time, ignoring any that
aren't in the dictionary, we can build up a list of items that we
actually want to retry, and use that afterwards without consulting
the dictionary again.


jskeet:
Hey Rishabh, if you don't find this easier to understand, I'm fine to just close it unmerged. It's just an option :)

Rishabh-V (approved):
Thank you Jon. This is much better and efficient. I also like this naming better.

PR #8948, opened 2022-07-29
feat: Prototype PR for fine grained access control

hemanshv:
Hi @jskeet @amanda-tarafa 
This is a prototype PR to demonstrate  high level approach for spanner fine grained access control. It is still very much a work in progress, and formatting, docs are not done yet. Also, there is an issue with the whitespace/line-ending and file-encoding in my visual studio, so there are few whitespace changes which are not intended. I am trying to fix it.
Please have a look at the approach and provide your valuable feedback.
Thank you


jskeet (commented):
Overall, it's not immediately clear to me that the role should be part of the session pool at all - it certainly seems a little out of place in the SessionPoolOptions, where everything else is about how long things last *in* the pool.

It's *possible* that it should be part of `SpannerClientCreationOptions` which is normally used to differentiate between session pools. At the moment it's not clear to me whether the expectation is that every session within the pool should have the same role. I assume the role does "belong" to the session, and can't be changed afterwards?

hemanshv:
Closing  this for #8995

PR #8962, opened 2022-08-03
DO NOT MERGE: Prototype of how we can convey more information to subscribers

jskeet:


jskeet:
@Rishabh-V This is *very* rough and ready, but is a starting point to at least discuss.

If users actually want to be able to react per-message to failures etc, we could potentially provide a public subclass which did all the "book-keeping" necessary - I'd need to see idealised customer code in order to know what to propose there.

jskeet:
Note for the breaking change: I'd forgotten that SubscriberClientImpl is public. We can easily fix this, but I'm busy with something else right now. It shouldn't change the general shape of the PR.

Rishabh-V:
> @Rishabh-V This is _very_ rough and ready, but is a starting point to at least discuss.
> 
> If users actually want to be able to react per-message to failures etc, we could potentially provide a public subclass which did all the "book-keeping" necessary - I'd need to see idealised customer code in order to know what to propose there.

Thank you very much Jon. I will create a sample code using this and discuss with Pub/Sub team and take it forward from here. Thanks again. This is very helpful.

Rishabh-V:
This is now incorporated in #8973 and can be closed.

PR #8965, opened 2022-08-04
fix : Improved tool's performance and enhanced it to accept bucket's configuration as input parameter.

anuragsrivstv:
In this PR:-
1. Improved tool's performance by modifying it to create only 1 file of specified size for all number of samples.
2. Enhanced tool to accept bucket's location , storage class and object versioning enabled flag as input parameter.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Configuration.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Configuration.cs
@@ -29,10 +29,21 @@ internal sealed class Configuration
     internal string OutputFolder { get; }
     internal int UploadChunkSizeInBytes { get; }
     internal int DownloadChunkSizeInBytes { get; }
-    internal string Bucket { get; } = IdGenerator.FromDateTime(prefix: "benchmark-");
+    internal string BucketStorageClass { get; }
+    internal string BucketLocation { get; }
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.BenchmarkingTool/Configuration.cs
```diff
@@ -29,29 +29,47 @@ internal sealed class Configuration
...
     internal int DownloadChunkSizeInBytes { get; }
-    internal string Bucket { get; } = IdGenerator.FromDateTime(prefix: "benchmark-");
+    internal string BucketStorageClass { get; }
+    internal string BucketLocation { get; }
+    internal bool ObjectVersioningEnabled { get; }
+    internal string BucketName { get; } = IdGenerator.FromDateTime(prefix: "benchmark-");
     internal string ObjectName => "obj_random_data";
 
-    private Configuration(int numOfSamples, int dataSizeInBytes, string outputFolder, int uploadChunkSizeInBytes, int downloadChunkSizeInBytes)
+    private Configuration(
```
jskeet, line 38: One option to consider is having a *single* command line argument which refers to a JSON file containing the configuration. You could then just use Json.NET or similar to deserialize the whole configuration. It would also make it simple to maintain multiple configurations, as separate files, rather than have to mess about with different options on the command line.

Something for a separate PR probably, but worth thinking about.

jskeet (approved):

PR #8973, opened 2022-08-05
feat: exactly once delivery (feature complete)

Rishabh-V:
Draft of all the changes with tests.
I will iterate over it one more time before asking for review but wanted to check if the changes are backward compatible.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
@@ -60,12 +60,27 @@ internal sealed class AckError
     internal IEnumerable<KeyValuePair<string, string>> PermanentFailures { get; }
 
     /// <summary>
+    /// Gets the AckIds/ExtendIds/NackIds that are successful.
+    /// This collection will be populated with AckIds/ExtendIds/NackIds only if they are successful,
+    /// else this collection will be empty (but not null).
+    /// </summary>
+    internal IEnumerable<string> SuccessfulIds => MessageIds.Except(PermanentFailureIds).Except(TemporaryFailureIds);
+
+    /// <summary>
+    /// Gets all the AckIds/ExtendIds/NackIds for which the response encountered a <see cref="RpcException"/>.
+    /// This collection is always populated with AckIds/ExtendIds/NackIds and is never null.
+    /// </summary>
+    internal IEnumerable<string> MessageIds { get; }
+
+    /// <summary>
     /// Initializes a new instance of the <see cref="AckError"/> class.
     /// </summary>
+    /// <param name="messageIds">The collection of AckId/NackId/ExtendId for which response is being processed.</param>
     /// <param name="temporaryFailures">The collection of temporary failed AckId/NackId/ExtendId as key and their corresponding failure reason as value.</param>
     /// <param name="permanentFailures">The collection of permanent failed AckId/NackId/ExtendId as key and their corresponding failure reason as value.</param>
-    internal AckError(IEnumerable<KeyValuePair<string, string>> temporaryFailures, IEnumerable<KeyValuePair<string, string>> permanentFailures)
+    internal AckError(IEnumerable<string> messageIds, IEnumerable<KeyValuePair<string, string>> temporaryFailures, IEnumerable<KeyValuePair<string, string>> permanentFailures)
     {
+        MessageIds = messageIds;
         TemporaryFailures = temporaryFailures;
         PermanentFailures = permanentFailures;
     }
@@ -81,8 +96,22 @@ internal sealed class AckError
     /// <returns>The sequence of <see cref="AcknowledgementException"/> if there are permanent failures, empty otherwise.</returns>
     internal IEnumerable<AcknowledgementException> GetPermanentExceptions() =>
         HasPermanentFailures
-            ? PermanentFailures.Select(AcknowledgementException.FromIdErrorPair).ToList()
+            ? PermanentFailures.Select(AcknowledgementException.FromIdPermanentErrorPair).ToList()
             : Enumerable.Empty<AcknowledgementException>();
+
+    /// <summary>
+    /// Gets the sequence of <see cref="AckNackResponse"/>
+    /// from the current instance of <seealso cref="AckError"/>.
+    /// It returns the AckNackResponse of only the successful or permanently failed message IDs.
+    /// </summary>
+    /// <remarks>
+    /// This method will always return a sequence of <see cref="AckNackResponse"/>. This sequence will be empty (but not null) if there are no successful responses or permanent failures
+    /// in the current instance of <seealso cref="AckError"/>.
+    /// </remarks>
+    /// <returns>The sequence of <see cref="AckNackResponse"/> if there are successful responses or permanent failures, empty otherwise.</returns>
+    internal IEnumerable<AckNackResponse> GetAckNackResponses() =>
+        PermanentFailures.Select(AckNackResponse.FromIdPermanentErrorPair)
+        .Concat(SuccessfulIds.Select(AckNackResponse.FromSuccessfulId));
 
     /// <summary>
     /// Gets the <see cref="AckError"/> from the <see cref="RpcException"/> while calling Acknowledge or ModifyAckDeadline RPC.
@@ -100,7 +129,7 @@ internal sealed class AckError
         if (retryAll)
         {
             // Return all ids in temporary failures and empty permanent failures, so that the caller can retry the whole temporary batch.
-            return new AckError(ids.ToDictionary(key => key, value => rpcException.Status.StatusCode.ToString()), Enumerable.Empty<KeyValuePair<string, string>>());
+            return new AckError(ids, ids.ToDictionary(key => key, value => rpcException.Status.StatusCode.ToString()), Enumerable.Empty<KeyValuePair<string, string>>());
         }
 
         // We can have complete or partial failure. We classify the error as temporary or permanent
@@ -121,13 +150,13 @@ internal sealed class AckError
         if (temporaryErrors.Any() || permanentErrors.Any())
         {
             // If temporary or permanent error messages are present, it means the entire request hasn't failed with a permanent error.
-            return new AckError(temporaryErrors, permanentErrors);
+            return new AckError(ids, temporaryErrors, permanentErrors);
         }
         else
         {
             // Now, if temporary or permanent error messages are *not* present, it means the entire request has failed with a permanent error.
             // It may be due to a gRPC error like a failed precondition or a permission denied error or other permanent error.
-            return new AckError(Enumerable.Empty<KeyValuePair<string, string>>(), ids.ToDictionary(key => key, value => rpcException.Status.StatusCode.ToString()));
+            return new AckError(ids, temporaryErrors, ids.ToDictionary(key => key, value => rpcException.Status.StatusCode.ToString()));
         }
     }
 
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckNackResponse.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckNackResponse.cs
@@ -1,0 +1,70 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using System.Collections.Generic;
+
+namespace Google.Cloud.PubSub.V1;
+
+/// <summary>
+/// A public representation of a response to an ACKnowledge (ACK) or Not ACKnowledge (NACK) request.
+/// </summary>
+public sealed class AckNackResponse
+{
+    /// <summary>
+    /// The ID of the message being ACK-ed or NACK-ed.
+    /// </summary>
+    public string MessageId { get; }
+
+    /// <summary>
+    /// The status of the ACK/NACK response.
+    /// </summary>
+    public AcknowledgementStatus Status { get; }
+
+    /// <summary>
+    /// The failure message, if any. This will be null if the response is successful.
+    /// </summary>
+    public string FailureMessage { get; }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AckNackResponse"/> class.
+    /// </summary>
+    /// <param name="messageId">The ID of the message being ACK-ed/NACK-ed.</param>
+    /// <param name="status">The status of ACK/NACK response.</param>
+    /// <param name="failureMessage">The failure message. This can be null if ACK/NACK is successful.</param>
+    public AckNackResponse(string messageId, AcknowledgementStatus status, string failureMessage)
+    {
+        MessageId = messageId;
+        Status = status;
+        FailureMessage = failureMessage;
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AckNackResponse"/> class from the pair of message ID and its corresponding permanent failure.
+    /// </summary>
+    /// <param name="idPermanentErrorPair">The <see cref="KeyValuePair{TKey, TValue}"/> of message ID as key and permanent error message as value.</param>
+    /// <returns>A new instance of the <see cref="AckNackResponse"/>.</returns>
+    internal static AckNackResponse FromIdPermanentErrorPair(KeyValuePair<string, string> idPermanentErrorPair)
+    {
+        var acknowledgementException = AcknowledgementException.FromIdPermanentErrorPair(idPermanentErrorPair);
+        return new AckNackResponse(idPermanentErrorPair.Key, acknowledgementException.Status, acknowledgementException.Message);
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AckNackResponse"/> class from the message ID that is successfully ACK-ed/NACK-ed.
+    /// </summary>
+    /// <param name="successfulId">The successful message ID.</param>
+    /// <returns>A new instance of the <see cref="AckNackResponse"/>.</returns>
+    internal static AckNackResponse FromSuccessfulId(string successfulId) =>
+        new AckNackResponse(successfulId, AcknowledgementStatus.Success, default);
+}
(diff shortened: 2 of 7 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1528,14 +1542,16 @@ void RetryAcksAndNacks(bool ack, AckError ackError)
                         RetryTemporaryFailures(idsToRetry, acksToRetry => queue.Locked(() => queue.Requeue(acksToRetry)), Enumerable.Empty<TimedId>(), null, 600);
                     }
 
-                    if (ackError.GetPermanentExceptions() is IEnumerable<AcknowledgementException> exception && exception.Any())
-                    {
-                        // For acks/nacks, bubble up the exception.
-                        throw new AggregateException(exception);
-                    }
+                    // TODO: Uncomment this code and log the permanent errors, while adding logging. This is commented as executing this path is not required without logging.
```
maheshgattani, line 1545: Do you mean to remove this before submitting?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1546,15 +1562,15 @@ void RetryExtends(IEnumerable<TimedId> extendIds, AckError ackError)
                         RetryTemporaryFailures(Enumerable.Empty<string>(), null, extendIdsToRetry, extendsToRetry => _extendQueue.Locked(() => _extendQueue.Requeue(extendsToRetry)), 10);
                     }
 
-                    if (ackError.GetPermanentExceptions() is IEnumerable<AcknowledgementException> exception && exception.Any())
-                    {
-                        // Don't bubble permanent failures for lease extends as they are not user initiated.
-                        // Just log them.
-                        // TODO: Log the permanent failures.                           
-                    }
+                    // TODO: Uncomment this code and log the permanent errors, while adding logging. This is commented as executing this path is not required without logging.
```
maheshgattani, line 1565: Same comment as above.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: I was planning to keep it so that we don’t lose track of it but happy to change based on feedback.

maheshgattani: We do not leave commented code in our repos, but is there a precedence or understanding in the client libs that its okay?

Rishabh-V: Sure. I shall remove it.  Will just leave a TODO, to keep track of it.

amanda-tarafa: No, this is not usual in the .NET library repos either.

@Rishabh-V are these instructions for us if we need to diagnose issues? Users usually will add a dependency on the library, i.e. compiled code, so this will hardly be of any use to them.

I'd like to wait if @jskeet has a better idea but I think that if anything, we just document this for ourselves somewhere (can we do that in a test for instance)? If, as the comment says, the errors are already available to the users, why would we want to log them? Users can already decide for themselves whether to log them or not, and so can we if we need to diagnose some failures.

Rishabh-V: Hey @amanda-tarafa - Earlier we were planning to log the permanent failures, but with the updated approach errors are already available to the user, so I had commented the code and added a TODO.  In the new commit, I have removed the commented code based on feedback. Thanks. cc: @maheshgattani

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1701,6 +1722,54 @@ private void HandleAckResponse(Task writeTask, List<string> ackIds, List<string>
...
+                            // Non-exactly once delivery scenario with RpcException.
+                            // All RpcExceptions are either recoverable or ignored. Non-RpcException would throw.
+                            // As a requirement from Pub/Sub team, all RpcExceptions in non-exactly once delivery scenario should be treated as fire and forget,
+                            // so we always treat them as successful (even when they are not). If there are recoverable errors, the ACK/NACK operation
+                            // will be retried, so the message may be sent multiple times with success status.
+                            return GetSuccessResponses(ids);
+                        }
+                        else
+                        {
+                            // Non-gRPC error will throw exception eventually, so don't return those IDs as success. Return an empty list.
```
pradn, line 1758: Can you explain this?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    /// <summary>
+    /// This handler function is called when an ACKnowledge (ACK) operation returns, indicating the status of successful or permanently failed messages.
+    /// </summary>
+    /// <param name="responses">The read-only collection of the <see cref="AckNackResponse"/>
+    /// indicating the <see cref="SubscriberClient.Reply.Ack"/> status of successful or permanently failed messages. This collection may be empty but never null.
+    /// For exactly once delivery subscription, this collection contains only the successful or permanently failed messages. In this case,
+    /// the messages that have failed with temporary errors are not included in this collection as their status may change in the future.
+    /// For non-exactly once subscriptions, this collection contains the ACK status as success for all the messages. In this case,
+    /// a message may appear multiple times with the same status.</param>
```
pradn, line 49: This is only if multiple acks were called for the same message, right? Please clarify this.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+    /// The users of subscription must implement this function to handle the messages.
+    /// </remarks>
+    public abstract Task<Reply> HandleMessage(PubsubMessage message, CancellationToken cancellationToken);
+
+    /// <summary>
+    /// This handler function is called when an ACKnowledge (ACK) operation returns, indicating the status of successful or permanently failed messages.
+    /// </summary>
+    /// <param name="responses">The read-only collection of the <see cref="AckNackResponse"/>
+    /// indicating the <see cref="SubscriberClient.Reply.Ack"/> status of successful or permanently failed messages. This collection may be empty but never null.
+    /// For exactly once delivery subscription, this collection contains only the successful or permanently failed messages. In this case,
```
pradn, line 46: We should strongly suggest the user implement and handle ack/nack failures for EOS subs. The EOS guarantees don't hold if they don't deal with permanent ack/nack failures.

pradn (changes requested):

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
Rishabh-V: For non exactly once subscription, as discussed in design doc, we are treating the ACK/NACK request as fire and forget for compatibility with other libraries, so we always send the Success status for ACK/NACK even if there may be a recoverable error like Deadline exceeded. In existing C# implementation, if there is a recoverable error like Deadline exceeded, the message is requeued for ACK/NACK. In such cases, the same message will appear in the ResponseHandler again and the status of request will be success. This will happen only in cases where the handler is used for non exactly once delivery subscriptions as we are not sharing the right status. In exactly once delivery subscription, we share the status of success and permanently failed messages only which are final, so it won't happen there.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckError.cs
```diff
@@ -81,9 +96,23 @@ internal AckError(IEnumerable<KeyValuePair<string, string>> temporaryFailures, I
...
+    /// It returns the AckNackResponse of only the successful or permanently failed message IDs.
+    /// </summary>
+    /// <remarks>
+    /// This method will always return a sequence of <see cref="AckNackResponse"/>. This sequence will be empty (but not null) if there are no successful responses or permanent failures
+    /// in the current instance of <seealso cref="AckError"/>.
+    /// </remarks>
+    /// <returns>The sequence of <see cref="AckNackResponse"/> if there are successful responses or permanent failures, empty otherwise.</returns>
+    internal IEnumerable<AckNackResponse> GetAckNackResponses() =>
+        PermanentFailures.Select(AckNackResponse.FromIdPermanentErrorPair)
+        .Concat(SuccessfulIds.Select(AckNackResponse.FromSuccessfulId)).ToList();
```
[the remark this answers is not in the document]
Rishabh-V: ToList() does eager evaluation while IEnumerable<T> is lazy. But given the caller of this method is already calling ToList(), I have removed ToList();

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
pradn: Which remarks section? I don't see a change for the comments for this method, nor for the class comment.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
pradn: Acks/modacks should not be retried if EOS is disabled. Please change the behavior for non-EOS subscriptions. The value of retrying a fire-and-forget operation like ack/modack upon DeadlineExceeded is negligible.

So, for EOS and non-EOS subscriptions, each status returned should correspond to one user-initiated action.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
pradn: I see, thanks for the clear explanation. Can you give me an example of a non-GRPC error for ack/modack?

pradn (changes requested):

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
Rishabh-V: This is in `HandleAckResponse` method in `SubscriberClient` class (line 1690), which has not been modified as part of current EOD changes and has been existing this way in C# library. It checks if the error is recoverable, i.e. StatusCode of gRPC exception is one of DeadlineExceeded, ResourceExhausted, Unavailable, Cancelled, Aborted etc, it treats that error as recoverable and re-queues the ackId/nackId/extendId for non exactly once delivery. If this is not the right behavior, I think we can track this as an issue and fix that in a different PR, so that exactly once changes and non-exactly once changes can be differentiated and we would be able to recollect later why the change is done. (I can do that change immediately after this exactly once delivery PR) Please let me know if you have any concerns with this. 
@jskeet - Please share if you have insights about this behavior and your thoughts on handling this change.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
Rishabh-V: Please see line number 50-55. It has a remarks tag and documentation. Please let me know if any additional message needs to be added and I can do it.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: There isn't one which I have observed or known. Essentially, the code just caters to the case that if non-gRPC error is thrown (if any) what would be the behavior.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
pradn: We need to indicate that when EOS is turned on, users SHOULD to implement that method. Otherwise it is optional.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
pradn: Yes, let's do this in a separate PR. You can copy this discussion into the issue for that one.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
pradn: We go into this path if writeTask is faulted and if `writeTask.Exception.As<RpcException>()` returns null. I believe this should logically not happen since GRPC should return only RpcExceptions. I don't think we can assume that an exception will come up elsewhere.

Since this path should not happen, per my understanding, can we instead log an error when it does? Or we can assert that `writeTask.Exception.As<RpcException>()` always returns non-null.

pradn (changes requested):

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: This code change (to only throw non-gRPC errors) was done earlier this year. Please see this [PR and discussion]( https://github.com/googleapis/google-cloud-dotnet/pull/8287#pullrequestreview-927576158). If we need to change this behavior, I am happy to change it. I can club it in the new issue, document the desired expected behavior in the issue and then do it as part of the next PR. Please let me know.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
Rishabh-V: Updated the documentation.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
Rishabh-V: Thanks Prad. I will create a new issue and refer this discussion.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckNackResponse.cs
```diff
@@ -0,0 +1,70 @@
...
+// limitations under the License.
+
+using System.Collections.Generic;
+
+namespace Google.Cloud.PubSub.V1;
+
+/// <summary>
+/// A public representation of a response to an ACKnowledge (ACK) or Not ACKnowledge (NACK) request.
+/// </summary>
+public class AckNackResponse
```
jskeet, line 22: Let's consider sealing this class.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1502,20 +1516,17 @@ private void HandleAckResponseForExactlyOnceDelivery(Task writeTask, List<string
                 // All methods below are local methods applicable to exactly once delivery subscriptions only.
                 // They are not used in non-exactly once delivery subscriptions.
 
-                // This method schedules the retry of ackIds with temporary errors and
-                // handles the permanent errors. Permanent errors are propagated to the caller for acks.
+                // This local method for exactly once delivery only, schedules the retry of ackIds with temporary errors.
```
jskeet, line 1519: There's a verb missing in "This local method for exactly once delivery only" here and below.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -1654,6 +1651,30 @@ void UpdateRetryableIds(IEnumerable<string> allIds, AckError ackError)
 
             private void HandleAckResponse(Task writeTask, List<string> ackIds, List<string> nackIds, List<TimedId> extendIds)
             {
+                bool hasAckIds = ackIds?.Count > 0;
+                bool hasNackIds = nackIds?.Count > 0;
+                var ids = hasAckIds ? ackIds
+                    : hasNackIds ? nackIds
+                    : Enumerable.Empty<string>().ToList();
```
jskeet, line 1658: Let's just use `new List<string>()` rather than calling an extension method on an empty sequence. Ditto below.

jskeet (commented):
Just a few comments along the way, but I'll schedule a meeting to discuss things further.
(Fundamentally I don't think I have enough context to go into the details here, and not enough time to get fully versed with the context.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Added "used" at the relevant places.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/AckNackResponse.cs
Rishabh-V: Done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Done.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriptionHandler.cs
pradn: This issue is tracked in https://github.com/googleapis/google-cloud-dotnet/issues/8992. We can close this comment.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
pradn: Please add a comment with the summary of our offline discussion. We can close this comment after that.

Rishabh-V: As discussed offline, updating the details for non-gRPC errors here. In case of any exception in the response of ack/nack request, we try and cast the exception to RpcException. If that is successful, it is a gRPC error else it will be a non-gRPC error. In case of non-gRPC error, we explicitly throw an exception [Please see here](https://github.com/googleapis/google-cloud-dotnet/pull/8973/files#:~:text=throw%20writeTask.Exception.FlattenIfPossible()%3B). Sharing the status of AckIds/NackIds with non-gRPC error as success would be incorrect as after a while exception would be thrown, hence  rather than passing the Status of these IDs as success, we send an empty list.

pradn (approved):
LGTM!

jskeet (approved):

PR #8980, opened 2022-08-09
ci: Updates Document AI V1 pre-generation script.

amanda-tarafa:
At least in some protos, the Location resource definition is the last resource definition so the previous scripts has stopped working.
This should fix the issues in #8979.


amanda-tarafa:
If this works I'll make the same change to Document AI V1Beta3.

(`-z` is GNU sed only, and I'm not sure if that's what we have).

jskeet (approved):

PR #8982, opened 2022-08-09
build: Remove duplicate and outdated dependency from Firestore.

amanda-tarafa:
Not sure how we missed this for 2 months.
The push action from #8981 failed because of this.


amanda-tarafa:
Hmmm, diff detection is failing because the package to compara to has duplicate dependencies, which, yes, it's what this PR is trying to fix.
Since it's failing, and not marking the change as breaking, the `Allow breaking change label` is not of use here. @jskeet you'll have to force merge I think.

jskeet:
As you say: how did we miss this? How has CI been building? That's really odd. Merging now.

amanda-tarafa:
> As you say: how did we miss this? How has CI been building? That's really odd. Merging now.

I have no idea. And I just checked several builds from last week, both unit and integration, Firestore is being run, it builds and tests pass normally.

PR #8984, opened 2022-08-10
build: Update location dependencies in Document AI.

amanda-tarafa:
- Updates preprocessing script to remove the resource definition for the documentai.googleapis.com/Location type without relying on it not being the last resource definition.
- Add Google.Cloud.Location dependencies as Location mixins are being generated now.
- This change apply to both Document AI V1 and V1Beta3 (the CL for V1Beta3 is coming).


amanda-tarafa:
#8980 didn't work, I'm guessing we don't have GNU sed after all.

The changes in this PR do work locally, and the libraries are generated correctly etc. Once this is merged, #8979 could be regenerated.

jskeet (approved):

PR #8987, opened 2022-08-15
feat: Regenerate DocumentAI manually

jskeet:
feat: Added field_mask to ProcessRequest object in document_processor_service.proto
feat: Added parent_ids to Revision object in document.proto
feat: Added integer_values, float_values and non_present to Entity object in document.proto
feat: Added corrected_key_text, correct_value_text to FormField object in document.proto
feat: Added OperationMetadata resource
feat!: Added Processor Management and Processor Version support to v1 library
docs: fix minor docstring formatting

BREAKING CHANGE: Changed the name field for ProcessRequest and BatchProcessorRequest to accept * so the name field can accept Processor and ProcessorVersion.


jskeet:
I'm expecting to see the breaking change cause a failure here - I'll create a 3.0.0 release after this is merged. (I shouldn't have released 2.0.0, unfortunately - the breaking change was originally discussed in May, but I'd forgotten about it by the time we did the GAX v4 releases.)

jskeet:
I've now added a commit to update the version to 3.0.0-beta00, which should mean that the breaking change doesn't cause a GitHub action issue. We can merge this, then do a release.

jskeet:
Merging this without review as it's all generated anyway.

PR #9023, opened 2022-08-25
chore: Pin Python dependencies with hashes

jskeet:
The Dockerfile needs to be changed significantly more, to use .NET 6 etc - but this commit is just about pip


jskeet:
Once this is in, we should do a release of some utility docs to check that works, then a package release. (The utility docs can find problems without worrying about a bogus tag etc.)

amanda-tarafa (approved):
LGTM

PR #9030, opened 2022-08-26
chore: Update pip requirements again

jskeet:
It looks like some dependency needs typing-extensions on Python 3.7
(which is what our Kokoro machine has right now).


amanda-tarafa (approved):
LGTM

PR #9062, opened 2022-09-01
chore: Remove Google.Cloud.BigQuery.DataExchange.Common

jskeet:
This has been removed from the protos, and the beta API no longer
uses it. The package will be delisted.


jskeet:
The CI failure is expected - we never got round to adjusting the CI scripts to handle API deletion, as it's so rare.

jskeet:
Scratch that - it's complaining about the remaining package. Fixing...

jskeet:
Now fixed - there was a reference to the now-deleted project file in the solution for the beta API.

amanda-tarafa (approved):
LGTM

PR #9071, opened 2022-09-02
tools: Change where the friendlyApiName item ends up in the devsite YAML

jskeet:
Instead of being directly in the root node, it should be in the first element
of the items root node.


amanda-tarafa (approved):
LGTM

PR #9089, opened 2022-09-09
feat: Retry conformance base code

anshuldavid13:
This is the first draft of Retry conformance base code. It still needs some more sanitation and optimizations, but would appreciate you feedback on this. 
Main TODO - Object test cases and fixing of one exception received from testbench. Most other test cases are working already.
Created this separate branch so that other individual tasks can be worked on and PRs can be raised against this code by other people while we collaborate to finish this.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/Google.Cloud.Storage.V1.RetryConformanceTests.csproj
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/Google.Cloud.Storage.V1.RetryConformanceTests.csproj
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+	<Project Sdk="Microsoft.NET.Sdk">
+	<PropertyGroup>
+		<TargetFrameworks>netcoreapp3.1;net462</TargetFrameworks>
+		<TestTargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">netcoreapp2.1</TestTargetFrameworks>
+		<IsPackable>false</IsPackable>
+		<NoWarn>1701;1702;1705;xUnit2004;xUnit2013;AD0001</NoWarn>
+	</PropertyGroup>
+
+	<ItemGroup>
+		<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.0.0" />
+		<PackageReference Include="xunit" Version="2.4.1" />
+		<PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+			<PrivateAssets>all</PrivateAssets>
+		</PackageReference>
+		<PackageReference Include="coverlet.collector" Version="3.1.0">
+			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+			<PrivateAssets>all</PrivateAssets>
+		</PackageReference>
+	</ItemGroup>
+
+	<ItemGroup>
+		<ProjectReference Include="..\Google.Cloud.Storage.V1.Tests\Google.Cloud.Storage.V1.Tests.csproj" />
+	</ItemGroup>
+</Project>
+
+<!-- TODO: Update the versions as required -->
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
@@ -1,0 +1,370 @@
+// Copyright 2022 Google LLC
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     https://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Apis.Storage.v1.Data;
+using Google.Cloud.ClientTesting;
+using Google.Cloud.Storage.V1.Tests.Conformance;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net.Http;
+using System.Text;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Storage.V1.RetryConformanceTests;
+
+public class RetryConformanceTest
+{
+    public static TheoryData<RetryTest> RetryTestData { get; } = StorageConformanceTestData.TestData.GetTheoryData(f => f.RetryTests);
+
+    internal static string TestBenchUrl { get; } = GetEnvironmentVariableOrDefault("TEST_BENCH_URL", "https://storage-testbench-vkcain7hhq-el.a.run.app/");
+
+    internal static string ProjectId { get; } = GetEnvironmentVariableOrDefault("PROJECT_ID", "test");
+
+    internal string Topic { get; } = GetEnvironmentVariableOrDefault("TOPIC", "test-topic");
+
+    internal string AccessId { get; set; }
+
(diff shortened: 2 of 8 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/Google.Cloud.Storage.V1.RetryConformanceTests.csproj
```diff
@@ -0,0 +1,28 @@
...
+		<PackageReference Include="xunit" Version="2.4.1" />
+		<PackageReference Include="xunit.runner.visualstudio" Version="2.4.3">
+			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+			<PrivateAssets>all</PrivateAssets>
+		</PackageReference>
+		<PackageReference Include="coverlet.collector" Version="3.1.0">
+			<IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+			<PrivateAssets>all</PrivateAssets>
+		</PackageReference>
+	</ItemGroup>
```
amanda-tarafa, line 21: Some of the dependencies here are outdated with respect to Google.Cloud.Storage.V1.Tests. That's why the build action is not passing I think.
Maybe for now remove the dependencies.

@jskeet we might need to include this one in regenerateProjects?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30709.132
+# Visual Studio Version 17
```
amanda-tarafa, line 2: Is it OK that we do this now? Everyone needs to update VS.

@jskeet what do you think?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/StorageResources.cs
```diff
@@ -0,0 +1,30 @@
...
+
+internal class StorageResources : List<StorageResource>
+{
+    internal string BucketName => this.FirstOrDefault(j => j.Resource == Resource.Bucket).Value;
+
+    internal string ObjectName => this.FirstOrDefault(j => j.Resource == Resource.Object).Value;
+
+    internal string HmacKey => this.FirstOrDefault(j => j.Resource == Resource.HmacKey).Value;
+
+    internal string Notification => this.FirstOrDefault(j => j.Resource == Resource.Notification).Value;
```
amanda-tarafa, line 29: So, for a list of resources, they all have the same bucketName, objectName, HmacKey, etc.?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -0,0 +1,370 @@
...
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Storage.V1.RetryConformanceTests;
+
+public class RetryConformanceTest
+{
+    public static TheoryData<RetryTest> RetryTestData { get; } = StorageConformanceTestData.TestData.GetTheoryData(f => f.RetryTests);
+
+    internal static string TestBenchUrl { get; } = GetEnvironmentVariableOrDefault("TEST_BENCH_URL", "https://storage-testbench-vkcain7hhq-el.a.run.app/");
```
amanda-tarafa, line 36: this URL is not behind any protection that I can see. Maybe don't make it public, I can modify Kokoro so that it is included in an env variable when we are ready to switch these tests on.

amanda-tarafa (changes requested):
Can you please add docs almost everywhere explaining what each helper class is, and what their properties represent, etc. For Instance, I'm not clear why we need a StorageResource class when we already have a Resource class defined in the tests protos, that is, what does the Value property in StorageResource is?
I've left some other sparse comments but the missing docs would go a long way to helping with review.

Also, all other conformance tests, we have them on the unit test project. Are these separete because we need to connect to the test bench service? Which is OK, I just want to confirm.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -0,0 +1,370 @@
...
+        response.EnsureSuccessStatusCode();
+        var responseMessage = await response.Content.ReadAsStringAsync();
+        return JsonConvert.DeserializeObject<TestResponse>(responseMessage);
+    }
+
+    private void RunRetryTest(TestResponse response, StorageResources resources, bool preConditionsPresent)
+    {
+        AddRetryIdHeader(response.Id);
+
+        // Use method mapping to call the StorageTestBench API.
```
tritone, line 132: Not a dotnet expert but I would expect this to go with the other code used to interact with the testbench -- I see you have TestResponse in a separate class/file, but it is possible to factor out all the code for interacting with the testbench so it's one abstraction basically?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -0,0 +1,370 @@
...
+        {
+            _methodMappings.TryAdd(method.Name, GetMappedFunction(method.Name));
+        }
+    }
+
+    private MethodInvocation GetMappedFunction(string name) => name.ToLowerInvariant() switch
+    {
+        // TODO: Add all mappings from the json file and throw exception in the default case. To be optimised
+
+        "storage.buckets.delete" => new MethodInvocation(s_clientType.GetMethod(nameof(StorageClient.DeleteBucket), new System.Type[] { typeof(string), typeof(DeleteBucketOptions) })) { BucketNameRequired = true },
```
tritone, line 200: Seems clever how all of this has been parameterized!

tritone (commented):
Need to look through this in more detail, but overall seems like a good start!

anshuldavid13:
Thanks for reviewing and providing you comments. I will work & respond on all of these shortly.
I am OOO this week , Is it possible to merge these changes to this branch(not main) in the meantime so that Anurag can continue to work on his changes for the test cases and push them as well for review? I will come back and then add another PR with all the review comment changes and further cleanup of the code as it is still in progress? Will  not be merging these changes in main currently. Kindly let me know if this could be a possible solution.

jskeet:
I'm going to fetch this, squash it to a single commit to avoid merge issues, rebase off current main to make things simpler for the future, then add another commit to make it build. At that point, I'll merge it.

jskeet (approved):
Approving to merge even though we'll want some further changes later.

amanda-tarafa (approved):
LGTM so it can be merged.

My previous comments still stand. Thanks.

anshuldavid13:
Thank you! Will be getting back to the review comments

anshuldavid13:
@amanda-tarafa 
 For Instance, I'm not clear why we need a StorageResource class when we already have a Resource class defined in the tests protos, that is, what does the Value property in StorageResource is ?   
     **>>> The storageResource is a wrapper created around the resource class to attach a value to the various types of resources created (multiple resources in certain test case scenarios)**
I've left some other sparse comments but the missing docs would go a long way to helping with review.
     **>>> Added the docs, will add it the classes as well ** 

Also, all other conformance tests, we have them on the unit test project. Are these separete because we need to connect to the test bench service? Which is OK, I just want to confirm.
    ** >>> Since these are not unit test cases but conformance test and need to connect to the test bench for execution, I have added them in a seperate project **

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/StorageResources.cs
anshuldavid13: Yes, they are using the same resources for running the tests sequentially

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: Thanks it was Rishabh's idea to implement it via reflection :)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: While sending the request, we use the C# Storage Client which internally takes care of the request. The response is serialized in a TestResponse class so that next set of actions can be taken based on the response. Could you please elaborate how do you want this as one abstraction. Thanks

PR #9106, opened 2022-09-14
chore: Make the PgJsonb type internal.

Rishabh-V:


amanda-tarafa (approved):
LGTM
@Rishabh-V wait for @jskeet to confirm, but yes, this is the only public change in the original PR so we should be good.

jskeet (approved):

jskeet:
Note: when we do a Spanner release, we should make sure we remove the change history element that would refer to this.

PR #9140, opened 2022-09-19
feat: Integration tests for PostgreSQL dialect.

Rishabh-V:
This draft PR contains the following:

- First pass at the integration tests for PostgreSQL dialect.
- Created a base class SpannerTestDatabaseBase from which SpannerTestDatabase and SpannerTestDatabasePostgreSql derives.
- Added Traits to all the test classes, so that test suites can be run based on category. This is again a first pass and not exhaustive.
- Updated .sh and .yml to not execute the tests that cannot be run on emulator.

The purpose of this PR is to gather feedback on how we want to go ahead with the Traits, base class and PostgreSQL dialect. I can split the changes into multiple commits/PRs after I gather feedback from this PR.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/Constants.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/Constants.cs
@@ -1,0 +1,26 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+namespace Google.Cloud.Spanner.Data.CommonTesting;
+
+public static class Constants
+{
+    public const string Category = nameof(Category);
+    public const string GSQL = nameof(GSQL);
+    public const string PostgreSQL = nameof(PostgreSQL);
+    public const string Admin = nameof(Admin);
+    public const string SkipOnEmulator = nameof(SkipOnEmulator);
+    public const string Yes = nameof(Yes);
+    public const string No = nameof(No);
+}
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2018 Google LLC
+// Copyright 2018 Google LLC
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,13 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using Google.Api.Gax;
-using Google.Api.Gax.ResourceNames;
 using Google.Cloud.ClientTesting;
-using Google.Cloud.Spanner.Admin.Instance.V1;
-using Google.Cloud.Spanner.Common.V1;
-using Google.Cloud.Spanner.V1.Internal.Logging;
-using Grpc.Core;
+using Google.Cloud.Spanner.Admin.Database.V1;
 using System;
 
 namespace Google.Cloud.Spanner.Data.CommonTesting
@@ -28,9 +23,8 @@ namespace Google.Cloud.Spanner.Data.CommonTesting
     /// (This may be used by multiple fixtures, each created and then disposed, within the same test run.)
     /// A tool is provided to clean up test databases.
     /// </summary>
-    public class SpannerTestDatabase
+    public class SpannerTestDatabase : SpannerTestDatabaseBase
     {
-        private static readonly object s_lock = new object();
         private static SpannerTestDatabase s_instance = null;
 
         /// <summary>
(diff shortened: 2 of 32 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -240,22 +242,25 @@ public async Task BindNumericEmptyArray()
...
+        [Trait(Constants.SkipOnEmulator, Constants.Yes)]
         public async Task BindJsonArray()
         {
             Skip.If(_fixture.RunningOnEmulator, "The emulator doesn't yet support the JSON type");
             await TestBindNonNull(
                 SpannerDbType.ArrayOf(SpannerDbType.Json),
                 new string[] { "{\"key\":\"value\"}", null, "{\"other-key\":\"other-value\"}" });
         }
-
+
```
Rishabh-V, line 261: I will fix the whitespace issues and the commented PgJsonb code as it is not supported yet in the actual PR.

Rishabh-V (commented):
The following inputs will help:

- Do we need more categories like "Admin", "Stress Tests", etc so that we could run only those?  Currently, I just came up with basic minimum.
- Feedback on Pg dialect setup and tests.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            Logger.DefaultLogger.Debug($"Instance {SpannerInstance} created successfully.");
+        }
+        catch (RpcException e) when (e.StatusCode == StatusCode.AlreadyExists)
+        {
+            // Ignore
+        }
+    }
+
+    protected virtual void CreatePostgreSqlDatabase()
```
jskeet, line 88: Is there a reason this is in the base class? It feels odd to have PG-specific code in the base class.

Would it make sense to have a method here that accepts the DatabaseDialect as a parameter, to avoid duplication?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabasePostgreSql.cs
```diff
@@ -0,0 +1,88 @@
...
+    private static readonly string s_generatedDatabaseName = IdGenerator.FromDateTime(prefix: "testdb_pg_", pattern: "yyyyMMdd't'HHmmss");
+
+    public override string SpannerDatabase { get; } = GetEnvironmentVariableOrDefault("TEST_SPANNER_POSTGRESQL_DATABASE", s_generatedDatabaseName);
+
+    // This is the simplest way of checking whether the environment variable was specified or not.
+    // It's a little ugly, but simpler than the alternatives.
+    /// <inheritdoc/>
+    public override bool Fresh => SpannerDatabase == s_generatedDatabaseName;
+
+    private SpannerTestDatabasePostgreSql(string projectId)
```
jskeet, line 58: Other than the call to CreatePostreSqlDatabase (which could be parameterised) it doesn't look like there's anything Postgres-specific here. I suspect we can move this into the base class (and remove the equivalent code from SpannerTestDatabase)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
```diff
@@ -101,72 +81,15 @@ private SpannerTestDatabase(string projectId)
             MaybeCreateInstanceOnEmulator(projectId);
             if (Fresh)
             {
-                using (var connection = new SpannerConnection(NoDbConnectionString))
-                {
-                    var createCmd = connection.CreateDdlCommand($"CREATE DATABASE {SpannerDatabase}");
-                    createCmd.ExecuteNonQuery();
-                    Logger.DefaultLogger.Debug($"Created database {SpannerDatabase}");
-                }
+                using var connection = new SpannerConnection(NoDbConnectionString);
```
jskeet, line 84: This looks like very different code to the code used to create a Postgres database. I'd expect us to be able to use the same for both.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -240,22 +242,25 @@ public async Task BindNumericEmptyArray()
             new DateTime?[] { });
 
         [SkippableFact]
+        [Trait(Constants.SkipOnEmulator, Constants.Yes)]
```
jskeet, line 245: With the trait in place, can we remove the Skip.If calls? (It feels like a recipe for inconsistency, requiring both.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/Constants.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace Google.Cloud.Spanner.Data.CommonTesting;
+
+public static class Constants
+{
+    public const string Category = nameof(Category);
+    public const string GSQL = nameof(GSQL);
+    public const string PostgreSQL = nameof(PostgreSQL);
+    public const string Admin = nameof(Admin);
+    public const string SkipOnEmulator = nameof(SkipOnEmulator);
+    public const string Yes = nameof(Yes);
```
jskeet, line 24: What are these Yes and No for? (It's not obvious to me what `[Trait(Constants.SkipOnEmulator, Constants.Yes)]` means for example.

Doc comments on all of these would be useful.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/GetSchemaTableTestsPostgreSql.cs
```diff
@@ -0,0 +1,109 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.IntegrationTests;
+
+[Collection(nameof(AllTypesTableFixturePostgreSql))]
+[CommonTestDiagnostics]
+[Trait(Constants.Category, Constants.GSQL)]
```
jskeet, line 27: Either this trait is wrong, or I don't understand them... or both.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/V1/ReliableStreamReaderTests.cs
```diff
@@ -22,6 +23,7 @@ namespace Google.Cloud.Spanner.V1.IntegrationTests
 {
     [Collection(nameof(ReadTableFixture))]
     [CommonTestDiagnostics]
+    [Trait(Constants.Category, Constants.GSQL)]
```
jskeet, line 26: It's not clear to me that we need a trait on *all* tests like this. I suspect this PR could be a lot smaller in terms of the number of files if we could clarify the meaning of the traits - I'm hoping a lot of tests don't need a trait at all.

jskeet (commented):
Some initial thoughts. Overall, I could do with a bit more of an idea of how we expect to use these. It would be good to put some more detail around that into the design doc.

Rishabh-V:
Thanks Jon. I will fix the base class and code duplication issues in next commit. Will update the document with thoughts around the Traits usage and update this PR based on the feedback there. Thanks.

Rishabh-V:
I will close this after the PR for integration tests for PostgreSQL is raised.

Rishabh-V:
Closing for #9164

PR #9143, opened 2022-09-20
feat: Generate Google.Maps.AddressValidation.V1

jskeet:
We won't merge this until we've had the go-ahead, but this is just in preparation.


jskeet:
Okay, I would like to merge this - we may still end up not releasing it, but I don't want it sitting around forever.

amanda-tarafa (approved):
LGTM

PR #9160, opened 2022-09-27
feat:  Disabled default ExponentialBackOffPolicy in message handler via StorageClientBuilder

anuragsrivstv:


jskeet (commented):
Do we actually need this to be configurable? Given that we're going to handle retrying ourselves, is there any reason not to just set it to none *always*?

anuragsrivstv:
> Do we actually need this to be configurable? Given that we're going to handle retrying ourselves, is there any reason not to just set it to none _always_?

I thought it might be useful to have configurable retry policy, may be for any unknown future use case, so made it configurable, but very happy to set it to None if that would be the better.

tritone (commented):
In other languages we have changed various aspects of retry policy to be configurable. Here are the options we provide for Go for example: https://pkg.go.dev/cloud.google.com/go/storage#RetryOption

We would expect the same level of configurability in dotnet. (I don't know enough to know whether this PR is the way to achieve this, though.)

jskeet:
@tritone: This PR is to disable the "default" retry handling inherited for any .NET REST library. I would expect any Storage-specific configuration to be very separate to that.

jskeet:
Please rename the PR and the the commit to describe the updated change (and note that it's Exponential, not Exponantial)

anuragsrivstv:
> Please rename the PR and the the commit to describe the updated change (and note that it's Exponential, not Exponantial)

Done.

jskeet (approved):

PR #9161, opened 2022-09-27
feat: using xUnit Trait to filter not supported tests on emulator

Rishabh-V:
This PR contains just the xUnit Trait decorated on the tests that are not supported on the emulator and the corresponding changes to filter unsupported tests in the .yml and shell script. Incorporates feedback from #9140 and the design document. Please share your feedback.

I will raise a new PR for the PostgreSQL infrastructure and tests.

Thanks.

### DIFF
--- a/.github/workflows/spanner-emulator-pr-push.yml
+++ b/.github/workflows/spanner-emulator-pr-push.yml
@@ -44,5 +44,5 @@ jobs:
           SPANNER_EMULATOR_HOST: localhost:9010
           TEST_PROJECT: emulator-test-project
       run: |
-        dotnet test -c Release ./apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests
+        dotnet test -c Release ./apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests --filter SupportedOnEmulator\!=No
         dotnet test -c Release ./apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Snippets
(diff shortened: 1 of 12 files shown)


jskeet (approved):
Looks fine. I was going to suggest that we could keep the Skip.If calls, but if the idea is that we can use traits for whole classes as well, and those wouldn't be handled by Skip.If, we might as well just use a single approach.

Rishabh-V:
> Looks fine. I was going to suggest that we could keep the Skip.If calls, but if the idea is that we can use traits for whole classes as well, and those wouldn't be handled by Skip.If, we might as well just use a single approach.

Thanks Jon. 

Here are my observations:
1) Trait works well with the test classes as well as test methods that need not be skipped based on a condition. They don't work well with the parameterized tests (conditional skipping). There are a few tests which accepts different types and need to be skipped only when a condition is met.
2) SkippableFact and Skip.If calls works well with conditions as well. However, if the entire test class needs to be skipped, it still needs all the individual test methods to be decorated with SkippableFact and Skip.If calls.

Since we have a few parameterized tests, we currently have a mix of both the approaches. I can understand more from you and make the required changes tomorrow. Thank you.

Rishabh-V:
> It's not really a matter of "required changes" - I was _considering_ whether we should keep `Skip.If` so that you should always be able to just "run all the tests" (and auto-skip them). But as you say, adding Skip.If to all the tests within a class feels a bit grim - so it's probably best to _just_ use xUnit traits, at least for the emulator aspect.
> 
> So feel free to merge.

Ok. Sure. Thanks Jon.

PR #9162, opened 2022-09-27
feat: Support custom GoogleCredential values in Spanner

jskeet:
This does *not* include DI; that can come in a future PR.


jskeet:
(Review one commit at a time...)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -456,7 +456,7 @@ private SpannerConnectionStringBuilder(SpannerConnectionStringBuilder other) : t
             CredentialOverride = other.CredentialOverride;
             SessionPoolManager = other.SessionPoolManager;
             EnvironmentVariableProvider = other.EnvironmentVariableProvider;
-            ConversionOptions = other.ConversionOptions;
+            // Note: ConversionOptions is populated by the connection string.
```
amanda-tarafa, line 463: Wondering if you considered implementing the `ConversionOptions` property with a similar approach to all the rest, i.e. `GetValueOrDefault`. I understand that `ConversionOptions` is a composition of several connection string parameters, so implementation of `GetValueOrDefault` wouldn't be exactly the same for `ConversionOptions` as it is for the rest, but we wouldn't have to make these exceptions here, and I believe it would be less prone to bugs in the future.

Rishabh-V (approved):
LGTM

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -468,13 +472,32 @@ private SpannerConnectionStringBuilder(SpannerConnectionStringBuilder other) : t
         /// Must not be null.
         /// </param>
         /// <param name="credentials">Optionally supplied credential to use for the connection.
-        /// If not set, then default application credentials will be used.
+        /// If this is null (the default parameter value), then default application credentials will be used.
+        /// Credentials can be retrieved from a file or obtained interactively.
+        /// See Google Cloud documentation for more information.
+        /// </param>
+        public SpannerConnectionStringBuilder(string connectionString, ChannelCredentials credentials = null) : this()
```
amanda-tarafa, line 479: I think it's better if the parameter is not optional an we add a constructor that takes the connection string only. We can then allow nulls or not for the credential parameters, but that's another consideration. (I think it's fine to allow nulls).

(Unless there's only one private constructor, I don't like constructors with optional parameters, it's a hazzle for overloading later on).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -486,6 +509,7 @@ public SpannerConnectionStringBuilder(string connectionString, ChannelCredential
         /// Must not be null.</param>
         /// <param name="credentials">The credential to use for the connection. May be null.</param>
         /// <param name="sessionPoolManager">The session pool manager to use. Must not be null.</param>
+        [Obsolete("Use overloads accepting credentials, then set the SessionPoolManager property", error: false)]
```
amanda-tarafa, line 512: We usually don't set `error: false` on obsolete attributes. (But I'm fine either way, just mentioning for consistency).

amanda-tarafa (approved):
LGTM with a couple of comments.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
jskeet: I looked into this... ConversionOptions is fetched a *lot* - I started looking at caching the constructed value along with the strings it was constructed from. I believe it's simpler to still populate it as we see the relevant properties.
(I'm actually *removing* the exception here, and instead treating it as being like any of the other properties which is derived from the connection string handling. It was already that way, but removing the assignment in the constructor handles it more uniformly.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
jskeet: This isn't a new constructor really - it's just a misleading diff. Options:

- Leave the constructor signature as it is (so no change to this PR)
- Add a SpannerConnectionStringBuilder(string) constructor, and make the argument here required. (So we'll get a compile-time change to which overload is called, but it'll do the same thing, so it's non-breaking.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
jskeet: I figured it would just be a little clearer - I can definitely remove it if we want.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
amanda-tarafa: Yes, I figured that it being used a lot might be a factor in wanting to have it permanently populated. It's fine then.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
amanda-tarafa: Ah yes, I see now that the constructor was there before.
I like the second option best. But it's fine if you merge as is.

amanda-tarafa (approved):
LGTM

PR #9164, opened 2022-09-28
feat: PostgreSQL dialect integration testing infrastructure and basic tests

Rishabh-V:


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
```diff
@@ -12,161 +12,55 @@
...
 using Google.Cloud.ClientTesting;
-using Google.Cloud.Spanner.Admin.Instance.V1;
-using Google.Cloud.Spanner.Common.V1;
-using Google.Cloud.Spanner.V1.Internal.Logging;
-using Grpc.Core;
+using Google.Cloud.Spanner.Admin.Database.V1;
 using System;
 
-namespace Google.Cloud.Spanner.Data.CommonTesting
+namespace Google.Cloud.Spanner.Data.CommonTesting;
```
Rishabh-V, line 19: I have used the file scoped namespace here, since the entire file was getting modified. Happy to revert, if it is difficult to review this way.

Rishabh-V:
Feedback from #9140 is incorporated.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,224 @@
...
+    /// </summary>
+    public virtual string SpannerPort { get; } = GetEnvironmentVariableOrDefault("TEST_SPANNER_PORT", null);
+
+    /// <summary>
+    /// The Spanner Instance name to connect to. It is read from the environment variable "TEST_SPANNER_INSTANCE".
+    /// </summary>
+    public virtual string SpannerInstance { get; } = GetEnvironmentVariableOrDefault("TEST_SPANNER_INSTANCE", "spannerintegration");
+
+    /// <summary>
+    /// The Spanner Database name to connect to. It is read from the environment variable "TEST_SPANNER_DATABASE".
```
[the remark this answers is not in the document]
amanda-tarafa: But I think you can have two protected abstract properites instead: SpannerDatabaseEnvVar, SpannerDatabaseGeneratedName. Both can return respective static fields on the derived classes, and then SpannerDatabase and Fresh can both be implemented here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,224 @@
...
+using System;
+
+namespace Google.Cloud.Spanner.Data.CommonTesting;
+
+/// <summary>
+/// Base class for Spanner Test database.
+/// </summary>
+public abstract class SpannerTestDatabaseBase
+{
+    protected static readonly object s_lock = new object();
```
amanda-tarafa, line 31: The lock shouldn't be shared across derived types.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,224 @@
...
+
+    protected SpannerTestDatabaseBase(string projectId, DatabaseDialect dialect)
+    {
+        TestLogger.Install();
+        ProjectId = projectId;
+        var builder = new SpannerConnectionStringBuilder
+        {
+            Host = SpannerHost,
+            DataSource = $"projects/{ProjectId}/instances/{SpannerInstance}",
+            EmulatorDetection = dialect == DatabaseDialect.Postgresql
```
amanda-tarafa, line 90: You have inheritance based on whether the dialect is PS or not. You shouldn't be needing to ask these questions here.

This constructor should receive a value for EmulatorDetection, not for the dialect.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            builder.Port = int.Parse(SpannerPort);
+        }
+        NoDbConnectionString = builder.ConnectionString;
+        SpannerClientCreationOptions = new SpannerClientCreationOptions(builder);
+        var databaseBuilder = builder.WithDatabase(SpannerDatabase);
+        ConnectionString = databaseBuilder.ConnectionString;
+        DatabaseName = databaseBuilder.DatabaseName;
+
+        if (dialect != DatabaseDialect.Postgresql)
```
amanda-tarafa, line 104: Here as well you can ask about the EmulatorDetection value received in the constructor and not the dialect.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            // We create database instance only on the emulator.
+            // In production, we expect the instance to already exist.
+            MaybeCreateInstanceOnEmulator(projectId);
+        }
+
+        if (Fresh)
+        {
+            CreateDatabase(dialect);
```
amanda-tarafa, line 113: And this method should receive no parameters, and be abstract in this class.
The derived classes, which know whether they are PS or not, should be able to implement this method without conditions.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,224 @@
...
+        }
+    }
+
+    protected static string GetEnvironmentVariableOrDefault(string name, string defaultValue)
+    {
+        string value = Environment.GetEnvironmentVariable(name);
+        return string.IsNullOrEmpty(value) ? defaultValue : value;
+    }
+
+    protected virtual void CreateDatabase(DatabaseDialect dialect)
```
amanda-tarafa, line 128: This should be abstract and with no parameters.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,224 @@
...
+    {
+        // PostgreSQL database can only be created by Admin APIs, which is not supported by Emulator for GSQL database creation.
+        // So, we need two different implementation for GSQL and PostgreSQL.
+        if (dialect == DatabaseDialect.Postgresql)
```
amanda-tarafa, line 132: Everything on this if should be the implementation of this method on the PS specific derived class.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,224 @@
...
+                Logger.DefaultLogger.Debug($"The database {SpannerDatabase} with dialect {dialect} created successfully.");
+            }
+            catch (RpcException e) when (e.StatusCode == StatusCode.AlreadyExists)
+            {
+                // Ignore.
+            }
+        }
+        else
+        {
+            using var connection = new SpannerConnection(NoDbConnectionString);
```
amanda-tarafa, line 161: And everything on this else should be the implementation of this method on the GoogleDialect specific derived class.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
amanda-tarafa: Actually, you probably don't even need to ask anything here, as `MaybeCreateInstanceOnEmulator` is already asking about `SpannerClientCreationOptions.UsesEmulator` which I think would reflect whatever you set on the connection builder?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
```diff
@@ -12,161 +12,55 @@
...
+
     /// <summary>
-    /// A database created on-demand for testing. This is never dropped, partly as it's hard to know when to do so.
-    /// (This may be used by multiple fixtures, each created and then disposed, within the same test run.)
-    /// A tool is provided to clean up test databases.
+    /// Fetches the database, creating it if necessary.
     /// </summary>
-    public class SpannerTestDatabase
+    /// <param name="projectId">The project ID to use, typically from a fixture.</param>
+    public static SpannerTestDatabase GetInstance(string projectId)
```
amanda-tarafa, line 34: Now that we are making changes here, please rename this method to Create, as Instance has Spanner specific meaning and it's confusing.
```suggestion
    public static SpannerTestDatabase Create(string projectId)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTestsPostgreSql.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.IntegrationTests;
+
+[Collection(nameof(SpannerDatabaseFixturePostgreSql))]
+[CommonTestDiagnostics]
+[Trait(Constants.SupportedOnEmulator, Constants.No)]
+public class BindingTestsPostgreSql
```
amanda-tarafa, line 27: I'm guessing we have an equivalent test class for Google dialect. Could we have a `BindingTestsBase` class that receives a base fixture and implements all the common tests, and then two dialect specific classes with the traits and with the tests for the dialect specific types? Just so that we don't have all the duplicated code.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/GetSchemaTableTestsPostgreSql.cs
```diff
@@ -0,0 +1,106 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.IntegrationTests;
+
+[Collection(nameof(AllTypesTableFixturePostgreSql))]
+[CommonTestDiagnostics]
+[Trait(Constants.SupportedOnEmulator, Constants.No)]
+public class GetSchemaTableTestsPostgreSql
```
amanda-tarafa, line 28: Same comment here for maybe a GetSchemaTableTestsBase class. In the base class you can have a `SchemaTestCommonData` that returns data for the shared types. And on each individual class you have a `SchemaTestData` that adds the specific data types to the TheoryData collection returned by `SchemaTestCommonData`?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTestsPostgreSql.cs
```diff
@@ -0,0 +1,348 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.IntegrationTests;
+
+[Collection(nameof(AllTypesTableFixturePostgreSql))]
+[CommonTestDiagnostics]
+[Trait(Constants.SupportedOnEmulator, Constants.No)]
+public class WriteTestsPostgreSql
```
amanda-tarafa, line 30: Same comment here about having a WriteTestsBase class.

amanda-tarafa (changes requested):
I've left several comments.

In particular, I've left comments about code duplication in tests, and suggestions on how to avoid that. But my suggestions are only valid if we decide to test PS so thoroughly. There are many tests here, for instance, most Write and Binding tests, that are not dialect specific, so I believe we shouldn't be testing all of that in both Google and PS.

I believe we agreed to test PS specifics thoroughly, like PgNumeric, PgJson, etc. and to have a couple of vanilla tests, let's say, a read and write on a Pg Database, but nothing else. Maybe I'm missing something though?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
Rishabh-V: Right Amanda. I did two things incorrectly. One was conditional creation in the base class and setting emulator detection based on dialect. I didn't understood EmulatorDetection correctly. I will make the change in the next commit.

Rishabh-V (commented):
Thanks Amanda. I will address the inheritance related comments and rather than testing the entire test suite for PostgreSQL dialect, I will write tests for only Pg specific types. This will remove the code duplication as well. Thanks. Will push the fixes in next commit by tomorrow.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
amanda-tarafa: This is a static declaration, which means there's exactly one value for it, it's not inherited, and it doesn't exist per instance.

If you have class A and class B that extends SpannerTestDatabaseBase, whe you use the lock in class A, it cannot be used in class B and viceversa, and, as you said, A and B in our case are independent, they are, let's say, mutually thread safe, so they shouldn't need to share a lock.
Tests may be run in parallel, so we would be adding unnecessary delays in this case. In other similar cases it could have worst implications of course. And yet in other cases you might really need a base class static lock to be shared amongst all inheriting classes.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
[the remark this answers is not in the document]
amanda-tarafa: What error? It's OK to use static members from within instance members.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
amanda-tarafa: Sounds good, thanks!

Rishabh-V (commented):
I have created the base classes and since most of the tests moved to the base test classes, we have tests for both the dialects. Please have a look.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerFixtureBase.cs
```diff
@@ -16,25 +16,24 @@
 using Google.Cloud.Spanner.Common.V1;
 using Google.Cloud.Spanner.V1.Internal.Logging;
 
-namespace Google.Cloud.Spanner.Data.CommonTesting
+namespace Google.Cloud.Spanner.Data.CommonTesting;
+
+/// <summary>
+/// Base classes for test fixtures for Spanner.
```
amanda-tarafa, line 22:
```suggestion
/// Base classes for test fixtures for Spanner Google dialect.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerFixtureBase.cs
```diff
@@ -16,25 +16,24 @@
...
+/// </summary>
+public abstract class SpannerFixtureBase : CloudProjectFixtureBase
```
amanda-tarafa, line 24: I think this one and SpannerFixtureBasePostgresSql can be further abstracted using generics on the Database type.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,177 @@
...
+        };
+        if (SpannerPort != null)
+        {
+            builder.Port = int.Parse(SpannerPort);
+        }
+        NoDbConnectionString = builder.ConnectionString;
+        SpannerClientCreationOptions = new SpannerClientCreationOptions(builder);
+        var databaseBuilder = builder.WithDatabase(SpannerDatabase);
+        ConnectionString = databaseBuilder.ConnectionString;
+        DatabaseName = databaseBuilder.DatabaseName;
```
amanda-tarafa, line 104: nit
```suggestion
        DatabaseName = databaseBuilder.DatabaseName;
        
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
```diff
@@ -0,0 +1,177 @@
...
+        // We create database instance only on the emulator.
+        // In production, we expect the instance to already exist.
+        MaybeCreateInstanceOnEmulator(projectId);
+
+        if (Fresh)
```
amanda-tarafa, line 109: I think this should be, it has probably always worked because where we run with the emulator the environment variable with the DB name is not set, but still, this is a bug.
```suggestion
        if (Fresh || SpannerClientCreationOptions.UsesEmulator)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabase.cs
```diff
@@ -13,160 +13,63 @@
...
-                        },
-                    }).PollUntilCompleted();
-                }
-                catch (RpcException e) when (e.StatusCode == StatusCode.AlreadyExists)
-                {
-                    // Ignore
-                }
-            }
-        }
+    public override string SpannerDatabase { get; } = GetEnvironmentVariableOrDefault("TEST_SPANNER_DATABASE", s_generatedDatabaseName);
```
amanda-tarafa, line 58:
```suggestion
    public override string SpannerDatabase { get; } = GetEnvironmentVariableOrDefault("TEST_SPANNER_DATABASE", GeneratedDatabaseName);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabasePostgreSql.cs
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    private static readonly string s_generatedDatabaseName = IdGenerator.FromDateTime(prefix: "testdb_pg_", pattern: "yyyyMMdd't'HHmmss");
+
+    /// <inheritdoc/>
+    public override string GeneratedDatabaseName => s_generatedDatabaseName;
+
+    /// <inheritdoc/>
+    public override string SpannerDatabase { get; } = GetEnvironmentVariableOrDefault("TEST_SPANNER_POSTGRESQL_DATABASE", s_generatedDatabaseName);
```
amanda-tarafa, line 61:
```suggestion
    public override string SpannerDatabase { get; } = GetEnvironmentVariableOrDefault("TEST_SPANNER_POSTGRESQL_DATABASE", GeneratedDatabaseName);
```

amanda-tarafa, line 61: But also, the things that really change here from one implementation to another are the database name prefix and the name of the env variable. Would you abstract those away instead?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -14,256 +14,77 @@
...
-        }
-
-        [Fact]
-        [Trait(Constants.SupportedOnEmulator, Constants.No)]
-        public async Task BindJsonEmptyArray()
-        {
-            // The emulator doesn't yet support the JSON type.
-            await TestBindNonNull(SpannerDbType.ArrayOf(SpannerDbType.Json), new string[] { });
-        }
+    [SkippableTheory]
```
amanda-tarafa, line 50: Is the trait not being used here because this is a theory? So we need to use skip instead?

Would it make sense to separate the test in two theories, one that is always supported and one that is not supported for the emulator so we can use the trait always? Are there other similar tests out there that we could apply the same rationale to?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/GetSchemaTableTestsPostgreSql.cs
```diff
@@ -0,0 +1,64 @@
...
+        _fixture = fixture;
+
+    protected override SpannerConnection GetConnection() =>
+        _fixture.GetConnection();
+
+    public static TheoryData<string, System.Type, SpannerDbType> SchemaTestData { get; } =
+        new TheoryData<string, System.Type, SpannerDbType>
+        {
+            // Base types.
+            { "PgNumericValue", typeof(PgNumeric), SpannerDbType.PgNumeric },
```
amanda-tarafa, line 44: Don't we have to test the Json PG as well?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTestsPostgreSql.cs
```diff
@@ -0,0 +1,69 @@
...
+    public async Task BindPgNumericArray() =>
+        await TestBindNonNull(SpannerDbType.ArrayOf(SpannerDbType.PgNumeric), new PgNumeric?[] { PgNumeric.Parse("0.0"), null, PgNumeric.Parse("1.0") });
+
+    [Fact]
+    public async Task BindPgNumericEmptyArray() =>
+        await TestBindNonNull(SpannerDbType.ArrayOf(SpannerDbType.PgNumeric), new PgNumeric[] { });
+
+    public static TheoryData<SpannerDbType> BindNullData { get; } = new TheoryData<SpannerDbType>
+    {
+        SpannerDbType.PgNumeric,
```
amanda-tarafa, line 57: And JsonPg?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionScopeTests.cs
```diff
@@ -371,7 +371,7 @@ public async Task ParallelWriteAsync()
             string[] keys = new string[] { IdGenerator.FromGuid(), IdGenerator.FromGuid(), IdGenerator.FromGuid() };
             await RetryHelpers.ExecuteWithRetryAsync(async () =>
             {
-                using var scope = new TransactionScope();
+                using var scope = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled);
```
amanda-tarafa, line 374: What's this change?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
Rishabh-V: Ok. Thanks for letting me know.

Rishabh-V (commented):
Thanks Amanda. Ack on complexity and time to run the tests.

1) I will make the changes so that base classes only has the helper, common methods and couple of good to have tests and move rest of the tests to the Google dialect based tests. This will reduce the time to run the tests.
2) PgJsonb tests are not included as PgJsonb changes will be GA on 14th October now and the tests would have failed in local. I will raise a separate PR for PgJsonb once the changes are G.A and base infrastructure is in place. I probably missed to state that clearly in our last sync up.
3) Have added few queries/observations. Please have a look and share your thoughts. 

I will raise a new commit with these items addressed based on your thoughts. Thanks.

PS: When I ran the tests in local for about 4-5 times, I observed an overall average increase in time by ~4 minutes after adding tests for Pg dialect, so I thought its a good idea to have all tests of these 3 test classes.

amanda-tarafa:
I'll take a look at your comments etc. tomorrow morning at the latest.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionScopeTests.cs
amanda-tarafa: Hmmm, we cannot change a test without understanding why.
I just ran these locally and they were green, so there might be an issue with your environment or maybe even with some of the code your are adding on this PR (although I really cannot think what).
Bottom line, it's OK to modfied tests when we understand what is happening and it makes sense, but not blindly, as we would very well be letting some bugs in.

amanda-tarafa (commented):
I've replied to your most recent comments.

About the tests for PG, I think +5 minutes is a nuisance on development to be honest, but that's really not the most important part. It's the maintenance burden as we would either need code duplication (first version of this PR) or add code complexity (current version of this PR), and there's not much to gain in my view, there's really no dialect specific library code except for the pg types, etc. I would keep it as simple as needed for now, and we can look into abstractions etc. as there's a neccesity for them.
For instance, we do need the SpannerTestDatabase hierarchy, and possibly the AllTypesTable hierarchy, but I'm not sure we need a hierarchy for the Binding, Writing and GetSchema tests. Can't we add a new class named maybe PgTests were we add all the PG specific tests? Yes, we can copy some code from the original Binding, Writing and GetSchema classes, or maybe even add some static methods in the original classes that Assert the common aspects, etc.
Not sure how that would like though, but I think it'll be cleaner and simpler than the alternatives.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
jskeet: Not sure - if we use the emulator but set up the table in a different tool for some reason, we wouldn't want to recreate it, would we?

Rishabh-V:
> I've replied to your most recent comments.
> 
> About the tests for PG, I think +5 minutes is a nuisance on development to be honest, but that's really not the most important part. It's the maintenance burden as we would either need code duplication (first version of this PR) or add code complexity (current version of this PR), and there's not much to gain in my view, there's really no dialect specific library code except for the pg types, etc. I would keep it as simple as needed for now, and we can look into abstractions etc. as there's a neccesity for them. For instance, we do need the SpannerTestDatabase hierarchy, and possibly the AllTypesTable hierarchy, but I'm not sure we need a hierarchy for the Binding, Writing and GetSchema tests. Can't we add a new class named maybe PgTests were we add all the PG specific tests? Yes, we can copy some code from the original Binding, Writing and GetSchema classes, or maybe even add some static methods in the original classes that Assert the common aspects, etc. Not sure how that would like though, but I think it'll be cleaner and simpler than the alternatives.

Thanks Amanda. As suggested by you, I will keep the SpannerTestDatabase and AllTypesTable hierarchy and get rid of all the test base classes (newly created in this PR). I will create a new test class namely PgTests that will have all the Pg dialect specific test cases and try to reuse the code by means of creating static methods in original GSQL classes. Will try to push a commit with these changes by Monday. Thanks.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
amanda-tarafa: @Rishabh-V, @jskeet and I talked about this offline and it's slightly more complicated. So that we don't mix fixing this in this PR, continue asking for Fresh only here, and please create an issue so that we remember to look at this separately. This is test code so its not a huge problem, and also, this bug wouldn't manifest itself in any of our current CI environments.

Rishabh-V: Sure. Thanks Amanda. I will log an issue.

Rishabh-V (commented):
Hi Amanda, I have addressed the comments. Wrote all tests in PgTests.cs. Updated few of the corresponding GSQL tests so that code can be reused. The execution time on an average is ~43 seconds for all the current Pg tests. PTAL. Thanks.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -40,65 +41,71 @@ public class BindingTests
...
+        };
+
+        // These SpannerDbTypes are unsupported on emulator.
+        public static TheoryData<SpannerDbType> BindUnsupportedNullData { get; } = new TheoryData<SpannerDbType>
+        {
+            SpannerDbType.Json,
             SpannerDbType.ArrayOf(SpannerDbType.Json),
         };
 
+        // TODO: xUnit v3 supports traits for DataAttributes. Use that instead of Skip when we migrate.
```
amanda-tarafa, line 61: Thanks for this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -40,65 +41,71 @@ public class BindingTests
...
+            Assert.Equal(DBNull.Value, reader.GetValue(0));
+            Assert.False(await reader.ReadAsync());
+        }
 
-                    Assert.False(await reader.ReadAsync());
-                }
-            }
+        private SpannerCommand GetSelectCommand(SpannerDbType parameterType, object value)
+        {
+            using var connection = _fixture.GetConnection();
```
amanda-tarafa, line 83: I have no idea how this works, but I think it shouldn't.
The connection is being disposed of right before returning the command, so by the time the command is executed the connection should alreayd be disposed?

Even if it's working this will be flaky I guess. And I'd very much like to understand why it's working.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -40,65 +41,71 @@ public class BindingTests
...
-                    Assert.True(await reader.ReadAsync());
+            var cmd = GetSelectCommand(parameterType, null);
+            await GetReaderAndAssertNull(cmd);
+        }
 
-                    Assert.True(reader.IsDBNull(0));
-                    Assert.Equal(DBNull.Value, reader.GetValue(0));
+        internal static async Task GetReaderAndAssertNull(SpannerCommand command)
+        {
+            using var reader = await command.ExecuteReaderAsync();
```
amanda-tarafa, line 74: I think it's better if this is done by the callers and this method just asserts the null.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -40,65 +41,71 @@ public class BindingTests
...
+            Assert.True(await reader.ReadAsync());
+            Assert.True(reader.IsDBNull(0));
+            Assert.Equal(DBNull.Value, reader.GetValue(0));
+            Assert.False(await reader.ReadAsync());
+        }
 
-                    Assert.False(await reader.ReadAsync());
-                }
-            }
+        private SpannerCommand GetSelectCommand(SpannerDbType parameterType, object value)
```
amanda-tarafa, line 81: Have this one execute the command and return the reader instead.

Mind you, you still cannot dispose of the connection etc., so probably this one needs the connection to be passed to it instead.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -40,65 +41,71 @@ public class BindingTests
...
-
-                        rowsRead++;
-                    }
+                    Assert.Equal(typeSpecificReader(reader), valueRead);
+                    Assert.False(reader.IsDBNull(0));
                 }
+
+                rowsRead++;
             }
             Assert.Equal(1, rowsRead);
```
amanda-tarafa, line 110: Do the same as in the other method so you can remove the while loop, etc.
```
Assert.True(await reader.ReadAsync());
// all the other code here
Assert.False(await reader.ReadAsync());
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionScopeTests.cs
amanda-tarafa: > TransactionScope doesn't work well with async/await and the recommended fix is to ...

Who says that TransactionScope does not work well with async/await? And who recommendas the fix? I'm just curious.

> We got rid of GetConnection() overload and used GetConnection(Logger logger = null, bool logCommitStats = false) instead.

OK, but why did this change had this side effect?

Sorry for insisting, but I don't yet understand what happened, it might be OK to make this change, but we really need to understand why.
I can take a look later as well if you prefer.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -48,6 +48,14 @@ private async Task<int> InsertAsync(SpannerParameterCollection values)
             }
         }
 
+        private SpannerCommand GetSelectCommand()
```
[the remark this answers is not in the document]
amanda-tarafa: And same suggestion as before, we probably want to pass in the connection so that the calle can just dispose of it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PgTests.cs
```diff
@@ -0,0 +1,165 @@
...
+
+    [Fact]
+    public async Task BindPgNumericArray() =>
+        await BindingTests.TestBindNonNull(SpannerDbType.ArrayOf(SpannerDbType.PgNumeric), new PgNumeric?[] { PgNumeric.Parse("0.0"), null, PgNumeric.Parse("1.0") }, GetSelectCommand);
+
+    [Fact]
+    public async Task BindPgNumericEmptyArray() =>
+        await BindingTests.TestBindNonNull(SpannerDbType.ArrayOf(SpannerDbType.PgNumeric), new PgNumeric[] { }, GetSelectCommand);
+
+    private SpannerCommand GetSelectCommand(SpannerDbType parameterType, object value)
```
amanda-tarafa, line 50: Maybe rename this one `GetSelectCommandForBinding`.
Although once these return the reader, it might be clearer, something like GetBindingTestReader or similar.

amanda-tarafa (changes requested):
Yep, I prefer this. A few comments still.

Rishabh-V (commented):
Hey Amanda, I have addressed your review comments in the last commit. I have also added my thoughts on the Transaction and GetConnection changes. That said, I will invest some time analyzing the `GetConnection` change later in my day and update the comment thread with my findings. PTAL. Thanks.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionScopeTests.cs
Rishabh-V: Update: I logged the ManagedThreadId after every line in the test, before and after making the changes in `GetConnection` and observed that before making the change the thread ID remains the same till the scope completion. After making the change, the thread ID changes before the scope.Complete() call.  This still doesn't explain why the 'GetConnection' changes triggered it. But I _think_ thread ID may or may not change after async/await call in the flow. This isn't predictable and we have one new async await call in the new `GetConnection`

amanda-tarafa: OK, having gone through the linked issue I understand and agree with having to make a change on this tests regardless. It's still not 100% clear why the issues manifested with this change, but that's not relevant. Our tests and production code is using async/await, so we should make this change.

One clarification, `SessionPoolManager.Create` is not calling `CreateClientAsync`, it's only passing it as a Func for it to be the client factory, but it's not being called there at all.

By looking at the `SessionPoolManager` code, though, I wonder if getting rid of the simple `GetConnection` method was a good idea. I think it was, as know we always have a fresh SessionPoolmanager for each test, which means tests are isolated from each other, which I believe is what should be. But still I'll let @jskeet chime in.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/GetSchemaTableTests.cs
```diff
@@ -42,33 +42,26 @@ public async Task GetSchemaTable_Default_ReturnsNull()
...
-                    Assert.Equal(spannerDbType, (SpannerDbType)row["ProviderType"]);
-                    // These fields are (currently) not filled as Spanner does not provided enough
-                    // information to fill them.
-                    Assert.True(row.IsNull("ColumnSize"));
-                    Assert.True(row.IsNull("NumericPrecision"));
-                    Assert.True(row.IsNull("NumericScale"));
-                }
-            }
+            string selectQuery = $"SELECT {columnName} FROM {_fixture.TableName}";
+            await GetSchemaTableWithFlagEnabledReturnsSchema(columnName, type, spannerDbType, _fixture.ConnectionString, selectQuery);
```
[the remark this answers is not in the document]
amanda-tarafa:
```suggestion
            await GetSchemaTable_WithFlagEnabled_ReturnsSchema_Impl(columnName, type, spannerDbType, _fixture.ConnectionString, selectQuery);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -116,25 +105,46 @@ private async Task TestBindNonNull<T>(SpannerDbType parameterType, T value, Func
             {
                 Assert.Equal(value, valueRead);
             }
+
+            Assert.False(await reader.ReadAsync());
         }
 
+        private async Task<SpannerDataReader> GetBindingTestReader(SpannerConnection connection, SpannerDbType parameterType, object value)
+        {
+            using var cmd = connection.CreateSelectCommand("SELECT @v");
```
amanda-tarafa, line 114: I'm not sure we can dispose of the command either before reading through the data reader. If I remember correctly, not all data is returned inmediately, insteadt it's lazily fetched.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PgTests.cs
```diff
@@ -0,0 +1,166 @@
...
+
+    private async Task ExecuteWriteValuesTest(Func<SpannerParameterCollection, Task<int>> insertCommand)
+    {
+        PgNumeric?[] numericArray = { PgNumeric.Parse("0.0"), null, PgNumeric.Parse("2.0") };
+
+        var parameters = new SpannerParameterCollection
+            {
+                { "PgNumericValue", SpannerDbType.PgNumeric, PgNumeric.Parse("2.0") },
+                { "PgNumericArrayValue", SpannerDbType.ArrayOf(SpannerDbType.PgNumeric), numericArray }
+            };
```
amanda-tarafa, line 142:
```suggestion
        var parameters = new SpannerParameterCollection
        {
            { "PgNumericValue", SpannerDbType.PgNumeric, PgNumeric.Parse("2.0") },
            { "PgNumericArrayValue", SpannerDbType.ArrayOf(SpannerDbType.PgNumeric), numericArray }
        };
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -48,6 +48,15 @@ private async Task<int> InsertAsync(SpannerParameterCollection values)
             }
         }
 
+        private SpannerConnection GetConnection() => _fixture.GetConnection();
+
+        private async Task<SpannerDataReader> GetWriteTestReader(SpannerConnection connection)
+        {
+            using var command = connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName} WHERE K=@k");
```
amanda-tarafa, line 55: Same here, this using is not correct here.

These tests might all be working just becuase we pull so little data that we only query the service once. But still...

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -56,18 +65,11 @@ private async Task<int> InsertAsync(SpannerParameterCollection values)
             record[_lastKey] = bytes;
         }
 
-        private async Task WithLastRowAsync(Action<SpannerDataReader> readerAction)
+        internal static async Task WithLastRowAsync(Action<SpannerDataReader> readerAction, SpannerConnection connection, Func<SpannerConnection, Task<SpannerDataReader>> getReader)
```
amanda-tarafa, line 68: Should this be as follows, or am I missunderstanding?
```suggestion
        internal static async Task WithLastRowAsync(Action<SpannerDataReader> lastRowAction, SpannerConnection connection, Func<SpannerConnection, Task<SpannerDataReader>> getReader)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -48,6 +48,15 @@ private async Task<int> InsertAsync(SpannerParameterCollection values)
             }
         }
 
+        private SpannerConnection GetConnection() => _fixture.GetConnection();
```
amanda-tarafa, line 51: Was this change necessary? (and in the other classes?)

amanda-tarafa (changes requested):
Thanks, looking better! A couple things, but nothing major.

@jskeet you might want to take a quick look.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/CloudSpannerFixtureBase.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Google.Cloud.Spanner.Data.CommonTesting;
+
+/// <summary>
+///  Base class for the test fixtures for Cloud Spanner.
+/// </summary>
+/// <typeparam name="TDatabase">The type of test database that derives from the <see cref="SpannerTestDatabaseBase"/> based on the dialect.</typeparam>
+public abstract class CloudSpannerFixtureBase<TDatabase> : CloudProjectFixtureBase where TDatabase : SpannerTestDatabaseBase
+{
+    // The type of test database derived from the SpannerTestDatabaseBase based on the dialect. 
+    public TDatabase Database { get; protected set; }
```
jskeet, line 28: Does this ever need to be set other than at the start?

If not, I'd just make it a readonly property and have a protected constructor with a TDatabase parameter. You'd then have, for instance:

```csharp
protected SpannerFixtureBasePostgreSql() : base(SpannerTestDatabasePostgreSql.Create(ProjectId)) {}
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
jskeet: Indeed. I suspect that disposing the command doesn't do much (possibly releases the session back to the pool?) but this is at least *logically* incorrect.

Given how short this is, I'd probably just inline it where you need it - but another option would be to pass in a delegate to do the reading/asserting while the command is open. Or just don't dispose the command, because I *believe* that SpannerDataReader.Dispose will dispose the command anyway.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -116,25 +105,46 @@ private async Task TestBindNonNull<T>(SpannerDbType parameterType, T value, Func
...
+            cmd.Parameters.Add("v", parameterType, value);
+            return await cmd.ExecuteReaderAsync();
+        }
+
+        internal static async Task TestBindNonNull<T>(SpannerDbType parameterType, T value, SpannerConnection connection, Func<SpannerConnection, SpannerDbType, object, Task<SpannerDataReader>> getReader, Func<SpannerDataReader, T> typeSpecificReader = null)
```
jskeet, line 119: I think it *might* be simpler (particularly in the light of the above) to keep the full implementation instead of trying to generalize. It's a very small amount of shared code, but avoiding duplication has introduced more complexity IMO. If you put the GetBindingTestReader code back into here *and* into TestBindNonNull, and just copy what you need in the Postgres tests, we don't need to worry about disposal etc.

I think I'd need to work with the code in VS to see though... I'm happy for it to go in as-is.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/GetSchemaTableTests.cs
```diff
@@ -105,12 +116,14 @@ public async Task GetSchemaTable_WithFlagEnabled_ReturnsColumnOrdinals()
                 {
                     var table = reader.GetSchemaTable();
                     // The table also contains the `K` column that is the primary key.
-                    var expectedRowCount = _fixture.RunningOnEmulator ? SchemaTestData.Count() - 1 : SchemaTestData.Count() + 1;
+                    var expectedRowCount = _fixture.RunningOnEmulator
```
jskeet, line 119: Presumably the  -1 / +1 is due to two tests not running on the emulator? I think it would be clearer to explain that via another property or something similar. (I haven't mentally mapped out exactly what I'd do, but I'm happy to do it in another PR; no need to change this one.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionScopeTests.cs
jskeet: I don't think I understand quite enough about what's been removed to comment - will chat with Amanda offline and see if that helps. (But yes, having test isolation is something I'm generally in favour of :)

Rishabh-V (commented):
Thanks Jon and Amanda for the review and inputs. I have addressed the comments in the new commit. I have one query on the suggestion in the CloudSpannerFixtureBase. Please help me with your thoughts. Rest of the changes are pushed. PTAL. Thanks.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/CloudSpannerFixtureBase.cs
Rishabh-V: Thanks Jon. Done.

amanda-tarafa (approved):
LGTM, thanks! This looks much better now.

Please squash all commits in a single one before merging.

Rishabh-V:
Thanks Amanda. Squashed into 1 commit. Merging.

PR #9205, opened 2022-10-12
chore: Update OwlBot Dockerfile

jskeet:
Although these are comment-only changes, the intention is that it will provoke a Cloud Build to update the Docker container to use the latest .NET SDKs. (We want 3.1.424 and 6.0.402.) This has been tested with a local build, but we want the automated update procedure to kick in.


amanda-tarafa (approved):
LGTM

PR #9256, opened 2022-10-31
tools: Fix API index parsing (ignore unknown fields)

jskeet:


jskeet:
(Will self-merge, in order to kick off another OwlBot run.)

PR #9277, opened 2022-11-03
chore: Update to generator v1.4.9

jskeet:
The new generator uses .NET 6.0 and doesn't generate unit tests.

The changes in this commit are mostly to remove post-generation fix-ups for unit tests, now that we don't have the unit tests.


jskeet:
This mustn't be merged until we release the new generator version and update to it, but then it's required in order to use that correctly.

jskeet:
Oooh... we'll need to remove all the test projects from solutions as well. Will work on that...

jskeet:
(And will need a version of GAX that supports WithHttpRuleOverrides... so we'll need to do a GAX release too.)

generateapis.sh
```diff
@@ -54,6 +54,8 @@ generate_microgenerator() {
   # Delete previously-generated files
   delete_generated apis/$1/$1
   delete_generated apis/$1/$1.Tests
+  rm -f apis/$1/$1.Tests/$1.Tests.csproj
+  rm -f apis/$1/$1.Tests/coverage.xml
```
amanda-tarafa, line 58: These are a one off, righ?

amanda-tarafa (approved):
LGTM

jskeet:
I think the easiest way to remove the test projects will be to remove *all* solution files under apis, then use generate-projects.sh to regenerate them from scratch :)

jskeet:
(Will rebase to handle the conflict.)

amanda-tarafa (approved):
LGTM

PR #9292, opened 2022-11-08
chore: Fix smoke test suggestion/execution

jskeet:
We no longer have a unit test project to publish - but we no longer need to publish it anyway, as that was only for the sake of getting gRPC native libraries in the right place. That's unnecessary now we're using Grpc.Net.Client.


amanda-tarafa (approved):
LGTM

PR #9313, opened 2022-11-10
feat: Add JSON DB type support in BigQuery

Rishabh-V:
This isn't complete and ready yet. 

I am figuring out a whitespace behavior observed while testing.  When I call the ListRows API on the table, whitespace in JSON string between field and value seems to be preserved while they are lost when using the StandardQuery (Select * from {table}).

I will investigate this, document my findings and request the review tomorrow. Thanks.

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryRow.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryRow.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2016 Google Inc. All Rights Reserved.
+// Copyright 2016 Google Inc. All Rights Reserved.
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -112,6 +112,7 @@ namespace Google.Cloud.BigQuery.V2
                 switch (type)
                 {
                     case BigQueryDbType.String:
+                    case BigQueryDbType.Json:
                         return ConvertArray(array, StringConverter);
                     case BigQueryDbType.Int64:
                         return ConvertArray(array, Int64Converter);
@@ -144,6 +145,7 @@ namespace Google.Cloud.BigQuery.V2
             switch (type)
             {
                 case BigQueryDbType.String:
+                case BigQueryDbType.Json:
                     return StringConverter((string)rawValue);
                 case BigQueryDbType.Int64:
                     return Int64Converter((string)rawValue);
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/BigQueryFixture.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/BigQueryFixture.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2016 Google Inc. All Rights Reserved.
+// Copyright 2016 Google Inc. All Rights Reserved.
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -234,6 +234,7 @@ namespace Google.Cloud.BigQuery.V2.IntegrationTests
                 { "single_numeric", BigQueryDbType.Numeric },
                 { "single_big_numeric", BigQueryDbType.BigNumeric },
                 { "single_geography", BigQueryDbType.Geography },
+                { "single_json", BigQueryDbType.Json },
                 { "single_record", recordSchema },
                 
                 // Repeated fields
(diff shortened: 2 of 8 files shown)


amanda-tarafa:
> When I call the ListRows API on the table, whitespace in JSON string between field and value seems to be preserved while they are lost when using the StandardQuery (Select * from {table}).

First thing I'll try to check is whether this is an API behaviour, it does looks like it.

Rishabh-V:
> > When I call the ListRows API on the table, whitespace in JSON string between field and value seems to be preserved while they are lost when using the StandardQuery (Select * from {table}).
> 
> First thing I'll try to check is whether this is an API behaviour, it does looks like it.

Right Amanda, as per [docs](https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#json_type:~:text=Whitespace%20characters%20are%20not%20preserved.) whitespace characters are not preserved. So, When I save "{\"x\": 10, \"y\": \"text\"}" , it is saved as "{\"x\":10,\"y\":\"text\"}" and same is returned from StandardQuery. However, while using ListRows API, I am getting  "{\"x\": 10, \"y\": \"text\"}" (with whitespace characters). I may be missing options or something else so I will explore further. But I think, I can seek feedback on this PR while I understand this behavior.

amanda-tarafa:
I'd like to insist this does look like an API behaviour, whith a quick look at the PR, I don't think the library is going into the fields and parsing the JSON in one way or another, we are just simply converting these to `string`. The easiest way to confirm that is by trying the `tabledata.list` operation on the [API explorer](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/list). If you observe there the same behaviour as with the library, then there's nothing for us to do, except maybe raise it with the API team.

I'll try to review later today, or first thing Monday. But it would be great if you could check with the API explorer and leave a comment with what you see. If the behaviour is caused by the library, then I'd prefer to review once that has been sorted out.

Rishabh-V:
> I'd like to insist this does look like an API behaviour, whith a quick look at the PR, I don't think the library is going into the fields and parsing the JSON in one way or another, we are just simply converting these to `string`. The easiest way to confirm that is by trying the `tabledata.list` operation on the [API explorer](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/list). If you observe there the same behaviour as with the library, then there's nothing for us to do, except maybe raise it with the API team.
> 
> I'll try to review later today, or first thing Monday. But it would be great if you could check with the API explorer and leave a comment with what you see. If the behaviour is caused by the library, then I'd prefer to review once that has been sorted out.

Sure Amanda. Thanks. I will check the behavior with the API explorer and update this thread.

Rishabh-V:
Thanks Amanda. This is indeed the API behavior.

ListRows API returns the whitespaces (as is) . I tested with no whitespace, single whitespace between key and value and multiple whitespace between key and value and the returned JSON string was exactly the same as the input. This is what I had expected (just that doc says whitespace characters are not preserved)

When standard query i.e. (Select * from table) is [executed](https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query), whitespace characters are lost in all the three cases above.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryRow.cs
```diff
@@ -169,6 +171,8 @@ private static object ConvertSingleValue(object rawValue, TableFieldSchema field
                     return GeographyConverter((string) rawValue);
                 case BigQueryDbType.Struct:
                     return ConvertRecord((JObject)rawValue, field);
+                case BigQueryDbType.Json:
```
jskeet, line 174: Include this in the BigQueryDbType.String case?

(In a separate commit, we could convert this to a switch expression, too. Don't do it in this PR though. I suspect a search for "switch" in the BQ libraries will give us lots of them... I'm happy to do that another time.)

jskeet (commented):
If the whitespace difference is an API thing, I think we should neither document nor test for it - I wouldn't be surprised if it changed over time.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryRow.cs
Rishabh-V: Thanks. Fixed.

Sure Jon. I can raise a new PR with with conversion to switch expression early next week.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryRow.cs
```diff
@@ -137,13 +137,16 @@ private static object ConvertSingleValue(object rawValue, TableFieldSchema field
                         return ConvertArray(array, BigNumericConverter);
                     case BigQueryDbType.Geography:
                         return ConvertArray(array, GeographyConverter);
+                    case BigQueryDbType.Json:
```
jskeet, line 140: Sorry, I missed this before - another one that can just move to the BigQueryDbType.String block.

jskeet (approved):
LGTM once the final switch change has been made. (Or even without, if we want to do that later.)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryRow.cs
Rishabh-V: Thanks for catching this as well. I apparently missed it twice even though I had observed and fixed it in BigQueryParameter.cs in the same PR.

jskeet (approved):
Please turn into a single commit before merging, but this looks fine.

amanda-tarafa (approved):

PR #9318, opened 2022-11-14
Customized options POC draft

anshuldavid13:


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Google.Cloud.Storage.V1;
+
+/// <summary>
+/// 
+/// </summary>
+public sealed class RetryOptions
```
jskeet, line 24: I would make this immutable if possible. It'll make a lot of things simpler.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryOptions.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public IList<int> ErrorCodes { get; set; }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    public bool RetryAlways { get; set; }
```
jskeet, line 54: Rather than having `RetryAlways` and `ErrorCodes`, if you have a `Func<int, bool>` (or potentially a `Func<HttpResponseMessage, bool>` so that the predicate can check any aspect of the response) then you gain more flexibility.

You can then add factory methods/properties for `RetryAlways` (which would be `_ => true`) or a method to create a predicate from a set of error codes.

anshuldavid13:
POC Completed

PR #9209, opened 2022-10-13
Retry conformance refactoring - second pass

jskeet:
This builds on the previous refactoring to remove the reflection entirely. We don't need to build up a map and invoke reflectively, building argument lists automatically - we can just invoke storage client methods directly, using the values from the context.

PR #9198, opened 2022-10-10
chore: Completed all mappings for remaining methods, added configs and fixed issues with the test cases

anshuldavid13:
Delete enter the commit message for your changes. Lines starting

PR #9148, opened 2022-09-22
Changes to run always idempotent retry tests.

anuragsrivstv:

PR #9328, opened 2022-11-15
feat: Retry Conformance Testing module

anshuldavid13:
This work is a combination of the following PRs:
#8945 
#9089 
#9148 
#9189 
#9198 
#9209 

Design details are present in go/gcs-retry-strategy-conformance-testing-design

Implementation details are added in go/gcsRetryConformanceImplementationDotnet


jskeet (commented):
I think it would be good to separate this into at least two commits - the first commit should *just* update the conformance test submodule and regenerate the protos. That can be basically reviewed as "Yes, that's generated code."

Then the second commit can be the actual retry conformance tests.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -0,0 +1,450 @@
...
+                {
+                    RunRetryTest(response, context, method.Group, specifyPreconditions);
+
+                    // storage.buckets.setIamPolicy has no preconditions implemented in .NET and hence will retry successfully
+                    if (!method.Name.Contains("buckets.setIamPolicy"))
+                    {
+                        Assert.False(response.Completed);
+                    }
+                }
+                catch (Exception ex) // To catch expected exception when retry should not happen.
```
jskeet, line 108: After this PR, it would be good to refactor this catch clause - it could be significantly simpler. But I'm happy enough for it to go in as-is for now.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/StorageResources.cs
```diff
@@ -0,0 +1,33 @@
...
+using Google.Cloud.Storage.V1.Tests.Conformance;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Google.Cloud.Storage.V1.RetryConformanceTests;
+
+/// <summary>
+/// Created to allow multiple resources to be created and sent together for single test cases as well
+/// </summary>
+internal class StorageResources : List<StorageResource>
```
jskeet, line 24: I don't think we're using this any more, are we? Can we delete the file?

jskeet (commented):
There's one file to delete, and please rename both of the commits:

- They should both be conventional commits, probably with "chore:" for the first and "tests:" for the second
- The first one has README.md at the start for no obvious reason. (I'd specify that it's about updating the conformance test submodule more than regenerating - it's the update that prompts the regeneration)

After that we should be okay to merge, although I've spotted a few nits that I'll fix up afterwards.

jskeet (approved):
I'm now not sure whether StorageResource is used any more either, but I can clear that up in a follow-up PR. Will merge now.

PR #9348, opened 2022-11-21
feat: Add support for count(*) aggregation query firestore

hemanshv:
Adding count(*) functionality in firestore

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/QueryTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/QueryTest.cs
@@ -431,6 +431,24 @@ namespace Google.Cloud.Firestore.IntegrationTests
             Assert.Equal(new[] { "a", "b", "d", "e" }, ids);
         }
 
+        [Fact]
+        public async Task Count_WithoutLimit()
+        {
+            CollectionReference collection = _fixture.HighScoreCollection;
+            var snapshot = await collection.Count().GetSnapshotAsync();
+            Assert.Equal(HighScore.Data.Length, snapshot.Count);
+        }
+
+        [Fact]
+        public async Task Count_WithLimit()
+        {
+            CollectionReference collection = _fixture.HighScoreCollection;
+            var snapshotWithoutLimit = await collection.Count().GetSnapshotAsync();
+            var snapshotWithLimit = await collection.Limit(2).Count().GetSnapshotAsync();
+            Assert.Equal(HighScore.Data.Length, snapshotWithoutLimit.Count);
+            Assert.Equal(2, snapshotWithLimit.Count);
+        }
+
         public static TheoryData<string, object, string[]> ArrayContainsTheoryData = new TheoryData<string, object, string[]>
         {
             { "StringArray", "x", new[] { "string-x,y", "mixed" } },
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQuerySnapshotTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQuerySnapshotTest.cs
@@ -1,0 +1,61 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using Google.Cloud.ClientTesting;
+using Xunit;
+using static Google.Cloud.Firestore.Tests.ProtoHelpers;
+
+namespace Google.Cloud.Firestore.Tests;
+public class AggregateQuerySnapshotTest
+{
+    private static readonly FirestoreDb s_db = FirestoreDb.Create("proj", "db", new FakeFirestoreClient());
+
+    [Fact]
+    public void VerifyAllMembers()
+    {
+        AggregateQuery aggQuery = s_db.Collection("col").Count();
+        var sampleReadTime = Timestamp.FromProto(CreateProtoTimestamp(1, 2));
+        int sampleCount = 10;
+        var aggQuerySnapshot = new AggregateQuerySnapshot(aggQuery, sampleReadTime, sampleCount);
+        Assert.Equal(sampleCount, aggQuerySnapshot.Count);
+        Assert.Equal(sampleReadTime, aggQuerySnapshot.ReadTime);
+        Assert.Equal(aggQuery, aggQuerySnapshot.Query);
+    }
+
+    [Fact]
+    public void Equality()
+    {
+        AggregateQuery aggQuery1 = s_db.Collection("col").Count();
+        AggregateQuery aggQuery2 = s_db.Collection("col1").Count();
+        var sampleReadTime = Timestamp.FromProto(CreateProtoTimestamp(1, 2));
+        int sampleCount = 10;
+        var control = new AggregateQuerySnapshot(aggQuery1, sampleReadTime, sampleCount);
+
+        EqualityTester.AssertEqual(control,
+            equal: new[] {
+                // All the members are equal.
+                new AggregateQuerySnapshot(s_db.Collection("col").Count(), Timestamp.FromProto(CreateProtoTimestamp(1, 2)), 10)
+            },
+            unequal: new[] {
+                // Unequal aggregate query
+                new AggregateQuerySnapshot(aggQuery2, sampleReadTime, sampleCount),
+                // Null count.
+                new AggregateQuerySnapshot(aggQuery2, sampleReadTime, null),
+                // Same aggregate query and count but different read time. 
+                new AggregateQuerySnapshot(s_db.Collection("col").Count(), Timestamp.FromProto(CreateProtoTimestamp(3, 4)), sampleCount),
+                // Same aggregate query but different read time and count.
+                new AggregateQuerySnapshot(s_db.Collection("col").Count(), Timestamp.FromProto(CreateProtoTimestamp(3, 4)), 20)
+            });
+    }
+}
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQueryTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQueryTest.cs
@@ -1,0 +1,88 @@
+﻿// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using Google.Api.Gax.Grpc;
+using Google.Cloud.ClientTesting;
+using Google.Cloud.Firestore.V1;
+using Moq;
+using System.Threading.Tasks;
+using Xunit;
+using static Google.Cloud.Firestore.Tests.ProtoHelpers;
+using static Google.Cloud.Firestore.V1.StructuredAggregationQuery.Types;
+using static Google.Cloud.Firestore.V1.StructuredAggregationQuery.Types.Aggregation.Types;
+
+namespace Google.Cloud.Firestore.Tests;
+public class AggregateQueryTest
+{
+    private static readonly FirestoreDb s_db = FirestoreDb.Create("proj", "db", new FakeFirestoreClient());
+    private static readonly Query s_query = s_db.Collection("col");
+
+    [Fact]
+    public void TestToStructuredAggregationQuery()
+    {
+        var expectedStructuredAggregationQuery = new StructuredAggregationQuery
+        {
+            StructuredQuery = s_query.ToStructuredQuery(),
+            Aggregations = { new Aggregation { Alias = "Count", Count = new Count() } }
+        };
+        Assert.Equal(expectedStructuredAggregationQuery, s_query.Count().ToStructuredAggregationQuery());
+    }
+
+    [Fact]
+    public async Task GetSnapshotAsync_VerifySnapshotMembers()
+    {
+        Mock<FirestoreClient> mock = new() { CallBase = true };
+        var db = FirestoreDb.Create("proj", "db", mock.Object);
+        var query = db.Collection("col").Select("Name");
+        var sampleReadTime = CreateProtoTimestamp(1, 3);
+        var request = new RunAggregationQueryRequest
+        {
+            Parent = query.ParentPath,
+            StructuredAggregationQuery = new StructuredAggregationQuery
+            {
+                StructuredQuery = query.ToStructuredQuery(),
+                Aggregations = { new Aggregation { Alias = "Count", Count = new Count() } }
+            }
+        };
+        var response = new FakeAggregationQueryStream(new[]
+        {
+            new RunAggregationQueryResponse { ReadTime = sampleReadTime, Result = new AggregationResult() }
+        });
+        mock.Setup(c => c.RunAggregationQuery(request, It.IsAny<CallSettings>())).Returns(response);
+        var aggregateQuery = query.Count();
+        var snapshot = await aggregateQuery.GetSnapshotAsync();
+        Assert.Equal(aggregateQuery, snapshot.Query);
+        Assert.Equal(Timestamp.FromProto(sampleReadTime), snapshot.ReadTime);
+    }
+
+    [Fact]
+    public void Equality()
+    {
+        var control = s_query.Count();
+        EqualityTester.AssertEqual(control,
+            equal: new[]
+            {
+                new AggregateQuery(s_db.Collection("col")).WithAggregation(Aggregates.CreateCountAggregate()),
+                // Aggregate query returned by count.
+                s_db.Collection("col").Count()
+            },
+            unequal: new[]
+            {
+                // same query but without any aggregation.
+                new AggregateQuery(s_db.Collection("col")),
+                // un-equal query.
+                new AggregateQuery(s_db.Collection("col1"))
+            });
+    }
+}
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/TransactionTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/TransactionTest.cs
@@ -87,6 +87,28 @@ namespace Google.Cloud.Firestore.Tests
         }
 
         [Fact]
+        public async Task GetSnaphsotAsync_AggregationQuery()
+        {
+            var db = CreateFirestoreDbExpectingNoCommits();
+            var transaction = await Transaction.BeginAsync(db, null, default);
+            var query = db.Collection("col");
+            var aggQuery = query.Count();
+            var snapshot = await transaction.GetSnapshotAsync(aggQuery);
+            Assert.Equal(aggQuery, snapshot.Query);
+            Assert.NotNull(snapshot);
+        }
+
+        [Fact]
+        public async Task GetSnaphotAsync_FailAfterWriteForAggregationQuery()
+        {
+            var db = CreateFirestoreDbExpectingNoCommits();
+            var transaction = await Transaction.BeginAsync(db, null, default);
+            var doc = db.Document("col/doc");
+            transaction.Delete(doc);
+            await Assert.ThrowsAsync<InvalidOperationException>(() => transaction.GetSnapshotAsync(doc.Parent));
+        }
+
+        [Fact]
         public async Task CommitAsync()
         {
             var client = new TransactionTestingClient();
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
@@ -1,0 +1,119 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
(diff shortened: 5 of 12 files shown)


hemanshv:
Hi @jskeet @amanda-tarafa ,
Its a draft PR for query count feature. Its still under process. 
I have raised the PR to get some internal feedback on the same.
Please do not review. Will let you know once the changes are finalized.
Thanks,

Rishabh-V:
I will have a look later today after going through the design document. Thanks.

Rishabh-V (commented):
First pass done. I have only reviewed the production code for now based on my limited understanding after reading the design doc. Will sync up with you early next week when you are back and we can do a second pass for the complete codebase after clarifying my understanding. Thanks.

Rishabh-V (commented):
I have reviewed based on my understanding. Once you address/respond to these comments, please feel free to remove the LEX-review label and request Jon and Amanda for their review. Thank you.

hemanshv:
Hi @jskeet  @amanda-tarafa ,
This is now ready for review. 
Please review and provide you feedback.

Thanks,

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,120 @@
...
+    {
+        if (ReferenceEquals(other, this))
+        {
+            return true;
+        }
+        if (other is null)
+        {
+            return false;
+        }
+        if (GetType() != other.GetType())
```
jskeet, line 114: No need for this - the type is sealed.

Again though, I'd encourage you to include the aggregations as part of the state, both here and for GetHashCode.
If we represent "we've asked for count" as state, even though it's the only available aggregation available now, it'll make it much easier to add support for other aggregations later.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuerySnapshot.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    public AggregateQuery Query { get; }
+
+    /// <summary>
+    /// The time at which the snapshot was read.
+    /// </summary>
+    public Timestamp ReadTime { get; }
+
+    /// <summary>
+    /// Number of documents that matches the query.
```
jskeet, line 36: Possibly add a note about limits?
What will this return in the future when we support non-count aggregations, and a count hasn't been specified?
(Perhaps the return type should actually be `long?` instead?) I'm genuinely not sure.

jskeet (commented):
Main comments are around the way that AggregateQuery is tested, and the current lack of state indicating the kind of aggregation being used.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuerySnapshot.cs
hemanshv: Thanks for your feedback @jskeet .Will update it and let you know.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/QueryTest.cs
```diff
@@ -431,6 +431,24 @@ public async Task WhereArrayContainsAny()
...
+            var snapshot = await collection.Count().GetSnapshotAsync();
+            Assert.Equal(snapshot.Count, HighScore.Data.Length);
+        }
+
+        [Fact]
+        public async Task Count_WithLimit()
+        {
+            CollectionReference collection = _fixture.HighScoreCollection;
+            var snapshotWithoutLimit = await collection.WhereGreaterThanOrEqualTo("Score", 100).Count().GetSnapshotAsync();
+            var snapshotWithLimit = await collection.WhereGreaterThanOrEqualTo("Score", 100).Limit(2).Count().GetSnapshotAsync();
```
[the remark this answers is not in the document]
hemanshv: No there is nothing in the result which indicate the same, but I think its trivial. If the limit provided is equal to the count then we can say that there is "at least n number of records".

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/QueryTest.cs
```diff
@@ -431,6 +431,24 @@ public async Task WhereArrayContainsAny()
...
+            CollectionReference collection = _fixture.HighScoreCollection;
+            var snapshot = await collection.Count().GetSnapshotAsync();
+            Assert.Equal(snapshot.Count, HighScore.Data.Length);
+        }
+
+        [Fact]
+        public async Task Count_WithLimit()
+        {
+            CollectionReference collection = _fixture.HighScoreCollection;
+            var snapshotWithoutLimit = await collection.WhereGreaterThanOrEqualTo("Score", 100).Count().GetSnapshotAsync();
```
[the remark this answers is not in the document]
hemanshv: No not really.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/QueryTest.cs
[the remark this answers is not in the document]
jskeet: The problem is that the code that supplies the limit may not be the code that handles the response. I don't think we need any change in this PR, but it *might* be worth asking the Firestore team to consider indicating somewhere in the response that the count is not necessarily accurate.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
hemanshv: Make Sense Jon will do.
Thanks

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuerySnapshot.cs
hemanshv: Yes Jon I will add a note about the limit thing. On return thing: When we support non-count aggregations and count hasn't been specified in that case the value of count would be "null". I have updated the return type to be long?.
Thanks

jskeet:
Will look when I get the chance - even coming back from one day off, I've got a bit of a backlog...

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,129 @@
...
+
+    private AggregateQuery(Query query, IReadOnlyList<Aggregation> aggregations)
+    {
+        _query = GaxPreconditions.CheckNotNull(query, nameof(query));
+        _aggregations = aggregations;
+    }
+
+    internal AggregateQuery WithAggregation(Aggregation aggregation)
+    {
+        if (aggregation.Count != null)
```
jskeet, line 54: Rather than splitting the initialization of the Count aggregation, why not create an internal `CreateCountAggregation` method in here, which can create it fully?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,129 @@
...
+        {
+            aggregation.Alias = CountAlias;
+        }
+        var newAggregations = _aggregations == null ? new List<Aggregation>() : new List<Aggregation>(_aggregations);
```
jskeet, line 58: Rather than allowing _aggregations to *ever* be null, I'd initialize it with an empty list (which you only need one instance of) in the single-parameter constructor. That should simplify all code using it.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,129 @@
...
+        IAsyncEnumerable<RunAggregationQueryResponse> responseStream = GetAggregationQueryResponseStreamAsync(transactionId, cancellationToken);
+        Timestamp? readTime = null;
+        Value count = null;
+        await responseStream.ForEachAsync(response =>
+        {
+            if (response.Result.AggregateFields != null)
+            {
+                response.Result.AggregateFields.TryGetValue(CountAlias, out count);
+            }
+            if (readTime == null && response.ReadTime != null)
```
jskeet, line 82: You can simplify this to

```csharp
readTime ??= Timestamp.FromProtoOrNull(response.ReadTime);
```

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,129 @@
...
+        GetSnapshotAsync(null, cancellationToken);
+
+    internal async Task<AggregateQuerySnapshot> GetSnapshotAsync(ByteString transactionId, CancellationToken cancellationToken)
+    {
+        IAsyncEnumerable<RunAggregationQueryResponse> responseStream = GetAggregationQueryResponseStreamAsync(transactionId, cancellationToken);
+        Timestamp? readTime = null;
+        Value count = null;
+        await responseStream.ForEachAsync(response =>
+        {
+            if (response.Result.AggregateFields != null)
```
jskeet, line 78: I'm not sure this is what you want - that means if there are multiple responses with `AggregateFields` populated, if the count is in the first one, we'll end up losing the count. You could do something like:

```csharp
if (count is null)
{
    response.Result.AggregateFields?.TryGetValue(CountAlias, out count);
}
```

jskeet: Actually, I'd also keep `count` as a `long?` variable, and have a check here - so something like:

```csharp
if (count is null && response.Result.AggregateFields?.TryGetValue(CountAlias, out var countValue) == true)
{
    // You probably need to change this - I can't remember exactly what it would be, but hopefully it'll be obvious enough in an IDE.
    GaxPreconditions.CheckState(countValue.ValueCase == Value.Types.ValueCase.Integer, "The count was not an integer.");
    count = countValue.IntegerValue;
}
```

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    private IAsyncEnumerable<RunAggregationQueryResponse> GetAggregationQueryResponseStreamAsync(ByteString transactionId, CancellationToken cancellationToken)
+    {
+        RunAggregationQueryRequest request = new RunAggregationQueryRequest
+        {
+            Parent = _query.ParentPath,
+            StructuredAggregationQuery = ToStructuredAggregationQuery()
+        };
+        if (transactionId != null)
```
jskeet, line 100: You could just use `Transaction = transactionId ?? ByteString.Empty` in the object initializer, I suspect.
(And yes, this may be something that could be simplified in Query too... I haven't checked.)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,129 @@
...
+    /// <returns>A <see cref="AggregateQuerySnapshot"/> which contains results of the Aggregate functions.</returns>
+    public Task<AggregateQuerySnapshot> GetSnapshotAsync(CancellationToken cancellationToken = default) =>
+        GetSnapshotAsync(null, cancellationToken);
+
+    internal async Task<AggregateQuerySnapshot> GetSnapshotAsync(ByteString transactionId, CancellationToken cancellationToken)
+    {
+        IAsyncEnumerable<RunAggregationQueryResponse> responseStream = GetAggregationQueryResponseStreamAsync(transactionId, cancellationToken);
+        Timestamp? readTime = null;
+        Value count = null;
+        await responseStream.ForEachAsync(response =>
```
jskeet, line 76: It might be worth extracting this lambda expression into a local method (e.g. `ProcessResponse`) for readability - I probably didn't do it when writing the Firestore client initially as they didn't exist back then ;)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuerySnapshot.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// </summary>
+    public AggregateQuery Query { get; }
+
+    /// <summary>
+    /// The time at which the snapshot was read.
+    /// </summary>
+    public Timestamp ReadTime { get; }
+
+    /// <summary>
+    /// Number of documents that matches the query. In-case when explicit limit is provided in the query.
```
jskeet, line 36: The second and third sentences should be a single sentence, something like:

"Note that when the query contains a limit, the count is restricted by that limit."

jskeet: I'd also document the conditions in which it will be null (anticipating that actually being feasible...)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -417,6 +419,13 @@ public Query Select(params FieldPath[] fieldPaths)
         public Query WhereNotIn(FieldPath fieldPath, IEnumerable values) =>
             Where(fieldPath, FieldOp.NotIn, ValidateWhereInValues(values));
 
+        /// <summary>
+        /// Returns an <see cref="AggregateQuery"/> with count(*) aggregation applied on this query.
+        /// </summary>
+        /// <returns>An instance of <see cref="AggregateQuery"/> with count(*) aggregation applied.</returns>
+        public AggregateQuery Count() =>
+            new AggregateQuery(this).WithAggregation(new Aggregation { Count = new Count() });
```
jskeet, line 427: Right, and this would become `new AggregateQuery(this).WithAggregation(AggregateQuery.CreateCountAggregate))`. Or you might introduce an `Aggregates` static class as a factory for aggregates. (You'd need to expose the count alias in there, of course.)

hemanshv:
Thanks for the feedback Jon. I will address them and let you know.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/TransactionTest.cs
```diff
@@ -86,6 +86,29 @@ public async Task GetSnaphotAsync_FailsAfterWrite()
...
+        }
+
+        [Fact]
+        public async Task GetSnaphotAsync_ShouldNotFailAfterWriteForAggregationQuery()
+        {
+            var db = CreateFirestoreDbExpectingNoCommits();
+            var transaction = await Transaction.BeginAsync(db, null, default);
+            var doc = db.Document("col/doc");
+            transaction.Delete(doc);
+            var snapshot = await transaction.GetSnapshotAsync(doc.Parent.Count());
```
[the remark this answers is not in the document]
hemanshv: I missed the validation of writes in our new method inside transaction. This should fail as per our current transaction semantics(all reads to be executed before writes). I will update that. Thanks for pointing this out.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQueryTest.cs
```diff
@@ -0,0 +1,68 @@
...
+        Assert.Equal(expectedQuery, snapshot.Query);
+        Assert.Equal(Timestamp.FromProto(sampleReadTime), snapshot.ReadTime);
+    }
+
+    [Fact]
+    public void Equality()
+    {
+        var aggQueryWithCount = s_mockQuery.Count();
+        var aggQueryWithoutCount = new AggregateQuery(s_mockQuery);
+        EqualityTester.AssertEqual(aggQueryWithCount, equal: new[] { new AggregateQuery(s_mockQuery).WithAggregation(new Aggregation { Count = new Count() }) }, unequal: new[] { new AggregateQuery(s_db.Collection("col1")), aggQueryWithoutCount });
```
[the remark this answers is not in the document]
hemanshv: Could you please elaborate a bit more on this.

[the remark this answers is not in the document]
jskeet: Which part? The first is just "use some separate local variables for clarity instead of doing everything in a single huge statement".
The second is about creating entirely independent but equal queries: `s_db.Collection("xyz")` and calling `s_db.Collection("xyz")` again, so equal underlying queries rather than a single Query object used in both cases.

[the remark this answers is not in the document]
hemanshv: Got it Thanks

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
hemanshv: While using this we are encountering StatusCode="InvalidArgument", Detail="Invalid transaction." error. The error is occurring because the setter of transactionId also sets consistencySelectorCase_  to ConsistencySelectorOneofCase.Transaction and a transaction with an empty transaction id is an invalid transaction. So I think we should stick to the old way.

jskeet: Ah, right. Yup, that makes sense.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQuerySnapshotTest.cs
```diff
@@ -0,0 +1,48 @@
...
+        var sampleReadTime = Timestamp.FromProto(CreateProtoTimestamp(1, 2));
+        int sampleCount = 10;
+        var aggQuerySnapshot = new AggregateQuerySnapshot(aggQuery, sampleReadTime, sampleCount);
+        Assert.Equal(sampleCount, aggQuerySnapshot.Count);
+        Assert.Equal(sampleReadTime, aggQuerySnapshot.ReadTime);
+        Assert.Equal(aggQuery, aggQuerySnapshot.Query);
+    }
+
+    [Fact]
+    public void Equality()
```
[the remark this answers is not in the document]
jskeet: Much better now, thanks!

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQuerySnapshotTest.cs
```diff
@@ -0,0 +1,61 @@
...
+        int sampleCount = 10;
+        var control = new AggregateQuerySnapshot(aggQuery1, sampleReadTime, sampleCount);
+
+        EqualityTester.AssertEqual(control,
+            equal: new[] {
+                // all the members are equal.
+                new AggregateQuerySnapshot(s_db.Collection("col").Count(), Timestamp.FromProto(CreateProtoTimestamp(1, 2)), 10)
+            },
+            unequal: new[] {
+                // unequal aggregate query
```
jskeet, line 51: Super-nit: Please start comments with a capital letter, even if they're just sentence fragments.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQuerySnapshotTest.cs
```diff
@@ -0,0 +1,61 @@
...
+                new AggregateQuerySnapshot(aggQuery2, sampleReadTime, sampleCount),
+                // null count.
+                new AggregateQuerySnapshot(aggQuery2, sampleReadTime, null),
+                // same aggregate query and count but different readtime.
```
jskeet, line 55: Super nit: "read time" instead of "readtime" (it's logically two words, which is why the T is capitalized in the PascalCasedName).

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    private IAsyncEnumerable<RunAggregationQueryResponse> GetAggregationQueryResponseStreamAsync(ByteString transactionId, CancellationToken cancellationToken)
+    {
+        RunAggregationQueryRequest request = new RunAggregationQueryRequest
+        {
+            Parent = _query.ParentPath,
+            StructuredAggregationQuery = ToStructuredAggregationQuery()
+        };
+        if(transactionId != null)
```
jskeet, line 90: Nit: space before `(`

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// <returns>A <see cref="AggregateQuerySnapshot"/> which contains results of the Aggregate functions.</returns>
+    public Task<AggregateQuerySnapshot> GetSnapshotAsync(CancellationToken cancellationToken = default) =>
+        GetSnapshotAsync(null, cancellationToken);
+
+    internal async Task<AggregateQuerySnapshot> GetSnapshotAsync(ByteString transactionId, CancellationToken cancellationToken)
+    {
+        IAsyncEnumerable<RunAggregationQueryResponse> responseStream = GetAggregationQueryResponseStreamAsync(transactionId, cancellationToken);
+        Timestamp? readTime = null;
+        long? count = null;
+        void ProcessResponse(RunAggregationQueryResponse response)
```
jskeet, line 68: Style nit we can discuss in our meeting - I personally tend to put local methods at the end of the method (with a line break before each one). I think that allows the code to stay in blocks of the same abstraction level. Let's talk about it though.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -0,0 +1,119 @@
...
+            if (count is null && response.Result.AggregateFields?.TryGetValue(Aggregates.CountAlias, out var countValue) == true)
+            {
+                GaxPreconditions.CheckState(countValue.ValueTypeCase == Value.ValueTypeOneofCase.IntegerValue, "The count was not an integer.");
+                count = countValue.IntegerValue;
+            }
+            readTime ??= Timestamp.FromProtoOrNull(response.ReadTime);
+        }
+        await responseStream.ForEachAsync(response => ProcessResponse(response), cancellationToken).ConfigureAwait(false);
+        GaxPreconditions.CheckState(readTime != null, "The stream returned from RunAggregationQuery did not provide a read timestamp.");
+        var snapshot = new AggregateQuerySnapshot(this, readTime.Value, count);
```
jskeet, line 79: Nit: Maybe just return directly rather than using another local variable here?

jskeet (approved):
Just tiny nits now - I've approved so that you can merge after fixing those, or we could even address them in a separate PR if you want. (Please squash to a single commit before merging though.)

PR #9363, opened 2022-11-25
tests: Check all GAPIC clients can be created successfully

jskeet:
This will prevent failures of HTTP configuration for REGAPIC clients.

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/CreateClientsCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/CreateClientsCommand.cs
@@ -1,0 +1,129 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using Google.Cloud.Tools.Common;
+using System;
+using System.IO;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using System.Threading;
+
+namespace Google.Cloud.Tools.ReleaseManager;
+
+public class CreateClientsCommand : CommandBase
+{
+    private const string PublishTargetFramework = "netstandard2.1";
+
+    public CreateClientsCommand()
+        : base("create-clients", "Checks that we can create clients for a pure-GAPIC package", "id")
+    {
+    }
+
+    protected override void ExecuteImpl(string[] args)
+    {
+        string id = args[0];
+        var catalog = ApiCatalog.Load();
+        if (!catalog.TryGetApi(id, out var api))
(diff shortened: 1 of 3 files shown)


jskeet:
Note: this initially deliberately breaks Compute, so we can see what a failure will look like.

tools/Google.Cloud.Tools.ReleaseManager/CreateClientsCommand.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            throw new UserErrorException($"No such API: {id}");
+        }
+        if (api.Generator != GeneratorType.Micro)
+        {
+            Console.WriteLine($"{id} is not a GAPIC-generated API; skipping client creation test.");
+        }
+
+        var assembly = PublishAndLoadAssembly(id);
+        if (assembly is null)
```
amanda-tarafa, line 48: nit: Maybe a note on when this could happen?

jskeet:
Humbug - failed for the wrong reason. Amending...

jskeet:
Okay, that's more like it. The deliberately-introduced failure mode reported this:

> Configuration error: Package 'Google.Cloud.Compute.V1' is broken: creation of client 'AcceleratorTypesClient' with transport 'Rest' failed: System.ArgumentException: Field no-such-name not found in message google.cloud.compute.v1.AggregatedListAcceleratorTypesRequest

Now reverting the Compute change...

tools/Google.Cloud.Tools.ReleaseManager/CreateClientsCommand.cs
```diff
@@ -0,0 +1,128 @@
...
+            .Where(t => t.IsClass)
+            .Where(t => t.Name.EndsWith("Client"))
+            .Where(t => assembly.GetType(t.FullName + "Builder") is Type builderType
+                        && builderType.BaseType.Name == "ClientBuilderBase`1")
+            .ToList();
+
+        if (clientTypes.Count == 0)
+        {
+            throw new UserErrorException($"API {id} has no client types.");
+        }
```
amanda-tarafa, line 78: Can we just do nothing here? Thinking of the log proto packages, etc., so we don't have to filter them out in bash possibly?

amanda-tarafa (approved):
LGTM

tools/Google.Cloud.Tools.ReleaseManager/CreateClientsCommand.cs
jskeet: They'll be filtered out earlier, by not using the GAPIC generator. (I believe, anyway. I'm running build-everything run now.)

tools/Google.Cloud.Tools.ReleaseManager/CreateClientsCommand.cs
jskeet: Ooh, this was left over from a previous iteration. Will remove.

tools/Google.Cloud.Tools.ReleaseManager/CreateClientsCommand.cs
jskeet: Ha... that was the *plan*, anyway. But:

```text
Google.Apps.Script.Type is not a GAPIC-generated API; skipping client creation test.
Configuration error: API Google.Apps.Script.Type has no client types.
```

jskeet: (The joys of a missing return statement.)

PR #9369, opened 2022-11-29
chore: Deleted unused/ uneccesary code

anshuldavid13:


Rishabh-V:
Hey Anshul, As discussed, the changes are not on top of the latest changes and there are conflicts. Could you please take the latest and then make these code changes? Thanks.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -85,31 +85,46 @@ private async Task RunTestCase(InstructionList instructionList, Method method, b
...
+                        {
+                            throw;
+                        }
+                    }
+                    else
+                    {
+                        throw;
+                    }
                 }
             }
```
Rishabh-V, line 129: There is probably a mismatch of braces in try catch finally block causing the build to fail. 

Also, I see there was a local method `InstructionContainsErrorCode` to handle different cases of exception. Do we really need to remove that?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -421,14 +346,5 @@ private static StringContent GetBodyContent(string methodName, InstructionList i
...
-    /// <summary>
-    /// Function created to be used for getting environmental variables
-    /// </summary>
-    private static string GetEnvironmentVariableOrDefault(string name, string defaultValue)
-    {
-        string value = Environment.GetEnvironmentVariable(name);
-        return string.IsNullOrEmpty(value) ? defaultValue : value;
-    }
-
     private static void SleepAfterBucketCreateDelete() => Thread.Sleep(2000);
```
Rishabh-V, line 349: nit: Should this now be just `SleepAfterBucketCreate()` as we have removed the bucket deletion code?

Rishabh-V (commented):
There were a couple of comments from Amanda on using Async suffix and Skip.IfNot methods. Do we plan to address them in this PR or later?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -243,26 +230,14 @@ private async Task<TestResponse> GetRetryTest(string id)
...
-
     /// <summary>
     /// Create the basic storage resources on storage test bench to enable running of the test cases
     /// </summary>
     private TestContext CreateTestContext(Method method)
     {
         var context = new TestContext(_fixture.ProjectId, _fixture.ServiceAccountEmail);
-        string bucketName = IdGenerator.FromDateTime(prefix: "retry-test-");
-        string objectName = "TestFile.json";
+        string bucketName = IdGenerator.FromDateTime(prefix: retryIdPrefix + "bucket-id-");
```
anuragsrivstv, line 239: retryIdPrefix is already generated using IdGenerator.FromDateTime(..) therefore may be we need not use it again for bucketName as long as we are creating max one bucket per retry test. It could be something like bucketName= $"{retryIdPrefix}-test-bucket"

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -292,9 +267,9 @@ private TestContext CreateTestContext(Method method)
         // For copying objects, need to create destination bucket and object as well
         if (method.Name == "storage.objects.rewrite")
         {
-            string destBucketName = IdGenerator.FromDateTime(prefix: "retry-test-dest-bucket-");
+            string destBucketName = retryIdPrefix + IdGenerator.FromDateTime(prefix: "dest-bucket-id");
```
anuragsrivstv, line 270: similarly like above comment it can be $"{retryIdPrefix}-dest-bucket"

jskeet:
I've created a PR label of "lex-review" and applied it here, so we can do this consistently.

Rishabh-V:
> I've created a PR label of "lex-review" and applied it here, so we can do this consistently.

Thanks Jon.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: The same bucket name is throwing error " the resource already exists" . So will have to keep different ids each time

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -38,6 +38,8 @@ public class RetryConformanceTest
     public static TheoryData<RetryTest> RetryTestData { get; } = StorageConformanceTestData.TestData.GetTheoryData(f => f.RetryTests);
 
     private readonly RetryConformanceTestFixture _fixture;
+    private readonly string retryIdPrefix = IdGenerator.FromDateTime(prefix: "test-id-", suffix: "-");
```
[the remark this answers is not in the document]
anshuldavid13: This is done in response to the comment from amanda "Maybe even create one random id per test and use that with suffixes for the resources of that test, and that may help with diagnosing issues in the future." [from email: Question about resource cleanup on Storage retry tests.] 
Hence I have added a new prefix id to all the resources pertaining to one test run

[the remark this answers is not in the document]
Rishabh-V: Ok. Updating docs may still be worth doing in a later PR so that others know the steps to run the tests.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -57,12 +59,10 @@ public async Task RetryTest(RetryTest test)
         {
             foreach (Method method in test.Methods)
             {
-                if (!ShouldRunMethod(method.Name))
+                if (ShouldRunMethod(method.Name))
```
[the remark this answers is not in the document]
Rishabh-V: `Skip.If` should just skip based on the condition. Here is a [sample code for reference](https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs#L67) . Ack on the code part.

But again, I'll let Jon and Amanda decide on this. Thanks.

Rishabh-V (commented):
Once others in the team have completed the review and you have addressed/responded to the current comments, please feel free to remove the lex-review label and request review from Jon and Amanda. Thanks.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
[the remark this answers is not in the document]
amanda-tarafa: The suggestions was not to use Skip.if, the suggestion was to do it as it currently is. We cannot skip in here because the loop wouln't continue, in this case we don't want to skip the whole test run, just the "subtests" that we know won't work and we cannot do that with skip.

LGTMing this change as it is now.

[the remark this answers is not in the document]
Rishabh-V: Ok. Thanks Amanda.

anuragsrivstv (commented):
LGTM

anshuldavid13:
Thank you LEX team for you review :)

anshuldavid13:
@jskeet @amanda-tarafa Can you please review this PR

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -81,49 +81,34 @@ public async Task RetryTest(RetryTest test)
...
 
-                    // storage.buckets.setIamPolicy has no preconditions implemented in .NET and hence will retry successfully
-                    if (!method.Name.Contains("buckets.setIamPolicy"))
-                    {
-                        Assert.False(response.Completed);
-                    }
-                }
-                catch (GoogleApiException ex) when (InstructionContainsErrorCode(ex.HttpStatusCode))
+                // storage.buckets.setIamPolicy has no preconditions implemented in .NET and hence will retry successfully
+                if (!method.Name.Contains("buckets.setIamPolicy"))
```
amanda-tarafa, line 102: Can you move this case to inside the if? See my comment there.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -81,49 +81,34 @@ public async Task RetryTest(RetryTest test)
...
     /// </summary>
-    private async Task RunTestCase(InstructionList instructionList, Method method, bool expectSuccess, bool specifyPreconditions)
+    private async Task RunTestCaseAsync(InstructionList instructionList, Method method, bool expectSuccess, bool specifyPreconditions)
     {
         var context = CreateTestContext(method);
-        var response = await CreateRetryTestResource(instructionList, method);
-        try
+        var response = await CreateRetryTestResourceAsync(instructionList, method);
+
+        if (expectSuccess)
```
amanda-tarafa, line 89:
```suggestion
        if (expectSuccess || method.Name.Contains("buckets.setIamPolicy"))
```

And include the commented note explaining why this can be here.
And maybe the link to the issue where this will be addressed.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -81,49 +81,34 @@ public async Task RetryTest(RetryTest test)
...
-                DeleteStorageResources(context);
-                RemoveRetryIdHeader();
-                await DeleteRetryTest(response.Id);
-            }
-            catch
-            {
-                // Swallow any exceptions during clean-up: it's best effort.
+                // The instructions specified that the given status code would be returned.
+                // We just need to check that we weren't expecting this specific call to succeed.
+                Assert.False(expectSuccess);
```
amanda-tarafa, line 111: This code is redundant here, expectSuccess will never be true at this point, we are in  the else of `if(expectSuccess)`.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -81,49 +81,34 @@ public async Task RetryTest(RetryTest test)
...
             }
         }
```
amanda-tarafa, line 114: I think the implementation of `InstructionContainsErrorCode` can be simplified:
```
=> instructionList.Instructions.Contains($"return-{(int)statusCode}");
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -165,7 +150,7 @@ private void RunRetryTest(TestResponse response, TestContext context, string gro
```
amanda-tarafa, line 150: Can you leave a note on line 145 explaining why you are skipping the ones that contain `notification`?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -165,7 +150,7 @@ private void RunRetryTest(TestResponse response, TestContext context, string gro
 
     // Note: not a local function as that cannot handle dynamic binding with generics.
     private static void ConsumeListOutput<TRequest, TResource>(PagedEnumerable<TRequest, TResource> pagedEnumerable) =>
-        pagedEnumerable.ReadPage(100);
+        pagedEnumerable.ReadPage(1000);
```
amanda-tarafa, line 153: Any reason why we went from 100 to 1000?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -235,34 +220,22 @@ Policy GetIamPolicyAndAddPublicViewer()
...
-
     /// <summary>
     /// Create the basic storage resources on storage test bench to enable running of the test cases
     /// </summary>
     private TestContext CreateTestContext(Method method)
     {
         var context = new TestContext(_fixture.ProjectId, _fixture.ServiceAccountEmail);
-        string bucketName = IdGenerator.FromDateTime(prefix: "retry-test-");
-        string objectName = "TestFile.json";
+        string bucketName = IdGenerator.FromDateTime(prefix: retryIdPrefix + "bucket-");
```
amanda-tarafa, line 237: I think retryIdPrefix is already created FromDataTime, so maybe this one can be from GUID. Else the ID's will have information of two different timestamps on them, and that'd be confusing.

But also, unlles you need more than one bucket per tests, you can simply do:

```suggestion
        string bucketName = $"{retryIdPrefix}bucket";
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -81,49 +81,34 @@ public async Task RetryTest(RetryTest test)
...
-            if (expectSuccess)
+            RunRetryTest(response, context, method.Group, specifyPreconditions);
+            var postTestResponse = await GetRetryTestAsync(response.Id);
+            Assert.True(postTestResponse.Completed, "Expected retry test completed to be true, but was false.");
+        }
+        else
+        {
+            try
             {
                 RunRetryTest(response, context, method.Group, specifyPreconditions);
```
amanda-tarafa, line 99: After this RunRetryTest call you need to add something like:
```suggestion
                RunRetryTest(response, context, method.Group, specifyPreconditions);
                Assert.True(false, "Failure was expected but retry succeeded");
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -292,17 +265,17 @@ private TestContext CreateTestContext(Method method)
         // For copying objects, need to create destination bucket and object as well
         if (method.Name == "storage.objects.rewrite")
```
amanda-tarafa, line 266: This is a comment about the note in line 257-258 and the code following from 260-263.

If the test is about retrying bucket creation, I don't think you should create the bucket on test setup. Creting the bucket is what you actually want to test.

Or am I missing something?

amanda-tarafa: And actually, from looking at the code on line 163, I don't think you need to do this at all.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -292,17 +265,17 @@ private TestContext CreateTestContext(Method method)
...
         {
-            string destBucketName = IdGenerator.FromDateTime(prefix: "retry-test-dest-bucket-");
+            string destBucketName = IdGenerator.FromDateTime(prefix: retryIdPrefix + "dest-bucket-");
```
amanda-tarafa, line 268: Same comment as before regarding using FromDateTime

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTestFixture.cs
```diff
@@ -56,6 +56,4 @@ private static string GetEnvironmentVariableOrDefault(string name, string defaul
         string value = Environment.GetEnvironmentVariable(name);
         return string.IsNullOrEmpty(value) ? defaultValue : value;
     }
-
-    internal static void SleepAfterBucketCreateDelete() => Thread.Sleep(2000);
```
amanda-tarafa, line 60: It probably should be in the fxiture, in case we ever use it in a diffirent test class.

amanda-tarafa (changes requested):
Thanks, a few more comments, but looking better.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: As discussed, this optimization is causing some test cases to fail, will try this approach further and will push the changes in another PR as required

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: Will implement this as well once the previous change is incorporated as they are linked

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: Same comment as above

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: Since we are not disposing off resources explicitly now, we need some more buffer  for the test cases to run together. Hence the increased limit.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: Done the creation from GUID. For each buckets though we still need differentiators as the same test suite needs multiple buckets to be created.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: As discussed, these changes for "insert.bucket" are also throwing errors in the test cases execution. Will be checking more on this and picking it up in a separate PR as required.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anshuldavid13: Replied to in the previous comment.

anshuldavid13 (commented):
Review comments implemented

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anuragsrivstv: Or may be we can directly write  Assert.Fail("Failure was expected but retry succeeded"); ? @amanda-tarafa what do you think ?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
anuragsrivstv: So basically , at line 163 the mandatory parameter for our CreateBucket API is a new bucketName not a new bucket itself.
As the _new_  bucket is not a  pre-requites for create/insert bucket operation, retry_tests.json dosent specify it. 

@anshuldavid13 , I _think_, that as we need to know in retry test that we need a bucket name before calling CreateBucket API, we should do handling to only create new bucket name not a new bucket.

anshuldavid13: Actually for insert bucket, the retry_tests.json does require a bucket as a pre requisite resource in test id 5.

anuragsrivstv: To me  bucket as a pre requisite to insert bucket in test id 5 seems like a mistake , also test id 5 is for non_retryable_errors 400 and 401 so even if the resource is wrong it might not be getting caught( may be we should verify this). Also again in test id 6 for mix_retryable_non_retryable_errors there is no resource for buckets.insert . 

Also one more thing, in the current code passing _IdGenerator.FromDateTime(prefix: retryIdPrefix + "insert-bucket-")_ as bucket name( at line 162 in create bucket) argument nullify the purpose of line number 259-261 as we are not using existing bucket name anywhere while creating bucket( also it would have given duplicate bucket error is we have used it)

anshuldavid13: Yes, hence will verify the "insert_bucket" scenarios separately and fix as required in the next PR.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -142,6 +140,7 @@ private void RunRetryTest(TestResponse response, TestContext context, string gro
         var result = ExecuteRpc(rpc, context, specifyPreconditions);
 
         // For list methods, we need to fetch a page in order to actually perform requests.
+        // Dont need to do this for notifications as those are getting consumed immidiately
         if (rpc.Contains("list") && !rpc.Contains("notification"))
         {
             ConsumeListOutput((dynamic) result);
```
amanda-tarafa, line 146: Even with your comment replying to my comment, I still don't understand why we went from 100 to 1000. Why is the fact that we are not deleting resources significant here?

amanda-tarafa (changes requested):

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
amanda-tarafa: Exactly what @anuragsrivstv is saying:

- You should need a bucket name, but not a created bucket for insert bucket tests. And the bucket that you are creating is not being used because you are passing a fresh bucket ID: if you remove the code in 259-261, everything will continue to work.
- You need to check with the Storage team whether test 5 has a mistake by requiring a bucket, or whether we have missunderstood what the test should do.

amanda-tarafa (approved):
We've discussed online, some of the remaining issues will be dealt with in upcoming PRs.
Approving so this can be merged, and work can continue on those separate PRs.

Please rebase everything in a single commit.

PR #9370, opened 2022-11-29
feat: Make GrafeasClient simpler to work with

jskeet:
Just because we don't have a channel pool doesn't really prevent us from using credentials with no scopes - we just create a channel each time. (This doesn't affect existing working clients.)

With this in place, we can run regular smoke tests, and the client creation tests can pass too.


apis/Grafeas.V1/Grafeas.V1/GrafeasClientBuilder.cs
```diff
@@ -24,18 +24,24 @@ public partial class GrafeasClientBuilder
...
-            // We can handle Endpoint + (credentials or scopes), or a CallInvoker
-            if (Endpoint is object && (ChannelCredentials is object || Scopes is object))
-            {
-                return;
-            }
-            if (CallInvoker is object)
+            // If an endpoint is specified but no scopes, it's still possible that the credentials
+            // will work. We let that fail normally. But we definitely need to either have a CallInvoker
+            // or Endpoint, as otherwise we have nowhere to connect to.
+            // Note that if no scopes have been set and no other
```
amanda-tarafa, line 30: Incomplete comment.

amanda-tarafa (approved):
LGTM

apis/Grafeas.V1/Grafeas.V1/GrafeasClientBuilder.cs
jskeet: Whoops, thanks, will fix.

amanda-tarafa (approved):
LGTM

PR #9406, opened 2022-12-06
chore: Updates for REST numeric enum support

jskeet:
Overnight we received a large PR with a bunch of generated changes due to BUILD.bazel files being updated to have rest_numeric_enums=true.

This PR:

- Adds tooling support for local generation (which I thought we already had, but never mind...)
- Updates apis.json to indicate which APIs have numeric enum support
- Regenerates all APIs, which just updates APIs which OwlBot would have regenerated (e.g. due to pregeneration.sh changes)

There are a couple of additional changes with a full regeneration, but I'll create a separate PR for those - this PR should *only* be about numeric enums.


amanda-tarafa (approved):
LGTM

PR #9407, opened 2022-12-06
chore: regenerate all APIs

jskeet:
I'm slightly unclear as to why these were missed before, but they seem fine to merge now.


amanda-tarafa (approved):
LGTM, yep, they do look fine.

PR #9420, opened 2022-12-08
feat: Enable self-signed JWTs for Storage and Translation clients

### LINKED ISSUE #9326: Enable UseSelfSignedJwts by default in Storage/Translation
Once GAX 2.3.0 is released, we should modify StorageClientBuilder and TranslationClientBuilder (and AdvancedTranslationClientBuilder) to enable the use of self-signed JWTs with scopes by default.

We can't do this *just* by default, as BigQuery doesn't appear to support self-signed JWTs with scopes at the moment, and we don't want to break users if they just update GAX. (Even if we explicitly disable it in the BQ library in an update, that won't help users who update GAX but not BQ.)

jskeet:
This requires GAX 4.3.0-beta01; if we need to do a GA release of Storage or Translation before GAX 4.3.0 is released, we'll need to comment out the line in each builder.

Fixes #9326


amanda-tarafa (approved):
LGTM

PR #9484, opened 2022-12-20
docs: Updated troubleshoot guide to show how to extract ErrorInfo from RpcException

anuragsrivstv:


docs/devsite-help/troubleshooting.md
```diff
@@ -141,6 +141,35 @@ In that case, the simplest fix is to add a direct dependency to
 `Grpc.Core` from your application, which will ensure that the
 native libraries are copied appropriately.
 
+## How to get ErrorInfo from SpannerException
+
+The ErrorInfo message provides error information that both humans and
+applications can depend on. If SpannerException contains RPCException
+as it's inner execption, we can extract ErrorInfo from SpannerException.
```
amanda-tarafa, line 148:
```suggestion
as its inner exception, we can extract ErrorInfo from SpannerException.
```

amanda-tarafa (commented):
I think this is better suited for Spanner specific documentation?
But I'll defer to @jskeet when he gets back.
Marking as `do not merge` for the time being.

amanda-tarafa:
This is not a new feature, this is a change in documentation, so the conventional commit prefix should be `docs:`. The commit and the PR name should both have it, the commit in particular is used to generate version history and classify commits according to their types.

amanda-tarafa:
FYI: https://www.conventionalcommits.org/en/v1.0.0/

Rishabh-V (commented):
I'll review this tomorrow morning. Added label `lex-review`, so that LEX team reviews it before requesting review from Jon and Amanda.

docs/devsite-help/troubleshooting.md
```diff
@@ -141,6 +141,35 @@ In that case, the simplest fix is to add a direct dependency to
 `Grpc.Core` from your application, which will ensure that the
 native libraries are copied appropriately.
 
+## How to get ErrorInfo from SpannerException
```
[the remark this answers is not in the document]
jskeet: No, "How to get..." isn't a question - it's just information. (It would be "How can I get ErrorInfo from SpannerException?" or similar.)

But yes, I'd make this primarily about RpcException.

(This is my one drive-by comment...)

docs/devsite-help/troubleshooting.md
```diff
@@ -214,6 +214,35 @@ In that case, the simplest fix is to add a direct dependency to
...
+
+The ErrorInfo message provides error information that both humans and
```
jskeet, line 219:
```suggestion
The `ErrorInfo` message provides error information that both humans and
```

docs/devsite-help/troubleshooting.md
```diff
@@ -214,6 +214,35 @@ In that case, the simplest fix is to add a direct dependency to
...
+applications can depend on. If SpannerException contains RPCException
```
jskeet, line 220:
```suggestion
applications can depend on. If a `SpannerException` contains an `RpcException`
```

docs/devsite-help/troubleshooting.md
jskeet: I'd suggest also using backticks for `ErrorInfo` and `SpannerException`

docs/devsite-help/troubleshooting.md
```diff
@@ -214,6 +214,35 @@ In that case, the simplest fix is to add a direct dependency to
...
+
+   // The reason for the error. Error reasons are unique within a particular domain of errors.
+   var reason = errorInfo.Reason;
+
+   // Additional structured details about this error.
+   var metadata = errorInfo.Metadata;
+
+   // Log domain , reason and metadata of ErrorInfo.
+    }
+```
```
jskeet, line 245:
```suggestion
```csharp
try
{
    // Code using Spanner
}
catch (SpannerException ex) when (ex.InnerException is RpcException rpcException )
{
    var errorInfo = rpcException.GetErrorInfo();

    // Domain is the logical grouping to which the "reason" belongs.   
    var domain = errorInfo.Domain;

    // The reason for the error. Error reasons are unique within a particular domain of errors.
    var reason = errorInfo.Reason;

    // Additional structured details about this error.
    var metadata = errorInfo.Metadata;

    // Use the domain, reason and metadata of ErrorInfo, e.g. for logging or taking remedial action.
}
```

jskeet (commented):
I would definitely suggest changing this to be focused on RpcException. We can have one example showing that in detail... then a subsection for SpannerException, with a brief example using the pattern match (and maybe having another catch clause for SpannerException when there *isn't* an inner RpcException?) that refers to the earlier example.

anuragsrivstv:
Will open a new PR #9560

PR #9509, opened 2023-01-06
feat: Support HMAC URL signing.

### LINKED ISSUE #6026: Cloud Storage UrlSigner support for HMAC sigs
Looking into the UrlSigner I noticed it specifically hardcodes the `GOOG4-RSA-SHA256` algo. Could an extension to the UrlSigner be made to support HMAC signing?

The reason for this ask is that RSA signing is rather computationally expensive at the amount of signatures we create per request and we would like to switch to HMAC, ideally without having to maintain the requisite hmac secret derivation and signing bits.

amanda-tarafa:
Closes #6026

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
@@ -47,8 +47,8 @@ namespace Google.Cloud.Storage.V1.Snippets
         {
             var bucketName = _fixture.BucketName;
             var objectName = _fixture.HelloStorageObjectName;
-            var credential = (await GoogleCredential.GetApplicationDefaultAsync()).UnderlyingCredential as ServiceAccountCredential;
-            var httpClient = new HttpClient();
+            var credential = (ServiceAccountCredential) (await GoogleCredential.GetApplicationDefaultAsync()).UnderlyingCredential;
+            using var httpClient = new HttpClient();
 
             // Sample: SignedURLGet
             // Additional: Sign(string,string,TimeSpan,*,*)
@@ -64,6 +64,47 @@ namespace Google.Cloud.Storage.V1.Snippets
             Assert.Equal(_fixture.HelloWorldContent, content);
         }
 
+        [Fact]
+        public async Task HmacSignedURLGet()
+        {
+            var bucketName = _fixture.BucketName;
+            var objectName = _fixture.HelloStorageObjectName;
+            var credential = (ServiceAccountCredential) (await GoogleCredential.GetApplicationDefaultAsync()).UnderlyingCredential;
+            using var httpClient = new HttpClient();
+
+            using var storageClient = await StorageClient.CreateAsync();
+            var hmacKey = await storageClient.CreateHmacKeyAsync(_fixture.ProjectId, credential.Id);
+            var hmacKeyId = hmacKey.Metadata.AccessId;
+            var hmacKeySecret = hmacKey.Secret;
+
+            try
+            {
+
+                // Sample: HmacSignedURLGet
+                // Additional: Sign(string,string,TimeSpan,*,*)
+                // Create an HmacBlobSigner from an HMAC Key ID and Secret.
+                UrlSigner.IBlobSigner blobSigner = UrlSigner.HmacBlobSigner.Create(hmacKeyId, hmacKeySecret);
+                // Create a URL signer from the HmacBlobSigner.
+                UrlSigner urlSigner = UrlSigner.FromBlobSigner(blobSigner);
+                // Create an HMAC signed URL which can be used to get a specific object for one hour.
+                string url = urlSigner.Sign(bucketName, objectName, TimeSpan.FromHours(1));
+
+                // Get the content at the created URL.
+                HttpResponseMessage response = await httpClient.GetAsync(url);
+                string content = await response.Content.ReadAsStringAsync();
+                // End sample
+
+                Assert.Equal(_fixture.HelloWorldContent, content);
+            }
+            finally
+            {
+                // We need to deactivate key before we can delete it.
+                hmacKey.Metadata.State = HmacKeyStates.Inactive;
+                await storageClient.UpdateHmacKeyAsync(hmacKey.Metadata);
+                await storageClient.DeleteHmacKeyAsync(_fixture.ProjectId, hmacKey.Metadata.AccessId);
+            }
+        }
+
         // See-also: Sign(string,string,TimeSpan,*,*)
         // Member: Sign(UrlSigner.RequestTemplate, UrlSigner.Options)
         // See [Sign](ref) for an example using an alternative overload.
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
```diff
@@ -64,6 +64,41 @@ public async Task SignedURLGet()
             Assert.Equal(_fixture.HelloWorldContent, content);
         }
 
+        [Fact]
+        public async Task HmacSignedURLGet()
+        {
+            var bucketName = _fixture.BucketName;
+            var objectName = _fixture.HelloStorageObjectName;
+            var credential = (await GoogleCredential.GetApplicationDefaultAsync()).UnderlyingCredential as ServiceAccountCredential;
```
jskeet, line 72: I'd suggest using a cast here instead of `as` so that we get a useful exception here instead of a NullReferenceException later on, if the credential isn't a service account.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
```diff
@@ -64,6 +64,41 @@ public async Task SignedURLGet()
...
+            var httpClient = new HttpClient();
+
+            using var storageClient = await StorageClient.CreateAsync();
```
jskeet, line 75: I wonder whether this should be in the sample?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.HmacBlobSigner.cs
```diff
@@ -0,0 +1,89 @@
...
+            (Id, _secret, _prefix) = (keyId.ThrowIfNullOrEmpty(nameof(keyId)), secret.ThrowIfNullOrEmpty(nameof(secret)), DefaultPrefix);
+
+        /// <summary>
+        /// Creates a new HMAC Blob Signer from an HMAC Key ID and Secret.
+        /// </summary>
+        /// <param name="keyId">The HMAC key ID. Must not be null or empty.</param>
+        /// <param name="secret">The HMAC key secret. Must not be null or empty.</param>
+        public static HmacBlobSigner Create(string keyId, string secret) => new HmacBlobSigner(keyId, secret);
+
+        /// <inheritdoc/>
```
jskeet, line 52: Can we move these properties to above the constructor?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.HmacBlobSigner.cs
```diff
@@ -0,0 +1,89 @@
...
+        private readonly string _secret;
+        private readonly string _prefix;
+
+        /// <summary>
+        /// Creates a new HMAC Blob Signer from an HMAC Key ID and Secret.
+        /// </summary>
+        /// <param name="keyId">The HMAC key ID. Must not be null or empty.</param>
+        /// <param name="secret">The HMAC key secret. Must not be null or empty.</param>
+        private HmacBlobSigner(string keyId, string secret) =>
+            (Id, _secret, _prefix) = (keyId.ThrowIfNullOrEmpty(nameof(keyId)), secret.ThrowIfNullOrEmpty(nameof(secret)), DefaultPrefix);
```
jskeet, line 43: Presumably the plan is that we can allow prefix customization at a later date if necessary...

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.HmacBlobSigner.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// <inheritdoc/>
+        public string Id { get; }
+
+        /// <inheritdoc/>
+        public string Algorithm => "GOOG4-HMAC-SHA256";
+
+        /// <inheritdoc/>
+        public string CreateSignature(byte[] data, BlobSignerParameters parameters)
+        {
+            data.ThrowIfNull(nameof(data));
```
jskeet, line 61: Nit: in google-cloud-dotnet we usually use `GaxPreconditions.CheckNotNull` instead.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.HmacBlobSigner.cs
```diff
@@ -0,0 +1,89 @@
...
+            parameters.ThrowIfNull(nameof(parameters));
```
jskeet, line 62: Should we validate that everything we use *within* parameters is non-null, or is that already handled for us?

jskeet (commented):
Just a few nits - generally happy.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
amanda-tarafa: Changing here and from where I copied it from :)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
amanda-tarafa: We have plenty of samples for managing HMAC keys and a docs page explaining (I've linked to it on the text that accompanies this example). So:

- I'd prefer not to duplicate this as a sample.
- Don't give the appearance of the need to create an HMAC key each time. They may be stored in files, secret mamangers, etc.

Let me know.

amanda-tarafa: As a side note, I'll add deletion of HMAC keys to the CleanData project. Only 5 can be created, and if deletion fails, then we'll reach quota soon.

And delete on finally here as well.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.HmacBlobSigner.cs
amanda-tarafa: Yes, exactly. The prefix itself could have been harcoded at the moment, but the constructor is private and the factory method is there precisely in case we need to allow other prefixes in the future.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.HmacBlobSigner.cs
amanda-tarafa: Yep, done.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.HmacBlobSigner.cs
amanda-tarafa: Ha, yes, done.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UrlSigner.HmacBlobSigner.cs
amanda-tarafa: Handled elsewhere.
BlobSignerParameters is inmutable, and can only be created from a timestamp, which is required, all the other fields have default values currently, just there in case they are customizable in the future.
This was part of the breaking changes we introduced when increasing major in preparation for HMAC support.

amanda-tarafa (commented):
Review addressed in latest commit.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/UrlSignerSnippets.cs
jskeet: All fine by me.

jskeet (approved):

PR #9530, opened 2023-01-11
chore: Update API catalog for Vmware Engine to include numeric enums

jskeet:
(This was missed when the API was added. We should probably automate updating this information from the BUILD.bazel file...)


amanda-tarafa (approved):
LGTM

PR #9531, opened 2023-01-11
chore: Update dependencies in project generation

jskeet:
This is mostly to handle the GAX 4.3.0 release, but there are a few other dependencies which we can update at the same time.


amanda-tarafa (approved):
LGTM

PR #9542, opened 2023-01-11
chore: Tool to update API catalog based on BUILD.bazel files

jskeet:
- Commit 1 is tooling code
- Commit 2 is the result of running the tool *and then rejecting some changes* (I need to investigate those)
- Commit 3 is the result of regenerating the affected APIs

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
@@ -1,0 +1,138 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using Google.Cloud.Tools.Common;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Google.Cloud.Tools.ReleaseManager;
+
+/// <summary>
+/// A class to update parameters in the API catalog based on BUILD.bazel files.
+/// (This is a stopgap measure until we've got a single configuration source.)
+/// </summary>
+public sealed class UpdateFromBazelCommand : CommandBase
+{
+    public UpdateFromBazelCommand() : base("update-from-bazel", "Updates the API catalog based on BUILD.bazel files", "id")
+    {
+    }
+
+    protected override void ExecuteImpl(string[] args)
+    {
+        string id = args[0];
+        var catalog = ApiCatalog.Load();
+
+        var ids = id == "all"
+            ? catalog.Apis.Where(api => api.ProtoPath is string && api.Generator == GeneratorType.Micro).Select(api => api.Id).ToArray()
+            : new[] { id };
+
+        bool anyUpdated = false;
+        foreach (var apiId in ids)
+        {
+            if (Update(catalog[apiId]))
+            {
+                anyUpdated = true;
+                Console.WriteLine($"Modified configuration for {apiId}");
+            }
+        }
+
+        if (anyUpdated)
+        {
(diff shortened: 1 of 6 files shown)


tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+        if (!File.Exists(bazelFile))
+        {
+            throw new UserErrorException($"Expected Bazel file does not exist: '{bazelFile}'");
+        }
+
+        var lines = File.ReadAllLines(bazelFile).ToList();
+        var start = lines.IndexOf("csharp_gapic_library(");
+        if (start == -1)
+        {
+            // This is effectively a warning. (AI Platform has it commented out, for example.)
```
amanda-tarafa, line 74: Wouldn't you still find it if it's commented out?

tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+            Console.WriteLine($"Bazel file for {api.Id} does not include csharp_gapic_library target.");
+            return false;
+        }
+        var end = lines.IndexOf(")", start);
+        // Second check is just a safety-net, in case the end of the target is formatted badly and we've picked up a later one.
+        if (end == -1 || end - start > 20)
```
amanda-tarafa, line 80: Is it fixed what can go in there? Or bounded by 20 characters I guess?

tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+            .Take(end - start)
+            .Select(line => line.TrimEnd(','))
+            .Where(line => line.Contains('='))
+            .Select(line => line.Split('=', 2))
+            .ToDictionary(bits => bits[0].Trim(), bits => bits[1].Trim('"', ' '));
+
+        bool updated = false;
+
+        bool restNumericEnums = properties.TryGetValue("rest_numeric_enums", out var restNumericEnumsValue) && restNumericEnumsValue == "True";
+        if (api.RestNumericEnums != restNumericEnums)
```
amanda-tarafa, line 97: Is there a check somewhere that once enabled, numeric enums cannot be disabled?
I know you are checking for True only above, which is fine, but if they were enabled once and are now disabled, our apis.json will be out of sync with the bazel files. I guess that disabling the numeric enums is a breaking change, so it should be caught upstream, just wondering.

tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            api.RestNumericEnums = restNumericEnums;
+            UpdateOrAddJson("restNumericEnums", restNumericEnums);
+            updated = true;
+        }
+
+        string transport = properties.GetValueOrDefault("transport", "grpc");
```
amanda-tarafa, line 104: (And disabling json is also a breaking change that should be catched upstream)

amanda-tarafa (approved):
LGTM

tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
jskeet: Nope - because then the line is `# csharp_gapic_library`.

tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
jskeet: 20 lines - basically I'd be astonished to see any *correct* rule go on for more than 20 lines. (And if it does, we can just extend this.) It's just an attempt at a bit of a safeguard.

tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
jskeet: Nope, if it's true before, but then set to "False" (or *anything* other than "True") then we'll propagate that to the config. (Eventually "true" may become the default, but we can deal with that later...)

tools/Google.Cloud.Tools.ReleaseManager/UpdateFromBazelCommand.cs
amanda-tarafa: Ah yes, you are looking for whole lines, not within a string. Duh.

PR #9550, opened 2023-01-13
docs: Add commentary on accidentally-optional parameters

### LINKED ISSUE #9488: StorageClient.CopyObjectAsync has an unfortunate signature
From [StorageObject.CopyObject.cs](https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.CopyObject.cs):

```csharp
        /// <summary>
        /// Creates a copy of an object synchronously, potentially to a different bucket. This method uses the
        /// <c>rewriteObject</c> underlying API operation for more flexibility and reliability.
        /// </summary>
        /// <param name="sourceBucket">The name of the bucket containing the object to copy. Must not be null.</param>
        /// <param name="sourceObjectName">The name of the object to copy within the bucket. Must not be null.</param>
        /// <param name="destinationBucket">The name of the bucket to copy the object to. Must not be null.</param>
        /// <param name="destinationObjectName">The name of the object within the destination bucket. Must not be null.</param>
        /// <param name="options">Additional options for the copy operation. May be null, in which case appropriate
        /// defaults will be used.</param>
        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
        /// <returns>A task representing the asynchronous operation, with a result returning the
        /// <see cref="Object"/> representation of the new storage object resulting from the copy.</returns>
        public virtual Task<Object> CopyObjectAsync(
            string sourceBucket,
            string sourceObjectName,
            string destinationBucket = null,
            string destinationObjectName = null,
            CopyObjectOptions options = null,
            CancellationToken cancellationToken = default)
        {
            throw new NotImplementedException();
        } 
```

We default `destinationBucket` and `destinationObjectName` to null, despite `null` not being a valid value for those parameters.

We probably can't fix this until the next major version, unfortunately - anything we do is likely to lead to ambiguity. I wanted to record this now, but we can probably close the issue once we've worked out what we'll do and how we'll remember to do it...

jskeet:
Fixes #9488 (as far as we can for now)

The required change for the next major version will be recorded internally.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.CopyObject.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.CopyObject.cs
@@ -53,6 +53,12 @@ namespace Google.Cloud.Storage.V1
         /// <param name="options">Additional options for the copy operation. May be null, in which case appropriate
         /// defaults will be used.</param>
         /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <remarks>
+        /// The fact that <paramref name="destinationBucket"/> and <paramref name="destinationObjectName"/> are
+        /// optional is a mistake. These default value of null for these parameters is invalid for this method,
+        /// meaning that any call which doesn't specify the parameters explicitly will fail. Making these parameters
+        /// required would be a compile-time breaking change; this will be implemented in the next major version of this library.
+        /// </remarks>
         /// <returns>A task representing the asynchronous operation, with a result returning the
         /// <see cref="Object"/> representation of the new storage object resulting from the copy.</returns>
         public virtual Task<Object> CopyObjectAsync(


Rishabh-V (approved):
LGTM.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.CopyObject.cs
```diff
@@ -53,6 +53,12 @@ public abstract partial class StorageClient
         /// <param name="options">Additional options for the copy operation. May be null, in which case appropriate
         /// defaults will be used.</param>
         /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <remarks>
+        /// The fact that <paramref name="destinationBucket"/> and <paramref name="destinationObjectName"/> are
+        /// optional is a mistake. These default value of null for these parameters is invalid for this method,
```
Rishabh-V, line 58: Just read it again and I think that "The default value of null for these parameters is invalid for this method..." sounds better, so maybe `These` => `The`. Please ignore if this doesn't make sense and sorry for an afterthought.

jskeet: Whoops, yes, it was intended to be "The". Will include this change in a separate commit but on a PR which is already dealing with other Storage matters.

PR #9552, opened 2023-01-13
tests: Update job labels tests to match new API behaviour.

amanda-tarafa:
Before, a label with a null value was ignored by the API. Now it's set with an empty string value.


jskeet (approved):

amanda-tarafa:
All tests are now green on local. Merging...

PR #9553, opened 2023-01-13
feat: Add DownloadValidationMode.Automatic

### LINKED ISSUE #9507: Skip hash validation when it's unsafe when downloading Storage objects
In #1641 I proposed a `HashValidationMode` of `WhenAvailable` or similar - but we couldn't implement it at the time.

With the new `x-goog-stored-content-encoding` header, we may be able to implement this - see the [Node PR](https://github.com/googleapis/nodejs-storage/pull/2023/files) for an example.

Ideally, we'd want to make this the default - but that should only happen at the next major version of Google.Cloud.Storage.V1.

jskeet:
Fixes #9507 (modulo the change to default in the next majro version)

This detects if the x-goog-stored-content-encoding header is set to gzip, but the actual content encoding of the response is non-gzip, and disables validation in that specific case - while validating in the same way as normal for everything else.

We will make this the default in the next major version. (This is recorded in the internal document.)

Implementation note: this now *always* creates a HashValidatingDownloader, and puts that in control of whether validation is actually performed or not. That's simpler than the previous split of the choice being in StorageClientImpl.DownloadObject.cs with the hashing in HashValidatingDownloader.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/HashValidatingDownloaderTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/HashValidatingDownloaderTest.cs
@@ -36,9 +36,80 @@ namespace Google.Cloud.Storage.V1.Tests
         [InlineData(Md5Value + "," + Crc32Value)]
         [InlineData(Crc32Value + "," + Md5Value)]
         [InlineData(Crc32Value)]
-        public void Valid(string headerValue)
+        public void Valid_ModeAlways(string headerValue)
         {
-            HashValidatingDownloader downloader = CreateDownloader(request => CreateResponse(request, headerValue));
+            HashValidatingDownloader downloader = CreateDownloader(request => CreateResponse(request, headerValue), DownloadValidationMode.Always);
+            AssertDownloadSucceeds(downloader);
+        }
+
+        [Theory]
+        [InlineData("gzip", "gzip")]
+        [InlineData(null, "gzip")]
+        [InlineData("plain", null)]
+        [InlineData("plain", "gzip")]
+        [InlineData("gzip", "plain")]
+        [InlineData("gzip", null)]
+        [InlineData(null, null)]
+        [InlineData("plain", "plain")]
+        public void Valid_ModeAutomatic(string storedContentEncoding, string contentEncoding)
+        {
+            // This test doesn't try different hashes exhaustively; the way it treats the header
+            // is the same as with Always.
+            string headerValue = Md5Value + "," + Crc32Value;
+            HashValidatingDownloader downloader = CreateDownloader(
+                request => CreateResponse(request, headerValue, storedContentEncoding, contentEncoding),
+                DownloadValidationMode.Automatic);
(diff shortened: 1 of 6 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/HashValidatingDownloaderTest.cs
```diff
@@ -36,9 +36,76 @@ public class HashValidatingDownloaderTest
...
+            AssertDownloadSucceeds(downloader);
+        }
+
+        [Theory]
+        [InlineData("gzip", "gzip")]
+        [InlineData(null, "gzip")]
+        [InlineData("plain", null)]
+        [InlineData("plain", "gzip")]
+        [InlineData("gzip", "plain")]
+        [InlineData("gzip", null)]
```
Rishabh-V, line 51: Would it be helpful to have 
```
[InlineData(null, null)]
[InlineData("plain","plain")]
```
or are these invalid?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/HashValidatingDownloaderTest.cs
```diff
@@ -97,12 +162,22 @@ private static Initializer GetInitializer(Func<HttpRequestMessage, HttpResponseM
...
+                response.Headers.Add(Crc32c.HashHeaderName, hashHeaderValue);
+            }
+            if (storedContentEncoding != null)
+            {
+                response.Headers.Add(HashValidatingDownloader.StoredContentEncodingHeaderName, storedContentEncoding);
+            }
+            if (contentEncoding != null)
             {
-                response.Headers.Add(Crc32c.HashHeaderName, headerValue);
+                response.Content.Headers.ContentEncoding.Add(contentEncoding);
```
Rishabh-V, line 180: 👍🏼

Rishabh-V (approved):
LGTM

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/HashValidatingDownloader.cs
```diff
@@ -24,54 +24,82 @@
...
+                        }
                     }
                 }
+                // The mode indicates that we'd like to validate the hash, but we don't have one to validate.
+                // (We could potentially have a mode of "fail if there isn't a hash to validate", in the future,
+                // but for the moment we just ignore it.)
+                return null;
             }
-            return null;
+
```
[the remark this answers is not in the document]
jskeet: Nope - we don't need either of them, in fact. Will fix.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/HashValidatingDownloaderTest.cs
jskeet: I don't think either would actually happen, but I'm happy to add them. (And to InvalidData_ModeAutomatic_Validated as well.)

PR #9560, opened 2023-01-15
docs: Updates troubleshoot guide to show how to extract `ErrorInfo` from `RpcException`.

anuragsrivstv:
- Updates troubleshoot file to show how to extract ErrorInfo from RpcException.

- Includes a sample to demo how to use RpcException property of SpannerException.


docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,62 @@ native libraries are copied appropriately.
...
+- `Metadata`- It is the additional structured details about the error. 
+It can help in identifying the exact cause of the error.
+
+```csharp
+using Google.Api.Gax.Grpc;
+try
+{
+  // Code that can throw the RpcException.
+}
+catch (RpcException  ex)
```
hemanshv, line 246: nit: multiple space before ex

hemanshv:
Added comments for 2 minor nits.
LGTM

docs/devsite-help/troubleshooting.md
Rishabh-V: yep. Should be just one whitespace.

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,62 @@ native libraries are copied appropriately.
...
+{
+   var errorInfo = ex.GetErrorInfo();   
+   var domain = errorInfo.Domain;
+   var reason = errorInfo.Reason;
+   var metadata = errorInfo.Metadata;
+
+   // Use the domain, reason and metadata of ErrorInfo, e.g. for logging or taking remedial action.
```
Rishabh-V, line 253: What remedial action?
Only reason I say this is that without concrete examples, this may paint an incomplete picture and the reader may end up being curious about it with no link or reference to the additional information. I 'll let Jon decide on it so please don't make any changes for now.

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,62 @@ native libraries are copied appropriately.
...
+}
+```
+
+### How can I get RpcException from SpannerException?
```
Rishabh-V, line 257: Given the changes have been pushed recently and not released yet, we *may* want to release those changes and then say that "starting version 4.X.X or higher", so that users of existing versions of library understand they need to upgrade and this doesn't create confusion.

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,62 @@ native libraries are copied appropriately.
...
+
+The `RpcException` property of `SpannerException` returns the underlying
+`RpcException` which caused the `SpannerException`. The returned 
+value may be null if the `SpannerException` was not caused by `RpcException`.
```
Rishabh-V, line 261: Maybe 
```suggestion
value may be null if the `SpannerException` was not caused by an `RpcException`.
```

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,62 @@ native libraries are copied appropriately.
...
+
+`RpcException` property of `SpannerException` can be used like:
+
+```csharp
+using Google.Api.Gax.Grpc;
+try
+{
+  // Code using Spanner
```
Rishabh-V, line 269: nit: `.` at the end of comment.

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,62 @@ native libraries are copied appropriately.
...
+```csharp
+using Google.Api.Gax.Grpc;
+try
+{
+  // Code using Spanner 
+}
+catch (SpannerException ex)
+{
+   var rpcException = ex.RpcException;   
+   if(rpcException != null)
```
Rishabh-V, line 274: missing space between `if` and `(`
```suggestion
   if (rpcException != null)
```

Rishabh-V: Also, I see in https://github.com/googleapis/google-cloud-dotnet/pull/9484#pullrequestreview-1234636109 that Jon mentions that
```
I would definitely suggest changing this to be focused on RpcException. We can have one example showing that in
detail. We can have one example showing that in detail... then a subsection for SpannerException, with a brief
example using the pattern match (and maybe having another catch clause for SpannerException when there isn't an
inner RpcException?) that refers to the earlier example.
```
so note on `SpannerException`  should be a subsection of the above `RpcException` section instead of a new section (I may be wrong, but wanted to check and be sure)

Rishabh-V (commented):
Few minor nits. Please have a look and request Jon's review to finalize the changes. Thank you.

docs/devsite-help/troubleshooting.md
anuragsrivstv: Make sense. So should we wait for its release before merging this PR ?

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,62 @@ native libraries are copied appropriately.
...
+
+- `Domain` - It is the logical grouping to which the error belongs. It is 
+typically the registered service name of the tool or product that 
+generates the error. Example: "pubsub.googleapis.com". If the error 
+is generated by some common infrastructure, the error domain must be 
+a globally unique value that identifies the infrastructure. 
+For Google API infrastructure, the error domain is "googleapis.com".
+
+- `Metadata`- It is the additional structured details about the error. 
+It can help in identifying the exact cause of the error.
```
jskeet, line 238: You need to specify the way in which this is constant. (I understand this text is taken from error_details.proto, but we can clarify here.)

I'd also avoid the "It is" wording here and below. Here's a potential rewriting:

```suggestion
- `Reason`: the proximate cause of the error. This is a stable unique identifier for this
   type of error within the domain, and can be used to programmatically determine
   how to handle the error.
- `Domain`: the logical grouping to which the error belongs. The error domain
   is typically the registered service name of the tool or product that
   generates the error. Example: "pubsub.googleapis.com". If the error 
   is generated by some common infrastructure, the error domain will be 
   a globally unique value that identifies the infrastructure. 
For Google API infrastructure, the error domain is "googleapis.com".
- `Metadata`: additional structured details about the error.
```

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,62 @@ native libraries are copied appropriately.
...
+
+```csharp
+using Google.Api.Gax.Grpc;
+try
+{
+  // Code that can throw the RpcException.
```
jskeet, line 244:
```suggestion
    // Code that can throw the RpcException.
```

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,62 @@ native libraries are copied appropriately.
...
+}
+catch (RpcException ex)
+{
+   var errorInfo = ex.GetErrorInfo();   
+   var domain = errorInfo.Domain;
+   var reason = errorInfo.Reason;
+   var metadata = errorInfo.Metadata;
```
jskeet, line 251:
```suggestion
    if (ex.GetErrorInfo() is ErrorInfo errorInfo)
    {
        var domain = errorInfo.Domain;
        var reason = errorInfo.Reason;
        var metadata = errorInfo.Metadata;
        // ...
    }
```

(It's entirely possible for `ex.GetErrorInfo()` to return null.

jskeet: We may want to have a comment about what the user can still do if there isn't an ErrorInfo, mind you. (They still have the status and the message.)

docs/devsite-help/troubleshooting.md
jskeet: That's really application-specific and API-specific. I think it's fine to leave the comment as is (within the suggestion above).

docs/devsite-help/troubleshooting.md
jskeet: Yes, or split the PR into two: one just for RpcException and one for SpannerException.
Alternatively, given that this is completely Spanner-specific, we could move it into the Spanner docs - and merge that now, so that it'll be published in the next Spanner release. I think that would be better.

anuragsrivstv: Sure. I'll remove SpannerException section from troubleshoot.md and include it in Spanner docs in a separate PR once the help page with RpcException usage is published.

docs/devsite-help/troubleshooting.md
anuragsrivstv: Incorporated changes.

docs/devsite-help/troubleshooting.md
anuragsrivstv: Sure.

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,42 @@ native libraries are copied appropriately.
...
+- `Domain`: the logical grouping to which the error belongs. The error domain
+   is typically the registered service name of the tool or product that
+   generates the error. Example: "pubsub.googleapis.com". If the error 
+   is generated by some common infrastructure, the error domain will be 
+   a globally unique value that identifies the infrastructure. 
+   For Google API infrastructure, the error domain is "googleapis.com".
+- `Metadata`: additional structured details about the error.
+
+```csharp
+using Google.Api.Gax.Grpc;
```
jskeet, line 239: This example still has unconventional indentation - it looks like it's three spaces for the first indent, and two for the second.

docs/devsite-help/troubleshooting.md
```diff
@@ -218,3 +218,42 @@ native libraries are copied appropriately.
...
+   {
+      var domain = errorInfo.Domain;
+      var reason = errorInfo.Reason;
+	  var metadata = errorInfo.Metadata;
+      // Use the domain, reason and metadata of ErrorInfo, e.g. for logging or taking remedial action.
+   }
+   else
+   {
+      // Use RpcException's other properties like status and message to handle the received exception
+      // as per the application logic demands.
```
jskeet, line 256:
```suggestion
      // Use RpcException's other properties (e.g. Status and Message) to handle the exception appropriately
```

docs/devsite-help/troubleshooting.md
anuragsrivstv: Reformatted the indentation.

jskeet (approved):

anuragsrivstv:
Merging now.

PR #9562, opened 2023-01-16
chore: Update API catalog from Bazel files

jskeet:
These changes do not affect the generated code, so it's a harmless change. (Both are "unusual" APIs, so I'm not surprised that there are some anomalies in terms of different information sources.)


amanda-tarafa (approved):
LGTM

PR #9564, opened 2023-01-16
chore: Upgrade to docfx 2.60.0, and use it as a dotnet tool

jskeet:
Once we've made sure this works with all our existing processes, we can then start moving to use Linux in CI for docs building, and update other repos to do the same.


amanda-tarafa (approved):
LGTM

PR #9604, opened 2023-01-17
chore: Update Bazel modification tool to limit modifications

jskeet:
- Restrict modified files to those we generate
- Restrict the number of files to modify per run

This will allow us to roll out releases in a controlled way.


amanda-tarafa (approved):
LGTM

PR #9625, opened 2023-01-18
chore: Update transport tool to require rest_numeric_enums=True

jskeet:
This will avoid manually reverting the 11 APIs which would otherwise be copied by the tool.


amanda-tarafa (approved):
LGTM

PR #9663, opened 2023-01-19
chore: Remove the hack that was in place for the previous round of releases

jskeet:
At some point it *may* be worth looking at times that we've needed something like this, and coming up with more of a framework - but I suspect the reasons vary too much.


amanda-tarafa (approved):
LGTM

PR #9777, opened 2023-01-24
chore: Remove manual smoke tests

jskeet:
Review one commit at a time.

I'm rerunning all the smoke tests (with `runintegrationtests.sh --smoke`) now.

PR #9779, opened 2023-01-24
fix: Fix ContainerAnalysis for REST

jskeet:
This will be a single commit once #9777 has been merged.


jskeet:
Will fix the merge conflict in the morning and merge

jskeet:
Looks like this is stuck... I'll amend the commit to try to prod it.

jskeet:
Looks like this is now worse than stuck - all .NET Core 3.1 installation attempts are failing. I'll leave it for half an hour and try again, in case it's transient... I know .NET Core 3.1 is no longer supported, but I'd hope that installation wouldn't be rug-pulled with no warning. We may need to swiftly update everything to *only* test on .NET 6.

PR #9810, opened 2023-02-02
chore: Use .NET 6.0 for testing

jskeet:
This means we don't need to install .NET Core 3.1 at all in GitHub Actions.


jskeet:
Okay, this is now ready for review

amanda-tarafa (approved):
LGTM

PR #9831, opened 2023-02-06
chore: Skip integration tests where all code is generated

jskeet:
(This is typically for generated snippets.)

We'll separate these out reasonably soon anyway, but we might as well start skipping purely-generated non-tests now.


amanda-tarafa (approved):
LGTM

jskeet:
Nope, it skips if there *isn't* anything that *isn't* generated :)

PR #9874, opened 2023-02-14
DO NOT MERGE

jskeet:
I'm *hoping* this is all that's required for the tricky aspect of DML that we've talked about.


jskeet:
Will investigate emulator failures after lunch - I'm not particularly surprised that there are failures, given that I've changed behavior here. Will run integration tests to see what's going on.

jskeet:
(I may also create a PR for the second commit, and get that merged, so it's out of the way.)

jskeet:
Running the integration tests I get 13 failures as well - having had a quick look, I believe they're all due to *expecting* the read to be lazy. So that would be fixed by the change to only be eager for DML.

jskeet:
@anuragsrivstv: Closing this now as I believe it's already in your PR.

PR #9875, opened 2023-02-14
feat: Populate product URL automatically from service config

jskeet:
Best to review one commit at a time - and if you're uncomfortable with any of it, it can wait until Amanda comes back.

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
@@ -14,6 +14,7 @@
 
 using Google.Cloud.Tools.ApiIndex.V1;
 using Google.Cloud.Tools.Common;
+using Google.Protobuf;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
 using Newtonsoft.Json.Linq;
@@ -22,6 +23,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using YamlDotNet.Serialization;
 
 namespace Google.Cloud.Tools.ReleaseManager
 {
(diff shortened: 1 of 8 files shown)


tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
```diff
@@ -123,5 +128,29 @@ protected override void ExecuteImpl(string[] args)
...
+
+        private static Api.Service ParseServiceConfigYaml(string path)
+        {
+            if (path is null)
+            {
+                return null;
+            }
+
+            var deserializer = new Deserializer();
+            using (var reader = File.OpenText(path))
```
Rishabh-V, line 140: Should we check if path is valid and if the file exists, before reading?

tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
```diff
@@ -123,5 +128,29 @@ protected override void ExecuteImpl(string[] args)
...
+            {
+                var yamlObject = deserializer.Deserialize(reader);
+                var serializer = new SerializerBuilder().JsonCompatible().Build();
+                var writer = new StringWriter();
```
Rishabh-V, line 144: Should we use using block?

Rishabh-V (approved):
Fortunately, the feature changes are simple to understand. Just couple of questions for my understanding. Thanks.

tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
jskeet: If it's not, generation will fail when we try to use it in what's inevitably the next command anyway - better to fail at this point, I'd say :)

tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
jskeet: No need for StringWriter - it's all in memory anyway. It wouldn't do any harm, but it wouldn't be useful either.

PR #9880, opened 2023-02-15
chore: Add Microsoft.Extensions.DependencyInjection to PubSub tests

jskeet:
This was previously in just the project file, which meant it was removed when project files were regenerated (e.g. by OwlBot).

Including it in apis.json means it's also in the snippet and integration test projects, but that's harmless.


Rishabh-V (approved):
LGTM

PR #9881, opened 2023-02-15
fix(PubSub): Ack/ModAck failures in non-exactly once delivery flow should not be retried.

### LINKED ISSUE #8992: Acks/ModAcks should not be retried if exactly once delivery is disabled.
This is identified in the code review of #8973 

In the existing code for non-exactly once delivery, the recoverable errors like DeadlineExceeded are retried. This shouldn't be the case. Acks/modacks should not be retried if exactly once delivery is disabled. 

Please follow the discussion for complete context - https://github.com/googleapis/google-cloud-dotnet/pull/8973#discussion_r945063154

_Originally posted by @pradn in https://github.com/googleapis/google-cloud-dotnet/pull/8973#discussion_r945160232_

Rishabh-V:
closes #8992


jskeet (approved):
How feasible is it to write a test to validate the change in behavior? If it's not too hard, that would be good to see. But change LGTM anyway.

Rishabh-V:
Thanks Jon. There were no existing tests for the deleted code, so there was no impact on existing tests. 

I did think about a new test.  When we have an ack/nack request, it is effectively treated as fire and forget.
Earlier the failed messages were re-queued for retry. Now failed messages are not re-queued for retry.
 
So for unit tests, behavior remains the same. The difference will manifest in the actual behavior when the messages will either be re-delivered or have delayed deliveries, which I suspect will be difficult to replicate deterministically in an integration test.

jskeet:
Righto, that's fine - please go ahead and merge.

PR #9882, opened 2023-02-15
chore: Regenerate all APIs locally

jskeet:
These are changes that didn't come through automatically for various (unconcerning) reasons.

This PR does not include ArtifactRegistry changes which were surfaced by local regeneration; I'm following up on whether a particular proto should be published or not.


jskeet:
Just noticed that Dialogflow adds a feature - will update commit comment.

Rishabh-V (approved):
LGTM

PR #9885, opened 2023-02-16
refactor: Add nested classes in new files.

Rishabh-V:
Splits the nested classes into new files.
I have used the year as 2023 for the newly created nested files and kept it as 2018 for the parent classes.
Used file scoped namespaces across all the changed files.
Added a new `<ItemGroup>` section in csproj, so that the new files show as dependent files.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.RequeueableQueue.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.RequeueableQueue.cs
@@ -1,0 +1,119 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+
+namespace Google.Cloud.PubSub.V1;
+
+public sealed partial class SubscriberClientImpl
+{
+    // Internal for testing.
+    /// <summary>
+    /// A queue that allows data to be requeued at the front of the queue.
+    /// Requeued items are queued after previously requeued items, but before queued items.
+    /// </summary>
+    internal class RequeueableQueue<T>
+    {
+        private readonly Queue<T> _q = new Queue<T>();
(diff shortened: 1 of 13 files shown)


Rishabh-V:
Hi Jon,

This is not urgent. PTAL and provide feedback. I have not done it currently for the tests but happy to do it in the next iteration if you would like those to be done as well.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.RequeueableQueue.cs
```diff
@@ -0,0 +1,119 @@
...
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+
+namespace Google.Cloud.PubSub.V1;
+
+public sealed partial class SubscriberClientImpl
+{
+    // internal for testing.
```
jskeet, line 22: Nit: let's capitalize the "i" while we're here :)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -0,0 +1,204 @@
...
+        _scheduler = settings.Scheduler ?? SystemScheduler.Instance;
+        _clock = settings.Clock ?? SystemClock.Instance;
+        _taskHelper = GaxPreconditions.CheckNotNull(taskHelper, nameof(taskHelper));
+        _flowControlSettings = settings.FlowControlSettings ?? DefaultFlowControlSettings;
+        _useLegacyFlowControl = settings.UseLegacyFlowControl;
+        _maxAckExtendQueue = (int)Math.Min(_flowControlSettings.MaxOutstandingElementCount ?? long.MaxValue, 20_000);
+        _disposeTimeout = settings.DisposeTimeout ?? DefaultDisposeTimeout;
+    }
+
+    private readonly object _lock = new object();
```
jskeet, line 81: Please could you move all these fields and properties to before the constructor while we're at it?

(This is very old code, before we'd adopted some of our current conventions.)

jskeet: (I didn't spot whether the same is true in other files, but it would be good to just go through them.)

jskeet (approved):
A couple of nits, but otherwise great - this looks like it'll make things so much easier for reviewing changes!

I assume this was a really "pure" refactoring, in terms of not having to change any logic etc? If there were any bits where you actually had to think carefully beyond "where to split the file", let me know and I'll review those more carefully.

Rishabh-V (commented):
Thanks Jon. I have addressed the review comments.

Yes, this is pure refactoring. In this commit, I have only reordered the fields and properties and made the structs read-only. PTAL. Thanks.

jskeet (approved):
Thanks for moving things around - that's really helpful :) (It's bugged me every time I've worked on the PubSub code, but it's never been worth fixing - but while we're moving files...)

Rishabh-V:
Thanks Jon. I 'll squash and merge.

PR #9887, opened 2023-02-16
chore: Fix OwlBot post processor to permit file-scoped namespaces

jskeet:
(Basically the "broken namespace" check is too picky at the moment.)


Rishabh-V (approved):
LGTM

PR #9889, opened 2023-02-16
chore: Fix line endings in mixed-ending files

jskeet:
(Will raise a separate PR to update the check for this.)


Rishabh-V (approved):
LGTM.

Out of curiosity, is there a script/tool that you have written to identify and fix these?

jskeet:
A mixture of `git ls-files --eol` (on a Linux box) and `sed` :)

PR #9891, opened 2023-02-16
chore: Fix the line ending check

jskeet:
This checks that there are no mixed- or crlf-ending files in the index.


Rishabh-V (approved):
LGTM

PR #9893, opened 2023-02-16
chore: Fix names in obsolete messages

### LINKED ISSUE #9890: Is this obsolete by mistake?
#### Environment details

  - OS: Mac
  - .NET version: 2.1
  - Package name and version: Google.Cloud.Spanner.Data v4.3.0

#### Steps to reproduce

  1. Use the call `spannerConnection.GetSessionPoolDatabaseStatistics()`
  2. Build the code
  3. See the Obsolete warning, and attempt to find the method the warning says you should use

I'm sure I've missed something but I have looked, and the warning suggests using `GetSessionPoolStatistics` but I can't find this method anywhere.

jskeet:
This uses the fact that as of C# 10, interpolated string literals can be constants, so long as they only *use* constants - which means it's okay to use `nameof` in an attribute argument.

Fixes #9890


jskeet:
Will fix the errors tomorrow - ignore until it's green :)

jskeet:
@Rishabh-V: This second commit should fix the build error. I believe this to be a compiler bug, and I'll add a link to this PR if I file an issue about it. PTAL at the second commit, and I'll squash it into the first if you're happy.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -666,19 +666,23 @@ public async Task ShutdownSessionPoolAsync(CancellationToken cancellationToken =
...
+        // in ObsoleteAttribute.
+#pragma warning disable CS0618 // Type or member is obsolete
         /// <summary>
         /// Retrieves the database-specific statistics for the session pool associated with the connection string. The connection string must
         /// include a database name.
         /// </summary>
         /// <returns>The session pool statistics, or <c>null</c> if there is no current session pool
         /// for the database specified in the connection string.</returns>
-        [Obsolete("Use GetSessionPoolStatistics instead. Both methods return the same data, but the GetSessionPoolStatistics name better reflects the fact that sessions are pooled on aspects other than database name.")]
+        [Obsolete($"Use {nameof(GetSessionPoolSegmentStatistics)} instead. Both methods return the same data, but the ${nameof(GetSessionPoolSegmentStatistics)} name better reflects the fact that sessions are pooled on aspects other than database name.")]
```
Rishabh-V, line 678: super nit: `${nameof(GetSessionPoolSegmentStatistics)}()` as we have used `()` in the above file?

Rishabh-V (approved):
Wow. This does look like a compiler bug. :+1:

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
jskeet: I think it's clearer without, to be honest. I'd probably have remove it from the other one, but it's internal so I wasn't as bothered :)

Rishabh-V: SG

jskeet:
Compiler bug reported in https://github.com/dotnet/roslyn/issues/66938

PR #9917, opened 2023-02-20
chore: Update docfx to 2.62.1

jskeet:
This is mostly so that we remember to remove the "-f" command line argument for `docfx metadata`, which is now invalid.


amanda-tarafa (approved):
LGTM

PR #9921, opened 2023-02-21
fix(BigTable): Fixes the issue with same qualifier across families.

### LINKED ISSUE #9870: Only data for one column is returned when reading row with same qualifier in multiple families
#### Environment details

  - OS: macOS Big Sur
  - .NET version: 7
  - Package name and version: Google.Cloud.Bigtable.V2 3.3.0

#### Steps to reproduce

  1. Create a table with two column families
  2. Create a row with the same column qualifier appearing in both column families 
  3. Try to read the row - only data for one of the columns is returned 

Seems that the issue could be related to RowAsyncEnumerator; https://github.com/googleapis/google-cloud-dotnet/blob/8e661bbe7bd29a47d7cab64126a5351b0f504a2e/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowAsyncEnumerator.cs#L335. It seems to me that the cell chunk data is not handled if two cell chunks with different families but equal qualifiers appears consecutively.

Rishabh-V:
Closes #9870 

I have also added a test case to test this scenario. I will paste the link of that PR in the comment.

### DIFF
--- a/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowAsyncEnumerator.cs
+++ b/apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowAsyncEnumerator.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2017 Google Inc. All rights reserved.
+// Copyright 2017 Google Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -321,6 +321,7 @@ namespace Google.Cloud.Bigtable.V2
                 }
                 owner.Assert(chunk.ValueSize >= 0, "NewCell valueSize can't be negative");
 
+                bool familyNameChanged = false;
                 if (chunk.FamilyName != null)
                 {
                     if (chunk.FamilyName != owner._currentCell.Family?.Name)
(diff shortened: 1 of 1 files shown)


Rishabh-V:
Conformance Test PR: https://github.com/googleapis/conformance-tests/pull/74

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowAsyncEnumerator.cs
```diff
@@ -321,18 +321,20 @@ public override RowMergeState HandleChunk(RowAsyncEnumerator owner, CellChunk ch
...
                         owner._currentCell.Family = new Family { Name = chunk.FamilyName };
                         Debug.Assert(!owner._currentFamilies.ContainsKey(chunk.FamilyName));
                         owner._currentFamilies[chunk.FamilyName] = owner._currentCell.Family;
+                        familyNameChanged = true;
                     }
                     owner.Assert(chunk.Qualifier != null, "NewCell has a familyName, but no qualifier");
                 }
 
-                if (chunk.Qualifier != null && chunk.Qualifier != owner._currentCell.Column?.Qualifier)
+                if (chunk.Qualifier != null && (chunk.Qualifier != owner._currentCell.Column?.Qualifier || familyNameChanged))
```
jskeet, line 337: What should happen if chunk.Qualifier is null, but the family name has changed? (I don't know enough about the underlying mechanics to know if that's something we should handle.)

Rishabh-V: My understanding is that this if condition determines if the current cell should be associated with a new column or not. If the family has changed and qualifier is null, it should be an invalid scenario as per this test case that states that a new column family must have a qualifier.  https://github.com/Rishabh-V/conformance-tests/blob/main/bigtable/v2/readrows.json#L50

jskeet: Works for me :)

jskeet (approved):

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/RowAsyncEnumerator.cs
Rishabh-V: Thanks Jon. I will wait for Igor to have a look as well. Meanwhile, [conformance tests](https://github.com/googleapis/conformance-tests/pull/74) with new test cases are failing, so that may require some more time to be merged.

PR #9923, opened 2023-02-21
chore: Update pip requirements again

jskeet:
This time, let's keep requirements.in, so that's easier to modify directly.


.kokoro/requirements.txt
```diff
@@ -329,6 +329,10 @@ importlib-metadata==6.0.0 \
     --hash=sha256:7efb448ec9a5e313a57655d35aa54cd3e01b7e1fbcf72dce1bf06119420f5bad \
     --hash=sha256:e354bedeb60efa6affdcc8ae121b73544a7aa74156d047311948f6d711cd378d
     # via keyring
```
amanda-tarafa, line 331: But you didn't change the actual extension on the file?

amanda-tarafa (approved):
LGTM, but with a question.

.kokoro/requirements.txt
jskeet: Nope. We're still keeping requirements.txt as the thing that pip install consumes - but we generate it (and commit) using pip-compile.

PR #9926, opened 2023-02-21
feat: Changes to run DML Return.

anuragsrivstv:


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -315,10 +315,15 @@ Task<long> ISpannerTransaction.ExecuteDmlAsync(ExecuteSqlRequest request, Cancel
                 var callSettings = SpannerConnection.CreateCallSettings(settings => settings.ExecuteStreamingSqlSettings, timeoutSeconds, cancellationToken);
                 using (var reader = _session.ExecuteSqlStreamReader(request, callSettings))
                 {
-                    Value value = await reader.NextAsync(cancellationToken).ConfigureAwait(false);
-                    if (value != null)
+                    var rowsAffected = 0;
+                    while ((await reader.NextAsync(cancellationToken).ConfigureAwait(false)) != null)
```
jskeet, line 319: I'd assume the stats are still available - I don't think we should bother reading through all the results if we haven't been asked to do so.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -436,8 +436,11 @@ private ReadOrQueryRequest GetReadOrQueryRequest()
                     case SpannerCommandType.Read:
                         request = ReadOrQueryRequest.FromReadRequest(GetReadRequest());
                         break;
+                    case SpannerCommandType.Dml:
```
jskeet, line 439: This has the same code as the SpannerCommandType.Select case, so we can just add this case to that one, unless there's another change coming later.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
```diff
@@ -133,13 +133,14 @@ async Task<int> Impl(SpannerTransaction transaction)
             }
         }
 
-        public Task<ReliableStreamReader> ExecuteReadOrQueryAsync(ReadOrQueryRequest request, CancellationToken cancellationToken, int timeoutSeconds /* ignored */)
+        public Task<ReliableStreamReader> ExecuteReadOrQueryAsync(ReadOrQueryRequest request, CancellationToken cancellationToken, int timeoutSeconds /* ignored */, bool isDmlReturn = false)
```
jskeet, line 136: In my own equivalent branch, I ended up putting the command type (as SpannerCommandType) in the EphemeralTransaction constructor. That meant I didn't need to modify any of the other transaction types.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
```diff
@@ -133,13 +133,14 @@ async Task<int> Impl(SpannerTransaction transaction)
...
         {
             return ExecuteHelper.WithErrorTranslationAndProfiling(Impl, "EphemeralTransaction.ExecuteReadOrQuery", _connection.Logger);
 
             async Task<ReliableStreamReader> Impl()
             {
-                PooledSession session = await _connection.AcquireSessionAsync(null, cancellationToken).ConfigureAwait(false);
+                var transactionOptions = isDmlReturn ? SpannerConnection.ReadWriteTransactionOptions : null;
```
jskeet, line 142: This means it's not automatically retried - whereas ExecuteNonQuery would use RunWithRetriableTransactionAsync. I suspect we'd want consistency there.

jskeet (commented):
Hmm... I'm not sure why these changes work but my mostly-equivalent ones didn't. It's possibly something to do with the RunWithRetriableTransaction that I was using (and think we should still use if possible).

anuragsrivstv:
> 



> Hmm... I'm not sure why these changes work but my mostly-equivalent ones didn't. It's possibly something to do with the RunWithRetriableTransaction that I was using (and think we should still use if possible).

With these changes also I'm not getting exact outcome. With `ExecuteReaderAsync` I'm able to run DML `Return` clause and was able to read affected/updated rows data one by one. But then when I ran `Select` query, the table was not updated. I _think_ , it has something to do with transaction not getting committed or something.

jskeet:
> But then when I ran Select query, the table was not updated. I think , it has something to do with transaction not getting committed or something.

That would make sense, as your approach at the moment doesn't commit the transaction - whereas my attempt with the RunInRetryableTransaction *would* commit it. (But apparently wasn't working for another reason.)

I can see a difficulty here: it's not clear to me whether we actually *can* commit the transaction and then return the SpannerDataReader. Is it invalidated by committing? I'm not sure, and we probably need to ask Knut. Basically we're outside the bounds of "normal" ordering at this point. (While we *could* load up all the data in memory, commit the transaction and then return an in-memory representation, that feels really inappropriate to me.)

One possibility would be to only allow ExecuteQuery with DML in explicit read/write transactions, at least for an initial feature launch.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -444,6 +444,13 @@ public override bool NextResult()
                             {
                                 throw new SpannerException(ErrorCode.DeadlineExceeded, "Read operation timed out");
                             }
+                            // This excpetion may occur if Spanner had to split returned data into multiple PartialResultSets.
+                            // This should not happen right now as DML statement can never execute more than 40,000 mutations but could increase in the future.
+                            catch (Exception ex) when (ex.Message.Contains("Read/query was cancelled due to Commit()/Rollback() on the enclosing transaction"))
```
anuragsrivstv, line 449: Hi @jskeet  and @olavloite , Please let me know that does this seems correct way to handle the case if the result was so big that Spanner had to split it into multiple PartialResultSets. If not how and where do you suggest us to do this error handling. Also what I understood was that right now this error is not reproducible , am I correct in saying so ?

jskeet: We should *definitely* only catch the right exception type - which I'd expect to be `RpcException`.
Ideally Spanner will provide an `ErrorInfo` that we can use to check the condition rather than checking text.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,42 +535,54 @@ public void PartitionedUpdate_Medium()
...
-                while (await reader.ReadAsync())
-                {
-                    actualValues.Add(reader.GetFieldValue<int>(reader.GetOrdinal("Value")));
-                }
-            });
-            var expectedValues = new List<int> { 2, 5 };
-            Assert.Equal(expectedValues, actualValues);
+                actualReturnedValues.Add(reader.GetFieldValue<int>(reader.GetOrdinal("Value")));
+            }
+            // TO Assert that DML Retrun command has returned expected values only.
```
jskeet, line 556:
```suggestion
            // Assert that the DML Return command has returned expected values only.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,42 +535,54 @@ public void PartitionedUpdate_Medium()
...
+            var expectedReturnedValues = new List<int> { 2, 5 };
+            Assert.Equal(expectedReturnedValues, actualReturnedValues);
+
+            // To Assert that DML Return command has actually updated the table.
```
jskeet, line 560:
```suggestion
            // Assert that the DML Return command has actually updated the table.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,42 +535,54 @@ public void PartitionedUpdate_Medium()
...
+            var actualDbValues = _fixture.FetchValues(key);
```
jskeet, line 561: Let's extract this code to another method, given that it's used for multiple tests.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -608,43 +622,50 @@ public async Task DMLReturn_ExecuteNonQueryAsync()
...
+                { 3, 3 }, // Not updated
+                { 4, 5 }  // Updated
+            };
+
+            Assert.False(readOnce);
+            Assert.Equal(expected, actual);
         }
 
-        [Fact(Skip = "This test is skipped until DML Returning changes are done.")]
+        // Without DML's return clause and ReadAsync call also ExecuteReaderAsync should update database table.
```
jskeet, line 656:
```suggestion
        // ExecuteReaderAsync still updates the database, even if we don't read from the returned reader and don't include "Then Return"
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -608,43 +622,50 @@ public async Task DMLReturn_ExecuteNonQueryAsync()
...
+        [Fact]
```
jskeet, line 657: We do still need another test for what happens if we call ExecuteReader and *do* read from the returned reader without any "With Return" part.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
```diff
@@ -52,6 +52,28 @@ async Task<long> Impl(SpannerTransaction transaction)
             }
         }
 
+        /// <summary>
+        /// Executes DML Return statement and returns affected requested rows values via ReliableStreamReader.
+        /// If the Return clause is not present with the DML command then the reader will not return any data,
+        /// but DML command will be executed.
```
jskeet, line 58:
```suggestion
        /// Executes a DML statement and returns affected rows values via ReliableStreamReader.
        /// If the Return clause is not present with the DML command then the reader will not return any data,
        /// but DML command will be executed.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ISpannerTransaction.cs
```diff
@@ -28,6 +28,9 @@ internal interface ISpannerTransaction
         // do if they need to return an int result type.
         Task<long> ExecuteDmlAsync(ExecuteSqlRequest request, CancellationToken cancellationToken, int timeoutSeconds);
 
+        // Note: this returns Task<ReliableStreamReader> to reflect requested affetced rows values from Spanner faithfully.
```
jskeet, line 31:
```suggestion
        // Note: this returns Task<ReliableStreamReader> to allow affected rows to be returned.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/ISpannerTransaction.cs
```diff
@@ -28,6 +28,9 @@ internal interface ISpannerTransaction
...
+        Task<ReliableStreamReader> ExecuteDmlReturnAsync(ExecuteSqlRequest request, CancellationToken cancellationToken, int timeoutSeconds);
```
jskeet, line 32: I'd rename this to `ExecuteDmlReader` or similar - because it will only ever be used with DML, it will always return a reader - but we don't actually know whether there's a "Then Return" clause.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -106,9 +106,12 @@ internal async Task<SpannerDataReader> ExecuteReaderAsync(CommandBehavior behavi
                 ValidateConnectionAndCommandTextBuilder();
                 ValidateCommandBehavior(behavior);
 
+                if (CommandTextBuilder.SpannerCommandType == SpannerCommandType.Dml)
```
jskeet, line 109: Please use braces for all `if` statement bodies, even when there's only a single statement.
(I'm surprised our editorconfig didn't pick this up...)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -106,9 +106,12 @@ internal async Task<SpannerDataReader> ExecuteReaderAsync(CommandBehavior behavi
...
+                    return  await ExecuteDmlReturnAsync(cancellationToken).ConfigureAwait(false);
```
jskeet, line 110: Nit: remove double space after return.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -223,6 +226,16 @@ private async Task<int> ExecuteDmlAsync(CancellationToken cancellationToken)
                 return checked((int) count);
             }
 
+            private async Task<SpannerDataReader> ExecuteDmlReturnAsync(CancellationToken cancellationToken)
```
jskeet, line 229: I'd rename this to `ExecuteDmlReaderAsync` for the same reasons as on ISpannerTransaction.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
jskeet: We're ignoring `behavior` and `singleUseReadSettings` here. Is that appropriate?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
jskeet: When we've heard back about this, we should have a test for it as well.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
jskeet: (If they're not supported, we should validate that they're not set.)

jskeet (commented):
Looks like we're getting close - a few things to change.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
jskeet: Yes, I wasn't suggesting removing this test - just adding another one.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
jskeet: > So may be we can log a warning for it if for DML behaviour value is other than default, rather than validating null value, what do you think ?

No, warnings are very unlikely to be seen (the only logger we have easily available is the Spanner semi-internal one.

If the user has specified a behavior other than the default, and we're ignoring that, we should throw an exception to indicate that that behavior isn't supported. (I don't think it can actually be null.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
olavloite: (Sorry, I have not found the time to give this a complete review)

It is possible to reproduce it, but not with a DML statement. What you can do is the following:

1. Create a table with a large number of rows/columns (basically enough data that Spanner will return multiple PartialResultSets)
2. Start a read/write transaction and execute `select * from my_table` using that transaction. Make sure to call `DataReader.Read` once while the transaction is still active.
3. Call `Transaction.Commit()`.
4. Try to fetch the rest of the data from the reader. This will fail with the above-mentioned error.

I don't think Spanner includes any unique `ErrorInfo` with this error, so I'm afraid that we have to rely on text matching.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
jskeet: Great thanks.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
anuragsrivstv: Sure.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDataReader.cs
```diff
@@ -446,7 +448,11 @@ public override bool NextResult()
                             }
                             // This excpetion may occur if Spanner had to split returned data into multiple PartialResultSets.
                             // This should not happen right now as DML statement can never execute more than 40,000 mutations but could increase in the future.
-                            catch (Exception ex) when (ex.Message.Contains("Read/query was cancelled due to Commit()/Rollback() on the enclosing transaction"))
+                            // TODO: Need to confirm behaviour of this catch block whenever we are able to produce the excpetion of multiple PartialResultSets.
+                            catch (Exception ex) when (
+                            ex is RpcException rpcException &&
+                            rpcException.GetErrorInfo() != null &&
+                            rpcException.GetErrorInfo().Reason.Contains("Read/query was cancelled due to Commit()/Rollback() on the enclosing transaction."))
```
anuragsrivstv, line 455: Hi @jskeet , does this seems fine ? If not, please let me know what do you have in you mind.

anuragsrivstv: I think based on latest Knut's comment, we may need to change this.

jskeet: Yes, the Reason would never be something like this - it would be an identifier, basically. I think it's worth consulting the Spanner term internally though - we should really try to avoid using text matching. It would be much better for the Spanner team to *start* using ErrorInfo if they're really not already.

Additionally, you shouldn't catch Exception and *then* check the type - just `catch (RpcException ex)` to start with.

anuragsrivstv: Sure. As we'll not get this particular exception right now, removing the catch block for now.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// Copyright 2017 Google Inc. All Rights Reserved.
```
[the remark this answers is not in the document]
jskeet: It's a one-time removal of the byte-order mark that VS used to include, but which is generally unpleasant. We've learned to just ignore changes like this :) (We could potentially fix all files in a single change, but I'm not sure it's worth it.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +535,114 @@ public void PartitionedUpdate_Medium()
             }
         }
 
-        [Fact(Skip = "This test is skipped until DML Returning changes are done.")]
+        // With DML's return clause ExecuteReaderAsync should return a reader containing affected rows data.
```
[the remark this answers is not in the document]
amanda-tarafa: And ideally with the retryable transaction as well. In particular, if there are any read errors on the DML returned data, the errors returned by Spanner shouldn't be retryable, i.e. we should never retry the DML after read errors.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
anuragsrivstv: Going through the existing code I believe that only `CommandBehavior.CloseConnection` is used in `ExecuteReader`  and that to decide that the connection is needed to be closed while disposing `SpannerDataReader` or not. This does not seems relevant with the current DML Implementations. Please correct me if I'm wrong.

Also in ExecuteReader here https://github.com/googleapis/google-cloud-dotnet/blob/799d9d1e2aff59c73c070fd8ceaecc89ade4d7ea/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs#L480 we already have a validation that  CommandBehaviour SchemaOnly is not supported by Cloud Spanner .

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -104,11 +104,27 @@ internal async Task<T> ExecuteScalarAsync<T>(CancellationToken cancellationToken
...
+                    {
+                        throw new NotSupportedException($"CommandBehaviour {behavior} is not supported by with DML command.");
+                    }
+
+                    return await ExecuteDmlReaderAsync(cancellationToken).ConfigureAwait(false);
+                }
+
                 ValidateCommandBehavior(behavior);
 
                 if (CommandTextBuilder.SpannerCommandType != SpannerCommandType.Select && CommandTextBuilder.SpannerCommandType != SpannerCommandType.Read)
```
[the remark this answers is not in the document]
anuragsrivstv: Sure.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
jskeet: Someone can call `ExecuteReader` and pass in `CommandBehavior.CloseConnection`, expecting (reasonably) that the connection should be closed. If we don't want to support it, then we should throw an exception if a client *tries* to do it - just ignoring their requested behavior seems wrong.

anuragsrivstv: Yes, we are throwing exception for it for DML at https://github.com/googleapis/google-cloud-dotnet/blob/579173ce98c110b54e354154be8e8e2311d6abc2/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs#L124

jskeet: Great - that wasn't there last time I'd reviewed :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/EphemeralTransaction.cs
```diff
@@ -52,6 +52,28 @@ async Task<long> Impl(SpannerTransaction transaction)
             }
         }
 
+        /// <summary>
+        /// Executes a DML statement and returns affected rows values via ReliableStreamReader.
+        /// If the Return clause is not present with the DML command then the reader will not return any data,
```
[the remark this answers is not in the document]
anuragsrivstv: Done.

anuragsrivstv (commented):
Hi All, Please let me know if the changes are fine to merge in the feature branch, or if there is any thing else you would like me to address. Thanks.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -103,14 +103,31 @@ internal async Task<T> ExecuteScalarAsync<T>(CancellationToken cancellationToken
...
             internal async Task<SpannerDataReader> ExecuteReaderAsync(CommandBehavior behavior, TimestampBound singleUseReadSettings, CancellationToken cancellationToken)
             {
+                if (CommandTextBuilder.SpannerCommandType != SpannerCommandType.Select
+                    && CommandTextBuilder.SpannerCommandType != SpannerCommandType.Read
+                    && CommandTextBuilder.SpannerCommandType != SpannerCommandType.Dml)
+                {
+                    throw new InvalidOperationException("ExecuteReader functionality is only available for queries, reads and DML commands.");
+                }
+
                 ValidateConnectionAndCommandTextBuilder();
```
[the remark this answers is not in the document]
amanda-tarafa: I think I would change the order of all of these validations for clarity and consistency with the other methods.

First: ValidateConnectionAndCommandTextBuilder()
Second: Check if it's DML, and do the DML specific validations and return the DML query.
Third: Call ValidateCommandBehaviour (this is the order in which is done in other methods)
Fourth: do the rest, including checking the command type, you don't have to check DML again, but you still need to include DML on the "ExecuteReader functionality is only available for queries, reads and DML commands." message.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -103,14 +103,31 @@ internal async Task<T> ExecuteScalarAsync<T>(CancellationToken cancellationToken
...
-                if (CommandTextBuilder.SpannerCommandType != SpannerCommandType.Select && CommandTextBuilder.SpannerCommandType != SpannerCommandType.Read)
+                if (CommandTextBuilder.SpannerCommandType == SpannerCommandType.Dml)
                 {
-                    throw new InvalidOperationException("ExecuteReader functionality is only available for queries and reads.");
+                    if (singleUseReadSettings != null)
+                    {
+                        throw new NotSupportedException("singleUseReadSettings cannot be used with DML command.");
+                    }
+
+                    if (behavior != CommandBehavior.Default)
```
amanda-tarafa, line 122: I don't see why we cannot support CommandBehaviour.CloseConnection here. The connection will be closed when the reader is closed, and that seems plausible here. What am I missing?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -223,6 +240,16 @@ private async Task<int> ExecuteDmlAsync(CancellationToken cancellationToken)
...
             }
 
+            private async Task<SpannerDataReader> ExecuteDmlReaderAsync(CancellationToken cancellationToken)
+            {
+                await Connection.EnsureIsOpenAsync(cancellationToken).ConfigureAwait(false);
+                var transaction = Transaction ?? Connection.AmbientTransaction ?? new EphemeralTransaction(Connection, Priority);
+                ExecuteSqlRequest request = GetExecuteSqlRequest();
+                var resultSet = await transaction.ExecuteDmlReaderAsync(request, cancellationToken, CommandTimeout).ConfigureAwait(false);
+                var enableGetSchemaTable = Connection.Builder.EnableGetSchemaTable;
+                return new SpannerDataReader(Connection.Logger, resultSet, Transaction?.ReadTimestamp, null, ConversionOptions, enableGetSchemaTable, CommandTimeout);
```
amanda-tarafa, line 250: We should support CommandBehaviour.CloseConnection here. I don't see whe we couldn't.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +535,218 @@ public void PartitionedUpdate_Medium()
...
-            RetryHelpers.ExecuteWithRetry(async () =>
+            string dml = $"UPDATE {_fixture.TableName} SET Value = OriginalValue + 1 WHERE UpdateMe AND Key=@key Then Return Value";
+            var command = connection.CreateDmlCommand(dml);
+            command.Parameters.Add("key", SpannerDbType.String, key);
+
+            List<int> actualReturnedValues = executeWithExplicitTransaction ?
+                await ExecuteWithExplicitTransaction(() => DmlAsyncWork(), connection, command) :
+                await ExecuteWithEphemeralTransaction(() => DmlAsyncWork());
+
+            async Task<List<int>> DmlAsyncWork()
```
amanda-tarafa, line 556: It's clearer for me if you move this to the end of the method.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +535,218 @@ public void PartitionedUpdate_Medium()
...
+            Assert.Equal(expectedReturnedValues, actualReturnedValues);
+
+            // Assert that the DML Return command has actually updated the table.
+            AssertActualAndUpdatedDbValues(key);
+        }
+
+        private async Task<T> ExecuteWithEphemeralTransaction<T>(Func<Task<T>> dmlAsyncWork)
+        {
+            return await dmlAsyncWork().ConfigureAwait(false);
+        }
```
amanda-tarafa, line 726:
```suggestion
        private Task<T> ExecuteWithEphemeralTransaction<T>(Func<Task<T>> dmlAsyncWork) => dmlAsyncWork();
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +535,218 @@ public void PartitionedUpdate_Medium()
...
+                await ExecuteWithEphemeralTransaction(() => DmlAsyncWork());
+
+            async Task<List<int>> DmlAsyncWork()
             {
-                string dml = $"UPDATE {_fixture.TableName} SET Value = OriginalValue + 1 WHERE UpdateMe AND Key=@key Then Return Value";
-                var command = connection.CreateDmlCommand(dml);
                 var reader = await command.ExecuteReaderAsync();
-            });
+                return null;
+            }
```
amanda-tarafa, line 598: async Task<bool> DmlAsyncWork()
{
    await command.ExecuteReaderAsync();
    return true;
}

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +535,218 @@ public void PartitionedUpdate_Medium()
...
-                { 0, 0 }, // Not updated
-                { 1, 2 }, // Updated
-                { 2, 2 }, // Not updated
-                { 3, 3 }, // Not updated
-                { 4, 5 }  // Updated
-            };
-            Assert.Equal(expected, actual);
+                var affectedRows = await command.ExecuteNonQueryAsync();
+                return affectedRows;
+            }
```
amanda-tarafa, line 624: async Task<int> DmlAsyncWork() => await command.ExecuteNonQueryAsync();

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +535,218 @@ public void PartitionedUpdate_Medium()
...
         }
 
-        [Fact(Skip = "This test is skipped until DML Returning changes are done.")]
-        public async Task DMLReturn_ExecuteReader_NoRead()
+        // With DML's return clause ExecuteReaderAsync should update table data even if Read is not invoked.
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        [Trait(Constants.SupportedOnEmulator, Constants.No)]
+        public async Task DMLReturn_ExecuteReader_NoRead(bool executeWithExplicitTransaction)
```
amanda-tarafa, line 580:
```suggestion
        [Trait(Constants.SupportedOnEmulator, Constants.No)]
        public async Task DMLReturn_ExecuteReader_NoRead([Combinatorial] bool executeWithExplicitTransaction)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +535,218 @@ public void PartitionedUpdate_Medium()
...
-                    //actualValues.Add(reader.GetFieldValue<int>(reader.GetOrdinal("Value")));
-                }
-            });
-            var expectedValues = new List<int> { 2, 5 };
-            Assert.Equal(expectedValues, actualValues);
+                var readOnce = await reader.ReadAsync();
+                return readOnce;
+            }
+
+            Assert.False(read);
```
amanda-tarafa, line 656: Do we get the affected rows anywhere? If yes, check that as well.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +535,218 @@ public void PartitionedUpdate_Medium()
...
-            RetryHelpers.ExecuteWithRetry(async () =>
+            string dml = $"UPDATE {_fixture.TableName} SET Value = OriginalValue + 1 WHERE UpdateMe AND Key=@key";
+            var command = connection.CreateDmlCommand(dml);
+            command.Parameters.Add("key", SpannerDbType.String, key);
+
+            _ = executeWithExplicitTransaction ?
+                await ExecuteWithExplicitTransaction(() => DmlAsyncWork(), connection, command) :
+                await ExecuteWithEphemeralTransaction(() => DmlAsyncWork());
+
+            async Task<List<int>> DmlAsyncWork()
```
amanda-tarafa, line 677: Do we get the affected rows anywhere? If yes, check that as well please.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +535,218 @@ public void PartitionedUpdate_Medium()
...
+                var reader = await command.ExecuteReaderAsync();
+                return null;
+            }
+
+            AssertActualAndUpdatedDbValues(key);
+        }
+
+        [Fact]
+        [Trait(Constants.SupportedOnEmulator, Constants.No)]
+        public async Task DmlReturnWithRetryableTransaction()
```
amanda-tarafa, line 688: I think we can run all above tests with retryable transaction as well. If you change the "DoAsyncWork" methods to recieve a transaction, that will be null with ephemeral transactions, it should be simple I think.

amanda-tarafa (changes requested):
Most comments are on tests.

And a question on production code. I don't see why we cannot support CommandBehaviour.CloseConnection for DML reads. The connection will be closed once the reader is closed, and not after the DML is executed. Am I missing something there?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
anuragsrivstv: Done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -106,12 +106,20 @@ internal async Task<SpannerDataReader> ExecuteReaderAsync(CommandBehavior behavi
                 ValidateConnectionAndCommandTextBuilder();
                 ValidateCommandBehavior(behavior);
 
-                if (CommandTextBuilder.SpannerCommandType != SpannerCommandType.Select && CommandTextBuilder.SpannerCommandType != SpannerCommandType.Read)
+                await Connection.EnsureIsOpenAsync(cancellationToken).ConfigureAwait(false);
+                var enableGetSchemaTable = Connection.Builder.EnableGetSchemaTable;
+                // When the data reader is closed, we may need to dispose of the connection.
+                IDisposable resourceToClose = (behavior & CommandBehavior.CloseConnection) == CommandBehavior.CloseConnection ? Connection : null;
```
amanda-tarafa, line 112: You don't need to do any of this before checking for DML.
The ExecuteDmlReadyAsync should receive a CommandBehaviour and check there.

I would move the if whith the DML check right after ValidateConnectionAndCommandTextBuilder(), and then leave the rest of the method as it was before. You can then do the DML Reader specific checks in `ExecuteDmlReaderAsync(CommandBehaviour, CancellationToken)`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -223,6 +226,14 @@ private async Task<int> ExecuteDmlAsync(CancellationToken cancellationToken)
                 return checked((int) count);
             }
 
+            private async Task<SpannerDataReader> ExecuteDmlReaderAsync(bool enableGetSchemaTable, IDisposable resourceToClose, CancellationToken cancellationToken)
```
amanda-tarafa, line 229:
```suggestion
            private async Task<SpannerDataReader> ExecuteDmlReaderAsync(CommandBehaviour behaviour, CancellationToken cancellationToken)
```
And calculate the others here, it's fine if that code is duplicated with ExceuteReaderAsync, its clearer to have the ExecuteDmlReaderAsync implementation completely separate from the ExecuteReaderAsync one. If they were sharing a lot more code that'd be different.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,192 @@ public void PartitionedUpdate_Medium()
...
+            List<int> actualReturnedValues = await Execute(DmlAsyncWork, connection, command, transactionType);
+
+            // Assert that the DML Return command has returned expected values only.
+            var expectedReturnedValues = new List<int> { 2, 5 };
+            Assert.Equal(expectedReturnedValues, actualReturnedValues);
+
+            // Assert that the DML Return command has actually updated the table.
+            AssertActualAndUpdatedDbValues(key);
+
+            async Task<List<int>> DmlAsyncWork(SpannerTransaction retryableTransaction = null)
```
amanda-tarafa, line 566: This might or might not be a retriable transaction.
```suggestion
            async Task<List<int>> DmlAsyncWork(SpannerTransaction transaction = null)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,192 @@ public void PartitionedUpdate_Medium()
...
             {
-                string dml = $"UPDATE {_fixture.TableName} SET Value = OriginalValue + 1 WHERE UpdateMe AND Key=@key Then Return Value";
-                var command = connection.CreateDmlCommand(dml);
+                var returnedValues = new List<int>();
+                if (retryableTransaction != null)
```
amanda-tarafa, line 569: You can skip this check, if you set it to null, is the same as not having set it at all.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,192 @@ public void PartitionedUpdate_Medium()
...
+        }
 
-            var actual = _fixture.FetchValues(key);
-            var expected = new Dictionary<int, int>
+        private async Task<T> Execute<T>(Func<SpannerTransaction, Task<T>> dmlAsyncWork, SpannerConnection connection, SpannerCommand command, TransactionType transactionType)
+        {
+            switch (transactionType)
+            {
+                case TransactionType.EphemeralTransaction:
+                    return await ExecuteWithEphemeralTransaction(() => dmlAsyncWork(null));
```
amanda-tarafa, line 697:
```suggestion
                    return await dmlAsyncWork(null);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,192 @@ public void PartitionedUpdate_Medium()
...
+                case TransactionType.ExplicitTransaction:
+                    return await ExecuteWithExplicitTransaction(() => dmlAsyncWork(null), connection, command);
```
amanda-tarafa, line 699:
```suggestion
                    return await ExecuteWithExplicitTransaction(dmlAsyncWork, connection);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,192 @@ public void PartitionedUpdate_Medium()
...
+                case TransactionType.RetryableTransaction:
+                    return await connection.RunWithRetriableTransactionAsync(transaction => dmlAsyncWork(transaction));
+                default:
+                    throw new ArgumentException($"Invalid argument '{transactionType}'.");
+            }
+        }
+
+        private async Task<T> ExecuteWithEphemeralTransaction<T>(Func<Task<T>> dmlAsyncWork) => await dmlAsyncWork().ConfigureAwait(false);
```
amanda-tarafa, line 707:
```suggestion
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,192 @@ public void PartitionedUpdate_Medium()
...
+
+        private async Task<T> ExecuteWithExplicitTransaction<T>(Func<Task<T>> dmlAsyncWork, SpannerConnection connection, SpannerCommand command)
```
amanda-tarafa, line 709:
```suggestion
        private async Task<T> ExecuteWithExplicitTransaction<T>(Func<SpannerTransaction, Task<T>> dmlAsyncWork, SpannerConnection connection)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,192 @@ public void PartitionedUpdate_Medium()
...
+        {
+            using var transaction = connection.BeginTransaction();
+            command.Transaction = transaction;
```
amanda-tarafa, line 712:
```suggestion
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,192 @@ public void PartitionedUpdate_Medium()
...
+            var result = await dmlAsyncWork().ConfigureAwait(false);
```
amanda-tarafa, line 713:
```suggestion
            var result = await dmlAsyncWork(transaction).ConfigureAwait(false);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,192 @@ public void PartitionedUpdate_Medium()
...
+        private async Task<T> Execute<T>(Func<SpannerTransaction, Task<T>> dmlAsyncWork, SpannerConnection connection, SpannerCommand command, TransactionType transactionType)
+        {
+            switch (transactionType)
+            {
+                case TransactionType.EphemeralTransaction:
+                    return await ExecuteWithEphemeralTransaction(() => dmlAsyncWork(null));
+                case TransactionType.ExplicitTransaction:
+                    return await ExecuteWithExplicitTransaction(() => dmlAsyncWork(null), connection, command);
+                case TransactionType.RetryableTransaction:
+                    return await connection.RunWithRetriableTransactionAsync(transaction => dmlAsyncWork(transaction));
```
amanda-tarafa, line 701:
```suggestion
                    return await connection.RunWithRetriableTransactionAsync(dmlAsyncWork);
```

amanda-tarafa (changes requested):
A few more comments for cleaning up tests, and on the ordering of the ExecuteReaderAsync method, but looking good. Thanks.

amanda-tarafa (approved):
LGTM, thanks for these changes.

Do wait for @jskeet 's review before merging. And also, once he has LGTMed, rebase all changed in a single commit.

Thanks

jskeet (approved):
One tiny nit, but otherwise (and even that can be ignored) I think we're good to merge.
(Please rebase to one commit first of course.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
```diff
@@ -535,127 +542,174 @@ public void PartitionedUpdate_Medium()
...
-            var actual = _fixture.FetchValues(key);
-            var expected = new Dictionary<int, int>
+        private async Task<T> Execute<T>(Func<SpannerTransaction, Task<T>> dmlAsyncWork, SpannerConnection connection, TransactionType transactionType)
+        {
+            switch (transactionType)
+            {
+                case TransactionType.EphemeralTransaction:
+                    return await dmlAsyncWork(null);
+                case TransactionType.ExplicitTransaction:
+                    return await ExecuteWithExplicitTransaction(dmlAsyncWork, connection);
```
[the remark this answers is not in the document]
anuragsrivstv: Done.

anuragsrivstv (commented):
Hey @jskeet , I've fixed the comment. Please have a look. If you are happy, I'll merge it. Thanks.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/DmlTests.cs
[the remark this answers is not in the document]
jskeet: I don't see any change - you've still got the local ExecuteWithExplicitTransaction method here, which I was suggesting inlining into this case.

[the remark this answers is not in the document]
anuragsrivstv: Sure, I misunderstood the comment. Fixed now. Please have a look.

jskeet (approved):

anuragsrivstv:
Merging the changes.

Issue #9939, opened 2023-02-22
Docs build is broken in multiple ways


jskeet:
- We need to either stop creating metadata for dependencies, or build them first
- docfx 2.61.0 and later have broken snippets in weird ways

jskeet:
While we're at it, we should investigate why we're still getting XML escaping in docs, e.g. https://cloud.google.com/dotnet/docs/reference/Google.Cloud.Dialogflow.V2/latest/Google.Cloud.Dialogflow.V2.AgentsClient#Google_Cloud_Dialogflow_V2_AgentsClient_GetAgent_System_String_Google_Api_Gax_Grpc_CallSettings_ - search for "Required. The project that the agent to fetch is associated with. Format"

jskeet:
Note to self: Google.Cloud.Tools.UnescapeDocfxMetadataXml is meant to unescape the metadata.

PR #9943, opened 2023-02-22
chore: Only generate metadata for the API we're creating docs for

jskeet:
This only changes the customer-facing result for non-Cloud APIs (which still have docs hosted on googleapis.dev), and only when they have other dependencies within this repo. We can live with that at the moment, and it avoids rebuilding things for no good reason.

Towards #9939


amanda-tarafa (approved):
LGTM

PR #9971, opened 2023-03-01
docs: Add help documentation for using API keys with GAPIC libraries

jskeet:
The change to the call settings override is just to make the test pass... it can easily take 5 seconds to create a topic...


amanda-tarafa (approved):
LGTM

PR #9980, opened 2023-03-02
feat: Draft add support for OR Queries

hemanshv:
Its a draft PR to showcase the implementation details for OR Queries. This only demonstrates **few filters**  like Equal, LessThan and GreatherThan the main PR will have all the remaining filters. It an aid to decide up on the best approach.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Filter.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Filter.cs
@@ -1,0 +1,240 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+
+using Google.Cloud.Firestore.V1;
+using System;
+using System.Collections.Generic;
+using FieldOp = Google.Cloud.Firestore.V1.StructuredQuery.Types.FieldFilter.Types.Operator;
+using UnaryOp = Google.Cloud.Firestore.V1.StructuredQuery.Types.UnaryFilter.Types.Operator;
+using ProtoFilter = Google.Cloud.Firestore.V1.StructuredQuery.Types.Filter;
+using static Google.Cloud.Firestore.V1.StructuredQuery.Types.CompositeFilter.Types;
+using System.Linq;
+using Google.Protobuf.WellKnownTypes;
+
+namespace Google.Cloud.Firestore;
+
+
+/// <summary>
+/// 
+/// </summary>
+public abstract class Filter
+{
+    private Filter() { }
+
+    internal abstract ProtoFilter ToProto(SerializationContext context);
+
+    /// <summary>
+    /// 
+    /// </summary>
+    internal class UnaryFilter : Filter, IEquatable<UnaryFilter>
+    {
+        public FieldPath Field { get; }
+
+        // The integer value of either the UnaryOp or FieldOp, depending on whether _value is null.
+        public int Op { get; }
+
+        // The value for a field (binary) operator, or null for a unary operator
+        public Object Value { get; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="fieldPath"></param>
+        /// <param name="op"></param>
+        /// <param name="value"></param>
+        public UnaryFilter(FieldPath fieldPath, FieldOp op, Object value)
+        {
+            Field = fieldPath;
+            var unaryOperator = GetUnaryOperator(value, op);
+            if (unaryOperator != UnaryOp.Unspecified)
+            {
+                Op = (int) unaryOperator;
+                Value = null;
+            }
+            else
+            {
+                Op = (int) op;
+                Value = value;
+            }
+        }
+
+        internal override ProtoFilter ToProto(SerializationContext context)
+        {
+            
+            return Value == null
+                ? new ProtoFilter { UnaryFilter = new StructuredQuery.Types.UnaryFilter { Field = Field.ToFieldReference(), Op = (UnaryOp) Op } }
+                : new ProtoFilter { FieldFilter = new StructuredQuery.Types.FieldFilter { Field = Field.ToFieldReference(), Op = (FieldOp) Op, Value = ValueSerializer.Serialize(context, Value) } };
+        }
+
+        private static UnaryOp GetUnaryOperator(object value, FieldOp op)
+        {
+            switch (value)
+            {
+                case null:
+                    return op switch
+                    {
+                        FieldOp.Equal => UnaryOp.IsNull,
+                        FieldOp.NotEqual => UnaryOp.IsNotNull,
+                        _ => throw new ArgumentException("Null values can only be used with the Equal/NotEqual operators", nameof(value))
+                    };
+                case double d when double.IsNaN(d):
+                case float f when float.IsNaN(f):
+                    return op switch
+                    {
+                        FieldOp.Equal => UnaryOp.IsNan,
+                        FieldOp.NotEqual => UnaryOp.IsNotNan,
+                        _ => throw new ArgumentException("Not-a-number values can only be used with the Equal/NotEqual operators", nameof(value))
+                    };
+                default:
+                    return UnaryOp.Unspecified;
+            }
+        }
+
+        public bool Equals(UnaryFilter other) => throw new NotImplementedException();
+    }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    internal class CompositeFilter : Filter, IEquatable<CompositeFilter>
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        public IEnumerable<Filter> Filters { get; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        public Operator Operator { get; }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="filters"></param>
+        /// <param name="op"></param>
+        public CompositeFilter(IEnumerable<Filter> filters, Operator op)
+        {
+            Filters = new List<Filter>(filters);
+            Operator = op;
+        }
+
+        public CompositeFilter WithAdditionalFilter(Filter filter)
+        {
+            var filters = new List<Filter>(Filters);
+            filters.Add(filter);
+            return new CompositeFilter(filters, Operator);
+        }
+
+        internal override ProtoFilter ToProto(SerializationContext context) =>
+            new ProtoFilter { CompositeFilter = new StructuredQuery.Types.CompositeFilter { Op = Operator, Filters = { Filters.Select(f => f.ToProto(context)) } } };
+
+        
+
+        public bool Equals(CompositeFilter other) => throw new NotImplementedException();
+    }
+
+    internal static bool IsOrderingFilter(UnaryFilter filter) =>
+        filter.Value is object &&
+        filter.Op == (int) FieldOp.GreaterThan ||
+        filter.Op == (int) FieldOp.GreaterThanOrEqual ||
+        filter.Op == (int) FieldOp.LessThan ||
+        filter.Op == (int) FieldOp.LessThanOrEqual;
+
+    internal static UnaryFilter GetOrderingFilter(CompositeFilter compositeFilter)
+    {
+        foreach (var filter in compositeFilter.Filters)
+        {
+            if (filter.GetType() == typeof(Filter.UnaryFilter))
+            {
+                if (IsOrderingFilter((UnaryFilter) filter))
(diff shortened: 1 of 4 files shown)


hemanshv:
Hi @jskeet,
It is the Draft PR for Query OR feature. Please have a look and provide your valuable feedback.

Thanks,

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Filter.cs
```diff
@@ -0,0 +1,144 @@
...
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    internal class UnaryFilter : Filter
+    {
+        public FieldPath Field { get; }
+
+        // The integer value of either the UnaryOp or FieldOp, depending on whether _value is null.
+        public FieldOp _op { get; }
```
jskeet, line 37: I'd expect this to either be a private field, or a property with a conventional name. (This sort of thing can be settled later though.)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -509,6 +511,50 @@ object ConvertValueToDocumentReferencesForInQuery()
...
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public Query Where(Filter filter)
+        {
+            var newFilters = _filters == null ? new List<InternalFilter>() : new List<InternalFilter>(_filters);
```
jskeet, line 521: The crux of my suggestion is that we get rid of the concept of having a list of filters in Query. Instead of the `_filters` field, we have a field which is just a single filter, *always* a `CompositeFilter` with a type of "and". An unfiltered query would have 0 elements and can be omitted; we can simplify a filter with 1 element to just that element; with more than 1 element we'd combine them in the normal way.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -1077,18 +1123,18 @@ internal InternalOrdering(FieldPath field, Direction direction)
             public bool Equals(InternalOrdering other) => Field.Equals(other.Field) && Direction == other.Direction;
         }
 
-        private struct InternalFilter : IEquatable<InternalFilter>
+        internal class InternalFilter : IEquatable<InternalFilter>
```
jskeet, line 1126: I'd expect to be able to get rid of this entirely - potentially having other subclasses of Filter if necessary.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -509,6 +511,50 @@ object ConvertValueToDocumentReferencesForInQuery()
...
+            }
+            else
+            {
+                newFilters.Add(ParseCompositeFilter((Filter.CompositeFilter) filter));
+            }
+            
+            return new Query(_root, _offset, _limit, _orderings, newFilters, _projections, _startAt, _endAt);
+        }
+
+        private InternalFilter ParseFieldFilter(Filter.UnaryFilter unaryFilter)
```
jskeet, line 534: This isn't really parsing anything in the normal sense. It may not be needed in end anyway, if everything is rewritten in terms of Filter.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Filter.cs
```diff
@@ -0,0 +1,144 @@
...
+            Field = fieldPath;
+            _op = op;
+            _value = value;
+        }
+    }
+
+    /// <summary>
+    /// 
+    /// </summary>
+    internal class CompositeFilter : Filter
```
jskeet, line 59: I suspect you'll want a "WithAdditionalFilter" method or similar in order to implement my suggestion of keeping this as the only filter field within Query.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Filter.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.Collections.Generic;
+using FieldOp = Google.Cloud.Firestore.V1.StructuredQuery.Types.FieldFilter.Types.Operator;
+
+namespace Google.Cloud.Firestore;
+
+
+/// <summary>
+/// 
+/// </summary>
+public class Filter
```
jskeet, line 27: I'd:

- Make this abstract
- Have a private constructor (to stop anyone else from trying to subclass it)
- Have an abstract ToProto method which each of the subclasses implements

(Basically this will replace InternalProto)

jskeet (commented):
Hopefully this clarifies the design I was thinking of. Let me know if it's unclear.
(Or if you weren't trying to implement my idea, I can re-review with that in mind!)

hemanshv:
Thanks @jskeet  for the elaboration. Its quite clear now. 
I will update the doc and implementation approach as well.

Thanks,

hemanshv:
Hi @jskeet ,
I've updated the PR to validate my understanding about the approach. It is not completed yet. Tests, documentation and some  methods are still needs to be added. Please have a look and share your thoughts.

Thanks,

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -27,16 +27,18 @@
...
-{
-    /// <summary>
-    /// A query against a collection.
-    /// </summary>
-    /// <remarks>
-    /// <see cref="CollectionReference"/> derives from this class as a "return-all" query against the
-    /// collection it refers to.
-    /// </remarks>
+{
+/// <summary>
```
jskeet, line 35: Not sure why all the documentation here has been outdented...

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -493,10 +496,9 @@ private Query Where(FieldPath fieldPath, FieldOp op, object value)
                 };
             }
 
-            InternalFilter filter = InternalFilter.Create(Database.SerializationContext, fieldPath, op, value);
-            var newFilters = _filters == null ? new List<InternalFilter>() : new List<InternalFilter>(_filters);
-            newFilters.Add(filter);
-            return new Query(_root, _offset, _limit, _orderings, newFilters, _projections, _startAt, _endAt);
+            var newFilter = (_filter == null) ? new Filter.CompositeFilter(new List<Filter>(), CompositeFilter.Types.Operator.And)
```
jskeet, line 499: Any reason this isn't just `return Where(new Filter.UnaryFilter(fieldPath, op, value))`? (I suspect eventually we'll move all of this logic somewhere into Filter anyway.)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -509,6 +511,38 @@ object ConvertValueToDocumentReferencesForInQuery()
...
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="filter"></param>
+        /// <returns></returns>
+        public Query Where(Filter filter)
+        {
+            ValidateFilterForSentinels(filter);
+            var newFilter = (_filter == null) ? new Filter.CompositeFilter(new List<Filter>(), CompositeFilter.Types.Operator.And)
```
jskeet, line 522: Presumably you could just use `new List<Filter> { filter }` rather than calling `WithAdditionalFilter`?

Or you could use:

```csharp
var newFilter = (_filter ?? new Filter.CompositeFilter(new List<Filter>(), CompositeFilter.Types.Operator.And))
    .WithAdditionalFilter(filter);
```

(We might want a simple `Filter.EmptyAnd` or similar factory method, if we use it a lot...)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -509,6 +511,38 @@ object ConvertValueToDocumentReferencesForInQuery()
...
+            .WithAdditionalFilter(filter) : _filter.WithAdditionalFilter(filter);
+            return new Query(_root, _offset, _limit, _orderings, _projections, _startAt, _endAt, newFilter);
+        }
+
+        private void ValidateFilterForSentinels(Filter filter)
+        {
+            if (filter.GetType() == typeof(Filter.UnaryFilter))
```
jskeet, line 529: Calling `GetType()` is usually a design smell. I think it would make more sense to have an abstract `ValidaterForSentinels` method in `Filter`.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -509,6 +511,38 @@ object ConvertValueToDocumentReferencesForInQuery()
...
+            {
+                Filter.UnaryFilter uFilter = (Filter.UnaryFilter) filter;
+                var convertedValue = ValueSerializer.Serialize(Database.SerializationContext, uFilter.Value);
+                ValidateNoSentinelsRecursively(convertedValue, "Sentinel values cannot be specified in filters");
+            }
+            else
+            {
+                Filter.CompositeFilter compositeFilter = (Filter.CompositeFilter) filter;
+                foreach(Filter.UnaryFilter f in compositeFilter.Filters)
```
jskeet, line 538: I'd expect a composite filter to be able to contain other composite filters.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Filter.cs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// <summary>
+    /// 
+    /// </summary>
+    internal class CompositeFilter : Filter, IEquatable<CompositeFilter>
+    {
+        /// <summary>
+        /// 
+        /// </summary>
+        public IEnumerable<Filter> Filters { get; }
```
jskeet, line 116: I'd suggest making all the members of internal types internal as well, unless they're overriding public behavior. It makes it easier to review changes later. (Removing an internal member is fine; removing a public member *looks* like it's breaking, until you notice it's in an internal type.)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -78,17 +79,17 @@ private protected Query(FirestoreDb database, DocumentReference parent, string c
         private Query(
             QueryRoot root,
             int offset, (int count, LimitType type)? limit,
-            IReadOnlyList<InternalOrdering> orderings, IReadOnlyList<InternalFilter> filters, IReadOnlyList<FieldPath> projections,
-            Cursor startAt, Cursor endAt)
+            IReadOnlyList<InternalOrdering> orderings, IReadOnlyList<FieldPath> projections,
```
jskeet, line 82: If you just replace the existing parameter by changing the name and type, there are far fewer changes elsewhere. (You're not logically changing what's needed for construction after all - just how it's expressed.)

Indeed, it might end up being cleanest to do this in two commits: one which changes from `IReadOnlyList<InternalFilter> filters` to `Filter.CompositeFilter filters`, and then another commit which *just* changes the name later. That way all the "churn" changes are isolated.

jskeet (commented):
Yes, this is definitely in the direction I was thinking of. I haven't reviewed carefully - really just for direction. Shout if there are any aspects you want me to take a more careful look at.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
hemanshv: make sense this would be a recursive call then, missed that will update it in the final PR.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
hemanshv: yups that is the eventually what the aim right now all the things are not in-place so its like this.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Filter.cs
```diff
@@ -0,0 +1,240 @@
...
+    internal static bool IsOrderingFilter(UnaryFilter filter) =>
+        filter.Value is object &&
+        filter.Op == (int) FieldOp.GreaterThan ||
+        filter.Op == (int) FieldOp.GreaterThanOrEqual ||
+        filter.Op == (int) FieldOp.LessThan ||
+        filter.Op == (int) FieldOp.LessThanOrEqual;
+
+    internal static UnaryFilter GetOrderingFilter(CompositeFilter compositeFilter)
+    {
+        foreach (var filter in compositeFilter.Filters)
```
hemanshv, line 158: @jskeet Want your feedback on this ordering part. We are doing DFS to find the first ordering filter and using it as default one. Please share your thoughts on this.

jskeet: That's not something I can answer - please ask the Firestore SDK team what their requirements are.

hemanshv: Thanks Jon will check with them. Thanks for the feedback. I will raise the final PR now with all the changes.

jskeet:
Can we close this now, as it's superseded by #10023?

hemanshv:
yeah sure Jon

Issue #9238, opened 2022-10-19
Pub/Sub client life-cycles should be more DI-friendly


jskeet:
This is effectively about addressing the latest comment in #5893.

If a singleton `PublisherClient` is registered in DI, I *suspect* it's entirely possible to lose published messages during shutdown due to buffering. We should work out what's required, and document it really clearly.

Rishabh-V:
Thanks Jon. I am on vacation for the entire next week. I will start work on it after I am back. Thanks.

Rishabh-V:
Will close it after docs are updated.

PR #9986, opened 2023-03-03
docs: Updates documentation to use the PublisherClient and SubscriberClient in a DI friendly way.

Rishabh-V:


Rishabh-V:
closes #9238

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,139 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
...
+```csharp
+services.AddPublisherClient(builder =>
+{
+    builder.TopicName = TopicName.FromProjectTopic("projectId", "topicId");
+    builder.CredentialsPath = "path/to/credentials.json";
+    // Other settings to customize the client.
+});
+```
+
+Using the `PublisherClient` registered with the dependency injection container is easy. To do so, simply create a constructor for the class in which the `PublisherClient`
```
amanda-tarafa, line 111: This only works in ASP.NET Core for the types that have automatic integration with DI. An example on how to do this in non ASP.NET application might be useful, maybe more so for subscribers that publishers, but they would be equivalent.

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,139 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
...
+    // Other settings to customize the client.
+});
+```
+
+The `SubscriberClient` is a long-running client and so it may be useful to use
+it in a background service. The background service can use the `SubscriberClient`
+registered with the dependency injection container and handle the messages in the background.
+
+In ASP.NET Core, the background services can be registered with the dependency injection container
+using the `AddHostedService` extension method as shown below:
```
amanda-tarafa, line 158: What package/namespace is this on? Use full names the first time you mention a type, please.

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,139 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
...
+    }
+}
+```
+
+During application shutdown, the `StopAsync` method of the `SubscriberService` is invoked, which in turn calls
+the `StopAsync` method of the `SubscriberClient` to gracefully shut down the client.
+
+## IAsyncDisposable
+
+Both `PublisherClient` and `SubscriberClient` implements the `IAsyncDisposable` interface,
```
amanda-tarafa, line 203:
```suggestion
Both `PublisherClient` and `SubscriberClient` implement the `IAsyncDisposable` interface,
```

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,139 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
...
+
+For the `PublisherClient`, when the `DisposeAsync` method is called, it asynchronously waits for the time interval
+specified in the `PublisherClient.Settings.DisposeTimeout` property for the `PublisherClient` to send any pending messages before aborting
+the clean shutdown process, that may leave some locally queued messages unsent. The time interval can be customized
+by setting the `PublisherClient.Settings.DisposeTimeout` property. If no value is specified, the default value of 5 seconds is used.
+
+For the `SubscriberClient`, when the `DisposeAsync` method is called, it asynchronously waits for the time interval
+specified in the `SubscriberClient.Settings.DisposeTimeout` property for the `SubscriberClient` to acknowledge the handled messages before aborting the
+the clean stop process, that may leave some handled messages un-acknowledged. The time interval can be customized
+by setting the `SubscriberClient.Settings.DisposeTimeout` property. If no value is specified, the default value of 5 seconds is used.
```
amanda-tarafa, line 216: This section is not a child section of the Dependency injection section so it shouldn't mention dependency injection aspects. This section only discribes the mechanis of disposing, and on the DI section, you may say something like "the DI container will dispose of all services registered as singletos, see the Dispose section for more information on what that means for Publisher and Subscriber".

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,139 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
 instance, it is recommended that a singleton client per topic is used for the lifetime of the
 application.
 
+## Dependency Injection
```
amanda-tarafa, line 85: Wondering outloud whether we sould add extension methods to configure a PublisherClient provider, so that publishers for different topics may be registered and fetched by the topic name. Same for subscribers.

This would just be a utility that users could built themselves of course.

@jskeet what do you think?

amanda-tarafa (changes requested):
Apart from the specific comments, there is a lot of embeded code in the docs that should be moved to snippets, ideally tested snippets (see the Google.Cloud.Diagnostics.AspNetCore3.Snippets for instance). But at the very least, this should be code that is built whenever we make changes to the library, else, we run the risk of getting outdated samples very soon.

Rishabh-V (commented):
Thanks Amanda. A couple of questions.

We have unit tests in place in the library for all the DI methods and the embedded code snippets are just showing how the extension method can be invoked. Should I create snippets that creates IServiceCollection and invoke the four extension methods?

For the BackgroundService class and DI resolution, I am not sure how I can fit them in a snippet, so I'll reach out to you early next week.

I 'll push a commit with all the other fixes post your inputs. Thank you.

amanda-tarafa:
> We have unit tests in place in the library for all the DI methods and the embedded code snippets are just showing how the extension method can be invoked. Should I create snippets that creates IServiceCollection and invoke the four extension methods?

Yes, we want to make sure that the code included in docs does not get outdated or starts failing. Unit tests do nothing for that as they are not testing the embeded code. Else, whenever we make a change we have to remember to check this embeded code to see if it's affected. If you move the code to snippets, even if you don't test (you should) we at least know that it builds.
 
> For the BackgroundService class and DI resolution, I am not sure how I can fit them in a snippet, so I'll reach out to you early next week.

Take a look first at the Google.Cloud.Diagnostics.ApsNetCore3.Snippets/IntegrationTests

Rishabh-V:
Sure Amanda. I 'll have a look; create snippets and refer them here. 

PS: I kept the snippets inline as they are simple extension methods on the `IServiceCollection`. I  thought of them similar to the XML documentation that we update as we modify the code. However, I understand your point that having a sample snippet will help us remember to update the documentation in case of breaking changes. Thanks.

apis/Google.Cloud.PubSub.V1/docs/index.md
Rishabh-V: I created a sample console app, configured the DI container and the construction injection works just fine as above. Please let me know if I am missing something.

amanda-tarafa: You have a controller example which is specific to ASP.NET Core.

`simply create a constructor for the class in which the `PublisherClient` this statement is true only if the type with the constructor is also resolved via DI. In APS.NET Core controllers are resolved via DI automatically, but not on other types of applicatons, i.e. you also need to configure the type with the constructor for DI.
So the better example I think is to show how to resolve a PublisherClient from DI directly, which is the more general example, and then maybe mention that in ASP.NET Core controller's (and other automatically resolved types like middlewares, etc.) that happens automatically. But just maybe.

Also, you have the link to MS's package, but I think it's best to add a link to https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection (to both is fine).

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,139 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
...
+
+Both `PublisherClient` and `SubscriberClient` can be easily integrated with the dependency injection
+container provided by the [Microsoft.Extensions.DependencyInjection](https://www.nuget.org/packages/Microsoft.Extensions.DependencyInjection/) package.
+
+### PublisherClient
+
+To register a singleton `PublisherClient` instance with default settings in the `IServiceCollection`, use the
+`AddPublisherClient` extension method as shown below:
```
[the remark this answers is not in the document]
amanda-tarafa: You've followed best practices and that's great, but users don't have to be familiar with best practices or don't have to asume you've followed them.

When you introduce these extension methods, a note saying something like "These methods are defined in the Google.Cloud.Pubsub.V1 package, in the namespace Microsoft. Extensions......" makes things clear.

apis/Google.Cloud.PubSub.V1/docs/index.md
amanda-tarafa: Again, mentioning the origin of a method is always useful. If it's weird to se the full name for an extension method (I agree it is), do something like (`AddHostedService`)[link to MS reference doc for this method].

Time and time again, we (across all languages) have received feedback from users that imports are not clear in samples.

amanda-tarafa (changes requested):

apis/Google.Cloud.PubSub.V1/docs/index.md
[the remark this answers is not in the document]
Rishabh-V: Sure Amanda.

apis/Google.Cloud.PubSub.V1/docs/index.md
Rishabh-V: Sure Amanda. Will add the link. 
For the non-ASP.NET Core apps, I think it depends on how the DI is configured. For example the below code will work fine in a Console app. However, retrieving a type from the ServiceProvider seems like an anti-pattern to show in docs. Please let me know if you had other thoughts and I can incorporate that. 

```
public class MyService
{
    private readonly PublisherClient _client;
    public MyService(PublisherClient client)
    {
        _client = client;
    }
    
    public void PublishMessage(string message)
    {
    }
}
-------
IServiceCollection services = new ServiceCollection();

// Add services to the container.
services.AddPublisherClient(topicName); 
services.AddSingleton<MyService>();

// Get the ServiceProvider.
var provider = services.BuildServiceProvider();

// Resolve the service and use it.
var service = provider.GetRequiredService<MyService>();
service.PublishMessage(message);

// Or directly resolve the PublisherClient.
var publisher = provider.GetRequiredService<PublisherClient>();
await publisher.PublishAsync(message);
```

apis/Google.Cloud.PubSub.V1/docs/index.md
Rishabh-V: Sure Amanda. I will do so. Thanks.

Rishabh-V (commented):
Thank you Amanda. This is helpful. Just one confirmation about non-ASP.NET Core apps and I will push the commit with fixes. Thanks.

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,80 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
...
+During application shutdown, the `StopAsync` method of the `SubscriberService` is invoked, which in turn calls
+the `StopAsync` method of the `SubscriberClient` to gracefully shut down the client.
+
+## IAsyncDisposable
+
+Both `PublisherClient` and `SubscriberClient` implement the `IAsyncDisposable` interface,
+allowing for asynchronous resource cleanup.
+
+For the `PublisherClient`, when the `DisposeAsync` method is called, it asynchronously waits for the time interval
+specified in the `PublisherClient.Settings.DisposeTimeout` property for the `PublisherClient` to send any pending messages before aborting
```
amanda-tarafa, line 150: Will this work well with DI? What happens if I set this to 100 hours for instance? Will DI just wait for it to be done or should I also configure DI to wait for that long?

apis/apis.json
```diff
@@ -3426,13 +3426,14 @@
...
         "Grpc.Core": "2.46.5"
       },
-      "testDependencies": {
-        "Google.Api.Gax.Testing": "4.3.0",
-        "Google.Apis.CloudResourceManager.v1": "1.60.0.2950",
-        "Microsoft.Extensions.DependencyInjection": "6.0.0",
-        "System.ValueTuple": "4.5.0",
-        "Xunit.Combinatorial": "1.4.1"
-      },
+        "testDependencies": {
```
amanda-tarafa, line 3429: Revert the indent changes here.

amanda-tarafa (changes requested):

apis/Google.Cloud.PubSub.V1/docs/index.md
jskeet: I suggest we wait until people ask for it. It's easy enough to add later.

Rishabh-V (commented):
I have tried to address all the open items. Please take a look.

jskeet:
Thanks Amanda - I'll review now as well. (I don't expect to find anything you've missed though :)

apis/apis.json
```diff
@@ -5071,4 +5073,4 @@
       ]
     }
   ]
-}
\ No newline at end of file
+}
```
jskeet, line 5076: I suspect it's worth removing this trailing newline otherwise it'll show up again when it's regenerated for some reason.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Snippets/SubscriberClientSnippets.cs
```diff
@@ -36,5 +41,92 @@ public void Emulator()
...
+
+        [Fact]
+        public async Task UseSubscriberServiceInConsoleApp()
+        {            
+            string projectId = "projectId";
+            string subscriptionId = "subscriptionId";
+
+            // Sample: UseSubscriberServiceInConsoleApp
+            // Add `using Microsoft.Extensions.Hosting;` in the using directives.
+            // Below code goes inside the Main method, i.e., public static async Task Main(string[] args)
```
jskeet, line 93: I'd remove this line to be honest - especially as with minimal API templates, people may well be using top-level statements anyway, where the Main method is implicit.

(It's not big deal if you decide to leave it though. I'd reword "Below code" to "The code below" though.)

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,99 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
...
+[Disposing of the publisher and subscriber clients](#disposing-of-the-publisher-and-subscriber-clients) for more information about what happens when disposing the `PublisherClient`.
+
+### SubscriberClient
+
+To register a singleton `SubscriberClient` instance with default settings in the `IServiceCollection`, use the
+`AddSubscriberClient` extension method as shown below:
+
+{{sample:SubscriberClient.AddSubscriberClient}}
+
+There is an overload of `AddSubscriberClient` method that takes `Action<SubscriberClientBuilder>` as a parameter
```
jskeet, line 126:
```suggestion
There is an overload of the `AddSubscriberClient` method that takes `Action<SubscriberClientBuilder>` as a parameter
```

(Or just "There is an overload of `AddSubscriberClient` that takes...")

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,99 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
...
+injection container in the `Microsoft.Extensions.DependencyInjection` namespace.
+
+### PublisherClient
+
+To register a singleton `PublisherClient` instance with default settings in the `IServiceCollection`, use the
+`AddPublisherClient` extension method as shown below:
+
+{{sample:PublisherClient.AddPublisherClient}}
+
+There is an overload of `AddPublisherClient` method that takes `Action<PublisherClientBuilder>` as a parameter
```
jskeet, line 99: Same comment as I've written below for AddSubscriberClient. (I just missed it when reading this the first time.)

apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,6 +82,99 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
...
+
+For the `SubscriberClient`, when the `DisposeAsync` method is called, it asynchronously waits for the time interval
+specified in the `SubscriberClient.Settings.DisposeTimeout` property for the `SubscriberClient` to acknowledge the handled messages before aborting the
+the clean stop process, that may leave some handled messages un-acknowledged. The time interval can be customized
+by setting the `SubscriberClient.Settings.DisposeTimeout` property. If no value is specified, the default value of 5 seconds is used.
+
+Please note that when working with the dependency injection container, if the timeout interval specified in the properties above is greater than the default value,
+the dependency injection container must be configured to have a timeout greater than or equal to this time. Failure to do so may result in the abrupt termination of the client shutdown process.
+
+See [ShutdownTimeout](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-6.0#shutdowntimeout) property for configuring the shutdown timeout in ASP.NET Core applications.
```
jskeet, line 173:
```suggestion
See [HostOptions.ShutdownTimeout](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host?view=aspnetcore-6.0#shutdowntimeout) for configuring the shutdown timeout in ASP.NET Core applications.
```

(Or have "the HostOptions.ShutdownTimeout property" if you want to include the word "property".

jskeet (approved):
A few nits, but basically we're there :)

Rishabh-V (commented):
Thanks Jon. I have addressed the review comments. PTAL. If this looks fine to you, I can squash into one commit and merge.  Thanks.

jskeet (approved):
Yup, those all look good, thanks.

Rishabh-V:
Thanks Jon. Merging.

PR #10006, opened 2023-03-07
chore: Update docfx and move to Markdig

jskeet:
(Markdig is the default Markdown engine in docfx now. We were only using dfm explicitly due to the previous Markdig integration not supporting overwrite files with multiple sections.)


amanda-tarafa (approved):
LGTM, nice they fixed this!

PR #10027, opened 2023-03-13
chore: CI changes

anshuldavid13:


anshuldavid13:
Retry conformance test run with detailed logs at : https://github.com/googleapis/google-cloud-dotnet/actions/runs/4410137035/jobs/7727278176

anshuldavid13:
Open for LEX team review

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.RetryConformanceTests/RetryConformanceTest.cs
```diff
@@ -68,19 +68,26 @@ public async Task RetryTest(RetryTestCase testCase)
 
         if (test.ExpectSuccess)
         {
+            Console.WriteLine("**************** STARTED TEST " + method.Name + " ********************");
```
hemanshv: may be we can use string interpolation here and other places.

hemanshv:
LGTM

jskeet (approved):
Looks good - please rebase to one commit, and ideally rename the PR to have a conventional commit name as well.

anshuldavid13:
> Looks good - please rebase to one commit, and ideally rename the PR to have a conventional commit name as well.
Is it ok if I keep both the commits in this one. The previous one has logs set that might come in handy to quickly debug if we come back to the skipped test cases later or get the storage test bench working without the workarounds. Can also compare it the existing logs in the past run of this PR which is kept for record. I will rename the PR to have conventional commits.

jskeet:
Hmm. I'm in two minds about this - if we want logging to be available, I'd prefer it to be available with cleaner code, and ideally in a way that lets us enable it *without* having to go back to an old commit. On the other hand, the cost of keeping the commit is pretty low.

On the balance, I'm okay with keeping it - but *if* we find we actually use it (even once) I'd like us to put a little more time (half an hour would be enough I suspect) to add logging in a less ad-hoc manner.

anshuldavid13:
> Hmm. I'm in two minds about this - if we want logging to be available, I'd prefer it to be available with cleaner code, and ideally in a way that lets us enable it _without_ having to go back to an old commit. On the other hand, the cost of keeping the commit is pretty low.
> 
> On the balance, I'm okay with keeping it - but _if_ we find we actually use it (even once) I'd like us to put a little more time (half an hour would be enough I suspect) to add logging in a less ad-hoc manner.

Thanks @jskeet

PR #10043, opened 2023-03-15
chore: update action to trigger conformance tests only for storage PRs

Rishabh-V:


Rishabh-V:
I will push a dummy commit in storage (and then revert in next commit) to see if it fires for storage.

jskeet (approved):
I suspect we don't need the `**` at the start, but it won't do any harm.

Rishabh-V:
Conformance tests are being run now. Will revert the whitespace change in next commit.

Rishabh-V:
Triggering conformance tests one more time to test it works with updated path without starting**. (Will revert it in next commit)

Rishabh-V:
> I suspect we don't need the `**` at the start, but it won't do any harm.

Yes Jon. You are correct. Removed starting ** and it is working :)

amanda-tarafa (approved):

PR #10050, opened 2023-03-17
chore: Update Firestore code to follow editorconfig requirements

jskeet:
There are no semantically-significant changes here.

When making some fields readonly, I've sometimes reordered so that all the readonly fields appear before the non-readonly ones.


Rishabh-V (approved):
LGTM

PR #10089, opened 2023-03-24
feat: Customize Retry options

anshuldavid13:
This has the core logic for processing the custom retry options available. It is yet to be implemented across all the other APIs, that will be done once this code logic is reviewed.


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// </summary>
+    public TimeSpan MaxBackoff { get; private set; }
+
+    /// <summary>
+    /// The multiplier to apply to the backoff on each iteration. It is always greater than or equal to 1.0.
+    /// </summary>
+    public double BackoffMultiplier { get; private set; }
+
+    /// <summary>
+    /// The default initial backoff time between the first attempt and the first retry. It is set to 1 second by default.
```
jskeet, line 40:
```suggestion
    /// The default initial backoff time between the first attempt and the first retry. The default is 1 second.
```

(Otherwise it sounds like you can modify the default.)

Same applies for the other defaults below.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+    {
+        InitialBackoff = InitialBackoff,
+        MaxBackoff = MaxBackoff,
+        BackoffMultiplier = BackoffMultiplier
+    };
+
+    /// <summary>
+    /// Returns the existing retry timings configurations using the specified initial backoff.
+    /// </summary>
+    /// <param name="initialBackoff">Custom initial Backoff for retry.</param>
```
jskeet, line 69:
```suggestion
    /// <param name="initialBackoff">The new value for <see cref="InitialBackoff" />.</param>
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// <returns>Retry timings with updated initial backoff.</returns>
```
jskeet, line 70:
```suggestion
    /// <returns>A new instance based on the existing values, but with the specified initial backoff.</returns>
```

"Update" sounds like the existing instance will be modified.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+
+/// <summary>
+/// These specify the custom timing configurations for retrying in case of failure of the API request.
+/// </summary>
+public sealed class RetryTimings
+{
+    /// <summary>
+    /// The backoff time between the first attempt and the first retry. It must be non-negative. 
+    /// </summary>
+    public TimeSpan InitialBackoff { get; private set; }
```
jskeet, line 27: I'd prefer to make all these properties *properly* read-only. Instead, we should always use the parameterized constructor accepting all values. That makes it easier to tell that it's really immutable.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Storage.V1;
+
+/// <summary>
+/// These specify the custom timing configurations for retrying in case of failure of the API request.
```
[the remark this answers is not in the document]
jskeet: We might also want to indicate (on this, RetryPredicate and RetryOptions) that it's immutable. That way users can be confident that they can reuse the same timings/predicate/options throughout their code.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// The default maximum backoff multiplier to be applied on each iteration. It is set to 2 by default.
+    /// </summary>
+    public static double DefaultBackoffMultiplier { get; } = 2;
+
+    /// <summary>
+    /// Returns the retry timing configurations all set to default.
+    /// </summary>
+    public static RetryTimings Default { get; } = new();
+
+    internal RetryTimings Clone() => new()
```
jskeet, line 59: I hope we'd be able to remove this method entirely, having added the private constructor accepting the three values.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// <param name="backoffMultiplier">Custom retry delay Backoff Multiplier for retry.</param>
+    /// <returns>Retry timings with updated backoff multiplier.</returns>
+    public RetryTimings WithBackoffMultiplier(double backoffMultiplier)
+    {
+        var clone = Clone();
+        clone.BackoffMultiplier = backoffMultiplier;
+        return clone;
+    }
+
+    /// <summary>
```
jskeet, line 102: I would remove this - we've already got the static `Default` property, which is more descriptive.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// </summary>
+    public RetryTimings()
+    {
+        InitialBackoff = RetryTimings.DefaultInitialBackoff;
+        MaxBackoff = RetryTimings.DefaultMaxBackoff;
+        BackoffMultiplier = RetryTimings.DefaultBackoffMultiplier;
+    }
+
+    /// <summary>
+    /// Constructor to pass the custom timing configurations for retrying.
```
jskeet, line 113:
```suggestion
    /// Creates an instance with the specified values.
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// </summary>
+    /// <param name="initialBackoff">Initial Backoff for retry.</param>
```
jskeet, line 115: Nit: Backoff => backoff in all comments

This is where we should specify the constraints in terms of "must" using `paramRef`.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// <param name="maxBackoff">Maximum Backoff for retry.</param>
+    /// <param name="backoffMultiplier">Retry delay Backoff Multiplier for retry.</param>
+    public RetryTimings(TimeSpan initialBackoff, TimeSpan maxBackoff, double backoffMultiplier)
+    {
+        InitialBackoff = initialBackoff >= TimeSpan.Zero ?initialBackoff : throw new ArgumentOutOfRangeException("InitialBackoff value must be non-negative.") ;
```
jskeet, line 120: Nit: space after `?`.

(It's a shame that GaxPreconditions doesn't actually help here.)

Please specify the parameter name (using `nameof`) as the first argument to the `ArgumentOutOfRange` constructor in each case. (And then we probably don't need the name in the message at all - but we might want to include the value.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+        MaxBackoff = maxBackoff >= initialBackoff ? maxBackoff : throw new ArgumentOutOfRangeException("MaxBackoff value must be at least as much as initialBackoff.");
+        BackoffMultiplier = backoffMultiplier >= 1 ? backoffMultiplier : throw new ArgumentOutOfRangeException("BackoffMultiplier value must atleast be 1.");
```
jskeet, line 122:
```suggestion
        BackoffMultiplier = backoffMultiplier >= 1 ? backoffMultiplier : throw new ArgumentOutOfRangeException("BackoffMultiplier value must at least be 1.");
```

(And the changes described above.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+    /// <summary>
+    /// Constructor to pass the custom timing configurations for retrying.
+    /// </summary>
+    /// <param name="initialBackoff">Initial Backoff for retry.</param>
+    /// <param name="maxBackoff">Maximum Backoff for retry.</param>
+    /// <param name="backoffMultiplier">Retry delay Backoff Multiplier for retry.</param>
+    public RetryTimings(TimeSpan initialBackoff, TimeSpan maxBackoff, double backoffMultiplier)
+    {
+        InitialBackoff = initialBackoff >= TimeSpan.Zero ?initialBackoff : throw new ArgumentOutOfRangeException("InitialBackoff value must be non-negative.") ;
+        MaxBackoff = maxBackoff >= initialBackoff ? maxBackoff : throw new ArgumentOutOfRangeException("MaxBackoff value must be at least as much as initialBackoff.");
```
jskeet, line 121:
```suggestion
        MaxBackoff = maxBackoff >= initialBackoff ? maxBackoff : throw new ArgumentOutOfRangeException("MaxBackoff value must be at least as long as initialBackoff.");
```

(And the changes described above.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryHandlerTest.cs
```diff
@@ -16,38 +16,37 @@
 using System.Net.Http;
 using Xunit;
 
-namespace Google.Cloud.Storage.V1.Tests
```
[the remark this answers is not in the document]
jskeet: We'll need more actual tests here to make sure the retry options are actually honoured though.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CopyObjectOptions.cs
```diff
@@ -128,6 +128,11 @@ public sealed class CopyObjectOptions
         /// </summary>
         public string UserProject { get; set; }
 
+        /// <summary>
+        /// Options to pass custom retry configuration for each API request.
```
jskeet, line 132: It's fine to use the term "custom" here and in other operation-specific options classes - but elsewhere it's just "the retry options" without differentiation between default or custom retry options.

(This comment is just intended to give some more background for other suggested changes.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -37,30 +37,35 @@ internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
...
-        private RetryHandler() { }
-
-        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request)
+            _retryPredicate = retryOptions.RetryPredicate;
+            _backoffMultiplier = retryOptions.RetryTimings.BackoffMultiplier;
+            _initialBackoff = retryOptions.RetryTimings.InitialBackoff;
+            _maxBackoff = retryOptions.RetryTimings.MaxBackoff;
+        }
+
+        // TODO: Provided null options for now, will remove it once the implementation across all APIs is complete
```
jskeet, line 53: I'd suggest having a separate overload without the RetryOptions, that we intend to remove at the end. Make `options` a required parameter here, and define what a value of null means. (I'd suggest either "never" or prohibit it entirely.)

jskeet (commented):
Please don't let the number of comments discourage you - I think the *core* design of this is fine. We'll definitely want more tests though... I think they can mostly be in RetryHandlerTest with a fake service to separate that out entirely from real storage operations.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryHandlerTest.cs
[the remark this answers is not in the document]
shivgautam: "Hide whitespace" under settings can help in this case.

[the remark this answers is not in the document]
shivgautam: I meant just for the ease of review - not ignoring the white space. Should have been clear. The formatting needs to be fixed.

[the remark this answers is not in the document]
jskeet: The formatting *is* okay in the long run - the problem is that the same commit is making changes to formatting (moving the test to use a file-scoped namespace, which is a positive change in itself) *and* making functional changes.
I agree that reviewing without whitespace would be a workaround - but then it would hide other changes that actually need to be fixed. I think it's cleaner to separate out the different kinds of change in different commits, so that we can review all commits with whitespace *shown* in order to see any issues, but without detracting from the review.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/CustomRetryImplTest.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Collections.Generic;
+using System.Net;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Storage.V1.Tests;
+public class CustomRetryImplTest
+{
+
+    [Fact]
```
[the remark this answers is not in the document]
anshuldavid13: I have modified the tests.. hope its better now

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+    public static RetryOptions IdempotentRetryOptions { get; } = new RetryOptions(RetryTimings.Default, RetryPredicate.DefaultErrorCodes);
+
+    /// <summary>
+    /// This configuration ensures that the api request is never retried on failure.
+    /// </summary>
+    public static RetryOptions Never { get; } = new RetryOptions(new RetryTimings(), RetryPredicate.EmptyPredicate);
+
+    internal RetryPredicate RetryPredicate { get; }
+    internal RetryTimings RetryTimings { get; }
+
```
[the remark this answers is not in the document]
anshuldavid13: Sure Jon, we can work on that in the future

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTimings.cs
```diff
@@ -0,0 +1,124 @@
...
+// limitations under the License.
+
+using System;
+
+namespace Google.Cloud.Storage.V1;
+
+/// <summary>
+/// These specify the custom timing configurations for retrying in case of failure of the API request.
+/// </summary>
+public sealed class RetryTimings
```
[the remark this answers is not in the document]
anshuldavid13: Sure

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryHandlerTest.cs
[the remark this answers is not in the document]
anshuldavid13: I have raised another PR for these changes as suggested

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryOptionsTest.cs
```diff
@@ -0,0 +1,35 @@
...
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using System;
+using Xunit;
+
+namespace Google.Cloud.Storage.V1.Tests;
+public class RetryOptionsTest
+{
+    [Fact]
```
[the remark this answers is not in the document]
anshuldavid13: Modified the test cases

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -69,14 +74,10 @@ public async Task<bool> HandleResponseAsync(HandleUnsuccessfulResponseArgs args)
...
-            // The first failure will have args.CurrentFailedTry set to 1,
-            // whereas we want the first delay to be 1 second. We use Math.Min on the power
-            // rather than on the result to obtain a max retry of 32 seconds without risking
-            // calling Math.Pow with a huge number.
-            int power = Math.Min(args.CurrentFailedTry - 1, 5);
-            double seconds = Math.Pow(2.0, power);
-            var delay = TimeSpan.FromSeconds(seconds);
-            await Task.Delay(delay, args.CancellationToken).ConfigureAwait(false);
+
+            TimeSpan delay = _initialBackoff + TimeSpan.FromSeconds((args.CurrentFailedTry - 1) * _backoffMultiplier);
```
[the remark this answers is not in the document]
anshuldavid13: Refactored and Updated the logic to make it exponential

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryPredicate.cs
```diff
@@ -0,0 +1,62 @@
...
+    public static RetryPredicate EmptyPredicate { get; } = RetryPredicate.FromErrorCodes();
+
+    /// <summary>
+    /// It configures retrying based on the specified custom retriable error codes.
+    /// In case the API request fails with an error code provided in this list, it is retried.
+    /// These error codes completely override the default error codes provided and are not appended to it.
+    /// </summary>
+    /// <param name="errorCodes">Error codes on which to retry.</param>
+    /// <returns>Returns the retry predicate with the custom error codes specified.</returns>
+    public static RetryPredicate FromErrorCodes(params int[] errorCodes) => new() { _predicate = x => errorCodes.Contains(x) };
```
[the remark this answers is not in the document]
anshuldavid13: Null is currently an accepted input as it shows it will not retry for any error codes ( as is used in case of Never) .. Should I be changing that?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryPredicate.cs
```diff
@@ -0,0 +1,64 @@
...
+
+    /// <summary>
+    /// Returns a Retry Predicate which will ensure that the operation will never retry in case of failure.
+    /// </summary>
+    public static RetryPredicate Never { get; } = RetryPredicate.FromErrorCodes();
+
+    /// <summary>
+    /// It configures retrying based on the specified custom retriable error codes.
+    /// In case the API request fails with an error code provided in this list, it is retried.
+    /// These error codes completely override the default error codes provided and are not appended to it.
```
jskeet, line 47:
```suggestion
    /// Returns a predicate which will retry on the specified HTTP error codes.
    /// Note that only these error codes will be retried; the default error codes are not used by the returned predicate.
```

(We're only creating a RetryPredicate here - we can't say how it's going to be used

jskeet: In general, try to avoid "It creates" or "It configures" language in documentation comments.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryPredicate.cs
```diff
@@ -0,0 +1,64 @@
...
+    /// </summary>
+    /// <param name="errorCodes">Error codes on which to retry.</param>
+    /// <returns>Returns the retry predicate with the custom error codes specified.</returns>
```
jskeet, line 50:
```suggestion
    /// <returns>Returns a predicate which retries on the specified error codes.</returns>
```

(Again, there's no need to mention "custom".)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryPredicate.cs
```diff
@@ -0,0 +1,64 @@
...
+    public static RetryPredicate FromErrorCodes(params int[] errorCodes) => new(x => errorCodes.Contains(x));
+
+    /// <summary>
+    /// It configures the conditions based on which the operation is retried in case of failure.
+    /// If configured, it completely override the default error codes provided and they are not considered anymore for retrying.
+    /// </summary>
+    /// <param name="predicate">Predicate based on which to decide whether to retry or not.</param>
+    /// <returns>Returns the retry predicate with the conditions specified for retrying.</returns>
```
jskeet, line 58:
```suggestion
    /// <summary>
    /// Returns a predicate which will retry based on the result of the specified delegate, which is provided with
    /// the HTTP error code for the response.
    /// </summary>
    /// <param name="predicate">Delete used to determine whether or not to retry.</param>
    /// <returns>Returns the retry predicate with the conditions specified for retrying.</returns>
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryPredicate.cs
```diff
@@ -0,0 +1,64 @@
...
+    public static RetryPredicate FromErrorCodePredicate(Func<int, bool> predicate) => new(predicate);
```
jskeet, line 59: Either here or on in the constructor, we should validate that the predicate is non-null. Otherwise we'll get a NullReferenceException in ShouldRetry.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTiming.cs
```diff
@@ -0,0 +1,109 @@
...
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using System;
+using System.Threading.Tasks;
+
+namespace Google.Cloud.Storage.V1;
+
+/// <summary>
+/// Options to control the delays between attempts when retrying failed API requests. These are immutable.
```
jskeet, line 21:
```suggestion
/// Options to control the delays between attempts when retrying failed API requests. This class is immutable.
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTiming.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// </summary>
+    public TimeSpan InitialBackoff { get; }
+
+    /// <summary>
+    /// Maximum backoff time between retries.
+    /// </summary>
+    public TimeSpan MaxBackoff { get; }
+
+    /// <summary>
+    /// The multiplier to apply to the backoff on each iteration.
```
jskeet, line 36:
```suggestion
    /// The multiplier to apply to the previous backoff on each iteration, leading to exponential backoff.
```

(Just to make it clear.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTiming.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// </summary>
+    public static double DefaultBackoffMultiplier { get; } = 2;
+
+    /// <summary>
+    /// Returns the retry timing configurations all set to default.
+    /// </summary>
+    public static RetryTiming Default { get; } = new();
+
+    /// <summary>
+    /// Returns the existing retry timings configurations using the specified initial backoff.
```
jskeet, line 61:
```suggestion
    /// Returns a new instance with the same values as the current instance, but with the specified initial backoff.
```

(And the same for the other methods.)

I know that basically repeats the `<return>` element, but I think it's clearer than "existing".

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTiming.cs
```diff
@@ -0,0 +1,109 @@
...
+        MaxBackoff = RetryTiming.DefaultMaxBackoff;
+        BackoffMultiplier = RetryTiming.DefaultBackoffMultiplier;
+    }
+
+    /// <summary>
+    /// Creates an instance with the specified values.
+    /// </summary>
+    /// <param name="initialBackoff">Initial backoff for retry. It must be non-negative.</param>
+    /// <param name="maxBackoff">Maximum backoff for retry. This is never smaller than <paramref name="initialBackoff" />.</param>
+    /// <param name="backoffMultiplier">Backoff multiplier for retry. It is always greater than or equal to 1.0.</param>
```
jskeet, line 96:
```suggestion
    /// <param name="initialBackoff">Initial backoff for retry. Must be non-negative.</param>
    /// <param name="maxBackoff">Maximum backoff for retry. Must be greater than or equal to <paramref name="initialBackoff" />.</param>
    /// <param name="backoffMultiplier">Backoff multiplier for retry. Must be greater than or equal to 1.0.</param>
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTiming.cs
```diff
@@ -0,0 +1,109 @@
...
+    public RetryTiming(TimeSpan initialBackoff, TimeSpan maxBackoff, double backoffMultiplier)
+    {
+        InitialBackoff = initialBackoff >= TimeSpan.Zero ? initialBackoff : throw new ArgumentOutOfRangeException(nameof(initialBackoff), $"'{InitialBackoff}' parameter value must be non-negative.");
+        MaxBackoff = maxBackoff >= initialBackoff ? maxBackoff : throw new ArgumentOutOfRangeException(nameof(maxBackoff), $"{MaxBackoff} parameter value must be at least as long as initialBackoff.");
+        BackoffMultiplier = backoffMultiplier >= 1 ? backoffMultiplier : throw new ArgumentOutOfRangeException(nameof(backoffMultiplier), $"{BackoffMultiplier} parameter value must at least be 1.");
+    }
+
+    internal TimeSpan GetDelay(int failureCount)
+    {
+        TimeSpan delay = failureCount > 1 ? InitialBackoff + TimeSpan.FromSeconds(Math.Pow(2.0, failureCount - 2) * BackoffMultiplier) : InitialBackoff;
```
jskeet, line 106: As noted earlier, this isn't correct. Please make sure you understand the expected meaning of BackoffMultiplier, write some tests, and then fix them appropriately.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -37,46 +36,47 @@ internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
...
-            request.AddUnsuccessfulResponseHandler(s_instance);
+            request.AddUnsuccessfulResponseHandler(retryHandler);
         }
 
-        // This function is designed to support asynchrony in case we need to examine the response content, but for now we only need the status code
-        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response) => 
-            Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
+        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request, RetryOptions options)
+        {
+            //In case null retry options or Never is passed, retry will not happen in case of failure
```
jskeet, line 56: Nit: space after `//` and a period at the end.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryHandlerTest.cs
jskeet, line 1: As discussed, I'd expect tests that check that RetryHandler observes the timing and predicate (and doesn't retry at all when the MarkAsRetriable method is provided null options).

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTimingTest.cs
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using System;
+using Xunit;
+
+namespace Google.Cloud.Storage.V1.Tests;
+public class RetryTimingTest
```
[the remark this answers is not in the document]
jskeet: (Discussed in meeting)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -37,30 +37,34 @@ internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
...
+            RetryHandler retryHandler = new RetryHandler(RetryOptions.IdempotentRetryOptions);
+
             // Note: we can't use ModifyRequest, as the x-goog-api-client header is added later by ConfigurableMessageHandler.
             // Additionally, that's only called once, and we may want to record the attempt number as well.
             request.AddExecuteInterceptor(InvocationIdInterceptor.Instance);
-            request.AddUnsuccessfulResponseHandler(s_instance);
+            request.AddUnsuccessfulResponseHandler(retryHandler);
+        }
+
+        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request, RetryOptions options)
```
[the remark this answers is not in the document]
jskeet: Please add a comment about the meaning of options being null as an XML comment (so that calling code can hover and see the details).

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -37,46 +36,47 @@ internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
...
+            // Note: we can't use ModifyRequest, as the x-goog-api-client header is added later by ConfigurableMessageHandler.
+            // Additionally, that's only called once, and we may want to record the attempt number as well.
+            request.AddExecuteInterceptor(InvocationIdInterceptor.Instance);
+            request.AddUnsuccessfulResponseHandler(retryHandler);
+        }
 
         public async Task<bool> HandleResponseAsync(HandleUnsuccessfulResponseArgs args)
         {
-            var retry = args.SupportsRetry && await IsRetriableResponse(args.Response).ConfigureAwait(false);
+            var retry = (_retryOptions != null) && args.SupportsRetry && _retryOptions.Predicate.ShouldRetry((int) args.Response.StatusCode);
```
jskeet, line 71: We don't need the nullity test now, as we'll never create an instance of RetryHandler where `_retryOptions` is null.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryOptions.cs
```diff
@@ -0,0 +1,50 @@
...
+    /// <summary>
+    /// This configuration ensures that the API request is never retried on failure.
+    /// </summary>
+    public static RetryOptions Never { get; } = new RetryOptions(RetryTiming.Default, RetryPredicate.Never);
+
+    internal RetryPredicate Predicate { get; }
+    internal RetryTiming Timing { get; }
+
+    internal static RetryOptions MaybeIdempotent(object condition) =>
```
jskeet, line 37: I'd add an XML comment for this, as it's not *hugely* clear how `condition` is going to be used.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryPredicate.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </summary>
+    public static RetryPredicate Never { get; } = RetryPredicate.FromErrorCodes();
+
+    /// <summary>
+    /// Returns a predicate which will retry on the specified HTTP error codes.
+    /// Note that only these error codes will be retried; the default error codes are not used by the returned predicate.
+    /// </summary>
+    /// <param name="errorCodes">Error codes on which to retry.</param>
+    /// <returns>Returns a predicate which retries on the specified error codes.</returns>
+    public static RetryPredicate FromErrorCodes(params int[] errorCodes) => (errorCodes == null) ? RetryPredicate.Never : new(x => errorCodes.Contains(x));
```
jskeet, line 50:
```suggestion
    public static RetryPredicate FromErrorCodes(params int[] errorCodes) => errorCodes is null ? RetryPredicate.Never : new(x => errorCodes.Contains(x));
```

(Prefer `is null` over `== null`, and there's no need for the parentheses.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryPredicate.cs
```diff
@@ -0,0 +1,63 @@
...
+
+    /// <summary>
+    /// Returns a predicate which will retry based on the result of the specified delegate, which is provided with
+    /// the HTTP error code for the response.
+    /// </summary>
+    /// <param name="predicate">Predicate used to determine whether or not to retry. In case of null predicate it will never retry.</param>
+    /// <returns>Returns the retry predicate with the conditions specified for retrying.</returns>
+    public static RetryPredicate FromErrorCodePredicate(Func<int, bool> predicate) => (predicate == null) ? RetryPredicate.Never : new(predicate);
```
jskeet, line 58: Same change as above.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryPredicate.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </summary>
+    /// <param name="errorCodes">Error codes on which to retry.</param>
+    /// <returns>Returns a predicate which retries on the specified error codes.</returns>
+    public static RetryPredicate FromErrorCodes(params int[] errorCodes) => (errorCodes == null) ? RetryPredicate.Never : new(x => errorCodes.Contains(x));
+
+    /// <summary>
+    /// Returns a predicate which will retry based on the result of the specified delegate, which is provided with
+    /// the HTTP error code for the response.
+    /// </summary>
+    /// <param name="predicate">Predicate used to determine whether or not to retry. In case of null predicate it will never retry.</param>
```
jskeet, line 56:
```suggestion
    /// <param name="predicate">Predicate used to determine whether or not to retry. May be null, in which case <see cref="Never" /> will be returned.</param>
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryPredicate.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// <returns>Returns the retry predicate with the conditions specified for retrying.</returns>
+    public static RetryPredicate FromErrorCodePredicate(Func<int, bool> predicate) => (predicate == null) ? RetryPredicate.Never : new(predicate);
+
+    internal bool ShouldRetry(int statusCode) => _predicate.Invoke(statusCode);
+
+    private RetryPredicate(Func<int, bool> predicate) => _predicate = predicate;
```
jskeet, line 62: Please move this to just below Never.

In general our order is:

- Static fields
- Instance fields (readonly then non-readonly)
- Static properties
- Instance properties
- Constructors
- Methods

(Where it makes sense to violate this, that's fine, but that's the normal convention.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTiming.cs
```diff
@@ -0,0 +1,107 @@
...
+    /// The default maximum backoff multiplier to be applied on each iteration. The default is 2.
+    /// </summary>
+    public static double DefaultBackoffMultiplier { get; } = 2;
+
+    /// <summary>
+    /// Returns the retry timing configurations all set to default.
+    /// </summary>
+    public static RetryTiming Default { get; } = new(RetryTiming.DefaultInitialBackoff, RetryTiming.DefaultMaxBackoff, RetryTiming.DefaultBackoffMultiplier);
+
+/// <summary>
```
jskeet, line 60: Please fix the indentation.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTiming.cs
```diff
@@ -0,0 +1,107 @@
...
+
+    internal TimeSpan GetDelay(int failureCount)
+    {
+        if (failureCount < 1)
+        {
+            return TimeSpan.Zero;
+        }
+
+        TimeSpan delay = TimeSpan.FromSeconds(InitialBackoff.TotalSeconds * Math.Pow(BackoffMultiplier, failureCount - 1));
+        return (delay > MaxBackoff) ? MaxBackoff : delay;
```
jskeet, line 105: Nit: remove the parentheses.

jskeet (commented):
Other than the tests we've discussed already, this is looking good - we're basically down to nits around comments and formatting...

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -37,46 +36,47 @@ internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
...
-                500, // Internal server error
-                502, // Bad gateway
-                503, // Service unavailable
-                504 // Gateway timeout
-        };
-        private static readonly RetryHandler s_instance = new RetryHandler();
+        private readonly RetryOptions _retryOptions;
 
-        private RetryHandler() { }
+        internal RetryHandler(RetryOptions retryOptions) => _retryOptions = retryOptions;
```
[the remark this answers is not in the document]
jskeet: I would expect the tests for RetryHandler to use MarkAsRetriable though...

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryTiming.cs
```diff
@@ -0,0 +1,107 @@
...
+    public RetryTiming(TimeSpan initialBackoff, TimeSpan maxBackoff, double backoffMultiplier)
+    {
+        InitialBackoff = initialBackoff >= TimeSpan.Zero ? initialBackoff : throw new ArgumentOutOfRangeException(nameof(initialBackoff), $"Parameter value '{InitialBackoff}' must at least be 1.");
+        MaxBackoff = maxBackoff >= initialBackoff ? maxBackoff : throw new ArgumentOutOfRangeException(nameof(maxBackoff), $"Parameter value '{MaxBackoff}' must be at least as long as initialBackoff.");
+        BackoffMultiplier = backoffMultiplier >= 1 ? backoffMultiplier : throw new ArgumentOutOfRangeException(nameof(backoffMultiplier), $"Parameter value '{BackoffMultiplier}' must at least be 1.");
+    }
+
+    internal TimeSpan GetDelay(int failureCount)
+    {
+        if (failureCount < 1)
```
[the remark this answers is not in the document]
jskeet: > Also this is now used in the unit test cases as the starting point (0th request)

Why would we unit test something that *should* never happen in production? We *should* never call this with a failure count less than 1, right? If we ever do, that's a bug - which is better expressed as an exception IMO.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTests.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        RetryOptions retryOptions = new RetryOptions(
+            retryTiming: new RetryTiming(initialBackoff: TimeSpan.FromSeconds(initialBackoff),
+            maxBackoff: TimeSpan.FromSeconds(maxBackoff), backoffMultiplier: backoffMultiplier),
+            retryPredicate: RetryPredicate.FromErrorCodes(errorCodes));
+
+        Retry(service => service.Buckets.Get("bucket"), client => client.GetBucket("bucket", new GetBucketOptions { RetryOptions = retryOptions }), retryOptions, totalTime, (HttpStatusCode)statusCode);
+    }
+
+    private static void Retry<T>(Func<StorageService, ClientServiceRequest<T>> requestProvider, Action<StorageClient> clientAction,
```
jskeet, line 42: I would expect this to have:

- An array of expected times of requests (first of which would be zero)
- An array of status codes to return
- Nothing optional
- Don't use retryOptions other than to pass to GetBucketOptions
- Don't have the Func or Action - just hard code
- Either another parameter for "should expect eventual failure or success" or just check whether the final response code is 200

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/FakeStorageService.cs
```diff
@@ -55,6 +55,16 @@ public void ExpectRequest<TResponse>(ClientServiceRequest<TResponse> request, Ht
             handler.ExpectRequest(httpRequest.RequestUri, httpRequest.Content?.ReadAsStringAsync()?.Result, responseMessage);
         }
 
+        public void ExpectRequests<TResponse>(ClientServiceRequest<TResponse> request, HttpStatusCode statusCode, int expectedCount)
+        {
+            for (int i = 0; i < expectedCount; i++)
+            {
+                var httpRequest = request.CreateRequest();
```
jskeet, line 62: Just call `ExpectRequest(request, statusCode)` in here? No need to duplicate the code, unless I'm missing something.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -0,0 +1,106 @@
...
+    {
+        RetryOptions retryOptions = new RetryOptions(
+            retryTiming: new RetryTiming(initialBackoff: TimeSpan.FromSeconds(2),
+            maxBackoff: TimeSpan.FromSeconds(6), backoffMultiplier: 2),
+            retryPredicate: RetryPredicate.FromErrorCodes(errorCodes));
+
+        RetryPredicateHelper(retryOptions, statusCode);
+    }
+
+    private static void RetryPredicateHelper(RetryOptions retryOptions, int statusCode)
```
jskeet, line 45: Please rename this to something more descriptive. It's not clear that we need two helper methods at all... I'd expect this to be easily handled by one method accepting:

- A RetryOptions
- The expected backoffs
- The status codes to return
- Whether the eventual result should be success or failure

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTimingTest.cs
```diff
@@ -0,0 +1,67 @@
...
+    [InlineData(2, 20, 3, 2, 6, 18, 20)]
+    [InlineData(0, 1, 2, 0)]
+    [InlineData(1, 2, 2, 1, 2)]
+    [InlineData(1, 4, 1, 1, 1, 1)]
+    [InlineData(8, 40, 1.5, 8, 12, 18, 27, 40, 40)]
+    public void GetDelayVariation(int initialBackoff, int maxBackoff, double backoffMultiplier, params int[] backOffs)
+    {
+        RetryTiming retryTiming = new RetryTiming(initialBackoff: TimeSpan.FromSeconds(initialBackoff), maxBackoff: TimeSpan.FromSeconds(maxBackoff), backoffMultiplier: backoffMultiplier);
+
+        for(int i=0;i<backOffs.Length;i++)
```
jskeet, line 52: Nit: Please add appropriate spaces here and in the call. I'd expect just getting VS to reformat it should work.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTimingTest.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            Assert.Equal(TimeSpan.FromSeconds(backOffs[i]), retryTiming.GetDelay(failureCount: i+1));
+        }
+    }
+
+    [Fact]
+    public void NullTimingCheck()
```
jskeet, line 59: Nit: rename to `NullTiming_EquivalentToDefault`

(The "check" is implicit in the fact that this is a test, but it's good to explain what the expected behavior is.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/TestMessageHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+using System;
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Google.Cloud.Storage.V1.Tests;
+
+internal class TestMessageHandler : ReplayingMessageHandler
```
jskeet, line 25: It's not clear to me that we need a new class here. I'd add the functionality to `ReplayingMessageHandler` - probably not making the assertions directly, but just recording the time of each request. Have a look at StorageClientImplRetryTest for an example of adding extra checking.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/TestMessageHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+    public void SetLastSeenTime()
+    {
+        _lastSeenTime = DateTime.UtcNow;
+    }
+
+    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+    {
+        var currTime = DateTime.UtcNow;
+        double actualBackoff = (currTime - _lastSeenTime).TotalSeconds;
+        Assert.Contains(_expectedTime[_numOfSendAsyncCalls], new List<double>() { Math.Floor(actualBackoff), Math.Ceiling(actualBackoff) });
```
jskeet, line 47: I suspect we can be a bit finer-grained than this. You can use `Assert.Equals(double, double, double)` to provide a tolerance. (250ms should be enough, I'd expect.) We can eventually replace this with a fake clock anyway, of course.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -37,46 +36,50 @@ internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
...
-        // This function is designed to support asynchrony in case we need to examine the response content, but for now we only need the status code
-        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response) => 
-            Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
+        /// <summary>
+        /// It marks the request as retriable with the retry options as provided.
+        /// In case null retry options or Never is passed, retry will not happen in case of failure.
+        /// </summary>
+        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request, RetryOptions options)
+        {
+            if (options is null || options.Predicate == RetryPredicate.Never || options == RetryOptions.Never)
```
jskeet, line 60: We were going to change this to make null invalid here, I believe.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryOptions.cs
jskeet: Please fill in the `<returns></returns>` part, or just remove it.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+    public static RetryOptions Never { get; } = new RetryOptions(RetryTiming.Default, RetryPredicate.Never);
+
+    internal RetryPredicate Predicate { get; }
+    internal RetryTiming Timing { get; }
+
+    /// <summary>
+    /// In case of partially idempotent operations, this method checks the conditions required to mark it as retriable.
+    /// If the condition is available, RetryOptions are set to default idempotent values. If not, they are set to Never.
+    /// </summary>
+    /// <param name="condition">This condition's presence determines whether the operation is retried or not</param>
```
jskeet, line 41: I don't understand this as a description - in particular, there's no mention of null, or what kind of value this is otherwise likely to be. (If we didn't touch this code for 5 years and came back to it having forgotten how it's used, I don't think we'd understand this.)

jskeet (commented):
Very nearly there now.

anshuldavid13:
> @anshuldavid13 is this ready for another review?

@jskeet  Yes please review

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -0,0 +1,89 @@
...
+        service.ExpectRequest(request, new Bucket());
+
+        DateTime startTime = DateTime.UtcNow;
+        client.GetBucket("bucket", new GetBucketOptions { RetryOptions = retryOptions });
+
+        Assert.Equal(expectedBackOffs.Count(), messageHandler.AttemptTimestamps.Count());
+
+        for (int i = 0; i < expectedTotalRetries; i++)
+        {
+            Assert.Equal(expectedBackOffs[i], (messageHandler.AttemptTimestamps[i]-startTime).TotalSeconds, 0.25);
```
jskeet, line 85: Nit: spaces around `-`.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -0,0 +1,89 @@
...
+
+namespace Google.Cloud.Storage.V1.Tests;
+
+/// <summary>
+/// Retries custom RetryTimings and RetryPredicate variants complete flows.
+/// </summary>
+public class RetryTest
+{
+    [Theory]
+    [InlineData(2, 6, 2, 502, 4, 502, 504)]
```
jskeet, line 31: This data looks really wrong for the parameters on the method. It looks like this was copy-pasted from a test where the first few parameters are about timing.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -0,0 +1,89 @@
...
+        service.HttpClient.MessageHandler.NumTries = expectedTotalRetries + 1;
+
+        var request = service.Buckets.Get("bucket");
+        var client = new StorageClientImpl(service);
+        retryOptions ??= RetryOptions.IdempotentRetryOptions;
+
+        // Retries for the the failures.
+        service.ExpectRequests(request, (HttpStatusCode) statusCode, expectedTotalRetries);
+
+        // Last call is a success.
```
jskeet, line 75: You're making it so, yes. I was suggesting before that we should have tests for success cases separately.
We can merge like this, but we're still missing tests IMO. In particular, I don't see any tests that check that we get an exception if the retry predicate doesn't match.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryHandler.cs
```diff
@@ -37,46 +36,55 @@ internal sealed class RetryHandler : IHttpUnsuccessfulResponseHandler
...
-        // This function is designed to support asynchrony in case we need to examine the response content, but for now we only need the status code
-        internal static Task<bool> IsRetriableResponse(HttpResponseMessage response) => 
-            Task.FromResult(s_retriableErrorCodes.Contains(((int)response.StatusCode)));
+        /// <summary>
+        /// It marks the request as retriable with the retry options as provided.
+        /// In case null retry options or Never is passed, retry will not happen in case of failure.
+        /// </summary>
+        internal static void MarkAsRetriable<TResponse>(StorageBaseServiceRequest<TResponse> request, RetryOptions options)
+        {
+            if (options is null)
```
jskeet, line 60: Please use `GaxPreconditions.CheckNotNull(options, nameof(options))` here.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/RetryOptions.cs
```diff
@@ -0,0 +1,56 @@
...
+
+    internal RetryPredicate Predicate { get; }
+    internal RetryTiming Timing { get; }
+
+    /// <summary>
+    /// In case of partially idempotent operations, this method checks the conditions required to mark it as retriable.
+    /// If the condition is available then the operation is retried.
+    /// </summary>
+    /// <param name="condition">This condition's presence determines whether the operation is retried or not.
+    /// If the condition is present, RetryOptions are set to default idempotent values. If its null, they are set to Never. </param>
```
jskeet, line 42:
```suggestion
    /// <param name="condition">
    /// An indicator of whether the operation should be retried or not.
    /// This method returns <see cref="IdempotentRetryOptions" /> if the value is not null, or
    /// <see cref="Never" /> if the value is null.
    /// </param>
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -67,23 +65,31 @@ private static void AssertAttempts(RetryOptions retryOptions, int statusCode, in
...
-        // Last call is a success.
-        service.ExpectRequest(request, new Bucket());
+        if (success)
+        {
+            // Last call is a success.
+            service.ExpectRequest(request, new Bucket());
 
-        DateTime startTime = DateTime.UtcNow;
-        client.GetBucket("bucket", new GetBucketOptions { RetryOptions = retryOptions });
+            DateTime startTime = DateTime.UtcNow;
```
jskeet, line 77: This should be before the `if` block...

jskeet (approved):
One thing to fix, then we can merge. I'll follow up (possibly next week) with a PR to use a fake scheduler for delays, but this will unblock adding the rest of the operations.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -67,23 +65,31 @@ private static void AssertAttempts(RetryOptions retryOptions, int statusCode, in
...
+            client.GetBucket("bucket", new GetBucketOptions { RetryOptions = retryOptions });
 
-        Assert.Equal(expectedBackOffs.Count(), messageHandler.AttemptTimestamps.Count());
+            Assert.Equal(expectedBackOffs.Count(), messageHandler.AttemptTimestamps.Count());
 
-        for (int i = 0; i < expectedTotalRetries; i++)
+            for (int i = 0; i < expectedTotalRetries; i++)
```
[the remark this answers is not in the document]
jskeet: Then we're missing the case of "the code retries a certain number of times but then fails" - either due to different status codes, or due to the limit in Google.Apis. Basically just because we retry doesn't mean we'll eventually succeed.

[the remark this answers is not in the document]
anshuldavid13: Updated the tests

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -35,61 +35,95 @@ public class RetryTest
...
         service.HttpClient.MessageHandler.GoogleApiClientHeader = "test/fake";
-        service.HttpClient.MessageHandler.NumTries = expectedTotalRetries + 1;
+        service.HttpClient.MessageHandler.NumTries = maximumRetries;
 
         var request = service.Buckets.Get("bucket");
         var client = new StorageClientImpl(service);
 
-        // Retries for the the failures.
-        service.ExpectRequests(request, (HttpStatusCode) statusCode, expectedTotalRetries);
+        // Retries for the the failures. Assumed that error code 504 is always included in the predicate for these tests.
```
jskeet, line 100: I would personally have expressed the test with an array of status codes to return, one per request. That way you can end with a success, or end with an error code that *isn't* matched by a predicate. It puts the flexibility in the hands of the caller, keeping this common method simpler.

anshuldavid13: As discussed, we can take up improving the integration tests in a separate PR. Merging the library changes for now.

PR #10122, opened 2023-03-30
chore: Update the backlog

### LINKED ISSUE #9250: StorageClient.UploadObject is difficult to use effectively
[The documentation](https://cloud.google.com/storage/docs/streaming) claims that `StorageClient.UploadObject[Async]` is useful for situations where "You want to upload data but don't know the final size at the start of the upload, such as when generating the upload data from a process, or when compressing an object on-the-fly."  But the API it provides is not suited to either scenario.  Its working is the polar opposite of the standard .NET idiom for generating and storing data: *the API gives you a writeable Stream, and you write the data to it.*

Instead, `StorageClient.UploadObject[Async]` *asks you for a readable Stream*, and then uploads the data from it.  The example given in the linked doc page even involves using `FileStream` to upload an existing file from the file system, which is the precise opposite of generating data in-process where you don't know the final size at the start of the upload.  The scenario of "compressing an object on-the-fly" that the documentation claims to support does not work at all here; compressors such as [GZipStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.gzipstream?view=net-7.0) or [BrotliStream](https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.brotlistream?view=net-7.0) require a writeable stream for their output.

While this API can certainly be useful for pre-existing data, it is clearly not fit for its advertised purpose.  A better API would look like this:

```
        public (Stream uploader, Task<Object> receipt) BeginUploadObject(
            string bucket,
            string objectName,
            string contentType,
            UploadObjectOptions options = null)

(plus a comparable Async variant)
```

This would return a writeable Stream to write data to.  The API will know that the data production is complete when `Close` or `Dispose` is called on the stream.  Generation of the Object metatata that serves as a receipt for your object storage would be the responsibility of an async Task, even for the synchronous API.

Right now, the only way to upload generated data without buffering it either in memory or in a temp file is to come up with some sort of hack to adapt a writeable Stream to a readable Stream.  [`System.IO.Pipelines.Pipe`](https://learn.microsoft.com/en-us/dotnet/standard/io/pipelines) seems to work, but that imposes additional overhead for moving the uploading code onto a thread pool thread.  It would be nice to instead have an API that's actually designed to fit the advertised workflow.

jskeet:
Adds #9250 to the backlog, and removes a couple of items that have been addressed.

Closes #9250

### DIFF
--- a/BACKLOG.md
+++ b/BACKLOG.md
@@ -36,12 +36,6 @@ the request and want to put work into it.
 - Status: No customer requests for this yet; can implement when
   required.
 
-## Storage: Better handling of compressed objects when compression is disabled
-
-- Issue: [#1784](https://github.com/googleapis/google-cloud-dotnet/issues/1784)
-- Issue: [#7154](https://github.com/googleapis/google-cloud-dotnet/issues/7154)
-- Status: All possible options need server changes in some form or other.
-
 ## Diagnostics: Log exceptions thrown in TimedBuffer
 
 - Issue: [#2182](https://github.com/googleapis/google-cloud-dotnet/issues/2182)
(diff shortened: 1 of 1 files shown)


amanda-tarafa (approved):
LGTM

PR #10123, opened 2023-03-30
Preview of removing most pre/mid generation changes

jskeet:
First commit is manual changes; the rest are from running the generator. (I've split it by API, but I regenerated everything.)

The options are available in the head commit of https://github.com/jskeet/googleapis/tree/generator-options
(The commit hash itself may change over time as I try to keep the branch just one ahead of the main head.)


jskeet:
Great :) Will leave this PR here for now as the first commit is handy... we can always close the PR but leave the branch if it looks like it'll get in the way.

jskeet:
This is now all completed.

PR #10184, opened 2023-04-15
feat: Custom Retry Test enhancement

anshuldavid13:


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -42,18 +40,16 @@ public void CustomRetryPredicateTest(int statusCode, bool success, params int[]
...
     }
 
     [Theory]
-    [InlineData(0, 1, 2, 0, 0)]
-    [InlineData(1, 2, 2, 2, 0, 1, 3)]
-    [InlineData(1, 4, 1, 3, 0, 1, 2, 3)]
-    public void CustomRetryTimingTest(int initialBackoff, int maxBackoff, double backoffMultiplier, int numOfFailures, params int[] expectedBackOffs)
+    [InlineData(0, 1, 2, 0, new[] { 200 }, new[] { 0 })] // It succeeds in first attempt with 0 backoff.
+    [InlineData(1, 2, 2, 2, new[] { 502, 504, 200 }, new[] { 0, 1, 3 })] // It retries twice for 502, 504 error code and then succeeds with 200 code
+    public void CustomRetryTimingTest(int initialBackoff, int maxBackoff, double backoffMultiplier, int numOfFaliures, int[] requestStatusCodes, int[] expectedBackOffs)
```
jskeet, line 52: numOfFaliures => failureCount - if we even need this.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -29,11 +28,10 @@ namespace Google.Cloud.Storage.V1.Tests;
...
     [Theory]
-    [InlineData(502, true, 502, 504)]
-    [InlineData(504, true, 502, 504)]
-    [InlineData(429, false, 502, 504)]
-    [InlineData(502, false)]
-    public void CustomRetryPredicateTest(int statusCode, bool success, params int[] errorCodes)
+    [InlineData(502, 200, true, 502, 504)] // It retries for 502 error code as it is present in retry predicate and then succeeds with 200 code.
+    [InlineData(502, 429, false, 502, 504)] // It fails because the secondStatusCode is not present in the retry predicate.
+    [InlineData(502, 200, false)] // It fails because the retry predicate doesnt have any retriable error code.
+    public void CustomRetryPredicateTest(int firstStatusCode, int secondStatusCode, bool success, params int[] errorCodes)
```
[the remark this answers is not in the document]
jskeet: (Although it might be better to just have an array of responseStatusCodes, as for the next test.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
jskeet: And requestStatusCodes => responseStatusCodes

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -81,14 +79,13 @@ public void ExceptionAfterRetryExhaustedTest()
...
-           statusCode: 502,
-           numOfFailures: 4,
+           requestStatusCodes: new[] { 502, 502, 502, 502, 200 },
            maximumRetries: 3,
            success: false,
            expectedBackOffs: expectedBackOffs);
     }
 
-    private static void AssertAttempts(RetryOptions retryOptions, int statusCode, int numOfFailures, int maximumRetries, bool success, params int[] expectedBackOffs)
+    private static void AssertAttempts(RetryOptions retryOptions, int[] requestStatusCodes, int maximumRetries, bool success, params int[] expectedBackOffs)
```
jskeet, line 88: Rename to responseStatusCodes (as that's where the status codes are reported).

jskeet: Do we actually need `maximumRetries`? That's not part of RetryHandler, so I think it would be best to remove it for now.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -101,27 +98,20 @@ private static void AssertAttempts(RetryOptions retryOptions, int statusCode, in
         var client = new StorageClientImpl(service, encryptionKey: null, scheduler);
```
jskeet, line 98: On line 94, you're setting the GoogleApiClientHeader, but I don't think that's actually needed. (It's in other code which tests the value of that header, but I wouldn't expect it to be required here.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -101,27 +98,20 @@ private static void AssertAttempts(RetryOptions retryOptions, int statusCode, in
...
-                service.ExpectRequest(request, new Bucket());
-
                 client.GetBucket("bucket", new GetBucketOptions { RetryOptions = retryOptions });
-
                 Assert.Equal(expectedBackOffs.Count(), messageHandler.AttemptTimestamps.Count());
                 service.Verify();
             }
             else
             {
-                // The call throws an exception
```
jskeet, line 124: We should still be calling service.Verify and asserting the expected backoffs even in the failure scenario.

`success` should *only* be used to determine whether the call eventually succeeds or not.

jskeet: > Tests will fail because in these cases all values of _requestResponses ( from ReplayingMessageHandler) are not dequeued as we are not consuming all loaded Status Codes.

So change the status codes provided by the test.

If you don't expect the client to retry at all, you should only provide a single status code.

We know exactly what the client should do, how many requests it should make, and the time between them. So that's what the mock needs to be "preprogrammed" with, and then we should verify that's exactly what happens.

anshuldavid13: Made the required changes. Kindly verify.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/RetryTest.cs
```diff
@@ -81,50 +77,41 @@ public void ExceptionAfterRetryExhaustedTest()
...
-           statusCode: 502,
-           numOfFailures: 4,
-           maximumRetries: 3,
+           responseStatusCodes: new[] { 502, 502, 502 },
            success: false,
            expectedBackOffs: expectedBackOffs);
     }
 
-    private static void AssertAttempts(RetryOptions retryOptions, int statusCode, int numOfFailures, int maximumRetries, bool success, params int[] expectedBackOffs)
+    private static void AssertAttempts(RetryOptions retryOptions, int[] responseStatusCodes, bool success, params int[] expectedBackOffs)
```
jskeet, line 85: This method would just accept an instance of the new test class that specifies everything that's involved.

There can then be one TheoryData for timing tests, and another for predicate tests - or we could just put everything together, to be honest.

See https://github.com/googleapis/gax-dotnet/blob/daa2e0b8591ac2a3e995f822b526315482ddd832/Google.Api.Gax.Grpc.IntegrationTests/ClientBuilderBaseTest.cs#L245 for an example of code that creates a TheoryData for a custom class.

jskeet (commented):
I think we're still missing the point of these tests in terms of the precision of "this is exactly what I expect and verify".

Let me know if you want me to create a prototype PR with the kind of thing I'm thinking of, as that may be simpler than just using comments.

anshuldavid13:
> Let me know if you want me to create a prototype PR with the kind of thing I'm thinking of, as that may be simpler than just using comments.

Hi Jon, yes it would be helpful to go through your prototype PR to understand exactly how you are expecting these test cases to look like and for me to move in that direction then rather than just trying to fix by each review comment. Will wait for it to proceed on more changes. Thank you

jskeet:
Okay, I'll do that in the next few hours.

jskeet:
See https://github.com/googleapis/google-cloud-dotnet/pull/10227

anshuldavid13:
Closing this PR as these tests are now covered under #10227

jskeet:
> Closing this PR as these tests are now covered under https://github.com/googleapis/google-cloud-dotnet/pull/10227

Just to clarify, #10227 was not attempting to be comprehensive in what's being tested. I would still like to see more actual test cases - the PR was just providing the infrastructure and some *examples* of tests.

PR #10190, opened 2023-04-17
feat: Log trace correlation

### LINKED ISSUE #9948: Consider adding a simple trace correlation mechanism to Google.Cloud.Logging.Console.
This would allow applications already in posession of a collected trace context to specify it on log entries so they are correlated.

Out of scope: any mechanism to create, augment, or collect trace contexts.

To provide an alternative for #9868 use case.

Rishabh-V:
Closes #9948

### DIFF
--- a/apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
+++ b/apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2020 Google LLC
+// Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@ using Microsoft.Extensions.Logging.Abstractions;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Diagnostics;
 using System.IO;
 using System.Text.Json;
 using Xunit;
@@ -25,17 +26,14 @@ namespace Google.Cloud.Logging.Console.Tests;
 
 public class GoogleCloudConsoleFormatterTest
 {
+    // The expected Json string for a simple log entry with no trace information and no scope information, comprising of just message, category and severity.
+    private const string ExpectedSimpleLogEntryJson = "{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\"}\n";
+    
     [Fact]
     public void SimplestLog()
     {
-        var formatter = CreateFormatter();
-        var logEntry = new LogEntry<string>(LogLevel.Information, "LogCategory", new EventId(1), "test", exception: null, (state, exception) => state);
-        var writer = new StringWriter { NewLine = "\n" };
-        formatter.Write(logEntry, scopeProvider: null, writer);
-        var expectedJson = "{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\"}\n";
-        var actualJson = writer.ToString();
-
-        Assert.Equal(expectedJson, actualJson);
+        var actualJson = LogSimpleLogEntry();
+        Assert.Equal(ExpectedSimpleLogEntryJson, actualJson);
     }
 
     [Fact]
@@ -65,18 +63,13 @@ public class GoogleCloudConsoleFormatterTest
     [Fact]
     public void Log_CorrectScopeInformation()
     {
-        var formatter = CreateFormatter(new GoogleCloudConsoleFormatterOptions { IncludeScopes = true });
-
-        var logEntry = new LogEntry<string>(LogLevel.Information, "LogCategory", new EventId(1), "test", exception: null, (state, exception) => state);
-        var writer = new StringWriter { NewLine = "\n" };
-
+        var options = new GoogleCloudConsoleFormatterOptions { IncludeScopes = true };
+        
         var scopeProvider = new LoggerExternalScopeProvider();
-
         scopeProvider.Push("Inner Scope");
-
-        formatter.Write(logEntry, scopeProvider: scopeProvider, writer);
+        
         var expectedJson = "{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\",\"scopes\":[\"Inner Scope\"]}\n";
-        var actualJson = writer.ToString();
+        var actualJson = LogSimpleLogEntry(options, scopeProvider);
 
         Assert.Equal(expectedJson, actualJson);
     }
@@ -84,18 +77,12 @@ public class GoogleCloudConsoleFormatterTest
     [Fact]
     public void Log_EmptyScopeArrayIsIgnored()
     {
-        var formatter = CreateFormatter(new GoogleCloudConsoleFormatterOptions { IncludeScopes = true });
-
-        var logEntry = new LogEntry<string>(LogLevel.Information, "LogCategory", new EventId(1), "test", exception: null, (state, exception) => state);
-        var writer = new StringWriter { NewLine = "\n" };
-
+        var options = new GoogleCloudConsoleFormatterOptions { IncludeScopes = true };
+               
         var scopeProvider = new LoggerExternalScopeProvider();
+        var actualJson = LogSimpleLogEntry(options, scopeProvider);
 
-        formatter.Write(logEntry, scopeProvider: scopeProvider, writer);
-        var expectedJson = "{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\"}\n";
-        var actualJson = writer.ToString();
-
-        Assert.Equal(expectedJson, actualJson);
+        Assert.Equal(ExpectedSimpleLogEntryJson, actualJson);
     }
 
     [Fact]
@@ -152,9 +139,68 @@ public class GoogleCloudConsoleFormatterTest
         Assert.Equal(expectedJsonWithScope, actualJsonWithScope);
     }
 
+    [Fact]
+    public void ConsoleLoggerOptions_NoProjectId_TraceInformationIsNotLogged()
+    {
+        // TraceGoogleCloudProjectId is not set.
+        
+        using var activity = new Activity("my-activity").Start();
+        var actualJson = LogSimpleLogEntry();
+
+        Assert.Equal(ExpectedSimpleLogEntryJson, actualJson);
+    }
+
+    [Fact]
+    public void Log_NoActivity_TraceInformationIsNotLogged()
+    {
+        // Set the TraceGoogleCloudProjectId.
+        var options = new GoogleCloudConsoleFormatterOptions { TraceGoogleCloudProjectId = "my-project-id" };
+
+        // Since we haven't started an activity, Activity.Current will be null.
+        var actualJson = LogSimpleLogEntry(options);
+
+        Assert.Equal(ExpectedSimpleLogEntryJson, actualJson);
+    }
+    
+    [Fact]
+    public void Log_TraceInformationIsLogged()
+    {
+        // Trace information is logged, only if:
+        // 1. TraceGoogleCloudProjectId is set.
+        // 2. Activity.Current is not null; and
+        // 3. Activity.Current.TraceId does not have the default value.
+
+        // Set the TraceGoogleCloudProjectId.
+        string traceProjectId = "my-project-id";
+        var options = new GoogleCloudConsoleFormatterOptions { TraceGoogleCloudProjectId = traceProjectId };
+
+        // Start an activity.
+        using var activity = new Activity("my-activity").Start();
+
+        // JsonWriter writes boolean as JSON literal true or false.
+        // We could use JsonConvert but for tests, direct literal is easy and simpler.
+        var expectedTraceSampled = activity.Recorded ? "true" : "false";
+
+        var expectedJson = $"{{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\",\"logging.googleapis.com/trace\":\"projects/{traceProjectId}/traces/{activity.TraceId}\",\"logging.googleapis.com/spanId\":\"{activity.SpanId}\",\"logging.googleapis.com/trace_sampled\":{expectedTraceSampled}}}\n";
+        var actualJson = LogSimpleLogEntry(options);
+        
+        Assert.Equal(expectedJson, actualJson);
+    }
+
     private static GoogleCloudConsoleFormatter CreateFormatter(GoogleCloudConsoleFormatterOptions options = null)
     {
         options ??= new GoogleCloudConsoleFormatterOptions();
         return new GoogleCloudConsoleFormatter(options);
+    }
+
+    // Logs a simple log entry to a StringWriter and returns the corresponding Json string based on specified GoogleCloudConsoleFormatterOptions and IExternalScopeProvider.
+    private static string LogSimpleLogEntry(GoogleCloudConsoleFormatterOptions options = null, IExternalScopeProvider scopeProvider = null)
+    {
+        // Create a simple log entry and return its Json string based on specified parameters.
+        LogEntry<string> logEntry = new LogEntry<string>(LogLevel.Information, "LogCategory", new EventId(1), "test", exception: null, (state, exception) => state);
+        var writer = new StringWriter { NewLine = "\n" };
+        var formatter = CreateFormatter(options);
+        formatter.Write(logEntry, scopeProvider: scopeProvider, writer);
+        return writer.ToString();
     }
 }
--- a/apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
+++ b/apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2020 Google LLC
+// Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ using Microsoft.Extensions.Logging.Console;
 using Microsoft.Extensions.Options;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
@@ -50,11 +51,19 @@ public sealed class GoogleCloudConsoleFormatter : ConsoleFormatter, IDisposable
     private static readonly JsonEncodedText s_scopesPropertyName = JsonEncodedText.Encode("scopes");
     private static readonly JsonEncodedText s_formatParametersPropertyName = JsonEncodedText.Encode("format_parameters");
 
+    // `trace`, `spanId` and `trace_sampled` are special JSON log fields used for log trace correlation.
+    // Please see https://cloud.google.com/logging/docs/structured-logging#special-payload-fields for more information.
+    private static readonly JsonEncodedText s_tracePropertyName = JsonEncodedText.Encode("logging.googleapis.com/trace");
+    private static readonly JsonEncodedText s_spanIdPropertyName = JsonEncodedText.Encode("logging.googleapis.com/spanId");
+    private static readonly JsonEncodedText s_traceSampledPropertyName = JsonEncodedText.Encode("logging.googleapis.com/trace_sampled");
+
     private static readonly JsonEncodedText s_debugSeverity = JsonEncodedText.Encode("DEBUG");
     private static readonly JsonEncodedText s_infoSeverity = JsonEncodedText.Encode("INFO");
     private static readonly JsonEncodedText s_warningSeverity = JsonEncodedText.Encode("WARNING");
     private static readonly JsonEncodedText s_errorSeverity = JsonEncodedText.Encode("ERROR");
     private static readonly JsonEncodedText s_criticalSeverity = JsonEncodedText.Encode("CRITICAL");
+
+    private static readonly string s_defaultTraceId = default(ActivityTraceId).ToHexString();
 
     private readonly IDisposable _optionsReloadToken;
     private GoogleCloudConsoleFormatterOptions _options;
@@ -100,6 +109,7 @@ public sealed class GoogleCloudConsoleFormatter : ConsoleFormatter, IDisposable
 
                 MaybeWriteFormatParameters(writer, logEntry.State);
                 MaybeWriteScopeInformation(writer, scopeProvider);
+                MaybeWriteTraceInformation(writer);
                 writer.WriteEndObject();
                 writer.Flush();
             }
@@ -175,6 +185,28 @@ public sealed class GoogleCloudConsoleFormatter : ConsoleFormatter, IDisposable
             writer.WriteString(key, ToInvariantString(pair.Value));
         }
         writer.WriteEndObject();
+    }
+
+    private void MaybeWriteTraceInformation(Utf8JsonWriter writer)
+    {
+        if (string.IsNullOrWhiteSpace(_options.TraceGoogleCloudProjectId) || Activity.Current is not Activity activity)
+        {
+            return;
+        }
+
+        var traceId = activity.TraceId.ToHexString();
+
+        // `Equals` implementation in `ActivityTraceId` has a bug. Please see https://github.com/dotnet/runtime/issues/85198 for more details.
+        // String comparision is unfortunate but currently the only available workaround.
+        if (traceId == s_defaultTraceId)
+        {
+            return;
+        }
+
+        var trace = $"projects/{_options.TraceGoogleCloudProjectId}/traces/{traceId}";
+        writer.WriteString(s_tracePropertyName, trace);
+        writer.WriteString(s_spanIdPropertyName, activity.SpanId.ToString());
+        writer.WriteBoolean(s_traceSampledPropertyName, activity.Recorded);
     }
 
     private static JsonEncodedText GetSeverity(LogLevel logLevel) =>
--- a/apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatterOptions.cs
+++ b/apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatterOptions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2020 Google LLC
+// Copyright 2020 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -19,10 +19,6 @@ namespace Google.Cloud.Logging.Console;
 /// <summary>
 /// Options for <see cref="GoogleCloudConsoleFormatter"/>.
 /// </summary>
-/// <remarks>
-/// This class does not currently introduce any new formatter options,
-/// but exists to allow further expansion without disruptive API changes.
-/// </remarks>
 public class GoogleCloudConsoleFormatterOptions : ConsoleFormatterOptions
 {
     // Possible options:
@@ -32,4 +28,16 @@ public class GoogleCloudConsoleFormatterOptions : ConsoleFormatterOptions
 
     // That said, for actual Google Cloud Logging, very few things are really flexible -
     // the property names for state and scopes are the only things that come to mind.
+
+    /// <summary>
+    /// The ID of the Google Cloud Project where trace data is being written to Google Cloud Trace.
+    /// Optional. Set this property to enable Google Cloud Trace and Logging correlation.
+    /// </summary>
+    /// <remarks>
+    /// This property has no effect on where log entries are written to.
+    /// It also has no effect on whether trace information is actually exported to Google Cloud Trace or to which Google Cloud Project traces are exported to.
+    /// When set, this property informs the Console Logger to include trace context information (if any) on the log entries, assuming the trace information is being stored on the specified Google Cloud Project.
+    /// Note that when running your code in Google Cloud, for instance in Google Cloud Run, trace information is automatically collected and exported by the runtime.
+    /// </remarks>
+    public string TraceGoogleCloudProjectId { get; set; }
 }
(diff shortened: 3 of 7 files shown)


Rishabh-V:
This is still a WIP and first draft. I will test this in a GCP Project tomorrow to see log and trace correlation and request review once it is ready for review. Thanks.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,6 +153,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
         Assert.Equal(expectedJsonWithScope, actualJsonWithScope);
     }
 
+    [Fact]
+    public void ConsoleLoggerOptions_EmptyProjectId_TraceInformationIsNotLogged()
```
jskeet, line 157: We're not starting an activity in this test, so presumably it wouldn't have anything to log anyway... I think we should start an activity here - basically the only difference between this and the test that *does* log trace information should be the empty project ID.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,6 +153,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
...
+
+        // Backup the generator, so that it can be reset after the test.
+        // It is null by default but just to keep it easy to understand.
+        var generator = Activity.TraceIdGenerator;
+
+        // Simulate the condition where the generated `TraceId` is equal to the default value of `ActivityTraceId`.
+        Activity.TraceIdGenerator = () => default;
+        using var activity = new Activity("my-activity").Start();
+
+        // Reset the generator, so that other tests aren't impacted.
```
jskeet, line 207: I'd put this in a finally block, to avoid issues if either of the previous lines throws an exception.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,6 +153,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
...
+        // 3. Activity.Current.TraceId does not have the default value.
+
+        // Set the TraceGoogleCloudProjectId.
+        string traceProjectId = "my-project-id";
+        var formatter = CreateFormatter(new GoogleCloudConsoleFormatterOptions { TraceGoogleCloudProjectId = traceProjectId });
+
+        // Start an activity.
+        using var activity = new Activity("my-activity").Start();
+
+        var logEntry = new LogEntry<string>(LogLevel.Information, "LogCategory", new EventId(1), "test", exception: null, (state, exception) => state);
```
jskeet, line 236: Can we extract this "Log a simple entry and return its JSON representation" as a separate method? It seems to be in common in all of these tests (and possibly some others).

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
```diff
@@ -177,6 +185,35 @@ private static void WriteKeyValuePairs(Utf8JsonWriter writer, IEnumerable<KeyVal
...
 
+    private void MaybeWriteTraceInformation(Utf8JsonWriter writer)
+    {
+        if (string.IsNullOrWhiteSpace(_options.TraceGoogleCloudProjectId))
+        {
+            return;
+        }
+
+        var activity = Activity.Current;
+        if (activity == null)
```
jskeet, line 196: Massively minor nit, doesn't even need fixing here - we generally prefer `is null` over `== null` these days.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
```diff
@@ -177,6 +185,35 @@ private static void WriteKeyValuePairs(Utf8JsonWriter writer, IEnumerable<KeyVal
...
+        {
+            return;
+        }
+
+        var traceId = activity.TraceId;
+
+        // The `Equals` method of the `ActivityTraceId` struct uses the _hexString field to compare two ActivityTraceId instances.
+        // The ActivityTraceId struct has a default value with an uninitialized _hexString field.
+        // So, `Equals` method doesn't work as expected. String comparision is unfortunate but currently the only available workaround.
+        if (traceId.ToString() == default(ActivityTraceId).ToString())
```
jskeet, line 206: Given that we're using the string representation just below, let's just call `ToString()` once and put it in a local variable.

Additionally, `default(ActivityTraceId).ToString()` presumably returns the same thing each time - so let's keep that in a static readonly variable, rather than calling `ToString()` each time.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatterOptions.cs
```diff
@@ -32,4 +31,13 @@ public class GoogleCloudConsoleFormatterOptions : ConsoleFormatterOptions
 
     // That said, for actual Google Cloud Logging, very few things are really flexible -
     // the property names for state and scopes are the only things that come to mind.
+
+    /// <summary>
+    /// The ID of the Google Cloud Project where trace data is being written to Google Cloud Trace.
+    /// This property is optional and should be specified if you want to navigate from a log entry to a trace.
```
jskeet, line 37: This sounds like it's purely for navigation - it won't affect whether traces are actually being recorded in Cloud Trace. Is that the case, or does enabling this mean you get tracing (via log entries) with no other work? I haven't been following the twists and turns of what actually creates the trace.

jskeet (commented):
Some initial comments - Amanda has much more context though.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
amanda-tarafa: And with a further tweak you can combine the two ifs:

```suggestion
        if (string.IsNullOrEmpty(_options.TraceGoogleCloudProjectId) || Activity.Current is not Activity activity)
        {
            return;
        }
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
```diff
@@ -177,6 +185,35 @@ private static void WriteKeyValuePairs(Utf8JsonWriter writer, IEnumerable<KeyVal
...
+        if (activity == null)
+        {
+            return;
+        }
+
+        var traceId = activity.TraceId;
+
+        // The `Equals` method of the `ActivityTraceId` struct uses the _hexString field to compare two ActivityTraceId instances.
+        // The ActivityTraceId struct has a default value with an uninitialized _hexString field.
+        // So, `Equals` method doesn't work as expected. String comparision is unfortunate but currently the only available workaround.
```
amanda-tarafa, line 205: I'm not sure what this means, that Equals does not behave as expected.

Regardless, I'm thinking maybe we should just check IsStopped, and we can do that, alongside the other conditions.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,6 +153,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
...
+    }
+
+    [Fact]
+    public void Log_NullActivity_TraceInformationIsNotLogged()
+    {
+        // Set the TraceGoogleCloudProjectId.
+        var formatter = CreateFormatter(new GoogleCloudConsoleFormatterOptions { TraceGoogleCloudProjectId = "my-project-id" });
+
+        // Set Activity.Current to null for clarity. (It is already null here as no Activity is created.)
+        Activity.Current = null;
```
amanda-tarafa, line 180: Let's not set it explicitly, if MS's default behaviour changes, we want to know about it.
```suggestion
        // Since we haven't started an activity, Activity.Current will be null.
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
amanda-tarafa: +1

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,6 +153,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
...
+        formatter.Write(logEntry, scopeProvider: null, writer);
+
+        var expectedJson = "{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\"}\n";
+        var actualJson = writer.ToString();
+
+        Assert.Equal(expectedJson, actualJson);
+    }
+
+    [Fact]
+    public void Log_NullActivity_TraceInformationIsNotLogged()
```
amanda-tarafa, line 174:
```suggestion
    public void Log_NoActivity_TraceInformationIsNotLogged()
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
amanda-tarafa, line 157:
```suggestion
    public void ConsoleLoggerOptions_NoProjectId_TraceInformationIsNotLogged()
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
amanda-tarafa: And if tests are running in parallel, this will be flaky.

But, maybe after my comment on IsStopped, we don't need this test anymore.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,6 +153,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
...
+        var logEntry = new LogEntry<string>(LogLevel.Information, "LogCategory", new EventId(1), "test", exception: null, (state, exception) => state);
+        var writer = new StringWriter { NewLine = "\n" };
+
+        formatter.Write(logEntry, scopeProvider: null, writer);
+
+        // JsonWriter writes boolean as JSON literal true or false.
+        // We could use JsonConvert but for tests, direct literal is easy and simpler.
+        var expectedTraceSampled = activity?.Recorded ?? false ? "true" : "false";
+
+        var expectedJson = $"{{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\",\"logging.googleapis.com/trace\":\"projects/{traceProjectId}/traces/{activity?.TraceId}\",\"logging.googleapis.com/spanId\":\"{activity?.SpanId}\",\"logging.googleapis.com/trace_sampled\":{expectedTraceSampled}}}\n";
```
amanda-tarafa, line 245:
```suggestion
        var expectedJson = $"{{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\",\"logging.googleapis.com/trace\":\"projects/{traceProjectId}/traces/{activity.TraceId}\",\"logging.googleapis.com/spanId\":\"{activity.SpanId}\",\"logging.googleapis.com/trace_sampled\":{expectedTraceSampled}}}\n";
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,6 +153,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
...
+        using var activity = new Activity("my-activity").Start();
+
+        var logEntry = new LogEntry<string>(LogLevel.Information, "LogCategory", new EventId(1), "test", exception: null, (state, exception) => state);
+        var writer = new StringWriter { NewLine = "\n" };
+
+        formatter.Write(logEntry, scopeProvider: null, writer);
+
+        // JsonWriter writes boolean as JSON literal true or false.
+        // We could use JsonConvert but for tests, direct literal is easy and simpler.
+        var expectedTraceSampled = activity?.Recorded ?? false ? "true" : "false";
```
amanda-tarafa, line 243: If the activity can be null at this point, we really need to fix the test.

```suggestion
        var expectedTraceSampled = activity.Recorded ? "true" : "false";
```

amanda-tarafa (changes requested):
A few changes, but overall looking good.

I'd also like to add documention an examples on a second commit on this same PR. See the SampleApp project, I think we can just modify that.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
Rishabh-V: Right. Earlier version of these tests used `ActivitySource` and if source didn't have a listener, it could lead to null activity. Now with the `Activity` constructor and start methods, activity won't be null, so the operator isn't needed. Fixed now.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
Rishabh-V: Done.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
Rishabh-V: Updated the comment with the issue link in `ActivityTraceId`

As discussed, `Activity.IsStopped` is available in .NET 7 and 8. Additionally, I think when IsStopped is called, it sets the `Activity.Current` to its Parent activity which may or may not be null.

Rishabh-V (commented):
Addressed all review comments in the second commit. I will update the sample app and documentation and push it in third commit. Thanks.

Rishabh-V:
I am not sure why breaking changes were detected in this commit. This seems to be unrelated to this PR @jskeet , @amanda-tarafa - Please let me know if this is not correct and I will have a look. Thanks.

amanda-tarafa:
@Rishabh-V I'll wait till you've added sample code to do a full review, just ping me here again.

For the breaking change diff, if you fetch upstream now and rebase your PR on that, it should go away.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -200,22 +177,20 @@ public void Log_DefaultTraceId_TraceInformationIsNotLogged()
...
-        
-        formatter.Write(logEntry, scopeProvider: null, writer);
-
-        var expectedJson = "{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\"}\n";
-        var actualJson = writer.ToString();
-
-        Assert.Equal(expectedJson, actualJson);
+        try
+        {
+            // Simulate the condition where the generated `TraceId` is equal to the default value of `ActivityTraceId`.
```
amanda-tarafa, line 182: I still don't like this. We need to check for the trace Id to be set or not in our production code, but given that it being unset is unlikely to happen (or there would be a bug in Activity) I think I'd prefer not to test this.

If we were to test it, I'd introduce an internal ActivityProvider of sorts that would default to Activity.Current, but that would allow for better testability.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -233,24 +208,33 @@ public void Log_TraceInformationIsLogged()
...
-    
+
     private static GoogleCloudConsoleFormatter CreateFormatter(GoogleCloudConsoleFormatterOptions options = null)
     {
         options ??= new GoogleCloudConsoleFormatterOptions();
         return new GoogleCloudConsoleFormatter(options);
     }
+
+    // Logs a simple log entry to a StringWriter and returns the corresponding Json string based on specified GoogleCloudConsoleFormatter and IExternalScopeProvider.
+    private static string SerializeSimpleLogEntryToJson(GoogleCloudConsoleFormatter formatter = null, IExternalScopeProvider scopeProvider = null)
```
amanda-tarafa, line 228: Is there any place where you are not passing the formatter? If that the case, then don't make formatter an optional parameter. If we ever need it in the future we can do so then.

Or just don't pass it everywhere you are using the formatter from `CreateFormatter`.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,9 +140,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
...
+    }
+
     private static GoogleCloudConsoleFormatter CreateFormatter(GoogleCloudConsoleFormatterOptions options = null)
     {
         options ??= new GoogleCloudConsoleFormatterOptions();
         return new GoogleCloudConsoleFormatter(options);
     }
+
+    // Logs a simple log entry to a StringWriter and returns the corresponding Json string based on specified GoogleCloudConsoleFormatter and IExternalScopeProvider.
+    private static string SerializeSimpleLogEntryToJson(GoogleCloudConsoleFormatter formatter = null, IExternalScopeProvider scopeProvider = null)
```
amanda-tarafa, line 228:
```suggestion
    private static string LogSimpleLogEntry(GoogleCloudConsoleFormatter formatter = null, IExternalScopeProvider scopeProvider = null)
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,9 +140,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
...
+        // Create a simple log entry and return its Json string based on specified parameters.
+        LogEntry<string> logEntry = new LogEntry<string>(LogLevel.Information, "LogCategory", new EventId(1), "test", exception: null, (state, exception) => state);
+        var writer = new StringWriter { NewLine = "\n" };
+        formatter ??= CreateFormatter();
+        formatter.Write(logEntry, scopeProvider: scopeProvider, writer);
+        return writer.ToString();
+    }
+
+    // Returns the expected Json string for a simple log entry with no trace information and no scope information, comprising of just message, category and severity.
+    private static string GetExpectedSimpleLogEntryJson() => "{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\"}\n";
```
amanda-tarafa, line 239:
```suggestion
    private const string ExpectedSimpleLogEntryJson = "{\"message\":\"test\",\"category\":\"LogCategory\",\"severity\":\"INFO\"}\n";
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
```diff
@@ -152,9 +140,101 @@ public void ConsoleLoggerOptions_OptionChange_IsReloaded()
...
+    {
+        // Create a simple log entry and return its Json string based on specified parameters.
+        LogEntry<string> logEntry = new LogEntry<string>(LogLevel.Information, "LogCategory", new EventId(1), "test", exception: null, (state, exception) => state);
+        var writer = new StringWriter { NewLine = "\n" };
+        formatter ??= CreateFormatter();
+        formatter.Write(logEntry, scopeProvider: scopeProvider, writer);
+        return writer.ToString();
+    }
+
+    // Returns the expected Json string for a simple log entry with no trace information and no scope information, comprising of just message, category and severity.
```
amanda-tarafa, line 238:
```suggestion
    // The expected Json string for a simple log entry with no trace information and no scope information, comprising of just message, category and severity.
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatter.cs
```diff
@@ -177,6 +187,29 @@ private static void WriteKeyValuePairs(Utf8JsonWriter writer, IEnumerable<KeyVal
...
 
+    private void MaybeWriteTraceInformation(Utf8JsonWriter writer)
+    {
+        if (string.IsNullOrWhiteSpace(_options.TraceGoogleCloudProjectId) || Activity.Current is not Activity activity)
+        {
+            return;
+        }
+
+        var traceId = activity.TraceId;
+        var traceIdentifier = traceId.ToHexString();
```
amanda-tarafa, line 198:
```suggestion
        var traceId = activity.TraceId.ToHexString();
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatterOptions.cs
```diff
@@ -20,8 +20,7 @@ namespace Google.Cloud.Logging.Console;
 /// Options for <see cref="GoogleCloudConsoleFormatter"/>.
 /// </summary>
 /// <remarks>
-/// This class does not currently introduce any new formatter options,
-/// but exists to allow further expansion without disruptive API changes.
+/// This class exists to allow further expansion without disruptive API changes.
```
amanda-tarafa, line 23: I think this is not necessary anymore. When options are empty it makes sense to explain why, but not we have options, so we can just remove the whole remarks section.

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatterOptions.cs
```diff
@@ -32,4 +31,16 @@ public class GoogleCloudConsoleFormatterOptions : ConsoleFormatterOptions
...
     // That said, for actual Google Cloud Logging, very few things are really flexible -
     // the property names for state and scopes are the only things that come to mind.
+
+    /// <summary>
+    /// The ID of the Google Cloud Project where trace data is being written to Google Cloud Trace.
+    /// Optional. Set this property to enable Google Cloud Trace and Logging correlation.
+    /// </summary>
+    /// <remarks>
+    /// This property has no effect on where log entries are written to.
+    /// It also has no effect on whether trace information is actually exported to Google Cloud Trace or to the Google Cloud Project they are exported to.
```
amanda-tarafa, line 41:
```suggestion
    /// It also has no effect on whether trace information is actually exported to Google Cloud Trace or to which Google Cloud Project traces are exported to.
```

apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console/GoogleCloudConsoleFormatterOptions.cs
```diff
@@ -32,4 +31,16 @@ public class GoogleCloudConsoleFormatterOptions : ConsoleFormatterOptions
...
+    /// When set to true, this property informs the Console Logger to include trace context information (if any) on the log entries, assuming the trace information is being stored on the specified Google Cloud Project.
```
amanda-tarafa, line 42:
```suggestion
    /// When set, this property informs the Console Logger to include trace context information (if any) on the log entries, assuming the trace information is being stored on the specified Google Cloud Project.
```

apis/Google.Cloud.Logging.Console/README.md
```diff
@@ -28,3 +28,12 @@ The server supports four URLs:
 - `/async-exception` throws an exception from an async method (which
   calls other async methods)
 
+## Log trace correlation
+
+To see log trace correlation in action in the SampleApp deployed in Google Cloud Run,
+ensure that the `TraceGoogleCloudProjectId` is set to a valid Google Cloud Project ID
+where trace data is being exported to. This can be done by setting the `TraceGoogleCloudProjectId`
+field in [appsettings.json](SampleApp/appsettings.json) before deployment.
```
amanda-tarafa, line 36: In Cloud Run this is very likely to be the same project where Cloud Run is running. You can check if there's a way to configure Cloud Run to store its own traces on a different project.

But also, this does not need to be deployed in Cloud Run, just in an environment that's configured to export traces, of which Cloud Run is an example.

apis/Google.Cloud.Logging.Console/README.md
```diff
@@ -28,3 +28,12 @@ The server supports four URLs:
 - `/async-exception` throws an exception from an async method (which
   calls other async methods)
 
+## Log trace correlation
```
amanda-tarafa, line 31: I would add a small paragraph saying that setting this library does not persist or start or otherwise manipulates trace information, etc...

apis/Google.Cloud.Logging.Console/SampleApp/Program.cs
```diff
@@ -18,8 +18,19 @@
 
 var builder = WebApplication.CreateBuilder(args);
 
+var config = new ConfigurationBuilder()
```
amanda-tarafa, line 21: I don't think you need to do this: https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-6.0.

Then you use the [builder.Configuration](https://learn.microsoft.com/en-gb/dotnet/api/microsoft.aspnetcore.builder.webapplicationbuilder.configuration?view=aspnetcore-6.0) property to access the config sections you need.

apis/Google.Cloud.Logging.Console/docs/index.md
```diff
@@ -31,6 +31,34 @@ var builder = WebApplication.CreateBuilder(args);
 builder.Logging.AddGoogleCloudConsole(options => options.IncludeScopes = true);
 ```
 
+## Log trace correlation
+
+The console formatter supports Google Cloud Trace and log correlation. This means that
+if the application using the console formatter is running in Google Cloud and exports traces
```
amanda-tarafa, line 37: Or if it's running anywhere else and it's somehow exporting traces to Google Cloud Trace.

apis/Google.Cloud.Logging.Console/docs/index.md
```diff
@@ -31,6 +31,34 @@ var builder = WebApplication.CreateBuilder(args);
...
+to Google Cloud Trace, the trace context information, if available, can be included in the log entry.
+
+To enable log trace correlation, set the `TraceGoogleCloudProjectId` property of the formatter options
+to the ID of the Google Cloud Project where trace data is being written to Google Cloud Trace.
+This property is optional, and if not set, the trace context information will not be
+included in the log entry. It has no effect on where log entries are written to or
+whether trace information is exported to Google Cloud Trace or to the Google Cloud Project they are exported to.
```
amanda-tarafa, line 44: Same change here as in the code comment.

apis/Google.Cloud.Logging.Console/docs/index.md
```diff
@@ -31,6 +31,34 @@ var builder = WebApplication.CreateBuilder(args);
...
+
+Note that when application runs in Google Cloud, for instance in Google Cloud Run,
+trace information is automatically collected and exported by the runtime.
+
+If the application is responsible for initiating the traces, it is application's responsibility to ensure
+that the traces are exported in the Google Cloud Trace, in the same Google Cloud Project as specified in the console formatter options.
```
amanda-tarafa, line 50:
```suggestion
that the traces are exported to Google Cloud Trace, in the same Google Cloud Project as specified in the console formatter options.
```

apis/Google.Cloud.Logging.Console/docs/index.md
```diff
@@ -31,6 +31,34 @@ var builder = WebApplication.CreateBuilder(args);
...
+
+To enable log trace correlation, the `TraceGoogleCloudProjectId` property of the options should be set to a valid Google Cloud Project ID
```
amanda-tarafa, line 52: There's a lot of repetition in these docs, this paragrah is almost identialca to the one starting in line 40.

amanda-tarafa (changes requested):

Rishabh-V (commented):
Review comments addressed/responded. PTAL

apis/Google.Cloud.Logging.Console/README.md
amanda-tarafa: Yes, I though this was the only docs when I first reviewed and only realized later that it was just the Sample docs.

Let's get the actual documentation clean and clear and then we can work on this one, I agree we can add a small explanation and the point to the actual documentation.

apis/Google.Cloud.Logging.Console/README.md
amanda-tarafa: Ah OK Cloud Run is fine then if this was meant for it anyways. Same applies here, let's get the actual docs clean and clear, and them we'll word this one.

Rishabh-V:
> LGTM, @jskeet will propose doc changes, if any, by raising a PR, see [#10190 (comment)](https://github.com/googleapis/google-cloud-dotnet/pull/10190#issuecomment-1524939375)

Thanks Jon and Amanda. I will squash into one commit and merge.

PR #10203, opened 2023-04-19
tests: Use a fake scheduler for retry tests

jskeet:
This reduces the time for StorageClientImplRetryTest from about a minute to about 5 seconds. (Even with a fake scheduler, it doesn't complete instantly, and there are a lot of tests.)


amanda-tarafa (approved):
LGTM

PR #10225, opened 2023-04-21
chore: Remove specs from namespace references in docfx

jskeet:
We may be able to remove this later; it's effectively a temporary workaround for https://github.com/dotnet/docfx/issues/8660


amanda-tarafa (approved):
LGTM

PR #10227, opened 2023-04-21
test: First pass at a custom-test-case class for retry tests

jskeet:
Using a custom class instead of just method parameters specified in attributes allows all of this to be much more explicit, and specify the whole RetryOptions in one place.

This PR currently has a single TheoryData and a single test - I expect we *can* cover everything we want to do that way, but also we *could* break it down into one TheoryData+test for custom predicates, one for custom timing etc. It won't matter much in the long run.


jskeet:
In terms of what to do with this: we can merge this as-is, but it doesn't attempt to cover all the tests that we currently have - we'd need a follow-up PR to finish that. Alternatively, you can use this to start a new PR.

anshuldavid13:
Hi Jon, this looks good. We can merge this PR and I can add another PR on top of this to add any more tests as needed. I will close my other PR in the meantime as we can go ahead from here now.

PR #10269, opened 2023-05-02
chore: Make the diff action fail if OwlBot still has work to do

jskeet:


amanda-tarafa (approved):
Yay!

jskeet:
(The second commit should break this PR - simulating what happens initially with OwlBot. I'll remove that once I've seen it fail appropriately.)

PR #10292, opened 2023-05-03
chore: Use local googleapis repo before trying GitHub for commit messages

jskeet:
This reduces the likelihood of us running into rate-limiting issues, without requiring the local googleapis repo to be *completely* up-to-date.

(This is only used during the release process. Creating a new Repository each time we want a commit isn't ideal, but it avoids adding state here.)


amanda-tarafa (approved):
Oh nice!

Issue #10132, opened 2023-04-05
Specify useInt64Timestamp in BigQuery


jskeet:
We closed #4031 at the time as we were doing the best we could with the API as it was exposed.

We should now be able to request timestamps as 64-bit integers instead of `double` values, which means we should be able to preserve microsecond precision. See https://cloud.google.com/bigquery/docs/reference/rest/v2/DataFormatOptions.

We should do this for both formatting and parsing, assuming it applies to both.

jskeet:
I've got a little bit of time available now, and having been mostly writing docs all week I'd like to get some hands-on coding... so I'll pick this up unless you object, @Rishabh-V.

Rishabh-V:
Sure Jon. Please go for it. Thank you.

jskeet:
Looks like the flag isn't actually present in the Discovery doc, which means it's not in our generated code, so we can't easily use it.

Internal bug reference: b/128710166

I got as far as writing a failing test before finding that we basically can't use it yet... I'll add that (as a skipped test).

PR #10301, opened 2023-05-04
test: Add a test for #10132 (and skip it for now)

jskeet:
We can't fix it yet due to the required option not being in the Discovery doc, but at least if we have a test, it'll be easy to make progress later.

Once this is merged, I'll close the issue for now.


Rishabh-V (approved):
LGTM

PR #10307, opened 2023-05-05
feat: Dispose Channel in shutdown

Rishabh-V:
This has 2 options.

Please review one commit at a time. 

Whichever looks fine, I can implement the same in `PublisherClientBuilder`. PTAL


jskeet:
Neither of these actually calls Dispose - they call Shutdown. Additionally, I don't know offhand whether Grpc.Core handles Shutdown appropriately - I'd rather keep calling ShutdownAsync as well

jskeet:
I tend to use a rule of 3 - I'm okay with code appearing twice, but at 3 times I want to centralise. I'm also hoping that we can revisit this and centralise in GAX for the next major version of that, at which point this code will go away :)

Rishabh-V:
> I tend to use a rule of 3 - I'm okay with code appearing twice, but at 3 times I want to centralise. I'm also hoping that we can revisit this and centralise in GAX for the next major version of that, at which point this code will go away :)

Okay. Good to know this. I will then update the code, test it out and raise a new commit by evening. Thanks.

Rishabh-V:
Hi @jskeet, I have tested this in Windows using TCPView. Earlier, the connection used to stay alive for a few seconds after the client shutdown. Now, the connection is closed as soon as client is shutdown.  PTAL. Thanks.

jskeet (approved):
LGTM, although I guess it would... :)

Happy to do a release for this later today.

jeffijoe:
> Earlier, the connection used to stay alive for a few seconds after the client shutdown.

Just a few seconds? For me it would be a lot longer, I would think 2 minutes since that's the default HTTP client idle timeout IIRC?

Rishabh-V:
> > Earlier, the connection used to stay alive for a few seconds after the client shutdown.
> 
> Just a few seconds? For me it would be a lot longer, I would think 2 minutes since that's the default HTTP client idle timeout IIRC?

Right. I added _a few seconds_ deliberately as that is what I observed consistently. I am not sure at the moment if it has something to do with my test application or me running the application on Windows, but 4-5 times I tried to reproduce the issue, the TCPView showed connection in established state only for a few seconds after shutdown and I had to grab the screenshot quickly before the connection was lost. 

I am done for the day (I work in IST timezone). I can probably check it again on coming Monday morning as it will be quick to check and report back if I observe any changes. Thanks.

jskeet:
@jeffijoe: I'm hoping to do a release of this in about 4 hours, so hopefu

PR #10352, opened 2023-05-12
chore: Fix Logging.V2 generation by renaming two RPCs

jskeet:
The names chosen are temporary (but not awful).


jskeet:
Once this is merged, I'll fetch #10312 locally, rebase and force push, then get OwlBot to regenerate. That should then generate a reasonable new API surface.

amanda-tarafa (approved):
LGTM

PR #10354, opened 2023-05-12
chore: Reset Google.Cloud.Batch.V1 to 2.0.0

jskeet:
The change in 2.1.0 was reverted, and the release (coincidentally) failed at the same time. We'll just go back to "this didn't happen".


amanda-tarafa (approved):
Did we tagged this one?
Not sure if the tag can de deleted?

jskeet:
Nope, it didn't end up being tagged, fortunately.

Issue #10318, opened 2023-05-08
Deadlock when disposing underlying gRPC channel in `SubscriberClient`


jeffijoe:
This is a follow-up of #10304 — now that clients are being disposed, a deadlock caused by what appears to be a race condition between cancelling a gRPC streaming call and disposing the channel has started to manifest.

I am not sure whether to report it here or to `grpc-dotnet`, but since it's occurring due to how the channels are used by this library, I figured I would start here.

This is difficult to reproduce reliably, but for my integration tests it happens quite often; I am still unable to run all my integration tests as mentioned in the linked issue, but now due to the teardown hanging because of this race condition.

#### Environment details

  - OS: macOS Ventura
  - .NET version: 7
  - Package name and version: Google.Cloud.PubSub.V1 @ 3.5.1

#### Steps to reproduce

Run the following application with a debugger attached, the latest version of `Google.Cloud.PubSub.V1` installed, and the emulator running on `localhost:8085`. When no more output is produced, pause the debugger and inspect the thread stacks.

```csharp
using Google.Api.Gax;
using Google.Cloud.PubSub.V1;
using Grpc.Core;

/*
 * Run this application with a debugger attached. When the output stops producing,
 * usually ending with  "Received message" and "Stopping subscriber", pause execution and inspect
 * the thread stacks. You will see the "cancelinator" thread being stuck
 * at `GrpcChannel.FinishActiveCall() -> Monitor.Enter`, and one of the .NET Worker Threads being
 * stuck at `GrpcCall.Cleanup() -> CancellationTokenSource.Registrations.WaitForCallbackToComplete`
 *
 * Be patient, it may take >1000 iterations to manifest depending on your setup. I don't know of
 * a more reliable way to reproduce this.
 */

// Emulator.
const string endpoint = "http://localhost:8085";

// Run ID, used to ensure unique topics and subs.
var runId = DateTimeOffset.UtcNow.ToUnixTimeSeconds();
var i = 1;
while (true)
{
    var iteration = i++;
    // Create the topic and subscription.
    var topic = await CreateTopic(iteration);
    var subscription = await CreateSubscription(topic);

    // Publish a message to the topic.
    var messageData = await PublishMessage(topic, iteration);

    // Set up a subscriber and wait for the message to be received.
    // That's how we know we have an open streaming call.
    var subscriber = await new SubscriberClientBuilder
    {
        Endpoint = endpoint,
[...]

PR #10357, opened 2023-05-15
fix: Temporary patch for PubSub Subscriber deadlock issue.

Rishabh-V:
This is a temporary patch for #10318 while we wait for the fix of https://github.com/grpc/grpc-dotnet/issues/2119

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
@@ -547,7 +547,7 @@ namespace Google.Cloud.PubSub.V1.Tests
                     Assert.Empty(fake.Subscribers[0].Nacks);
                     Assert.Empty(fake.Subscribers[0].Extends);
                     Assert.Equal(new[] { fake.Time0 + TimeSpan.FromSeconds(1) }, fake.Subscribers[0].WriteCompletes);
-                    Assert.Equal(new[] { fake.Time0 + TimeSpan.FromSeconds(1) }, fake.ClientShutdowns);
+                    Assert.Equal(new[] { fake.Time0 + TimeSpan.FromSeconds(3) }, fake.ClientShutdowns);
                 });
             }
         }
@@ -581,7 +581,7 @@ namespace Google.Cloud.PubSub.V1.Tests
                     Assert.Empty(fake.Subscribers[0].Nacks);
                     Assert.Empty(fake.Subscribers[0].Extends);
                     Assert.Equal(new[] { fake.Time0 + TimeSpan.FromSeconds(1) }, fake.Subscribers[0].WriteCompletes);
-                    Assert.Equal(new[] { fake.Time0 + TimeSpan.FromSeconds(1) }, fake.ClientShutdowns);
+                    Assert.Equal(new[] { fake.Time0 + TimeSpan.FromSeconds(3) }, fake.ClientShutdowns);
                 });
             }
         }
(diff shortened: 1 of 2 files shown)


jskeet (approved):
Looks fine to me, but please wait for @amanda-tarafa to review as well, as she has more context.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -903,7 +903,7 @@ public void LeaseMaxExtension()
                     var doneTask = fake.Subscriber.StartAsync(async (msg, ct) =>
                     {
                         // Emulate a hanging message-processing task.
-                        await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromHours(24), ct));
+                        await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromHours(23), ct));
```
amanda-tarafa, line 906: I don't understand this change.

amanda-tarafa: And will this test fail back when we remove the delay. If not, add the TODO here as well.

jeffijoe:
The grpc-dotnet PR was merged, how often do they do patch releases? Maybe we can wait for it.

amanda-tarafa:
> The grpc-dotnet PR was merged, how often do they do patch releases? Maybe we can wait for it.

I've asked https://github.com/grpc/grpc-dotnet/issues/2119#issuecomment-1548704542. Let's see how long it takes and if that works for you, then yes, we would also prefer to not release this patch. Note that after Grpc.Net.Client has been released, it will take us maybe a couple of days to trickle the dependency update and do a release of Google.Cloud.PubSub.V1, but you'd still be able to force the newer Grpc.Net.Client dependency though by including it explicitly.

amanda-tarafa:
@Rishabh-V hold off on merging until we have heard back from Grpc.Net.Client and know whether @jeffijoe is OK with waiting or not. (+ review comments :) )

jeffijoe:
"Couple of weeks" sounds like it could be anything from 2-8 weeks, so I'd still appreciate the temporary patch 😄

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
Rishabh-V: I had thought about putting a TODO here but didn't as the behavior of this test isn't altered. Here is why:

- This test emulates a hanging subscriber client, by adding a delay of 24 hours before acknowledging the message.
- After simulated 12 hours, it stops the client. So message is never acknowledged and the client is stopped. The test checks that the lease extensions happened only for the 60 minutes (DefaultMaxTotalAckExtension is 60 minutes)
- The `TestScheduler` used to execute these tests have a [simulated timeout of 24 hours](https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/Tasks/TestScheduler.cs#L503). The scheduler advances the clock based on delays and checks that [timeout isn't reached](https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/Tasks/TestScheduler.cs#L547)
- After adding this 2 second delay before the channel shutdown, the simulated test execution time becomes 24 hours and 2 seconds, which is greater than the stimulated timeout of 24 hours so this test was getting the timeout error. 
- Since the client is shutdown after simulated 12 hours, any value of delay in acking the message greater than 12 hours (and less than simulated timeout) will work just fine as long as test doesn't times out.

So, there is no change in behavior of the test by changing the delay from 24 hours to 23 hours before acknowledging the message as the client stops after 12 hours. 

The test will not fail after this 2 second delay in shutdown is reverted, but it makes no difference.

Let me know if you would still like to add a TODO and I can do that. Thanks.

amanda-tarafa (approved):

PR #10371, opened 2023-05-17
fix: Correct the auto-extend lease interval for exactly-once delivery.

Rishabh-V:
Fixes the issue of duplicate message delivery in exactly once delivery due to incorrect auto-extend interval which prevents the message from being auto-extended.

I am creating a document which explains the root cause of issues and the proposed fix. I should be able to share it by the end of my day today. Thanks.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
@@ -862,15 +862,15 @@ namespace Google.Cloud.PubSub.V1.Tests
             }
         }
 
-        [Fact]
-        public void LeaseExtension()
+        [Theory, CombinatorialData]
+        public void LeaseExtension(bool isExactlyOnceDelivery)
         {
             var msgs = new[] { new[] {
                 ServerAction.Data(TimeSpan.Zero, new[] { "1" }),
                 ServerAction.Data(TimeSpan.FromSeconds(5), new[] { "2" }),
                 ServerAction.Inf()
             } };
-            using (var fake = Fake.Create(msgs, ackDeadline: TimeSpan.FromSeconds(30), ackExtendWindow: TimeSpan.FromSeconds(10)))
+            using (var fake = Fake.Create(msgs, isExactlyOnceDelivery: isExactlyOnceDelivery, ackDeadline: TimeSpan.FromSeconds(30), ackExtendWindow: TimeSpan.FromSeconds(10)))
             {
                 fake.Scheduler.Run(async () =>
                 {
@@ -890,14 +890,14 @@ namespace Google.Cloud.PubSub.V1.Tests
             }
         }
 
-        [Fact]
-        public void LeaseMaxExtension()
+        [Theory, CombinatorialData]
+        public void LeaseMaxExtension(bool isExactlyOnceDelivery)
         {
             var msgs = new[] { new[] {
                 ServerAction.Data(TimeSpan.Zero, new[] { "1" }),
                 ServerAction.Inf()
             } };
-            using (var fake = Fake.Create(msgs, ackDeadline: TimeSpan.FromSeconds(30), ackExtendWindow: TimeSpan.FromSeconds(10)))
+            using (var fake = Fake.Create(msgs, isExactlyOnceDelivery: isExactlyOnceDelivery, ackDeadline: TimeSpan.FromSeconds(30), ackExtendWindow: TimeSpan.FromSeconds(10)))
             {
                 fake.Scheduler.Run(async () =>
                 {
@@ -913,7 +913,7 @@ namespace Google.Cloud.PubSub.V1.Tests
                     Assert.Equal(1, fake.Subscribers.Count);
                     // Check that the lease was extended for 60 minutes only.
                     // +1 is due to initial lease extension at time=0
-                    Assert.Equal((int)SubscriberClient.DefaultMaxTotalAckExtension.TotalSeconds / 20 + 1, fake.Subscribers[0].Extends.Count);
+                    Assert.Equal((int) SubscriberClient.DefaultMaxTotalAckExtension.TotalSeconds / 20 + 1, fake.Subscribers[0].Extends.Count);
                 });
             }
         }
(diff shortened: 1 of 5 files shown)


Rishabh-V:
I will request review after the document is done and shared. Thanks.

Rishabh-V:
Hi @jskeet, @amanda-tarafa, I have updated this PR to reflect approach#3 from the design document. PTAL and share your feedback. Thanks.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -862,15 +862,15 @@ public void OnlyOneStart()
             }
         }
 
-        [Fact]
-        public void LeaseExtension()
+        [Theory, CombinatorialData]
+        public void LeaseExtension([CombinatorialValues(true, false)] bool isExactlyOnceDelivery)
```
amanda-tarafa, line 866: You don't need this, the test is already marked as CombinatorialData, so bools and I believe enums, etc, it will combine the values automatically.

```suggestion
        public void LeaseExtension(bool isExactlyOnceDelivery)
```

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -884,20 +884,23 @@ public void LeaseExtension()
                     await fake.TaskHelper.ConfigureAwait(doneTask);
                     Assert.Equal(1, fake.Subscribers.Count);
                     DateTime S(int seconds) => fake.Time0 + TimeSpan.FromSeconds(seconds);
-                    Assert.Equal(new[] { S(0), S(5), S(20), S(25), S(40), S(45), S(60), S(65) }, fake.Subscribers[0].Extends.Select(x => x.Time));
+                    var expectedExtends =
+                        isExactlyOnceDelivery ? new[] { S(0), S(5), S(10), S(15), S(20), S(25), S(30), S(35), S(40), S(45), S(50), S(55), S(60), S(65) }
```
amanda-tarafa, line 888: I think I'd prefer this test to be split in two distinct tests. Theories are not a way to reuse setup code, theories are a way to prove to a whole set of inputs (a theory) behave in the same way.

And the same for the rest of the modifications below.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -1109,6 +1149,48 @@ public void SilentlyFixedParameters()
             Assert.Equal(TimeSpan.FromSeconds(59), sub4.AutoExtendInterval);
         }
 
+        [Fact]
+        public void ExactlyOnceDelivery_SilentlyFixedParameters()
```
amanda-tarafa, line 1153: Hmm, I think this test should be split in plenty (and same for the one above from where you copied this one, sorry).

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -62,6 +62,15 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
         _modifyDeadlineSeconds = (int)((settings.AckDeadline ?? DefaultAckDeadline).TotalSeconds);
```
amanda-tarafa, line 62: I'm wondering how big of a change would it be to rename _modifyDeadlineSeconds to _ackDeadlineSeconds.
and honestly a lot of the other names here are somewhat confusing, for instance autoExtendInterval would be clearer if it was called delayBeforeAutoExtend (as it is really the delay before )
Obviously if we were to make some of those changes, they would be on their own commits or even on a separate PR, but I think renaming some things here would help a lot.
@jskeet for thoughts.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -62,6 +62,15 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
...
         var autoExtendInterval = TimeSpan.FromSeconds(_modifyDeadlineSeconds) - (settings.AckExtensionWindow ?? DefaultAckExtensionWindow);
         var autoExtendIntervalForExactlyOnceDelivery = settings.AckExtensionWindow ?? MinimumAckExtensionWindowForExactlyOnceDelivery;
```
amanda-tarafa, line 64: These two here have a different meaning, so their names definetely shouldn't be similar (and maybe other things shouldn't be similar). If these two values are being used for the same purpose elsewhere, that's certainly going to yield wrong results.

autoExtendInterval => is the delay before we attempt to extende the lease.
autoExtendIntervalForExactlyOnceDelivery => is the window we have for extending the lease.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -62,6 +62,15 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
...
+        // The value of 15 seconds ensures that we get ample time before the Ack deadline expires and that we don't send too frequent lease extensions.
+        var preferredExtendInterval = TimeSpan.FromSeconds(_modifyDeadlineSeconds - 15);
```
amanda-tarafa, line 66: And now
preferredExtendInterval => is a delay before we can attempt to extend the lease.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -62,6 +62,15 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
...
+        if (autoExtendIntervalForExactlyOnceDelivery > preferredExtendInterval)
```
amanda-tarafa, line 67: And this is comparing pears with apples:

autoExtendIntervelForExactlyOnceDelivery => window for extending the lease.
preferredExtendInterval => delay before we can attempt to extend the lease.

amanda-tarafa: And line 78 is also going to be wrong at least some of the time, i.e when autoExtendIntervelForExactlyOnceDelivery retains it's original value instead of being assigned preferredExtendInterval. In that case Math.Max value will be comparing a "windows for extending the lease" with a "delay before we can attempt to extend the lease", and those are complementary values not equivalents.

I wonder where else we have these same mismatches, and how these values are actually used.

amanda-tarafa (changes requested):
I think there are some errors in this PR that are also present in the original implementaiton of exactly-once delivery. I'm not 100% certain though, as my understanding of this code is limited. I've left comments trying to explain.
@Rishabh-V I think it might be better to leave the PR as is until @jskeet can take a look tomorrow and sanity check my reasoning. In case I'm wrong you don't have to make changes.
@jskeet If you can sanity check my comments on production code, that'd be great.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
Rishabh-V: While I agree that names can be better, I just thought, I will add more context here:
autoExtendInterval -> It's the interval after which lease of unhandled messages is automatically extended for non-exactly once subscriptions
autoExtendIntervalForExactlyOnceDelivery -> Is the same, but for exactly-once delivery subscriptions. It is just that the interval calculation comes from AckExtensionWindow.

jskeet: No, it's not the same. `settings.AckExtensionWindow` is "How long we're meant to allow before the deadline".

Put it this way: if someone increases `settings.AckExtensionWindow`, then in non-EOD that means we extend the lease earlier; in EOD that means we extend the lease later. That's not right!

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
Rishabh-V: Now the test is common for eod and regular subs.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
Rishabh-V: Renamed them as _ackDeadlineSeconds and _autoExtendDelay.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
Rishabh-V: Thanks Jon and Amanda. This is now addressed.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
Rishabh-V: Thanks Amanda. This is now addressed.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -1070,43 +1070,72 @@ public void ValidParameters()
...
         [Fact]
-        public void SilentlyFixedParameters()
+        public void MinimumDelayIsUsedWhenMinimumAckDeadlineIsSpecified()
         {
             var subscriptionName = new SubscriptionName("project", "subscriptionId");
             var clients = new[] { new FakeEmptySubscriberServiceApiClient() };
 
-            // Test MinimumLeaseExtensionDelay is honoured when minimum ack-deadline used.
-            var settingsAckExtension1 = new SubscriberClient.Settings
+            // Test that MinimumLeaseExtensionDelay is honoured when minimum ack-deadline is specified.
```
jskeet, line 1078: Given the confusion we've had, let's rename this to "when the ack-deadline is specified with the minimum possible value".

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -76,6 +76,11 @@ public enum Reply
     /// </summary>
     public static TimeSpan DefaultAckDeadline { get; } = TimeSpan.FromSeconds(60);
 
+    /// <summary>
+    /// The default message Acknowledgement deadline of 60 seconds for exactly-once delivery subscriptions.
```
jskeet, line 80: Let's work out whether we want it to be "Acknowledgement", "ACKnolwedgement" or "acknowledgement" - in lines 80, 85 and 90 we've got all three...

jskeet (approved):
A few nits, and one genuine bug (I think) - I'm not sure why the bug wasn't seen in tests; I'm happy for you to make the decision about whether to add a new test to catch it before fixing. (Throttling may be hard to test.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -1070,43 +1070,72 @@ public void ValidParameters()
...
+            var settingsAckExtension = new SubscriberClient.Settings
```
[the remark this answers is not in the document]
Rishabh-V: done. Here and in below tests as well. Thanks.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Made it consistent with "acknowledgement" across the public API docs.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -612,10 +616,10 @@ private void HandleExtendLease(HashSet<string> msgIds, LeaseCancellation cancell
                     _eventPush.Set();
                     // Some ids still exist, schedule another extension.
                     // The overall `_maxExtensionDuration` is maintained by passing through the existing `cancellation`.
-                    Add(_scheduler.Delay(_exactlyOnceDeliveryEnabled ? _autoExtendIntervalForExactlyOnceDelivery : _autoExtendInterval, _softStopCts.Token), Next(false, () => HandleExtendLease(msgIds, cancellation)));
+                    Add(_scheduler.Delay(_exactlyOnceDeliveryEnabled ? _autoExtendDelayForExactlyOnceDelivery : _autoExtendDelay, _softStopCts.Token), Next(false, () => HandleExtendLease(msgIds, cancellation)));
                     // Increment _extendThrottles.
                     _extendThrottleHigh += 1;
-                    Add(_scheduler.Delay(_extendQueueThrottleInterval, _softStopCts.Token), Next(false, () => _extendThrottleLow += 1));
+                    Add(_scheduler.Delay(_exactlyOnceDeliveryEnabled ? _extendQueueThrottleInterval : _extendQueueThrottleIntervalForExactlyOnceDelivery, _softStopCts.Token), Next(false, () => _extendThrottleLow += 1));
```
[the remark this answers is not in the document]
Rishabh-V: Ah. Thanks. Fixed now.

Rishabh-V (commented):
Thanks Jon. Added two more commits. First commit has the code review comments addressed and second one makes "acknowledgement" consistent across the Public API docs. PTAL. Thanks.

jskeet (approved):

jskeet:
Merging as I believe all the comments have been addressed. We should wait until @amanda-tarafa is back before *releasing* though.

amanda-tarafa (commented):
I've left a couple of comments, but around docs. I'll check for those specifically in #10463 as things may have changed.

PR #10383, opened 2023-05-18
feat: Support `dataBoostEnabled` field in partitioned Read/ExecuteSql

anuragsrivstv:
- Adds a new overload of public method SpannerCommand.GetReaderPartitionsAsync
- Adds a new class `PartitionOptions` to be taken as a mandatory parameter in the new overloaded method.
- Updates existing and adds new unit tests.
- Updates existing and adds new integration tests.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// Copyright 2017 Google Inc. All Rights Reserved.
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Cloud.Spanner.Data.CommonTesting;
+using System;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -29,42 +31,72 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
         public PartitionedReadTests(PartitionedReadTableFixture fixture) =>
             _fixture = fixture;
 
-        [CombinatorialData]
-        [Theory]
-        public async Task DistributedReadAsync(bool query)
+        [Theory, CombinatorialData]
+        [Trait(Constants.SupportedOnEmulator, Constants.No)]
+        public async Task DistributedReadAsync(bool query, bool dataBoostEnabled)
         {
-            int numRows;
-            using (var connection = _fixture.GetConnection())
+            int numRows = await GetNumberOfRows();
+
+            using var connection = new SpannerConnection(_fixture.ConnectionString);
+            await connection.OpenAsync();
+
+            using var transaction = await connection.BeginReadOnlyTransactionAsync();
+            using var cmd = query
+                ? connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName}")
+                : connection.CreateReadCommand(_fixture.TableName, ReadOptions.FromColumns(_fixture.ColumnNames), KeySet.All);
+            transaction.DisposeBehavior = DisposeBehavior.CloseResources;
+            cmd.Transaction = transaction;
+            var partitions = await cmd.GetReaderPartitionsAsync(PartitionOptions.Default.WithPartitionSizeBytes(100).WithDataBoostEnabled(dataBoostEnabled));
+            var transactionId = transaction.TransactionId;
+
+            if (query)
             {
-                using (var cmd = connection.CreateSelectCommand($"SELECT COUNT(*) FROM {_fixture.TableName}"))
-                {
-                    numRows = await cmd.ExecuteScalarAsync<int>();
-                }
+                Assert.Equal(dataBoostEnabled, partitions.FirstOrDefault().Request.ExecuteSqlRequest.DataBoostEnabled);
+            }
+            else
+            {
+                Assert.Equal(dataBoostEnabled, partitions.FirstOrDefault().Request.ReadRequest.DataBoostEnabled);
             }
 
-            using (var connection = new SpannerConnection(_fixture.ConnectionString))
-            {
-                await connection.OpenAsync();
+            await Task.WhenAll(partitions.Select(
+                    partitions => DistributedReadWorkerAsync(partitions, transactionId)))
+                .ConfigureAwait(false);
 
-                using (var transaction = await connection.BeginReadOnlyTransactionAsync())
-                using (var cmd = query
-                    ? connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName}")
-                    : connection.CreateReadCommand(_fixture.TableName, ReadOptions.FromColumns(_fixture.ColumnNames), KeySet.All))
-                {
-                    transaction.DisposeBehavior = DisposeBehavior.CloseResources;
-                    cmd.Transaction = transaction;
-                    var partitions = await cmd.GetReaderPartitionsAsync(1000);
-                    var transactionId = transaction.TransactionId;
+            Assert.Equal(numRows, _rowsRead);
+        }
 
-                    //we simulate a serialization/deserialization step in the call to the subtask.
-                    await Task.WhenAll(partitions.Select(
-                            x => DistributedReadWorkerAsync(CommandPartition.FromBase64String(x.ToBase64String()),
-                                            TransactionId.FromBase64String(transactionId.ToBase64String()))))
-                        .ConfigureAwait(false);
-                }
+        [Theory, CombinatorialData, Obsolete]
 
-                Assert.Equal(numRows, _rowsRead);
-            }
+        public async Task DistributedReadAsyncWithoutOptions(bool query)
+        {
+            int numRows = await GetNumberOfRows();
+
+            using var connection = new SpannerConnection(_fixture.ConnectionString);
+            await connection.OpenAsync();
+
+            using var transaction = await connection.BeginReadOnlyTransactionAsync();
+            using var cmd = query
+                ? connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName}")
+                : connection.CreateReadCommand(_fixture.TableName, ReadOptions.FromColumns(_fixture.ColumnNames), KeySet.All);
+            transaction.DisposeBehavior = DisposeBehavior.CloseResources;
+            cmd.Transaction = transaction;
+            var partitions = await cmd.GetReaderPartitionsAsync(1000);
+            var transactionId = transaction.TransactionId;
+
+            //we simulate a serialization/deserialization step in the call to the subtask.
+            await Task.WhenAll(partitions.Select(
+                    x => DistributedReadWorkerAsync(CommandPartition.FromBase64String(x.ToBase64String()),
+                                    TransactionId.FromBase64String(transactionId.ToBase64String()))))
+                .ConfigureAwait(false);
+
+            Assert.Equal(numRows, _rowsRead);
+        }
+
+        private async Task<int> GetNumberOfRows()
+        {
+            using var connection = _fixture.GetConnection();
+            using var cmd = connection.CreateSelectCommand($"SELECT COUNT(*) FROM {_fixture.TableName}");
+            return await cmd.ExecuteScalarAsync<int>();
         }
 
         /// <summary>
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/PartitionOptionsTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/PartitionOptionsTests.cs
@@ -1,0 +1,47 @@
+// Copyright 2023 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+using Google.Cloud.Spanner.V1;
+using System;
+using System.Threading;
+using Xunit;
+
+namespace Google.Cloud.Spanner.Data.Tests;
+
+public class PartitionOptionsTests
+{
+    [Fact]
+    public void CheckDefaultValues()
+    {
+        var defaultOptions = PartitionOptions.Default;
+
+        Assert.Null(defaultOptions.MaxPartitions);
+        Assert.Null(defaultOptions.PartitionSizeBytes);
+        Assert.False(defaultOptions.DataBoostEnabled);
+    }
+
+    [Fact]
+    public void WithMethods()
+    {
+        var options = PartitionOptions.Default.WithPartitionSizeBytes(10).WithMaxPartitions(20).WithDataBoostEnabled(false);
+        var newOptions = options.WithPartitionSizeBytes(100).WithMaxPartitions(200).WithDataBoostEnabled(true);
+
+        Assert.Equal(10, options.PartitionSizeBytes);
+        Assert.Equal(20, options.MaxPartitions);
+        Assert.Equal(false, options.DataBoostEnabled);
+
+        Assert.Equal(100, newOptions.PartitionSizeBytes);
+        Assert.Equal(200, newOptions.MaxPartitions);
+        Assert.Equal(true, newOptions.DataBoostEnabled);
+    }
+}
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// Copyright 2017 Google Inc. All Rights Reserved.
 // 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -959,8 +959,46 @@ namespace Google.Cloud.Spanner.Data.Tests
                 It.IsAny<CallSettings>()));
         }
 
-        [Fact]
-        public async Task CanExecuteReadPartitionedReadCommand()
+        [Theory, CombinatorialData]
+        public async Task CanExecuteReadPartitionedReadCommand(bool dataBoostEnabled)
+        {
+            var spannerClientMock = SpannerClientHelpers
+                .CreateMockClient(Logger.DefaultLogger);
+            spannerClientMock
+                .SetupBatchCreateSessionsAsync()
+                .SetupBeginTransactionAsync()
+                .SetupPartitionAsync()
+                .SetupStreamingRead();
+
+            var connection = BuildSpannerConnection(spannerClientMock);
+            var transaction = await connection.BeginReadOnlyTransactionAsync();
+            var command = connection.CreateReadCommand("Foo", ReadOptions.FromColumns("Col1", "Col2").WithLimit(10), KeySet.All);
+            command.Transaction = transaction;
+            var partitions = await command.GetReaderPartitionsAsync(PartitionOptions.Default.WithPartitionSizeBytes(0).WithMaxPartitions(10).WithDataBoostEnabled(dataBoostEnabled));
+
+            Assert.Equal(dataBoostEnabled, CommandPartition.FromBase64String(partitions.FirstOrDefault().ToBase64String()).Request.DataBoostEnabled);
+
+            foreach (var partition in partitions)
+            {
+                // Normally we would send this information to another client to read, but we are just simulating it here
+                // by serializing and deserializing the information locally.
+                var tx = connection.BeginReadOnlyTransaction(TransactionId.FromBase64String(transaction.TransactionId.ToBase64String()));
+                var cmd = connection.CreateCommandWithPartition(CommandPartition.FromBase64String(partition.ToBase64String()), tx);
+                var reader = await cmd.ExecuteReaderAsync();
+                Assert.True(reader.HasRows);
+            }
+
+            spannerClientMock.Verify(client => client.StreamingRead(
+                It.Is<ReadRequest>(request =>
+                    !request.PartitionToken.IsEmpty &&
+                    request.DataBoostEnabled == dataBoostEnabled &&
+                    object.Equals(request.Transaction.Id.ToBase64(), transaction.TransactionId.Id)),
+                It.IsAny<CallSettings>()), Times.Exactly(10));
+        }
+
+        // This unit test will be removed once we remove the obsolete method.
+        [Fact, Obsolete]
+        public async Task CanExecuteReadPartitionedReadCommand_WithoutOptions()
         {
             var spannerClientMock = SpannerClientHelpers
                 .CreateMockClient(Logger.DefaultLogger);
@@ -975,6 +1013,7 @@ namespace Google.Cloud.Spanner.Data.Tests
             var command = connection.CreateReadCommand("Foo", ReadOptions.FromColumns("Col1", "Col2").WithLimit(10), KeySet.All);
             command.Transaction = transaction;
             var partitions = await command.GetReaderPartitionsAsync(0, 10);
+
             foreach (var partition in partitions)
             {
                 // Normally we would send this information to another client to read, but we are just simulating it here
@@ -986,7 +1025,10 @@ namespace Google.Cloud.Spanner.Data.Tests
             }
 
             spannerClientMock.Verify(client => client.StreamingRead(
-                It.Is<ReadRequest>(request => !request.PartitionToken.IsEmpty && object.Equals(request.Transaction.Id.ToBase64(), transaction.TransactionId.Id)),
+                It.Is<ReadRequest>(request =>
+                    !request.PartitionToken.IsEmpty &&
+                    request.DataBoostEnabled == false &&
+                    object.Equals(request.Transaction.Id.ToBase64(), transaction.TransactionId.Id)),
                 It.IsAny<CallSettings>()), Times.Exactly(10));
         }
 
@@ -1091,7 +1133,7 @@ namespace Google.Cloud.Spanner.Data.Tests
                  "Foo", ReadOptions.FromColumns(new List<string> { "Col1" }),
                  KeySet.FromParameters(new SpannerParameterCollection
                      {
-                        { "p", dbType, value } 
+                        { "p", dbType, value }
                      }));
             using var reader = command.ExecuteReader();
             Assert.True(reader.HasRows);
@@ -1123,7 +1165,7 @@ namespace Google.Cloud.Spanner.Data.Tests
             var sessionPoolManager = new SessionPoolManager(sessionPoolOptions, spannerClient.Settings, spannerClient.Settings.Logger, (_o, _s, _l) => Task.FromResult(spannerClient));
             sessionPoolManager.SpannerSettings.Scheduler = spannerClient.Settings.Scheduler;
             sessionPoolManager.SpannerSettings.Clock = spannerClient.Settings.Clock;
-            
+
             builder.DataSource = DatabaseName.Format(SpannerClientHelpers.ProjectId, SpannerClientHelpers.Instance, SpannerClientHelpers.Database);
             builder.SessionPoolManager = sessionPoolManager;
             return new SpannerConnection(builder);
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
@@ -1,0 +1,86 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License").
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at 
+//
+// https://www.apache.org/licenses/LICENSE-2.0 
+//
+// Unless required by applicable law or agreed to in writing, software 
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using System.Threading;
+
+namespace Google.Cloud.Spanner.Data;
+
+/// <summary>
+/// An immutable class representing partition options.
+/// </summary>
+/// <remarks>
+/// <para>
+/// The specified values of `PartitionSizeBytes` and `MaxPartitions`
+/// are utilized in the RPC call while generating the partition tokens.
+/// </para>
+/// <para>
+/// The specified flag for `DataBoostEnabled` will be stored in the request property
+/// of the <see cref="CommandPartition"/> object and is used when creating a request
+/// to read from the partition. If `DataBoostEnabled` is set to true, the request will be
+/// executed using Spanner-independent compute resources.
+/// </para>
+/// </remarks>
+public sealed class PartitionOptions
+{
+    /// <summary>
+    /// Returns a default instance of <see cref="PartitionOptions"/>.
+    /// </summary>
+    public static PartitionOptions Default { get; } = new PartitionOptions(null, null, false);
+
+    /// <summary>
+    /// Returns a new instance with the same options as this one, but with the specified new value for <see cref="DataBoostEnabled" />.
+    /// </summary>
+    /// <returns>A new instance of <see cref="PartitionOptions"/>.</returns>
+    public PartitionOptions WithDataBoostEnabled(bool? dataBoostEnabled) => new(PartitionSizeBytes, MaxPartitions, dataBoostEnabled);
+
+    /// <summary>
+    /// Returns a new instance with the same options as this one, but with the specified new value for <see cref="PartitionSizeBytes"/>.
+    /// </summary>
+    /// <returns>A new instance of <see cref="PartitionOptions"/>.</returns>
+    public PartitionOptions WithPartitionSizeBytes(long? partitionSizeBytes) => new(partitionSizeBytes, MaxPartitions, DataBoostEnabled);
+
+    /// <summary>
+    /// Returns a new instance with the same options as this one, but with the specified new value for <see cref="MaxPartitions"/>.
+    /// </summary>
+    /// <returns>A new instance of <see cref="PartitionOptions"/>.</returns>
+    public PartitionOptions WithMaxPartitions(long? maxPartitions) => new(PartitionSizeBytes, maxPartitions, DataBoostEnabled);
+
+    /// <summary>
+    /// The desired data size for each partition generated.
+    /// The default value for this is null, in which case the server will decide the data size for each partition generated.
+    /// </summary>
+    public long? PartitionSizeBytes { get; }
+
+    /// <summary>
+    /// The desired maximum number of partitions to return.
+    /// The default value for this is null, in which case the server will decide how many partitions to return.
+    /// </summary>
+    public long? MaxPartitions { get; }
+
+    /// <summary>
+    /// If set to true, the request will be executed using Spanner independent compute resources.
+    /// The default value for this option is false.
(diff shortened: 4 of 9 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/StorageClassTest.cs
```diff
@@ -103,7 +104,7 @@ public void CreateBucketAndObjects()
             string objectName2 = IdGenerator.FromGuid();
 
             var obj2 = client.UploadObject(bucketName, objectName2, "application/octet-stream", TestHelpers.GenerateData(100));
-            Assert.Equal(updatedBucketStorageClass, obj2.StorageClass);
+            Assert.NotEqual(updatedBucketStorageClass, obj2.StorageClass);
```
jskeet, line 107: Is this change intentional? My guess is that it isn't.

anuragsrivstv: No It's not intended. I'm reverting it.

anuragsrivstv: Removed now.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -400,7 +400,21 @@ public override void Prepare()
...
-            CreateExecutableCommand().GetReaderPartitionsAsync(partitionSizeBytes, maxPartitions, cancellationToken);
+            GetReaderPartitionsAsync(new PartitionOptions(partitionSizeBytes, maxPartitions, false, cancellationToken));
+
+        /// <summary>
+        /// Creates a set of <see cref="CommandPartition"/> objects that are used to execute a query or read
+        /// operation in parallel.  Each of the returned command partitions are used
+        /// by <see cref="SpannerConnection.CreateCommandWithPartition"/> to create a new <see cref="SpannerCommand"/>
+        /// that returns a subset of data.
+        /// </summary>
+        /// <param name="options"> An instance of <see cref="PartitionOptions"/> in which options can be specified for use when generating
```
Rishabh-V, line 411: We can probably reword this.

anuragsrivstv:
Hi @jskeet and @amanda-tarafa  Please have a look at this PR whenever you have time. Thanks.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <summary>
+    /// Returns a default instance of <see cref="PartitionOptions"/> with databoost enabled flag set to false.
+    /// </summary>
+    public static PartitionOptions Default { get; } = new PartitionOptions(null, null, false);
+
+    /// <summary>
+    /// Returns a new instance with the same options as this one, but with databoost enabled flag set to true.
+    /// </summary>
+    /// <returns> A new instance of <see cref="PartitionOptions"/> class.</returns>
+    public PartitionOptions WithDataBoostEnabled => new(PartitionSizeBytes, MaxPartitions, true, CancellationToken);
```
jskeet, line 39: I'd expect this to be a method rather than a property:

```csharp
public PartitionOptions WithDataBoostEnabled(bool dataBoostEnabled) => new(PartitionSizeBytes, MaxPartitions, dataBoostEnabled, CancellationToken);
```

I'd expect similar methods for the other aspects too.

See QueryOptions.cs for an example of a similar design.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+    internal long? MaxPartitions { get; }
+
+    internal bool DataBoostEnabled { get; }
+
+    internal CancellationToken CancellationToken { get; }
+
+    /// <summary>
+    /// Creates a new instance of <see cref="PartitionOptions"/> class.
+    /// </summary>
+    /// <param name="partitionSizeBytes">
```
jskeet, line 52: These descriptions can be in the properties, after making them public.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// is currently 10,000. The maximum value is currently 200,000.  This is only
+    /// a hint.  The actual number of partitions returned may be smaller or larger than
+    /// this maximum count request.
+    /// </param>
+    /// <param name="dataBoostEnabled">
+    /// If set to true, the request will be executed using Spanner independent compute resources.
+    /// The default value for this option is false.
+    /// </param>
+    /// <param name="cancellationToken">An optional token for canceling the call.</param>
+    public PartitionOptions(long? partitionSizeBytes, long? maxPartitions, bool? dataBoostEnabled, CancellationToken cancellationToken = default)
```
jskeet, line 69: I'd prefer this to be internal. In our experience, public constructors end up being annoying when we want to add more parameters - particularly as one is currently optional here. (We made that mistake with CallSettings.)

Having a static property for the default, then "With" methods that return the new values, is a nicely extensible pattern.

(There's an exception if we're *really* certain that no additional parameters will be needed, but it feels like this is a case where there may be more in the future.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -164,23 +164,24 @@ private Task<ReliableStreamReader> ExecuteReadOrQueryRequestAsync(TimestampBound
                 return effectiveTransaction.ExecuteReadOrQueryAsync(request, cancellationToken, CommandTimeout);
             }
 
-            internal async Task<IReadOnlyList<CommandPartition>> GetReaderPartitionsAsync(long? partitionSizeBytes, long? maxPartitions, CancellationToken cancellationToken)
+            internal async Task<IReadOnlyList<CommandPartition>> GetReaderPartitionsAsync(PartitionOptions options)
```
jskeet, line 167: This is where I'd expect a cancellation token as well - and in the public methods.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -164,23 +164,24 @@ private Task<ReliableStreamReader> ExecuteReadOrQueryRequestAsync(TimestampBound
...
                 var tokens = await Transaction.GetPartitionTokensAsync(
-                        readOrQueryRequest, partitionSizeBytes, maxPartitions, cancellationToken, CommandTimeout)
+                        readOrQueryRequest, options.PartitionSizeBytes, options.MaxPartitions, options.CancellationToken, CommandTimeout)
                     .ConfigureAwait(false);
                 return tokens.Select(
                     x =>
                     {
                         var request = readOrQueryRequest.CloneRequest();
                         request.PartitionToken = x;
+                        request.DataBoostEnabled = options.DataBoostEnabled;
```
jskeet, line 184: I'd move this to just after we've set up `readOrQueryRequest` - this lambda expression should (IMO) be about partition-specific changes.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -985,8 +988,13 @@ public async Task CanExecuteReadPartitionedReadCommand()
                 Assert.True(reader.HasRows);
             }
 
+            bool expectedDataBoostEnabledValue = dataBoostEnabled.HasValue && dataBoostEnabled.Value;
```
jskeet, line 991: This can just be `dataBoostEnabled ?? false` - which I'd probably inline below.

jskeet (commented):
Mostly fine - the key change I'd like to see is in the design of PartitionOptions itself.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+{
+    /// <summary>
+    /// Returns a default instance of <see cref="PartitionOptions"/> with databoost enabled flag set to false.
+    /// </summary>
+    public static PartitionOptions Default { get; } = new PartitionOptions(null, null, false);
+
+    /// <summary>
+    /// Returns a new instance with the same options as this one, but with the databoost enabled flag updated to the provided value.
+    /// </summary>
+    /// <returns> A new instance of <see cref="PartitionOptions"/> class.</returns>
```
jskeet, line 38: Super nit: Please remove the leading space from each of these `<returns>` bits of documentation.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+/// </summary>
+public sealed class PartitionOptions
+{
+    /// <summary>
+    /// Returns a default instance of <see cref="PartitionOptions"/> with databoost enabled flag set to false.
+    /// </summary>
+    public static PartitionOptions Default { get; } = new PartitionOptions(null, null, false);
+
+    /// <summary>
+    /// Returns a new instance with the same options as this one, but with the databoost enabled flag updated to the provided value.
```
jskeet, line 36: I'd use `<see cref="DataBoostEnabled" />` instead of "the databoost enabled flag".

I'd also avoid saying "update to the provided value" as it could cause confusion about whether it changes this instance. How about:

```csharp
/// Returns a new instance with the same options as this one, but with the specified new value for <see cref="DataBoostEnabled" />.
```

(And ditto the ones below.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+// See the License for the specific language governing permissions and 
+// limitations under the License.
+
+using System.Threading;
+
+namespace Google.Cloud.Spanner.Data;
+
+/// <summary>
+/// An immutable class representing partition options.
+/// The specified values of `PartitionSizeBytes`, `MaxPartitions` and `CancellationToken`
```
jskeet, line 21: We need to remove CancellationToken here.

I'd probably split this into a summary of just the first line, then a remarks section with two "para" elements, one for this sentence and one for DataBoostEnabled.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+/// are utilized in the RPC call while generating the partition tokens.
+/// The specified flag for `DataBoostEnabled` will be stored in the request property
+/// of the <see cref="CommandPartition"/> object and is used when executing the
+/// partition object. If `DataBoostEnabled` is set to true, the request will be
+/// executed using Spanner independent compute resources.
```
jskeet, line 26: Nit: Spanner-independent perhaps? (To be honest, it's not entirely clear to me what that means, but I acknowledge it's probably product wording.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+    public PartitionOptions WithPartitionSizeBytes(long? partitionSizeBytes) => new(partitionSizeBytes, MaxPartitions, DataBoostEnabled);
+
+    /// <summary>
+    /// Returns a new instance with the same options as this one, but with the max partitions updated to the provided value.
+    /// </summary>
+    /// <returns> A new instance of <see cref="PartitionOptions"/> class.</returns>
+    public PartitionOptions WithMaxPartitions(long? maxPartitions) => new(PartitionSizeBytes, maxPartitions, DataBoostEnabled);
+
+    /// <summary>
+    /// The desired data size for each partition generated.  The default for this
```
jskeet, line 54: Nit: Please replace all double-spaces after periods with just a single space. There's no need to double-space.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+/// are utilized in the RPC call while generating the partition tokens.
+/// The specified flag for `DataBoostEnabled` will be stored in the request property
+/// of the <see cref="CommandPartition"/> object and is used when executing the
+/// partition object. If `DataBoostEnabled` is set to true, the request will be
+/// executed using Spanner independent compute resources.
+/// </summary>
+public sealed class PartitionOptions
+{
+    /// <summary>
+    /// Returns a default instance of <see cref="PartitionOptions"/> with databoost enabled flag set to false.
```
jskeet, line 31: I'd stop after "a default instance of <see cref="PartitionOptions"/>" - the defaults are called out in the individual properties. No need to prioritize databoost here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// The desired data size for each partition generated.  The default for this
+    /// option is currently 1 GiB.  This is only a hint. The actual size of each
+    /// partition may be smaller or larger than this size request.
+    /// </summary>
+    public long? PartitionSizeBytes { get; }
+
+    /// <summary>
+    /// The desired maximum number of partitions to return.  For example, this may
+    /// be set to the number of workers available.  The default for this option
+    /// is currently 10,000. The maximum value is currently 200,000.  This is only
```
jskeet, line 63: The default for this property is null, and I think we should say so - and say that that means the server gets to decide. We *might* want to document the currently defaulting, but I'm not sure whether it's worth it (and it could easily get out of date).

Ditto for PartitionSizeBytes.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -400,7 +400,27 @@ public override void Prepare()
...
-            CreateExecutableCommand().GetReaderPartitionsAsync(partitionSizeBytes, maxPartitions, cancellationToken);
+            GetReaderPartitionsAsync(
+                PartitionOptions.Default.
+                WithPartitionSizeBytes(partitionSizeBytes).
+                WithMaxPartitions(maxPartitions),
+                cancellationToken);
+
+        /// <summary>
+        /// Creates a set of <see cref="CommandPartition"/> objects that are used to execute a query or read
+        /// operation in parallel.  Each of the returned command partitions are used
```
jskeet, line 411: Again, let's remove double-spacing.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/PartitionOptionsTests.cs
```diff
@@ -0,0 +1,101 @@
...
+
+namespace Google.Cloud.Spanner.Data.Tests;
+
+public class PartitionOptionsTests
+{
+    [Fact]
+    public void CheckDefaultValues()
+    {
+        var options = PartitionOptions.Default;
+
```
jskeet, line 27: Possibly just call `AssertDefaultOptions(PartitionOptions.Default)` for this test?
(If we introduce a new property, it would be good to only have to update one place.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -986,7 +989,10 @@ public async Task CanExecuteReadPartitionedReadCommand()
             }
 
             spannerClientMock.Verify(client => client.StreamingRead(
-                It.Is<ReadRequest>(request => !request.PartitionToken.IsEmpty && object.Equals(request.Transaction.Id.ToBase64(), transaction.TransactionId.Id)),
+                It.Is<ReadRequest>(request =>
+                !request.PartitionToken.IsEmpty &&
```
jskeet, line 993: These three lines should be indented IMO.

jskeet (commented):
Much better with with With methods, thanks. Down to just nits, really.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
```diff
@@ -53,7 +65,9 @@ public async Task DistributedReadAsync(bool query)
                 {
                     transaction.DisposeBehavior = DisposeBehavior.CloseResources;
                     cmd.Transaction = transaction;
-                    var partitions = await cmd.GetReaderPartitionsAsync(1000);
+                    var partitions = dataBoostEnabled == null ?
+                        await cmd.GetReaderPartitionsAsync(1000) :
+                        await cmd.GetReaderPartitionsAsync(PartitionOptions.Default.WithPartitionSizeBytes(1000).WithDataBoostEnabled(dataBoostEnabled));
```
amanda-tarafa, line 70: We need to add some asserts to this test (and once you do that, I'm not certain that reusing the same method is a good idea re clarity, maybe copy paste the common bits).

- We need to assert that the requests in the CommandPartitions returned in line 68 have DataBoostEnabled set to the correct value.
- We also need to assert that after serialization/deserialization the CommandPartitions built in line 75 have DataBostEnabled set to the correct value, possibly assert this before using the partitions in DistributedReadWorkerAsync.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerCommandTests.cs
```diff
@@ -959,8 +959,9 @@ public async Task CanExecuteReadCommandWithLimit()
                 It.IsAny<CallSettings>()));
         }
 
-        [Fact]
-        public async Task CanExecuteReadPartitionedReadCommand()
+        [CombinatorialData]
+        [Theory]
```
amanda-tarafa, line 963:
```suggestion
        [Theory]
        [CombinatorialData]
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
amanda-tarafa: Maybe a link to product documentation here, instead of attempting to explain this?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -400,7 +400,22 @@ public override void Prepare()
             long? partitionSizeBytes = null,
```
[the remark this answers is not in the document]
amanda-tarafa: I'd vote for making obsolete, now that we have the options, all new parameters will be added there.
@jskeet what do you think?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -164,17 +164,18 @@ private Task<ReliableStreamReader> ExecuteReadOrQueryRequestAsync(TimestampBound
...
+            internal async Task<IReadOnlyList<CommandPartition>> GetReaderPartitionsAsync(PartitionOptions options, CancellationToken cancellationToken)
             {
                 ValidateConnectionAndCommandTextBuilder();
 
                 GaxPreconditions.CheckState(Transaction?.Mode == TransactionMode.ReadOnly,
                     "GetReaderPartitions can only be executed within an explicitly created read-only transaction.");
 
                 await Connection.EnsureIsOpenAsync(cancellationToken).ConfigureAwait(false);
-                var readOrQueryRequest = GetReadOrQueryRequest();
+                var readOrQueryRequest = GetReadOrQueryRequest().CloneRequest();
```
amanda-tarafa, line 175: I don't think we need to clone here.

amanda-tarafa (changes requested):

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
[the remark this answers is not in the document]
jskeet: I don't mind... it does add unnecessary work for customers, but it'll be cleaner without.

jskeet:
I'll have another look tomorrow morning.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
```diff
@@ -29,9 +31,52 @@ public class PartitionedReadTests
...
+                    ? connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName}")
+                    : connection.CreateReadCommand(_fixture.TableName, ReadOptions.FromColumns(_fixture.ColumnNames), KeySet.All))
+                {
+                    transaction.DisposeBehavior = DisposeBehavior.CloseResources;
+                    cmd.Transaction = transaction;
+                    var partitions = await cmd.GetReaderPartitionsAsync(PartitionOptions.Default.WithPartitionSizeBytes(100).WithDataBoostEnabled(dataBoostEnabled));
+                    var transactionId = transaction.TransactionId;
+
+                    var serDserPartition = CommandPartition.FromBase64String(partitions.FirstOrDefault().ToBase64String());
+                    var request = serDserPartition.Request.ExecuteSqlRequest;// TODO add ternary operator
```
jskeet, line 62: I don't understand this comment

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
```diff
@@ -29,9 +31,52 @@ public class PartitionedReadTests
...
+                using (var cmd = query
+                    ? connection.CreateSelectCommand($"SELECT * FROM {_fixture.TableName}")
+                    : connection.CreateReadCommand(_fixture.TableName, ReadOptions.FromColumns(_fixture.ColumnNames), KeySet.All))
+                {
+                    transaction.DisposeBehavior = DisposeBehavior.CloseResources;
+                    cmd.Transaction = transaction;
+                    var partitions = await cmd.GetReaderPartitionsAsync(PartitionOptions.Default.WithPartitionSizeBytes(100).WithDataBoostEnabled(dataBoostEnabled));
+                    var transactionId = transaction.TransactionId;
+
+                    var serDserPartition = CommandPartition.FromBase64String(partitions.FirstOrDefault().ToBase64String());
```
jskeet, line 61: I would leave the serialization and deserialization testing to a simple unit test. Just using the first partition directly here would be fine, IMO.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
```diff
@@ -29,9 +31,52 @@ public class PartitionedReadTests
...
+                    var request = serDserPartition.Request.ExecuteSqlRequest;// TODO add ternary operator
+                    Assert.Equal(request.DataBoostEnabled, dataBoostEnabled);
+
+
+                    //we simulate a serialization/deserialization step in the call to the subtask.
+                    await Task.WhenAll(partitions.Select(
+                            x => DistributedReadWorkerAsync(CommandPartition.FromBase64String(x.ToBase64String()),
```
jskeet, line 68: Again, this code would be much simpler without the serialization/deserialization - which is easy to unit test separately.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
```diff
@@ -29,9 +31,52 @@ public class PartitionedReadTests
...
+                    cmd.Transaction = transaction;
+                    var partitions = await cmd.GetReaderPartitionsAsync(PartitionOptions.Default.WithPartitionSizeBytes(100).WithDataBoostEnabled(dataBoostEnabled));
+                    var transactionId = transaction.TransactionId;
+
+                    var serDserPartition = CommandPartition.FromBase64String(partitions.FirstOrDefault().ToBase64String());
+                    var request = serDserPartition.Request.ExecuteSqlRequest;// TODO add ternary operator
+                    Assert.Equal(request.DataBoostEnabled, dataBoostEnabled);
+
+
+                    //we simulate a serialization/deserialization step in the call to the subtask.
```
jskeet, line 66: If we *do* keep the serialization/deserialization so we need to keep this comment, please add a space after `//` and capitalize the `w`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
```diff
@@ -29,9 +31,52 @@ public class PartitionedReadTests
...
 
-        [CombinatorialData]
-        [Theory]
-        public async Task DistributedReadAsync(bool query)
+        [Theory, CombinatorialData]
+        [Trait(Constants.SupportedOnEmulator, Constants.No)]
+        public async Task DistributedReadAsync(bool query, bool dataBoostEnabled)
+        {
+            int numRows;
+            using (var connection = _fixture.GetConnection())
```
jskeet, line 39: Given that this code is in multiple tests, let's extract it out as a method.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
jskeet: (Ah, it looks like this was possibly a Ben comment from a long time ago. I'd say it's fine to write the *new* test cleanly, and I can go back and fix up older ones later...)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/PartitionOptionsTests.cs
```diff
@@ -0,0 +1,97 @@
...
+    private void AssertDefaultOptions(PartitionOptions defaultOptions)
+    {
+        Assert.Null(defaultOptions.MaxPartitions);
+        Assert.Null(defaultOptions.PartitionSizeBytes);
+        Assert.False(defaultOptions.DataBoostEnabled);
+    }
+
+    [Theory, CombinatorialData]
+    public void WithDataBoostEnabled_ShouldNotAlterExistingOptionInstance(bool dataBoostEnabled)
```
jskeet, line 73: I really don't think we need this to be a combinatorial test.

I'd actually suggest changing the three earlier tests to separate out "what are the defaults" from "what do the with methods do":

For each property:
- Create a completely customized options object like in the first line of this test
- Use WithXyz to create a new object
- Assert that all values in the new object are correct

(I actually wouldn't bother checking the old ones, given that it's clearly immutable.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,85 @@
...
+
+/// <summary>
+/// An immutable class representing partition options.
+/// </summary>
+/// <remarks>
+/// <para>The specified values of `PartitionSizeBytes` and `MaxPartitions`
+/// are utilized in the RPC call while generating the partition tokens.</para>
+/// <para>The specified flag for `DataBoostEnabled` will be stored in the request property
+/// of the <see cref="CommandPartition"/> object and is used when executing the
+/// partition object.For more information see
```
jskeet, line 27: If you *do* want to refer to the properties in the protos, you can do so with a `<see cref="..." />` - but I probably wouldn't do that anyway. If there's product documentation, refer to that - if there isn't, just remove the sentence.

(If we keep the sentence, add a space after the period.)

Personally I find this sort of documentation more readable if you put the `<para>` opening and closing tags on their own lines - that way the fact that it's in two paragraphs becomes more obvious in the source code.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    /// <summary>
+    /// Returns a default instance of <see cref="PartitionOptions"/>.
+    /// </summary>
+    public static PartitionOptions Default { get; } = new PartitionOptions(null, null, false);
+
+    /// <summary>
+    /// Returns a new instance with the same options as this one, but with the specified new value for <see cref="DataBoostEnabled" />.
+    /// </summary>
+    /// <returns>A new instance of <see cref="PartitionOptions"/> class.</returns>
```
jskeet, line 41: No need for the " class" part. (And if you *do* want it, we need "A new instance of the PartitionOptions class" rather than just "A new instance of PartitionOptions class.".) Same for ther other methods.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// The desired data size for each partition generated.
+    /// The default value for this is null, in which case the server will decide the value of
+    /// [PartitionSizeBytes](https://cloud.google.com/dotnet/docs/reference/Google.Cloud.Spanner.V1/latest/Google.Cloud.Spanner.V1.PartitionOptions#Google_Cloud_Spanner_V1_PartitionOptions_PartitionSizeBytes).
+    /// </summary>
+    public long? PartitionSizeBytes { get; }
+
+    /// <summary>
+    /// The desired maximum number of partitions to return.
+    /// The default value for this is null, in which case the server will decide the value of
+    /// [MaxPartitions](https://cloud.google.com/dotnet/docs/reference/Google.Cloud.Spanner.V1/latest/Google.Cloud.Spanner.V1.PartitionOptions#Google_Cloud_Spanner_V1_PartitionOptions_MaxPartitions).
```
jskeet, line 66: Again, use `<see>` to refer to other C# code - but I don't think we need a reference at all.

Just:

> The default value for this is null, in which case the server will decide how many partitions to return.

... would be clearer IMO. Likewise for PartitionSizeBytes.

jskeet:
Btw, I've renamed the PR to make it more concise - think of it in terms of release notes:

- "Changes" is redundant because any PR is going to be a change
- We don't usually need to say which library it affects, because it'll be in the release notes for that specific library. (This is something I forget quite often, admittedly.)
- The rest is just rewording for brevity

jskeet (commented):
Just a couple of nits and questions.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/PartitionOptionsTests.cs
```diff
@@ -31,40 +31,7 @@ public void CheckDefaultValues()
...
-        var newOptions = options.WithPartitionSizeBytes(100);
-
-        Assert.Equal(100, newOptions.PartitionSizeBytes);
-        Assert.Equal(20, newOptions.MaxPartitions);
-        Assert.Equal(false, newOptions.DataBoostEnabled);
-    }
-
-    [Fact]
-    public void CheckOptionsAreImmutable()
+    public void CheckWithMethods()
```
jskeet, line 34: Again, I'd just rename it to `WithMethods` - we're in a test file, so the "check" aspect can be taken for granted.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ReadOrQueryRequest.cs
```diff
@@ -55,6 +55,11 @@ internal interface IReadOrQueryRequest
         /// </summary>
         ByteString ResumeToken { set; }
 
+        /// <summary>
+        /// See <see cref="ReadRequest.DataBoostEnabled"/> and <see cref="ExecuteSqlRequest.DataBoostEnabled"/>
+        /// </summary>
+        bool DataBoostEnabled { get; set; }
```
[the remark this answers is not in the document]
jskeet: Right. I guess ReadOrQueryRequest is unlikely to be used by customers directly anyway, so it doesn't matter much.

jskeet (approved):
One nit, but then I *think* this can be merged (after rebasing to one commit).
I'd quite like Amanda to have a final look before we *release* the change, unless there's a huge amount of urgency to get it out this week.

anuragsrivstv:
> 

Sure Jon. Thanks for reviewing it !

jskeet (approved):
I'd missed the "CheckDefaultValues" name, but it's not worth doing another round just for that.

jskeet:
Merging without Amanda's re-approval; Amanda can double-check before release.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/PartitionedReadTests.cs
```diff
@@ -29,42 +31,72 @@ public class PartitionedReadTests
         public PartitionedReadTests(PartitionedReadTableFixture fixture) =>
             _fixture = fixture;
 
-        [CombinatorialData]
-        [Theory]
-        public async Task DistributedReadAsync(bool query)
+        [Theory, CombinatorialData]
+        [Trait(Constants.SupportedOnEmulator, Constants.No)]
+        public async Task DistributedReadAsync(bool query, bool dataBoostEnabled)
```
amanda-tarafa, line 36: I think this test should be split in two, DistributedReadAsync_Select, DistributedReadAsync_Read.

I understand it was like that before, but test in general shouldn't have conditional setup and/or asserts.

(It's fine if the obsolete one remains as is, as we will remove it)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/PartitionOptions.cs
```diff
@@ -0,0 +1,86 @@
...
+/// The specified values of `PartitionSizeBytes` and `MaxPartitions`
+/// are utilized in the RPC call while generating the partition tokens.
+/// </para>
+/// <para>
+/// The specified flag for `DataBoostEnabled` will be stored in the request property
+/// of the <see cref="CommandPartition"/> object and is used when creating a request
+/// to read from the partition. If `DataBoostEnabled` is set to true, the request will be
+/// executed using Spanner-independent compute resources.
+/// </para>
+/// </remarks>
```
amanda-tarafa, line 33: nit: I think I would put all of this info on each of the individual properties. Maybe leaving a note here saying, "Some of the options are used when generating the partition tokens, while others are used when reading the partitions. Please refer to each of the options documentation for more information".

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -377,30 +377,51 @@ public override void Prepare()
...
+        /// <summary>
+        /// Creates a set of <see cref="CommandPartition"/> objects that are used to execute a query or read
+        /// operation in parallel. Each of the returned command partitions are used
+        /// by <see cref="SpannerConnection.CreateCommandWithPartition"/> to create a new <see cref="SpannerCommand"/>
+        /// that returns a subset of data.
+        /// </summary>
+        /// <param name="options">An instance of <see cref="PartitionOptions"/> class in
+        /// which we can set the maxPartitions, partitionSizeBytes, dataBoostEnabled and
+        /// cancellationToken options for generating and executing partitions.
+        /// </param>
```
amanda-tarafa, line 419: This should be something like:

`"The <see cref="PartitionOptions"/> for generating and reading partitions`

If you list here all the actual options we need to continue to update this doc when we add new options.
Also, the cancellation token is a parameter on its own, not part of the PartitionOptions.

amanda-tarafa (commented):
There's an error in docs that should ideally be fixed before releasing. And, given that this cannot be released yet it might be a good idea to fix the couple other items as well. FYI @anuragsrivstv

Issue #10133, opened 2023-04-05
Validate use of streaming gRPC calls


jskeet:
As reported in https://github.com/grpc/grpc-dotnet/issues/2086, streaming gRPC calls can cause a memory leak if the original gRPC call isn't disposed of *and* we don't read to the end of the response stream. (And if we dispose of the call without reading to the end of the response stream, that will abort the request which probably isn't ideal.)

We should check our hand-written code for uses of bidi/server-streaming calls (I'm thinking particularly of Firestore and Spanner, but there may be more) to check that we're disposing of everything correctly. (Separately, our wrapper types should implement IDisposable.)

meredithslota:
Probably also Pub/Sub but I'm not clear on the scope of handwritten code there.

Rishabh-V:
Right @meredithslota. We have already identified the code changes needed in BigTable, Firestore, PubSub and Spanner. Jon has already implemented the `IDisposable` on the bidi/server streams in GAX. Once the GAX changes are in, we can make the relevant changes in these libraries. Thanks.

PR #10389, opened 2023-05-19
fix: PubSub disposal of gRPC streaming calls.

Rishabh-V:
Towards #10133

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
@@ -223,7 +223,8 @@ namespace Google.Cloud.PubSub.V1.Tests
                     _scheduler = scheduler;
                     _writeAsyncPreDelay = writeAsyncPreDelay; // delay within the WriteAsync() method. Simulating network or server slowness.
                     var responseStream = new En(msgs, scheduler, taskHelper, clock, useMsgAsId, ct, isExactlyOnceDelivery);
-                    _call = new AsyncDuplexStreamingCall<StreamingPullRequest, StreamingPullResponse>(null, responseStream, Task.FromResult(new Metadata()), null, null, null);
+                    // Set disposeAction parameter of AsyncDuplexStreamingCall to No-op as it is called internally while disposing stream.
+                    _call = new AsyncDuplexStreamingCall<StreamingPullRequest, StreamingPullResponse>(null, responseStream, Task.FromResult(new Metadata()), null, null, () => { });
                     _clock = clock;
                     _writeCompletes = writeCompletes;
                     _streamPings = streamPings;
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
@@ -276,10 +276,19 @@ public sealed partial class SubscriberClientImpl
                 // Ignore all errors; the stream may be in any state.
                 try
                 {
-                    _registerTaskFn(_pull.WriteCompleteAsync());
+                    Add(_pull.WriteCompleteAsync(), Next(false, () =>
+                        {
+                            try
+                            {
+                                _pull.Dispose();
+                            }
+                            finally
+                            {
+                                _pull = null;
+                            }
+                        }));
                 }
                 catch { }
-                _pull = null;
             }
         }
 


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -279,7 +279,11 @@ private void StopStreamingPull()
                     _registerTaskFn(_pull.WriteCompleteAsync());
                 }
                 catch { }
-                _pull = null;
+                finally
+                {
+                    _pull.Dispose();
```
amanda-tarafa, line 284: This one need to be registered as well, and as a continuation of the WriteCompleteAsync.
Else Dispose will likely be called before WriteCompleteAsync and WriteCompleteAsync will fail.

amanda-tarafa (changes requested):

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
Rishabh-V: Right. 
I had been registering it all this while (for deadlock issue and even today) and just before raising the PR, I changed it. Reverted now. PTAL.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -276,10 +276,13 @@ private void StopStreamingPull()
                 // Ignore all errors; the stream may be in any state.
                 try
                 {
-                    _registerTaskFn(_pull.WriteCompleteAsync());
+                    Add(_pull.WriteCompleteAsync(), Next(false, _pull.Dispose));
```
amanda-tarafa, line 279: Does the `false` mean that the previous task does not need to have succeede? Checking as we'd need Dispose to execute regardless of whether WriteCompleteAsync succeeded or not.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -223,7 +223,8 @@ private class PullStream : StreamingPullStream
                     _scheduler = scheduler;
                     _writeAsyncPreDelay = writeAsyncPreDelay; // delay within the WriteAsync() method. Simulating network or server slowness.
                     var responseStream = new En(msgs, scheduler, taskHelper, clock, useMsgAsId, ct, isExactlyOnceDelivery);
-                    _call = new AsyncDuplexStreamingCall<StreamingPullRequest, StreamingPullResponse>(null, responseStream, Task.FromResult(new Metadata()), null, null, null);
+                    // Set disposeAction parameter of AsyncDuplexStreamingCall to No-op as it is called internally while disposing stream.
```
amanda-tarafa, line 226: If AsyncDuplexStreamingCall is in our control, shouldn't we instead check if the disposeAction is null and do nothing in that case?

Is there anywhere in production code that we are creating an AsyncDuplexStreamingCall passing a null disposeAction that will nos start failing as this test?

amanda-tarafa (commented):
I have more questions.

And the main one, is there an analysis doc that describes how this is the only place we need to dispose of the streaming call?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
Rishabh-V: This is in our test infrastructure. Earlier, we were not disposing the stream so `disposeAction` being null wasn't an issue. However, now that we `Dispose` the stream, this action is internally called by Grpc.Core.Api while calling and since this action is null, the `Dispose` fails.

Rishabh-V:
> I have more questions.
> 
> And the main one, is there an analysis doc that describes how this is the only place we need to dispose of the streaming call?

Sure Amanda. I hadn't created a detailed document. I have my notes, but I can create one and share by Tuesday morning (I may be out on Monday) and we can take it up from there. Thanks.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
amanda-tarafa: Ok my main question is if Dispose will execute even if WriteCompleteAsync fails? That's now looping the loop but just to make sure we really dispose whenever we can. If not, maybe register a single task that calls WriteCompleteAsync in a try and Dispose in the finally?

Rishabh-V: Thanks Amanda. To be honest, I wasn't 100% sure. But, I have updated second commit and now I am 100% sure that Dispose will be called even if WriteCompleteAsync fails. PTAL. Thanks.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
amanda-tarafa: Thanks for the explanation

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -276,10 +276,19 @@ private void StopStreamingPull()
...
+                        {
+                            try
+                            {
+                                _pull.Dispose();
+                            }
+                            finally
+                            {
+                                _pull = null;
+                            }
+                        }, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, _taskHelper.TaskScheduler));
```
amanda-tarafa, line 289: I'm not sure why we'd need the dispose to Execute on the same thread as the WriteCompleteAsync?

Rishabh-V: I believe that `SingleChannel` is effectively single-threaded.  This code keeps it consistent.

amanda-tarafa: Then in that case I would prefer this code relying on whatever existing code is there that enforces the single threadness. That seems to be `Add(Task task, NextAction next)` that takes the first and second action. If you follow the code, you'll see that `next` will execute even if `task` failed.

amanda-tarafa (approved):
LGTM, thanks!

PR #10394, opened 2023-05-22
chore: Update to protoc v23

jskeet:
Will self-merge this (it's all basically generated) to get in before a storm of OwlBot PRs.


jskeet:
(Merged with override to avoid slew of generated changes.)

PR #10399, opened 2023-05-23
fix: Firestore disposal of gRPC streaming calls

jskeet:
Fixes the Firestore part of #10133.


jskeet:
Looks like there's one integration test that will need fixing - effectively the timing of an exception being thrown is changed. (I think that's reasonable to consider non-breaking, really.)

jskeet:
Integration test is now fixed in second commit. Alternatively we could separate out the validation from the iterator block to keep it eager, but I'm not convinced that's worth the complexity.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/WatchStream.cs
```diff
@@ -200,7 +200,7 @@ async Task CloseStreamAsync()
                         // being dropped, at which point completing the stream will fail; we don't want the listener to stop at that
                         // point. Instead, it will reconnect.
                     }
-                    underlyingStream.GrpcCall.Dispose();
```
Rishabh-V, line 203: nit: finally block here?

Rishabh-V (approved):
LGTM

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/WatchStream.cs
jskeet: Effectively done, but with a bit of refactoring. It's in a new commit.

PR #10434, opened 2023-05-25
chore: Added comments about fields, their meaning and possible issues.

Rishabh-V:


Rishabh-V:
This is for discussion, gathering feedback and to list out possible things that we need to address.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -49,6 +49,12 @@ public sealed partial class SubscriberClientImpl : SubscriberClient
     public SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<SubscriberServiceApiClient> clients, Settings settings, Func<Task> shutdown)
         : this(subscriptionName, clients, settings, shutdown, TaskHelper.Default) { }
 
+    // _modifyDeadlineSeconds is the AckDeadline in seconds before which the message must be acknowledged or have its lease extended. If the user hasn't specified a value of AckDeadline, the default of 60 seconds is used.
```
jskeet, line 52: Rather than specifying the details of each field, I was expecting to see comments right at the top (even before the class declaration) about the *conventions* for field names:

- An "ackDeadline" is the amount of time we extend the lease by in each auto-extend request
- A suffix of "window" is the amount of time we allow for an auto-extend request to take place
- A suffix of "interval" is a time we wait between sending one auto-extend request and the next; this is derived from the ackDeadline and the window.

... and then the rules about what can be done with each value:

- An ack deadline can be limited by min/max ack deadlines
- A window can be limited by min/max windows
- An interval can be calculated as "ack deadline - window"
- An interval can be limited by min/max intervals

Any other computation is forbidden.

(I'm considering experimenting with enshrining this in the type system, with some internal structs...)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -60,7 +66,11 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
         settings.Validate();
         // These values are validated in Settings.Validate() above, so no need to re-validate here.
         _modifyDeadlineSeconds = (int)((settings.AckDeadline ?? DefaultAckDeadline).TotalSeconds);
+        // FIXME: This should be named autoExtendDelay. 
         var autoExtendInterval = TimeSpan.FromSeconds(_modifyDeadlineSeconds) - (settings.AckExtensionWindow ?? DefaultAckExtensionWindow);
+        // FIXME: This should be named autoExtendDelayForExactlyOnceDelivery.
+        // FIXME: MinimumAckExtensionWindowForExactlyOnceDelivery should be renamed to MinimumLeaseExtensionDelay or MinimumLeaseExtensionSchedule with the default value of 60 seconds.
```
jskeet, line 72: Unfortunately it's public - I think we'll need to introduce a new property, and mark the old one as obsolete (and explain in a comment). Fortunately it's read-only, so I don't think it will actually break anyone.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -60,7 +66,11 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
         settings.Validate();
         // These values are validated in Settings.Validate() above, so no need to re-validate here.
         _modifyDeadlineSeconds = (int)((settings.AckDeadline ?? DefaultAckDeadline).TotalSeconds);
+        // FIXME: This should be named autoExtendDelay.
```
jskeet, line 69: I'd prefer to keep the renaming part separate from the calculation part - the renaming of "interval" to "delay" is an optional nice-to-have when the names are clear and the computations have been corrected, but the real "fixme" is that the computation at the moment doesn't make sense.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
Rishabh-V: Thanks. This rule part makes it clear what you and Amanda are thinking.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -66,10 +71,8 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
         settings.Validate();
         // These values are validated in Settings.Validate() above, so no need to re-validate here.
         _modifyDeadlineSeconds = (int)((settings.AckDeadline ?? DefaultAckDeadline).TotalSeconds);
-        // FIXME: This should be named autoExtendDelay. 
         var autoExtendInterval = TimeSpan.FromSeconds(_modifyDeadlineSeconds) - (settings.AckExtensionWindow ?? DefaultAckExtensionWindow);
-        // FIXME: This should be named autoExtendDelayForExactlyOnceDelivery.
-        // FIXME: MinimumAckExtensionWindowForExactlyOnceDelivery should be renamed to MinimumLeaseExtensionDelay or MinimumLeaseExtensionSchedule with the default value of 60 seconds.
+        // FIXME: MinimumAckExtensionWindowForExactlyOnceDelivery is incorrectly named. It should be renamed to MinimumLeaseExtensionDelay or MinimumLeaseExtensionSchedule with the default value of 60 seconds.
```
jskeet, line 75: I'd still expect both of these comments to be just one of "FIXME: This is assigning a window to an interval, which is prohibited".
We don't need to go into the impact of this yet - it's just working out what doesn't make sense, so we can then revisit.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -29,6 +29,17 @@
 
 namespace Google.Cloud.PubSub.V1;
 
+// The class follows following conventions for the field names:
+// - An "ackDeadline" is the amount of time before which the message must be acknowledged or have its lease extended.
+// - A suffix of "window" indicates the duration of time before the ackDeadline at which the lease is extended.
+// - A suffix of "interval" indicates the delay between sending one auto-lease extension request and the next; This is derived from the ackDeadline and the window.
+// The rules for manipulating the field values are:
+// - The "ackDeadline" should be within the MinAckDeadline and MaxAckDeadline. If not specified the DefaultAckDeadline is used.
```
jskeet, line 37: This is definitely better, but I wouldn't start talking about *which* min, max and default values are used. My intention was that this is effectively a set of rules for the *types* of values... exactly how those rules are used may well vary between EOD and non-EOD.

jskeet (commented):
Definitely getting there.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -29,6 +29,17 @@
 
 namespace Google.Cloud.PubSub.V1;
 
+// The class follows following conventions for the field names:
+// - An "ackDeadline" is the amount of time before which the message must be acknowledged or have its lease extended.
+// - A suffix of "window" indicates the duration of time before the ackDeadline at which the lease is extended.
+// - A suffix of "interval" indicates the delay between sending one auto-lease extension request and the next; This is derived from the ackDeadline and the window.
+// The rules for manipulating the field values are:
+// - The "ackDeadline" should be within a minimum and maximum value.
```
jskeet, line 37: These are still "checks" rather than operations, really. What I'm looking for is comments that mean when we're looking at a line of code, we can see whether the value is being used or computed in a valid way.

So it should be valid to say "compute the min of two deadlines" or "compute the max of two deadlines" (and ditto for windows and intervals", but it's not valid to say "compute the min of deadline X and window Y" or "assign this deadline value to an interval variable".

Later I'm going to try codify this in terms of internal-only structs - if I get it right, we should be able to see anywhere that we're violating those manipulation rules. It won't tell us whether the business rules about "must be at least X" are valid, but it'll tell us when the code is self-consistent.

Rishabh-V:
Closing for #10371.

PR #10455, opened 2023-05-26
Prototype to fix auto-extend delay in eod and incorporate missing feature

Rishabh-V:
This is a rough draft for discussion. This isn't polished yet. I will now update the document to list out the changes.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
@@ -82,8 +82,17 @@ public abstract partial class SubscriberClient : IAsyncDisposable
     public static TimeSpan MinimumAckExtensionWindow { get; } = TimeSpan.FromMilliseconds(50);
 
     /// <summary>
+    /// The default minimum message Acknowledgement deadline extension of 60 seconds for exactly-once delivery subscriptions.
+    /// </summary>
+    public static TimeSpan MinimumAckDeadlineExtensionForExactlyOnceDelivery { get; } = TimeSpan.FromSeconds(60);
+
+    /// <summary>
     /// The minimum message acknowledgement extension window of 60 seconds for exactly once delivery subscriptions.
     /// </summary>
+    /// <remarks>
+    /// This property is deprecated. Use <see cref="MinimumAckDeadlineExtensionForExactlyOnceDelivery"/> instead.
+    /// </remarks>
+    [Obsolete("Use MinimumAckDeadlineExtensionForExactlyOnceDelivery instead.")]
     public static TimeSpan MinimumAckExtensionWindowForExactlyOnceDelivery { get; } = TimeSpan.FromSeconds(60);
 
     /// <summary>
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
@@ -149,11 +149,13 @@ public sealed partial class SubscriberClientImpl
         private readonly CancellationTokenSource _pushStopCts;
         private readonly CancellationTokenSource _softStopCts;
         private readonly SubscriptionName _subscriptionName;
-        private readonly int _modifyDeadlineSeconds; // Seconds to add to deadling on lease extension.
-        private readonly TimeSpan _autoExtendInterval; // Delay between auto-extends for non-exactly once delivery subscriptions.
-        private readonly TimeSpan _autoExtendIntervalForExactlyOnceDelivery; // Delay between auto-extends for exactly once delivery subscriptions.
-        private readonly TimeSpan _maxExtensionDuration; // Maximum duration for which a message lease will be extended.
+        private readonly int _modifyDeadlineSeconds; // Seconds to add to deadline on lease extension.
+        private readonly int _modifyDeadlineSecondsForExactlyOnceDelivery; // Seconds to add to deadline on lease extension for exactly-once delivery subscriptions.
+        private readonly TimeSpan _autoExtendDelay; // Delay between auto-extends for non-exactly once delivery subscriptions.
+        private readonly TimeSpan _autoExtendDelayForExactlyOnceDelivery; // Delay between auto-extends for exactly once delivery subscriptions.
+        private readonly TimeSpan _maxExtensionDuration; // Maximum duration for which a message lease will be extended.        
         private readonly TimeSpan _extendQueueThrottleInterval; // Throttle pull if items in the extend queue are older than this.
+        private readonly TimeSpan _extendQueueThrottleIntervalForExactlyOnceDelivery; // Throttle pull if items in the extend queue are older than this for exactly-once delivery subscriptions.
         private readonly int _maxAckExtendQueueSize; // Soft limit on push queue sizes. Used to throttle pulls.
         private readonly int _maxAckExtendSendCount; // Maximum number of ids to include in an ack/nack/extend push RPC.
         private readonly int _maxConcurrentPush; // Mamimum number (slightly soft) of concurrent ack/nack/extend push RPCs.
@@ -194,11 +196,13 @@ public sealed partial class SubscriberClientImpl
             _softStopCts = subscriber._globalSoftStopCts;
             _subscriptionName = subscriber.SubscriptionName;
             _modifyDeadlineSeconds = subscriber._modifyDeadlineSeconds;
+            _modifyDeadlineSecondsForExactlyOnceDelivery = subscriber._modifyDeadlineSecondsForExactlyOnceDelivery;
             _maxAckExtendQueueSize = subscriber._maxAckExtendQueue;
-            _autoExtendInterval = subscriber._autoExtendInterval;
-            _autoExtendIntervalForExactlyOnceDelivery = subscriber._autoExtendIntervalForExactlyOnceDelivery;
+            _autoExtendDelay = subscriber._autoExtendDelay;
+            _autoExtendDelayForExactlyOnceDelivery = subscriber._autoExtendDelayForExactlyOnceDelivery;
             _maxExtensionDuration = subscriber._maxExtensionDuration;
-            _extendQueueThrottleInterval = TimeSpan.FromTicks((long)((TimeSpan.FromSeconds(_modifyDeadlineSeconds) - _autoExtendInterval).Ticks * 0.5));
+            _extendQueueThrottleInterval = TimeSpan.FromTicks((long)((TimeSpan.FromSeconds(_modifyDeadlineSeconds) - _autoExtendDelay).Ticks * 0.5));
+            _extendQueueThrottleIntervalForExactlyOnceDelivery = TimeSpan.FromTicks((long) ((TimeSpan.FromSeconds(_modifyDeadlineSecondsForExactlyOnceDelivery) - _autoExtendDelayForExactlyOnceDelivery).Ticks * 0.5));
             _maxAckExtendSendCount = Math.Max(10, subscriber._maxAckExtendQueue / 4);
             _maxConcurrentPush = 3; // Fairly arbitrary.
             _flow = flow;
@@ -318,7 +322,7 @@ public sealed partial class SubscriberClientImpl
             Task initTask = _pull.WriteAsync(new StreamingPullRequest
             {
                 SubscriptionAsSubscriptionName = _subscriptionName,
-                StreamAckDeadlineSeconds = _modifyDeadlineSeconds,
+                StreamAckDeadlineSeconds = _exactlyOnceDeliveryEnabled ? _modifyDeadlineSecondsForExactlyOnceDelivery : _modifyDeadlineSeconds,
                 MaxOutstandingMessages = _useLegacyFlowControl ? 0 : _flow.MaxOutstandingElementCount,
                 MaxOutstandingBytes = _useLegacyFlowControl ? 0 : _flow.MaxOutstandingByteCount
             });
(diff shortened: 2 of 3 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -81,9 +81,25 @@ public enum Reply
     /// </summary>
     public static TimeSpan MinimumAckExtensionWindow { get; } = TimeSpan.FromMilliseconds(50);
 
+    /// <summary>
+    /// The default minimum message Acknowledgement deadline extension of 0 seconds.
```
jskeet, line 85: 0 seems unlikely to be useful. We should consult with the PubSub team about this.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -81,9 +81,25 @@ public enum Reply
...
+    /// </summary>
+    public static TimeSpan DefaultMinimumAckDeadlineExtension { get; } = TimeSpan.FromSeconds(0);
+
+    /// <summary>
+    /// The default maximum message Acknowledgement deadline extension of 0 seconds.
```
jskeet, line 90: That suggests we can never extend at all - should this be an hour or something like that?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -81,9 +81,25 @@ public enum Reply
...
+
+    /// <summary>
+    /// The default minimum message Acknowledgement deadline extension of 60 seconds for exactly-once delivery subscriptions.
+    /// </summary>
+    public static TimeSpan DefaultMinimumAckDeadlineExtensionForExactlyOnceDelivery { get; } = TimeSpan.FromSeconds(60);
+
     /// <summary>
     /// The minimum message acknowledgement extension window of 60 seconds for exactly once delivery subscriptions.
     /// </summary>
+    [Obsolete("Use DefaultMinimumAckDeadlineExtensionForExactlyOnceDelivery instead.")]
```
jskeet, line 102: I'd add some more description to the summary as well.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -149,11 +149,13 @@ internal RetryInfo(DateTime firstTimeOfFailureInUtc, TimeSpan? backoff = null)
         private readonly CancellationTokenSource _pushStopCts;
         private readonly CancellationTokenSource _softStopCts;
         private readonly SubscriptionName _subscriptionName;
-        private readonly int _modifyDeadlineSeconds; // Seconds to add to deadling on lease extension.
-        private readonly TimeSpan _autoExtendInterval; // Delay between auto-extends for non-exactly once delivery subscriptions.
-        private readonly TimeSpan _autoExtendIntervalForExactlyOnceDelivery; // Delay between auto-extends for exactly once delivery subscriptions.
-        private readonly TimeSpan _maxExtensionDuration; // Maximum duration for which a message lease will be extended.
+        private readonly int _modifyDeadlineSeconds; // Seconds to add to deadline on lease extension.
```
jskeet, line 152: Given that we've got three pairs of fields, perhaps we should have a type of LeaseExtensionSettings or similar? (That could be done later.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.Settings.cs
```diff
@@ -65,6 +65,20 @@ public sealed class Settings
         /// </summary>
         public TimeSpan? AckDeadline { get; set; }
 
+        /// <summary>
```
jskeet, line 68: It feels a little odd for a minimum and maximum to be configured by the user. I'd normally expect the user to specify an AckDeadlineExtension, and for that to be constrained by a system min/max value. This may all make sense when I come to the code later, but it's work thinking about in terms of user understanding. How does this relate to `AckDeadline` above? Is there a difference between `AckDeadline` and `AckDeadlineExtension`?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -180,6 +195,43 @@ private async Task StopCompletionAsync(Task[] subscriberTasks, HashSet<Task> reg
...
+        return TimeSpan.FromTicks(Math.Max(delay.Ticks, MinimumLeaseExtensionDelay.Ticks));
+    }
+
+    /// <summary>
+    /// Returns the effective Acknowledgement Deadline based on the Maximum and/or Min Acknowledgement Deadline settings provided in the settings.
+    /// </summary>
+    /// <param name="specifiedAckDeadlineSeconds">The Ack deadline in seconds for which effective deadline is to be calculated.</param>
+    /// <param name="exactlyOnceDelivery">A boolean flag indicating if subscription is exactly-once delivery or not.</param>
+    /// <returns>The effective AckDeadline to use.</returns>
+    private TimeSpan GetEffectiveAckDeadline(int specifiedAckDeadlineSeconds, bool exactlyOnceDelivery)
```
jskeet, line 216: My feelings earlier on are still present - why would a user say "I want to extend by exactly 10 seconds, but not less than 5 seconds and not more than 20 seconds"? That basically allows them to set *any* extension value, doesn't it, just by setting the min to 0 and the max to infinity.

jskeet (commented):
Okay, I haven't reviewed much of the implementation detail yet, but that's because I'm confused about why we'd allow users to specify the min/max of a value as well as the value itself.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
```diff
@@ -59,14 +59,16 @@ internal SubscriberClientImpl(SubscriptionName subscriptionName, IEnumerable<Sub
...
-        _modifyDeadlineSeconds = (int)((settings.AckDeadline ?? DefaultAckDeadline).TotalSeconds);
-        var autoExtendInterval = TimeSpan.FromSeconds(_modifyDeadlineSeconds) - (settings.AckExtensionWindow ?? DefaultAckExtensionWindow);
-        var autoExtendIntervalForExactlyOnceDelivery = settings.AckExtensionWindow ?? MinimumAckExtensionWindowForExactlyOnceDelivery;
-        // For exactly once delivery subscription, minimum ack extension window value should be the default value of 60 seconds or the user provided value.
+        _ackExtensionWindow = settings.AckExtensionWindow ?? DefaultAckExtensionWindow;
+        _modifyDeadlineSeconds = (int) ((settings.AckDeadline ?? DefaultAckDeadline).TotalSeconds);
+        // For exactly-once delivery subscriptions, AckDeadline value should be the default value of 60 seconds or the user provided value. 
+        // Due to both DefaultAckDeadline and MinimumAckDeadlineExtensionForExactlyOnceDelivery being 60 second,
+        // AckDeadline for regular and eod subs have same value currently. But this is just coincidence. It is better to keep them seperate.
```
amanda-tarafa, line 66: I'm not sure this is right. I think that the  effective AckDeadline for eod cannot be less than 60 seconds even if the user specifies so. `MinimumAckDeadlineExtensionForExactlyOnceDelivery` is not a default value, it's a constraint on a value. So for me it would go something like:

```
var maybeInvalid = settings.Ackdealine ?? DefaultAckDeadline;
_modifyDeadlineSeconds... = maybeInvalid < MinimumAckDeadline... ? MinimumAckDeadline : maybeInvalid;
```

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
```diff
@@ -81,9 +81,18 @@ public enum Reply
     /// </summary>
     public static TimeSpan MinimumAckExtensionWindow { get; } = TimeSpan.FromMilliseconds(50);
 
+    /// <summary>
+    /// The default minimum message Acknowledgement deadline extension of 60 seconds for exactly-once delivery subscriptions.
```
amanda-tarafa, line 85: Is it default or is it minimum?
From all the internal emails, I understand it's minimum not default.

What the implications are on failing or replacing when user provided values are outside the range, is not clear to me, and I'm ultimately fine with either.

Note SubscriberClient.Settings has a Validate method that is not taking eod restrictions into account. 
That might be fine, as Settings seems to not know whether eod or non-eod, but then, we need to make those validations or replacements elsewhere.

amanda-tarafa (commented):
I've left a couple of comments, and I could still use with a top level description of the types of values we need, and how they relate to each other.
That would help a lot with the refactoring Jon is proposing and I second.

The other thing I would ask about is what happens if a subscription is change mid-flight (while being read from) from eod to non-eod or viceversa, or even if tthat is possible. Because depending on the answers things can become simpler I think.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
Rishabh-V: Sure Amanda.  I am very happy to change.

Here are my thoughts:

- For non-eod subs, ackDeadline is user provided AckDeadline or the default (which happens to be 60 seconds)

- For eod subs, ackDeadline is user provided AckDeadline or the minimum Ack Deadline (which happens to be 60 seconds).

- Earlier, because of the above defaults, I had reused the _modifyAckDeadline across both the subs. With both the defaults matching, we do not require the restart of subscriber client even if sub type is changed mid-flight.

- While prototyping this fix, I felt if tomorrow (this is hypothetical and violates YAGNI) the MinimumAckDeadline is changed to say 75 seconds and we have default AckDeadline as 60 seconds, then we will have different values of AckDeadline for eod and non-eod subs and so reusing _modifyAckDeadline will require subscriber to restart the client. Keeping them separate prevents the need to restart. 

- We have had an issue reported in which subscription is changed in-flight while the client is running. This was when the eod subs was first introduced. 

Please let me know what you think.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs
Rishabh-V: The requirement for eod is that the `MinAckDeadline` should be either:

- A user specified value in range (10, 600), or
- The default minimum Ack Deadline extension of 60 seconds if user hasn't specified the MinAckDeadline.

Now, this requirement comes in the document considering the libraries that takes the MinimumAckDeadline, MaximumAckDeadline, TotalMaxAckDeadline for lease management. They don't take AckDeadline as input from the user.

We do have default AckDeadline as 60 seconds which is what we had used earlier. But keeping in view of above requirements, I changed to use MinimumAckDeadline instead of the DefaultAckDeadline if the user has not provided AckDeadline.

Rishabh-V:
Sure Amanda. I can add top level descriptions in the next commit. This was just for prototyping so hadn't included that. I have responded to all the questions/comments. PTAL. Thanks.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.cs
Rishabh-V: Please share your thoughts and feedback on the following:
1) I have kept 2 separate AckDeadlines for regular and eod subscriptions. (With the current defaults, both will be same but wanted to get the thoughts if we should keep two AckDeadlines or just one). Due to the defaults, the same variable will work for both the subscriptions, even if subscription is changed in-flight. 
2) I have added a new MinimumAckDeadlineForExactlyOnceDelivery and marked the DefaultMinimumAckExtensionWindowForExactlyOnceDelivery as Obsolete. Given we have take only AckDeadline and (not MinimumAckDeadline) would it make sense to create this or should be just use DefaultAckDeadline (as it has same default value as DefaultMinimumAckDeadline) and mark DefaultMinimumAckExtensionWindowForExactlyOnceDelivery as Obsolete. We can always change this if there is a requirement to change the defaults.

The requirement to support Minimum/Maximum AckDeadline is only for libraries that don't allow users to specify AckDeadline. In those libraries, the AckDeadline is calculated as a percentile value based on latency and then to ensure that the AckDeadline is within range, they make use of Maximum and MinimumAckDeadline values. In our library, we already allow user to set AckDeadline which is within a range (10,600), so we don't need this.

amanda-tarafa: The tldr is

I'm totally fine with using two variables for values that have different meaning even if currently they have the same value. My comments were not about that.

My comments were about the 60 seconds minimum for EOD ack deadline. You are not using that as a minimum, you are using it as a default. In the internal emails it says clearly that should be used as a minimum. Something like "even though the C# library does not have a heuristic to calculate the deadline extension, an thus it doesn't need a min and max to anchor the heuristic result, the deadline extension is constrained by the overall EOD min extension of 60 seconds".

I'm quoting that from memory but that's effectively what the Pub/Sub team said on the last email thread that you started.

The other thing I want to know is, where is the requirement of "not restarting the subscriber if the subscription changes from EOD to non EOD or viceversa" is coming from. Is that a requirement that's explicitly coming from the Pub/Sub team?

Rishabh-V: > My comments were about the 60 seconds minimum for EOD ack deadline. You are not using that as a minimum, you are using it as a default. In the internal emails it says clearly that should be used as a minimum. Something like "even though the C# library does not have a heuristic to calculate the deadline extension, an thus it doesn't need a min and max to anchor the heuristic result, the deadline extension is constrained by the overall EOD min extension of 60 seconds".

Right Amanda. This is there in the email summary. However, this is not complete. This is the default minimum if user hasn't specified the minimumAckDeadline value. User can provide any value in range (10, 600) and override this default minimum. Since we do not have MinimumAckDeadline, so either we use the DefaultAckDeadline (which is 60 seconds) or we change it to DefaultMinimumAckDeadlineForExactlyOnceDelivery and use that. - What would be your preference?  (I will also paste the requirement in chat space for clarity)

>The other thing I want to know is, where is the requirement of "not restarting the subscriber if the subscription changes from EOD to non EOD or viceversa" is coming from. Is that a requirement that's explicitly coming from the Pub/Sub team?

Yes, this came as a bug logged by PubSub team. They explicitly stated to change the subscription in flight with subscriber on. I will share this in our chat as well. Thanks.

Please let me know your thoughts and I will update as per your suggestion/preference. Thanks.

Rishabh-V: Thanks Jon for the detailed response.

The requirement is to have default minimum AckDeadline of 60 seconds which can be overridden by the user to any value in the valid range. However, this requirement is for libraries that support minimum/maximum AckDeadline. So, I am happy naming and using the default value as DefaultAckDeadlineForExactlyOnceDelivery as suggested above. I think we can do with just one AckDeadline property in user settings. 

 The reason for using two variables for modifyAckDeadline was that we get to know if subscription is eod or not only later when we get StreamingPullResponse that sets the flag if eod is enabled.

But yes with the above code everything is achievable. Please let me know if this sounds fine and I can implement that. Thank you.

jskeet: > I think we can do with just one AckDeadline property in user settings.

So we don't need to allow users to configure different deadlines based on whether it turns out that it's EOD or not? If that's the case, that's great.

Rishabh-V:
Closing for #10371. All the feedback is incorporated there. This was just a prototype of the fix for discussion. Thanks.

PR #10462, opened 2023-05-31
feat: DS Sum/Avg changes [DO NOT MERGE]

anshuldavid13:
Design doc: [go/ds-sumavg-dotnet-sdk](http://goto.google.com/ds-sumavg-dotnet-sdk)

Protos are yet to be published to allow merging

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
@@ -18,6 +18,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Xunit;
+using static Google.Cloud.Datastore.V1.Aggregations;
 using static Google.Cloud.Datastore.V1.QueryResultBatch.Types;
 
 namespace Google.Cloud.Datastore.V1.IntegrationTests
@@ -441,7 +442,7 @@ namespace Google.Cloud.Datastore.V1.IntegrationTests
             var query = new Query("CountTestStQuery");
             AggregationQuery aggQuery = new AggregationQuery(query)
             {
-                Aggregations = { Aggregations.Count("count")}
+                Aggregations = { Count("count") }
             };
             AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
             long count = results["count"].IntegerValue;
@@ -449,9 +450,184 @@ namespace Google.Cloud.Datastore.V1.IntegrationTests
         }
 
         [Fact]
-        public void OrQueries()
+        public void Sum_WithGqlQuery()
+        {
+            var db = _fixture.DatastoreTestDb;
+
+            var gql = new GqlQuery { QueryString = "SELECT sum(age) as `sumage` FROM Students" };
+            AggregationQueryResults results = db.RunAggregationQuery(gql);
+            Assert.Equal(49, results["sumage"].IntegerValue);
+
+            var gqlD = new GqlQuery { QueryString = "SELECT sum(height) as `heightsum` FROM Students" };
+            AggregationQueryResults resultsD = db.RunAggregationQuery(gqlD);
+            Assert.Equal(18.8, resultsD["heightsum"].DoubleValue);
+        }
+
+        [Fact]
+        public void Sum_WithStructuredQuery()
+        {
+            var db = _fixture.DatastoreTestDb;
+
+            var query = new Query("Students");
+            AggregationQuery aggQuery = new AggregationQuery(query)
+            {
+                Aggregations = { Sum("age") }
+            };
+            AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
+            long sumAge = results["property_1"].IntegerValue;
+            Assert.Equal(49, sumAge);
+
+            AggregationQuery aggQueryD = new AggregationQuery(query)
+            {
+                Aggregations = { Sum("height") }
+            };
+            AggregationQueryResults resultsD = db.RunAggregationQuery(aggQueryD);
+            Assert.Equal(18.8, resultsD["property_1"].DoubleValue);
+        }
+
+        [Fact]
+        public void Sum_WithAlias()
+        {
+            var db = _fixture.DatastoreTestDb;
+
+            var query = new Query("Students");
+            AggregationQuery aggQuery = new AggregationQuery(query)
+            {
+                Aggregations = { Sum("age", "SumOfAge") }
+            };
+            AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
+            long sumAge = results["SumOfAge"].IntegerValue;
+            Assert.Equal(49, sumAge);
+
+            AggregationQuery aggQueryD = new AggregationQuery(query)
+            {
+                Aggregations = { Sum("height", "SumOfHeight") }
+            };
+            AggregationQueryResults resultsD = db.RunAggregationQuery(aggQueryD);
+            Assert.Equal(18.8, resultsD["SumOfHeight"].DoubleValue);
+        }
+
+        [Fact]
+        public void Sum_WithNaNValues()
         {
             var db = _fixture.CreateDatastoreDb();
+            var keyFactory = db.CreateKeyFactory("StudentsNaN");
+            var entities = new[]
+            {
+                new Entity { Key = keyFactory.CreateKey("1"), ["height"] = 5 },
+                new Entity { Key = keyFactory.CreateKey("2"), ["height"] = Double.NaN }
+            };
+            db.Insert(entities);
+
+            var query = new Query("StudentsNaN");
+            AggregationQuery aggQuery = new AggregationQuery(query)
+            {
+                Aggregations = { Sum("height") }
+            };
+            AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
+            Assert.Equal(Double.NaN, results["property_1"].DoubleValue);
+        }
+
+        [Fact]
+        public void Avg_WithGqlQuery()
+        {
+            var db = _fixture.DatastoreTestDb;
+
+            var gql = new GqlQuery { QueryString = "SELECT avg(age) as `avgage` FROM Students" };
+            AggregationQueryResults results = db.RunAggregationQuery(gql);
+            Assert.Equal(12.25, results["avgage"].DoubleValue);
+        }
+
+        [Fact]
+        public void Avg_WithStructuredQuery()
+        {
+            var db = _fixture.DatastoreTestDb;
+
+            var query = new Query("Students");
+            AggregationQuery aggQuery = new AggregationQuery(query)
+            {
+                Aggregations = { Average("age") }
+            };
+            AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
+            Assert.Equal(12.25, results["property_1"].DoubleValue);
+        }
+
+        [Fact]
+        public void Avg_WithAlias()
+        {
+            var db = _fixture.DatastoreTestDb;
+
+            var query = new Query("Students");
+            AggregationQuery aggQuery = new AggregationQuery(query)
+            {
+                Aggregations = { Average("height", "AvgOfHeight") }
+            };
+            AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
+            Assert.Equal(4.7, results["AvgOfHeight"].DoubleValue);
+        }
+
+        [Fact]
+        public void Avg_WithNaNValues()
+        {
+            var db = _fixture.DatastoreTestDb;
+            var keyFactory = db.CreateKeyFactory("StudentsNaN");
+            var entities = new[]
+            {
+                new Entity { Key = keyFactory.CreateKey("K"), ["height"] = 5 },
+                new Entity { Key = keyFactory.CreateKey("Z"), ["height"] = Double.NaN }
+            };
+            db.Insert(entities);
+
+            var query = new Query("StudentsNaN");
+            AggregationQuery aggQuery = new AggregationQuery(query)
+            {
+                Aggregations = { Average("height", "Avg_height") }
+            };
+            AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
+            Assert.Equal(Double.NaN, results["Avg_height"].DoubleValue);
+        }
+
+        [Fact]
+        public void MultipleAggregations()
+        {
+            var db = _fixture.DatastoreTestDb;
+
+            var query = new Query("Students");
+            AggregationQuery aggQuery = new AggregationQuery(query)
+            {
+                Aggregations = { Sum("age", "Sum_age"), Average("age", "Avg_age"), Count("Count") }
+            };
+            AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
+            Assert.Equal(49, results["Sum_age"].IntegerValue);
+            Assert.Equal(12.25, results["Avg_age"].DoubleValue);
+            Assert.Equal(4, results["Count"].IntegerValue);
+        }
+
+        [Fact]
+        public void Aggregation_NonNumericValue()
+        {
+            var db = _fixture.DatastoreTestDb;
+            var keyFactory = db.CreateKeyFactory("StudentsNN");
+            var entities = new[]
+            {
+                new Entity { Key = keyFactory.CreateKey("A"), ["height"] = 5 },
+                new Entity { Key = keyFactory.CreateKey("D"), ["height"] = 4 }
+            };
+            db.Insert(entities);
+
+            var query = new Query("StudentsNN");
+            AggregationQuery aggQuery = new AggregationQuery(query)
+            {
+                Aggregations = { Sum("Key", "nonNum") }
+            };
+            AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
+            Assert.Equal(0, results["nonNum"].IntegerValue);
+        }
+
+        [Fact]
+        public void OrQueries()
+        {
+            var db = _fixture.DatastoreTestDb;
             var keyFactory = db.CreateKeyFactory("OrQuery");
             var entities = new[]
             {
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
@@ -450,184 +450,54 @@ namespace Google.Cloud.Datastore.V1.IntegrationTests
         }
 
         [Fact]
-        public void Sum_WithGqlQuery()
-        {
-            var db = _fixture.DatastoreTestDb;
-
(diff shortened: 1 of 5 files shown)


hemanshv:
LGTM

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
```diff
@@ -441,17 +442,192 @@ public void Count_WithStructuredQuery()
             var query = new Query("CountTestStQuery");
             AggregationQuery aggQuery = new AggregationQuery(query)
             {
-                Aggregations = { Aggregations.Count("count")}
+                Aggregations = { Count("count")}
```
jskeet, line 445: Nit: space after the `)`

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
```diff
@@ -441,17 +442,192 @@ public void Count_WithStructuredQuery()
...
             };
             AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
             long count = results["count"].IntegerValue;
             Assert.Equal(2, count);
         }
 
         [Fact]
-        public void OrQueries()
+        public void Sum_WithGqlQuery()
```
jskeet, line 453: I really don't think we need all of these tests. I would create *one* test for a structured query, and *one* test for GQL, specifying two sums and an average with some mixture of aliased and not aliased. It's just a general integration test to make sure everything basically works... we don't need to be exhaustive, given that most of what's being tested is the server rather than the client.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreFixture.cs
```diff
@@ -31,12 +31,30 @@ public sealed class DatastoreFixture : CloudProjectFixtureBase, ICollectionFixtu
...
 
+        public DatastoreDb DatastoreTestDb { get; }
+
         public string NamespaceId { get; }
         public PartitionId PartitionId => new PartitionId { ProjectId = ProjectId, NamespaceId = NamespaceId };
 
         public DatastoreFixture()
         {
             NamespaceId = IdGenerator.FromDateTime(prefix: "test-");
+            DatastoreTestDb = CreateTestStudents();
```
jskeet, line 42: I would remove all the changes from DatastoreFixture - with the above suggestion with only one test, we can populate the data in the test itself, as we do for the other integration tests.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
```diff
@@ -32,4 +32,56 @@ public static class Aggregations
...
+    /// as a double if it cannot fit within a 64-bit signed integer. When this
+    /// occurs, the returned value will lose precision.
+    ///
+    /// * When underflow occurs, floating-point aggregation is non-deterministic.
+    /// This means that running the same query repeatedly without any changes to
+    /// the underlying values could produce slightly different results each
+    /// time. In those cases, values should be stored as integers over
+    /// floating-point numbers.
+    /// </summary>
+    /// <param name="property">Property for which Sum is to calculated</param>
```
jskeet, line 58: Property for which the sum is to be calculated

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
```diff
@@ -32,4 +32,56 @@ public static class Aggregations
...
+    /// <param name="alias">A string used to retrieve the result of this aggregation.
+    /// If not provided, Datastore will pick a default name following the format `property_&lt;incremental_id++>`
+    /// eg. property_1
```
jskeet, line 61: e.g.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
```diff
@@ -32,4 +32,56 @@ public static class Aggregations
...
+    ///
+    /// * When underflow occurs, floating-point aggregation is non-deterministic.
+    /// This means that running the same query repeatedly without any changes to
+    /// the underlying values could produce slightly different results each
+    /// time. In those cases, values should be stored as integers over
+    /// floating-point numbers.
+    /// </summary>
+    /// <param name="property">Property for which Sum is to calculated</param>
+    /// <param name="alias">A string used to retrieve the result of this aggregation.
+    /// If not provided, Datastore will pick a default name following the format `property_&lt;incremental_id++>`
```
jskeet, line 60: Again, XML comment format rather than Markdown - `<c>` is appropriate here.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
```diff
@@ -32,4 +32,56 @@ public static class Aggregations
...
+    /// eg. property_1
+    /// </param>
+    /// <returns>A Sum(property) aggregation.</returns>
+    public static Aggregation Sum(string property, string alias = "") =>
+        new() { Sum = new Sum() { Property = new PropertyReference(property) }, Alias = alias };
```
jskeet, line 65: Might as well use target-typed new in both places:

```csharp
new() { Sum = new() { Property = new PropertyReference(property) }, Alias = alias };
```

Or if you *do* want to keep `new Sum`, remove the `()` after it to be consistent with our style for object initializers.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
```diff
@@ -32,4 +32,56 @@ public static class Aggregations
     /// <returns>A count(*) aggregation.</returns>
     public static Aggregation Count(string alias = "") =>
         new Aggregation { Count = new Count(), Alias = alias };
+
+    /// <summary>
```
jskeet, line 36: I'm slightly surprised we don't have a unit test for Count already. Given how trivial each of these methods is, I'm not overly bothered - but it wouldn't hurt to have a couple of unit tests for each method (with and without alias, basically).

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
```diff
@@ -32,4 +32,56 @@ public static class Aggregations
...
+    /// Sum of the values of the requested property.
```
jskeet, line 37: Please use XML documentation formatting instead of Markdown for lists and paragraphs. It will show up better in VS.

See https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.LongRunning/Google.LongRunning/Operation.cs#L25 for an example.

If there's a Datastore product documentation page with all this information, I'd actually prefer a link to that, so that we don't need to worry about keeping this up to date.

(This all applies to Sum as well.)

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
anshuldavid13: do think that for the trivial amount that we have added in client libraries for this, the integration tests already cover everything that needs to be tested and we shouldn't just be adding more test for the sake of it, as is already done in count.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
```diff
@@ -441,13 +442,58 @@ public void Count_WithStructuredQuery()
...
+                new Entity { Key = keyFactory.CreateKey("7"), ["age"] = 12, ["height"] = 4.6  },
+                new Entity { Key = keyFactory.CreateKey("8"), ["age"] = 14, ["height"] = 4  },
+                new Entity { Key = keyFactory.CreateKey("9"), ["age"] = 11, ["height"] = 5.2  }
+            };
+            db.Insert(entities);
+
+            var query = new Query("Student");
+            AggregationQuery aggQuery = new AggregationQuery(query)
+            {
+                Aggregations = { Sum("age", "Sum_age"), Average("age"), Count("Count") }
```
jskeet, line 489: Super-nit: everywhere else, our aliases are lower-case. I'd suggest `sum_age` and `count`, assuming that's the convention that's normally see in Datastore docs etc. (Likewise, we've got `avgheight` above - is using an underscore conventional or not?)

Not massively important of course, as this is only a test rather than a sample, but it's good to avoid distracting inconsistencies even in tests.

(These two tests are definitely better than the previous large set though - thanks.)

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
jskeet: I'm not going to insist on it, but I do generally like to see unit tests for things that are *so* easy to test like this. (See FilterTest for examples. I'm happy with light integration test coverage *because* we generally have good unit test coverage - meaning that integration tests can really be about *integration*: "I'm confident in the proto I'm sending due to the unit tests, now I can test that that has the effect I expect."

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
```diff
@@ -32,4 +33,73 @@ public static class Aggregations
...
+    ///     This means that running the same query repeatedly without any changes to
+    ///     the underlying values could produce slightly different results each
+    ///     time. In those cases, values should be stored as integers over
+    ///     floating-point numbers.</description>
+    ///   </item>
+    /// </list>
+    /// </summary>
+    /// <param name="property">Property for which the sum is to be calculated.</param>
+    /// <param name="alias">A string used to retrieve the result of this aggregation.
+    /// If not provided, Datastore will pick a default name following the format `property_incremental_id++`
```
jskeet, line 70: Nit: we've still got the backticks here.

How about:

```text
If not provided, Datastore will pick a default name with a prefix of <c>property_</c>
followed by an auto-incremented ID, e.g. <c>property_1</c>.
```

jskeet (commented):
Just a couple more nits.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreFixture.cs
```diff
@@ -1,4 +1,4 @@
-﻿// Copyright 2016 Google Inc. All Rights Reserved.
+// Copyright 2016 Google Inc. All Rights Reserved.
```
amanda-tarafa, line 1: nit: Maybe fully revert the changes to this file.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
amanda-tarafa: I have to agree with Jon here, in particular with:

> I'm happy with light integration test coverage because we generally have good unit test coverage- meaning that integration tests can really be about integration

For instance, with no unit tests, if the integration test start failing, it's harder to know whether that's a service failure or our own code failure. In general, we should have thorough unit testing and only representative integration testing.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
amanda-tarafa: +1 for waiting for Datastore to have public documentation around Avg and Sum and link that here. This is the kind of documentation that tends to get out of date really quickly.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreTransactionTest.cs
```diff
@@ -152,27 +153,32 @@ private async Task CommitTest(Func<DatastoreTransaction, Task<CommitResponse>> c
...
+                new Entity { Key = keyFactory.CreateKey("11"), ["age"] = 12, ["height"] = 5  },
+                new Entity { Key = keyFactory.CreateKey("21"), ["age"] = 12, ["height"] = 4.6  },
+                new Entity { Key = keyFactory.CreateKey("31"), ["age"] = 14, ["height"] = 4  },
+                new Entity { Key = keyFactory.CreateKey("41"), ["age"] = 11, ["height"] = 5.2  }
             };
             db.Insert(entities);
             using var transaction = db.BeginTransaction();
-            var gqlQuery = new GqlQuery { QueryString = "SELECT count(*)  as `count` FROM CountTestTransaction " };
-            var query = new Query("CountTestTransaction");
+            var gqlQuery = new GqlQuery { QueryString = "SELECT count(*)  as `count` FROM AggTestTransaction " };
```
amanda-tarafa, line 169: We probably want the gqlQuery to have sum and avg as well.

And I would prefere these tests as separate ones, one for GQL and one for AggregationQuery (even for before when the only aggregation was count).

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
amanda-tarafa: +1 to being consistent with alias naming to avoid unnecesary differences. And in particular `avgage` is pretty hard to read.

And as for the tests, I would have the ones in the classes `DatastoreDbTest` and `DatastoreTransactionTest` have the same structure:

- One GQL test that includes the 3 aggregations
- One structured test that includes the 3 aggregations

This is because:

- Clarity. Unnecesary differences are distracting and distractions hinder maintenability. When we land here in a year, we'll be asking ourselves why the tests in `DatastoreDbTest` and `DatastoreTransactionTest` are structured differently when they could be structured the same.
- Testing practices: The 2 types of queries (GQL and structured) are implemented differently, so we should test for them separately. The 3 types of aggregations are implemented the same (for the same type of query) so it's OK that we test them together, at least in integration tests (unit tests should definetely test them separately as well).

amanda-tarafa (changes requested):

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreFixture.cs
anshuldavid13: I already did, this is git issue.

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
```diff
@@ -32,4 +33,73 @@ public static class Aggregations
     /// <returns>A count(*) aggregation.</returns>
     public static Aggregation Count(string alias = "") =>
         new Aggregation { Count = new Count(), Alias = alias };
+
```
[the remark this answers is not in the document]
anshuldavid13: Done

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreTransactionTest.cs
anshuldavid13: Done

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/Aggregations.cs
anshuldavid13: Done

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.Tests/AggregationsTest.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Collections.Generic;
+using Xunit;
+using static Google.Cloud.Datastore.V1.AggregationQuery.Types;
+using static Google.Cloud.Datastore.V1.Aggregations;
+
+namespace Google.Cloud.Datastore.V1.Tests;
+public class AggregationsTest
+{
+    [Fact]
+    public void AggregationTest()
```
jskeet, line 24: This isn't really testing what Average and Sum do at all.

To see what I mean, try change your Average and Sum implementations to do the wrong thing - to ignore the alias, or even make Average use the Sum operator and vice versa. This test is currently just equivalent to:

```csharp
Assert.Equal(Average("age"), Average("age"));
Assert.Equal(Sum("height"), Sum("height"));
```

jskeet (commented):
I think this is okay to stay until the docs and protos have been published, at which point presumably it will compile, and we can replace the long comments with a link to the product docs.

amanda-tarafa (commented):
Looks good to me as well. Just the bit to add a link to public docs when that's possible.
Thanks!

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1.IntegrationTests/DatastoreDbTest.cs
```diff
@@ -441,13 +442,58 @@ public void Count_WithStructuredQuery()
...
             };
             AggregationQueryResults results = db.RunAggregationQuery(aggQuery);
             long count = results["count"].IntegerValue;
             Assert.Equal(2, count);
         }
 
+        [Fact]
+        public void Aggregation_WithGqlQuery()
+        {
+            var db = _fixture.CreateDatastoreDb();
```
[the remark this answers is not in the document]
anshuldavid13: Regarding the "same" query , I have used age and height in sum in different methods for variations in tests to check sum returning both integer and double values , just like we are doing for alias and non alias testing and hence they are not exactly the "same" in both.
I will actually remove the other column from the dataset to make it cleaner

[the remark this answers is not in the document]
anshuldavid13: Including aggregation on multiple columns in the same query adds the overhead of creating "composite indexes" for these "kinds". Currently I dont have anyway of doing it via code, we will have to create the index in a yaml file and set locally for running now and then for running these tests on CI, we will have to set it in that DB (whosoever has rights). Hence I just separated it out in  different queries as the final result achieved was the same.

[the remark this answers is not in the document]
jskeet: Ah, okay - it might be worth adding a comment with that info. Definitely not worth adding a composite index for that.

jskeet:
Closing in favour of #10580.

PR #10464, opened 2023-05-31
feat: Add support for glob pattern storage

hemanshv:
Adds support for glob storage in storage ListObjects api. This will resolve https://github.com/googleapis/google-cloud-dotnet/issues/10460.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListObjectsTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListObjectsTest.cs
@@ -72,6 +72,17 @@ namespace Google.Cloud.Storage.V1.IntegrationTests
             await AssertObjects(prefix, options, expectedNames.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries));
         }
 
+        [Theory]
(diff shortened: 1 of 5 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListObjectsTest.cs
```diff
@@ -72,6 +72,17 @@ public async Task PrefixAndDelimiter(string prefix, string expectedNames)
             await AssertObjects(prefix, options, expectedNames.Split(new[] { ',' }, StringSplitOptions.RemoveEmptyEntries));
         }
 
+        [Theory]
+        [InlineData("la*", "large.txt")]
+        [InlineData("a/*.txt", "a/o1.txt,a/o2.txt")]
+        [InlineData("a/x/*.txt", "a/x/o3.txt,a/x/o4.txt")]
+        public void MatchGlob(string globPattern, string expectedNames)
```
jskeet, line 79: I believe if you make this 

```csharp
public void MatchGlob(string globPattern, params string[] expectedNames)
```

you can just list the expected names directly, without needing to split on commas in `AssertObjectNames`.

jskeet (approved):
LGTM - one possible simplification for a test, but that's all. Thanks for the swift turnaround :)

PR #10467, opened 2023-06-01
test: [wip] test transaction commit behavior

olavloite:
Test whether committing a transaction is enough to return the session to the pool.


amanda-tarafa:
@olavloite FYI: Integration tests are not run on presubmits. So, if you were relying on this PR to run the new test, that's not happening at the moment.

olavloite:
> @olavloite FYI: Integration tests are not run on presubmits. So, if you were relying on this PR to run the new test, that's not happening at the moment.

@amanda-tarafa Thanks for the heads up. I did not realize that (in the sense that I assumed that the integration tests on the emulator would be run). I verified that the test succeeds locally on the emulator. Is there any way to trigger that on a PR?

amanda-tarafa:
> > @olavloite FYI: Integration tests are not run on presubmits. So, if you were relying on this PR to run the new test, that's not happening at the moment.
> 
> @amanda-tarafa Thanks for the heads up. I did not realize that (in the sense that I assumed that the integration tests on the emulator would be run). I verified that the test succeeds locally on the emulator. Is there any way to trigger that on a PR?

@olavloite my bad, I should have been more explicit. The test does run and has run on the emulator. But it doesn't run against the service (which is probably irrelevant for this particular test).

olavloite:
> > > @olavloite FYI: Integration tests are not run on presubmits. So, if you were relying on this PR to run the new test, that's not happening at the moment.
> > 
> > 
> > @amanda-tarafa Thanks for the heads up. I did not realize that (in the sense that I assumed that the integration tests on the emulator would be run). I verified that the test succeeds locally on the emulator. Is there any way to trigger that on a PR?
> 
> @olavloite my bad, I should have been more explicit. The test does run and has run on the emulator. But it doesn't run against the service (which is probably irrelevant for this particular test).

Ah, good to hear. And indeed; for this test, running against the emulator is enough.

jskeet:
Is there a plan for the future of this PR? No urgent pressure, I just don't want it hanging around forever :)

olavloite:
Replaced by #10673

PR #10495, opened 2023-06-07
feat: Generate (but don't release) Google.Cloud.MigrationCenter.V1

jskeet:


amanda-tarafa:
There's a build failure though.

apis/apis.json
```diff
@@ -3025,6 +3025,7 @@
       "shortName": "migrationcenter",
       "serviceConfigFile": "migrationcenter_v1.yaml",
       "restNumericEnums": true
```
amanda-tarafa, line 3027: Missing comma, which fails the build.

jskeet: Yup, have repushed. And it looks like we should be able to release fairly soon.

PR #10496, opened 2023-06-07
fix: Makes transactions marked as DisposedBehaviour.Detach actually detach from the session pool.

amanda-tarafa:
The test I added was red before the fix (because of an unexpected active session) and it's now green.

I'm running the rest of the tests locally now.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
@@ -59,6 +59,8 @@ namespace Google.Cloud.Spanner.Data
     public sealed class SpannerTransaction : SpannerTransactionBase, ISpannerTransaction
     {
         private readonly List<Mutation> _mutations = new List<Mutation>();
+        // This value will be true if and only if this transaction was created by RetriableTransaction.
+        private readonly bool _isRetriable = false;
         private DisposeBehavior _disposeBehavior = DisposeBehavior.ReleaseToPool;
         private bool _disposed = false;
 
@@ -192,7 +194,8 @@ namespace Google.Cloud.Spanner.Data
             SpannerConnection connection,
             TransactionMode mode,
             PooledSession session,
-            TimestampBound timestampBound)
+            TimestampBound timestampBound,
+            bool isRetriable)
         {
             SpannerConnection = GaxPreconditions.CheckNotNull(connection, nameof(connection));
             CommitTimeout = SpannerConnection.Builder.Timeout;
@@ -200,6 +203,7 @@ namespace Google.Cloud.Spanner.Data
             Mode = mode;
             _session = GaxPreconditions.CheckNotNull(session, nameof(session));
             TimestampBound = timestampBound;
+            _isRetriable = isRetriable;
         }
 
         /// <summary>
@@ -460,6 +464,19 @@ namespace Google.Cloud.Spanner.Data
                 return;
             }
             _disposed = true;
+
+            if (_isRetriable)
+            {
+                // If this transaction is being used by RetriableTransaction, we want to dispose of this instance
+                // but we don't want to do anything with the session, as the RetriableTransaction will attempt to
+                // reuse it with a fresh transaction.
+                // If acquiring a fresh transaction with the existing session fails, the session will be disposed
+                // and a new one with a fresh transaction will be obtained.
+                // If acquiring a fresh transaction succeeds, then the session will be disposed after the RetriableTransaction
+                // succeeds or we have stopped retrying.
+                return;
+            }
+
             switch (DisposeBehavior)
             {
                 case DisposeBehavior.CloseResources:
@@ -475,8 +492,16 @@ namespace Google.Cloud.Spanner.Data
                     }
                     _session.ReleaseToPool(forceDelete: false);
                     break;
+                case DisposeBehavior.Detach:
+                    // A detached transaction is expected to be explicitly shared across processes.
+                    // So we don't release it back to the pool, but we need to mark it as disposed and stop counting it as active
+                    // if it was created in the targeted pool. The first time a detached transaction is created, it will be created
+                    // in the targeted pool, subsequent times (as the transaction ID is used) it will be created in the detached pool
+                    // which doesn't keep track of active transactions/sessions.
+                    _session.DetachFromPool();
+                    break;
                 default:
-                    // Default for detached or unknown DisposeBehavior is to do nothing.
+                    // Default for unknown DisposeBehavior is to do nothing.
                     break;
             }
         }
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2018 Google LLC
+// Copyright 2018 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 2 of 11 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.ISessionPool.cs
```diff
@@ -30,6 +30,7 @@ internal interface ISessionPool
             IClock Clock { get; }
             Task<PooledSession> WithFreshTransactionOrNewAsync(PooledSession session, TransactionOptions transactionOptions, CancellationToken cancellationToken);
             void Release(PooledSession session, ByteString transactionToRollback, bool deleteSession);
+            void Detach(PooledSession session);
```
jskeet, line 33: None of the implementations use the session - that sounds like it's either a bug (probably in TargetedSessionPool) or that we can remove the parameter.

jskeet (commented):
I'll need to think about this more - probably opening up the code. That may well be tomorrow.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -59,6 +59,8 @@ private protected SpannerTransactionBase()
     public sealed class SpannerTransaction : SpannerTransactionBase, ISpannerTransaction
     {
         private readonly List<Mutation> _mutations = new List<Mutation>();
+        // true if this transaction is used by RetriableTransaction.
```
jskeet, line 62: "Is used by" or "Was created by"?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -460,6 +464,15 @@ protected override void Dispose(bool disposing)
                 return;
             }
             _disposed = true;
+
+            if (_isRetriable)
```
jskeet, line 468: What *will* return the session back to the pool?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.ISessionPool.cs
jskeet: Ping on this one.

amanda-tarafa: It's not a bug, TargetetSessionPool does not keep track of PooledSession instances themselves, just through counters.

None of this is public so we can definetely remove the parameter. But then I think I'd like to call this method something like DetachActiveSession. Does that sound good?

amanda-tarafa (commented):
Review comments addressed in second commit. Thanks!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -460,6 +464,19 @@ protected override void Dispose(bool disposing)
...
+            {
+                // If this transaction is being used by RetriableTransaction, we want to dispose of this instance
+                // but we don't want to do anything with the session, as the RetriableTransaction will attempt to
+                // reuse it with a fresh transaction.
+                // If acquiring a fresh transaction with the existing session fail, the session will be disposed
+                // and a new one with a fresh transaction will be obtained.
+                // If acquiting a fresh transaction succeeds, then the session will be diposed after the RetriableTransaction
```
jskeet, line 475: Nit: diposed => disposed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
jskeet: Thanks, that's much clearer :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.ISessionPool.cs
jskeet: Either that, or we could document that the parameter is *currently* unused, but could be useful for debugging. (It feels like it's a bit of a coincidence that at the moment we don't need to say which session is being detached.)

jskeet (approved):
I *think* all this looks good, but I'd wait for @olavloite to approve as well.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
amanda-tarafa: Done, and also acquiting => acquiring.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.ISessionPool.cs
amanda-tarafa: Yes, I like leaving the parameter in but documenting we don't use it. Indeed, it's totally a detail implementaiton that we don't need to use the actual session we are detaching, to detach it.

amanda-tarafa (commented):
Comments address, thanks! Will wait for @olavloite review as well.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.ISessionPool.cs
```diff
@@ -30,6 +30,14 @@ internal interface ISessionPool
             IClock Clock { get; }
             Task<PooledSession> WithFreshTransactionOrNewAsync(PooledSession session, TransactionOptions transactionOptions, CancellationToken cancellationToken);
             void Release(PooledSession session, ByteString transactionToRollback, bool deleteSession);
+            /// <summary>
+            /// Detaches the given session from the pool, meaning the pool stops tracking this session as active,
+            /// but does nothing else, in particular it doesn't return the session to the pool or attempts to delete it.
```
olavloite, line 35:
```suggestion
            /// but does nothing else, in particular it doesn't return the session to the pool or attempt to delete it.
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -231,6 +231,34 @@ public void ReleaseToPool(bool forceDelete)
...
+        /// <see cref="SessionPool.CreateDetachedSession(SessionName, ByteString, ModeOneofCase)"/> (or an overload) to create an instance
+        /// of <see cref="PooledSession"/> representing the shared transaction and then call <see cref="ReleaseToPool(bool)"/> passing true
+        /// to force session deletion and clean up resources.
+        /// Else, the application can rely on Spaner service garbage collection to clean up this session once it becomes stale.
+        /// </remarks>
+        public void DetachFromPool()
+        {
+            if (MarkAsDisposed())
+            {
+                _pool.Detach(AfterReset());
```
olavloite, line 254: Why are we calling `AfterReset()` here? Won't that remove the transaction from the session, which effectively makes it unusable from this point on? Or am I missing something here?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -460,6 +464,19 @@ protected override void Dispose(bool disposing)
                 return;
             }
             _disposed = true;
+
+            if (_isRetriable)
+            {
+                // If this transaction is being used by RetriableTransaction, we want to dispose of this instance
+                // but we don't want to do anything with the session, as the RetriableTransaction will attempt to
+                // reuse it with a fresh transaction.
+                // If acquiring a fresh transaction with the existing session fail, the session will be disposed
```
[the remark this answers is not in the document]
amanda-tarafa: Done.

amanda-tarafa (commented):
@olavloite I've addressed your comments, PTAL, thanks!

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
olavloite: (Sorry for the slow response, I'm on-call this week).

Sorry, I thought this method was called already when you would call `Transaction.DisposeBehavior = Detach`. That is not correct, it is called when the transaction is disposed. And then it is obvious that you can't use the transaction anymore. Thanks for the explanation.

olavloite (approved):

PR #10497, opened 2023-06-07
chore: Run all smoke tests instead of stopping on failure

jskeet:


jskeet:
(The first commit is just plumbing to allow us to return an exit code.)

PR #10505, opened 2023-06-09
tests: Skips DiscoveryEngine tests that continue to fail

amanda-tarafa:


jskeet (approved):
Hmm... that's interesting - the existing smoke tests pass for me, with just the one skipped test. Quite possibly an allow-list somewhere. Happy to skip these for now though.

PR #10539, opened 2023-06-16
WIP: topics for rest library docs

jskeet:
This is just a starting point for comments on topics I've no doubt missed.

It includes *most* (but not deliberately all) of what's in https://github.com/googleapis/google-api-dotnet-client/blob/main/FAQ.md - once this is all published, we should update that.

### DIFF
--- a/docs/devsite-help/legacy-rest-libraries.md
+++ b/docs/devsite-help/legacy-rest-libraries.md
@@ -1,0 +1,35 @@
+# Legacy HTTP/1.1 + REST libraries
+
+As described in the [library types page](library-types.md), the legacy
+HTTP/1.1 + REST libraries are API-specific libraries with source code in
+the [google-api-dotnet-client repository](https://github.com/googleapis/google-api-dotnet-client),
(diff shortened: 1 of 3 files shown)


docs/devsite-help/legacy-rest-libraries.md
```diff
@@ -0,0 +1,35 @@
+# Legacy HTTP/1.1 + REST libraries
```
amanda-tarafa, line 1: If I were to use YouTube or Drive, I would be slightly scared of the legacy word here. Not certain how to address that, but maybe simply aknowledging that these are the only libraries for such APIs is fine.

amanda-tarafa (commented):
Yep, this looks good

docs/devsite-help/legacy-rest-libraries.md
jskeet: That's a good point. I think we can still keep the term in the library types doc, but I'll remove it from here. I can mention that these are effectively legacy for GCP.

jskeet:
Closing as a PR now, as it's far from ready for review.

PR #10580, opened 2023-06-26
feat: Add sum and average aggregates

jskeet:
(This is currently the same code as #10462; will review and modify accordingly when the feature is released, before merging.)


jskeet:
Will investigate why the CLA bot has fired here - that's very odd...

jskeet:
This may now be ready to release - need to check that the tests pass, and consult the product team.

jskeet:
Integration tests still fail with:

> 'SUM' aggregations are not supported.

hemanshv (approved):

PR #10534, opened 2023-06-15
feat: Firestore sum avg change

anshuldavid13:
Design doc: V3 changes in [design doc](https://docs.google.com/document/d/1xsYj_kraMwSe2Q2qSdO57oWWdnSjlIrO-ZDW296zYOs/edit?pli=1#heading=h.yj2y1op9kxa1) 

Note: I have made minimal changes to count implementation in this. Will take another pass to make all relevant FS count changes. This PR is focused on the FS V3 Sum & Avg implementation

PR #10581, opened 2023-06-26
feat: Add sum and average aggregates

jskeet:
(This is currently the same code as #10534; I'll be making changes to the first commit as per the review on that PR before merging.)


jskeet:
Amended each commit to make myself the author, to keep CLA-bot happy.

jskeet:
Work still to do:

- Check that the feature works
- Handle long field names
- Review/fix integration tests (e.g. `Students` class)

jskeet:
@amanda-tarafa This is now ready for review. One commit at a time is probably beneficial - happy to receive suggestions as to how we want to break it up for actual merging.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuerySnapshot.cs
```diff
@@ -80,6 +80,9 @@ public T GetValue<T>(AggregateField aggregateField)
         return GetValue<T>(aggregateField.Alias);
     }
 
+    // Note: TryGetValue is internal right now, as we're not 100% convinced we want it in the public API.
```
amanda-tarafa, line 83: Hmmm, I'd leave it in, after all users can always catch the exception from GetValue and wrap it in their own TryGetValue.
I think this is different from when were talking about returning null given a non-existent alias.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/AggregateQueryTest.cs
```diff
@@ -109,4 +109,14 @@ public async Task MultipleAggregations()
         Assert.Equal(Student.Data.Average(c => c.MathScore), snapshot.GetValue<double>("avg_score"));
     }
 
+    [Fact]
+    public async Task NestedAggregations()
+    {
+        CollectionReference collection = _fixture.StudentCollection;
+        var snapshot = await collection.Aggregate(AggregateField.Sum("Nested.Math"), Average("Nested.Math", "avg_score"), Count()).GetSnapshotAsync();
```
amanda-tarafa, line 116: I'd also like to check that the alias "Sum_Nested.Math" works, which is the one we would generate by default, right? I mean, I'm not certain if there are restrictions on how the alias can look like (apart from length). I tried to find it but couldn't.

amanda-tarafa (approved):
This looks good.

As for how to split it, I'd say that the few refactoring commits can remain, the rest, feature and tests, in a single commit I think.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateField.cs
```diff
@@ -0,0 +1,85 @@
...
+    /// </summary>
+    /// <param name="field">The field on which the sum aggregation is performed. Must not be null.</param>
+    /// <param name="alias" cref="Aggregation.Alias">Optional alias to identify the result of the aggregation in <see cref="AggregateQuerySnapshot"/>.
+    /// If this is null, an alias of the form "Sum_{field}" is used.
+    /// </param>
+    /// <returns>Returns a new instance of <see cref="AggregateField"/>.</returns>
+    public static AggregateField Sum(string field, string alias = null)
+    {
+        GaxPreconditions.CheckNotNull(field, nameof(field));
+        return new(new Aggregation { Sum = new Sum { Field = FieldPath.FromDotSeparatedString(field).ToFieldReference() }, Alias = alias ?? $"Sum_{field}" });
```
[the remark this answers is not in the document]
jskeet: Discussed offline: this is now irrelevant, as we no longer use the aliases in RPCs. (It *was* relevant in the first commit, which was being reviewed...)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -424,7 +423,21 @@ public Query Select(params FieldPath[] fieldPaths)
         /// </summary>
         /// <returns>An instance of <see cref="AggregateQuery"/> with count(*) aggregation applied.</returns>
         public AggregateQuery Count() =>
```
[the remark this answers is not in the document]
jskeet: Discussed offline: we'll leave this as it is.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuerySnapshot.cs
jskeet: Discussed offline, and we'll keep it internal for now.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/AggregateQueryTest.cs
jskeet: I went to put that in, and we've already got that :)

```csharp
Assert.Equal(Student.Data.Sum(c => c.Nested.Math), snapshot.GetValue<double>("Sum_Nested.Math"));
```

amanda-tarafa (approved):
All good!

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/AggregateQueryTest.cs
```diff
@@ -0,0 +1,122 @@
...
+using Google.Cloud.Firestore.V1;
+using System.Linq;
+using System.Threading.Tasks;
+using Xunit;
+using static Google.Cloud.Firestore.AggregateField;
+
+namespace Google.Cloud.Firestore.IntegrationTests;
+
+[Collection(nameof(FirestoreFixture))]
+public class AggregateQueryTest
```
MarkDuckworth, line 25: Depending on the result, both `snapshot.GetValue<double>(...)` and `snapshot.GetValue<long>(...)` can be lossy. It might be worth adding a few tests to determine if the caller can get the precise result using `snapshot.GetValue<object>(...)` (or similar).

```
// `a` sums to a value that cannot be represented precisely as Int64
// `b` sums to an integer value that cannot be represented precisely as a double
addDoc({a: 0.1, b: (Int)2^53})
addDoc({a: 0.1, b: 1})
s = collection.aggregate(sum("a"), sum("b"))

var sumA = s.GetValue<Object>("sum_a")
Assert.Equal(sumA.GetType(), typeof(Double));

// Expected result from the server is an integer only if all input
// is integer values and the result does not overflow Int64
var sumB = s.GetValue<Object>("sum_b")
Assert.Equal(sumB.GetType(), typeof(Double));

File: apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQuerySnapshotTest.cs [hunk 6]
```diff
@@ -23,54 +26,147 @@ public class AggregateQuerySnapshotTest
...
+        var aggQuerySnapshot = new AggregateQuerySnapshot(aggQuery, sampleReadTime, data);
+
+        // Alias key
+        Assert.Equal(20, aggQuerySnapshot.GetValue<long>("Sum_bar"));
+
+        // AggregateField key
+        Assert.Equal(20, aggQuerySnapshot.GetValue<long>(Sum("bar")));
+
+        // Object type output
+        Assert.Equal(20L, aggQuerySnapshot.GetValue<object>(Sum("bar")));
```
MarkDuckworth, line 119: This is probably sufficient to cover my previous comment about testing `snap.GetValue<object>(...)`, but I'll leave that comment in case you also want to add an integration test.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQueryTest.cs
```diff
@@ -53,37 +78,117 @@ public async Task GetSnapshotAsync_VerifySnapshotMembers()
...
         });
         mock.Configure().RunAggregationQuery(request, Arg.Any<CallSettings>()).Returns(response);
+
         var aggregateQuery = query.Count();
         var snapshot = await aggregateQuery.GetSnapshotAsync();
+
+        Assert.Equal(aggregateQuery, snapshot.Query);
+        Assert.Equal(Timestamp.FromProto(sampleReadTime), snapshot.ReadTime);
+        // We didn't provide a count in the response.
+        Assert.Null(snapshot.Count);
```
MarkDuckworth, line 96: I think we wrote this case to throw, so it would have consistent behavior with `snapshot.GetValue<long>("Count")`. This might be worth further discussion.

MarkDuckworth (approved):
LGTM. I left a few comments for your consideration.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/AggregateQueryTest.cs
jskeet: Right - yes, that makes sense to add as another integration test.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/AggregateQueryTest.cs
```diff
@@ -0,0 +1,122 @@
...
+    [Fact]
+    public async Task AvgWithFilter()
+    {
+        CollectionReference collection = _fixture.HighScoreCollection;
+        var snapshot = await collection.WhereGreaterThan("Level", 20).Aggregate(Average("Level")).GetSnapshotAsync();
+        Assert.Equal(HighScore.Data.Where(x => x.Level > 20).Average(c => c.Level), snapshot.GetValue<double>("Avg_Level"));
+    }
+
+    [Fact]
+    public async Task MultipleAggregations()
```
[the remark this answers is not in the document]
jskeet: I'd prefer not to add that as an integration test, otherwise we'll get a failure if that limit is increased later.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQuerySnapshotTest.cs
jskeet: My reason for adding an integration test would be as a reminder to ourselves that the "obvious" type isn't always used as the underlying result type.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQueryTest.cs
jskeet: This is an unfortunate result of us not meeting when designing Count - the Unity SDK has a non-nullable property; the .NET SDK has a nullable property. I think we should continue in the spirit of that discrepancy rather than fixing it by having a nullable property which never returns null. Happy to discuss further if you feel strongly about it though.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/AggregateQuery.cs
```diff
@@ -16,43 +16,38 @@
...
-    internal AggregateQuery WithAggregation(Aggregation aggregation)
-    {
-        var newAggregations = new List<Aggregation>(_aggregations);
-        newAggregations.Add(aggregation);
-        return new AggregateQuery(_query, newAggregations);
+        var aliasSet = new HashSet<string>(StringComparer.Ordinal);
+        foreach (var field in aggregateFields)
+        {
+            GaxPreconditions.CheckArgument(field is not null, nameof(aggregateFields), "All aggregate fields must be non-null.");
+            GaxPreconditions.CheckArgument(aliasSet.Add(field.Alias), nameof(aggregateFields), "All aliases in an aggregate query must be unique.");
```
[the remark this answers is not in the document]
jskeet: Right, good to know - but yes, I'll keep the implementation as it is for simplicity.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
```diff
@@ -424,7 +424,35 @@ public Query Select(params FieldPath[] fieldPaths)
...
+        /// <summary>
+        /// Calculates the specified aggregations to return the aggregate query. Multiple aggregations can be performed in a single query.
+        /// </summary>
+        /// <param name="aggregateField">Specifies the <see cref="AggregateField"/> to be calculated. Must not be null.</param>
+        /// <param name="aggregateFields">Additional aggregations to be included in the query, if any. Must not be null, but may be empty.</param>
+        /// <returns>Returns an <see cref="AggregateQuery" /> that performs aggregations on the documents in the result set of this query.</returns>
+        public AggregateQuery Aggregate(AggregateField aggregateField, params AggregateField[] aggregateFields)
+        {
+            GaxPreconditions.CheckNotNull(aggregateField, nameof(aggregateField));
+            GaxPreconditions.CheckNotNull(aggregateFields, nameof(aggregateFields));
```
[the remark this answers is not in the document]
jskeet: Just by calling `query.Aggregate(Count(), null)` for example.

The compiler only creates an array for a `params` parameter when the compile-time type of the argument isn't implicitly convertible to the array type.

jskeet:
Right - now down to one main commit, and then one just to remove using directives. (The refactoring of moving existing tests to a new file would have been tricky to extract as a new commit, and I don't think it's worth doing.)

There may be some changes required when the feature is GA - I haven't run the integration tests against a live service. But otherwise, I think this is ready to go.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/Query.cs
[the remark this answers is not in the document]
MarkDuckworth: Right. Thanks for the reminder.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/AggregateQueryTest.cs
MarkDuckworth: I'm fine continuing as is.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.IntegrationTests/FirestoreFixture.cs
```diff
@@ -60,32 +65,57 @@ public class FirestoreFixture : CloudProjectFixtureBase, ICollectionFixture<Fire
...
             HighScoreCollection = FirestoreDb.Collection(CollectionPrefix + "-high-scores");
+            StudentCollection = FirestoreDb.Collection(CollectionPrefix + "-students");
             ArrayQueryCollection = FirestoreDb.Collection(CollectionPrefix + "-array-query");
             CollectionGroupQueryCollection = FirestoreDb.Collection(CollectionPrefix + "-collection-groups");
             Task.Run(PopulateCollections).Wait();
         }
 
         private async Task PopulateCollections()
         {
+            await CreateIndex(HighScoreCollection, AscendingField("Score"), AscendingField("Level"));
```
jskeet, line 92: As an aside, we might want to parallelize these later...

amanda-tarafa (approved):
Yep, looks good.

PR #10587, opened 2023-06-27
refactor: Convert DeserializationContext to interface IDeserializationContext

jskeet:
Previously, we've only needed a deserialization context for a DocumentSnapshot, so it's been fine to create a wrapper around that. However, when AggregateQuerySnapshot is used for deserialization (for sum/average), it makes more sense to make the snapshot the context itself, implementing an internal interface. This can deliberately throw exceptions if it's used in the wrong place: if you try to deserialize an aggregate query snapshot to a .NET type which is decorated with an attribute to be auto-populated with a document ID, for example, it's appropriate for that to fail.

The ability to deserialize using an AggregateQuerySnapshot is not used in this commit; it's preparing for future functionality.


jskeet:
(@amanda-tarafa once you're happy with this, I'll ask the Unity folks to take a look as well.)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDeserializationConfigurationAttribute.cs
```diff
@@ -53,12 +53,12 @@ public sealed class FirestoreDeserializationConfigurationAttribute : Attribute
...
         /// <param name="context">The current deserialization context.</param>
         /// <param name="map">The map to (possibly) update.</param>
-        internal void ApplyContext(DeserializationContext context, Dictionary<string, object> map)
+        internal void ApplyContext(IDeserializationContext context, Dictionary<string, object> map)
         {
             MaybePopulate(DocumentIdKey, context.DocumentReference);
-            MaybePopulate(CreateTimestampKey, context.Snapshot.CreateTime);
-            MaybePopulate(UpdateTimestampKey, context.Snapshot.UpdateTime);
-            MaybePopulate(ReadTimestampKey, context.Snapshot.ReadTime);
+            MaybePopulate(CreateTimestampKey, context.CreateTime);
```
amanda-tarafa, line 59: I think this needs to be something like:
```suggestion
            MaybePopulate(CreateTimestampKey, () => context.CreateTime);
```

And then the function below:

```
void MaybePopulate(string key, Func<object> valueGetter)
{
    ...
}
```

to defer attempting to obtain the value until we know the key is not null. Else this will always fail even if the attribute is not configured to deserialize the non-general values?

But I might be missing something here.

amanda-tarafa (approved):
I have a specific question on the implementation, but also:

Having read the docs in [FirestoreDocumentCreateTimestampAttribute.cs](https://github.com/googleapis/google-cloud-dotnet/blob/b606f4c5293289554f5a035f3f8e0ce4c71ef3fe/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDocumentCreateTimestampAttribute.cs#L28), I do wonder if we should apply the same logic to aggregate snapshots that we do to missing document snapshots, i.e. not fail on this document only fields.
I'm imagining the use case where the user wants to represent the aggregates of an entity set using the same type as the one for the entity itself?

But I'm not certain that my understanding of this is correct, so take this with a pinch of salt.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDeserializationConfigurationAttribute.cs
jskeet: Ooh, you're absolutely right. Will write a test first, then fix it.

jskeet:
> Having read the docs in FirestoreDocumentCreateTimestampAttribute I do wonder if we should apply the same logic to aggregate snapshots that we do to missing document snapshots, i.e. not fail on this document only fields.

I wondered about that too. We definitely *could* say that those properties always return null for non-documents. We *could* do the same for DocumentId as well. I think I generally prefer the fail-fast approach, but let's see what the Unity team thinks.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDeserializationConfigurationAttribute.cs
jskeet: Done now in a second commit - PTAL.

amanda-tarafa (approved):
Yep, LGTM, thanks for the test as well.

jskeet:
Great, thanks - will contact the Unity SDK team to see what they think.

PR #10604, opened 2023-06-29
docs: Initial iteration of a help page for REST libraries

jskeet:
I removed the general credential section that was going to be included, as we have a reference to the OAuth docs. We might want a quick cheat-sheet of methods on GoogleCredential, but that can come later.

### DIFF
--- a/docs/devsite-help/library-types.md
+++ b/docs/devsite-help/library-types.md
@@ -137,3 +137,6 @@ Examples include:
 
 (In general, packages starting "Google.Apis" followed by a product
 name and a version number fall into this category.)
+
+More information about various aspects of these libraries can be found in the
+[dedicated page](rest-libraries.md).
--- a/docs/devsite-help/rest-libraries.md
+++ b/docs/devsite-help/rest-libraries.md
@@ -1,0 +1,165 @@
+# HTTP/1.1 + REST libraries
+
+As described in the [library types page](library-types.md), the
+HTTP/1.1 + REST libraries are API-specific libraries with source code in
+the [google-api-dotnet-client repository](https://github.com/googleapis/google-api-dotnet-client),
+generated from [Discovery documents](https://developers.google.com/discovery/v1/getting_started)
+instead of API protobuf representations. API calls are made over HTTP/1.1 with
+JSON content.
+
+These libraries depend heavily on the support libraries of Google.Apis and
+Google.Apis.Core, which have their source code in the same repository.
+
+These libraries are still supported and automatically updated as the API
+surfaces change, but the underlying code and generator is not being actively worked
+on, and only high-impact feature requests are considered.
+
+Some features are not as discoverable as we'd like them to be; this page
+describes the use of these features.
+
+## Service object reuse
+
+Service client classes (e.g. `DriveService`) are thread-safe and reusable.
+Typically your application should create a single instance and use it for all operations.
+
+By default, every instantiation of a service client class creates a new instance of
+[HttpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient),
+and creating many of these is, in general, a bad idea. See, for example,
+[this post](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/) for why.
+
+Sometimes it *is* required to create an instance per request. In this case make sure
+your code calls `Dispose` on each instance when it has finished with it.
+This will call `Dispose` on the `HttpClient`.
+
+### GoogleWebAuthorizationBroker
+
+[GoogleWebAuthorizationBroker](https://github.com/googleapis/google-api-dotnet-client/blob/main/Src/Support/Google.Apis.Auth/OAuth2/GoogleWebAuthorizationBroker.cs) requires interaction with the end-user, so can only be used in client-side code.
+
+OAuth authentication requires the end-user to interact with a web browser, which can only be done client-side.
+
+It is possible to use OAuth authentication for end users in a web application, but not using
+`GoogleWebAuthorizationBroker`. For these purposes, you can use one of the ASP.NET integration packages
+we provide, such as [`Google.Apis.Auth.AspNetCore3`](https://www.nuget.org/packages/Google.Apis.Auth.AspNetCore3).
+
+You can find general information about Google APIs and OAuth 2.0 in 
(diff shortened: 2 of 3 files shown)


docs/devsite-help/rest-libraries.md
```diff
@@ -0,0 +1,168 @@
...
+
+By default, every instantiation of a service client class creates a new instance of
+[HttpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient),
+and creating many of these is, in general, a bad idea. See, for example,
+[this post](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/) for why.
+
+Every instantiation of a client class also uses more CPU time and requires more memory allocation,
+which are generally best avoided if possible.
+
+Sometimes it *is* required to create an instance per request; for example, if different
```
amanda-tarafa, line 33: Actually, Apiary does cleanly support per-call credentials.

So I wonder if it's best to say, create a single client, and if you need to pass different credentials per call, do this...

docs/devsite-help/rest-libraries.md
```diff
@@ -0,0 +1,168 @@
...
+Service client classes (e.g. `StorageService`) are thread-safe and reusable.
+Typically your application should create a single instance and use it for all operations.
+
+By default, every instantiation of a service client class creates a new instance of
+[HttpClient](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient),
+and creating many of these is, in general, a bad idea. See, for example,
+[this post](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/) for why.
+
+Every instantiation of a client class also uses more CPU time and requires more memory allocation,
+which are generally best avoided if possible.
```
amanda-tarafa, line 31: I'm not sure we'd need to justify why? We don't justify this for GAPICs, so, maybe just: "we recommend you use a single instace of a service client"?

docs/devsite-help/rest-libraries.md
```diff
@@ -0,0 +1,168 @@
...
+## Integration with IHttpClientFactory
+
+For historical reasons, there are two (incompatible) `IHttpClientFactory` interfaces:
+
+- [`Google.Apis.Http.IHttpClientFactory`](https://github.com/googleapis/google-api-dotnet-client/blob/main/Src/Support/Google.Apis.Core/Http/IHttpClientFactory.cs) in the REST support library
+- [`System.Net.Http.IHttpClientFactory`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.ihttpclientfactory) in .NET
+
+The .NET interface is widely used, and supported in the community with projects such as
+[Polly](https://github.com/App-vNext/Polly). Unfortunately, the REST support library predates this
+interface and it's can't be directly retro-fitted without breaking changes.
```
amanda-tarafa, line 90:
```suggestion
interface and it can't be directly retro-fitted without breaking changes.
```

docs/devsite-help/rest-libraries.md
```diff
@@ -0,0 +1,168 @@
...
+
+The `HttpClientFromMessageHandlerFactory` provides a bridge between the two interfaces. Integration is still
```
amanda-tarafa, line 92:
```suggestion
The `HttpClientFromMessageHandlerFactory` provides a bridge between the two interfaces. Usage is still
```

docs/devsite-help/library-types.md
```diff
@@ -136,4 +136,7 @@ Examples include:
 - Google.Apis.YouTube.v3
 
 (In general, packages starting "Google.Apis" followed by a product
-name and a version number fall into this category.)
\ No newline at end of file
+name and a version number fall into this category.)
+
+More information about various aspects of these libraries can be found in the
+[dedicated page](legacy-rest-libraries.md).
```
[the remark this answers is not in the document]
jskeet: Done locally.

docs/devsite-help/rest-libraries.md
```diff
@@ -0,0 +1,168 @@
...
+These libraries are still supported and automatically updated as the API
+surfaces change, but the underlying code and generator is not being actively worked
+on, and only high-impact feature requests are considered.
+
+Some features are not as discoverable as we'd like them to be; this page
+describes the use of these features.
+
+## Service object reuse
+
+Service client classes (e.g. `StorageService`) are thread-safe and reusable.
```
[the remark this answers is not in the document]
jskeet: Good plan. Done locally.

docs/devsite-help/rest-libraries.md
jskeet: This was copied from the FAQ, I believe :) But I agree - deleted. (I think it's fine to include the details of it creating a new HttpClient, but not the paragraph starting at line 30.) We've already got the recommendation on line 23.

docs/devsite-help/rest-libraries.md
jskeet: I haven't included the "and if you need to pass different credentials per call" part as I wasn't sure exactly what you meant, but I've just removed the example. (It does beg the question about the circumstances for when that's appropriate though.)

docs/devsite-help/rest-libraries.md
jskeet: Thanks - again, from the FAQ. (We should delete all the duplicate content and refer to this page once it's published.)

docs/devsite-help/rest-libraries.md
jskeet: I prefer integration here - if the issue is the duplication of "integration" later on the sentence, I think I'd probably prefer to remove it there rather than change this. Happy to chat about that or other options though.

jskeet (commented):
Changes in a second commit. Let's talk about the remaining aspects in our meeting.

docs/devsite-help/rest-libraries.md
amanda-tarafa: Yep, it's because of the duplication of integration, but I think it's actually fine, it really makes the most sense as it is.

amanda-tarafa (approved):

docs/devsite-help/rest-libraries.md
amanda-tarafa: (Couldn't include this as part of the review).

The most common case for needing a different client is when you need different credentials. But in the case of Apiary, service clients support per call credentials cleanly. I think we can add a section on "Per call credentials" (now, or later) and here simply not specify other corner cases for which you may need different clients.

PR #10626, opened 2023-07-07
chore: Disambiguate test cases in Spanner integration tests

jskeet:
I believe XUnit just uses a string representation of each theory argument to create test IDs. DateTime doesn't include the "kind" in the string representation (or in equality) so test cases that vary only by kind are deemed duplicates.

Two test cases were removed as they *were* duplicates (just different ways of creating the exact same value).


jskeet:
I'm rerunning the full integration test suite to check there aren't others, but I *think* this is all that's needed.

amanda-tarafa (approved):
Oh that's interesting.

PR #10630, opened 2023-07-10
feat: BigTable benchmarking script

hemanshv:
Adds BigTable benchmarking script with Raw stubs.

### DIFF
--- a/issues/BigtablePerfBenchmarking/BigtablePerfBenchmarking.csproj
+++ b/issues/BigtablePerfBenchmarking/BigtablePerfBenchmarking.csproj
@@ -8,7 +8,6 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Google.Cloud.Bigtable.Common.V2" Version="3.0.0" />
     <PackageReference Include="Google.Cloud.Bigtable.V2" Version="3.6.0" />
   </ItemGroup>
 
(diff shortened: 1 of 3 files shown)


hemanshv:
Hi @jskeet ,
I have added the bigtable benchmarking script over here. Please have a look and share your feedback.

Thanks,

issues/BigtablePerfBenchmarking/BigtablePerfBenchmarking.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Google.Cloud.Bigtable.Common.V2" Version="3.0.0" />
```
jskeet, line 11: I don't think we should need this reference - it should already be implicit by virtue of having Google.Cloud.Bigtable.V2

jskeet (approved):
It's *possible* that we'll want to expand this to a solution with multiple projects testing different things in the future, but I think this is fine for now.

issues/BigtablePerfBenchmarking/BigtablePerfBenchmarking.csproj
hemanshv: done

PR #10810, opened 2023-08-07
chore: Update to xunit 2.5.1

jskeet:
This allows us to revert some of the changes we made when updating to 2.5.0.

I've verified this works with 2.5.1-pre.12 - we just need to wait for 2.5.1 to be released.


amanda-tarafa (approved):
LGTMing so that it's ready when xunit 2.5.1 is out.

PR #10885, opened 2023-08-16
chore: First part of removing Moq from Diagnostics testing

jskeet:
One chunk at a time... I'm finding it much easier to understand the tests when they're written without mocks, personally.

These two commits represent about half of the uses of mocks within Diagnostics - I'm hoping to get to the other half tomorrow.


amanda-tarafa (approved):
LGTM. Yep, I also like tests better with the fake.

PR #10900, opened 2023-08-18
chore: Fix diagnostics by removing Moq

### LINKED ISSUE #10896: Build is broken having removed Moq
I'd not taken into account that the AspNetCore test projects depend on the Common test project...

Basically, we've got more Moq to remove.

I think we can probably live with the broken state until tomorrow morning - I'll fix it then.

jskeet:
Fixes #10896.


jskeet:
(Merging before @amanda-tarafa reviews, to fix the build. We can make any other changes later.)

amanda-tarafa:
Just for the record, this all looks good.

PR #10915, opened 2023-08-22
chore: update autorelease.sh to use the new publish reporter tool

hemanshv:
This updates the autorelease.sh script to use trigger in house(.Net) release publish reporter tool.

### DIFF
--- a/.kokoro/autorelease.sh
+++ b/.kokoro/autorelease.sh
@@ -15,12 +15,7 @@ populate_all_secrets
 python -m pip install --require-hashes -r pip-requirements.txt
 python -m pip install --require-hashes -r requirements.txt
 
-# The publish reporter script uses "python3" which doesn't exist on Windows.
-# Work out what we should use instead, and fix up the script.
-python -m releasetool publish-reporter-script > /tmp/publisher-script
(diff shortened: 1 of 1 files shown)


.kokoro/autorelease.sh
```diff
@@ -15,12 +15,7 @@ populate_all_secrets
 python -m pip install --require-hashes -r pip-requirements.txt
 python -m pip install --require-hashes -r requirements.txt
 
-# The publish reporter script uses "python3" which doesn't exist on Windows.
-# Work out what we should use instead, and fix up the script.
-python -m releasetool publish-reporter-script > /tmp/publisher-script
-PYTHON3=$(source ../toolversions.sh && echo $PYTHON3)
-sed -i "s/python3/$PYTHON3/g" /tmp/publisher-script
-source /tmp/publisher-script
+../tools/Google.Cloud.Tools.ReleaseTool/publish_reporter.sh
```
jskeet, line 18: I suspect this should use "source" - otherwise it'll start a new shell and report when *that* completes.

hemanshv: Done

PR #10956, opened 2023-08-30
chore: Remove python packages entirely

jskeet:
We previously used docuploader and releasetool; these are both now reimplemented in .NET.


jskeet:
This PR should only be merged after we've successfully done a package release with the new tools. But it's good to get some review before then.

amanda-tarafa (approved):
Great!! LGTM

jskeet:
A release has succeeded, so we're good to go.

PR #10980, opened 2023-09-01
chore: Remove last vestige of Python

jskeet:
Tested by running build.sh (but cancelling pretty quickly) and generateapis.sh (all the way through).


amanda-tarafa (approved):

jskeet:
(Next week we should release something to check that part still works...)

PR #11024, opened 2023-09-12
chore: Remove trailing whitespace from all handwritten files

jskeet:
As a test, `git show --ignore-space-at-eol` shows no changes.


jskeet:
(This change *doesn't* rewrite all files to avoid having a BOM at the start of the file - but any file that was going to be edited anyway is "fixed" in that way.)

amanda-tarafa (approved):

PR #11043, opened 2023-09-18
chore: Update xUnit to 2.5.1

jskeet:
@amanda-tarafa: Please review the last three commits.


jskeet:
Humbug - it looks like the changes for collection assertions are still needed (which is weird, as I thought that had been fixed). Removing the first commit...

amanda-tarafa (approved):

PR #11066, opened 2023-09-20
fix: Make Log4Net type searching more robust

jskeet:
After recent updates (possibly .NET SDK 6.0.414; possibly dependencies) tests have been failing on net462, due to a loader error in this code. It turns out that while `GetAssemblies()` returns, calling `GetTypes()` on some of those assemblies would fail. This change isolates each `Assembly.GetTypes()` call, so that the search can progress to other assemblies.


amanda-tarafa (approved):
Oh wow

PR #11116, opened 2023-09-29
poc: Deferring client creation

amanda-tarafa:
This is a proof of concept so it's incomplete.
I've sent an email about this.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherServiceApiClient.g.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherServiceApiClient.g.cs
@@ -30,6 +30,10 @@ using scg = System.Collections.Generic;
 using sco = System.Collections.ObjectModel;
 using st = System.Threading;
 using stt = System.Threading.Tasks;
+using System;
+using Google.Api.Gax;
+using Grpc.Core;
+using System.Threading.Tasks;
 
 namespace Google.Cloud.PubSub.V1
 {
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherServiceApiClient.g.cs
```diff
@@ -1825,6 +1842,54 @@ public override DetachSubscriptionResponse DetachSubscription(DetachSubscription
...
+        private readonly Lazy<stt::Task<(PublisherServiceApiClient, ChannelBase)>> _publisherServiceApiClientTask;
+
+        internal PublisherServiceApiClientDeferred(Func<stt::Task<(PublisherServiceApiClient, ChannelBase)>> publisherServiceApiClientProvider) =>
+            _publisherServiceApiClientTask = new Lazy<stt.Task<(PublisherServiceApiClient, ChannelBase)>>(publisherServiceApiClientProvider);
+
+        /// <inheritdoc/>
+        public override Topic CreateTopic(Topic request, gaxgrpc::CallSettings callSettings = null)
+        {
+            // Note:  this is blocking if this is the first operation ever called on this client.
+            // But so is the sync call anyway.
```
jskeet, line 1857: The sync call is blocking, but I don't think it's blocking in the same way as this. (We'd need to check what actual gRPC implementations do.) I think we need a lot of caution here, and testing in apps with single-thread synchronization contexts. (Using a sync RPC method in a WPF/WinForms thread is a really idea of course, but if it currently works, even badly, we don't want to change that behavior to be a deadlock.)

amanda-tarafa: [Grpc.Net.Client blocks in this same way](https://github.com/grpc/grpc-dotnet/blob/7e62218646dfe16ef0a98802886bb6a7f5507a80/src/Grpc.Net.Client/Internal/HttpClientCallInvoker.cs#L154) as far as I understand it. But I'm not certain about Grpc.Core.

But yes, regardless of how we do this, we need to be cautios with sync calls.

jskeet:
Closing (without deleting the branch) as we've done what we wanted to.

PR #11135, opened 2023-10-03
chore: Create indexes in parallel

jskeet:
(On my box, for .NET 6 integration tests, this chops the time down from 9m 24s to 5m 25s... so definitely worth doing. There's one other CreateIndex call, but that's for a collection only used in a single test, so it's harder to parallelize.)


jskeet:
I was going to leave this until after the release, but it turned out to be sufficiently easy that I thought I might as well do it now.

amanda-tarafa (approved):

Issue #11092, opened 2023-09-26
Using built-in DI registration functions like `AddPublisherServiceApiClient` can potentially cause thread starvation during high load at startup.


flagbug:
Using the built-in DI registration functions like `AddPublisherServiceApiClient` can potentially cause thread starvation during high load at startup.

#### Environment details

  - OS: Debian 11 in Docker
  - .NET version: 7.0
  - Package name and version: `Google.Cloud.PubSub.V1->3.2.0`
  - Running on Google Cloud Run

#### Steps to reproduce

Unfortunately there is no way to reliably reproduce it, we've encountered this issue only twice in the past two months

1. Create a basic ASP.NET Core with the DI registration code in `Program.cs`: `builder.Services.AddPublisherServiceApiClient();`

2. Add a basic controller that accepts an `PublisherServiceApiClient` in the constructor and publishes Pub/Sub messages using `await PublisherServiceApiClient.PublishAsync`
3. Send a ton of traffic over a longer period of time. 

When a new instance starts (e.g due to scaling), under rare circumstances this can cause the exact same issue as described here https://github.com/googleapis/google-api-dotnet-client/issues/1807, where the fetching of the Application Default Credentials might cause thread starvation.

To summarize this previous issue:
- Soon after application startup, you start receiving requests that try to obtain the Application Default Credentials from DI.
- All but one of the request thread blocks until the singleton is ready, which will be ready once the default credential is fetched. 
- More requests keep coming in, and keep blocking on either of the DI singletons. The thread pool starts to deplete.
- The "lucky" thread that went on to fetch the credential gets to a point where it needs to check whether we are running on Compute, and for that an HTTP request is made to the metadata server. This is an async operation that doesn't block the thread, which is freed to do other things (probably attent to yet another incoming request).
- Still more requests keep coming in and more threads keep getting blocked waiting for the credential to be fetched, stressing the thread pool even more. The thread pool is probably growing at this point, but very slowly at first. Requests start being queued up, and some may start timing out.
- When the async HTTP request to the metadata server ends it needs a thread to resume back into, and it's now competing for a thread from a starved thread pool. Basically, we need a thread to finish fetching the credential, while most threads are blocked waiting for the credential to be fetched.
- Eventually the thread pool grows enough, the credential fetching can resume (if under the time we timeout at else, you get the error you initially reported), and the rest of the threads unblock with time.

[...]

PR #11142, opened 2023-10-04
docs: Document thread starvation scenario workaround.

amanda-tarafa:
Towards #11092

### DIFF
--- a/docs/devsite-help/client-lifecycle.md
+++ b/docs/devsite-help/client-lifecycle.md
@@ -101,6 +101,28 @@ builder.Services
     .AddExampleClient();
 ```
 
+#### Dependency injection in high-load-at-startup environments.
+
+If your application uses Application Default Credentials while running in a Google runtime (Compute Engine,
+GKE, Cloud Functions, etc.) and you expect high request load on application startup you may run into a known
+issue that leads to thread starvation and high response latency. We are looking into a permanent fix,
+but in the meantime you can use the following workaround: make certain you request Application
+Default Credentials **during** startup configuration (you don't have to use them). For example:
+
+```csharp
+var builder = WebApplication.CreateBuilder(args);
+builder.Services
+    .AddRazorPages()
+    .AddExampleClient();
+// Workaround for possible thread starvation scenario.
+GoogleCredential.GetApplicationDefaultCredential();
+```
+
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.PubSub.V1/docs/index.md
```diff
@@ -82,76 +82,6 @@ restart listening for messages with `StartAsync(...)` again. Due to the expense
 instance, it is recommended that a singleton client per topic is used for the lifetime of the
 application.
 
-## Dependency Injection
```
jskeet, line 85: I'm not sure about this - the AddSubscriberClient and AddPublisherClient methods do have slightly different overloads to the pure-GAPIC ones. I suspect we don't need as much detail as we currently have, but I think some is important - including the aspects about automatic stopping.

docs/devsite-help/client-lifecycle.md
```diff
@@ -101,6 +101,28 @@ builder.Services
     .AddExampleClient();
 ```
 
+#### Dependency injection in high-load-at-startup environments.
+
+If your application uses Application Default Credentials while running in a Google runtime (Compute,
```
jskeet, line 106: Compute Engine instead of just Compute?

docs/devsite-help/client-lifecycle.md
```diff
@@ -101,6 +101,28 @@ builder.Services
...
+issue that leads to thread starvation and high response latency. We are looking into a permanent fix,
+but in the meantime you can use the following workaround: make certain you request Application
+Default Credentials **during** startup configuration (you don't have to use them). For example:
+
+```csharp
+var builder = WebApplication.CreateBuilder(args);
+builder.Services
+    .AddRazorPages()
+    .AddExampleClient();
+// Workarounf for possible thread starvation scenario.
```
jskeet, line 117: Workarounf => Workaround

jskeet (commented):
Can we reduce this PR to just the second commit for now, and then consider the first commit separately?

docs/devsite-help/client-lifecycle.md
amanda-tarafa: Done.

amanda-tarafa (commented):
I've updated this PR to be only about the library wide changes in docs.

jskeet (approved):

PR #11147, opened 2023-10-05
Some of Inline Begin Transaction changes

anuragsrivstv:
PooledSession changes in ILB.

### DIFF
--- a/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
+++ b/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
@@ -18,7 +18,7 @@ using Google.Cloud.Spanner.V1.Internal;
 using Google.Protobuf;
 using Google.Protobuf.WellKnownTypes;
 using System;
-using System.Runtime.CompilerServices;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using static Google.Cloud.Spanner.V1.TransactionOptions;
@@ -42,7 +42,7 @@ namespace Google.Cloud.Spanner.V1
     /// </remarks>
     public sealed class PooledSession : IDisposable
     {
-        private readonly Session _session;
+        internal Session Session { get; }
 
         /// <summary>
         /// The name of the session. This is never null.
@@ -52,23 +52,22 @@ namespace Google.Cloud.Spanner.V1
         /// <summary>
         /// The ID of the transaction. May be null.
         /// </summary>
-        public ByteString TransactionId { get; }
+        public ByteString TransactionId { get; set; }
 
         /// <summary>
-        /// The mode of the transaction. (Always None iff TransactionId is null.)
+        /// The options of the transaction.
         /// </summary>
-        internal ModeOneofCase TransactionMode { get; }
+        public TransactionOptions TransactionOptions { get; set; }
 
         /// <summary>
-        /// The read timestamp of the transaction. (Always <c>null</c> if
-        /// ReturnReadTimestamp = false or if TransactionMode != ReadOnly.)
+        /// The read timestamp of the transaction. (Always <c>null</c> if ReturnReadTimestamp = false)
         /// </summary>
         public Timestamp ReadTimestamp { get; }
 
         /// <summary>
         /// Indicates whether the server has told us that the session has expired.
         /// </summary>
-        internal bool ServerExpired => _session.Expired;
+        internal bool ServerExpired => Session.Expired;
 
         /// <summary>
         /// The time (in ticks since 0001-01-01T00:00:00Z) at which to refresh this session.
@@ -84,6 +83,11 @@ namespace Google.Cloud.Spanner.V1
         /// </summary>
         private readonly DateTime _evictionTime;
 
+        /// <summary>
+        /// Create a single instance of the semaphore with a value of 1. It means only one thread will be allowed access at a time.
+        /// </summary>
+        private static readonly SemaphoreSlim _semaphoreSlim = new SemaphoreSlim(1, 1);
+
         // Just for convenience...
         private SpannerClient Client => _pool.Client;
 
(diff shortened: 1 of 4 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -52,23 +52,22 @@ public sealed class PooledSession : IDisposable
...
-        /// The mode of the transaction. (Always None iff TransactionId is null.)
+        /// The options of the transaction.
         /// </summary>
-        internal ModeOneofCase TransactionMode { get; }
+        public TransactionOptions TransactionOptions { get; set; }
 
         /// <summary>
-        /// The read timestamp of the transaction. (Always <c>null</c> if
-        /// ReturnReadTimestamp = false or if TransactionMode != ReadOnly.)
+        /// The read timestamp of the transaction. (Always <c>null</c> if ReturnReadTimestamp = false)
```
anuragsrivstv, line 63: ...Or if TransactionOptions is not of ReadOnly type.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -42,7 +42,7 @@ namespace Google.Cloud.Spanner.V1
     /// </remarks>
     public sealed class PooledSession : IDisposable
     {
-        private readonly Session _session;
+        internal Session Session { get; }
```
anuragsrivstv, line 45: Made it internal because now we are passing entire PooledSession object in ResultStream instead of directly this session instance. Therefore we now need to access this via PooledSession object in ResultStream class.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ResultStream.cs
```diff
@@ -50,7 +51,7 @@ internal sealed class ResultStream : IAsyncStreamReader<PartialResultSet>, IDisp
         private readonly LinkedList<PartialResultSet> _buffer;
         private readonly SpannerClient _client;
         private readonly ReadOrQueryRequest _request;
-        private readonly Session _session;
+        private readonly PooledSession _session;
```
anuragsrivstv, line 54: Reason for this change: Because we need to invoke ExecuteWithTransactionSelector from PooledSession Class.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -467,5 +498,95 @@ private void CheckNotDisposed()
...
+
+            // Null Transaction Id means a new trasaction needs to inline begin/started.
+            // Only one thread should inline begin the transction.
+
+            var enteredSemaphoreSlim = false; // Intially false for all threads.
+            if (TransactionId == null)
+            {
+                await _semaphoreSlim.WaitAsync().ConfigureAwait(false); // All except one thrad will wait here for semaphore to release.
+                enteredSemaphoreSlim = true; // True for first thread trying to inline begin.
+            }
```
amanda-tarafa, line 538: I don't think this works as you think it does:

Say 5 commands are all executed at "the same time". For all of them TransactionId will be null so all of them will attempt to grab the semaphore. Only one of them will grab the semaphore.

That one will execute and inline the transaction. In the meantime, 3 more commands were executed, and for all of those TransactionId was null because the first command that inlined the transaction is not done yet. You now have 7 commands waiting for the semaphore.

Now th first command returns, sets TransactionId to some value and releases the semaphore. One of the 7 commands waiting for the semaphore grabs it, while the other 6 continue to be blocked there.

That one command that grabbed the semaphore gets to execute with the new transaction Id and after execution, releases the semaphore, so one of the 6 remaining gets to execute, and then one of the 5 remaining, etc.

The problem is that after the first command that inlined the transaction returns, all 7 waiting commandas can start executing in parallel, and this implementation doesn't allow that.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -270,9 +282,16 @@ public async Task<CommitResponse> CommitAsync(CommitRequest request, CallSetting
         {
             CheckNotDisposed();
             GaxPreconditions.CheckNotNull(request, nameof(request));
-            GaxPreconditions.CheckState(TransactionId != null, "Cannot commit a PooledSession with no associated transaction");
+            var transactionId = TransactionId;
+            if (transactionId == null)
```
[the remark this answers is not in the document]
jskeet: And given that that involves making a request, it can't just be a matter of taking a lock. I think we need to take a step back and work out the situations in which this can reasonably be called multiple times from different threads. We could potentially prevent that from happening (safely remember whether commit has been called, and throw an exception if it has) which would at least make things easier. But I don't know whether that's reasonable.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
anuragsrivstv: Yes, that's true for all the commands executed before TransactionId generation. 
So the scenario you have explained, to handle that we can release semaphore from the thread as soon as transactionId is not null for next threads.
That is adding if block at Line number 541 like below:
```
// Transaction was aready inline begin by another thread in parallel while this was on hold. 
            if (TransactionId != null)
            {
             if (enteredSemaphoreSlim)
                {
                    _semaphoreSlim.Release();
                    enteredSemaphoreSlim=false;
                }
                return await asyncWork(new TransactionSelector { Id = TransactionId }).ConfigureAwait(false);
            }
            ```

anuragsrivstv: Do you think it'll work ?

amanda-tarafa: Not really. Yes, the blocked commands don't have to wait for the other ex-blocked commands to finish executing, but still, they have to wait to acquiere the semaphore. Basically, we'll have waiting commands that don't need to be waiting, and that's not good even if for a small amount of time.

amanda-tarafa:
Closing as we'll be going in a different direction.

PR #11167, opened 2023-10-12
docs: Update ConfigurationModifier property documentation

jskeet:
(This is initially just one property, but once we've tweaked it until we're happy, I'll roll it out to all the equivalent properties.)

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadJsonOptions.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadJsonOptions.cs
@@ -77,8 +77,17 @@ namespace Google.Cloud.BigQuery.V2
         public SchemaUpdateOption? DestinationSchemaUpdateOptions { get; set; }
 
         /// <summary>
-        /// Delegate that is used to modify the underlying <see cref="JobConfigurationLoad"/>.
+        /// Optional action to perform after preparing the request. If this property is non-null,
+        /// the <see cref="JobConfigurationLoad"/> used for a request will be passed to the delegate
+        /// before the request is executed. This allows for fine-grained modifications which aren't
+        /// otherwise directly supported by the properties in this options type.
         /// </summary>
+        /// <remarks>
+        /// Prefer the properties on this type over this modifier to prepare the request.
+        /// Only use this modifier to configure aspects for which there are no properties available.
+        /// This modifier is applied to the request after all properties on this type have been applied.
+        /// The delegate is only called once per operation, even if the request is automatically retried.
+        /// </remarks>
         public Action<JobConfigurationLoad> ConfigurationModifier { get; set; }
 
         internal void ModifyConfiguration(JobConfigurationLoad loadRequest)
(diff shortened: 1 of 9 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadJsonOptions.cs
```diff
@@ -77,8 +77,16 @@ public sealed class UploadJsonOptions : JobCreationOptions
...
 
         /// <summary>
-        /// Delegate that is used to modify the underlying <see cref="JobConfigurationLoad"/>.
+        /// Optional action to perform after preparing the request. If this property is non-null,
+        /// the <see cref="JobConfigurationLoad"/> used for a request will be passed to the delegate
+        /// before the request is executed. This allows for fine-grained modifications which aren't
+        /// otherwise directly supported by the properties in this options type.
         /// </summary>
+        /// <remarks>
+        /// Properties within this type are used to prepare the request before the delegate is called,
```
Rishabh-V, line 86: It looks good. 👍🏼 This conveys the message and helps me use this property correctly. 
I think we can break the sentence for even better clarity. Something like:
```Properties within this type are used to prepare the request before the delegate is called. Therefore, the delegate should generally not modify .........```
But it is fine without that change as well.

Rishabh-V (approved):
LGTM.

Also shared my thoughts inline.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadJsonOptions.cs
amanda-tarafa: I think something along the lines of preference would be clearer:

"Prefer the properties on this type over this modifier to prepare the request. Only use this modifier to configure aspects for which there are no properties available. This modifier is applied to the request after all properties on this type have been applied."

amanda-tarafa (commented):
I've proposed an alternative, but happy as is as well.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UploadJsonOptions.cs
jskeet: @amanda-tarafa: Works for me. So that's just changing the first sentence in the remarks, and I'm okay to keep the one about "The delegate is only called once [...]"?

amanda-tarafa: Yes, exactly. And also yes, keep the one about the delegate being called only once.

jskeet:
I've now applied Amanda's suggested change, and then applied the change to all jobs. PTAL.

amanda-tarafa (approved):
Thanks!

jskeet:
(Will wait until the morning to merge, in case the Lex team has further comments.)

Rishabh-V:
> (Will wait until the morning to merge, in case the Lex team has further comments.)

This looks good to me. Thanks.

PR #11194, opened 2023-10-18
chore: GenerateSBOM tool

SurferJeffAtGoogle:


.gitattributes
```diff
@@ -1 +1,3 @@
 generateapis.sh eol=lf
+*.csproj text eol=lf
```
jskeet, line 2: I would rather not make this change in this PR. We may well want to do it, but it's not part of the tool.

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
```diff
@@ -1,329 +1,329 @@
...
+// The classes below are not general purpose Spdx classes.  Their single
+// purpose is to generate json SBOMs for nupkgs generated from this repository,
+// so they have default values convenient for that purpose.
+// Reusing these classes for any other purpose would require a refactor.
+
+/// <summary>
+/// SPDX format.
+/// See https://github.com/spdx/spdx-spec/blob/development/v2.3.1/schemas/spdx-schema.json
+/// </summary>
+internal class Spdx
```
jskeet, line 33: There's quite a lot I'd want to change here in terms of C# conventions etc, but I think that's going to be best done as a follow-up PR after getting the initial logic in. Basically, I'll avoid nits in this review.

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
```diff
@@ -1,329 +1,329 @@
...
+        = new List<SpdxRelationship> { new() };
+}
+
+internal class SpdxPackage
+{
+    public string name { get; set; } = "";
+    public string SPDXID { get; set; } = "";
+    public string versionInfo { get; set; } = "";
+    public string description { get; set; } = "";
+    public string supplier { get; set; } = "Person: Google LLC";
```
jskeet, line 56: Google isn't a person - I'd expect this to be Organization, according to https://spdx.github.io/spdx-spec/v2.3/package-information/#751-description

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
```diff
@@ -1,329 +1,329 @@
...
+    public string originator { get; set; } = "Person: Google LLC";
+    public string downloadLocation { get; set; }
+        = "https://github.com/googleapis/google-cloud-dotnet.git";
+    public string licenseConcluded { get; set; } = "Apache-2.0";
```
jskeet, line 60: It's worth noting that some .NET packages we ship - not from this repo - are MIT. We'll need to bear that in mind if we start using this for other repos.

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
```diff
@@ -1,329 +1,329 @@
...
+        var spdx = new Spdx
+        {
+            name = nuspecMetadata.Id + "@" + nuspecMetadata.Version,
+            documentNamespace = "https://spdx.google/" + uuid,
+            creators = new List<string> { "Organization: Google LLC" },
+            created = DateTime.UtcNow.ToString("yyyy-MM-ddTHH:mm:ssZ",
+                System.Globalization.CultureInfo.InvariantCulture),
+            packages = new List<SpdxPackage> {
+                new() {
+                    name = fileName,
```
jskeet, line 250: In https://spdx.github.io/spdx-spec/v2.3/package-information/#71-package-name-field, the example doesn't include the version number... so I'd expect this just to be the NuGet package name, e.g. Google.Cloud.Firestore.

There's a separate Package File Name property we could use for the filename: https://spdx.github.io/spdx-spec/v2.3/package-information/#74-package-file-name-field

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
```diff
@@ -1,329 +1,329 @@
...
+    public string SPDXID { get; set; } = "";
+    public string versionInfo { get; set; } = "";
+    public string description { get; set; } = "";
+    public string supplier { get; set; } = "Person: Google LLC";
+    public string originator { get; set; } = "Person: Google LLC";
+    public string downloadLocation { get; set; }
+        = "https://github.com/googleapis/google-cloud-dotnet.git";
+    public string licenseConcluded { get; set; } = "Apache-2.0";
+    public string licenseDeclared { get; set; } = "Apache-2.0";
+    public string copyrightText { get; set; } = "NOASSERTION";
```
jskeet, line 62: This surprises me... shouldn't it be copyright Google LLC?

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
```diff
@@ -1,329 +1,329 @@
...
+                System.Globalization.CultureInfo.InvariantCulture),
+            packages = new List<SpdxPackage> {
+                new() {
+                    name = fileName,
+                    SPDXID = "SPDXRef-Package-" + fileName + "-" + uuid,
+                    versionInfo = nuspecMetadata.Version,
+                    description = nuspecMetadata.Description,
+                    checksums = new List<SpdxChecksum> {
+                        new() {
+                            algorithm = "SHA1",
```
jskeet, line 256: Do we definitely want to use SHA-1 rather than SHA-256? My understanding is that SHA-1 should really not be used for anything new these days. Even if other SBOM creation tools use SHA-1, could we not use SHA-256?

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
SurferJeffAtGoogle: Ok.  Thanks.

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
SurferJeffAtGoogle: No checksum is required.  Since the code already calculates a SHA1 to build the UUID, I reckoned it was easy to record in the SBOM.

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
SurferJeffAtGoogle: A question for Emrick.

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
SurferJeffAtGoogle: Done.

tools/Google.Cloud.Tools.GenerateSBOM/Program.cs
jskeet: Makes sense. I'm done for the day on my work laptop, but will send a message tomorrow.

jskeet (approved):
Thinking about it more, if I'm going to be asking questions about the format anyway, I think it's probably simplest to squash and merge this PR now, and then I can make any changes. No need to go back and forth.

PR #11206, opened 2023-10-23
chore: generate SBOMs during release

SurferJeffAtGoogle:
How do I test this?

### DIFF
--- a/.kokoro/release.sh
+++ b/.kokoro/release.sh
@@ -57,6 +57,10 @@ echo "Building with commit $COMMITTISH"
 
 if [[ $SKIP_NUGET_PUSH = "" ]]
 then
+  echo "Generating SBOMs"
+  nuget_dir = "$(pwd)/releasebuild/nuget"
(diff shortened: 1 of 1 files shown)


jskeet:
I'd really like to pause this effort for a bit - I'll need to put some significant time into it, and right now I have other high priority work.

jskeet:
Have been thinking about this overnight - I think I'd like to turn this into a dotnet tool (as we have for a couple of other tools) before we start invoking it. But we don't need to do the refactoring before that. (We do need to work out whether to keep the current "do everything in a directory" or require specific filenames to be passed - I'm personally in favour of the latter.)

@amanda-tarafa Does that sound about right to you? (It'll make it easy to generate SBOMs in other repos, e.g. for Apiary, *and* we don't need to worry about invoking dotnet from the right directory.)

amanda-tarafa:
Yep, doing one file a time matches better with out setup here and in other repos.

jskeet:
Closing this as I'll be sending out a replacement PR once I've released the SBOM generator as a dotnet tool.

PR #11214, opened 2023-10-25
chore: Use the new SBOM generator tool during the release process

jskeet:
Any NuGet package that we're going to push, we first generate the SBOM for.


jskeet:
(I'll wait until the tool is published before merging this - shouldn't be long now.)

Rishabh-V (approved):

PR #11217, opened 2023-10-25
(DO NOT MERGE) Prototype of using API key support in GAX

jskeet:
This isn't *expected* to build, btw - but it works locally, where I have the prototype GAX package.

I suspect we'd want to add something in the service config to instruct the generator to generate things differently, rather than do it by hand. (This modifies existing generated code rather than being purely additive.)


amanda-tarafa (commented):
+1 on adding something in service config, but at worst, we can generate always the call to `GetEffectiveSettings` (it does the right thing if there's no ApiKey) and then adding the public property is an addition to the partial client builder class.

jskeet:
Closing for now. I've created internal bugs covering the rest of what needs to be done for this.

PR #11222, opened 2023-10-26
feat: Add Google.Cloud.Storage.V2 (highly experimental)

jskeet:
This adds and generates Google.Cloud.Storage.V2, but this should *not* be taken as an indication that the API is ready to use. We're adding it to the repo to make it easier to explore how we want to augment it with handwritten code, but it's *absolutely* not ready for production use, and we'd recommend not even experimenting with it yet.


jskeet:
(Expecting this new push to fail on "build and test" - creating a client should fail at the moment...)

amanda-tarafa (approved):

PR #11247, opened 2023-10-30
refactor: SBOM tool generation refactor

amanda-tarafa:
@jskeet as always one commit at a time. I'll all refactoring in a single commit before merging.

PR #11248, opened 2023-10-30
chore: Makes SBOM generator usable from other repos.

amanda-tarafa:
@jskeet only the last commit is specific to this PR, the rest are from #11247.

### DIFF
--- a/tools/Google.Cloud.Tools.SbomGenerator/NuspecPackage.cs
+++ b/tools/Google.Cloud.Tools.SbomGenerator/NuspecPackage.cs
@@ -38,5 +38,11 @@ public class NuspecPackage
 
         [XmlElement("description")]
         public string Description { get; set; }
+
+        [XmlElement("projectUrl")]
+        public string ProjectUrl { get; set; }
+
+        [XmlElement("license")]
+        public string License { get; set; }
     }
 }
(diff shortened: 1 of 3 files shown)


tools/Google.Cloud.Tools.SbomGenerator/NuspecPackage.cs
```diff
@@ -38,5 +38,11 @@ public class NuspecMetadata
 
         [XmlElement("description")]
         public string Description { get; set; }
+
+        [XmlElement("projectUrl")]
+        public string ProjectUrl { get; set; }
+
+        [XmlElement("license")]
+        public string License { get; set; }
```
jskeet, line 46: This is assuming that the license type is "expression". Unfortunately I suspect that with XML serialization, that's relatively hard to validate...

jskeet (approved):
Wow, that's easier than I'd expected. Hooray :)
(It's a pity that the license expression isn't in its own XML element, but I think it's okay - we're only intending to use this for our own packages, after all.)

tools/Google.Cloud.Tools.SbomGenerator/NuspecPackage.cs
amanda-tarafa: Yes, I did look into it briefly, but couldn't find an evident way to do this.

PR #11288, opened 2023-11-02
fix(pubsub): set x-goog-request-params for streaming pull request

kamalaboulhosn:
This is usually set within the auto-generated code, but since streaming pull doesn't have an http path associated with it, it does not get auto-generated.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
@@ -33,6 +33,8 @@ namespace Google.Cloud.PubSub.V1;
 
 public sealed partial class SubscriberClientImpl
 {
+    private const string XGoogRequestParamsHeaderKey = "x-goog-request-params";
+
     /// <summary>
     /// Controls a single <see cref="ChannelBase"/>/<see cref="SubscriberClient"/> within this
     /// <see cref="SubscriberClientImpl"/>. This class controls the pulling of messages, and
@@ -310,7 +312,8 @@ public sealed partial class SubscriberClientImpl
         // Backoff delay (if present) has already been done; no need to delay here.
         private void HandleStartStreamingPullWithoutBackoff()
         {
-            _pull = _client.StreamingPull(CallSettings.FromCancellationToken(_softStopCts.Token));
+            var callSettings = CallSettings.FromHeader(XGoogRequestParamsHeaderKey, string.Format("subscription={0}", _subscriptionName));
+            _pull = _client.StreamingPull(callSettings.MergedWith(CallSettings.FromCancellationToken(_softStopCts.Token)));
             // Cancellation not needed in this WriteAsync call. The StreamingPull() cancellation
             // (above) will cause this call to cancel if _softStopCts is cancelled.
             Task initTask = _pull.WriteAsync(new StreamingPullRequest


jskeet (commented):
We should be able to do this within a partial class (and using dedicated code for x-goog-request-params, that knows about URL-encoding the value etc). I'll create a counter-PR tomorrow.

jskeet:
Notes to myself as much as anything else... We have this as a partial method:

`partial void Modify_StreamingPullApiCall(ref gaxgrpc::ApiBidirectionalStreamingCall<StreamingPullRequest, StreamingPullResponse> call);`

... but `ApiBidirectionalStreamingCall` doesn't have `WithCallSettingsOverlay` like the unary equivalents.

I still have options I'm thinking about though... will try to dive into this tomorrow.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -310,7 +312,8 @@ private void StartStreamingPull()
         // Backoff delay (if present) has already been done; no need to delay here.
         private void HandleStartStreamingPullWithoutBackoff()
         {
-            _pull = _client.StreamingPull(CallSettings.FromCancellationToken(_softStopCts.Token));
+            var callSettings = CallSettings.FromHeader(XGoogRequestParamsHeaderKey, string.Format("subscription={0}", _subscriptionName));
+            _pull = _client.StreamingPull(callSettings.MergedWith(CallSettings.FromCancellationToken(_softStopCts.Token)));
```
jskeet, line 316: Instead of calling `callSettings.MergedWith(...)`, just `callSettings.WithCancellationToken(_softStopCts.Token)`.

(I'd create suggestions for these, but they'd end up with the wrong email address causing more bureaucracy...)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -310,7 +312,8 @@ private void StartStreamingPull()
         // Backoff delay (if present) has already been done; no need to delay here.
         private void HandleStartStreamingPullWithoutBackoff()
         {
-            _pull = _client.StreamingPull(CallSettings.FromCancellationToken(_softStopCts.Token));
+            var callSettings = CallSettings.FromHeader(XGoogRequestParamsHeaderKey, string.Format("subscription={0}", _subscriptionName));
```
jskeet, line 315: Please add a TODO to call CallSettings.FromGoogleRequestParamsHeader when https://github.com/googleapis/gax-dotnet/issues/733 is fixed and released, and change the second argument to:

```csharp
$"subscription={Uri.EscapeDataString(_subscriptionName.ToString())}"`
```

We could potentially create this whole CallSettings once as a field, but I guess we're not calling HandleStartStreamingPullWithoutBackoff that often...

jskeet (commented):
We can take this change for the moment, but there's actually a better option in terms of putting this into the SubscriberServiceApiClientSettings when we create the client.

However, I *think* at the moment there's a bug in how we're (not) using the API settings anyway, which I need to look into. I'll see if I can create a fix for that now, but we could get this in instead if this is urgent and if the other fix will take a while.

jskeet:
> However, I think at the moment there's a bug in how we're (not) using the API settings anyway, which I need to look into. I'll see if I can create a fix for that now, but we could get this in instead if this is urgent and if the other fix will take a while.

Nope, there isn't a bug - we should be able to do this quite cleanly. I'll send an alternative PR to you.

PR #11293, opened 2023-11-03
fix: Set x-goog-request-params for streaming pull requests from SubscriberClient

jskeet:
This is an alternative to #11288, setting a single CallSettings that is applied to every StreamingPull call.

(It's significantly more code, but feels more appropriate to me in terms of setting things up a single time so that any call to StreamingPull just does the right thing naturally.)

I'd like review from both @kamalaboulhosn and @Rishabh-V. I tried to work out a way to test it, but there are no simple injection points for this, and it didn't seem worth adding them just for this fix.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.ClientCreationSettings.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.ClientCreationSettings.cs
@@ -25,7 +25,7 @@ public abstract partial class SubscriberClient
     /// This type is now obsolete; please use <see cref="SubscriberClientBuilder"/> which provides an
     /// API surface consistent with other clients (as well as additional Pub/Sub-specific properties such as <see cref="SubscriberClientBuilder.ClientCount"/>).
     /// </summary>
-    [Obsolete("Use PublisherClientBuilder to customize client settings.")]
+    [Obsolete("Use SubscriberClientBuilder to customize client settings.")]
     public sealed class ClientCreationSettings
     {
         /// <summary>
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientBuilder.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientBuilder.cs
@@ -144,6 +144,30 @@ public sealed class SubscriberClientBuilder : ClientBuilderBase<SubscriberClient
 
             return channel.ShutdownAsync();
         }
+
+    }
+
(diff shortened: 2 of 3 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClient.ClientCreationSettings.cs
```diff
@@ -25,7 +25,7 @@ public abstract partial class SubscriberClient
     /// This type is now obsolete; please use <see cref="SubscriberClientBuilder"/> which provides an
     /// API surface consistent with other clients (as well as additional Pub/Sub-specific properties such as <see cref="SubscriberClientBuilder.ClientCount"/>).
     /// </summary>
-    [Obsolete("Use PublisherClientBuilder to customize client settings.")]
+    [Obsolete("Use SubscriberClientBuilder to customize client settings.")]
```
jskeet, line 28: Note: this is just a fix to a typo that I noticed at the same time. I can pull it out to a separate PR if desired.

Rishabh-V: Ah! Right. Thanks for spotting it.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientBuilder.cs
```diff
@@ -144,6 +144,30 @@ static Task DisposeChannelAsync(Grpc.Core.ChannelBase channel)
...
+        }
+
+        // We never modify a settings object that the user code has specified explicitly.
+        var settingsToModify = ApiSettings?.Clone() ?? new SubscriberServiceApiSettings();
+
+        // TODO: Use CallSettings.FromGoogleRequestParamsHeader when
+        // https://github.com/googleapis/gax-dotnet/issues/733 is fixed and released.
+        settingsToModify.StreamingPullSettings = settingsToModify.StreamingPullSettings
+            .WithHeader("x-goog-request-params", $"subscription={Uri.EscapeDataString(subscriptionName.ToString())}");
+        return settingsToModify;
```
Rishabh-V, line 170: 👍🏼  Yep. We have the method for it but is internal. This is clean and once we have the method exposed, it will be easier to refactor this.

Rishabh-V (approved):

kamalaboulhosn:
Thanks for your work on this, @jskeet! I'll close my other PR for this.

jskeet:
Great, thanks. Merged, and @Rishabh-V is going to perform a test - then we can cut a release with this change.

PR #11320, opened 2023-11-09
feat: Adds support for dotnet data protection with secret manager

hemanshv:
Adds support for .NET data protection using GCP Secret manager


Rishabh-V:
I 'll have a look tomorrow morning. You may want to check out the failing tests in the meantime.

apis/apis.json
```diff
@@ -55,19 +55,36 @@
...
-      "description": "Version-agnostic types for Apps Script APIs.",
-      "tags": [
-        "gsuite"
-      ],
-      "dependencies": {
-        "Google.Api.CommonProtos": "2.10.0"
-      },
-      "generator": "proto",
-      "protoPath": "google/apps/script/type"
+        "id": "Google.Apps.Script.Type",
```
Rishabh-V, line 58: Indentation is changed. Please revert it. Only the newly added changes should be shown in diff.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/docs/index.md
```diff
@@ -0,0 +1,17 @@
+## Getting Started
+
+The `Google.Cloud.AspNetCore.DataProtection.SecretManager` package allows storing ASP.NET keys in [Google Cloud Secret Manager](https://cloud.google.com/secret-manager).
```
Rishabh-V, line 3: ASP.NET or ASP.NET Core?

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/docs/index.md
```diff
@@ -0,0 +1,17 @@
...
+The keys can be shared across several instances of a webapp. Apps can share authentication cookies
+or CSRF protection across multiple servers
+
+Follow the examples below to see how the library can be integrated into your application.
+
+## Sample code
+```csharp
+public void ConfigureServices(IServiceCollection services)
```
Rishabh-V, line 11: Thinking if we should have a snippet that can be referenced here. This will answer the obvious questions like what using directives to import etc. Please don't make any changes yet and wait for suggestions from Jon and Amanda.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/IntegrationTests/CloudSecretManagerXmlRepositoryTest.cs
```diff
@@ -0,0 +1,62 @@
...
+        repo.StoreElement(data2, "ignore");
+        var retrivedData = repo.GetAllElements();
+
+        XElement element = Assert.Single(retrivedData);
+        Assert.Equal("id2", element.Attribute("id")?.Value);
+    }
+
+    private CloudSecretManagerXmlRepository CreateXmlRepository(string secretName)
+    {
+        var client = SecretManagerServiceClient.Create();
```
Rishabh-V, line 59: Do we need to create client in every test method?

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,128 @@
...
+                    Automatic = new Replication.Types.Automatic(),
+                },
+            },
+            ParentAsProjectName = _projectName,
+            SecretId = _secretName.SecretId
+        };
+        _secretManagerServiceClient.CreateSecret(createSecretRequest, _callSettings);
+    }
+
+    private bool IsSecretExists()
```
Rishabh-V, line 116: The name sounds grammatically incorrect. Maybe we can use just `SecretExists`

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/.repo-metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+    "distribution_name": "Google.Cloud.AspNetCore.DataProtection.SecretManager",
+    "release_level": "stable",
```
Rishabh-V, line 3: Couple of questions:

.NET PowerPack repository was released as alpha, we are directly migrating that to a stable release?
  
I am not versed with the values of library_type. Could you please point me to the documentation you referred to provide the value as INTEGRATION (this is for my reference)

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        try
+        {
+            var res = _secretManagerServiceClient.GetSecret(_secretName);
```
Rishabh-V, line 120: We can surely have better name than `res`

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            return Array.Empty<XElement>();
+        }
+        var accessSecretVersionResponse = GetSecretVersionResponse();
+
+        if (accessSecretVersionResponse != null)
+        {
+            var secretPayload = accessSecretVersionResponse.Payload.Data.ToStringUtf8();
+            XDocument document = XDocument.Parse(secretPayload);
+            var res = document?.Root?.Elements().ToList();
```
Rishabh-V, line 71: Please use a descriptive name here as well.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/GoogleCloudDataProtectionBuilderExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+using Google.Cloud.SecretManager.V1;
+using Microsoft.AspNetCore.DataProtection;
+using Microsoft.AspNetCore.DataProtection.KeyManagement;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+
+namespace Google.Cloud.AspNetCore.DataProtection.SecretManager;
+
+/// <summary>
+/// Contains extension methods for modifying a <see cref="IDataProtectionBuilder"/> to work with Google Cloud Secret Manager.
```
Rishabh-V, line 25: Can we say - `Extension methods for ......` or `Static class to provide extension methods....`

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/GoogleCloudDataProtectionBuilderExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+/// </summary>
+public static class GoogleCloudDataProtectionBuilderExtensions
+{
+    /// <summary>
+    /// Configures data protection system to persist keys in Google Cloud Secret Manager.
```
Rishabh-V, line 30: nit: builder instead of system?  (here and below)
Maybe use cref so that it becomes intuitive to the user at the time of using the method?

Rishabh-V (commented):
A quick first pass. Will have a look again in detail later in the day. Thanks.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/.repo-metadata.json
hemanshv: yes @Rishabh-V on the release part I am also a bit nervous on that I think we can start up with `alpha`  first.

on the library_type part you can refer to [go/library-types](https://goto.google.com/library-types)

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/docs/index.md
hemanshv: done

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
hemanshv: done

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
hemanshv: done

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/GoogleCloudDataProtectionBuilderExtensions.cs
hemanshv: done

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,128 @@
...
+            return Array.Empty<XElement>();
+        }
+        var accessSecretVersionResponse = GetSecretVersionResponse();
+
+        if (accessSecretVersionResponse != null)
+        {
+            var secretPayload = accessSecretVersionResponse.Payload.Data.ToStringUtf8();
+            XDocument document = XDocument.Parse(secretPayload);
+            var xElements = document?.Root?.Elements().ToList();
+            return xElements ?? (IReadOnlyCollection<XElement>) Array.Empty<XElement>();
```
anuragsrivstv, line 72: I think .ToList() will never return null.

jskeet: ToList won't, but if `document` or `document.Root` is null (which is explicitly handled in the line above) then `xElements` will still be null. That said, I don't believe `XDocument.Parse` ever returns null, or returns a document with no root.

hemanshv: got it I will address it.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/.repo-metadata.json
Rishabh-V: Thanks. Reading the document, I am still not clear if `library_type` should be `INTEGRATION`. Let's get this reviewed from Jon and Amanda and we'll know if this the is right value to be used.

Rishabh-V:
> A quick first pass. Will have a look again in detail later in the day. Thanks.

Having another look now.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager.csproj
```diff
@@ -0,0 +1,20 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <Version>1.0.0-alpha01</Version>
+        <TargetFramework>netstandard2.1</TargetFramework>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
+        <Description>Recommended Google client library to use Google Cloud Secret Manager to store data protection keys for ASP.NET Core.</Description>
+        <PackageTags>Secret Manager;Google;Cloud;Security</PackageTags>
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
Rishabh-V, line 10: Do we need this section? Please remove them.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/GoogleCloudDataProtectionBuilderExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+/// <summary>
+/// Static class to provide extension methods for modifying a <see cref="IDataProtectionBuilder"/> to work with Google Cloud Secret Manager.
+/// </summary>
+public static class GoogleCloudDataProtectionBuilderExtensions
+{
+    /// <summary>
+    /// Configures data protection system to persist keys in Google Cloud Secret Manager.
+    /// </summary>
+    /// <param name="builder">The data protection builder to configure. Must not be null.</param>
+    /// <param name="projectName">The project id in which the keys are stored. Must not be null</param>
```
Rishabh-V, line 33: Nit: Can we just name it as projectId so that name and description is consistent? (here and in other relevant places)

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/.repo-metadata.json
```diff
@@ -0,0 +1,7 @@
+{
+    "distribution_name": "Google.Cloud.AspNetCore.DataProtection.SecretManager",
+    "release_level": "preview",
+    "client_documentation": "https://cloud.google.com/dotnet/docs/reference/",
```
Rishabh-V, line 4: Is this url correct?

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,128 @@
...
+/// This class is configured by <see cref="GoogleCloudDataProtectionBuilderExtensions.PersistKeysToGoogleCloudSecretManager(Microsoft.AspNetCore.DataProtection.IDataProtectionBuilder, string, string)"/>
+/// (and other overloads).
+/// </summary>
+internal class CloudSecretManagerXmlRepository : IXmlRepository
+{
+    private readonly SecretManagerServiceClient _secretManagerServiceClient;
+    private readonly SecretName _secretName;
+    private readonly ProjectName _projectName;
+    private readonly SecretVersionName _version;
+    private readonly CallSettings _callSettings = CallSettingsExtensions.WithRetry(CallSettings.FromExpiration(Expiration.FromTimeout(TimeSpan.FromMilliseconds(10000))), RetrySettings.FromExponentialBackoff(maxAttempts: 5, initialBackoff: TimeSpan.FromMilliseconds(0.2), maxBackoff: TimeSpan.FromMilliseconds(10000), backoffMultiplier: 1.5, retryFilter: RetrySettings.FilterForStatusCodes(StatusCode.DeadlineExceeded, StatusCode.Unavailable)));
```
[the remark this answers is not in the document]
hemanshv: The default setting are just specifying the Expiration time but here I want to implement retry as well alongside. Its helps incase we are trying to read the value but the serivce was unavailable. Its better to retry for while instead of throwing an error rightaway (likewise in othercases) .
Thoughts?

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/.repo-metadata.json
hemanshv: No its not I was getting an error added it for debugging reasons. Forgot to remove it afterward will remove it. Thanks for pointing out. Right now we don't have any client docs for it.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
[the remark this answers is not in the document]
Rishabh-V: I think for `AccessSecretVersion` we already have retry in place out of the box for `Unavailable` and `ResourceExhausted` error codes. Do we need it for `AddSecretVersion` and `CreateSecret` ? Have we encountered any issues or are we doing it defensively for better resiliency? 

It may be fine and I 'll leave it for Jon/Amanda to take a call.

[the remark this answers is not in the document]
hemanshv: Ok got it we can simply the `AccessSecretVersion` then for other I did it defensively only haven't encountered any issues.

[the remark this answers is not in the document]
Rishabh-V: (So the question arises if retry could be configured for one API in the client, it could have been done for others as well but not done. So, there must be a reason for not retrying the Create and Add methods.) Let's take inputs from Jon/Amanda before making a change.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,128 @@
...
+            var secretPayload = accessSecretVersionResponse.Payload.Data.ToStringUtf8();
+            XDocument document = XDocument.Parse(secretPayload);
+            var xElements = document.Root.Elements().ToList();
+            return xElements ?? (IReadOnlyCollection<XElement>) Array.Empty<XElement>();
+        }
+        return Array.Empty<XElement>();
+
+        AccessSecretVersionResponse GetSecretVersionResponse()
+        {
+            var response = _secretManagerServiceClient.AccessSecretVersion(_version, _callSettings);
```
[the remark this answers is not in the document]
hemanshv: done

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,128 @@
...
+            return Array.Empty<XElement>();
+        }
+        var accessSecretVersionResponse = GetSecretVersionResponse();
+
+        if (accessSecretVersionResponse != null)
+        {
+            var secretPayload = accessSecretVersionResponse.Payload.Data.ToStringUtf8();
+            XDocument document = XDocument.Parse(secretPayload);
+            var xElements = document.Root.Elements().ToList();
+            return xElements ?? (IReadOnlyCollection<XElement>) Array.Empty<XElement>();
```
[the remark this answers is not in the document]
hemanshv: done

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager.csproj
hemanshv: done

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/GoogleCloudDataProtectionBuilderExtensions.cs
hemanshv: done

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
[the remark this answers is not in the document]
hemanshv: I've removed the logic for retry. As other are non idempotency operations and it can lead to inconsistency. 
Happy to get feedback on the same.

hemanshv:
Hi @jskeet and @amanda-tarafa ,
We are done with the internal review on this. Please have a look and share your valuable feedback.

Thanks,

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Xml.Linq;
+using Secret = Google.Cloud.SecretManager.V1.Secret;
+
+namespace Google.Cloud.AspNetCore.DataProtection.SecretManager;
+
+/// <summary>
+/// Implementation of <see cref="IXmlRepository"/> that stores the protected elements in a Google Cloud Secret Manager.
```
jskeet, line 31: Remove "a" from "in a Google Cloud Secret Manager".

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+        // Version is populated in such a way that it always refers to the most latest version.
+        // For more information on version naming refer to
+        // https://cloud.google.com/dotnet/docs/reference/Google.Cloud.SecretManager.V1/latest/Google.Cloud.SecretManager.V1.AccessSecretVersionRequest#Google_Cloud_SecretManager_V1_AccessSecretVersionRequest_Name
+        _version = SecretVersionName.FromProjectSecretSecretVersion(projectId, secretName, "latest");
+    }
+
+    /// <inheritdoc />
+    public IReadOnlyCollection<XElement> GetAllElements()
+    {
+        if (!SecretExists())
```
jskeet, line 59: Rather than make an RPC just to check if it exists, can we call AccessSecretVersion unconditionally, and catch whatever failure we get if it doesn't exist?

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            return Array.Empty<XElement>();
+        }
+        var accessSecretVersionResponse = _secretManagerServiceClient.AccessSecretVersion(_version);
+
+        if (accessSecretVersionResponse != null)
```
jskeet, line 65: RPCs never return null, so you can remove this. (Someone mocking the service could cause a NullReferenceException, but that's their own fault...)

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+            var xElements = document.Root.Elements().ToList();
+            return xElements;
+        }
+        return Array.Empty<XElement>();
+    }
+
+    /// <inheritdoc />
+    public void StoreElement(XElement element, string friendlyName)
+    {
+        if (!SecretExists())
```
jskeet, line 78: There's a race condition here if multiple clients try to do this at the same time.
Maybe it's worth just unconditionally creating the secret, and catching the specific "it already exists" failure?
If we're worried that's expensive, we could check for existence, then try to create the secret but *still* catch the "it already exists" failure.

@amanda-tarafa thoughts welcome here.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+        {
+            CreateSecret();
+        }
+        XDocument document = new XDocument(new XElement("root"));
+        document.Root.Add(element);
+        SecretPayload payload = new SecretPayload
+        {
+            Data = ByteString.CopyFrom(document.ToString(), Encoding.UTF8)
+        };
+        _secretManagerServiceClient.AddSecretVersion(_secretName, payload);
```
jskeet, line 88: Doesn't this overwrite anything that was previously there? I'd expect it to just add a new top-level element.

Compare this with https://github.com/dotnet/aspnetcore/blob/main/src/DataProtection/DataProtection/src/Repositories/FileSystemXmlRepository.cs - where multiple elements are basically stored as separate files.

I'd expect an integration test of:

- Create new secret
- Add element 1
- Add element 2
- Fetch all elements (both should be there)

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/GoogleCloudDataProtectionBuilderExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Google.Cloud.AspNetCore.DataProtection.SecretManager;
+
+/// <summary>
+/// Static class to provide extension methods for modifying a <see cref="IDataProtectionBuilder"/> to work with Google Cloud Secret Manager.
+/// </summary>
+public static class GoogleCloudDataProtectionBuilderExtensions
+{
+    /// <summary>
+    /// Configures data protection system to persist keys in Google Cloud Secret Manager.
```
jskeet, line 30: I'd remove "system" here, *or* change it to "Configures the data protection system" (adding "the").
Same for other methods.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/GoogleCloudDataProtectionBuilderExtensions.cs
```diff
@@ -0,0 +1,78 @@
...
+        builder.Services.AddSingleton<IConfigureOptions<KeyManagementOptions>>(services =>
+        {
+            if (client == null)
+            {
+                client = services.GetService<SecretManagerServiceClient>();
+                if (client == null)
+                {
+                    var credential = services.GetService<GoogleCredential>();
+                    // If credential is null, this will use the default credentials automatically.
+                    client = SecretManagerServiceClient.Create();
```
jskeet, line 70: You're not using credential here at all.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/IntegrationTests/CloudSecretManagerXmlRepositoryTest.cs
```diff
@@ -0,0 +1,59 @@
...
+    [Fact]
+    private void StoreMultipleVersion_AndReadLatestVersion()
+    {
+        var data = new XElement("foo", new XAttribute("id", "id1"));
+        var data2 = new XElement("foo", new XAttribute("id", "id2"));
+        var secretName = _secretManagerFixture.GenerateSecretName();
+        var repo = CreateXmlRepository(secretName);
+        repo.StoreElement(data, "ignore");
+        repo.StoreElement(data2, "ignore");
+        var retrivedData = repo.GetAllElements();
```
jskeet, line 51: Nit: retrived => retrieved

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/IntegrationTests/CloudSecretManagerXmlRepositoryTest.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        XElement element = Assert.Single(retrivedData);
```
jskeet, line 53: This is where I'd expect to see multiple elements.
We might even want an integration test using the file system provider to validate how that behaves, so we're consistent.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/IntegrationTests/SecretManagerFixture.cs
```diff
@@ -0,0 +1,56 @@
...
+        foreach(var secret in _generatedSecrets)
+        {
+            try
+            {
+                SecretName secretName = new SecretName(ProjectId, secret);
+                _client.DeleteSecret(secretName);
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Exception while deleting secret {secret}");
```
jskeet, line 51: I'd remove these Console.WriteLine calls, and just have a comment saying we're swallowing the exception.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/IntegrationTests/SecretManagerFixture.cs
```diff
@@ -0,0 +1,56 @@
...
+        _generatedSecrets.Add(secretName);
+        return secretName;
+    }
+
+    public SecretManagerFixture() =>
+        _client = SecretManagerServiceClient.Create();
+
+    public override void Dispose()
+    {
+        foreach(var secret in _generatedSecrets)
```
jskeet, line 42: Nit: space after foreach

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/docs/index.md
jskeet: Yes, it would be better to have snippets so we know the code at least compiles - it doesn't need to be part of a test necessarily.

As it happens, I don't think this code *would* compile, because SecretManagerServiceClient is abstract...

apis/apis.json
```diff
@@ -18,6 +18,23 @@
...
+      "type": "other",
+      "description": "Recommended Google client library to use Google Cloud Secret Manager to store data protection keys for ASP.NET Core.",
+      "tags": [
+        "Secret Manager",
+        "Google",
+        "Cloud",
+        "Security"
+      ],
+      "dependencies": {
+          "Google.Cloud.SecretManager.V1": "2.1.0",
```
jskeet, line 34: Nit: unindent to two spaces. (I suspect that running generateprojects.sh will do that for you automatically.)

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/IntegrationTests/SecretManagerFixture.cs
```diff
@@ -0,0 +1,56 @@
+// Copyright 2023 Google LLC
```
jskeet, line 1: Please rename IntegrationTests to something like DataProtection.SecretManager.IntegrationTests; the project generator looks for a suffix of ".IntegrationTests.csproj" so it needs the period.

I'd then use a *namespace* which matches our normal convention - so Google.Cloud.AspNetCore.DataProtection.SecretManager. It's only the directory/file name that needs to be unconventional.

jskeet (commented):
First round of comments. The most important part is the "add vs replace" semantics.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
hemanshv: Got it Jon Thanks for pointing this out. I missed this part.
Will rectifie it.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/docs/index.md
hemanshv: Added it.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
hemanshv: done

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
hemanshv: updated the logic Please have a look and share your thoughts

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/GoogleCloudDataProtectionBuilderExtensions.cs
hemanshv: rectified

jskeet:
Will review again when I can - but it may well not be until Monday.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/DataProtection.SecretManager.IntegrationTests/CloudSecretManagerXmlRepositoryTest.cs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    [Fact]
+    private void StoreAndRead()
+    {
+        var data = new XElement("foo", new XAttribute("id", "id1"));
+        var secretName = _secretManagerFixture.GenerateSecretName();
+        var repo = CreateXmlRepository(secretName);
+        repo.StoreElement(data, "ignore");
+        var retrivedData = repo.GetAllElements();
```
jskeet, line 48: Nit: retrieved, not retrived

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/DataProtection.SecretManager.IntegrationTests/CloudSecretManagerXmlRepositoryTest.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        XElement element = Assert.Single(retrivedData);
+        Assert.Equal("id1", element.Attribute("id")?.Value);
+    }
+
+    [Fact]
+    private void CompareWithFileSystemXmlRepository()
+    {
+        var loggerFactory = LoggerFactory.Create(builder =>
```
jskeet, line 57: Nit: this can be simpler as just:

```csharp
var loggerFactory = LoggerFactory.Create(builder => builder.AddConsole());
```

(In general, single-statement lambda expressions can usually be written with an expression body.)

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/DataProtection.SecretManager.IntegrationTests/CloudSecretManagerXmlRepositoryTest.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        var data = new XElement("foo", new XAttribute("id", "id1"));
+        var data2 = new XElement("foo", new XAttribute("id", "id2"));
+        var secretName = _secretManagerFixture.GenerateSecretName();
+        var repo = CreateXmlRepository(secretName);
+        repo.StoreElement(data, "ignore");
+        repo.StoreElement(data2, "ignore");
+        var retrievedData = repo.GetAllElements();
+
+        Assert.True(retrievedData.Count == 2);
```
jskeet, line 89: `Assert.Equal(2, retrievedData.Count);`

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/DataProtection.SecretManager.IntegrationTests/CloudSecretManagerXmlRepositoryTest.cs
```diff
@@ -0,0 +1,122 @@
...
+        fs.StoreElement(data2, "");
+        var res1 = fs.GetAllElements();
+
+        var secretName = _secretManagerFixture.GenerateSecretName();
+        var repo = CreateXmlRepository(secretName);
+        repo.StoreElement(data, "ignore");
+        repo.StoreElement(data2, "ignore");
+        var res2 = repo.GetAllElements();
+
+        Assert.True(CompareCollection(res1, res2));
```
jskeet, line 75: Can we not just use `Assert.Equal`, passing in the XNodeEqualityComparer?

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/DataProtection.SecretManager.Snippets/DataProtectionExtensionMethodsSnippets.cs
```diff
@@ -0,0 +1,61 @@
...
+namespace Google.Cloud.AspNetCore.DataProtection.SecretManager.Snippets;
+
+public class DataProtectionExtensionMethodsSnippets
+{
+    public void RegisterSecretManagerClient()
+    {
+        IServiceCollection services = new ServiceCollection();
+
+        // Sample: RegisterSecretManagerClient
+
```
jskeet, line 28: Nit: Remove blank line (here and for other samples).

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
```diff
@@ -46,75 +44,56 @@ internal CloudSecretManagerXmlRepository(SecretManagerServiceClient client, stri
...
 
     /// <inheritdoc />
     public IReadOnlyCollection<XElement> GetAllElements()
     {
-        if (!SecretExists())
+        try
         {
-            return Array.Empty<XElement>();
+            var elements = new List<XElement>();
+            foreach (SecretVersion secretVersion in _secretManagerServiceClient.ListSecretVersions(_secretName))
```
jskeet, line 55: I don't know enough about SecretManager to know whether this is a sensible way of achieving the aim of "storing a new element adds a new one". I'd have *expected* SecretManager to be built on "each version is intended to be complete", but I could be wrong. If I'm right though, I'd suggest one of two options:

- Create a new secret per element, using the secret name as a prefix
- Fetch the existing secret, parse it as an XDocument, add the element, and store the result (with suitable race condition handling)

I'd suggest talking to the SecretManager team about this.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/DataProtection.SecretManager.IntegrationTests/CloudSecretManagerXmlRepositoryTest.cs
hemanshv: They return a list and the file repo is reading the list the directory using some command internally so the order of element may be different in the 2 list.

jskeet: In that case we can just create sets rather than lists. (Or we could order each list by something stable.) I'm sure we don't need to implement a collection comparison ourselves.

apis/Google.Cloud.AspNetCore.DataProtection.SecretManager/Google.Cloud.AspNetCore.DataProtection.SecretManager/CloudSecretManagerXmlRepository.cs
hemanshv: Thanks Jon for sharing your input. 
I am referring to the existing FileSystemXmlRespository code  and  trying to make our implementation look similar to that. In the same we can see that they are basically adding every XElement in a separate file 
ref: https://docs.google.com/document/d/1iVHN6VuF0pTDDoai1YUSgyVJ6rxrqL6yCMdwfk66740/edit#heading=h.bbjct62hz474

and while reading they are reading all the files with the extension .xml inside the folder.
ref: https://github.com/dotnet/aspnetcore/blob/main/src/DataProtection/DataProtection/src/Repositories/FileSystemXmlRepository.cs#L84
In our implementation we can think of replacing directory with Secret and the file with the secret version. So every secret version will contain the XElement and while returning we will just return the list of versions available.

let me know your thoughts on the same.

jskeet: My point is that I don't think those are comparable concepts. There's a difference between "a directory with multiple independent files" and "multiple versions of one resource". Yes, you *can* list secret versions and get independent data, but I'm not convinced that's how SecretManager expects versions to be used. (Would you use separate commits in git to store fundamentally distinct data in the same file, for example? That would seem odd to me.)

Again, I'd suggest reaching out to the SecretManager team to discuss how multiple independent items should be stored in SecretManager.

hemanshv: Ok Jon I got the point thanks for explanation Will reach out to the secret manager team as next step.

jskeet:
Closing this PR as we're expecting to probably go back to using KMS instead of SecretManager (but still move the code from the current PowerPack repo.)

PR #11350, opened 2023-11-21
tests: Modify tests to account for XUnit issue.

amanda-tarafa:
Assert.Equal gives a false negative if dictionaries being compared contain null elements.

See https://github.com/xunit/xunit/issues/2824


jskeet (approved):
Hopefully we'll be able to undo this later :)

amanda-tarafa:
> Hopefully we'll be able to undo this later :)

Yes!

PR #11360, opened 2023-11-23
feat: Inline Begin Transaction

amanda-tarafa:
@jskeet this is ready for review.

The first three commits are refactors for making things easier later on. And then inlining is introduced gradually. Tests passed after each commit.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PooledSessionTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PooledSessionTests.cs
@@ -18,7 +18,6 @@ using Google.Api.Gax.Testing;
 using Google.Cloud.Spanner.V1.Internal;
 using Google.Cloud.Spanner.V1.Internal.Logging;
 using Google.Protobuf;
-using Google.Protobuf.WellKnownTypes;
 using Grpc.Core;
 using NSubstitute;
 using NSubstitute.Extensions;
@@ -101,16 +100,15 @@ namespace Google.Cloud.Spanner.V1.Tests
             clock.Advance(halfRefresh);
 
             // Make a successful request
-            var response = new Transaction();
             pool.Client.Configure()
-                .BeginTransactionAsync(Arg.Any<BeginTransactionRequest>(), Arg.Any<CallSettings>())
-                .Returns(Task.FromResult(response));
-            await pooledSession.WithFreshTransactionAsync(s_readOnlyOptions, default);
+                .ExecuteSqlAsync(Arg.Any<ExecuteSqlRequest>(), Arg.Any<CallSettings>())
+                .Returns(Task.FromResult(new ResultSet()));
+            await pooledSession.ExecuteSqlAsync(new ExecuteSqlRequest(), default);
 
             // The request will have extended the refresh time.
             Assert.Equal(clock.GetCurrentDateTimeUtc() + options.IdleSessionRefreshDelay, pooledSession.RefreshTimeForTest);
 
-            _ = pool.Client.Received(1).BeginTransactionAsync(Arg.Any<BeginTransactionRequest>(), Arg.Any<CallSettings>());
+            _ = pool.Client.Received(1).ExecuteSqlAsync(Arg.Any<ExecuteSqlRequest>(), Arg.Any<CallSettings>());
         }
 
         [Fact]
@@ -122,64 +120,110 @@ namespace Google.Cloud.Spanner.V1.Tests
             // Make a request which fails due to the session not being found (because it has expired).
             var exception = CreateSessionExpiredException();
             pool.Client.Configure()
-                .BeginTransactionAsync(Arg.Any<BeginTransactionRequest>(), Arg.Any<CallSettings>())
-                .Returns(Task.FromException<Transaction>(exception));
+                .ExecuteSqlAsync(Arg.Any<ExecuteSqlRequest>(), Arg.Any<CallSettings>())
+                .Returns(Task.FromException<ResultSet>(exception));
 
-            await Assert.ThrowsAsync<RpcException>(() => pooledSession.WithFreshTransactionAsync(s_readOnlyOptions, default));
+            await Assert.ThrowsAsync<RpcException>(() => pooledSession.ExecuteSqlAsync(new ExecuteSqlRequest(), default));
             Assert.True(pooledSession.ServerExpired);
 
-            _ = pool.Client.Received(1).BeginTransactionAsync(Arg.Any<BeginTransactionRequest>(), Arg.Any<CallSettings>());
+            _ = pool.Client.Received(1).ExecuteSqlAsync(Arg.Any<ExecuteSqlRequest>(), Arg.Any<CallSettings>());
         }
 
         [Fact]
-        public async Task WithFreshTransactionAsync()
+        public async Task NoTransactionOptions()
         {
             var pool = new FakeSessionPool();
             var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName);
-            var expectedTransaction = new Transaction
-            {
-                Id = ByteString.CopyFromUtf8("transaction"),
-                ReadTimestamp = DateTimeOffset.UtcNow.ToTimestamp()
-            };
 
             // Make a successful request
             pool.Client.Configure()
-                .BeginTransactionAsync(Arg.Any<BeginTransactionRequest>(), Arg.Any<CallSettings>())
-                .Returns(Task.FromResult(expectedTransaction));
+                .ExecuteSqlAsync(Arg.Any<ExecuteSqlRequest>(), Arg.Any<CallSettings>())
+                .Returns(Task.FromResult(new ResultSet()));
 
-            var newSession = await pooledSession.WithFreshTransactionAsync(s_readOnlyOptions, default);
+            await pooledSession.ExecuteSqlAsync(new ExecuteSqlRequest(), default);
 
-            _ = pool.Client.Received(1).BeginTransactionAsync(
-                Arg.Do<BeginTransactionRequest>(request =>
+            _ = pool.Client.Received(1).ExecuteSqlAsync(
+                Arg.Do<ExecuteSqlRequest>(request =>
                 {
                     Assert.Equal(s_sampleSessionName, request.SessionAsSessionName);
-                    Assert.Equal(ModeOneofCase.ReadOnly, request.Options.ModeCase);
+                    Assert.Null(request.Transaction);
                 }),
                 Arg.Any<CallSettings>());
-
-            Assert.NotSame(newSession, pooledSession);
-            Assert.Equal(pooledSession.SessionName, newSession.SessionName);
-            Assert.Equal(s_readOnlyOptions, newSession.TransactionOptions);
-            Assert.Equal(expectedTransaction.Id, newSession.TransactionId);
-            Assert.Equal(expectedTransaction.ReadTimestamp, newSession.ReadTimestamp);
         }
 
-
         [Fact]
-        public async Task RequestSessionIsPopulated()
+        public async Task SingleUseTransactionOptions()
         {
             var pool = new FakeSessionPool();
-            var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName);
+            var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName).WithTransactionOptions(s_readOnlyOptions, true);
 
             // Make a successful request
             pool.Client.Configure()
-                .BeginTransactionAsync(Arg.Any<BeginTransactionRequest>(), Arg.Any<CallSettings>())
-                .Returns(Task.FromResult(new Transaction()));
+                .ExecuteSqlAsync(Arg.Any<ExecuteSqlRequest>(), Arg.Any<CallSettings>())
+                .Returns(Task.FromResult(new ResultSet()));
 
-            await pooledSession.WithFreshTransactionAsync(s_readOnlyOptions, default);
+            await pooledSession.ExecuteSqlAsync(new ExecuteSqlRequest(), default);
 
-            _ = pool.Client.Received(1).BeginTransactionAsync(
-                Arg.Do<BeginTransactionRequest>(request => Assert.Equal(s_sampleSessionName, request.SessionAsSessionName)),
+            _ = pool.Client.Received(1).ExecuteSqlAsync(
+                Arg.Do<ExecuteSqlRequest>(request =>
+                {
+                    Assert.Equal(s_sampleSessionName, request.SessionAsSessionName);
+                    Assert.Equal(s_readOnlyOptions, request.Transaction.SingleUse);
+                }),
+                Arg.Any<CallSettings>());
+        }
+
+        [Fact]
+        public async Task ExplicitTransaction()
+        {
+            var transactionId = ByteString.CopyFromUtf8("transaction");
+            var pool = new FakeSessionPool();
+            var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName).WithTransaction(transactionId, s_readWriteOptions, false);
+
+            // Make a successful request
+            pool.Client.Configure()
+                .ExecuteSqlAsync(Arg.Any<ExecuteSqlRequest>(), Arg.Any<CallSettings>())
+                .Returns(Task.FromResult(new ResultSet()));
+
+            await pooledSession.ExecuteSqlAsync(new ExecuteSqlRequest(), default);
+
+            _ = pool.Client.Received(1).ExecuteSqlAsync(
+                Arg.Do<ExecuteSqlRequest>(request =>
+                {
+                    Assert.Equal(s_sampleSessionName, request.SessionAsSessionName);
+                    Assert.Equal(transactionId, request.Transaction.Id);
+                }),
+                Arg.Any<CallSettings>());
+        }
+
+        [Fact]
+        public async Task InlinedTransaction()
+        {
+            var transactionId = ByteString.CopyFromUtf8("transaction");
+            var pool = new FakeSessionPool();
+            var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName).WithTransactionOptions(s_readWriteOptions, false);
+
+            // Make a successful request
+            pool.Client.Configure()
+                .ExecuteSqlAsync(Arg.Any<ExecuteSqlRequest>(), Arg.Any<CallSettings>())
+                .Returns(Task.FromResult(new ResultSet
+                {
+                    Metadata = new ResultSetMetadata
+                    {
+                        Transaction = new Transaction { Id = transactionId }
+                    }
+                }));
+
+            await pooledSession.ExecuteSqlAsync(new ExecuteSqlRequest(), default);
+
+            Assert.Equal(transactionId, pooledSession.TransactionId);
+
+            _ = pool.Client.Received(1).ExecuteSqlAsync(
+                Arg.Do<ExecuteSqlRequest>(request =>
+                {
+                    Assert.Equal(s_sampleSessionName, request.SessionAsSessionName);
+                    Assert.Equal(s_readWriteOptions, request.Transaction.Begin);
+                }),
                 Arg.Any<CallSettings>());
         }
 
@@ -190,114 +234,9 @@ namespace Google.Cloud.Spanner.V1.Tests
             var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName);
             var transactionId = ByteString.CopyFromUtf8("transaction");
             var sessionWithTransaction = pooledSession.WithTransaction(transactionId, s_partitionedDml, singleUseTransaction: false);
-            Assert.Equal(s_partitionedDml, sessionWithTransaction.TransactionOptions);
+            Assert.Equal(s_partitionedDml.ModeCase, sessionWithTransaction.TransactionMode);
             Assert.Equal(transactionId, sessionWithTransaction.TransactionId);
             Assert.Equal(s_sampleSessionName, sessionWithTransaction.SessionName);
-        }
-
-        [Fact]
-        public async Task RequestTransactionIsPopulated()
-        {
-            var pool = new FakeSessionPool();
-            var transactionId = ByteString.CopyFromUtf8("transaction");
-            var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName);
-            var sessionWithTransaction = pooledSession.WithTransaction(transactionId, s_readWriteOptions, singleUseTransaction: false);
-
-            // Make a successful request
-            var request = new CommitRequest();
-            pool.Client.Configure().CommitAsync(request, Arg.Any<CallSettings>())
-                .Returns(Task.FromResult(new CommitResponse()));
-            await sessionWithTransaction.CommitAsync(request, null);
-
-            // The call modifies the request. (We can't easily check that it was modified before the RPC)
-            Assert.Equal(s_sampleSessionName, request.SessionAsSessionName);
-            Assert.Equal(transactionId, request.TransactionId);
-
-            _ = pool.Client.Received(1).CommitAsync(request, Arg.Any<CallSettings>());
-        }
-
-        // TODO: Revisit the names of the following 4 tests.
-
-        [Fact]
-        public async Task ExecuteSqlAsync_RequestTransactionIsPopulatedWhenNotPresent()
-        {
-            var pool = new FakeSessionPool();
-            var transactionId = ByteString.CopyFromUtf8("transaction");
-            var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName);
-            var sessionWithTransaction = pooledSession.WithTransaction(transactionId, s_readWriteOptions, singleUseTransaction: false);
-
-            // Make a successful request
-            var request = new ExecuteSqlRequest();
-            pool.Client.Configure().ExecuteSqlAsync(request, Arg.Any<CallSettings>())
-                .Returns(Task.FromResult(new ResultSet()));
-            await sessionWithTransaction.ExecuteSqlAsync(request, null);
-
-            // The call modifies the request. (We can't easily check that it was modified before the RPC)
-            Assert.Equal(s_sampleSessionName, request.SessionAsSessionName);
-            Assert.Equal(transactionId, request.Transaction.Id);
-
-            _ = pool.Client.Received(1).ExecuteSqlAsync(request, Arg.Any<CallSettings>());
-        }
-
-        [Fact]
-        public async Task ExecuteSqlAsync_RequestTransactionIsLeftAloneWhenPresent()
-        {
-            var pool = new FakeSessionPool();
-            var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName);
-
-            // Make a successful request
-            var request = new ExecuteSqlRequest { Transaction = new TransactionSelector { Begin = s_readOnlyOptions } };
-            pool.Client.Configure().ExecuteSqlAsync(request, Arg.Any<CallSettings>())
-                .Returns(Task.FromResult(new ResultSet()));
-            await pooledSession.ExecuteSqlAsync(request, null);
-
-            // The call modifies the request's session, but not transaction.
-            Assert.Equal(s_sampleSessionName, request.SessionAsSessionName);
-            Assert.Equal(TransactionSelector.SelectorOneofCase.Begin, request.Transaction.SelectorCase);
-            Assert.Equal(s_readOnlyOptions, request.Transaction.Begin);
-
-            _ = pool.Client.Received(1).ExecuteSqlAsync(request, Arg.Any<CallSettings>());
-        }
-
-        [Fact]
-        public async Task ExecuteBatchDmlAsync_RequestTransactionIsPopulatedWhenNotPresent()
-        {
-            var pool = new FakeSessionPool();
-            var transactionId = ByteString.CopyFromUtf8("transaction");
-            var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName);
-            var sessionWithTransaction = pooledSession.WithTransaction(transactionId, s_readWriteOptions, singleUseTransaction: false);
-
-            // Make a successful request
-            var request = new ExecuteBatchDmlRequest();
-            pool.Client.Configure().ExecuteBatchDmlAsync(request, Arg.Any<CallSettings>())
-                .Returns(Task.FromResult(new ExecuteBatchDmlResponse()));
-            await sessionWithTransaction.ExecuteBatchDmlAsync(request, null);
-
-            // The call modifies the request. (We can't easily check that it was modified before the RPC)
-            Assert.Equal(s_sampleSessionName, request.SessionAsSessionName);
-            Assert.Equal(transactionId, request.Transaction.Id);
-
-            _ = pool.Client.Received(1).ExecuteBatchDmlAsync(request, Arg.Any<CallSettings>());
-        }
-
-        [Fact]
-        public async Task ExecuteBatchDmlAsync_RequestTransactionIsLeftAloneWhenPresent()
-        {
-            var pool = new FakeSessionPool();
-            var pooledSession = PooledSession.FromSessionName(pool, s_sampleSessionName);
-
-            // Make a successful request
-            var request = new ExecuteBatchDmlRequest { Transaction = new TransactionSelector { Begin = s_readOnlyOptions } };
-            pool.Client.Configure().ExecuteBatchDmlAsync(request, Arg.Any<CallSettings>())
-                .Returns(Task.FromResult(new ExecuteBatchDmlResponse()));
-            await pooledSession.ExecuteBatchDmlAsync(request, null);
-
-            // The call modifies the request's session, but not transaction.
-            Assert.Equal(s_sampleSessionName, request.SessionAsSessionName);
-            Assert.Equal(TransactionSelector.SelectorOneofCase.Begin, request.Transaction.SelectorCase);
-            Assert.Equal(s_readOnlyOptions, request.Transaction.Begin);
-
-            _ = pool.Client.Received(1).ExecuteBatchDmlAsync(request, Arg.Any<CallSettings>());
         }
 
         [Fact]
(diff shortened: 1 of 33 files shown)


amanda-tarafa:
@olavloite In case you want to review.
@bshaffer FYI

amanda-tarafa:
(And the breaking changes are expected)

jskeet:
Thanks - I'll try to get to the review today; I may well send out review notes on a per-commit basis.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PooledSessionTests.cs
```diff
@@ -35,6 +35,9 @@ public class PooledSessionTests
         private static readonly SessionName s_sampleSessionName = new SessionName("project", "instance", "database", "session");
         // In various tests we don't care whether things are evicted "on" or "after" a particular tick, so add one tick.
         private static readonly TimeSpan OneTick = TimeSpan.FromTicks(1);
+        private static readonly TransactionOptions s_readWriteOptions = new TransactionOptions { ReadWrite = new Types.ReadWrite() };
```
jskeet, line 38: Semi-nit, but worth bearing in mind for the future - oneof cases like this can use target-typed new effectively:

```csharp
new TransactionOptions { ReadWrite = new() };
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -55,14 +55,22 @@ public sealed class PooledSession : IDisposable
...
+        /// </remarks>
+        internal TransactionOptions TransactionOptions { get; }
 
         /// <summary>
-        /// The read timestamp of the transaction. (Always <c>null</c> if
-        /// ReturnReadTimestamp = false or if TransactionMode != ReadOnly.)
+        /// The read timestamp of the transaction. May be null.
         /// </summary>
+        /// <remarks>
+        /// Will be set iif <see cref="TransactionOptions.ReadOnly"/> have <see cref="Types.ReadOnly.ReturnReadTimestamp"/>
```
jskeet, line 103: have => has? And possibly rewrite it to

```csharp
> /// iif <see cref="TransactionOptions>" represents read-only options with <see cref="Types.ReadOnly.ReturnReadTimestamp"/>
> set to true, of if [...]
```

(And change "and a values was provide" to "and a value was provided".)

I'm curious about the ReturnReadTimestamp bit though... if the options say what kind of transaction the session *will* represent, but we haven't already begun the transaction, how do we have a timestamp?

What about if it's read-only with a specified timestamp in the timestamp bound?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -42,6 +41,7 @@ namespace Google.Cloud.Spanner.V1
     /// </remarks>
     public sealed class PooledSession : IDisposable
     {
+        internal static readonly TransactionOptions NoTransactionOptions = new TransactionOptions();
```
jskeet, line 44: This makes me slightly nervous, because it's mutable. Internal readonly fields are just about okay if they're immutable, but it's hard to check that nothing's mutating this.

Could we make it private, but allow null in the constructors, in which case we use NoTransactionOptions (without even cloning, potentially)?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -55,14 +55,22 @@ public sealed class PooledSession : IDisposable
...
         /// <summary>
-        /// The mode of the transaction. (Always None iff TransactionId is null.)
+        /// The options for the transaction that is or will be associated with this session. Won't be null.
         /// </summary>
-        internal ModeOneofCase TransactionMode { get; }
+        /// <remarks>
+        /// Will be <see cref="TransactionOptions.ModeOneofCase.None"/> if <see cref="TransactionId"/>
+        /// is null.
+        /// </remarks>
+        internal TransactionOptions TransactionOptions { get; }
```
jskeet, line 90: Should everyone avoid mutating this?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
```diff
@@ -227,7 +227,17 @@ public PooledSession CreateDetachedSession(SessionName sessionName, ByteString t
         {
             GaxPreconditions.CheckNotNull(sessionName, nameof(sessionName));
             GaxPreconditions.CheckNotNull(transactionId, nameof(transactionId));
-            return PooledSession.FromSessionName(_detachedSessionPool, sessionName).WithTransaction(transactionId, transactionMode, readTimestamp);
+            return PooledSession.FromSessionName(_detachedSessionPool, sessionName).WithTransaction(transactionId, BuildTransactionOptions(), readTimestamp);
+
+            // TODO: Maybe just consider the breaking change and directly receive TransactionOptions.
+            TransactionOptions BuildTransactionOptions() => transactionMode switch
```
jskeet, line 250: Lots of scope for target-typed new here :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.cs
```diff
@@ -227,7 +227,17 @@ public PooledSession CreateDetachedSession(SessionName sessionName, ByteString t
         {
             GaxPreconditions.CheckNotNull(sessionName, nameof(sessionName));
             GaxPreconditions.CheckNotNull(transactionId, nameof(transactionId));
-            return PooledSession.FromSessionName(_detachedSessionPool, sessionName).WithTransaction(transactionId, transactionMode, readTimestamp);
+            return PooledSession.FromSessionName(_detachedSessionPool, sessionName).WithTransaction(transactionId, BuildTransactionOptions(), readTimestamp);
+
+            // TODO: Maybe just consider the breaking change and directly receive TransactionOptions.
```
jskeet, line 249: Am I right in saying detached sessions are likely to change a bit in other ways anyway?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -152,18 +166,13 @@ internal PooledSession WithTransaction(ByteString transactionId, ModeOneofCase t
         /// session. For instance, when retrying aborted commits it is better if the transaction work and commit
         /// are retried with the same session, because after each abort the sessions' lock priority increments.
         /// </remarks>
-        /// <returns>A new instance of <see cref="PooledSession"/>.</returns>
-        /// <exception cref="InvalidOperationException">If this <see cref="PooledSession.TransactionMode"/>
-        /// is <see cref="ModeOneofCase.None"/>.</exception>
-        public Task<PooledSession> WithFreshTransactionOrNewAsync(TransactionOptions transactionOptions, CancellationToken cancellationToken)
+        public Task<PooledSession> WithFreshTransactionOrNewAsync(CancellationToken cancellationToken)
```
jskeet: I'm really hoping this either goes away in a later commit, or makes more sense to me.

jskeet (commented):
Comments from commit 1.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -179,14 +178,14 @@ private async Task<PooledSession> AcquireSessionImplAsync(TransactionOptions tra
...
-                    if (session.TransactionMode == transactionMode)
+                    if (session.TransactionOptions.ModeCase == transactionMode)
                     {
                         Interlocked.Increment(ref _rwTransactionRequestsPrewarmed);
                     }
                 }
 
                 // If we've already got the right transaction mode, we're done.
-                if (session.TransactionMode == transactionMode)
+                if (session.TransactionOptions.ModeCase == transactionMode)
```
jskeet: Is that actually the case? If we've been given transaction options for a read-only transaction, but with a different read_timestamp to the one in the pool for example, we can't just use the same pooled session, presumably?

jskeet (commented):
One more comment from the first commit.

jskeet (commented):
Reviewed second commit: I like it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -134,9 +134,10 @@ internal async Task<SpannerDataReader> ExecuteReaderAsync(CommandBehavior behavi
                 {
                     throw new InvalidOperationException("singleUseReadSettings cannot be used within another transaction.");
                 }
-                effectiveTransaction = effectiveTransaction ?? new EphemeralTransaction(Connection, Priority);
+                effectiveTransaction = effectiveTransaction
+                    ?? new EphemeralTransaction(Connection, Priority, singleUseReadSettings is null ? null : singleUseReadSettings.ToTransactionOptions());
```
jskeet, line 138: Can we just use `singleUseReadSettings?.ToTransactionOptions()` for this?

jskeet (commented):
3rd commit: I don't understand what a single use transaction is. Aren't all transactions single-use?

I think I'll probably need a video call to chat about that one.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Internal/ExecuteHelper.cs
```diff
@@ -89,5 +90,55 @@ private static string GetResourceInfoTypeFromTrailers(RpcException exception)
                 return null;
             }
         }
+
+        /// <summary>
```
jskeet, line 94: This looks familiar - I know we've got the TCS version in GAX, but does this exist somewhere else?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -44,15 +44,40 @@ public sealed class PooledSession : IDisposable
...
         /// </summary>
         public SessionName SessionName { get; }
 
+        private Transaction _transaction;
         /// <summary>
         /// The ID of the transaction. May be null.
         /// </summary>
-        public ByteString TransactionId { get; }
+        /// <remarks>
+        /// Transactions are acquiered when they are needed so this will be null in the
```
jskeet, line 61: acquiered => acquired

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -44,15 +44,40 @@ public sealed class PooledSession : IDisposable
...
+        /// following cases:
+        /// <list type="bullet">
+        /// <item>
+        /// <see cref="TransactionOptions"/> is <see cref="TransactionOptions.ModeOneofCase.None"/>,
+        /// inlcuding when the session is idle in the pool.
```
jskeet, line 66: including

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -44,15 +44,40 @@ public sealed class PooledSession : IDisposable
...
+        /// </item>
+        /// <item>
+        /// <see cref="SingleUseTransaction"/> is true. No transactions will exist client side.
+        /// </item>
+        /// <item>
+        /// The session has been acquired from the pool with some trasaction options but no command execution has
```
jskeet, line 72: transaction

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+                    ReadTimestamp = readTimestamp
+                };
+            }
             _session = new Session { SessionName = SessionName };
             _evictionTime = evictionTime;
             _refreshTicks = refreshTicks;
         }
 
+        /// <summary>
+        /// Decides when we need a transaction ID or not and when that can be obtained by transaction inlining.
```
jskeet, line 170: when => whether in both cases, I think.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+        /// <param name="transactionSelectorSetter">Called by this method to set the correct transaction selector.</param>
+        /// <param name="commandAsync">The command (RPC) that will be executed after the transaction selector has maybe been set.
+        /// Callers may fail if command is executed but no transaction selector has been set.</param>
+        /// <param name="inlinedTransactionExtractor">Will extract transaction information from the command's response if
+        /// transaction inlining was succesful. May be null, which indicates that command does not support transaction inlining.</param>
+        /// <param name="skipTransactionCreation">If true, transaction creation may be skipped. This is used by commit and rollback
+        /// so that a transaction is not created just for inmediate commit or rollback. Note that if there are pending mutations, commit
+        /// should set this parameter to false.</param>
+        /// <param name="cancellationToken">The cancellation token for the operation.</param>
+        /// <returns></returns>
```
jskeet, line 185: Fill in return and exception comments.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+            // we'll be awaiting on the task that does the modifying.
+            if (TransactionId is ByteString transactionId)
+            {
+                transactionSelectorSetter(new TransactionSelector { Id = transactionId });
+                return await commandAsync().ConfigureAwait(false);
+            }
+
+            // We now know that we don't have a transaction ID but we need one to execute
+            // the command we have been given.
+            // There are
```
jskeet, line 219: Incomplete comment

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+                {
+                    // We need to create a transaction.
+                    if (!skipTransactionCreation)
+                    {
+                        // The calling command does not support inilining
+                        // or the transaction mode Partitioned DML, which cannot be inline.
+                        // Either way we need to create a transaction explicitly.
+                        if (inlinedTransactionExtractor is null || TransactionOptions.ModeCase == ModeOneofCase.PartitionedDml)
+                        {
+                            _transactionCreationTask = Task.Run(() => SetExplicitTransactionAsync(cancellationToken), cancellationToken);
```
jskeet, line 247: I'm nervous about calling Task.Run in a lock. There are always weird cases where it runs synchronously. Not sure of an alternative though.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+
+            lock (_transactionCreationTaskLock)
+            {
+                // We are not creating a transaction. We might need to do so.
+                if (_transactionCreationTask is null)
+                {
+                    // We need to create a transaction.
+                    if (!skipTransactionCreation)
+                    {
+                        // The calling command does not support inilining
```
jskeet, line 242: inlining (or maybe inline transactions)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+                                    // We only wait for the _transactionCreationTaks to be done to guarantee
+                                    // that the transaction ID obtained via inlining has been stored and can
+                                    // be access via TransactionId. In turn this guarantees command executors
+                                    // that there's an ID available inmediately after a successful command execution.
+                                    await _transactionCreationTask.ConfigureAwait(false);
+                                    return response;
+                                }
+                                catch (Exception ex)
+                                {
+                                    Client.Settings.Logger.Warn($"Command execution with transaction inlining failed. " +
```
jskeet, line 297: Remove the $ here and below, given that it's just a plain string literal?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+                                    Client.Settings.Logger.Warn($"Transaction creation via inlining failed. " +
+                                        $"Attempting to begin an explicit transaction. ",
+                                        ex);
+                                    await SetExplicitTransactionAsync(cancellationToken).ConfigureAwait(false);
+                                }
+                            }, cancellationToken);
+
+                            // The command execution task is the combination of attempting inlining,
+                            // and if that fails, executing the command with the explicitly created transaction.
+                            commandMaybeWithTransactionAsync = async () =>
```
jskeet, line 282: Perhaps move this to a local function? It's weird to review this but know that it's actually *not* called in the lock.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+                        // The calling command does not support inilining
+                        // or the transaction mode Partitioned DML, which cannot be inline.
+                        // Either way we need to create a transaction explicitly.
+                        if (inlinedTransactionExtractor is null || TransactionOptions.ModeCase == ModeOneofCase.PartitionedDml)
+                        {
+                            _transactionCreationTask = Task.Run(() => SetExplicitTransactionAsync(cancellationToken), cancellationToken);
+                            commandMaybeWithTransactionAsync = () => CommandWithTransactionAsync(cancellationToken);
+                        }
+                        // The calling command supports inlining.
+                        // We attempt inlining but if that fails, we create a transaction explicitly.
```
jskeet, line 251: What sort of failure are we talking about here?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+                            // and if that fails, explicitly creating a transaction.
+                            _transactionCreationTask = Task.Run(async () =>
+                            {
+                                try
+                                {
+                                    await SetInlinedTransactionAsync(commandWithInliningTask).ConfigureAwait(false);
+                                }
+                                catch (Exception ex)
+                                {
+                                    Client.Settings.Logger.Warn($"Transaction creation via inlining failed. " +
```
jskeet, line 273: Remove $ here and below; remove trailing space from second line.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -115,20 +140,236 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
...
+                                    // that the transaction ID obtained via inlining has been stored and can
+                                    // be access via TransactionId. In turn this guarantees command executors
+                                    // that there's an ID available inmediately after a successful command execution.
+                                    await _transactionCreationTask.ConfigureAwait(false);
+                                    return response;
+                                }
+                                catch (Exception ex)
+                                {
+                                    Client.Settings.Logger.Warn($"Command execution with transaction inlining failed. " +
+                                        $"Waiting for an explicit transaction to be created to attempt command execution. ",
```
jskeet, line 298: Remove trailing space.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -315,17 +556,49 @@ public void DetachFromPool()
...
+                }
+            }
+
+            async Task<CommitResponse> CommitAsync()
+            {
+                // If a transaction had been started, by now SetTransaction should have been called with a transaction ID.
+                // If not, there's an attempt to commit a non-existent transaction.
+                if (request.TransactionId is null || request.TransactionId.IsEmpty)
+                {
+                    throw new InvalidOperationException("Cannot commit a PooledSession with no associated transaction." +
```
jskeet, line 580: Nit: add space at end.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -315,17 +556,49 @@ public void DetachFromPool()
...
-            MarkAsCommittedOrRolledBack();
-            return response;
+            return ExecuteMaybeWithTransactionSelectorAsync(
+                transactionSelectorSetter: SetTransaction,
+                commandAsync: CommitAsync,
+                inlinedTransactionExtractor: null, // Commit does not support inline transactions.
+                skipTransactionCreation: request.Mutations.Count == 0, // If there are only mutations we won't have a transaction but we need one.
+                callSettings?.CancellationToken ?? default);
+
+            void SetTransaction(TransactionSelector transactionSelector)
```
jskeet, line 559: Maybe rename this (and similar methods) to SetTransactionInRequest or similar, so that we know "SetTransaction" always means "set transaction in the pooled session".

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -315,17 +556,49 @@ public void DetachFromPool()
...
+                switch (transactionSelector.SelectorCase)
+                {
+                    case TransactionSelector.SelectorOneofCase.Id:
+                        request.TransactionId = transactionSelector.Id;
+                        break;
+                    case TransactionSelector.SelectorOneofCase.Begin:
+                        throw new InvalidOperationException("Commit does not support inline transactions. This is a bug in library code.");
+                    case TransactionSelector.SelectorOneofCase.SingleUse:
+                        throw new InvalidOperationException("A single use transaction cannot be committed.");
+                    default: throw new InvalidOperationException("Cannot commit a PooledSession with no associated transaction");
```
jskeet, line 570: Nit: New line for the exception.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -378,11 +687,43 @@ internal Task<PartitionResponse> PartitionReadOrQueryAsync(PartitionReadOrQueryR
...
+            {
+                switch (transactionSelector.SelectorCase)
+                {
+                    case TransactionSelector.SelectorOneofCase.Id:
+                    case TransactionSelector.SelectorOneofCase.Begin:
+                        request.Transaction = transactionSelector;
+                        break;
+                    case TransactionSelector.SelectorOneofCase.SingleUse:
+                        throw new InvalidOperationException("A single use transaction cannot be used for creating partitioned reads or queries.");
+                    default: throw new InvalidOperationException("Cannot call PartitionReadOrQueryAsync with no associated transaction.");
```
jskeet, line 696: Nit: New line for the exception.

jskeet (commented):
4th commit review: I'll need to look at this more closely, but I believe I understand.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/PooledSessionTests.cs
```diff
@@ -434,7 +434,7 @@ private PooledSession CreateWithTransaction(SessionPool.ISessionPool pool, Trans
             return PooledSession.FromSessionName(pool, s_sampleSessionName).WithTransaction(transactionId, options, singleUseTransaction: false);
         }
 
-        private class FakeSessionPool : SessionPool.ISessionPool
+        internal class FakeSessionPool : SessionPool.ISessionPool
```
jskeet, line 377: Maybe promote to a top-level class? (Doesn't need to be in this PR.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -52,6 +51,8 @@ public sealed class PooledSession : IDisposable
         /// </summary>
         public SessionName SessionName { get; }
 
+        internal Session Session { get; }
```
jskeet, line 54: Doc comment would be handy here - along with the relationship to `SessionName`. (Can we make SessionName just delegate to Session.SessionName?)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ResultStream.cs
```diff
@@ -131,16 +131,50 @@ private async Task<PartialResultSet> ComputeNextAsync(CancellationToken cancella
 
                 try
                 {
-                    if (_grpcCall == null)
+                    bool hasNext = false;
```
jskeet, line 134: It's a pity we have to have this here rather than within PooledSession. I assume it's just rather tricky to put it in PooledSession...

jskeet (commented):
Commit 5 review: it feels unfortunate, but probably unavoidable.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -395,29 +395,16 @@ private PooledSession AfterReset()
...
-                {
-                    Options = transactionOptions,
-                    SessionAsSessionName = SessionName,
-                };
-                var callSettings = Client.Settings.BeginTransactionSettings
-                    .WithExpiration(Expiration.FromTimeout(_pool.Options.Timeout))
-                    .WithCancellationToken(cancellationToken);
-                return RecordSuccessAndExpiredSessions(Client.BeginTransactionAsync(request, callSettings));
-            }
+            MarkAsDisposed();
```
jskeet, line 406: This prevents it from being reused, right? Was it a bug that we didn't have that before?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -175,29 +167,16 @@ private async Task<PooledSession> AcquireSessionImplAsync(TransactionOptions tra
...
-                    Interlocked.Increment(ref _rwTransactionRequests);
-                    if (session.TransactionOptions.ModeCase == transactionMode)
-                    {
-                        Interlocked.Increment(ref _rwTransactionRequestsPrewarmed);
-                    }
-                }
-
-                // Whatever session we got it either has vanilla transaction options or the wrong ones.
-                return session.WithTransactionOptions(transactionOptions, singleUseTransaction);
+                // While in the pool, sessions have TransactionOptions.ModeCase == ModeOneofCase.None and singleUseTransaction == false.
```
jskeet, line 170: This makes me wonder whether we've got the model quite right... whether we should consider refactoring to "sessions idle in the pool" and "sessions from the pool, now being used" as different types. But let's chat about that.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SessionPool.TargetedSessionPool.cs
```diff
@@ -426,24 +380,10 @@ private void EvictAndRefreshSessions()
...
-                    EvictSession(session);
-                }
-                foreach (var session in staleSessions)
-                {
-                    Parent.ConsumeBackgroundTask(RefreshAsync(session), "session refresh");
-                }
-
-                void RemoveStaleOrExpiredItemsFromStack(ConcurrentStack<PooledSession> stack)
-                {
+                    // Remove stale or expired sessions from the stack
```
jskeet, line 383: Possibly remove any notion of a "stack"?

jskeet (commented):
Okay, I think I broadly understand, but I definitely want to chat more before we merge.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -61,15 +60,20 @@ public sealed class PooledSession : IDisposable
         /// Will be <see cref="TransactionOptions.ModeOneofCase.None"/> if <see cref="TransactionId"/>
         /// is null.
         /// </remarks>
-        internal TransactionOptions TransactionOptions { get; }
+        private TransactionOptions TransactionOptions { get; }
+
+        /// <summary>
+        /// The transaction mode for the transaction that is or may be associated with this session.
```
jskeet, line 96: And is this *only* for testing in the end? (If so, let's say so.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -61,15 +60,20 @@ public sealed class PooledSession : IDisposable
...
+        /// </summary>
+        internal TransactionOptions.ModeOneofCase TransactionMode => TransactionOptions.ModeCase;
 
         /// <summary>
         /// The read timestamp of the transaction. May be null.
         /// </summary>
         /// <remarks>
-        /// Will be set iif <see cref="TransactionOptions.ReadOnly"/> have <see cref="Types.ReadOnly.ReturnReadTimestamp"/>
+        /// Will be set iif <see cref="TransactionOptions"/> represents read-only options with <see cref="Types.ReadOnly.ReturnReadTimestamp"/>
```
jskeet: Even if ReturnReadTimestamp is true, this will only return a non-null value if we've already created the transaction, right? Or do we never end up with a pooled session for read-only transaction options but without the transaction already populated?

(I'm more concerned about my own mental model than nit-picking the docs here.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -101,14 +105,14 @@ public sealed class PooledSession : IDisposable
 
         private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, ByteString transactionId, TransactionOptions transactionOptions, Timestamp readTimestamp, DateTime evictionTime, long refreshTicks)
```
jskeet: It's probably worth adding docs to this constructor, in terms of when it's used.
Because we can end up with transaction options but no transaction ID, as far as I can see (unless I've misunderstood) - it's just we presumably use a different constructor in that case?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -124,18 +128,18 @@ internal static PooledSession FromSessionName(SessionPool.ISessionPool pool, Ses
             var now = pool.Clock.GetCurrentDateTimeUtc();
             var refreshDelay = options.SessionRefreshJitter.GetDelay(options.IdleSessionRefreshDelay);
             var evictionDelay = options.SessionEvictionJitter.GetDelay(options.PoolEvictionDelay);
-            return new PooledSession(pool, sessionName, transactionId: null, NoTransactionOptions, readTimestamp: null, now + evictionDelay, now.Ticks + refreshDelay.Ticks);
+            return new PooledSession(pool, sessionName, transactionId: null, transactionOptions: null, readTimestamp: null, now + evictionDelay, now.Ticks + refreshDelay.Ticks);
         }
 
         /// <summary>
-        /// Returns a PooledSession for the same session as this one with <see cref="TransactionOptions"/> set to <see cref="NoTransactionOptions"/>.
+        /// Returns a PooledSession for the same session as this one with <see cref="TransactionOptions"/> set to <see cref="TransactionOptions.ModeOneofCase.None"/>.
```
jskeet, line 396: `<see cref="TransactionOptions"/>` => `<see cref="TransactionMode"/>`?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
jskeet: (Unless this happens later on, of course - or my mental model is wrong. Will check, and resolve this comment if it becomes obvious.)

jskeet (commented):
A few nits and bits for us to chat about.

jskeet (approved):

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -91,22 +141,245 @@ public sealed class PooledSession : IDisposable
...
+            GaxPreconditions.CheckArgument(
+                readTimestamp is null || transactionId is not null,
+                nameof(readTimestamp),
+                "A read timestamp can only be specified if a transaction ID is also being specified.");
+
+            SingleUseTransaction = singleUseTransaction;
+            if (transactionId is not null)
+            {
+                // We don't need to lock here, this is the constructor.
+                _transaction = new Transaction
```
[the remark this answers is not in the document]
amanda-tarafa: I think this is so that users doing partitions need only remember the transaction ID and not the rest of the transaction fields? But possibly we can handle that elsewhere and here just receive the whole Transaction. I'll leave a TODO to explore that when I look at detached transactions.

jskeet (approved):

amanda-tarafa:
Closing and reopening for the breaking change detector benefit.

PR #11368, opened 2023-11-29
feat: Storage object retention lock

### LINKED ISSUE #11227: storage: support object retention feature
Please implement the object retention feature. The following operations must be supported:

- Create a bucket with object retention enabled by using the `enableObjectRetention` bucket create boolean option at bucket create time
- Get the object retention enabled status of a bucket
- Create an object with retention enabled (on an retention enabled bucket) by populating the `retention` object at object create time (retention object has the fields `mode` which can be Unlocked or Locked, and `retainUntilTime` which is a timestamp)
- Update retention for an existing object with a patch operation, by passing the `overrideUnlockedRetention` boolean patch option
- Get the retention policy of an object
- Remove retention for an existing object the same way as above

Details in go/gcs-object-retention-client-request

You can use the [java implementation](https://github.com/googleapis/java-storage/pull/2277) as a reference.

The kokoro test project should be allowlisted shortly, let me know if you need access to an allowlisted project for your own testing. 

We're aiming to get this in by the November code freeze, let me know if you need any help.

anuragsrivstv:
Fixes #11227.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
@@ -111,6 +111,64 @@ namespace Google.Cloud.Storage.V1.IntegrationTests
         }
 
         /// <summary>
+        /// Tests for object retention policy.
+        /// </summary>
+        [Fact]
+        public void ObjectRetentionTests()
+        {
+            var client = _fixture.Client;
+            var bucketName = _fixture.GenerateBucketName();
+
+            // Create a bucket with object retention enabled.
+            var createBucketoptions = new CreateBucketOptions { EnableObjectRetention = true };
+            var bucket = _fixture.CreateBucket(bucketName, false, createBucketoptions);
+            bucket = client.GetBucket(bucketName);
+            Assert.Equal("Enabled", bucket.ObjectRetention?.Mode);
+
+            var baseDateTime = DateTimeOffset.UtcNow;
+
+            string objectName = "object.txt";
+            // Create an object, with retention configured.\
+            var newObject = new Object
+            {
+                Bucket = bucketName,
+                Name = objectName,
+                ContentType = "text/plain",
+                Retention = new Object.RetentionData
+                {
+                    Mode = "Unlocked",
+                    RetainUntilTimeDateTimeOffset = baseDateTime.AddDays(5),
+                }
+
+            };
+            CreateObject(newObject);
+            var obj = client.GetObject(bucketName, objectName);
+            Assert.Equal("Unlocked", obj.Retention?.Mode);
+            Assert.Equal(baseDateTime.AddDays(5).Date, obj.Retention.RetainUntilTimeDateTimeOffset.Value.Date);
+
+            Assert.Throws<GoogleApiException>(() => client.DeleteObject(bucketName, objectName));
+
+            // Update the retainUntilTime of the object
+            var updateObjectOptions = new UpdateObjectOptions { OverrideUnlockedRetention = true, };
+            obj.Retention.Mode = "Unlocked";
+            obj.Retention.RetainUntilTimeDateTimeOffset = baseDateTime.AddDays(2);
+            client.UpdateObject(obj, updateObjectOptions);
+            var updatedObj = client.GetObject(bucketName, objectName);
+            Assert.Equal(baseDateTime.AddDays(2).Date, updatedObj.Retention?.RetainUntilTimeDateTimeOffset.Value.Date);
+
+            Assert.Throws<GoogleApiException>(() => client.DeleteObject(bucketName, objectName));
+
+            // Remove retention for existing object.
+            updatedObj.Retention = null;
+            var removeobjectRetentionOption = new UpdateObjectOptions { OverrideUnlockedRetention = true, };
+            client.UpdateObject(updatedObj, removeobjectRetentionOption);
+            var retentionRemovedObj = client.GetObject(bucketName, objectName);
+            Assert.Null(retentionRemovedObj.Retention);
+
+            client.DeleteObject(bucketName, objectName);
+        }
+
+        /// <summary>
         /// Test Set/Update temporary hold on an object and attempt to delete an object.
         /// It should fail until temporaryHold is set to false.
         /// </summary>
@@ -205,6 +263,13 @@ namespace Google.Cloud.Storage.V1.IntegrationTests
             string text = $"This is the content of {objectName}";
             var bytes = Encoding.UTF8.GetBytes(text);
             return _fixture.Client.UploadObject(bucketName, objectName, "text/plain", new MemoryStream(bytes));
+        }
+
+        private Object CreateObject(Object obj)
+        {
+            string text = $"This is the content of {obj.Name}";
+            var bytes = Encoding.UTF8.GetBytes(text);
+            return _fixture.Client.UploadObject(obj, new MemoryStream(bytes));
         }
     }
 }
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CreateBucketOptions.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CreateBucketOptions.cs
@@ -43,6 +43,11 @@ namespace Google.Cloud.Storage.V1
         /// </summary>
         public RetryOptions RetryOptions { get; set; }
 
+        /// <summary>
+        /// When set to true, object retention is enabled for this bucket.
+        /// </summary>
+        public bool? EnableObjectRetention { get; set; }
+
         internal void ModifyRequest(BucketsResource.InsertRequest request)
         {
             if (PredefinedAcl != null)
@@ -58,6 +63,10 @@ namespace Google.Cloud.Storage.V1
             if (Projection != null)
             {
                 request.Projection = GaxPreconditions.CheckEnumValue((ProjectionEnum) Projection, nameof(Projection));
+            }
+            if (EnableObjectRetention != null)
+            {
+                request.EnableObjectRetention = EnableObjectRetention;
             }
         }
     }
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UpdateObjectOptions.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UpdateObjectOptions.cs
@@ -88,6 +88,12 @@ namespace Google.Cloud.Storage.V1
         /// </summary>
         public RetryOptions RetryOptions { get; set; }
 
+        /// <summary>
+        /// Must be true to remove the retention configuration, reduce its unlocked retention period, or change its
+        /// mode from unlocked to locked.
+        /// </summary>
+        public bool? OverrideUnlockedRetention { get; set; }
+
         private bool AnyExplicitPreconditions =>
             IfGenerationMatch != null || IfGenerationNotMatch != null || IfMetagenerationMatch != null || IfMetagenerationNotMatch != null;
 
@@ -148,6 +154,10 @@ namespace Google.Cloud.Storage.V1
             if (UserProject != null)
             {
                 request.UserProject = UserProject;
+            }
+            if (OverrideUnlockedRetention != null)
+            {
+                request.OverrideUnlockedRetention = OverrideUnlockedRetention;
             }
         }
     }
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/StorageFixture.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/StorageFixture.cs
@@ -242,9 +242,9 @@ namespace Google.Cloud.Storage.V1.IntegrationTests
 
         }
 
-        internal Bucket CreateBucket(string name, bool multiVersion)
+        internal Bucket CreateBucket(string name, bool multiVersion, CreateBucketOptions options = null)
         {
-            var bucket = Client.CreateBucket(ProjectId, new Bucket { Name = name, Versioning = new Bucket.VersioningData { Enabled = multiVersion } });
+            var bucket = Client.CreateBucket(ProjectId, new Bucket { Name = name, Versioning = new Bucket.VersioningData { Enabled = multiVersion } }, options);
             SleepAfterBucketCreateDelete();
             RegisterBucketToDelete(name);
             return bucket;
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/CreateBucketOptionsTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/CreateBucketOptionsTest.cs
@@ -39,12 +39,14 @@ namespace Google.Cloud.Storage.V1.Tests
             {
                 PredefinedAcl = PredefinedBucketAcl.AuthenticatedRead,
                 PredefinedDefaultObjectAcl = PredefinedObjectAcl.BucketOwnerFullControl,
-                Projection = Projection.Full
+                Projection = Projection.Full,
+                EnableObjectRetention = true
             };
             options.ModifyRequest(request);
             Assert.Equal(PredefinedAclEnum.AuthenticatedRead, request.PredefinedAcl);
             Assert.Equal(PredefinedDefaultObjectAclEnum.BucketOwnerFullControl, request.PredefinedDefaultObjectAcl);
             Assert.Equal(ProjectionEnum.Full, request.Projection);
+            Assert.True(request.EnableObjectRetention);
         }
     }
 }


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UploadObject.cs
```diff
@@ -48,6 +48,19 @@ public sealed partial class StorageClientImpl : StorageClient
...
+                destination.Retention = new Object.RetentionData();
+                if (options.RetentionMode != null)
+                {
+                    destination.Retention.Mode = options.RetentionMode.ToString();
+                }
+                if (options.RetainUntilTime != null)
+                {
+                    destination.Retention.RetainUntilTimeDateTimeOffset = options.RetainUntilTime;
+                }
+            }
```
amanda-tarafa, line 63: Can't we not do this on the UploadObjectOptions.ModifyMediaUpload method? That's why it's there for.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UpdateObject.cs
```diff
@@ -41,6 +41,19 @@ private ObjectsResource.UpdateRequest CreateUpdateObjectRequest(Object obj, Upda
...
+            {
+                obj.Retention = new Object.RetentionData
+                {
+                    Mode = options.RetentionMode,
+                    RetainUntilTimeDateTimeOffset = options.RetainUntilTime
+                };
+            }
+            else
+            {
+                obj.Retention = null;
```
amanda-tarafa, line 55: Do we need to set null explicitly?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UpdateObject.cs
```diff
@@ -41,6 +41,19 @@ private ObjectsResource.UpdateRequest CreateUpdateObjectRequest(Object obj, Upda
...
+            if (options?.RetentionMode != null || options?.RetainUntilTime != null)
+            {
+                obj.Retention = new Object.RetentionData
+                {
+                    Mode = options.RetentionMode,
+                    RetainUntilTimeDateTimeOffset = options.RetainUntilTime
+                };
+            }
+            else
+            {
```
amanda-tarafa, line 54: We can do all of this on the ModifyRequest method of the options as well.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CreateBucketOptions.cs
```diff
@@ -43,6 +43,11 @@ public sealed class CreateBucketOptions
         /// </summary>
         public RetryOptions RetryOptions { get; set; }
 
+        /// <summary>
+        /// When set to true, object retention is enabled for this bucket.
+        /// </summary>
+        public bool EnableObjectRetention { get; set; }
```
amanda-tarafa, line 49: This probably should be, and then we only set it if it has a value. This is to allow for explicitly setting one value or another.
```suggestion
        public bool? EnableObjectRetention { get; set; }
```

amanda-tarafa (changes requested):

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UpdateObject.cs
anuragsrivstv: Yes, we need to. This is for the use case where we need to remove retention from an object. So in that case fetched bucket object will already have a retention configure on it, which we need to mark as null and update.

jskeet:
(FYI, I'll wait until @amanda-tarafa is happy, then I'll do a final pass.)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UpdateObject.cs
anuragsrivstv: Okay, on thinking more on it, there could be also a use case where user is updating an object and at the same time wants existing retention of the object to be same as it is( not removed ). In that case they may not provide  RetentionMode and RetainUntilTime.
To Handle such cases as well do you suggest that we do not mark existing retention property of fetched object null here. Introduce a new bool property say options.RemoveRetention and mark retention of existing object as null only if it is set to true ? What do you think ?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UpdateObjectOptions.cs
```diff
@@ -149,6 +165,27 @@ internal void ModifyRequest(UpdateRequest request, Object obj)
             {
                 request.UserProject = UserProject;
             }
+            if (OverrideUnlockedRetention != null)
+            {
+                request.OverrideUnlockedRetention = OverrideUnlockedRetention;
+            }
+            if (RetentionMode != null || RetainUntilTime != null)
```
amanda-tarafa, line 172: Actually, I don't think we need these bits in the optios as calling code is able to build the object itself and set there whatever they need.

We only need OverrideUnlockRetention because that's a request field, and its the request what we build internally for the users.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UploadObjectOptions.cs
```diff
@@ -185,6 +195,18 @@ internal void ModifyMediaUpload(CustomMediaUpload upload)
...
+            if (RetentionMode != null || RetainUntilTime != null)
+            {
+                upload.Body.Retention = new Object.RetentionData();
+                if (RetentionMode != null)
+                {
+                    upload.Body.Retention.Mode = RetentionMode;
+                }
+                if (RetainUntilTime != null)
+                {
+                    upload.Body.Retention.RetainUntilTimeDateTimeOffset = RetainUntilTime;
```
amanda-tarafa, line 207: Similarly here, we have a pubic overload that acceots the object to upload so users can use that directly to set these.

amanda-tarafa (changes requested):
As for your question of how to unset a value vs. how to leave it as is withou explicitly specifying it, that's why we have both update and patch operations.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -206,5 +264,12 @@ private Object CreateObject(string bucketName, string objectName)
             var bytes = Encoding.UTF8.GetBytes(text);
             return _fixture.Client.UploadObject(bucketName, objectName, "text/plain", new MemoryStream(bytes));
         }
+
+        private Object CreateObject(Object obj)
```
amanda-tarafa, line 268: This one is probably better in the Fixture. Seems like we have one already that's local to another method, but we might be able to easily make it internal.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -110,6 +110,64 @@ public void LockRetentionPolicy()
...
+                {
+                    Mode = "Unlocked",
+                    RetainUntilTimeDateTimeOffset = baseDateTime.AddDays(5),
+                }
+
+            };
+            CreateObject(newObject);
+            var obj = client.GetObject(bucketName, objectName);
+            Assert.Equal("Unlocked", obj.Retention?.Mode);
+            Assert.Equal(baseDateTime.AddDays(5).Date, obj.Retention.RetainUntilTimeDateTimeOffset.Value.Date);
```
amanda-tarafa, line 147: Or am I missing something?

```suggestion
            Assert.Equal(baseDateTime.AddDays(5), obj.Retention.RetainUntilTimeDateTimeOffset.Value);
```

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -110,6 +110,64 @@ public void LockRetentionPolicy()
...
+
+            Assert.Throws<GoogleApiException>(() => client.DeleteObject(bucketName, objectName));
+
+            // Update the retainUntilTime of the object
```
amanda-tarafa, line 151: This should be another test.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -110,6 +110,64 @@ public void LockRetentionPolicy()
...
+            var updateObjectOptions = new UpdateObjectOptions { OverrideUnlockedRetention = true, };
+            obj.Retention.Mode = "Unlocked";
+            obj.Retention.RetainUntilTimeDateTimeOffset = baseDateTime.AddDays(2);
+            client.UpdateObject(obj, updateObjectOptions);
+            var updatedObj = client.GetObject(bucketName, objectName);
+            Assert.Equal(baseDateTime.AddDays(2).Date, updatedObj.Retention?.RetainUntilTimeDateTimeOffset.Value.Date);
+
+            Assert.Throws<GoogleApiException>(() => client.DeleteObject(bucketName, objectName));
+
+            // Remove retention for existing object.
```
amanda-tarafa, line 161: And this should be another test.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UpdateObjectOptions.cs
```diff
@@ -88,6 +88,12 @@ public sealed class UpdateObjectOptions
         /// </summary>
         public RetryOptions RetryOptions { get; set; }
 
+        /// <summary>
+        /// Must be true to remove the retention configuration, reduce its unlocked retention period, or change its
+        /// mode from unlocked to locked.
+        /// </summary>
+        public bool? OverrideUnlockedRetention { get; set; }
```
amanda-tarafa, line 95: You are not unit testing this change specifically.

amanda-tarafa (changes requested):
Looking better now, still some issues with tests.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
amanda-tarafa: And ping on this comment?

amanda-tarafa (commented):
@anuragsrivstv I've left a couple more comments, but it's fine if you prefer one of us to address them. Thanks!!

Also, @jskeet this is mostly ready for your review, save for my comments around tests.

jskeet:
I'll *try* to review this in the next couple of days, but there's a lot going on.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -110,6 +110,64 @@ public void LockRetentionPolicy()
...
+            {
+                Bucket = bucketName,
+                Name = objectName,
+                ContentType = "text/plain",
+                Retention = new Object.RetentionData
+                {
+                    Mode = "Unlocked",
+                    RetainUntilTimeDateTimeOffset = baseDateTime.AddDays(5),
+                }
+
```
jskeet, line 142: Nit: remove blank line.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -110,6 +110,64 @@ public void LockRetentionPolicy()
...
+            // Create a bucket with object retention enabled.
+            var createBucketoptions = new CreateBucketOptions { EnableObjectRetention = true };
+            var bucket = _fixture.CreateBucket(bucketName, false, createBucketoptions);
+            bucket = client.GetBucket(bucketName);
+            Assert.Equal("Enabled", bucket.ObjectRetention?.Mode);
+
+            var baseDateTime = DateTimeOffset.UtcNow;
+
+            string objectName = "object.txt";
+            // Create an object, with retention configured.\
```
jskeet, line 131: Nit: remove trailing backslash.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
jskeet: I would suggest truncating `baseDateTime` to the millisecond to start with (with a comment), then performing an exact comparison.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -110,6 +110,64 @@ public void LockRetentionPolicy()
...
+        /// Tests for object retention policy.
+        /// </summary>
+        [Fact]
+        public void ObjectRetentionTests()
+        {
+            var client = _fixture.Client;
+            var bucketName = _fixture.GenerateBucketName();
+
+            // Create a bucket with object retention enabled.
+            var createBucketoptions = new CreateBucketOptions { EnableObjectRetention = true };
```
jskeet, line 123: So what happens if the code below is tried *without* object retention enabled? I think that would probably be worth another test.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -110,6 +110,64 @@ public void LockRetentionPolicy()
...
+                Retention = new Object.RetentionData
+                {
+                    Mode = "Unlocked",
+                    RetainUntilTimeDateTimeOffset = baseDateTime.AddDays(5),
+                }
+
+            };
+            CreateObject(newObject);
+            var obj = client.GetObject(bucketName, objectName);
+            Assert.Equal("Unlocked", obj.Retention?.Mode);
```
jskeet, line 146: I'd remove the `?` here - that way we can tell the difference between `obj.Retention` being null (NRE) and `obj.Retention.Mode` being null (assertion failure).

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
jskeet: I'd probably remove the method entirely, and just inline `_fixture.Client.UploadObject(obj, new MemoryStream(_fixture.SmallContent));` unless we really need the specific text.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -110,6 +110,64 @@ public void LockRetentionPolicy()
...
+            Assert.Throws<GoogleApiException>(() => client.DeleteObject(bucketName, objectName));
+
+            // Remove retention for existing object.
+            updatedObj.Retention = null;
+            var removeobjectRetentionOption = new UpdateObjectOptions { OverrideUnlockedRetention = true, };
+            client.UpdateObject(updatedObj, removeobjectRetentionOption);
+            var retentionRemovedObj = client.GetObject(bucketName, objectName);
+            Assert.Null(retentionRemovedObj.Retention);
+
+            client.DeleteObject(bucketName, objectName);
```
jskeet, line 168: I'd add a comment here to indicate that this is really part of the test, not just clean-up: once we've removed the lock, we should be able to delete the object.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/RetentionPolicyTest.cs
```diff
@@ -110,6 +110,64 @@ public void LockRetentionPolicy()
...
+            obj.Retention.RetainUntilTimeDateTimeOffset = baseDateTime.AddDays(2);
+            client.UpdateObject(obj, updateObjectOptions);
+            var updatedObj = client.GetObject(bucketName, objectName);
+            Assert.Equal(baseDateTime.AddDays(2).Date, updatedObj.Retention?.RetainUntilTimeDateTimeOffset.Value.Date);
+
+            Assert.Throws<GoogleApiException>(() => client.DeleteObject(bucketName, objectName));
+
+            // Remove retention for existing object.
+            updatedObj.Retention = null;
+            var removeobjectRetentionOption = new UpdateObjectOptions { OverrideUnlockedRetention = true, };
```
jskeet, line 163: Nit: remove trailing comma.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UpdateObjectOptions.cs
jskeet: +1 for this.

jskeet:
I'll fetch this locally and make the changes that Amanda and I had requested.

PR #11414, opened 2023-12-08
docs: Update getting-started.md with actionable auth guidance

piaxc:
Adding actionable guidance for setting up ADC and removing reference to explicit credentials, since this is a niche use case.

### DIFF
--- a/docs/devsite-help/getting-started.md
+++ b/docs/devsite-help/getting-started.md
@@ -68,24 +68,19 @@ dotnet add package Google.Cloud.Translate.V3
 > - Grpc.Auth, Grpc.Core.Api, Grpc.Net.Client, Grpc.Net.Common: support for the [gRPC](https://grpc.io/) RPC protocol
 > - Google.LongRunning: support for [long-running operations](long-running-operations.md)
 
-## Set up Application Default Credentials
-
-This step sets up Application Default Credentials to use your user credentials in a local development
-environment. 
-
-```sh
-gcloud auth application-default login
-```
-
-When you create a client, it automatically detects and uses these credentials.
-For other ways to set up Application Default Credentials, see
-[Set up Application Default Credentials](https://cloud.google.com/docs/authentication/provide-credentials-adc).
-
 ## Create a client
 
 The first step in making any API calls is to create a client. Some libraries have multiple clients
 for operations involving different resources; others have a single client. In the Translation API
 we're using, there's just `TranslationServiceClient`.
+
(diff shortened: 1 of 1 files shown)


jskeet:
Assigning Amanda to review as well, but personally I absolutely do *not* regard explicit credentials as a niche use case.

piaxc:
Fair enough! But for getting started, ADC is the way to go.

docs/devsite-help/getting-started.md
```diff
@@ -68,17 +68,25 @@ dotnet add package Google.Cloud.Translate.V3
...
 
 The first step in making any API calls is to create a client. Some libraries have multiple clients
 for operations involving different resources; others have a single client. In the Translation API
 we're using, there's just `TranslationServiceClient`.
 
-Clients can be configured in a number of ways, but in many cases the defaults are fine. The most
-common reason to use explicit configuration is to use specific credentials for
-[authentication](https://cloud.google.com/docs/authentication/use-cases). For this example, we'll just use the
-[application default credentials](https://cloud.google.com/docs/authentication/production#automatically).
-
```
amanda-tarafa, line 81: Similar to @jskeet I'd prefer if we rephased this section instead of removing it and even instead of adding a whole new section on setting up  ADC.

I think something along the lines of: "... For this example, we'll just use ADC (with links to ADC and general auth at Google). To set up ADC in your local development environment [link]".

Note that instead of including the gcloud command verbatim, I'd prefer a link to a document that does.

amanda-tarafa (changes requested):
Reasons for my changes request:

- Specific credentials being a niche use case does not remotely match our experience with .NET users. I agree that ADC is good getting-starting advice, and that's why we were using it in this example to begin with. But mentioning "specific credentials" is wholly proportional with .NET usage, even in a "getting started" doc.
- I prefer including links to the authoritave information containing current best practices and steps instead of duplicating those here. That's because best practices change often and then we have to play catch up.

docs/devsite-help/getting-started.md
amanda-tarafa: Aditionally, this section is not just talking about specific credentials. It's also mentioning that "clients can be configured in other ways but defaults are mostly fine", and that's relevant information to include here, regardles of ADC usage.

amanda-tarafa (approved):
Thanks, this looks good!

Please wait for @jskeet to confirm as well before merging.

jskeet (approved):
Much better, thanks!

PR #11457, opened 2024-01-02
chore: Regenerate OwlBot YAML files with a 2024 copyright.

jskeet:
(Will self-merge.)


jskeet:
Diff breakage is due to breaking change in Spanner. (We might want to update the version to the next major, beta00.)

PR #11461, opened 2024-01-02
fix: [Google.Cloud.PubSub.V1] Use message ordering enabled property that comes with streaming pull responses

kamalaboulhosn:
Use message ordering enabled property that comes with streaming pull responses so that messages are only delivered to the callback one at a time in order when ordering is actually enabled. This ensures there is no latency hit on processing messages when the subscription does not have ordering enabled.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
@@ -174,6 +174,7 @@ public sealed partial class SubscriberClientImpl
         private long _extendThrottleHigh = 0; // Incremented on extension, and put on extend queue items.
         private long _extendThrottleLow = 0; // Incremented after _extendQueueThrottleInterval, checked when throttling.
         private bool _exactlyOnceDeliveryEnabled = false; // True if subscription is exactly once, else false.
+        private bool _messageOrderingEnabled = false; // True if subscription has ordering enabled, else false.
         private TimeSpan? _pullBackoff = null;
 
         internal SingleChannel(SubscriberClientImpl subscriber,
@@ -397,6 +398,7 @@ public sealed partial class SubscriberClientImpl
                 {
                     current = _pull.GrpcCall.ResponseStream.Current;
                     _exactlyOnceDeliveryEnabled = current.SubscriptionProperties?.ExactlyOnceDeliveryEnabled ?? false;
+                    _messageOrderingEnabled = current.SubscriptionProperties?.MessageOrderingEnabled ?? false;
                 }
                 catch (Exception e) when (e.As<RpcException>()?.IsRecoverable() ?? false)
                 {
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -310,7 +313,8 @@ private void StartStreamingPull()
         // Backoff delay (if present) has already been done; no need to delay here.
         private void HandleStartStreamingPullWithoutBackoff()
         {
-            _pull = _client.StreamingPull(CallSettings.FromCancellationToken(_softStopCts.Token));
+            var callSettings = CallSettings.FromHeader(XGoogRequestParamsHeaderKey, string.Format("subscription={0}", _subscriptionName));
```
jskeet, line 316: This should use `Uri.EscapeDataString` on the subscription name - or we can wait until there's a new GAX release, hopefully soon, and then use `CallSettings FromGoogleRequestParamsHeader`. (We can refactor to that later easily enough.)

jskeet (commented):
I think this looks okay, but please could you rebase it to a single commit from the current HEAD? (We try to avoid all kinds of merge commits, and while we *can* squash the PR when it's merged, we prefer to merge it as-is if possible.)

I'm assuming that the existing behavior of `_flow.Process` is okay?

Have you already looked at potentially adding a unit test but decided against it, or is that something we should look into?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
kamalaboulhosn: Sorry, that was left over from a previous change that isn't meant to be here. It is now removed.

kamalaboulhosn:
@jskeet I have now rebased. The existing `_flow.Process` is okay, yes. The goal here is treat messages with ordering keys as messages without ordering keys in terms of delivery order if the subscription has message ordering disabled.

For unit tests, it turns out I had to make changes to ensure the ordered test still works, which I have now done. Testing that messages can come out of order when the feature is disabled is a bit more tricky. The only way I was able to do it effectively in the [Java client](https://github.com/googleapis/java-pubsub/pull/1851/files#diff-3f2adab9cf24e056119c51b74361d57d404eb463229c3f8a918ef3259ef49d75) was to put a sleep in and ensure multiple messages were processed simultaneously. There was already precedent for that kind of manipulation in other tests, so it wasn't too out of place, but I'm not sure it really fits as well here in these tests. At the end of the day, the most important thing to ensure is that ordering is preserved when the subscription has ordering enabled, which the existing tests should cover. If there is a good way to test that two callbacks for the same key can execute concurrently if they have the same ordering key, but message ordering is disabled on the subscription, I'm happy to do so, but it doesn't look like the tests are set up to make that kind of thing easy.

jskeet (approved):
Thanks for the changes. I'll probably reformat a couple of things later, but that's easier for me to do than describe :)

Now that we're out of a release freeze, we can release this at any time. Presumably it's a really safe change, so we don't need to release a beta beforehand or anything?

kamalaboulhosn:
@jskeet It should be a safe change and a release would be great! The change was prompted by a customer noting high latency when ordering keys were used even when ordering was disabled on the subscription.

PR #11467, opened 2024-01-03
chore: Reformatting of SubscriberClientTest

jskeet:
This basically just introduces some more line breaks to avoid lines getting quite so long. We could do significantly more should we wish to.


jskeet:
(Self-merging - no real code changes.)

PR #11470, opened 2024-01-03
chore: Batch API regeneration for copyright lines

jskeet:
This change will make OwlBot-authored changes smaller. In the future we may want to work out a way of avoiding updating copyright years at all.

There are a few APIs that are out of date or don't generate for one reason or another; I'll deal with those separately.


jskeet:
I'll merge this on green - I've validated locally that the only changes are copyright lines.

PR #11473, opened 2024-01-03
chore: Regenerate all APIs

jskeet:
(Getting down to very few APIs which generate differently locally from Bazel...)


amanda-tarafa (approved):
LGTM

Just in case it's unexpected, BigQuery.Logging.V1 is adding a new field, but there doesn't seem to be anything wrong with that.

jskeet:
> Just in case it's unexpected, BigQuery.Logging.V1 is adding a new field, but there doesn't seem to be anything wrong with that.

Yes, that was known - I need to track down why OwlBot hadn't previously created a change for it, but it's not a matter of the generator doing the wrong thing.

jskeet:
(I've now added a third commit to regenerate Google.Maps.Routing.V2 - I believe I had a local change to the protos when regenerating a while ago.)

PR #11479, opened 2024-01-04
tools: Streamline local generation

jskeet:
A surprising amount of time is taken getting information from the API catalog.

This change:
- Makes the query-api-catalog command bypass reflectively creating instances of all commands
- Avoids deserialization of the API catalog during query-api-catalog
- Avoids using "dotnet run" to invoke ReleaseManager

The whole thing is still surprisingly slow though - it's possible that a new option on query-api-catalog to output a Bash associative array might help, so we only need to call it once per API instead of 7 times. In the meantime, I believe this is a ~30% performance improvement overall, which I'll take.

### DIFF
--- a/generateapis.sh
+++ b/generateapis.sh
@@ -21,13 +21,20 @@ fi
 # Allow pre/post-generation scripts to know where to find the repos
 export GOOGLEAPIS
 
+# We call ReleaseManager frequently; dotnet run takes a while to start it,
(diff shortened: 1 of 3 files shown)


jskeet:
Time taken for full local regeneration with these changes: 59 minutes. I think it was about 90 minutes yesterday...

amanda-tarafa (approved):

PR #11486, opened 2024-01-05
chore: Use the new generator logging option

jskeet:
(We can't merge this change until we're using a generator that support it.)


amanda-tarafa (approved):
LGTM, so it'll be ready once we update the generator.

jskeet:
Rebasing, then we should be able to merge...

Issue #11522, opened 2024-01-11
Support the STORAGE_EMULATOR_HOST environment variable for StorageClient


EricStG:
**Is your feature request related to a problem? Please describe.**
We're using the firebase emulator, and we want to be able to configure the StorageClient via environment variables

 **Describe the solution you'd like**
We'd like to be able to configure the StorageClient's host via the STORAGE_EMULATOR_HOST environment variable

 **Describe alternatives you've considered**
We are currently setting the address via code, as described [here](https://github.com/hemanshv/dotnet-docs-samples/blob/main/storage/api/Storage.Samples.Tests/SetClientEndpointTest.cs).

 **Additional context**
This is already supported by the [GO](https://cloud.google.com/go/docs/reference/cloud.google.com/go/storage/latest) and [Node](https://github.com/googleapis/nodejs-storage/issues/2069) clients

jskeet:
Will have a look at this. One thing that the .NET libraries are consistent about is not *just* using an environment variable. We'd do this in the same way - see https://cloud.google.com/dotnet/docs/reference/help/emulators for examples.

(I wasn't aware that the Firebase emulator implemented a full Storage service... Will need to test it thoroughly.)

jskeet:
Hmmm... I've just had made a first attempt at this, and the emulator just rejects all requests. I need to look in more detail as to what it *will* accept.

jskeet:
Ah - my mistake, I'd used the emulator UI port instead of the API port. Doh. Looks like it's working now - with the limitations around no bucket operations, etc. Will see if I can get it more production-ready and create a PR.

EricStG:
We have it working here without any changes, at least for basic upload/download
Except for a workaround on UpdateObjectAsync, which fails (we haven't dug too much into it for our needs though)

jskeet:
Will look into UpdateObjectAsync as part of the work - thanks for calling that out.

EricStG:
We were having errors trying to update both `CacheControl` and `Metadata` with `UpdateObjectAsync`.
The error was:
```
The service storage has thrown an exception.
HttpStatusCode is NotImplemented.
No JSON error details were specified.
Raw error details are: Not Implemented
```
Could be a limitation on the emulator, but it would not be [documented](https://firebase.google.com/docs/emulator-suite/connect_storage#differences_from_google_cloud_storage).

jskeet:
[...]

PR #11535, opened 2024-01-16
 feat: Add support for the Firebase Storage emulator

jskeet:
Despite being large, much of this is copy-pasted from the gRPC emulator support (as this is the first REST-based emulator we've looked at). This is only a prototype at the moment, but it appears to work.
I'm looking for feedback about what I may be missing...

Towards #11522


amanda-tarafa (approved):
LGTM

jskeet:
Right, will have another pass through it myself and convert this into a PR to actually merge...

amanda-tarafa (approved):

PR #11559, opened 2024-01-19
chore: Stop sending the legacy routing header for Datastore and Firestore

amanda-tarafa:


jskeet (approved):
Slightly alarmed that there are no Datastore tests to delete, but hey.
Always nice when whole partial classes can be deleted...

amanda-tarafa:
> Slightly alarmed that there are no Datastore tests to delete, but hey. Always nice when whole partial classes can be deleted...

Yep, and whole test projects (in Firestore).

apis/Google.Cloud.Firestore.V1/Google.Cloud.Firestore.V1/FirestoreClientPartial.cs
```diff
@@ -37,115 +37,6 @@ public partial class FirestoreClientImpl
...
-        /// </summary>
-        /// <remarks>
-        /// This should be set to the database resource name ("projects/{projectId}/databases/{databaseId}") for any RPC.
-        /// For non-streaming calls, <see cref="FirestoreClientImpl"/> performs this automatically. This cannot be performed
-        /// automatically for streaming calls due to the separation between initializing the stream and sending requests, so
-        /// client code should set the value in a <see cref="CallSettings"/>. Typically this is performed with either the
-        /// <see cref="CallSettings.FromHeader(string, string)"/> factory method or the
-        /// <see cref="CallSettingsExtensions.WithHeader(CallSettings, string, string)"/> extension method.
-        /// </remarks>
-        public const string ResourcePrefixHeader = "google-cloud-resource-prefix";
```
jskeet, line 54: Right, we'll need to reinstate this because it's public. We can make it obsolete though. I'll need to check what we need to do in WatchStream though.

PR #11560, opened 2024-01-19
fix: Fix Firestore build, which still depends on ResourcePrefixHeader

jskeet:
This makes the constant obsolete, but continues to use it in WatchStream for now. When we've had the next GAX release and consulted with the Firestore team, we should remove its use from WatchStream (but keep it in FirestoreClientImpl until the next major release).


jskeet:
I'll self-merge this to fix the build (as it's really simple) but @amanda-tarafa should review when she's back online.

jskeet:
Diff is failing because the build was broken in the previous commit.

PR #11561, opened 2024-01-19
feat: All BatchGetDocuments RPCs to have customized retry settiings (per-FirestoreDb)

### LINKED ISSUE #11322: Better way for setting operation timeout and retries in Firestore SDK
Hello,

I'm looking for a convenient way of setting timeouts and optionally retries for basic Firestore operations.

As an example, I have a read document operation that in some _weird_ cases may be running for more than a minute:
```
firestoreDb.Collection(collectionName).Document(key).GetSnapshotAsync()
```

I am aware that there is a way to provide a cancellation token to some of methods but in my opinion, it is not a clean way to set timeout moreover using `CreateLinkedTokenSource` adds additional overhead and requires advanced technics (like pooling and disposing) to be used.

Naive example:
```
 var timeoutCts = new CancellationTokenSource();
 timeoutCts.CancelAfter(timeSpan);
 var combinedCts = CancellationTokenSource.CreateLinkedTokenSource(timeoutCts.Token, cancellationToken);
 return combinedCts.Token;
```

I can be wrong but I'm under impression that in .NET SDK timeouts and retries were hard-coded.
https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs#L78-L86

As a suggestion I would like to highlight Python SDK where setting timeouts and retries requires no effort.

https://cloud.google.com/python/docs/reference/firestore/latest/query
```
retry ([google.api_core.retry.Retry](https://googleapis.dev/python/google-api-core/latest/retry.html#google.api_core.retry.Retry)) – Designation of what errors, if any, should be retried. Defaults to a system-specified policy.
timeout ([float](https://python.readthedocs.io/en/latest/library/functions.html#float)) – The timeout for this request. Defaults to a system-specified value.
```

JFYI: There was an old issue https://github.com/googleapis/google-cloud-dotnet/issues/2471 but I think it is different.

jskeet:
fix: Use FirestoreSettings.BatchGetDocuments for batch timing

This will change the default timeout from 10 minutes to 5 minutes, but that will bring us in line with other clients. Any customer observing problems after this change can manually set the timeout in FirestoreSettings. I regard this as a "sufficiently safe" (but still *potentially* breaking) change to not merit a new major version.

Fixes #11322.

(While it would be nice to have per-RPC settings, that's probably too niche a requirement to be worth the API surface upheaval, at least for now.)

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/FirestoreDbTest.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/FirestoreDbTest.cs
@@ -337,6 +337,7 @@ namespace Google.Cloud.Firestore.Tests
             await Assert.ThrowsAsync<InvalidOperationException>(() => db.GetDocumentSnapshotsAsync(new[] { db.Document("col1/doc1") }, transactionId: null, fieldMask: null, default));
         }
 
+        // This tests the default retry settings.
         [Fact]
         public async Task GetDocumentSnapshotsAsync_Retry()
         {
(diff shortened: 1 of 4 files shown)


amanda-tarafa (approved):
LGTM

PR #11564, opened 2024-01-19
feat: Add configurable retry timing for RunTransactionAsync

### LINKED ISSUE #10513: Google.Cloud.Firestore: Backoff on FirestoreDb.RunTransactionAsync retry
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Regarding [RunTransactionAsync](https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs#L389) on the FirestoreDb class in the Google.Cloud.Firestore package.

When we call it with `MaxAttempts > 1` option, then we would like to have some sort of backoff policy, in our case for data contention, so that it does not spam the document when it is already in contention.
(note there is a `backoff` variable that is unused in the source code, not really relevant to the solution proposed below though)

Either a backoff policy similar to what we see in the `_batchGetCallSettings` field or being able to pass an optional backoff policy to the method / client.
There's the issue of backwards compatibility, so maybe it does need to configurable by the user, defaulting to current behavior (no backoff).

Currently we're implementing our own backoff policy, wrapping around the firestoredb, but since you already have a backoff concept, it seemed reasonable to add it here as well.

 **Additional context**
I admit that I'm a bit unsure if there's some internal backoff policy inside the lower internal functions that I haven't been able to divine, but from what I could see there's nothing obvious going on, so please correct me if there's anything I've misunderstood.

jskeet:
There are further changes to come in terms of transaction retry, but this is at least a start. Note that this changes the default backoff from "none at all" to "100ms initial, with a multiplier of 1.3". That's a more reasonable default, and it seems unlikely that customers would actually *depend* on there being no backoff.

Fixes #10513

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs
@@ -387,9 +387,12 @@ namespace Google.Cloud.Firestore
         public async Task<T> RunTransactionAsync<T>(Func<Transaction, Task<T>> callback, TransactionOptions options = null, CancellationToken cancellationToken = default)
         {
             ByteString previousTransactionId = null;
-            options = options ?? TransactionOptions.Default;
-            var attemptsLeft = options.MaxAttempts;
-            TimeSpan backoff = TimeSpan.FromSeconds(1);
+            options ??= TransactionOptions.Default;
+
+            var retrySettings = options.RetrySettings;
+            var attemptsLeft = retrySettings.MaxAttempts;
+            TimeSpan backoff = retrySettings.InitialBackoff;
+            var scheduler = Client.Settings.Scheduler ?? SystemScheduler.Instance;
 
             while (true)
             {
@@ -408,8 +411,12 @@ namespace Google.Cloud.Firestore
                     }
                     catch (RpcException e) when (CheckRetry(e, ref rollback))
                     {
-                        // On to the next iteration...
+                        // On to the next iteration after a backoff.
                     }
+
+                    // This is essentially the inner loop of RetryAttempt.CreateRetrySequence.
+                    await scheduler.Delay(retrySettings.BackoffJitter.GetDelay(backoff), cancellationToken).ConfigureAwait(false);
+                    backoff = retrySettings.NextBackoff(backoff);
                 }
                 finally
                 {
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/TransactionOptions.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/TransactionOptions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2017, Google Inc. All rights reserved.
+// Copyright 2017, Google Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -13,6 +13,8 @@
 // limitations under the License.
 
 using Google.Api.Gax;
+using Google.Api.Gax.Grpc;
+using System;
 
 namespace Google.Cloud.Firestore
 {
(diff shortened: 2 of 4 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs
```diff
@@ -410,8 +420,12 @@ public async Task<T> RunTransactionAsync<T>(Func<Transaction, Task<T>> callback,
                     }
                     catch (RpcException e) when (CheckRetry(e, ref rollback))
```
amanda-tarafa, line 421: I think there's a bug here that was there before.

If commit fails with aborted and there are attempts left, we never rollback that failed transaction right? Or is it OK not to rollback it because it already failed?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs
```diff
@@ -410,8 +420,12 @@ public async Task<T> RunTransactionAsync<T>(Func<Transaction, Task<T>> callback,
...
                     {
-                        // On to the next iteration...
+                        // On to the next iteration after a backoff.
                     }
+
+                    // This is essentially the inner loop of RetryAttempt.CreateRetrySequence.
+                    await scheduler.Delay(timing.BackoffJitter.GetDelay(backoff), cancellationToken).ConfigureAwait(false);
+                    backoff = timing.NextBackoff(backoff);
```
amanda-tarafa, line 428: I think this is clearer inside the catch.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/TransactionOptions.cs
```diff
@@ -24,27 +25,53 @@ public sealed class TransactionOptions
         /// <summary>
         /// The transaction options that are used if nothing is specified by the caller.
         /// </summary>
-        public static TransactionOptions Default { get; } = new TransactionOptions(5);
+        public static TransactionOptions Default { get; } = new TransactionOptions(5, null);
 
         /// <summary>
         /// The number of times the transaction will be attempted before failing.
         /// </summary>
         public int MaxAttempts { get; }
```
amanda-tarafa, line 33: Could we deprecate this one instead of ignoring the one in RetrySettings?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs
jskeet: I'm generally nervous of doing stuff in catch blocks. There are some oddities about how code executes there - not that I can remember details. Let's chat next week.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs
jskeet: I think it's okay, but we'll need to check. We definitely need to make changes here in general though - this PR is only attempting to affect the timing.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/TransactionOptions.cs
jskeet: Feels a bit like overkill if most people won't need to touch the timing - creating a retry setting is relatively complex. Happy to chat about it though.

jskeet:
Revamped to just use an underlying RetrySettings - PTAL.

amanda-tarafa (approved):
LGTM, thanks!

jskeet:
I've added tests, and this is now ready to merge if you're happy with it. (Again, it's not the final word on transaction retries - but it's at least an improvement in terms of timing.)

PR #11581, opened 2024-01-24
chore: Update GAX dependencies for *all* libraries.

amanda-tarafa:
So that we don't have to manually update GAX to get the newest common protos?

@jskeet  I've updated dependencies un apis.json and regenerated all projects (by running release manager's update-dependencies). I believe you only ran regenerate-projects which updates dependencies for alphas and betas only.
I understand you may have done that on purpose so let's chat about it. I figure we so rarely release a patch that we can assume that the next release of each library will be a minor.

Once this is in, #11580 and any others like it will be OK.


amanda-tarafa:
I've sent #11581 for this.

jskeet:
Yes, the possibility of a patch is the reason for pinning common versions - but I agree, it hardly ever happens. Let's do this for now, but consider removing the current logic that prevents defaulting for stable versions.

jskeet (approved):

amanda-tarafa:
> Yes, the possibility of a patch is the reason for pinning common versions - but I agree, it hardly ever happens. Let's do this for now, but consider removing the current logic that prevents defaulting for stable versions.

Yep, makes sense, I've created an internal issue for that. Merging now.

PR #11648, opened 2024-02-02
feat: Bigtable supports universe domain

amanda-tarafa:


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
```diff
@@ -212,15 +212,15 @@ public GcpCallInvoker CreateGcpCallInvoker()
         {
```
jskeet, line 212: Could you add docs for this method while you're here? (I'm not sure why the method is public, but I guess we now can't change that. I'll add it to the list of things we might want to change in the next major version.)

jskeet (approved):

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2/BigtableServiceApiClientPartial.cs
amanda-tarafa: Yes, here, and in the async version as well.

I don't remember noticing the missing docs, but I do remember wondering why these were public to beging with. But yes, making them internal or private on the next major version sounds like a good plan.

PR #11651, opened 2024-02-03
build: PubSub depends on latest GAX

amanda-tarafa:


amanda-tarafa:
@jskeet I'm bumping the GAX dependency on PubSub explicitly so that we can share the universe domain demo (after we've released a beta).

jskeet (approved):

PR #11654, opened 2024-02-06
feat: Spanner leader routing

amanda-tarafa:
@jskeet as always one commit at a time. Although most of these will be squashed in a single commit for merging.


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientPartial.cs
```diff
@@ -12,14 +12,21 @@
...
 using Google.Api.Gax.Grpc;
 using Google.Cloud.Spanner.Common.V1;
-using System;
 
 namespace Google.Cloud.Spanner.V1
 {
-    // Partial class to set up resource-based routing.
+    public partial class SpannerClient
+    {
+        /// <summary>
```
jskeet, line 24: This comment is weird, given that it's not a CallSettings, but a SpannerSettings. Not sure whether it *should* be a CallSettings, or a SpannerSettings with better docs. (I realize this commit is just moving it.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
```diff
@@ -2,25 +2,26 @@
...
+using Google.Api.Gax.Grpc;
 using System;
 using System.Threading;
 using System.Threading.Tasks;
 
 namespace Google.Cloud.Spanner.V1
 {
-    public partial class SpannerClientBuilder
-    {
+public partial class SpannerClientBuilder
```
jskeet, line 23: There are some weird changes to indentation in this file...

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
```diff
@@ -76,5 +87,17 @@ public SpannerClientBuilder MaybeCreateEmulatorClientBuilder()
             builder.CopySettingsForEmulator(this);
             return builder;
         }
+
+        internal new T GetEffectiveSettings<T>(T settings) where T : ServiceSettingsBase, new()
```
jskeet, line 91: Humbug - I guess we should have made the method virtual in GAX? Not that it's a problem, so long as we don't call it *from* GAX.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientPartial.cs
```diff
@@ -19,12 +19,31 @@ namespace Google.Cloud.Spanner.V1
...
+            {
+                MaybeApplyRouteToLeaderHeader(ref settings);
+            }
+        }
+
+        internal void MaybeApplyRouteToLeaderHeader(ref CallSettings settings)
+        {
+            if (Settings.LeaderRoutingEnabled)
+            {
+                settings = settings.WithHeader(LeaderRoutingHeader, true.ToString());
```
jskeet, line 44: I think it would be clearer to put the constant in here, rather than calling `true.ToString()`. (It will also return "True" - I'd expect "true" to be more idiomatic, although I'm not sure the value actually matters.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerSettingsPartial.cs
```diff
@@ -34,4 +35,14 @@ public Logger Logger
...
+
+    /// <summary>
+    /// Specifies whether leader routing is enabled or not. This is true by default.
+    /// </summary>
+    /// <remarks>
+    /// If this value is true some operations will always be explicitly routed to the leader,
+    /// some operations will never be explicitly routed to the leader, and some operations will
+    /// be routed to the leader depending on the transaction type they are using.
+    /// </remarks>
+    internal bool LeaderRoutingEnabled { get; set; } = true;
```
jskeet, line 47: I'm slightly surprised this is internal rather than public. Is there a reason for that?

jskeet (approved):
What's there looks fine, but I'm slightly surprised there isn't connection string support. Is the idea that this would be disabled so rarely that we don't need to make it easy? (I can't remember exactly how SpannerConnection construction works - is there a way of passing in arbitrary settings?)

amanda-tarafa (commented):
@jskeet all review comments addressed in new ordered commits prefixed "Address review". Plus a new commit at the end adding connection string support (I was waiting for confirmation this was wanted).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -436,6 +437,20 @@ public EmulatorDetection EmulatorDetection
             }
         }
 
+        /// <summary>
+        /// Options to control leader routing. This is true by default.
```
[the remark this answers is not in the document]
amanda-tarafa: Yes! thanks for catching this. I'll add a test as well.

amanda-tarafa (commented):
@jskeet  all previous commits squashed into one plus a new commit fixing the leader routing default in the connection string (I'll squash that one as well before mergin.) PTAL.

jskeet (approved):

PR #11752, opened 2024-02-16
feat: Check script permissions in a GitHub action

jskeet:
Fixes b/325560863

### DIFF
--- a/.github/workflows/check-script-permissions.yml
+++ b/.github/workflows/check-script-permissions.yml
@@ -1,0 +1,14 @@
+name: Check that all pre and post-generation scripts are executable
+
+on: [pull_request, push]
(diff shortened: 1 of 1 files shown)


jskeet:
I don't *think* I can get the workflow to run in a PR from my fork to this one - if you're happy that it looks broadly okay, I suggest we merge as-is, and then I create a new PR which just deliberately introduces a "bad" script... we can check that that breaks the build, and then close the PR.

jskeet:
Note that both this and "check the API catalog" are very small workflows. We could potentially introduce a small "extra checks" workflow containing both of them instead.

amanda-tarafa (approved):

PR #11755, opened 2024-02-16
Add a bad script - no execute permissions

jskeet:


amanda-tarafa:
This worked, we can no close it.

PR #11763, opened 2024-02-19
feat: Implement int64-based timestamps

### LINKED ISSUE #10132: Specify useInt64Timestamp in BigQuery

jskeet:
Fixes #10132 and b/323942760

As every, review each commit separately. (In particular, the first commit removes some legacy code that was only used by tests.)

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetQueryResultsOptions.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetQueryResultsOptions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2016 Google Inc. All Rights Reserved.
+// Copyright 2016 Google Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -56,7 +56,20 @@ namespace Google.Cloud.BigQuery.V2
         /// </summary>
         public string PageToken { get; set; }
 
-        internal void ModifyRequest(GetQueryResultsRequest request)
+        /// <summary>
+        /// Whether to use 64-bit integers for timestamps, instead of
+        /// floating point, in order to preserve precision. If not set, this is effectively true.
+        /// </summary>
+        public bool? UseInt64Timestamp { get; set; }
+
+        // This doesn't follow the usual pattern, as we want to modify the request even if options is null.
+        internal static void ModifyRequest(GetQueryResultsOptions options, GetQueryResultsRequest request)
+        {
+            options?.ModifyRequest(request);
+            request.FormatOptionsUseInt64Timestamp = options?.UseInt64Timestamp ?? true;
+        }
+
+        private void ModifyRequest(GetQueryResultsRequest request)
         {
             if (PageToken != null && StartIndex != null)
             {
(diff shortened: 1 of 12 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetQueryResultsOptions.cs
```diff
@@ -56,7 +56,20 @@ public sealed class GetQueryResultsOptions
         /// </summary>
         public string PageToken { get; set; }
 
-        internal void ModifyRequest(GetQueryResultsRequest request)
+        /// <summary>
+        /// Whether to use 64-bit integers for timestamps, instead of
+        /// floating point, in order to preserve precision. If not set, this is effectively true.
+        /// </summary>
+        public bool? UseInt64Timestamp { get; set; }
```
amanda-tarafa, line 63: I think I'd do this instead of moving away from the usual pattern.

```suggestion
        /// <summary>
        /// Whether to use 64-bit integers for timestamps, instead of
        /// floating point, in order to preserve precision. Defaults to true.
        /// </summary>
        public bool UseInt64Timestamp { get; set; } = true;
```

And then anywhere in our public surface that we can recieve these options, do `options ?? new GetQueryResultsOptions()` or `options ?? GetQueryResultsOptions.Default`.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/ListRowsOptions.cs
```diff
@@ -40,13 +40,26 @@ public sealed class ListRowsOptions
...
+        }
+
+        private void ModifyRequest(ListRequest request)
         {
             if (PageToken != null && StartIndex != null)
             {
                 throw new ArgumentException($"Cannot specify both {nameof(PageToken)} and {nameof(StartIndex)}");
             }
-
+            request.FormatOptionsUseInt64Timestamp = true;
```
amanda-tarafa, line 62: This is not done on the GetQueryResultsOptions, and iit's not needed here either I think.

amanda-tarafa (approved):
I think I'd do this differently but it's fine as is as well. There's just one tiny comment otherwise.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetQueryResultsOptions.cs
jskeet: The normal pattern for this project is to use a nullable type - I *think* we've done that for every option. It means that the options are composable. Not that we end up actually using that ability at the moment, but it's why we have it.

We could still use the "have a default set of options and move ModifyRequest back to being an instance method though", if you want.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/ListRowsOptions.cs
jskeet: Whoops, I think this is left over from an earlier version. Removing.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/GetQueryResultsOptions.cs
amanda-tarafa: Yep, it's eaither we brake one pattern or the other. But yes, composability would be harder to achieve if we make some options non-nullable. So yes, this is better in the long run.

amanda-tarafa (approved):
LGTM

jskeet:
Right - will rebase this and then merge; we can change how we *use* the options later as that's only internal.

PR #9243, opened 2022-10-21
feat: PgJsonb integration tests.

Rishabh-V:
The PR contains basic integration tests for PgJsonb. PgJsonb[] is yet not supported and hence PR doesn't have them.

PR #9085, opened 2022-09-05
feat: PgJsonb and its unit tests.

Rishabh-V:
PgJsonb and its unit tests.

Will raise a new PR for its integration tests along with the setup of infrastructure for PostgreSQL dialect.

PR #11777, opened 2024-02-20
feat(spanner): add PG.OID support

skuruppu:
The PG.OID type is an int64 value with a PG_OID type annotation, mostly used by pg_catalog tables. This PR adds support for representing PG.OID values in the C# client.

Tests are added for being able to query a PG.OID value.

User tables can't be created with PG.OID as a column type so no tests were added for DML or mutations. Since the currently available set of pg_catalog tables don't have PG.OID columns, we also don't test being able to read a PG.OID from a table.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
@@ -34,9 +34,11 @@ namespace Google.Cloud.Spanner.Data.Tests
             yield return new object[] { SpannerDbType.Numeric, DbType.VarNumeric, true };
             yield return new object[] { SpannerDbType.Unspecified, DbType.Object, true };
             yield return new object[] { SpannerDbType.String, DbType.String, true };
-            // There is no DbType that will map automatically to SpannerDbType.Json or SpannerDbType.PgJsonb.
+            // There is no DbType that will map automatically to SpannerDbType.Json, SpannerDbType.PgJsonb
+            // or SpannerDbType.PgOid.
             yield return new object[] { SpannerDbType.Json, DbType.String, false };
             yield return new object[] { SpannerDbType.PgJsonb, DbType.String, false };
+            yield return new object[] { SpannerDbType.PgOid, DbType.Int64, false };
         }
 
         [Theory]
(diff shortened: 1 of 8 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
```diff
@@ -31,6 +31,8 @@ public static IEnumerable<object[]> GetDbTypeConversions()
             yield return new object[] { SpannerDbType.Timestamp, DbType.DateTime, true };
             yield return new object[] { SpannerDbType.Float64, DbType.Double, true };
             yield return new object[] { SpannerDbType.Int64, DbType.Int64, true };
+            // There is no DbType that will map automatically to SpannerDbType.PgOid.
+            yield return new object[] { SpannerDbType.PgOid, DbType.Int64, false };
```
amanda-tarafa, line 35: nit: Can you move these to after line 41 and update the comment in line 39? I think it's clearer if we have all the types that are not mapped automatically together.

amanda-tarafa (approved):
Hi @skuruppu:
I've left a tiny comment, otherwise looks good.
Please wait till @jskeet can take a look as well in case I missed something.

jskeet (approved):
Please address the comment from Amanda, but otherwise this looks okay to me. All the changes you *have* made look fine - but I can't vouch for whether there are other places that have been missed. Can I assume you looked for "jsonb" as an example?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
skuruppu: That makes sense. All done.

skuruppu:
> Please address the comment from Amanda, but otherwise this looks okay to me. All the changes you _have_ made look fine - but I can't vouch for whether there are other places that have been missed. Can I assume you looked for "jsonb" as an example?

Yeah exactly. I found all the PG.JSONB PRs (#9085, #9243, #9492) and followed exactly what those did. I think PG.JSONB is the closest type to PG.OID.

What I skipped was adding the type to `apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AllTypesTableFixturePostgreSql.cs` since you shouldn't be able to create a user table with this thype.

skuruppu:
@amanda-tarafa and @jskeet  I'll let you merge this when you're ready.

PR #11810, opened 2024-02-28
chore: Compute all GAPIC generator options in C#

jskeet:
(This is only a single commit after #11809 has been merged.)

This reduces the total time on Windows by about 10 minutes. (It doesn't appear to have much impact on Linux, where generation runs *much* faster anyway...)


amanda-tarafa (approved):
On nice third commit!

PR #11879, opened 2024-02-28
Release Google.Cloud.DevTools.Common version 3.1.0

jskeet:
Changes in this release:

No API surface changes; just dependency updates.

### DIFF
--- a/apis/Google.Cloud.DevTools.Common/docs/history.md
+++ b/apis/Google.Cloud.DevTools.Common/docs/history.md
@@ -1,5 +1,9 @@
 # Version history
 
+## Version 3.1.0, released 2024-02-28
+
+No API surface changes; just dependency updates.
+
 ## Version 3.0.0, released 2022-06-08
 
 This is the first version of this package to depend on GAX v4.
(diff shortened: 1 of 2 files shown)


amanda-tarafa (approved):

amanda-tarafa:
@jskeet this one failed because although the version was changed in apis.json, it was not changed on the project file so we attempted to publish the same version that existed, i.e. we attempted to publish Google.Cloud.DevTools.Common version 3.0.0.

I didn't caught this on review.

jskeet:
Gosh, that's odd. I wonder what happened.

jskeet:
GenerateProjectsCommand *still* isn't updating the version here, when I run it manually. Investigating further...

jskeet:
Aha - it's because there are no *source files* directly in the production directory - only a subdirectory. We use the presence of `*.cs` to work out where to actually generate project files. This is a highly unusual project. It's simple enough to detect the presence of an existing csproj file as well - I'll do that. We can then release a patch version of Google.Cloud.DevTools.Common rather than trying to retrofit the current one.

jskeet:
Removed the label to keep the issue from coming back.

PR #12049, opened 2024-03-05
Release Google.Cloud.Translation.V2 version 3.3.0

jskeet:
Changes in this release:

No API surface changes; just dependency updates.

### DIFF
--- a/apis/apis.json
+++ b/apis/apis.json
@@ -5156,7 +5156,7 @@
       "id": "Google.Cloud.Translation.V2",
       "productName": "Google Cloud Translation",
       "productUrl": "https://cloud.google.com/translate/",
-      "version": "3.2.0",
+      "version": "3.3.0",
       "type": "rest",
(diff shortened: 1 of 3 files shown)


apis/apis.json
```diff
@@ -5156,7 +5156,7 @@
       "id": "Google.Cloud.Translation.V2",
       "productName": "Google Cloud Translation",
       "productUrl": "https://cloud.google.com/translate/",
-      "version": "3.2.0",
+      "version": "3.3.0",
       "type": "rest",
       "description": "Recommended Google client library to access the Translate v2 API. It wraps the Google.Apis.Translate.v2 client library, making common operations simpler in client code. The Translate API translates text from one language to another.",
       "dependencies": {
```
amanda-tarafa, line 5162: I think we should update the Discovery dependency to Google.Apis.Translate.v2 1.67.0.875

jskeet: Right, will do that and BigQuery in a separate PR, then do the release again.

Issue #12034, opened 2024-03-03
Way to specify additional headers to be sent along ResumableUpload.InitiateSessionAsync()


Et3rnos:
There is currently no way to pass extra headers to be sent alongside ResumableUpload.InitiateSessionAsync().

I'm currently creating a website that should allow users to upload big files directly to the Google Storage bucket, and I would like to specify the size of the file to be uploaded ahead of the time. Your documentation says this is possible by passing a X-Upload-Content-Length header, but after spending some time looking at this package's documentation and even its source, seems like this is not possible.

> From https://cloud.google.com/storage/docs/performing-resumable-uploads#json-api:
> 
> Optional headers that you can add to the request include [X-Upload-Content-Type](https://cloud.google.com/storage/docs/json_api/v1/parameters#xuploadcontenttype) and [X-Upload-Content-Length](https://cloud.google.com/storage/docs/json_api/v1/parameters#xuploadcontentlength).

Are there plans to make this possible? Is there any workaround? Thanks.

Et3rnos:
Looks like a possible workaround is:

```cs
uploader.Service.HttpClient.DefaultRequestHeaders.TryAddWithoutValidation("X-Upload-Content-Length", "1");
```

Would be nice to have a dedicated option for it though.

Et3rnos:
Hi @jskeet and thanks for the quick response. I'll explain why that header is so important for me and give you a bit more of information:

First of all, I'm using Google.Cloud.Storage.V1 (v4.8.0), as you pointed out. This was the .NET client library recommended by the docs page. If you have any other alternative that does what I'm about to explain I'm down for it though :)

I'm building a website that should allow users to upload huge files, but each user also has a quota he has to respect. I'm currently using resumable uploads to upload files, and the flow is similar to the following:

- user manifests his intent to upload a file by sending a request to one of my endpoints, I run something similar to the following on the server:

```cs
var uploader = storageClient.CreateObjectUploader(bucketName, key, request.ContentType, new MemoryStream([]), new UploadObjectOptions
{
      Origin = "<the site url>",
});
var session = await uploader.InitiateSessionAsync();
return Ok(new
{
    Url = session.AbsoluteUri,
});
```

- user gets the session url and uploads the file using a PUT request
- ... (the rest isn't really relevant)

[...]

PR #12054, opened 2024-03-05
feat: Allow a new upload session to be initiated as a single method call

jskeet:
This allows the content length to be (optionally) specified, which is then propagated via X-Upload-Content-Length.

This is currently just a prototype for comment - I'd want to add integration tests, some unit tests, and obviously XML docs.

See #12034 for background.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.UploadObject.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.UploadObject.cs
@@ -49,8 +49,8 @@ namespace Google.Cloud.Storage.V1
         /// <summary>
         /// Creates an instance which is capable of starting a resumable upload for an object.
         /// </summary>
-        /// <param name="destination">Object to create or update. Must not be null, and must have the name,
-        /// bucket and content type populated.</param>
+        /// <param name="destination">Object to create or update. Must not be null, and must have the name
+        /// and bucket populated.</param>
         /// <param name="source">The stream to read the data from. Must not be null.</param>
         /// <param name="options">Additional options for the upload. May be null, in which case appropriate
         /// defaults will be used.</param>
@@ -116,8 +116,8 @@ namespace Google.Cloud.Storage.V1
         /// <summary>
         /// Uploads the data for an object in storage synchronously, from a specified stream.
         /// </summary>
-        /// <param name="destination">Object to create or update. Must not be null, and must have the name,
-        /// bucket and content type populated.</param>
+        /// <param name="destination">Object to create or update. Must not be null, and must have the name
+        /// and bucket populated.</param>
         /// <param name="source">The stream to read the data from. Must not be null.</param>
         /// <param name="options">Additional options for the upload. May be null, in which case appropriate
         /// defaults will be used.</param>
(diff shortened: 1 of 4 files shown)


amanda-tarafa (commented):
I think it would be useful to include the UploadObjectOptions as a parameter to these new methods, and then include the length in the UploadObjectOptions. We'd use the options slightly diferent in these methods, as the length field would be used to create the source (empty) stream instead of to modify the uploader, but I think that's fine, options are still surfaced in a consistent manner. And we would have to fail if both a stream and a lenght are provided, which is also fine I think. It's ugly implementation wise though.

Also, we could update [this snippet to be simpler](https://github.com/googleapis/google-cloud-dotnet/blob/897e7116143ffc286287047d0986ee6870051ec4/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/StorageClientSnippets.cs#L349)

jskeet:
> I think it would be useful to include the UploadObjectOptions as a parameter to these new methods, and then include the length in the UploadObjectOptions. We'd use the options slightly diferent in these methods, as the length field would be used to create the source (empty) stream instead of to modify the uploader, but I think that's fine, options are still surfaced in a consistent manner. And we would have to fail if both a stream and a lenght are provided, which is also fine I think. It's ugly implementation wise though.
> 
> Also, we could update [this snippet to be simpler](https://github.com/googleapis/google-cloud-dotnet/blob/897e7116143ffc286287047d0986ee6870051ec4/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Snippets/StorageClientSnippets.cs#L349)

I'll have a look at how much of UploadObjectOptions would actually be relevant. I *thought* most of it wouldn't be, but I could be wrong. Will look tomorrow.

jskeet:
Right, having looked, I can see that UploadObjectOptions does indeed have a bunch of stuff in that would be relevant.

I think my personal preference would still be to have a separate method accepting the length and an UploadObjectOptions rather than putting the length into the options directly. I'll amend the PR with my preference for this, and we can then discuss it further. (I may have misunderstood your proposal.)

amanda-tarafa (commented):
Yep, this looks good.

jskeet:
Great - will do all the rest of the work tomorrow, thanks !

jskeet:
Okay, this is now ready for proper review.

It's currently only the async methods. I suspect we *could* add synchronous methods just by calling `.ResultWithUnwrappedExceptions()` - that's what we do in `SignedUrlResumableUpload`. (We may also want to provide the ability to specify a length in `SignedUrlResumableUpload` as well at some point.)

I'm happy to either add the sync methods in here, or wait until they're requested.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.UploadObject.cs
```diff
@@ -153,5 +153,51 @@ public abstract partial class StorageClient
...
+        /// Initiates an upload session, optionally specifying the length of the content to be uploaded.
+        /// The resulting URI can be used with <see cref="ResumableUpload.CreateFromUploadUri"/>.
+        /// </summary>
+        /// <param name="destination">Object to create or update. Must not be null, and must have the name
+        /// and bucket populated.</param>
+        /// <param name="contentLength">The length of the content to upload later. This may be null, in which
+        /// case any length of content may be uploaded. If the value is non-null, it must be strictly positive
+        /// (not zero), and the content uploaded later must be exactly this length.</param>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        /// <param name="options">Additional options for the upload. May be null, in which case appropriate
```
amanda-tarafa, line 167: super nit: swap these two? I thought Visual Studio would catch that.

amanda-tarafa (approved):
Tiny nit only.

I'm fine with adding the sync versions later if there's demand for it.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClient.UploadObject.cs
jskeet: Whoops, yes, will do.

PR #12170, opened 2024-03-20
Release Google.Cloud.Language.V1 version 3.6.0-alpha01

jskeet:
Changes in this release:

Experimental release targeting netstandard2.0 instead of netstandard2.1.


jskeet:
I'm fully expecting the "diff" action to fail. We need to change detect-pr-changes.sh.

jskeet:
The create-clients test is also a known breakage that we can address separately.

PR #12212, opened 2024-03-21
feat: Added TopicName and SubsciptionName resolution from container as an extension point.

maw136:
This PR provides functionality to resolve TopicName and SubscriptionName from the IServiceProvider. 
It is a better way to resolve such dependency because one can reuse Configuration validation through the IOptions<> registration.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/ServiceCollectionExtensionsTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/ServiceCollectionExtensionsTest.cs
@@ -1,4 +1,4 @@
-// Copyright 2023 Google LLC
+// Copyright 2024 Google LLC
 //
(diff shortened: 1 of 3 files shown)


maw136:
Should those useless whitespaces remain?

jskeet:
> Should those useless whitespaces remain?

Yes, for now. I'm happy to create a separate PR later for reformatting etc, but it helps to keep the change focused if it's *just* the code that actually needs to be changed.

maw136:
I've created the issue: https://github.com/googleapis/google-cloud-dotnet/issues/12213
And I am reverting to whitespace changes as we speak.
I have my editor set to delete unnecessary whitespace at save.
In my experience it is the best option.

maw136:
Actually the issue was the line ending option. For the future the .editorconfig could enforce LF option for C# source files.

maw136:
Thanks in advance for current and future support.

maw136:
Better approach https://github.com/googleapis/google-cloud-dotnet/pull/12212

Issue #12213, opened 2024-03-21
feat: Add TopicName and SubsciptionName resolution from container as an extension point.


maw136:
I've created a PR: https://github.com/googleapis/google-cloud-dotnet/pull/12212
that resolved the issue of not being able to resolve TopicName/SubscriptionName from IoC container.
One such use-case for this new feature is:
Imagine a registration of IOptions<T> where T is the configuration for said pub/sub, there are validation options enabled at startup and maybe other dependencies for the configuration.
Current resolution requires one to create additional IServiceProvider instance inside Action<TBuilder>, alas there is perfectly good instance accessible inside virtual base.Configuration method of either Builder classes.

amanda-tarafa:
I believe the best approach is to add overloads in [ServiceCollectionExtensions](https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/ServiceCollectionExtensions.cs) as follows:

`public static IServiceCollection AddSubscriberClient(this IServiceCollection services, Action<IServiceProvider, SubscriberClientBuilder> action)`

`public static IServiceCollection AddPublisherClient(this IServiceCollection services, Action<IServiceProvider,PublisherClientBuilder> action)`

This will allow you to resolve from DI not just the topic name but any other aspect, without us needing to add new "resolver" properties for each aspect in the builder that anyone wants to resolve from DI. Let me know if I'm missing something and this does not fully resolve your use case. Also @jskeet for thoughts.

maw136:
Yes yours is more general solution and the better approach. 
I considered it and opted for the one in PR because of less intrusiveness, and possible breaking changes.
I am fully on board for either approach.

amanda-tarafa:
I don't think there'd be breaking changes with my proposal, we'd be adding a couple of overloads? Can you clarify that? Also, what do you mean with intrusiveness? Again, I'm not certain if I'm missing something here.

I do strongly prefer my approach over yours, so if there's nothing else, I think we should go with that.

maw136:
Just being pragmatic. I had to adjust existing unit tests to handle your approach. This is the only thing that got me thinking.
I prefer your approach myself too TBH.
Here is the implementation: https://github.com/googleapis/google-cloud-dotnet/pull/12215

amanda-tarafa:
[...]

PR #12215, opened 2024-03-21
feat: Added TopicName and SubsciptionName resolution from container as an extension point.

maw136:


maw136:
https://github.com/googleapis/google-cloud-dotnet/issues/12213 better aproach

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/ServiceCollectionExtensions.cs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 Google LLC
+// Copyright 2024 Google LLC
```
jskeet, line 1: Please revert this change - we leave copyright years from when the file was first introduced, where possible. (It's harder for generated code, which is generated from scratch each time.)

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/ServiceCollectionExtensions.cs
```diff
@@ -35,12 +35,29 @@ public static partial class ServiceCollectionExtensions
...
+    /// </summary>
+    /// <param name="services">
+    /// The <see cref="IServiceCollection"/> to add the singleton client to.
+    /// </param>
+    /// <param name="action">
+    /// An action delegate to invoke on the <see cref="PublisherClientBuilder"/> for configuring the <see cref="PublisherClient"/>. This is invoked before <paramref name="services"/> are used.
+    /// Must not be null and at-least <see cref="PublisherClientBuilder.TopicName"/> must be set.
+    /// </param>
+    /// <returns>The updated <see cref="IServiceCollection"/>, for method chaining.</returns>
+    public static IServiceCollection AddPublisherClient(this IServiceCollection services, Action<IServiceProvider, PublisherClientBuilder> action)
```
jskeet, line 54: This is *strictly speaking* a breaking change due to an existing call becoming ambiguous, in the following scenarios:

- Using a null literal - I'm not bothered about that, as we're validating that the action is non-null anyway (so it would only break something at compile-time that would already be broken at execution time)
- Using a method group conversion where there are two overloads: I think this is really, really unlikely, and the benefits of this outweigh the risk

jskeet (changes requested):
I *think* this will be okay - I'll discuss the breaking change aspect with Amanda.
Separately, I'll fix up whitespace that was spotted in the other PR.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/ServiceCollectionExtensions.cs
maw136: Will do. I haven't changed it manually. There may be some build time script or task that updates it.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/ServiceCollectionExtensions.cs
maw136: Ah, the sweet harmony of shared concerns. If you're nodding in agreement, I'm all in.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/ServiceCollectionExtensions.cs
maw136: Yes, I have some minimal reformat rules on save enabled. .editorconfig is probably one of them rules sources.

PR #12227, opened 2024-03-25
docs: Add example of DI using the provider for configuration

jskeet:
(Let me know if I've missed the point, or if it can be better expressed in a sample.)


amanda-tarafa (approved):
Yes, this looks right.

PR #12231, opened 2024-03-25
Release Google.LongRunning version 3.2.0

jskeet:
Changes in this release:

### New features

This library now targets netstandard2.0 instead of netstandard2.1. This should be compatible with existing libraries that depend on it, but will allow new libraries to also target netstandard2.0.


amanda-tarafa (approved):

jskeet:
Diff failure was in tooling, not in the actual library.

PR #12240, opened 2024-03-26
chore: Update libraries to target netstandard2.0 where feasible

jskeet:
As ever, review one commit at a time.

In this case, it's worth reviewing the last two commits even though they're autogenerated. The "feat" one should only contain prod projects where the target has changed (as well as some test projects). The "chore" commit should *not* contain any target changes.

Once this PR is merged, I'll run a batch release which should pick up all the libraries which have changed targets - we'll need to manually move to a new minor for previously-patch-released libraries.

After that, we can change or remove the temporary hack, as there should be very few libraries left which *can't* target netstandard2.0. (Mostly those targeting Logging.V2 I suspect, as that's a particularly long dependency chain due to Logging.Type.)


amanda-tarafa (approved):
LGTM.

Fine for Spanner to target both, also, depending on what has changed, we might take advantage of the upcoming major version to just target netstandard2.0.

PR #12241, opened 2024-03-26
docs: Provide updated details on target frameworks

jskeet:
I would suggest we merge this as soon as we're happy with it, but only publish the results when we're at least "nearly there" with releases.


amanda-tarafa (commented):
Couple of suggestions and one change. Otherwise LGTM

amanda-tarafa (approved):

PR #12557, opened 2024-04-10
feat!: Add support for FLOAT32

amanda-tarafa:
BREAKING CHANGE: The default mapping for values of CLR type float was FLOAT64 and it is now FLOAT32.


amanda-tarafa:
@jskeet I'm still running integration tests locally, it'll be a while.
It's also not clear to me if the emulator supports FLOAT32, so that check may fail here. In that case I'll skip FLOAT32 tests when running on the emulator.
Otherwise, this is ready for review.

amanda-tarafa:
OK, so FLOAT32 is not supported on the emulator. I'll skip those tests tomorrow.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerParameterTests.cs
```diff
@@ -34,8 +34,9 @@ public static IEnumerable<object[]> GetDbTypeConversions()
             yield return new object[] { SpannerDbType.Numeric, DbType.VarNumeric, true };
             yield return new object[] { SpannerDbType.Unspecified, DbType.Object, true };
             yield return new object[] { SpannerDbType.String, DbType.String, true };
-            // There is no DbType that will map automatically to SpannerDbType.Json, SpannerDbType.PgJsonb
-            // or SpannerDbType.PgOid.
+            // There is no DbType that will map automatically to SpannerDbType.Float32,
```
jskeet, line 37: What about DbType.Single? (I may well be missing something.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
```diff
@@ -338,7 +347,11 @@ public static SpannerDbType FromClrType(System.Type type)
             {
                 return Timestamp;
             }
-            if (type == typeof(float) || type == typeof(double) || type == typeof(decimal))
+            if (type == typeof(float))
+            {
+                return Float32;
+            }
+            if (type == typeof(double) || type == typeof(decimal))
```
jskeet, line 354: If we're making breaking changes anyway, maybe we should be using SpannerNumeric for decimal? (In a separate PR no doubt, but let's discuss)

jskeet (approved):
One concern around DbType, and a point for us to discuss about the default mapping for decimal, but otherwise fine. Approving in anticipation of the DbType bit being simple to fix, and the decimal mapping being non-blocking.

jskeet:
Filed b/333670648 for the decimal discussion.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
```diff
@@ -272,5 +274,9 @@ private async Task TestBindNonNull<T>(SpannerDbType parameterType, T value, Func
         public async Task BindJsonEmptyArray() => await TestBindNonNull(
             SpannerDbType.ArrayOf(SpannerDbType.Json),
             new string[] { });
+
+        internal static void MaybeSkipIfOnEmulator(bool runningOnEmulator, SpannerDbType spannerDbType) =>
+            Skip.If(runningOnEmulator && BindUnsupportedNullData.Contains([spannerDbType]),
```
jskeet, line 279: Is `[spannerDbType]` a collection expression here, and if so, are we able to use those at the moment? (I thought we'd have to wait until we're using the .NET 8 SDK.) I may just be missing something though.

jskeet (commented):
Just seen the GitHub actions failures - I think they're due to the collection expression. (I'm looking forward to when we *can* use those...)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerDbType.cs
amanda-tarafa: Yes, we've agreed to do this, I'll send a PR later.

amanda-tarafa (commented):
@jskeet All comments addressed. All but the first commit are new. Thanks!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/BindingTests.cs
amanda-tarafa: Yes, I just happily accepted Visual Studio's suggestion for this.

amanda-tarafa:
These emulator failures are just because some FLOAT32 tests that I didn't skip. Doing that now.

amanda-tarafa:
All green on the emulator now.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
```diff
@@ -251,5 +219,30 @@ private void ValidateAndParseClrToSpannerTypeMappings(string input, SpannerConve
                 }
             }
         }
+
+        private static void CheckMappingsFormat(string[] mappings, string mappingType)
```
jskeet, line 239: It took me a while to work out the purpose of this - and it looks like we always use this just before CheckValidMappings. Maybe combine the two, and document it as "validates that 'mappings' contains at least one mapping, and all mappings are valid"? (Or I may have missed something...)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConversionOptionsTests.cs
```diff
@@ -67,35 +67,41 @@ public void WithSpannerToClrMappings()
...
+            builder.ClrToSpannerTypeDefaultMappings = "singletofloat32,decimaltofloat64,datetimeTodate";
+            Assert.Equal(SpannerDbType.Float32, builder.ConversionOptions.SingleToConfiguredSpannerType);
             Assert.Equal(SpannerDbType.Float64, builder.ConversionOptions.DecimalToConfiguredSpannerType);
             Assert.Equal(SpannerDbType.Date, builder.ConversionOptions.DateTimeToConfiguredSpannerType);
         }
 
         [Theory]
         [InlineData("DecimalToFloat64,DecimalToNumeric,DateTimeToDate")] // Multiple mappings for a type.
         [InlineData("DecimalToFloat64,DecimalToPgNumeric,DateTimeToDate,DateTimeToTimestamp")] // Multiple mappings for a type.
+        [InlineData("SingleToFloat64,SingleToFloat32,DateTimeToDate,DateTimeToTimestamp")] // Multiple mappings for a type.
```
jskeet, line 104: I think all of these tests should be changed so that there's only *one* bad mapping involved... otherwise the fact that we've got DateTimeToDate and DateTimeToTimestamp together could be enough to cause the problem, even without us having the check for Single multiple mappings, if you see what I mean.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConversionOptionsTests.cs
```diff
@@ -67,35 +67,41 @@ public void WithSpannerToClrMappings()
         public void WithClrToSpannerMappings()
         {
             var options = SpannerConversionOptions.Default;
-            options = options.WithClrToSpannerMappings("DecimalToNumeric,DateTimeToDate");
+            options = options.WithClrToSpannerMappings("SingleToFloat32,DecimalToNumeric,DateTimeToDate");
```
jskeet, line 70: (The line isn't relevant here.)

Is there already a test for the defaults somewhere? It feels like a test that just uses SpannerConversionOptions.Default and asserts all the defaults would be useful. It's probably covered by the other test classes, but it feels like it would be useful documentation.

jskeet (approved):
A few refactoring/test suggestions which can be done later, but I *think* this is all good to go... it's possible I missed something from earlier reviews, but I'm pretty sure you'll have got all those covered.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConversionOptions.cs
amanda-tarafa: Yes, done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConversionOptionsTests.cs
amanda-tarafa: I've added a test for defaults, nice catch, there was none explicitly.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConversionOptionsTests.cs
amanda-tarafa: Yep, you are right, done.

amanda-tarafa (commented):
Comments addressed in ordered `review` commits. Thanks!

PR #12559, opened 2024-04-10
tools: By default, pull the latest version from googleapis when adding a new API

jskeet:
Fixes b/333667199

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
@@ -14,6 +14,7 @@
 
 using Google.Cloud.Tools.Common;
 using Google.Protobuf;
+using LibGit2Sharp;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Converters;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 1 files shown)


tools/Google.Cloud.Tools.ReleaseManager/AddCommand.cs
```diff
@@ -46,6 +54,22 @@ protected override int ExecuteImpl(string[] args)
...
+            {
+                Console.WriteLine($"Pulling googleapis directory");
+                using var repo = new Repository(googleapis);
+                string oldSha = repo.Head.Tip.Sha;
+                var merger = new Signature("Ignored", "ignored@google.com", DateTimeOffset.UtcNow);
+                var options = new PullOptions
+                {
+                    MergeOptions = new MergeOptions { FastForwardStrategy = FastForwardStrategy.FastForwardOnly, FailOnConflict = true }
+                };
+                Commands.Pull(repo, merger, options: null);
```
maw-wmc, line 68: Created `PullOptions` not used in the `Pull` call.

jskeet: Doh, thanks. Interesting that it works anyway, but I didn't validate that it fails if there are conflicts etc. Will use the options.

amanda-tarafa (approved):

PR #12561, opened 2024-04-10
feat: Added keyed service registration and some tests

maw136:


jskeet:
I really don't think it's worth creating a PR for this right now. When we can update to .NET, we'll be doing it across many repos, and it'll just be confusing to have PRs which do it in specific repos, possibly not in quite the same way.

.github/workflows/build-pr.yml
```diff
@@ -11,17 +11,17 @@ jobs:
...
         submodules: true
         fetch-depth: 100
-        
-    - name: Setup .NET 6.0
+
+    - name: Setup .NET 8.0
       uses: actions/setup-dotnet@v4
       with:
-        dotnet-version: 6.0.x
+        dotnet-version: 8.0.x
```
amanda-tarafa, line 24: This is not the reason why we can't build our libraries for .NET 8. We know how to do this.

The blocker at the moment is with internal infrastructure that we depend on for CI and other automation. We cannot move to .NET 8 at the moment. We are actively working on it and we expect to be able to do so sometime in Q2/Q3 this year.

Note (and for other readers) that you can use all libraries in applications targetting .NET 8. But we cannot yet build ourselves againts .NET 8, so we cannot have dependencies on .NET 8 specific features.

Once we are able to build against .NET 8, for this feature to be possible, we'd need to decide whether we explicitly target .NET 8 in libraries, as those are two different things (building against and targetting), that each carry maintenance burden and user facing commitments.

For all these reasons we cannot accept this PR at the moment. I'm sorry about that. I'll be closing this PR soon after I submit this comment.

PR #12564, opened 2024-04-11
docs: Add note around Google.Apis.Logging beeing hidden from Intellisense

amanda-tarafa:
Because we are releasing googleapis/google-api-dotnet-client#2707


jskeet (approved):
Thanks - I'll add something in the cloud.google.com docs too.

amanda-tarafa:
> Thanks - I'll add something in the cloud.google.com docs too.

This is the sample that is shown there, and the text really doesn't mention anything about ILogger at the moment. It's fine either way though.

PR #12565, opened 2024-04-11
tools: Reimplement API generation in C# (instead of bash)

jskeet:
This still delegates to protoc for most of the actual generation, but doing it all within C#:

- Makes it much faster on a Google Windows laptop
- Makes it easier to tweak logic later
- Means we can remove the get-generator-options command

Starting processes is generally messier in C# than in a bash script, but we don't really have *that* much to do.

This change also implements a few optimizations:

- We generate directly to the target directory, instead of into a temporary directory. We need to avoid "losing" the previous project files, but that's easy enough to do.
- We don't generate test files any more, so we don't need to clean out generated files.

The change doesn't make much difference to regenerate-all timing on Linux (about 18 minutes down to 15) but on Windows it reduces the time from over an hour to about 16 minutes.


jskeet:
Note: this removes the --check-compatibility

PR #12570, opened 2024-04-12
chore: Update test dependencies

jskeet:


jskeet:
It looks like we need to update at least one Spanner test for this change.

amanda-tarafa (approved):

PR #12684, opened 2024-05-02
fix: Retry more exceptions during transactions

### LINKED ISSUE #12669: Transaction failing due to error "The service is temporarily unavailable. Please retry with exponential backoff."
#### Environment details

  - OS: Hosting on a Cloud Run
  - .NET version: 6.0
  - Package name and version: Google.Cloud.Firestore from Google LLC 3.4.0

#### Steps to reproduce

Sadly no steps.

#### Description

A couple of weeks ago, I have started to receive errors with Firestore transactions with the following message:

`
The referenced transaction has expired or is no longer valid. (at Google.Api.Gax.Grpc.ApiCallRetryExtensions.b__0>d.MoveNext()\n--- End of stack trace from previous location ---\n at Google.Cloud.Firestore.WriteBatch.CommitAsync(ByteString transactionId, CancellationToken cancellationToken)\n at Google.Cloud.Firestore.FirestoreDb.RunTransactionAsync[T](Func 2 callback, TransactionOptions options, CancellationToken cancellationToken)\n at Google.Cloud.Firestore.FirestoreDb.RunTransactionAsync[T](Func 2 callback, TransactionOptions options, CancellationToken cancellationToken))
`

I don't think there is a problem in the code, since some transactions that fail are pretty simple and have not been modified in a long time. I also have seen that these failures occur during small periods of time (see times below), which makes me think that it is something related to the moment of the request/the availability of the service.

2024-04-20 05:21:24.737 CEST
2024-04-20 06:18:13.679 CEST
2024-04-20 16:01:29.846 CEST
2024-04-20 16:01:43.408 CEST
2024-04-20 16:01:57.471 CEST
2024-04-20 16:02:31.865 CEST
2024-04-20 16:02:57.235 CEST

We have activated the Audit logs and have seen that transactions are failing due to this error:
  The service is temporarily unavailable. Please retry with exponential backoff.
After that, the transaction retries but fails due to the previous error:
  The referenced transaction has expired or is no longer valid.

The first error occurs very often, what could be causing this issue? Could it be something related to the availability of Firestore, if so, is there any way to upgrade the Firestore to sustain more movement?

Errors "The referenced transaction has expired or is no longer valid." catched by the server:
image
Errors "The service is temporarily unavailable. Please retry with exponential backoff." catched by the Audit logs:
image

Stacktrace from today (2024-04-30 05:02:49.338 CEST):
`
"Message": "Status(StatusCode=\"InvalidArgument\", Detail=\"The referenced transaction has expired or is no longer valid.\")",
    "Data": {},
    "InnerException": null,
    "HelpLink": null,
    "Source": "System.Private.CoreLib",
    "HResult": -2146233088,
    "StackTrace": "   at Google.Api.Gax.Grpc.ApiCallRetryExtensions.<>c__DisplayClass0_0 2.<<WithRetry>b__0>d.MoveNext()\n--- End of stack trace from previous location ---\n   at Google.Cloud.Firestore.WriteBatch.CommitAsync(ByteString transactionId, CancellationToken cancellationToken)\n   at Google.Cloud.Firestore.FirestoreDb.RunTransactionAsync[T](Func 2 callback, TransactionOptions options, CancellationToken cancellationToken)\n   at Google.Cloud.Firestore.FirestoreDb.RunTransactionAsync[T](Func 2 callback, TransactionOptions options, CancellationToken cancellationToken)"
`

On another topic, we have experienced a ton of errors, a total of 1180, on 27-04-2024 between 7:00 am and 8:00 am CEST. The error was "Error connecting to subchannel." and it happened when trying to read a document. I know it is not the topic of the issue, but I am not sure if it is related to the problem. It looks like a Firestore was failing on that time but I don't know where I can check the status of Firestore, since there is no Firestore on the status page (https://status.firebase.google.com/).

Thank you in advance.

jskeet:
Almost all of these also cause a roll-back, but if the RPC failure is an InvalidArgument, we only retry if the message contains "transaction has expired", and in that case we don't roll back.

This brings the .NET implementation *closer* in line with other languages, but more work is required in the long term in terms of retrying BeginTransaction, aligning the need for rollback, and handling rollback failures.

Fixes #12669.

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs
@@ -431,6 +431,11 @@ namespace Google.Cloud.Firestore
             {
                 switch (e.Status.StatusCode)
                 {
+                    case StatusCode.Unknown:
+                    case StatusCode.Internal:
+                    case StatusCode.Unavailable:
+                    case StatusCode.Unauthenticated:
+                    case StatusCode.ResourceExhausted:
                     case StatusCode.Aborted:
                         // Definitely rollback, retry if we have any attempts left.
                         rollback = true;
@@ -440,6 +445,10 @@ namespace Google.Cloud.Firestore
                         // No retry, but we do want to roll back.
                         rollback = true;
                         return false;
+                    // If the transaction has expired, don't roll back, but do retry.
+                    case StatusCode.InvalidArgument when e.Message?.IndexOf("transaction has expired", StringComparison.Ordinal) >= 0:
+                        rollback = false;
+                        return true;
                     default:
                         return false;
                 }
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/FirestoreDbTest.RunTransactionAsync.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Tests/FirestoreDbTest.RunTransactionAsync.cs
@@ -58,10 +58,16 @@ namespace Google.Cloud.Firestore.Tests
         }
 
         // In-depth testing is only performed for the main work method
-        [Fact]
-        public async Task RunTransactionAsync_RollbackAndRetry()
+        [Theory]
+        [InlineData(StatusCode.Aborted)]
+        [InlineData(StatusCode.Unknown)]
+        [InlineData(StatusCode.Internal)]
+        [InlineData(StatusCode.Unavailable)]
+        [InlineData(StatusCode.Unauthenticated)]
+        [InlineData(StatusCode.ResourceExhausted)]
+        public async Task RunTransactionAsync_RollbackAndRetry(StatusCode code)
         {
-            var client = new TransactionTestingClient(2, CreateRpcException(StatusCode.Aborted));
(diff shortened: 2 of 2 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore/FirestoreDb.cs
```diff
@@ -440,6 +445,10 @@ bool CheckRetry(RpcException e, ref bool rollback)
                         // No retry, but we do want to roll back.
                         rollback = true;
                         return false;
+                    // If the transaction has expired, don't roll back, but do retry.
+                    case StatusCode.InvalidArgument when e.Message?.IndexOf("transaction has expired", StringComparison.Ordinal) >= 0:
```
amanda-tarafa, line 449: Hope we could add a test upstream to make certain this text doesn't change in the future.

jskeet: Ideally we can add a machine-readable error description in the trailers :)
(But changing the message is certainly regarded as a breaking change.)

jskeet:
Thanks @tom-andersen - yes, we definitely need to keep bringing things into closer alignment.

PR #12961, opened 2024-05-15
tests: Add full repro for issue 11793

jskeet:


philvmx1:
LGTM, this is exactly the configuration we've seen that causes the issue to manifest. Thank you!

NOTE: needs both Ordering and ExactlyOnce.

amanda-tarafa (approved):

PR #12974, opened 2024-05-17
tools: Add show-pending-releases command

jskeet:
Note that the "pending" here means "tagged but not released" - so the "autorelease: tagged" label, not "autorelease: pending".


amanda-tarafa (approved):
This is great!

PR #12995, opened 2024-05-23
feat: Improve Pub/Sub streaming pull retry handling

jskeet:
Just a draft at the moment, for a check on whether this is in the right direction before I add tests. (That'll be painful.)

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
@@ -23,6 +23,7 @@ using Grpc.Core;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
@@ -1784,7 +1785,7 @@ namespace Google.Cloud.PubSub.V1.Tests
         }
 
         [Fact]
-        public void StreamingPullRetry_InternalErrorRetriesForever()
+        public void StreamingPullRetry_InternalErrorContinuesRetrying()
         {
             // A regular internal failure that's not due to an auth error.
             var exception = new RpcException(new Status(StatusCode.Internal, "Bang"));
@@ -1798,6 +1799,24 @@ namespace Google.Cloud.PubSub.V1.Tests
                 Assert.False(subscriberTask.IsCompleted);
                 await fake.TaskHelper.ConfigureAwait(fake.Subscriber.StopAsync(CancellationToken.None));
                 await subscriberTask;
+            });
+        }
+
+        [Fact]
+        public void StreamingPullRetry_RetriableErrorEventuallyFails()
+        {
+            // A regular internal failure that's not due to an auth error.
+            var exception = new RpcException(new Status(StatusCode.Internal, "Bang"));
+
+            // When we've reached a limit of the number of exceptions we're happy to retry, we'll eventually fail.
+            // (This will take a long time, with all the backoffs involved...)
+            using var fake = Fake.Create(CreateBadMoveNextSequence(TimeSpan.FromSeconds(1), exception, 100, includeTrailing: true));
+
+            fake.Scheduler.Run(async () =>
+            {
+                var subscriberTask = fake.Subscriber.StartAsync((msg, ct) => throw new Exception("No messages should be provided"));
+                var subscriberEx = await Assert.ThrowsAsync<RpcException>(() => subscriberTask);
+                Assert.Equal(exception.Status, subscriberEx.Status);
             });
         }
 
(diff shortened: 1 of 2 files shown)


amanda-tarafa (commented):
Yep, this looks good.

jskeet:
Great - I'll see how feasible it is to test it tomorrow :)
(I'll almost certainly need an internal property in SubscriberSettings to have the retry timing, just to disable jitter, but that's okay.)

jskeet:
Okay, I *think* this is now ready, including tests. The first three commits are really clean-up of existing issues (or refactoring). Then there's the production code, then the tests.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -451,8 +450,8 @@ internal TestSubscriptionHandler(bool ackOrNack)
                 Responses = new List<AckNackResponse>();
             }
 
-            public override async Task<SubscriberClient.Reply> HandleMessage(PubsubMessage message, CancellationToken cancellationToken) =>
-                await Task.FromResult(_ackOrNack ? SubscriberClient.Reply.Ack : SubscriberClient.Reply.Nack);
+            public override Task<SubscriberClient.Reply> HandleMessage(PubsubMessage message, CancellationToken cancellationToken) =>
+                Task.FromResult(_ackOrNack ? SubscriberClient.Reply.Ack : SubscriberClient.Reply.Nack);
```
amanda-tarafa, line 467: tiny nit: but `_ackOrNack` => `_ack`?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -1109,61 +1109,121 @@ private void UpdateReceiptModAckStatus(IEnumerable<string> ids, bool? status)
...
+                _firstExceptionTimestamp ??= now;
+                // This won't be used if we decide not to retry anyway.
+                Backoff = _backoffTiming.NextBackoff(Backoff ?? TimeSpan.Zero);
+
+                // Don't retry any non-RpcExceptions
+                if (exception.As<RpcException>() is not RpcException rpcEx)
+                {
+                    return false;
+                }
+                if (_exceptions.Count < ExceptionLimit)
```
amanda-tarafa, line 1163: I don't see `_exceptions.Count` being used anywhere else? I'm probably misising something but it seems like after the limit, we continue retrying as usual except that we don't store the exceptions?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -1109,61 +1109,121 @@ private void UpdateReceiptModAckStatus(IEnumerable<string> ids, bool? status)
...
+                // If the exception isn't generally recoverable, don't retry.
+                if (!rpcEx.IsRecoverable())
+                {
+                    return false;
+                }
+
+                // If the exception was a failure due to auth, and we've seen some before, don't retry.
+                if (IsAuthException(rpcEx))
+                {
+                    var count = _exceptions.Count(IsAuthException);
```
amanda-tarafa, line 1188: Corner case likely as we've been failing anyway, but shouldn't the auth exceptions be consecutive for us to stop retrying? If there was an Auth exception and then 10 other exceptions, and then an Auth exception, etc. it might be that the auth error is really transient (and there's something else going on).

amanda-tarafa (approved):
LGTM, I left a couple of comments, but that if at all, would affect corer cases where the pull is failing anyway.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
jskeet: Discussed offline - we're going to leave this for now.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
jskeet: Discussed offline: we're going to leave this, as if there's a mix of errors it means things are pretty bad and we're unlikely to recover properly.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
jskeet: Discussed offline: while that *was* the previous intended behavior (just to prevent a memory leak), we're now going to fail (no retry) after this many exceptions.

amanda-tarafa (approved):
Thanks!

PR #13078, opened 2024-06-03
tools: Allow override of GAPIC_GENERATOR_VERSION

jskeet:
If it's already specified as an environment variable, just leave it.

Fixes b/344481559


amanda-tarafa (approved):
Nice!

PR #13131, opened 2024-06-04
tools: Ignore Ocktokit OverflowException when parsing the response after adding a comment

amanda-tarafa:


jskeet:
Another recently reported issue: https://github.com/octokit/octokit.net/issues/2927

jskeet (approved):

PR #13151, opened 2024-06-10
chore: Update Octokit to 12.0.0

jskeet:
This fixes recent issues we've had with the responses when parsing comments.


amanda-tarafa (approved):
Shouls we remove the try-catch, or just leave it in case this happens again in the future?

jskeet:
> Shouls we remove the try-catch, or just leave it in case this happens again in the future?

Let's leave it there.

PR #13281, opened 2024-07-16
refactor: Avoid storing a RowAsyncEnumerator unnecessarily

jskeet:
We only need to know whether or not ReadRowsStream has already been enumerated - we don't need the original enumerator. This makes it clearer that there's nothing for ReadRowsStream to dispose.

Addresses b/352471768


amanda-tarafa (approved):
Oh nice!

PR #13477, opened 2024-09-05
chore(deps): Update Microsoft.CodeAnalysis dependencies

amanda-tarafa:
This should make #13474 closable


jskeet:
We can't upgrade those packages though - see the comment above the version lines. Once we've moved to .NET 8 we'll be fine, but I don't think we can do it until then.

amanda-tarafa:
> We can't upgrade those packages though - see the comment above the version lines. Once we've moved to .NET 8 we'll be fine, but I don't think we can do it until then.

Ah yes, just realized again.

PR #13596, opened 2024-09-26
chore: Update to latest version of GAX

jskeet:


amanda-tarafa (approved):

jskeet:
We should already be allowing breaking changes in a beta, which I *believe* is the only breaking change here... but I'll check.

jskeet:
Ah - it's Google.Cloud.Kms.V1 that's causing the failure here. We should do a release of that carefully, considering whether to do a major bump or not.

PR #13656, opened 2024-10-04
feat: Support connection sharing across gRPC channels

amanda-tarafa:
@jskeet the first commit is some helpful refactoring.

For the second commit, naming ideas are welcomed.

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
@@ -57,6 +57,12 @@ public sealed class PublisherClientBuilder : ClientBuilderBase<PublisherClient>
     }
 
     /// <summary>
+    /// Whether the <see cref="PublisherServiceApiClient"/>s may share gRPC connections or not.
+    /// Defaults to false, meaning, each client has its own gRPC connection.
+    /// </summary>
+    public bool ShareClientConnections { get; set; }
+
+    /// <summary>
     /// Additional settings for batching, message ordering etc. Default settings will be used if this is null.
     /// </summary>
     public PublisherClient.Settings Settings { get; set; }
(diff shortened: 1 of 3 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientBuilder.cs
```diff
@@ -18,6 +18,7 @@
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using static Google.Cloud.PubSub.V1.ClientBuilderChannelHelper;
```
jskeet, line 21: I think it *might* be clearer to just qualify the calls to the methods. It's probably clear in VS, but I was a bit confused to start with.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientBuilder.cs
```diff
@@ -65,6 +56,12 @@ public int? ClientCount
         }
     }
 
+    /// <summary>
+    /// Whether the <see cref="PublisherServiceApiClient"/>s may share gRPC connections or not.
+    /// Defaults to false, meaning, each client has its own gRPC connection.
+    /// </summary>
+    public bool ClientsShareConnections { get; set; }
```
jskeet, line 63: Consider ShareClientConnections instead? I'm not massively concerned about either option though.

jskeet (approved):
Please change the message for the second commit before merging as that'll end up in release notes.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientBuilder.cs
amanda-tarafa: Yes, I'll change to that.

amanda-tarafa:
Latest changes done:

- Property name changed from ClientsShareConnections => ShareClientConnections
- Commit message and PR title changed as well

Merging now, thanks!

PR #13888, opened 2024-11-27
tools: Add a --unconfigured option for API generation

jskeet:
This allows generation (at least for most APIs) of code without an API being part of the API catalog. This is *not* code that's ready to ship - various things may be wrong - but can still be useful as part of an API producer's workflow. Running `./generateapis.sh --unconfigured` will generate every API it can find under googleapis - or APIs can be specified as *directories* (e.g. google/cloud/functions/v2) instead of as the .NET package name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,6 @@ build_timing_log.txt
 
 # Temporary local GAX
 gax-dotnet
+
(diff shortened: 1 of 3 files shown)


amanda-tarafa (approved):
Yep, I get what this is doing.

PR #13991, opened 2024-12-13
tools: Enable the result of "unconfigured" generation to be built afterwards

jskeet:


amanda-tarafa (approved):
This looks good, thanks!!

PR #13995, opened 2024-12-13
chore: Refactor GenerateSolutionFiles to avoid stuttering

jskeet:
We already provide the same name in the directory as in the csproj file, so let's use that to avoid duplication.


amanda-tarafa (approved):
Nice!

amanda-tarafa (approved):

PR #14006, opened 2024-12-14
fix: Evaluate RootLayout.ForCurrentDirectory lazily

### LINKED ISSUE #14005: Don't fail ReleaseManager.CommandBase ctor if getting repo root fails
When running a container command, we're usually not in a repo root... but in order to work out which command to use, we instantiate all commands.

We can probably just make the RootContainer property lazy, so that it fails appropriately if any path actually uses it, but doesn't fail in other scenarios.

jskeet:
This prevents start-up failures in the container environment (and any other time we're not in a repo).

Fixes #14005

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/CommandBase.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/CommandBase.cs
@@ -14,8 +14,10 @@
 
 using Google.Cloud.Tools.Common;
 using LibGit2Sharp;
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 
 namespace Google.Cloud.Tools.ReleaseManager
 {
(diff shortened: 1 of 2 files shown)


tools/Google.Cloud.Tools.ReleaseManager/CommandBase.cs
```diff
@@ -28,6 +30,10 @@ public abstract class CommandBase : ICommand
         // releases, but fundamentally it's "the current source of truth we're basing this release on".
         internal const string PrimaryBranch = "main";
 
+        // This is lazy so that we can construct instead of CommandBase even when the current directory isn't
```
amanda-tarafa, line 33:
```suggestion
        // This is lazy so that we can construct instances of CommandBase even when the current directory isn't
```

amanda-tarafa (approved):

amanda-tarafa (approved):

amanda-tarafa (approved):

PR #14012, opened 2024-12-16
feat: Allow a container-clean api-path of "none"

jskeet:
When provided "none", the container should clean just non-API-specific files.

(This is needed for the configure CLI command.)

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/CleanCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/CleanCommand.cs
@@ -25,6 +25,7 @@ namespace Google.Cloud.Tools.ReleaseManager.ContainerCommands;
 /// - repo-root: path to the root of the clone
 /// - api-path: path to the API to clean generated files from, e.g. google/cloud/functions/v2
 ///   This is optional; when omitted, all configured APIs are cleaned.
+///   The special value of "none" means "clean non-API-specific files only".
 /// </summary>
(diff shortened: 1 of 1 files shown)


tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/CleanCommand.cs
```diff
@@ -25,6 +25,7 @@ namespace Google.Cloud.Tools.ReleaseManager.ContainerCommands;
 /// - repo-root: path to the root of the clone
 /// - api-path: path to the API to clean generated files from, e.g. google/cloud/functions/v2
 ///   This is optional; when omitted, all configured APIs are cleaned.
+///   The special value of "none" means "clean non-API-specific files only".
```
amanda-tarafa, line 28: For later: Wondering if `api-path` should be required, with a couple of magic values "none" and "all"?

Alternative, with hide that from the surface and have three commands: clean-api, clean-all, clean-common?

Definetely something for another PR though.

amanda-tarafa (approved):

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/CleanCommand.cs
jskeet: You read my mind (first option): https://github.com/googleapis/generator/issues/67

amanda-tarafa: Nice :)

PR #14018, opened 2024-12-17
fix: Expect CommonResourcesConfig.json in generator input, not googleapis

jskeet:
Will self-merge as this is a trivial change within a code path which isn't exercised in normal usage.


jskeet:
@amanda-tarafa FYI...

PR #14019, opened 2024-12-17
fix: Set environment variables used by pre/postgeneration scripts

jskeet:
These were removed when refactoring GenerateApisCommand, as that no longer picks the values up from the environment when provided with a root layout - but we still want to be able to refer to them from the scripts.


jskeet:
Self-merging to unblock, but worth reviewing later @amanda-tarafa. (Would it be better to have something within GenerateApisCommand to do this based on a root layout? Then the environment variables could probably be private.)

amanda-tarafa:
> Self-merging to unblock, but worth reviewing later @amanda-tarafa. (Would it be better to have something within GenerateApisCommand to do this based on a root layout? Then the environment variables could probably be private.)

Yep, I think it could be worth it, specially if some of these env vars end up being needed later from other places? But definetely something that can wait.

PR #14081, opened 2025-01-10
Very early prototype of ExplainOptions for Datastore

jskeet:
This is just an attempt at the shape of things.

Still to do:

- Naming!
- Documentation
- Testing
- We can refactor a lot of other code to use AdvancedQuery
- Aggregation queries
- Transactions (hopefully put transaction ID support into AdvancedQuery and then just add overloads)


jskeet:
Looking into the test failure. I suspect I've inadvertently introduced an infinite loop.

amanda-tarafa (commented):
Yep, I like this.

jskeet:
Whoops, this is now obsolete.

PR #14117, opened 2025-01-17
feat: Allow Grafeas.V1 to use the default endpoint

jskeet:


jskeet:
The "breaking change" here isn't really - it's removing *overrides* of existing methods. I suspect that making the diff tool aware of that subtlety might be tricky. For now, I'll just mark this as breaking-changes-allowed.

amanda-tarafa (approved):

PR #14141, opened 2025-01-24
chore: Clean non-source files when generating locally

jskeet:
(This should probably never make a difference, but it's good to be sure.)


jskeet:
(I'm locally regenerating everything to make sure that it *doesn't* actually change anything, but I'm reasonably confident.)

amanda-tarafa (approved):

PR #14367, opened 2025-03-01
feat: BigTable Conformance Tests GRPC Proxy Implementation

cy-yun:
Reference: https://github.com/googleapis/java-bigtable/blob/main/test-proxy/src/main/java/com/google/cloud/bigtable/testproxy/CbtTestProxy.java

Implement conformance tests for BigTable.


apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,207 @@
+using Google.Api.Gax;
```
amanda-tarafa, line 1: Licenses

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,207 @@
...
+
+namespace Google.Cloud.Bigtable.V2.ConformanceTests;
+
+public class CloudBigtableV2TestProxyImpl : CloudBigtableV2TestProxy.CloudBigtableV2TestProxyBase
+{
+    private readonly Dictionary<string, CbtClient> _idClientMap;
+
+    private CloudBigtableV2TestProxyImpl() => _idClientMap = new();
+
+    private static void CheckArgument(bool condition, string message, ServerCallContext context)
```
amanda-tarafa, line 20: GaxPreconditions.CheckArgument

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,207 @@
...
+        {
+            BigtableServiceApiSettings settings = new BigtableServiceApiSettings
+            {
+                AppProfileId = request.AppProfileId
+            };
+
+            if (request.PerOperationTimeout is not null)
+            {
+                Expiration exp = Expiration.FromTimeout(TimeSpan.FromSeconds(request.PerOperationTimeout.Seconds));
+                settings.CallSettings = CallSettings.FromExpiration(exp);
```
amanda-tarafa, line 82: Per operation

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,207 @@
...
+            };
+            if (dataTarget != "emulator")
+            {
+                string accessToken = request.SecurityOptions.AccessToken;
+                builder.Endpoint = dataTarget;
+                builder.ChannelCredentials = GetChannelCredentials(securityOptions.UseSsl, securityOptions.SslRootCertsPem); ;
+                builder.GrpcChannelOptions = (securityOptions.UseSsl && securityOptions.SslEndpointOverride is not null)
+                    ? GrpcChannelOptions.Empty.WithCustomOption("grpc.ssl_target_name_override", securityOptions.SslEndpointOverride)
+                    : GrpcChannelOptions.Empty; ;
+                builder.GoogleCredential = accessToken is not null ? GoogleCredential.FromAccessToken(accessToken) : GoogleCredential.GetApplicationDefault();
```
cy-yun, line 97: don't set if accessToken is null

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,207 @@
...
+            }
+            else
+            {
+                builder.EmulatorDetection = EmulatorDetection.EmulatorOnly;
+            }
+            InstanceName instanceName = new InstanceName(projectId, instanceId);
+            GcpCallInvoker callInvoker = builder.CreateGcpCallInvoker();
+            builder.CallInvoker = callInvoker;
```
amanda-tarafa, line 105: Instead of doing this, after calling builder.BuildAsync you may have access to something called `LastCreatedChannel`.

amanda-tarafa (commented):
Looking good! I'll leave some more comments later.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,224 @@
...
+
+            BigtableServiceApiClientBuilder builder = new BigtableServiceApiClientBuilder
+            {
+                Settings = settings
+            };
+            if (dataTarget != "emulator")
+            {
+                string accessToken = request.SecurityOptions.AccessToken;
+                builder.Endpoint = dataTarget;
+                builder.ChannelCredentials = GetChannelCredentials(securityOptions.UseSsl, securityOptions.SslRootCertsPem); ;
```
amanda-tarafa, line 108: We cannot have ChannelCredentials and GoogleCredentials at the same time, the builder does not allow it.

If we have cases like those, we need to build the ChannelCredential to include the access token. I can show you how.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,224 @@
...
+        }
+        context.Status = new Status(StatusCode.OK, "CloseClient succeeded");
+        return new CloseClientResponse();
+    }
+
+    public override async Task<RemoveClientResponse> RemoveClient(RemoveClientRequest request, ServerCallContext context)
+    {
+        string clientId = request.ClientId;
+        GaxPreconditions.CheckArgument(clientId is not ("" or null), "ClientId", "client id must be provided", context);
+        bool removed = await Task.Run(() => _idClientMap.Remove(clientId)).ConfigureAwait(false);
```
amanda-tarafa, line 164: We don't need to run this asynchronously. Just call the remove method normally and then...

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,224 @@
...
+        if (!removed)
+        {
+            context.Status = new Status(StatusCode.NotFound, $"Client {clientId} not found.");
+            throw new RpcException(context.Status);
+        }
+        context.Status = new Status(StatusCode.OK, "RemoveClient succeeded");
+        return new RemoveClientResponse();
```
amanda-tarafa, line 171:
```suggestion
        return Task.FromResult(new RemoveClientResponse());
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,224 @@
...
+        catch (Exception e)
+        {
+            context.Status = new Status(StatusCode.Internal, e.Message, e);
+            throw new RpcException(context.Status);
+        }
+        context.Status = new Status(StatusCode.OK, "CloseClient succeeded");
+        return new CloseClientResponse();
+    }
+
+    public override async Task<RemoveClientResponse> RemoveClient(RemoveClientRequest request, ServerCallContext context)
```
amanda-tarafa, line 160:
```suggestion
    public override Task<RemoveClientResponse> RemoveClient(RemoveClientRequest request, ServerCallContext context)
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
amanda-tarafa, line 164: One question here is, we don't shutdown clients that we remove?
And we don't remove clients that we shutdown?

That may be the case, and the test runner could simply be calling Close and then Remove, but let's double check.

amanda-tarafa (commented):
I've left a few more comments. Happy to chat about them.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,224 @@
...
+        {
+            context.Status = new Status(StatusCode.NotFound, $"Client {clientId} not found.");
+            throw new RpcException(context.Status);
+        }
+        return _idClientMap[clientId];
+    }
+
+    private void OverrideTimeoutSetting(long timeout, BigtableServiceApiSettings settings)
+    {
+        settings.CheckAndMutateRowSettings = CallSettings.FromExpiration(Expiration.FromTimeout(TimeSpan.FromSeconds(timeout)));
```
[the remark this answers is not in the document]
cy-yun: got it. I had it like that initially but changed it to this implementation bc I was worried they might be mutable.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
cy-yun: Could you please link code samples or docs?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
cy-yun: I understand. I need the method to be asynchronous though because it's overriding an asynchronous virtual method that returns Task<object>.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
cy-yun: ah. This is nice. thank you

cy-yun: ah this is nice thank you!

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/CloudBigtableV2TestProxyImpl.cs
cy-yun: I copied the Java implementation for this.

.kokoro/bigtable-conformance.sh
```diff
@@ -0,0 +1,67 @@
...
+## Get the directory of the build script
+scriptDir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+## cd to the parent directory, i.e. the root of the git repo
+cd ${scriptDir}/..
+
+set +e
+
+# Build the proxy
+pushd .
+cd apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2ConformanceTest
```
amanda-tarafa, line 27:
```suggestion
cd apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTest
```

.kokoro/bigtable-conformance.sh
```diff
@@ -0,0 +1,54 @@
...
+# limitations under the License.
+
+set -eo pipefail
+
+## Get the directory of the build script
+scriptDir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+## cd to the parent directory, i.e. the root of the git repo
+cd ${scriptDir}/..
+
+set +e
```
amanda-tarafa, line 23: Remove this line, I think we still want an error to inmediately terminate the script?

.github/workflows/bigtable-conformance.yml
```diff
@@ -0,0 +1,42 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# Github action job to test core java library features on
+# downstream client libraries before they are released.
+on:
+  push:
+    branches:
+    - main
+  pull_request:
```
amanda-tarafa, line 20: Add here BigTable paths so that this action only executes when we make changes to BigTable. See the [Spanner emulator action](https://github.com/googleapis/google-cloud-dotnet/blob/main/.github/workflows/spanner-emulator-pr-push.yml) as an example.

.kokoro/bigtable-conformance.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
```
amanda-tarafa, line 1: This file shouldn't be in the .kokoro folder, because it's not being run by Kokoro. I think it should go on the Google.Cloud.BigTable.V2.ConformanceTests folder.

.kokoro/bigtable-conformance.sh
```diff
@@ -0,0 +1,54 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+## Get the directory of the build script
+scriptDir=$(realpath $(dirname "${BASH_SOURCE[0]}"))
+## cd to the parent directory, i.e. the root of the git repo
+cd ${scriptDir}/..
```
amanda-tarafa, line 21: In all other scripts in this repo we use the following instead. Use that for consistency here as well.

```
SCRIPT=$(readlink -f "$0")
SCRIPT_DIR=$(dirname "$SCRIPT")
```

amanda-tarafa: And possibly then `cd $SCRIPT_DIR`, so that way you can inmediately call `dotnet build` without having to pushd etc.

.kokoro/bigtable-conformance.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set +e
+
+# Build the proxy
+pushd .
+cd apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests
```
amanda-tarafa, line 27: This will change once you move the script into the ConformanceTests folder, and we won't need pushd and popd.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Protos/test_proxy.proto
```diff
@@ -0,0 +1,352 @@
...
+package google.bigtable.testproxy;
+
+import "google/api/client.proto";
+import "google/bigtable/v2/bigtable.proto";
+import "google/bigtable/v2/data.proto";
+import "google/protobuf/duration.proto";
+import "google/rpc/status.proto";
+
+option go_package = "./testproxypb";
+option csharp_namespace = "google.bigtable.testproxy";
```
amanda-tarafa, line 26: Since we control this one, let's do `Google.Cloud.Bigtable.V2.ConformanceTests.TestProxy` and regenerate.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Google.Cloud.Bigtable.V2.ConformanceTests.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <Nullable>enable</Nullable>
```
amanda-tarafa, line 5: Remove, we are not using nullable at the moment.
```suggestion
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Google.Cloud.Bigtable.V2.ConformanceTests.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Google.Cloud.Bigtable.V2\Google.Cloud.Bigtable.V2.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Grpc.AspNetCore" VersionOverride="2.40.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Core" VersionOverride="2.3.0" />
```
amanda-tarafa, line 15: These versions need to be added in `Directory.Packages.props` and here you'd only need: ****
```suggestion
    <PackageReference Include="Grpc.AspNetCore" />
    <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel.Core" />
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/appsettings.json
```diff
@@ -0,0 +1,19 @@
...
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
+    }
+  },
+  "AllowedHosts": "*",
+  "Kestrel": {
+    "EndpointDefaults": {
+      "Protocols": "Http2"
+    },
```
amanda-tarafa, line 12: Do we need to do this in the end, or just for MacOS?

If it's just for MacOs, then maybe move it to the appsettings.Development.json file?

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/appsettings.Development.json
```diff
@@ -0,0 +1,8 @@
+{
```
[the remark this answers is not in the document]
amanda-tarafa: Unless we need to add MacOs specific config here or similar.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+var builder = WebApplication.CreateBuilder(args);
+
+builder.WebHost.ConfigureKestrel(options =>
+{
+  // Setup a HTTP/2 endpoint without TLS.
+  // options.ListenLocalhost(7238, o => o.Protocols =
+  //     HttpProtocols.Http2);
+});
+// Additional configuration is required to successfully run gRPC on macOS.
+// For instructions on how to configure Kestrel and gRPC clients on macOS, visit https://go.microsoft.com/fwlink/?linkid=2099682
```
amanda-tarafa, line 13: Remove the commented code.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Program.cs
```diff
@@ -0,0 +1,24 @@
...
+
+// Add services to the container.
+builder.Services.AddGrpc();
+
+var app = builder.Build();
+
+// Configure the HTTP request pipeline.
+app.MapGrpcService<CloudBigtableV2TestProxyImpl>();
+app.MapGet("/", () => "Communication with gRPC endpoints must be made through a gRPC client. To learn how to create a client, visit: https://go.microsoft.com/fwlink/?linkid=2086909");
```
amanda-tarafa, line 22:
```suggestion
app.MapGet("/", () => "gRPC service for running conformance tests for Google.Cloud.Bigtable.V2");
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,435 @@
...
+
+        GaxPreconditions.CheckArgument(clientId is not ("" or null), "ClientId", "client id must be provided");
+        GaxPreconditions.CheckArgument(projectId is not ("" or null), "ProjectId", "project id must be provided");
+        GaxPreconditions.CheckArgument(instanceId is not ("" or null), "InstanceId", "instance id must be provided");
+        GaxPreconditions.CheckArgument(dataTarget is not ("" or null), "DataTarget", "data target must be provided");
+        GaxPreconditions.CheckArgument(securityOptions is null || (!securityOptions.UseSsl
+            || (securityOptions.SslRootCertsPem is not ("" or null))), "SecurityOptions",
+            "security_options.ssl_root_certs_pem must be provided if security_options.use_ssl is true");
+
+#pragma warning disable CS8604 // Possible null reference argument.
```
amanda-tarafa, line 59: When you remove `nullable` from the csproj, you can remove all of these.

amanda-tarafa (changes requested):
I've done a first pass. I'll continue tomorrow, fine if you want to wait until I'm done.

The changes in CloudBigtableV2TestProxyImpl apply to all methods there, not just the one where I left them.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,393 @@
...
+            throw new RpcException(context.Status);
+        }
+        context.Status = new Status(StatusCode.OK, "CloseClient succeeded");
+        return new CloseClientResponse();
+    }
+
+    public override Task<RemoveClientResponse> RemoveClient(RemoveClientRequest request, ServerCallContext context)
+    {
+        string clientId = request.ClientId;
+        GaxPreconditions.CheckArgument(clientId is not ("" or null), "ClientId", "client id must be provided", context);
```
amanda-tarafa, line 139: Just a reminder, remove all of these validations. Let's assume the text engine is passing the right test data.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,393 @@
...
+            return new RowResult
+            {
+                Status = SetExceptionStatus(e, context)
+            };
+        }
+    }
+
+    public override async Task<ExecuteQueryResult> ExecuteQuery(ExecuteQueryRequest request, ServerCallContext context)
+    {
+        CbtClient cbtClient = GetClient(request.ClientId, context);
```
amanda-tarafa, line 323: Following BitTable's advice let's throw Unimplemented exception here and skip all the execute query tests.

amanda-tarafa (changes requested):
A few more changes on the proxy implementation but nothing huge.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
cy-yun: I can do that. But if I remove the CloseClient implementation and throw NotImplementedException, the entire conformance test will fail. So I will only get rid of the ExecuteQuery implementation

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Google.Cloud.Bigtable.V2.ConformanceTests.csproj
cy-yun: I'm not sure if I understand this feedback!

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/bigtable-conformance.sh
```diff
@@ -14,13 +14,10 @@
 # limitations under the License.
 
 set -eo pipefail
+set +e
```
cy-yun, line 17: left this here because the github action will fail without it

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,365 @@
...
+            }
+            return result;
+        }
+        catch (RpcException e)
+        {
+            result.Status = SetExceptionStatus(e);
+            if (e.StatusCode == StatusCode.DeadlineExceeded)
+            {
+                int i = 0;
+                foreach (Google.Cloud.Bigtable.V2.MutateRowsRequest.Types.Entry entry in request.Request.Entries)
```
amanda-tarafa, line 249: Does this mean that the library could return a different code for each row? Let's check with the BT team.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,365 @@
...
+    private CbtClient GetClient(string clientId, ServerCallContext context)
+    {
+        if (!s_clientMap.ContainsKey(clientId))
+        {
+            context.Status = new Status(StatusCode.NotFound, $"Client {clientId} not found.");
+            throw new RpcException(context.Status);
+        }
+
+        return s_clientMap[clientId];
+    }
```
amanda-tarafa, line 346:
```suggestion
    private CbtClient GetClient(string clientId)
    {
        if (!s_clientMap.ContainsKey(clientId))
        {
            throw new RpcException(new Status(StatusCode.NotFound, $"Client {clientId} not found."));
        }

        return s_clientMap[clientId];
    }
```

amanda-tarafa (changes requested):
A few changes but mostly tyding things up.

This is great!

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Protos/test_proxy.proto
cy-yun: Go code does depend on it. The namespace for our proxy is hard coded in multiple places. One example here: https://github.com/googleapis/cloud-bigtable-clients-test/blob/main/testproxypb/test_proxy_grpc.pb.go#L36 . If we change this on our end, I'm certain that the test will not be able to find our proxy.

jskeet: That's the fully-qualified *protobuf* name though, which is based on the protobuf package declaration. Nothing else should care about the csharp_namespace.

cy-yun: I see what you mean. If the namespace for CloudBigtableV2TestProxyImpl.cs is what the test expects ("google.bigtable.testproxy") , the csharp namespace defined here can be anything else? I would need to add import statements in the CloudBigtableV2TestProxyImpl.cs though since the namespace for these dependencies would change.

What do you suggest we make the csharp_namespace here? I still think google.bigtable.testproxy is the best for consistency with CloudBigtableV2TestProxyImpl.cs

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/bigtable-conformance.sh
cy-yun: so something like this?

`pushd .
cd cloud-bigtable-clients-test/tests
# Cookie, RetryInfo, ExecuteQuery, ReverseScans and FeatureGap are known failures of new features that we don't yet support.
# CloseClient we don't support as expected, but we support it in a valid manner.
# For the others we have issues to investigate, see comments in b/372509076 .
set +e
eval "go test -v -proxy_addr=:7238 ${configFlag} -skip _Retry_WithRoutingCookie\|_Retry_WithRetryInfo\|_CloseClient\|_ReverseScans\|TestFeatureGap\|TestExecuteQuery\|TestReadRows_NoRetry_ErrorAfterLastRow\|TestReadRows_Retry_PausedScan\|TestReadRows_Retry_LastScannedRow_Reverse\|TestReadRow_Generic_DeadlineExceeded"
returnCode=$?
set -e
popd`

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Protos/test_proxy.proto
jskeet: It should be an idiomatic C# namespace. I'd expect it to be Google.Cloud.Bigtable.V2.ConformanceTests.Protos, probably. That way it fits in with the folder structure.

I don't know enough about CloudBigtableV2TestProxyImpl to know what's going on here (this is very much a drive-by comment). I'd expect you'll need to regenerate the C# code after changing the option - but then it should be fine.

amanda-tarafa: I think `Google.Cloud.Bigtable.V2.ConformanceTests.TestProxy` matches well here. When you change that yes, you have to regenerate the protos and then change the `using clause` on  CloudBigtableV2TestProxyImpl.cs.

Nothing else should be affected by this change. If you changed the `package` bit on this file, then we would be breaking things.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,365 @@
...
+            {
+                await cbtClient.LastCreatedChannel.ShutdownAsync();
+            }
+        }
+        catch (Exception e)
+        {
+            context.Status = new Status(StatusCode.Internal, e.Message, e);
+            throw new RpcException(context.Status);
+        }
+        return new CloseClientResponse();
```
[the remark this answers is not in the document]
cy-yun: Yes. CloseClient is being called after each test. I will remove the implementation and just leave 
```
return Task.FromResult(new CloseClientResponse());
```

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
cy-yun: Yes. That is correct.

https://github.com/googleapis/cloud-bigtable-clients-test/blob/main/tests/mutaterows_test.go#L196

each &mutateRowsAction represents a row. You will see that, in other tests, there are multiple of those defined. The delay is defined per row.

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,365 @@
...
+                settings.ReadRowsSettings = callSettings;
+                settings.ReadModifyWriteRowSettings = callSettings;
+                settings.SampleRowKeysSettings = callSettings;
+                settings.CallSettings = callSettings;
+            }
+        }
+
+        ChannelCredentials GetChannelCredentials()
+        {
+            // securityOptions.UseSsl, securityOptions.SslRootCertsPem, securityOptions.AccessToken
```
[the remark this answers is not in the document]
cy-yun: gotcha

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Services/CloudBigtableV2TestProxyImpl.cs
```diff
@@ -0,0 +1,365 @@
...
+using System.Diagnostics;
+
+namespace google.bigtable.testproxy;
+
+public sealed class CloudBigtableV2TestProxyImpl : CloudBigtableV2TestProxy.CloudBigtableV2TestProxyBase
+{
+    private class CbtClient
+    {
+        public BigtableClient Client { get; }
+        public ChannelBase LastCreatedChannel { get; }
```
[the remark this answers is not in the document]
cy-yun: yep. I got rid of CbtClient class all together because we didn't need InstanceName either

apis/Google.Cloud.Bigtable.V2/Google.Cloud.Bigtable.V2.ConformanceTests/Google.Cloud.Bigtable.V2.ConformanceTests.csproj
cy-yun: Got it. I understand now

amanda-tarafa (approved):
Amazing!

PR #14386, opened 2025-03-04
feat: Expose query planning and statistics

jskeet:
This approach wraps ExplainOptions (as an input, that may need to be reused) but not ExplainMetrics, PlanSummary etc which are output-only.


jskeet:
This is a simpler alternative to #14079, as it doesn't involve wrapping any output protos.

amanda-tarafa (approved):
I like this quite a bit.

PR #14401, opened 2025-03-06
tests: Fix Bucket soft-delete tests.

amanda-tarafa:
The soft-deleted buckets were being tracker by the fixture. The fixture tracks buckets for deletion which is done on a best effort basis, so that's not the problem. But, tracked buckets are also use as the expected result for ListAllBuckets tests, so these soft-deleted buckets were being expected but they were not showing up in actual lists.

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/GetBucketTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/GetBucketTest.cs
@@ -32,7 +32,8 @@ public class GetBucketTest
     public async Task SoftDeleted()
     {
         var bucketName = _fixture.GenerateBucketName();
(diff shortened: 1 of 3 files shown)


amanda-tarafa:
Fixes b/401146198

jskeet (approved):
Just as a thought, could we infer registerForDeletion as false if softDelete is true? Or do we sometimes create them but not delete them?

amanda-tarafa:
> Just as a thought, could we infer registerForDeletion as false if softDelete is true? Or do we sometimes create them but not delete them?

In one test we create a soft-delete bucket, delete it, and then restore it, which means that needs to be registered for deletion.
Note, this could actually cause flakiness in the case the Restore and ListAll tests are executing at the same time, where Restore gets to the point of deleting the bucket and ListAll executes before the bucket has been restored. I think that chances of that happening are pretty slim, but I'll keep an eye out.

PR #14491, opened 2025-03-19
tools: Allow the container "build" command to accept a library ID

jskeet:
This is a temporary measure to unblock the releasing part of the pipeline. We may well end up with two separate options, api-path and library-id, or we could *just* refer to library-id (which will cause difficulties for building "unconfigured" generated code).

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/BuildCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/BuildCommand.cs
@@ -65,6 +65,7 @@ public class BuildCommand : IContainerCommand
             processArguments.Add("--notests");
         }
 
+        List<string> apiIds = new();
         // By default, we build all configured APIs.
         // If an API path is configured, we build just that.
         // While build.sh without any arguments will query the API catalog, we might as well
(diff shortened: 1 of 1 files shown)


tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/BuildCommand.cs
```diff
@@ -74,16 +75,35 @@ private int BuildConfigured(Dictionary<string, string> options, string repoRoot)
             var targetApi = catalog.Apis.SingleOrDefault(api => api.ProtoPath == apiPath);
             if (targetApi is null)
             {
-                Console.WriteLine($"API path '{apiPath}' is not configured for any API.");
-                return 1;
+                // Maybe what was passed in wasn't a proto path, but a library ID (as per the pipeline-state.json, so it may be a package group ID)...
```
amanda-tarafa, line 78: It may be a single library ID as well, right? Although those are likely to be specified through the proto-path instead?

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/BuildCommand.cs
```diff
@@ -74,16 +75,35 @@ private int BuildConfigured(Dictionary<string, string> options, string repoRoot)
...
+                // This is probably a temporary measure - we'll probably end up with different options for api-path vs library-id.
```
amanda-tarafa, line 79: Yep +1 for this to be temporary.

What I don't fully understand is, why we need both. There's a mapping between API and library (groups) in the other PR, so this could just take a library ID (single or group) and librarian could do the mapping from proto path to library ID with the state?

amanda-tarafa (approved):

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/BuildCommand.cs
jskeet: When Librarian is running generation - which may be "raw" (e.g. in CI, with no context of google-cloud-dotnet) it wants to build what has just been generated, and there may be no library. When Librarian is creating a release PR, it wants to be able to build the library that it's preparing to release.

It may well be that we actually want to separate the two container build commands - but for now, this feels like a quick way of reusing the existing code while we're exploring the release process.

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/BuildCommand.cs
jskeet: Only if it's got state - which it won't have when doing raw generation...

PR #14500, opened 2025-03-20
tools: Separate "build" container command into "build-raw" and "build-library"

jskeet:
Librarian will work out which of these to call based on context.

(We're still discussing whether to do this, but it feels cleaner to me at the moment.)


jskeet:
@ldetmer is okay with this, so it's ready to merge if you're happy, @amanda-tarafa. (I'll then rebase #14501 to be a single commit.)

amanda-tarafa (approved):
Yes, I agree, this is definetely cleaner.

PR #14503, opened 2025-03-20
chore: Fix build

jskeet:
This *shouldn't* be necessary, but it appears to be - possibly due to the .NET SDK version.


amanda-tarafa (approved):
Yikes

PR #14576, opened 2025-04-04
chore: Add dependency on Google.Cloud.Location

jskeet:
Fixes regeneration.


jskeet:
As a side note, I do wonder whether we might be able to add dependencies dynamically as part of project generation if we're smart, via loading the service config. Will think about that in the background.

amanda-tarafa (approved):

PR #14606, opened 2025-04-10
feat: Introduce a new environment variable for local Librarian testing

jskeet:
If INTEGRATION_TEST_SERVICE_ACCOUNT_JSON is set, then the integration-test-library command will set up a service account file with the content of the environment variable. This is not intended for normal Librarian usage, where integration tests are only run for releasing, but it helps for Librarian/container testing.

The following command makes the JSON content available as a single line.

```sh
export INTEGRATION_TEST_SERVICE_ACCOUNT_JSON=$(cat $GOOGLE_APPLICATION_CREDENTIALS | sed -z 's/\n/ /g')
```

Fixes b/409731843

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/IntegrationTestLibraryCommand.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/IntegrationTestLibraryCommand.cs
@@ -15,6 +15,7 @@
 using Google.Cloud.Tools.Common;
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 
 namespace Google.Cloud.Tools.ReleaseManager.ContainerCommands;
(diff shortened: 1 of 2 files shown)


jskeet:
Astonishingly, this worked first time...

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/IntegrationTestLibraryCommand.cs
```diff
@@ -44,4 +46,36 @@ public int Execute(ContainerOptions options)
         }
         return 0;
     }
+
+    private void MaybeSaveServiceAccount()
```
amanda-tarafa, line 50: This is fine, but I wonder if it ain't better to configure Docker to use the host's ADC setup when running the container locally?

Or maybe I'm misunderstanding something.

amanda-tarafa (approved):

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/IntegrationTestLibraryCommand.cs
jskeet: That would be possible, but given the many ways that ADC can work, it could be pretty involved. If I've got an environment variable of GOOGLE_APPLICATION_CREDENTIALS defined to point to a file in some random directory, would we mount that directory, or copy the file, or something else? What about if it's just via gcloud? etc etc..

I'd be more inclined to put significant work into this if we expected running like this to be common - but currently I'm only expecting this to be used when developing either Librarian or a container.

PR #14632, opened 2025-04-17
tests: Remove need for a separate Firestore project

jskeet:
We still can't have a *default* database for Firestore and Datastore in the same project, so this changes all our Firestore integration tests and snippets to use a named database instead. (Changing Datastore would have been more disruptive.)


amanda-tarafa (approved):

jskeet:
Tests all now pass; merging.

PR #14666, opened 2025-04-25
ci: Restore committish override and repo cloning for legacy release process

amanda-tarafa:
This affects the legacy release process only.
Kokoro does not update submodules for GitHub repos when cloning such repos. We have stopped cloning the repo ourselves so now we need to update the submodules.

Fixes b/413558650

### DIFF
--- a/.kokoro/release.sh
+++ b/.kokoro/release.sh
@@ -16,6 +16,10 @@ source $SCRIPT_DIR/populatesecrets.sh
 
 git config --global core.longpaths true
 
+# We don't clone the repo anymore, we release in place.
+# But Kokoro is not fetching submodules when cloning the repo itself.
+git submodule update --init --recursive
+
 # Only populate secrets if we have to.
 # Else, we assume secrets have already been populated by the caller.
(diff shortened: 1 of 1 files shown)


.kokoro/release.sh
```diff
@@ -40,6 +44,19 @@ DOCS_CREDENTIALS="$SECRETS_LOCATION/docuploader_service_account"
...
+  COMMITTISH=$(git rev-parse HEAD)
+else
+  COMMITTISH=$COMMITTISH_OVERRIDE
+fi
+
+git clone https://github.com/googleapis/google-cloud-dotnet.git releasebuild -c core.autocrlf=input --recursive
+
+cd releasebuild
+  git checkout $COMMITTISH
+
```
jskeet, line 59:
```suggestion
COMMITTISH=$COMMITTISH_OVERRIDE
if [[ $COMMITTISH_OVERRIDE != "" ]]
then
  git checkout $COMMITTISH
fi
```

jskeet (approved):

PR #14669, opened 2025-04-25
tools: Docuploader supports ADC

amanda-tarafa:


jskeet (approved):
Fine as far as it goes, but we'll need to update generate-devsite-utilities.sh and generate-devsite-help.sh as well. That can come a bit later though. Another alternative might be to remove the support for the command line flag entirely, and let callers just set GOOGLE_APPLICATION_CREDENTIALS beforehand first?

jskeet (approved):
This will work for now - we should probably revisit it all when we've finished the double migration.

PR #14679, opened 2025-04-28
test: Simple App to test legacy errors

purva9413:
This is a simple app which will help us debug and visualize legacy errors that Compute throws in its client library. We can then draw comparisons to the same error scenario from the Compute APIs accessed through different platforms like gcloud or the GCP portal and check how we can better propagate these errors to the customer in our client library.

The error scenario reproduced by this app is trying to create an already existing disk resource in GCP. The disk can easily be created beforehand, before running this app.

Commands: 
```
## Use default auth from terminal of vscode
gcloud auth application-default login

## Run command
dotnet run project-id={your-project-id} zone={your-zone} disk-name={your-disk-name}
```


issues/TestLegacyErrors.sln
```diff
@@ -0,0 +1,22 @@
+﻿
```
amanda-tarafa, line 1: There should be a folder within issues that contains the solution file and the project folder. You can name the folder Issue408122240

issues/TestLegacyErrors/Program.cs
```diff
@@ -0,0 +1,100 @@
+﻿using Google.Cloud.Compute.V1;
```
amanda-tarafa, line 1: Add the copyright notice that all the others have.

issues/TestLegacyErrors/TestLegacyErrors.csproj
```diff
@@ -0,0 +1,15 @@
...
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Google.Apis.Auth" Version="1.69.0" />
```
amanda-tarafa, line 11: I don't think you need this, you will get it transitively from Google.Cloud.Compute.V1.

issues/TestLegacyErrors/Program.cs
```diff
@@ -0,0 +1,100 @@
...
+                    Console.WriteLine($"No errors found {result}");
+                }
+            }
+            catch (RpcException e)
+            {
+                Console.WriteLine("Got Exception!!");
+                Console.WriteLine($"Rpc Status: {e.GetRpcStatus()}");
+                Console.WriteLine($"Error Info: {e.GetErrorInfo()}");
+                // GetAllStatusDetails() should be available Google.Api.Gax v4.11.0 onwards
+                // Console.WriteLine($"Error Details: {e.GetAllStatusDetails()}");
```
amanda-tarafa, line 93: You can add the explicit GAX dependency to the version you need so you can uncomment this line.

amanda-tarafa (changes requested):

issues/TestLegacyErrors.sln
purva9413: Done!

issues/TestLegacyErrors/Program.cs
purva9413: Done!

issues/TestLegacyErrors/Program.cs
purva9413: I think latest GAX is at 4.10.0 - https://github.com/googleapis/gax-dotnet/releases/tag/Google.Api.Gax-4.10.0

We need this commit in order for this method to be there - https://github.com/googleapis/gax-dotnet/commit/46758edbecdf9ee2e08b42b15a1675b2661b8075

amanda-tarafa: Yep you are right.

amanda-tarafa (approved):
LGTM, but you need to rebase this PR on upstream/main into a single commit. The Diff Checker failures should go away then.

issues/TestLegacyErrors/TestLegacyErrors.csproj
purva9413: Done!

PR #14702, opened 2025-05-06
feat: Add EmitNumericTypes option for format parameters and scopes

### LINKED ISSUE #14693: Logging.Console to log format_parameters with type awareness
The items in `jsonPayload.format_parameters` are always logged as strings, e.g. `jsonPayload.format_parameters.ElapsedTime` is logged as `"123"`.  This prevents us from using numeric comparisons in queries.

The logging could be done with some awareness of the parameter's type.  The obvious case (and maybe the only one that's really needed) is that numbers should be logged without quotes, i.e. `jsonPayload.format_parameters.ElapsedTime` is logged as `123`.

The only downside I can see is a potential performance hit, but I'm not sure if this would be a big deal.  I considered putting something like the following into `WriteKeyValuePairs`:

```
            switch (pair.Value)
            {
                case string:
                    writer.WriteString(key, ToInvariantString(pair.Value)); // String is the most common so check it first (for performance).
                    break;
                case int intValue:
                    writer.WriteNumber(key, intValue);
                    break;               
                case decimal decimalValue:
                    writer.WriteNumber(key, decimalValue);
                    break;
                // etc...
                default:
                    writer.WriteString(key, ToInvariantString(pair.Value));
                    break;
            }
```

jskeet:
Fixes #14693

### DIFF
--- a/apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
+++ b/apis/Google.Cloud.Logging.Console/Google.Cloud.Logging.Console.Tests/GoogleCloudConsoleFormatterTest.cs
@@ -20,6 +20,7 @@ using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.IO;
 using System.Text.Json;
+using System.Text.Json.Nodes;
 using Xunit;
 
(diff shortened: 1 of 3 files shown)


amanda-tarafa (approved):
Oh this is good.

PR #14733, opened 2025-05-14
tests: Make it easier to diagnose some Compute tests failures.

amanda-tarafa:


apis/Google.Cloud.Compute.V1/Google.Cloud.Compute.V1.IntegrationTests/InstancesTest.cs
```diff
@@ -66,7 +66,7 @@ void CreateInstanceNonAsciiDescription()
                 };
                 var insertOp = instancesClient.Insert(projectId, zone, instanceResource);
                 var completed = insertOp.PollUntilCompleted(metadataCallback: metadata => _output.WriteLine($"Called back; metadata name={metadata.Name}"));
-                Assert.Null(completed.Exception);
+                Assert.True(completed.Exception is null, completed.Exception.ToString());
```
jskeet, line 69: This should be `completed.Exception?.ToString()` otherwise if the exception *is* null, we'll throw when evaluating the second argument.

amanda-tarafa: Yes, duh, Fixed.

PR #14742, opened 2025-05-15
tests: Fix tests for restricted environment

jskeet:
We need to detect the absence, not the presence, of GOOGLE_APPLICATION_CREDENTIALS.


amanda-tarafa (approved):
Oh duh to me.

PR #14755, opened 2025-05-16
tests: Updates to Storage tests for new CI

amanda-tarafa:


jskeet (approved):
Will check about the normalization resources, but everything else looks good.

PR #14793, opened 2025-05-23
tools: Remove all uses of the API index

jskeet:
This should definitely be reviewed one commit at a time.

While we're not *definitely* getting rid of the API index, .NET is probably the main user of it - and doesn't need to be. This puts us in a much better position to remove it.

This PR removes a few ReleaseManager commands which aren't needed in the new production pipeline; there's a lot more code that we should be able to delete over time. Additionally, GenerateApis can almost certainly be streamlined a lot now, removing much of the differentiation between "generate configured" and "generate unconfigured" - that will come in a later PR.


amanda-tarafa (approved):
This looks good!

jskeet:
Woot! Next week I'll start going further...

PR #14794, opened 2025-05-23
fix(BigTable): ReadRow/s respects rowsLimit and retries on Aborted exception

cy-yun:
Fixes TestReadRows_Retry_PausedScan, TestReadRows_NoRetry_ErrorAfterLastRow, and (partially) TestReadRow_Generic_DeadlineExceeded.

As of 5/23/2025, TestReadRow_Generic_DeadlineExceeded still fails not because DotNet BT library is behaving incorrectly, but the response status that the DotNet test proxy reports (OK) differs from Go client's expectations (DeadlineExceeded). This is because the Dotnet BT library swallows and retries DeadlineExceeded RPC exception and does not bubble the exception up to the Proxy. To make the it bubble up, we will need to make structural changes to the response model for ReadRow/s operations. However, I don't think this is worth it because such change will be irrelevant to our customers and we'd only be doing it to get 1 assert statement in the Go test to pass; other more important assert statements pass. I think the correct thing to do is to remove [this line](https://github.com/googleapis/cloud-bigtable-clients-test/blob/main/tests/readrow_test.go#L121) from the Go client test.


amanda-tarafa (approved):
This is great! Just two requests:

- Wait for @mutianf 's review before merging.
- Split each fix into its own commit, so in the future, it's clearer what fixes what.

PR #14807, opened 2025-05-29
chore: Update to .NET 8 SDK

jskeet:


jskeet:
> LGTM, but out of curiosity, why do we need the GH workflows to install .NET 6?

We still use .NET 6 for tests at the moment. We can change that in a bit, but this is a smaller step.

PR #14829, opened 2025-06-02
tools: Update the Bigtable client generator to .NET 8

jskeet:
This allows us to remove the MSBuildLocator entirely, and update our Roslyn dependencies (which had issues with .NET 6).


jskeet:
@amanda-tarafa Quick ping for this and #14831 - I think after this, #13810 *might* go away...

PR #14835, opened 2025-06-02
test: App to test UniverseDomain setting in Spanner connection string

purva9413:
The app needs setting of the required configs like project, instance etc before running. It tests the create table and insert row operations using the universe domain set through the connection string.


amanda-tarafa (commented):
This is fine, but I'm not sure it makes sense for us to merge it. The only difference from any other testing code we have is that it sets the universe domain. And since we will be doing that locally anyway, we can do it in the sample/test/... of our choosing.

What do you think @purva9413 ? Happy to chat about it tomorrow.

purva9413:
Closing this out, we do not need to merge this into the test apps but good to have this PR as a reference for anyone.

PR #14843, opened 2025-06-04
DO NOT MERGE: regenerate all APIs

jskeet:
This is basically to test the impact of https://github.com/googleapis/gapic-generator-csharp/pull/875


jskeet:
The changes to the Merchant IssueResolution API are due to https://github.com/googleapis/googleapis/commit/d4b3a8f0a9080724a2fe0866effc8478ce3efec2

The other non-Compute changes will need looking into separately, but I don't believe they're due to the generator change.

jskeet:
Merchant IssueResolution is failing due to a missing dependency; I'll add that in a new commit, but we'll want to add that in a PR that actually gets merged, too.

jskeet:
The ScratchDisks breaking change is an *expected* breaking change - see b/416655944

PR #14846, opened 2025-06-04
feat: Spanner support for isolation level

purva9413:


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -516,27 +516,32 @@ public Task<SpannerTransaction> BeginTransactionAsync(CancellationToken cancella
         /// <inheritdoc />
         protected override DbTransaction BeginDbTransaction(IsolationLevel isolationLevel)
```
amanda-tarafa: For this methos use :
```suggestion
        protected override DbTransaction BeginDbTransaction(IsolationLevel isolationLevel) =>
```

and wrap the long line below in one or two lines.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -516,27 +516,32 @@ public Task<SpannerTransaction> BeginTransactionAsync(CancellationToken cancella
...
-                throw new NotSupportedException(
-                    $"Cloud Spanner only supports isolation levels {IsolationLevel.Serializable} and {IsolationLevel.Unspecified}.");
-            }
-            return Task.Run(() => BeginTransactionAsync()).ResultWithUnwrappedExceptions();
+            return Task.Run(() => BeginTransactionAsync(SpannerTransactionCreationOptions.ReadWrite.WithIsolationLevel(ConvertToSpannerIsolatonLevel(isolationLevel)),transactionOptions: null, default)).ResultWithUnwrappedExceptions();
         }
 
 #if NETSTANDARD2_1_OR_GREATER
         /// <inheritdoc />
         protected override async ValueTask<DbTransaction> BeginDbTransactionAsync(IsolationLevel isolationLevel, CancellationToken cancellationToken)
```
amanda-tarafa: Same comment for this one as for the one above.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -516,27 +516,32 @@ public Task<SpannerTransaction> BeginTransactionAsync(CancellationToken cancella
...
         {
-            if (isolationLevel != IsolationLevel.Unspecified && isolationLevel != IsolationLevel.Serializable)
+            return await BeginTransactionAsync(SpannerTransactionCreationOptions.ReadWrite.WithIsolationLevel(ConvertToSpannerIsolatonLevel(isolationLevel)), transactionOptions: null, cancellationToken).ConfigureAwait(false);
+        }
+#endif
```
amanda-tarafa, line 526:
```suggestion
#endif

```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -516,27 +516,32 @@ public Task<SpannerTransaction> BeginTransactionAsync(CancellationToken cancella
...
+        private Google.Cloud.Spanner.V1.TransactionOptions.Types.IsolationLevel ConvertToSpannerIsolatonLevel(IsolationLevel isolationLevel)
```
amanda-tarafa: Use a switch expression instead: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -28,19 +29,19 @@ public sealed class SpannerTransactionCreationOptions
     /// Options that will result in a read-write transaction.
     /// </summary>
     public static SpannerTransactionCreationOptions ReadWrite { get; } = new SpannerTransactionCreationOptions(
-        timestampBound: null, transactionId: null, isDetached: false, isSingleUse: false, isPartitionedDml: false, excludeFromChangeStreams: false);
+        timestampBound: null, transactionId: null, isDetached: false, isSingleUse: false, isPartitionedDml: false, excludeFromChangeStreams: false, isolationLevel: IsolationLevel.Unspecified);
```
amanda-tarafa, line 34: Double check in the design that these are the right defaults.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -516,27 +516,32 @@ public Task<SpannerTransaction> BeginTransactionAsync(CancellationToken cancella
...
+        {
+            switch (isolationLevel)
             {
-                throw new NotSupportedException(
-                    $"Cloud Spanner only supports isolation levels {IsolationLevel.Serializable} and {IsolationLevel.Unspecified}.");
+                case IsolationLevel.Unspecified:
+                    return Google.Cloud.Spanner.V1.TransactionOptions.Types.IsolationLevel.Unspecified;
```
amanda-tarafa: Use an alias like `SpannerIsolationLevel`: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
amanda-tarafa: I think Serializable is the right default.

amanda-tarafa (commented):
I've left some comments. But looking more at the docs for Spanner's isolation level enum and System.Data.IsolationLevel, I think that in Google.Cloud.Spaner.Data we shouldn't expose Spanner's isolation level, instead we should expose a System.Data.IsolationLevel property and document that only Serializable and RepeatableRead are supportes. Let's chat tomorrow about this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
purva9413: Yes I think in the design they say Serializable but in the compiled proto file it is Unspecified with a conversion to Serializable in the backend - https://github.com/googleapis/google-cloud-dotnet/blob/0f4ee4ce5358ef119904c5ae80761b5b872e4230/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Transaction.g.cs#L868-L874

All the other languages are also using Unspecified afaict so thought we should keep this consistent.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: Done and moved to SpannerTransactionCreationOptions

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -14,7 +14,11 @@
 
 using Google.Api.Gax;
 using Google.Cloud.Spanner.V1;
+using System;
+//using System.Data;
```
amanda-tarafa:
```suggestion
```

And also check the rest in case we can clean them further.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -92,6 +96,12 @@ public sealed class SpannerTransactionCreationOptions
     /// </summary>
     public bool IsPartitionedDml { get; }
 
+    /// <summary>
+    /// Snapshot Isolation Level set to <see cref="TransactionOptions.Types.IsolationLevel"/>.
```
amanda-tarafa: I think this is: 
```suggestion
    /// Snapshot Isolation Level set to <see cref="System.Data.IsolationLevel"/>.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -137,6 +147,12 @@ private SpannerTransactionCreationOptions(TimestampBound timestampBound, Transac
             nameof(excludeFromChangeStreams),
             "Only read-write and partioned DML transactions can be marked for change stream exclusion.");
         ExcludeFromChangeStreams = excludeFromChangeStreams;
+        Console.WriteLine($"TransactionMode {TransactionMode}, IsolationLevel {isolationLevel}");
```
amanda-tarafa:
```suggestion
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -137,6 +147,12 @@ private SpannerTransactionCreationOptions(TimestampBound timestampBound, Transac
...
+        GaxPreconditions.CheckArgument(
+            TransactionMode == TransactionMode.ReadWrite || isolationLevel == IsolationLevel.Unspecified,
+            nameof(isolationLevel),
+            $"Isolation Level can only be specified for read-write transactions. {TransactionMode}, {isolationLevel}");
```
amanda-tarafa:
```suggestion
            $"Isolation Level can only be specified for read-write transactions. This transaction would be {TransactionMode} and the specified level is {isolationLevel}");
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -201,12 +220,29 @@ public SpannerTransactionCreationOptions WithIsDetached(bool isDetached) =>
...
+
+    private SpannerIsolationLevel ConvertToSpannerIsolatonLevel(IsolationLevel isolationLevel) => isolationLevel switch
+    {
+        IsolationLevel.Unspecified => SpannerIsolationLevel.Unspecified,
+        IsolationLevel.Serializable => SpannerIsolationLevel.Serializable,
+        IsolationLevel.RepeatableRead => SpannerIsolationLevel.RepeatableRead,
+        IsolationLevel.Snapshot => SpannerIsolationLevel.RepeatableRead,
+        _ => throw new NotSupportedException(
+            $"Cloud Spanner currently does not support {nameof(isolationLevel)}"),
+    };
```
amanda-tarafa: You can inline this method within GetTransactionOptions, i.e. a method within a method, as that's the only place you are ever going to use it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -78,6 +78,7 @@ public sealed class SpannerTransaction : SpannerTransactionBase, ISpannerTransac
         private int _lastDmlSequenceNumber = 0;
 
         /// <inheritdoc />
+        /// This is currently unused, but if ever it is used we need to convert it to <see cref="Google.Cloud.Spanner.V1.TransactionOptions.Types.IsolationLevel" />
```
amanda-tarafa: I think this needs changing, we should return the value on the options now, right?
And I do wonder if we should return `Serializable` if the options are set to `Unspecified`, as that's what has been supported until now.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
```diff
@@ -250,6 +251,78 @@ public async Task AbortedThrownCorrectly()
...
+            // Requires multiple read/write transactions which is not supported on the emulator.
+
+            // Note: deeply nested using statements to ensure that we dispose of everything even in the case of failure,
+            // but we manually dispose of both tx1 and connection1.
+            using (var connection1 = new SpannerConnection(_fixture.ConnectionString))
+            {
+                using (var connection2 = new SpannerConnection(_fixture.ConnectionString))
+                {
+
+                    await Task.WhenAll(connection1.OpenAsync(), connection2.OpenAsync());
```
amanda-tarafa: You don't need to do this, begin transaction will open the connection if it's closed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
```diff
@@ -250,6 +251,78 @@ public async Task AbortedThrownCorrectly()
...
+
+                            // TX1 WRITE/COMMIT
+                            using (var cmd = connection1.CreateUpdateCommand(_fixture.TableName))
+                            {
+                                cmd.Parameters.Add("k", SpannerDbType.String, _key);
+                                cmd.Parameters.Add("Int64Value", SpannerDbType.Int64, 0);
+                                cmd.Transaction = tx1;
+                                await cmd.ExecuteNonQueryAsync();
+                                await tx1.CommitAsync();
+                                tx1.Dispose();
```
amanda-tarafa: Commit will dispose also, you don't need to dispose here explicitly.

I think you don't need the nested using statements, you can just use using declarations to make certain we dispose of things in the case of failures. Like this `using var cmd = ...;` notice no braces.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
```diff
@@ -250,6 +251,78 @@ public async Task AbortedThrownCorrectly()
...
+                            }
+                            connection1.Dispose();
+
+                            // TX2 READ again should not throw an exception
+                            using (var cmd = CreateSelectAllCommandForKey(connection2))
+                            {
+                                cmd.Transaction = tx2;
+                                using (var reader = await cmd.ExecuteReaderAsync())
+                                {
+                                    // Best we can do is assert the read is successful even after TX1 commits write as TX2 should have snapshot isolation
```
amanda-tarafa: And check that the new data is not visible here?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
```diff
@@ -219,4 +226,22 @@ public void ExcludeFromChangeStream_PartitionedDml()
         var options = SpannerTransactionCreationOptions.PartitionedDml.WithExcludeFromChangeStreams(true);
         Assert.True(options.ExcludeFromChangeStreams);
     }
+
+    [Fact]
+    public void IsolationLevel_ReadOnly() =>
+        Assert.Throws<ArgumentException>(() => SpannerTransactionCreationOptions.ReadOnly.WithIsolationLevel(IsolationLevel.RepeatableRead));
+
+    [Fact]
+    public void IsolationLevel_ReadWrite()
```
amanda-tarafa, line 236: We need to check that the ADO.NET option is correctly translated to the Spanner option. So you need two theories (one for success and one for failures) that attempt to set the ADO.NET isolation level and check that the resulting Spanner transaction options contains the correct Spanner isolation level.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
amanda-tarafa: Thinking more about this, I don't think we should have an integration test for testing backend behavior. I think we only need the unit tests I requested below, we are testing elsewhere that the correct Spanner transaction options are sent when starting a Spanner transaction. For this change, we only need to check that those transaction options contain the right isolation level option.

amanda-tarafa (changes requested):
Changes are mostly nits. But we can drop the integration test and add more unit tests.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
purva9413: Good idea!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
purva9413: Resolving since we are removing this integration test.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
purva9413: Resolving since we are removing this integration test.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
purva9413: Ack, this integration test caught the issue discussed in b/331523252 so thought I'd put it in. But if we have integration tests to test correct setting of the TransactionOptions all the way to the Spanner backend I think it should be sufficient for this feature once that issue is resolved. 

Will remove this integration test and add in more unit tests.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
purva9413: Re: converting the Unspecified to Serializable, my two cents would be it might be confusing for users who explicitly set the IsolationLevel to `Unspecified`, and when they access it they get `Serializable` instead. 

But for Spanner IsolationLevel there is an implicit conversion frm `Unspecified` to `Serializable` [here](https://github.com/googleapis/google-cloud-dotnet/blob/7e4f3d1203e5b3c7acdc8f8ad161a70b46f8f8d8/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Transaction.g.cs#L868-L874).

We could return Serializable when Unspecified is set making this explicit to the user or just document it for clarity. Lmk your thoughts.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
purva9413: Ack, added in the unit tests.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
amanda-tarafa: Yes, I'm torn as well. 

My thinking for asking this is, up until now, users would always see Seriablizable, regardless of whether they specified isolation level or not. So, in principle, calling code may have been creating transactions without specifying isolation level and seeing here serializable. We are now changing that behaviour.

Yep, OK, I think it's acceptable to return what's in the options.

Also, remove the bit of documentaiton that says "this is unused".

purva9413: Done, updated the comment! Also opening this PR up for official review from its draft state.

amanda-tarafa:
Closing in favor of #14987

PR #14848, opened 2025-06-05
feat: Support the Interval type

amanda-tarafa:
Closes #14254

@Hectorhammett as FYI.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerDbTypeTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerDbTypeTests.cs
@@ -83,20 +83,22 @@ namespace Google.Cloud.Spanner.Data.Tests
             yield return new object[] { "TIMESTAMP", SpannerDbType.Timestamp };
             yield return new object[] { "NUMERIC", SpannerDbType.Numeric };
             yield return new object[] { "NUMERIC{PG}", SpannerDbType.PgNumeric };
+            yield return new object[] { "INTERVAL", SpannerDbType.Interval };
 
             yield return new object[] { " STRING  ", SpannerDbType.String };
-            yield return new object[] { "  BOOL  ", SpannerDbType.Bool };
-            yield return new object[] { "  BYTES  ", SpannerDbType.Bytes };
-            yield return new object[] { "  JSON  ", SpannerDbType.Json };
-            yield return new object[] { "  JSONB{PG}  ", SpannerDbType.PgJsonb };
-            yield return new object[] { "  DATE  ", SpannerDbType.Date };
-            yield return new object[] { "  FLOAT32  ", SpannerDbType.Float32 };
-            yield return new object[] { "  FLOAT64  ", SpannerDbType.Float64 };
-            yield return new object[] { "  INT64  ", SpannerDbType.Int64 };
-            yield return new object[] { "  OID{PG}", SpannerDbType.PgOid };
-            yield return new object[] { "  TIMESTAMP  ", SpannerDbType.Timestamp };
-            yield return new object[] { "  NUMERIC  ", SpannerDbType.Numeric };
-            yield return new object[] { "  NUMERIC{PG}  ", SpannerDbType.PgNumeric };
+            yield return new object[] { " BOOL  ", SpannerDbType.Bool };
+            yield return new object[] { " BYTES  ", SpannerDbType.Bytes };
+            yield return new object[] { " JSON  ", SpannerDbType.Json };
+            yield return new object[] { " JSONB{PG}  ", SpannerDbType.PgJsonb };
+            yield return new object[] { " DATE  ", SpannerDbType.Date };
+            yield return new object[] { " FLOAT32  ", SpannerDbType.Float32 };
+            yield return new object[] { " FLOAT64  ", SpannerDbType.Float64 };
+            yield return new object[] { " INT64  ", SpannerDbType.Int64 };
+            yield return new object[] { " OID{PG}  ", SpannerDbType.PgOid };
+            yield return new object[] { " TIMESTAMP  ", SpannerDbType.Timestamp };
+            yield return new object[] { " NUMERIC  ", SpannerDbType.Numeric };
+            yield return new object[] { " NUMERIC{PG}  ", SpannerDbType.PgNumeric };
+            yield return new object[] { " INTERVAL  ", SpannerDbType.Interval };
 
             yield return new object[] { "STRING(2)", SpannerDbType.String.WithSize(2) };
             yield return new object[] { "STRING(100)", SpannerDbType.String.WithSize(100) };
@@ -123,6 +125,7 @@ namespace Google.Cloud.Spanner.Data.Tests
             yield return new object[] { "ARRAY<INT64>", SpannerDbType.ArrayOf(SpannerDbType.Int64) };
             yield return new object[] { "ARRAY<OID{PG}>", SpannerDbType.ArrayOf(SpannerDbType.PgOid) };
             yield return new object[] { "ARRAY<TIMESTAMP>", SpannerDbType.ArrayOf(SpannerDbType.Timestamp) };
+            yield return new object[] { "ARRAY<INTERVAL>", SpannerDbType.ArrayOf(SpannerDbType.Interval) };
 
             yield return new object[] { "ARRAY<STRING(5)>", SpannerDbType.ArrayOf(SpannerDbType.String), false };
             yield return new object[] { "ARRAY<BYTES(5)>", SpannerDbType.ArrayOf(SpannerDbType.Bytes), false };
@@ -182,9 +185,10 @@ namespace Google.Cloud.Spanner.Data.Tests
                 { "F10", SpannerDbType.Json, null },
                 { "F11", SpannerDbType.PgNumeric, null },
                 { "F12", SpannerDbType.PgJsonb, null },
-                { "F13", SpannerDbType.PgOid, null }
+                { "F13", SpannerDbType.PgOid, null },
+                { "F14", SpannerDbType.Interval, null}
             };
-            yield return new object[] { "STRUCT<F1:STRING,F2:INT64,F3:BOOL,F4:BYTES,F5:DATE,F6:FLOAT32,F7:FLOAT64,F8:TIMESTAMP,F9:NUMERIC,F10:JSON,F11:NUMERIC{PG},F12:JSONB{PG},F13:OID{PG}>", sampleStruct.GetSpannerDbType() };
+            yield return new object[] { "STRUCT<F1:STRING,F2:INT64,F3:BOOL,F4:BYTES,F5:DATE,F6:FLOAT32,F7:FLOAT64,F8:TIMESTAMP,F9:NUMERIC,F10:JSON,F11:NUMERIC{PG},F12:JSONB{PG},F13:OID{PG},F14:INTERVAL>", sampleStruct.GetSpannerDbType() };
 
             sampleStruct = new SpannerStruct
             {
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
@@ -1,0 +1,337 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Numerics;
+using Xunit;
+
+namespace Google.Cloud.Spanner.V1.Tests
+{
+    public class IntervalTests
+    {
+        private static readonly BigInteger NanonosecondsInASecond = (BigInteger)1_000_000_000;
+        private static readonly BigInteger NanonosecondsInAMillisecond = (BigInteger)1_000_000;
+        private static readonly BigInteger NanonosecondsInAMicrosecond = (BigInteger)1_000;
+
+        [Theory]
+        [InlineData(5)]
+        [InlineData(-5)]
+        [InlineData(13)]
+        [InlineData(-13)]
+        [InlineData(24)]
+        [InlineData(-24)]
+        [InlineData(120_000)]
+        [InlineData(-120_000)]
+        public void FromMonths(int months)
+        {
+            Interval interval = Interval.FromMonths(months);
+            Assert.Equal(months, interval.Months);
+            Assert.Equal(0, interval.Days);
+            Assert.Equal(0, interval.Nanoseconds);
+        }
+
+        [Theory]
+        [InlineData(120_001)]
+        [InlineData(-120_001)]
+        [InlineData(int.MaxValue)]
+        public void FromMonths_OutOfRange(int months) =>
+            Assert.Throws<ArgumentOutOfRangeException>(() => Interval.FromMonths(months));
+
+        [Theory]
+        [InlineData(5)]
+        [InlineData(-5)]
+        [InlineData(30)]
+        [InlineData(-30)]
+        [InlineData(31)]
+        [InlineData(-31)]
+        [InlineData(3_660_000)]
+        [InlineData(-3_660_000)]
+        public void FromDays(int days)
+        {
+            Interval interval = Interval.FromDays(days);
+            Assert.Equal(days, interval.Days);
+            Assert.Equal(0, interval.Months);
+            Assert.Equal(0, interval.Nanoseconds);
+        }
+
+        [Theory]
+        [InlineData(3_660_001)]
+        [InlineData(-3_660_001)]
+        [InlineData(int.MaxValue)]
+        public void FromDays_OutOfRange(int days) =>
+            Assert.Throws<ArgumentOutOfRangeException>(() => Interval.FromDays(days));
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(-1)]
+        [InlineData(60)]
+        [InlineData(-60)]
+        [InlineData(90)]
+        [InlineData(-90)]
+        [InlineData(32_767)]
+        public void FromSeconds_Short(short seconds)
+        {
+            Interval interval = Interval.FromSeconds(seconds);
+            Assert.Equal(seconds * NanonosecondsInASecond, interval.Nanoseconds);
+            Assert.Equal(0, interval.Months);
+            Assert.Equal(0, interval.Days);
+        }
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(-1)]
+        [InlineData(60)]
+        [InlineData(-60)]
+        [InlineData(90)]
+        [InlineData(-90)]
+        [InlineData(2_147_483_647)]
+        public void FromSeconds_Int(int seconds)
+        {
+            Interval interval = Interval.FromSeconds(seconds);
+            Assert.Equal(seconds * NanonosecondsInASecond, interval.Nanoseconds);
+            Assert.Equal(0, interval.Months);
+            Assert.Equal(0, interval.Days);
+        }
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(-1)]
+        [InlineData(60)]
+        [InlineData (-60)]
+        [InlineData(90)]
+        [InlineData(-90)]
+        [InlineData(316_224_000_000)]
+        public void FromSeconds_Long(long seconds)
+        {
+            Interval interval = Interval.FromSeconds(seconds);
+            Assert.Equal(seconds * NanonosecondsInASecond, interval.Nanoseconds);
+            Assert.Equal(0, interval.Months);
+            Assert.Equal(0, interval.Days);
+        }
+
+        [Theory]
+        [InlineData("1")]
+        [InlineData("-1")]
+        [InlineData("60")]
+        [InlineData("-60")]
+        [InlineData("90")]
+        [InlineData("-90")]
+        [InlineData("316224000000")]
+        [InlineData("316223000000.999999999")]
+        [InlineData("0.123456789")]
+        public void FromSeconds_Decimal(string secondsText)
+        {
+            var seconds = decimal.Parse(secondsText);
+            Interval interval = Interval.FromSeconds(seconds);
+            Assert.Equal(new BigInteger(seconds * (long) NanonosecondsInASecond), interval.Nanoseconds);
+            Assert.Equal(0, interval.Months);
+            Assert.Equal(0, interval.Days);
+        }
+
+        [Theory]
+        [InlineData("316224000001")]
+        [InlineData("316224000000.1")]
+        [InlineData("0.1234567891")]
+        [InlineData("-316224000001")]
+        [InlineData("-316224000000.1")]
+        [InlineData("-0.1234567891")]
+        public void FromSeconds_OutOfRange(string secondsText)
+        {
+            var seconds = decimal.Parse(secondsText);
+            Assert.Throws<ArgumentOutOfRangeException>(() => Interval.FromSeconds(seconds));
+        }
+
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(-1)]
+        [InlineData(1000)]
+        [InlineData(-1000)]
+        [InlineData(999)]
+        [InlineData(-999)]
+        [InlineData(316_224_000_000_000)]
+        public void FromMilliseconds(long milliseconds)
+        {
+            Interval interval = Interval.FromMilliseconds(milliseconds);
+            Assert.Equal(milliseconds * NanonosecondsInAMillisecond, interval.Nanoseconds);
+            Assert.Equal(0, interval.Months);
+            Assert.Equal(0, interval.Days);
+        }
+
+        [Theory]
+        [InlineData(316_224_000_000_001)]
+        [InlineData(long.MaxValue)]
+        [InlineData(-316_224_000_000_001)]
+        [InlineData(long.MinValue)]
+        public void FromMilliseconds_OutOfRange(long milliseconds) =>
+            Assert.Throws<ArgumentOutOfRangeException>(() => Interval.FromMilliseconds(milliseconds));
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(-1)]
+        [InlineData(1_000_000)]
+        [InlineData(-1_000_000)]
+        [InlineData(999_999)]
+        [InlineData(-999_999)]
+        [InlineData(316_224_000_000_000_000)]
+        public void FromMicroseconds(long microseconds)
+        {
+            Interval interval = Interval.FromMicroseconds(microseconds);
+            Assert.Equal(microseconds * NanonosecondsInAMicrosecond, interval.Nanoseconds);
+            Assert.Equal(0, interval.Months);
+            Assert.Equal(0, interval.Days);
+        }
+
+        [Theory]
+        [InlineData(316_224_000_000_000_001)]
+        [InlineData(long.MaxValue)]
+        [InlineData(-316_224_000_000_000_001)]
+        [InlineData(long.MinValue)]
+        public void FromMicroseconds_OutOfRange(long microseconds) =>
+            Assert.Throws<ArgumentOutOfRangeException>(() => Interval.FromMicroseconds(microseconds));
+
+        [Theory]
+        [InlineData(1)]
+        [InlineData(-1)]
+        [InlineData(1_000_000_000)]
+        [InlineData(-1_000_000_000)]
+        [InlineData(999_999_999)]
+        [InlineData(-999_999_999)]
+        public void FromNanoseconds(long nanoseconds)
+        {
+            Interval interval = Interval.FromNanoseconds(nanoseconds);
+            Assert.Equal(nanoseconds, interval.Nanoseconds);
+            Assert.Equal(0, interval.Months);
+            Assert.Equal(0, interval.Days);
+        }
+
+        [Theory]
+        [InlineData("316224000000000000000")]
+        [InlineData("-316224000000000000000")]
+        public void FromBigIntegerNanoseconds(string nanosecondsText)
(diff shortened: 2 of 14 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+using System.Globalization;
+using System.Numerics;
+using System.Text;
+
+namespace Google.Cloud.Spanner.V1;
+
+/// <summary>
+/// Represents the Spanner Interval type.
+/// </summary>
+public class Interval
```
jskeet: I assume this could be sealed?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+{
+    /// <summary>
+    /// Represents the zero Interval value.
+    /// </summary>
+    public static readonly Interval Zero = new Interval(0, 0, 0);
```
jskeet: I'd make these properties (just by adding `{ get; }`).

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+    public static readonly Interval MinValue = new Interval(MinMonths, MinDays, MinNanoseconds);
+
+    internal const int MaxMonths = 120_000;
+    internal const int MinMonths = -MaxMonths;
+    internal const int MaxDays = 3_660_000;
+    internal const int MinDays = -MaxDays;
+    internal static readonly BigInteger MaxNanoseconds = BigInteger.Parse("316224000000000000000");
+    internal static readonly BigInteger MinNanoseconds = -MaxNanoseconds;
+
+    internal const long NanosecondsInASecond = 1_000_000_000;
```
jskeet: I'd rename all of these to use "Per" instead of "InA" - similar to https://learn.microsoft.com/en-us/dotnet/api/system.timespan.microsecondsperday

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+
+    /// <summary>
+    /// Creates an interval from the given days.
+    /// </summary>
+    public static Interval FromDays(int days) => new Interval(0, days, 0);
+
+    /// <summary>
+    /// Creates an interval from the given seconds.
+    /// </summary>
+    public static Interval FromSeconds(decimal seconds) => new Interval(0, 0, SecondsToNanoseconds(seconds));
```
jskeet: I wonder whether it's worth adding int-parameter overloads for all of these. We'd probably want to check whether that creates any issues if you have a `short` argument for example - I don't know offhand if the conversion to `int` would be "better than" the conversion to `BigInteger`.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+    /// Parses the string represantion in ISO8601 format of a duration returning its Interval equivalent.
+    /// </summary>
+    /// <param name="text">
+    /// The string representation of a duration in ISO8601 format: P[n]Y[n]M[n]DT[n]H[n]M[n[.fraction]]S.
+    /// </param>
+    public static Interval Parse(string text)
+    {
+        GaxPreconditions.CheckNotNull(text, nameof(text));
+
+        var state = new IntervalParsingState();
```
jskeet: Maybe move all the actual logic to IntervalParsingState, so that this line becomes:

```csharp
return new IntervalParsingState().Parse(text);
```

(The fact that that could remove 65 occurrences of `state.` suggests it would help...)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+        {
+            throw new ArgumentOutOfRangeException($"Interval has a nanosecond precision, only 9 decimal places are allowed for seconds.");
+        }
+
+        return new BigInteger(seconds * NanosecondsInASecond);
+    }
+
+    private class IntervalParsingState
+    {
+        internal static char[] P = new char[] { 'P' };
```
jskeet: Let's make all of these read-only - and if we *do* move the logic here, everything can become private.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+        /// </summary>
+        public bool MayBeTerminal { get; set; }
+
+        /// <summary>
+        /// True if we are a state that must be terminal.
+        /// </summary>
+        public bool IsTerminal { get; set; }
+
+        /// <summary>
+        /// The delimeters that are allowed next.
```
jskeet: Nit: delimiters

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
jskeet: Thinking about it more... do we actually need the separate class at all? Can we not just move the fields from there to local variables in this method?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+                    state.NextAllowed = IntervalParsingState.AfterMins;
+                    break;
+                case 'S':
+                    state.MayBeTerminal = true;
+                    state.IsTerminal = true;
+                    state.Seconds = ParseSeconds();
+                    state.NextAllowed = null;
+                    break;
+                default:
+                    throw new InvalidOperationException("There's a bug in Interval.Parse.");
```
jskeet: Maybe include `text[end]` in the message? (Maybe other bits of state?)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+    /// The resulting Interval value, if parsing was succesful. null otherwise.
+    /// </param>
+    public static bool TryParse(string text, out Interval interval)
+    {
+        try
+        {
+            interval = Parse(text);
+            return true;
+        }
+        catch (Exception)
```
jskeet: Just `catch` here?

Alternatively, we might want to restructure this to avoid the performance penalty of the exception - although we can do that later.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+        // We are at a terminal state but we haven't parsed the whole string yet.
+        if (state.IsTerminal && state.Start < text.Length)
+        {
+            throw new FormatException($"Can't parse {text.Substring(state.Start)} after reaching the end of a valid duration.");
+        }
+
+        // We parsed the whole string but we ended up at a state that's not terminal.
+        if (!state.MayBeTerminal)
+        {
+            throw new FormatException($"{text} ending is not correctly formatted.");
```
jskeet: I'd suggest using `$"'{text}' ending etc"` to make it clear which part is the value. Ditto above.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+            }
+
+            state.Start = end + 1;
+
+        } while (state.Start < text.Length && !state.IsTerminal);
+
+        // We are at a terminal state but we haven't parsed the whole string yet.
+        if (state.IsTerminal && state.Start < text.Length)
+        {
+            throw new FormatException($"Can't parse {text.Substring(state.Start)} after reaching the end of a valid duration.");
```
jskeet: Interval rather than duration?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+        var state = new IntervalParsingState();
+        int end;
+        do
+        {
+            end = text.IndexOfAny(state.NextAllowed, state.Start);
+
+            // We couldn't find any of the allowed characters of which we needed to find one.
+            if (end == -1)
+            {
+                throw new FormatException($"Expected one of {string.Join(", ", state.NextAllowed)} in {text.Substring(state.Start)}.");
```
jskeet: See later comments about putting the bad text in quotes.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+            }
+            // P[n]Y[n]M[n]DT[n]H[n]M[n[.fraction]]S
```
jskeet, line 183: Are we happy with *any* aspect of this being negative independent of other aspects, e.g. `P5H-2.2S`?

Additionally, the text `P[n]Y[n]M[n]DT[n]H[n]M[n[.fraction]]S` in the comment earlier doesn't make it clear what can be included/omitted. Maybe some examples would help.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+    /// <summary>
+    /// Creates an interval from the given nanoseconds.
+    /// </summary>
+    public static Interval FromNanoseconds(BigInteger nanoseconds) => new Interval(0, 0, nanoseconds);
+
+    /// <summary>
+    /// Parses the string represantion in ISO8601 format of a duration returning its Interval equivalent.
+    /// </summary>
+    /// <param name="text">
+    /// The string representation of a duration in ISO8601 format: P[n]Y[n]M[n]DT[n]H[n]M[n[.fraction]]S.
```
jskeet: Nit: ISO8601 => ISO 8601

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+                    state.MayBeTerminal = true;
+                    state.IsTerminal = false;
+                    state.Years = ParseInt();
+                    state.NextAllowed = IntervalParsingState.AfterY;
+                    break;
+                case 'M' when !state.IsTime:
+                    state.MayBeTerminal = true;
+                    state.IsTerminal = false;
+                    state.Months = ParseInt();
+                    state.NextAllowed = IntervalParsingState.AfterMonth;
```
jskeet: Nit: let's rename this to AfterMonths to make it clear that it's a number of months, not a month number within a year. (So 25 months is fine, for example.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        int totalMonths =
+            state.Years * MonthsInAYear
+            + state.Months;
+
+        BigInteger totalNanoseconds =
+            state.Hours * (BigInteger)NanosecondsInAnHour
+            + state.Minutes * (BigInteger)NanosecondsInAMinute
+            + SecondsToNanoseconds(state.Seconds);
```
jskeet: This will throw ArgumentOutOfRangeException rather than FormatException if more than 9 decimal places have been specified.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+    public override int GetHashCode() =>
+        GaxEqualityHelpers.CombineHashCodes(Days.GetHashCode(), Months.GetHashCode(), Nanoseconds.GetHashCode());
+
+    private static BigInteger SecondsToNanoseconds(decimal seconds)
+    {
+        // We only accept 9 decimal places.
+        decimal rounded = Math.Round(seconds, 9);
+        if (seconds != rounded)
+        {
+            throw new ArgumentOutOfRangeException($"Interval has a nanosecond precision, only 9 decimal places are allowed for seconds.");
```
jskeet: Possibly just ArgumentException rather than ArgumentOutOfRangeException?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+
+    /// <inheritdocs/>
+    public override int GetHashCode() =>
+        GaxEqualityHelpers.CombineHashCodes(Days.GetHashCode(), Months.GetHashCode(), Nanoseconds.GetHashCode());
+
+    private static BigInteger SecondsToNanoseconds(decimal seconds)
+    {
+        // We only accept 9 decimal places.
+        decimal rounded = Math.Round(seconds, 9);
+        if (seconds != rounded)
```
jskeet: This will allow more than 9 decimal places to be specified, so long as they're zero. Do we want that? (I think it might give the wrong impression.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
jskeet: And the "new char[]" part is redundant - just `private static readonly char[] P = { 'P' };` should be fine, for example.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+    /// Returns the ISO8601 string representation of this Interval.
+    /// </summary>
+    public override string ToString()
+    {
+        if (Equals(Zero))
+        {
+            return "P0Y";
+        }
+
+        int years = Months / MonthsInAYear ;
```
jskeet: Nit: space before semicolon.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
```diff
@@ -0,0 +1,229 @@
...
+
+namespace Google.Cloud.Spanner.V1.Tests
+{
+    public class IntervalTests
+    {
+        private static readonly BigInteger NanonosecondsInASecond = (BigInteger)1_000_000_000;
+        private static readonly BigInteger NanonosecondsInAMillisecond = (BigInteger)1_000_000;
+        private static readonly BigInteger NanonosecondsInAMicrosecond = (BigInteger)1_000;
+
+        [Theory]
```
jskeet, line 27: Tests for FromXyz out of bounds? (I think they're only covered by parsing at the moment.)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
```diff
@@ -0,0 +1,229 @@
...
+        [InlineData("PT-316224000000S", "PT-87840000H")]
+        [InlineData("PT0.999999999S", "PT0.999999999S")]
+        [InlineData("PT0.000000009S", "PT0.000000009S")]
+        public void ParseRoundTrip(string intervalString, string expectedString)
+        {
+            Interval interval = Interval.Parse(intervalString);
+            Assert.Equal(expectedString, interval.ToString());
+        }
+
+        [Theory]
```
jskeet, line 289: Tests for more than 9 decimal places in seconds?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -0,0 +1,381 @@
...
+    /// </summary>
+    public static Interval FromNanoseconds(BigInteger nanoseconds) => new Interval(0, 0, nanoseconds);
+
+    /// <summary>
+    /// Parses the string represantion in ISO8601 format of a duration returning its Interval equivalent.
+    /// </summary>
+    /// <param name="text">
+    /// The string representation of a duration in ISO8601 format: P[n]Y[n]M[n]DT[n]H[n]M[n[.fraction]]S.
+    /// </param>
+    public static Interval Parse(string text)
```
purva9413, line 154: Curious to know if we have already thought about using a regex and named groups here to parse for the various time and period values? Are there some cases which were difficult to cover with a regex that made us go for this approach of explicit parsing?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Yes, done.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done, here and everywhere

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Yes, regex ended up being harder to fine tune and understand. It got quite convoluted. Past discussion:

- https://github.com/googleapis/google-cloud-dotnet/pull/14254#discussion_r1974586638
- https://github.com/googleapis/google-cloud-dotnet/pull/14254#discussion_r1982811716

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Removed the class.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Yes, signs are individual for each of the parts. I've expanded the documentation and included some examples.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done here and everywhere.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done just catch. I'll look into restructuring with a later PR.

But if I understand you correctly, I'd worry about the stacktrace of the exception that'd we'd be able to throw in `Parse`.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done, I've removed the class and these are now private static properties with setters only.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Removed these when removing the class.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: If the years, months, etc. are out of range (min/max) we throw ArgumentOutOfRange. I understand precission is not exactly the same as range, but close enough to use the same exception?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Yes, see my note below.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Yes you are right. I think it's fine for when the `FromSeconds(decimal seconds) ` factory method is used: for the purpose of Interval `FromSeconds(0.123456789)` should be equivalent to `FromSeconds(0.123456789000000000)`.

But it's not the same when parsing. I'll check it on parsing, but then, that brings me back to the type of the exception. For `"PT0.123456789123456789"` I still like using ArgumentOutOfRangeException but not for `"PT0.123456789000000"`, that's definetely a format exception. I will try to distinguish the cases in parse.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: I don't think we need to. I've added tests like FromSeconds_Short, FromSeconds_Int, ... and it seems fine. See if you can think of tricky value to test with.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Ok done, we now throw format exception for extra zeroes and argument out of range exception for extra non-zeroes. I think the distictions is worthwhile but happy to reconsider.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -94,136 +94,203 @@ private Interval(int months, int days, BigInteger nanoseconds)
...
+            years * MonthsPerYear
+            + months;
 
         BigInteger totalNanoseconds =
-            state.Hours * (BigInteger)NanosecondsInAnHour
-            + state.Minutes * (BigInteger)NanosecondsInAMinute
-            + SecondsToNanoseconds(state.Seconds);
+            hours * (BigInteger)NanosecondsPerHour
+            + minutes * (BigInteger)NanosecondsPerMinute
+            + SecondsToNanoseconds(seconds, validatePreccion: false); // We already validated precission when parsing, to fail sooner.
```
jskeet: Nit: precission => precision

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
```diff
@@ -320,62 +387,18 @@ obj is Interval other
     public override int GetHashCode() =>
         GaxEqualityHelpers.CombineHashCodes(Days.GetHashCode(), Months.GetHashCode(), Nanoseconds.GetHashCode());
 
-    private static BigInteger SecondsToNanoseconds(decimal seconds)
+    private static BigInteger SecondsToNanoseconds(decimal seconds, bool validatePreccion)
```
jskeet: validatePreccion => validatePrecision

But given that this is only called from two places, I'm tempted to suggest that you just inline the validation into FromSeconds.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
```diff
@@ -66,14 +112,45 @@ public void FromDays(int days)
...
+        public void FromSeconds_Decimal(decimal seconds)
+        {
+            Interval interval = Interval.FromSeconds(seconds);
+            Assert.Equal(new BigInteger(seconds * (long) NanonosecondsInASecond), interval.Nanoseconds);
+            Assert.Equal(0, interval.Months);
+            Assert.Equal(0, interval.Days);
+        }
+
+        [Theory]
+        [InlineData(316_224_000_001)]
```
jskeet: I suspect all of these should be specified as strings that are parsed with BigDecimal. (It's possible that xUnit will do that itself.) Otherwise we're ending with binary floating point values for some and integers for others.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
```diff
@@ -66,14 +112,45 @@ public void FromDays(int days)
...
+        [InlineData(316_224_000_000.1)]
+        [InlineData(0.123456789123456789)]
```
jskeet: Change this to just 0.1234567891 so that we check that the valid/invalid precision boundary is correct.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerDbTypeTests.cs
```diff
@@ -97,6 +98,7 @@ public static IEnumerable<object[]> GetSpannerStringConversions()
             yield return new object[] { "  TIMESTAMP  ", SpannerDbType.Timestamp };
             yield return new object[] { "  NUMERIC  ", SpannerDbType.Numeric };
             yield return new object[] { "  NUMERIC{PG}  ", SpannerDbType.PgNumeric };
+            yield return new object[] { " INTERVAL ", SpannerDbType.Interval };
```
jskeet: Is there a reason for the uneven spacing for some of these types?

jskeet (commented):
Just a few nits, but I think changing all the tests that expect a decimal to accept a string is fairly important.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/Interval.cs
amanda-tarafa: Yep, inlining sunds more straight forwards now. Done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/V1/IntervalTests.cs
amanda-tarafa: Oh yes, you are right, and we can't even do `239_238_293_898_239M` because that's not "constant enough" to be an attribute argument here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerDbTypeTests.cs
amanda-tarafa: Not that I can think off, fixing... I'm leaving one prefix space and two suffix spaces.

amanda-tarafa:
Thanks, I've rebased in two commits and I'll merge on green.

PR #14852, opened 2025-06-05
fix: update AR exit gate to use the correct project for images

ldetmer:


jskeet (approved):
Not sure if we'll be able to push to this project, but I'm very happy for us to test it.

PR #14857, opened 2025-06-06
fix: Forget ordering keys that have no pending messages.

### LINKED ISSUE #14675: Pub/Sub KeyState memory leak
#### Environment details

  - OS: Linux (Alpine)
  - .NET version: 7
  - Package name and version: `Google.Cloud.PubSub.V1` version 3.9.1 (also applies to all subsequent versions)

#### Use-case

We're routing chat messages so are setting the ordering key to a GUID chat identifier, making sure the messages won't come in out of order. Each unique conversation has its own chat identifier.
A typical conversation doesn't contain many messages going back and forth, so relatively speaking we are creating a lot of unique ordering keys for a small amount of messages.

#### Observations

Image

Continuously growing memory usage, resulting in `OutOfMemoryException`s and the program restarting.

A memory profile of the .NET process shows a large amount of `KeyState` objects:

Image

#### Cause

The [`_keyedState` dictionary](https://github.com/googleapis/google-cloud-dotnet/blob/93d932d18ebcc9e776dd14693e5f4e5ee60b6db6/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientImpl.cs#L153) grows with each unique conversation we're handling as a new `KeyState` object is [being added to it](https://github.com/googleapis/google-cloud-dotnet/blob/93d932d18ebcc9e776dd14693e5f4e5ee60b6db6/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientImpl.cs#L229-L231). Only when the `KeyState` encountered an error [is it cleaned up](https://github.com/googleapis/google-cloud-dotnet/blob/93d932d18ebcc9e776dd14693e5f4e5ee60b6db6/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientImpl.cs#L262-L266).

The used memory is never freed as the `PublisherClient` is registered as a singleton, as recommended in [these docs](https://cloud.google.com/dotnet/docs/reference/Google.Cloud.PubSub.V1/latest#coding-considerations) and there's no other clean up function present.

#### Steps to reproduce

  1. Configure a Pub/Sub push subscription with ordering enabled
  2. Create a basic .NET app registering a singleton`PublisherClient`
  3. Use the `PublisherClient` to send messages with different ordering keys, resulting in lots of `KeyState` objects being created
  4. Reach an `OutOfMemoryException` given enough throughput and time

#### Additional context

I've connected with Google Support via our partner DoiT and got confirmation that "a cleaning mechanism is under consideration". As the Google Support ticket is now closed I'd like to keep track of the progress via this issue. It also serves as a way to discuss potential solutions. Let me know if another way of handling this is preferred.

amanda-tarafa:
Fixes #14675

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientImpl.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientImpl.cs
@@ -187,6 +187,22 @@ public sealed class PublisherClientImpl : PublisherClient
     private readonly CancellationTokenSource _hardStopCts;
     private readonly TaskCompletionSource<int> _shutdownTcs;
 
+    /// <summary>
+    /// The number of ordering keys that are being currently tracked by the publisher,
+    /// which may include the empty string ordering key used for unordered messages.
+    /// For testing purposes only.
+    /// </summary>
+    internal int PendingKeysCount
+    {
+        get
+        {
+            lock(_lock)
+            {
+                return _keyedState.Count;
+            }
+        }
+    }
+
     /// <inheritdoc/>
     public override TopicName TopicName { get; }
 
@@ -534,6 +550,17 @@ public sealed class PublisherClientImpl : PublisherClient
                 }
                 // Trigger a new send, if there is any more data ready to send.
                 TriggerSend();
+
+                // For ordered messages we store the state on a dictionary by the ordering key.
+                // Ordering keys are usually short lived, so we remove states from the dictionary that have no pending messages,
+                // to avoid unbounded memory consumption.
+                // If messages for a removed key are published again, the key and its state will be added back to the dictionary,
+                // as if it were a new key.
+                // Note that if unordered messages have been sent the state dictionary will forever contain the entry with empty string as a key.
+                if (state.HasOrderingKey && state.State == OrderingKeyState.Normal && batches.Count == 0)
+                {
+                    _keyedState.Remove(state.OrderingKey);
+                }
             }
             postLockAction();
         }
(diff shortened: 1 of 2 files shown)


amanda-tarafa:
@Martijnvos FYI.

amanda-tarafa:
Failures are unrelated to this PR and because we need to update the target of the testing projects, which I'm doing right now. I'll then rebase.

cy-yun (approved):

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientImpl.cs
```diff
@@ -534,6 +550,17 @@ async Task Send()
                 }
                 // Trigger a new send, if there is any more data ready to send.
                 TriggerSend();
+
+                // For ordered messages we store the state on a dictionary by the ordering key.
+                // Ordering keys are usually short lived, so we remove states from the dictionary that have no pending messages,
+                // to avoid unbounded memory consumption.
+                // If messages for a removed key are published again, they key and its state will be added back to the dictionary,
+                // as if it were a new key.
+                // Note that if unordered messages have been sent the state dictionary will forever containg the entry with empty string as a key.
```
Martijnvos: Nit: typo in `containg`

Martijnvos (commented):
Thanks for these changes! Logically they makes sense to me.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/PublisherClientImpl.cs
amanda-tarafa: Thanks! Fixed.

amanda-tarafa:
Typos fixed. The PR has been rebased on the latests .NET 8 related changes and checks are now green as expected.

@jskeet I'll leave open till tomorrow in case you want to take a look. Thanks.

amanda-tarafa:
Thanks, yes, I did check all things very carefully as well. Integration/unit tests, of which we have a few that send messages in parallel in different combinations, etc. are all green. I believe this should be fine.

PR #14860, opened 2025-06-09
tools: First pass at building and releasing non-google-cloud-dotnet

jskeet:


tools/Google.Cloud.Tools.ReleaseManager/ContainerCommand.cs
```diff
@@ -66,6 +67,10 @@ public int Execute(string[] args)
 
         try
         {
+            if (options.RepoRoot is string repoRoot && !File.Exists(Path.Combine(repoRoot, ApiCatalog.PathInRepository)))
+            {
+                return NonGoogleCloudDotnetCommands.Execute(args[0], options);
+            }
```
amanda-tarafa, line 73: This is fine. But maybe at some point, in the per repo Librarian config (pipeline state maybe) you have a list of repo supported commands?

amanda-tarafa (approved):

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommand.cs
jskeet: Possibly. I think it's going to get more complex if we consider generating google-cloudevents-dotnet (with a different API repo!), and possibly even the common protos in GAX. Worth bearing in mind, certainly.

PR #14862, opened 2025-06-09
fix: Avoid reaching Firestore database operations per minute quota.

amanda-tarafa:
Fixes b/423643728

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.CleanTestData/Program.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.CleanTestData/Program.cs
@@ -50,6 +50,9 @@ internal class Program
             {
                 await adminClient.DeleteDatabaseAsync(dbName);
                 Console.WriteLine($"Success deleting {dbName.DatabaseId}");
+                // Firestore has a default quota of 60 database operations per minute.
+                // We wait 1 second in between database deletes.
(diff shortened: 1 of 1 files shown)


apis/Google.Cloud.Firestore/Google.Cloud.Firestore.CleanTestData/Program.cs
```diff
@@ -50,6 +50,9 @@ async Task DeleteDatabaseAsync(DatabaseName dbName)
             {
                 await adminClient.DeleteDatabaseAsync(dbName);
                 Console.WriteLine($"Success deleting {dbName.DatabaseId}");
+                // Firestore has a default quota of 60 database operations per minute.
+                // We wait 1 second in between database deletes.
+                await Task.Delay(1000);
```
purva9413, line 55: Is it possible we have multiple tests running which might be triggering other db operations? If yes, we might want to increase this to more than a second to account for the database operations apart from the delete operation (like inserts etc). Feel free to resolve and get this in if that is not the case. Thanks!

amanda-tarafa: We run data clean up at times where nothing else in runing, including during a weekend. If we were running togethter, it would be worst than the quota, clean-up could delete resources that were actively being used by tests.

PR #14867, opened 2025-06-10
tools: First pass at tests for CleanCommand

jskeet:
We'll need to expand the framework over time, for sure, but this was the easiest command to start with.

Towards https://github.com/googleapis/librarian/issues/304

We also need to run this regularly, of course. Thoughts on that welcome - just make runintegrationtests.sh do it?


amanda-tarafa (approved):
LGTM.

As for when to run this, let's setup a new job daily_container_tests or daily_tools_tests just so we don't get things mixed. I can help with that.

PR #14887, opened 2025-06-16
feat(BigQuery): Add JobConfigurationModifier field to JobCreationOptions

cy-yun:
https://buganizer.corp.google.com/issues/414634563

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
@@ -30,6 +30,7 @@ namespace Google.Cloud.BigQuery.V2.Tests
     {
         // Default location for tests that need one.
         private const string DefaultLocation = "default-location";
+        private const string Reservation = "Reservation";
 
         [Fact]
         public void PrettyPrintEnabledOnClientIsPropagatedToRequest()
@@ -1037,6 +1038,13 @@ namespace Google.Cloud.BigQuery.V2.Tests
             var job = CreateJobWithSampleConfiguration(new UploadCsvOptions { JobId = "custom_job_id" });
             Assert.Equal("project", job.JobReference.ProjectId);
             Assert.Equal("custom_job_id", job.JobReference.JobId);
+        }
+
+        [Fact]
+        public void CreateJob_JobConfigurationModifier()
+        {
+            var job = CreateJobWithSampleConfiguration(new UploadCsvOptions { JobConfigurationModifier = jobConfiguration => jobConfiguration.Reservation = Reservation });
+            Assert.Equal("Reservation", job.Configuration.Reservation);
         }
 
         [Fact]
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.JobCrud.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.JobCrud.cs
@@ -204,7 +204,6 @@ namespace Google.Cloud.BigQuery.V2
             string projectId = options?.ProjectId ?? ProjectId;
             string jobId = options?.JobId;
             string jobIdPrefix = options?.JobIdPrefix;
-            IDictionary<string, string> jobLabels = options?.Labels;
             if (jobId != null && jobIdPrefix != null)
             {
                 throw new ArgumentException("Only one of JobId or JobIdPrefix can be specified for a single operation.");
@@ -215,10 +214,8 @@ namespace Google.Cloud.BigQuery.V2
                 jobId = prefix + Guid.NewGuid().ToString().Replace("-", "_");
             }
             string location = options?.JobLocation ?? DefaultLocation;
-            if (jobLabels != null && jobLabels.Count > 0)
-            {
-                configuration.Labels = jobLabels;
-            }
+
+            options?.ModifyJobConfiguration(configuration);
             return new Job { Configuration = configuration, JobReference = GetJobReference(projectId, jobId, location) };
         }
 
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.Queries.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.Queries.cs
@@ -191,7 +191,7 @@ namespace Google.Cloud.BigQuery.V2
                 GaxPreconditions.CheckArgument(parameters.All(p => !string.IsNullOrEmpty(p.Name)), nameof(parameters),
                     $"When using a parameter mode of '{nameof(BigQueryParameterMode.Named)}', all parameters must have names");
             }
-            return CreateInsertJobRequest(new JobConfiguration { Query = jobConfigurationQuery, DryRun = options?.DryRun }, options);
+            return CreateInsertJobRequest(new JobConfiguration { Query = jobConfigurationQuery }, options);
         }
 
         private TabledataResource.ListRequest CreateListRequest(TableReference tableReference, ListRowsOptions options, TableSchema schema)
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.csproj
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.csproj
@@ -10,7 +10,7 @@
   <ItemGroup>
     <PackageReference Include="ConfigureAwaitChecker.Analyzer" PrivateAssets="All" />
     <PackageReference Include="Google.Api.Gax.Rest" />
-    <PackageReference Include="Google.Apis.Bigquery.v2" VersionOverride="[1.68.0.3400, 2.0.0.0)" />
+    <PackageReference Include="Google.Apis.Bigquery.v2" VersionOverride="[1.69.0.3783, 2.0.0.0)" />
   </ItemGroup>
   <ItemGroup>
     <Compile Update="BigQueryClient.*.cs">
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Apis.Bigquery.v2.Data;
+using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Google.Cloud.BigQuery.V2
 {
@@ -59,5 +62,28 @@ namespace Google.Cloud.BigQuery.V2
         /// Label keys must start with a letter.
         /// </summary>
         public IDictionary<string, string> Labels { get; set; }
+
+        /// <summary>
+        /// Optional action to perform after preparing the request. If this property is non-null,
+        /// the <see cref="JobConfiguration"/> used for a request will be passed to the delegate
+        /// before the request is executed. This allows for fine-grained modifications which aren't
+        /// otherwise directly supported by the properties in this options type or derived option types.
+        /// </summary>
+        /// <remarks>
+        /// Prefer the properties on this type over this modifier to prepare the request.
+        /// Only use this modifier to configure aspects for which there are no properties available.
+        /// This modifier is applied to the request after all properties on this type have been applied.
+        /// The delegate is only called once per operation, even if the request is automatically retried.
+        /// </remarks>
+        public Action<JobConfiguration> JobConfigurationModifier { get; set; }
+
+        internal virtual void ModifyJobConfiguration(JobConfiguration jobConfiguration)
+        {
+            if (Labels is not null && Labels.Count > 0)
+            {
+                jobConfiguration.Labels = Labels;
+            }
+            JobConfigurationModifier?.Invoke(jobConfiguration);
+        }
     }
 }
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/QueryOptions.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/QueryOptions.cs
@@ -15,6 +15,7 @@
 using Google.Apis.Bigquery.v2.Data;
 using System;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 namespace Google.Cloud.BigQuery.V2
 {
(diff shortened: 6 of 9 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
```diff
@@ -59,5 +61,25 @@ public abstract class JobCreationOptions
         /// Label keys must start with a letter.
         /// </summary>
         public IDictionary<string, string> Labels { get; set; }
+
+        /// <summary>
+        /// Optional action to perform after preparing the request. If this property is non-null,
+        /// the <see cref="JobConfiguration"/> used for a request will be passed to the delegate
+        /// before the request is executed. This allows for fine-grained modifications which aren't
+        /// otherwise directly supported by the properties in this options type.
```
amanda-tarafa:
```suggestion
        /// otherwise directly supported by the properties in this options type or derived option types.
```

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
```diff
@@ -59,5 +61,25 @@ public abstract class JobCreationOptions
...
+        /// This modifier is applied to the request after all properties on this type have been applied.
+        /// The delegate is only called once per operation, even if the request is automatically retried.
+        /// </remarks>
+        public Action<JobConfiguration> JobConfigurationModifier { get; set; }
+
+        /// <summary>
+        /// Execute the JobConfigurationModifier, if it is not null.
+        /// </summary>
+        /// <param name="jobConfiguration"></param>
+        public void ModifyJobConfiguration(JobConfiguration jobConfiguration) => JobConfigurationModifier?.Invoke(jobConfiguration);
```
amanda-tarafa:
```suggestion
        internal void ModifyJobConfiguration(JobConfiguration jobConfiguration) => JobConfigurationModifier?.Invoke(jobConfiguration);
```

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.csproj
```diff
@@ -10,7 +10,7 @@
   <ItemGroup>
     <PackageReference Include="ConfigureAwaitChecker.Analyzer" PrivateAssets="All" />
     <PackageReference Include="Google.Api.Gax.Rest" />
-    <PackageReference Include="Google.Apis.Bigquery.v2" VersionOverride="[1.68.0.3400, 2.0.0.0)" />
+    <PackageReference Include="Google.Apis.Bigquery.v2" VersionOverride="[1.69.0.3783, 2.0.0.0)" />
```
amanda-tarafa, line 13: You need to change [apis.json](https://github.com/googleapis/google-cloud-dotnet/blob/f745828670e87492e5331f134d403006de72810a/generator-input/apis.json#L1141) and then execute [generateprojects.sh](https://github.com/googleapis/google-cloud-dotnet/blob/main/generateprojects.sh)

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.Queries.cs
```diff
@@ -191,7 +191,13 @@ private JobsResource.InsertRequest CreateInsertQueryJobRequest(string sql, IEnum
                 GaxPreconditions.CheckArgument(parameters.All(p => !string.IsNullOrEmpty(p.Name)), nameof(parameters),
                     $"When using a parameter mode of '{nameof(BigQueryParameterMode.Named)}', all parameters must have names");
             }
-            return CreateInsertJobRequest(new JobConfiguration { Query = jobConfigurationQuery, DryRun = options?.DryRun }, options);
+            var jobConfiguration = new JobConfiguration
+            {
+                Query = jobConfigurationQuery,
+                DryRun = options?.DryRun
```
amanda-tarafa: I think we can now move this change to ModifyJobConfiguration of the descendants of JobConfiguration that have the DryRun property. You can make the ModifyJobConfiguration method virtual and override it on the relevant descendants.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.JobCrud.cs
```diff
@@ -288,8 +292,9 @@ private InsertRequest CreateLoadJobRequest(IEnumerable<string> sourceUris, Table
 
             var load = new JobConfigurationLoad { SourceUris = sourceList, DestinationTable = destination, Schema = schema };
             options?.ModifyRequest(load);
-            var request = CreateInsertJobRequest(new JobConfiguration { Load = load }, options);
-            return request;
+            var jobConfiguration = new JobConfiguration { Load = load };
+            options?.ModifyJobConfiguration(jobConfiguration);
+            return CreateInsertJobRequest(jobConfiguration, options);
         }
```
amanda-tarafa: You missed one in what is now line 320.

amanda-tarafa: And you missed a couple on BigQueryClientImp.InsertData.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
```diff
@@ -135,8 +137,9 @@ public void GetJob_JobLabels()
         {
             var client = BigQueryClient.Create(_fixture.ProjectId);
             var table = client.GetTable(_fixture.ProjectId, _fixture.DatasetId, _fixture.HighScoreTableId);
-            var options = new QueryOptions { Labels = JobLabels };
+            var options = new QueryOptions { Labels = JobLabels, JobConfigurationModifier = jobConfiguration => jobConfiguration.Reservation = Reservation };
             var jobToFind = client.CreateQueryJob($"SELECT * FROM {table}", null, options);
+            Assert.Equal(Reservation, jobToFind.Resource.Configuration.Reservation);
```
amanda-tarafa: Is this passing? Integration tests are not run on presubmits and the format for the reservation value does not match what's specified on documentation.

Also, I think the change in this commit needs unit tests, both for testing that ModifyJobConfiguratioin is called everywhere where a new JobConfiguration is created, and for testing the ModifyJobConfiguration itself will call the callback when set.

With that, we don't need integration tests.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/UpdateDatasetOptions.cs
```diff
@@ -21,8 +22,18 @@ namespace Google.Cloud.BigQuery.V2
     /// </summary>
     public sealed class UpdateDatasetOptions
     {
-        internal void ModifyRequest(DatasetsResource.UpdateRequest request)
+
+        /// <summary>
+        /// Optional action to perform after preparing the request. If this property is non-null,
+        /// the <see cref="UpdateRequest"/> used for a request will be passed to the delegate
+        /// before the request is executed. This allows for fine-grained modifications which aren't
+        /// otherwise directly supported by the properties in this options type.
```
amanda-tarafa: I would drop the "... by the properties..." bit on the account that there are no other properties on this class.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
```diff
@@ -114,13 +116,27 @@ public void GetDataset()
         {
             var projectId = "project";
             var datasetId = "dataset";
+            var accessPolicyVersion = 1;
```
amanda-tarafa: I think it'd be clearer if we have specific tests for checking that the callback is used, like `GetDataset_WithModifiedRequest` and `UpdateDataset_WithModifiedRequest`.

Also, add unit tests for the option types you modified.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/PatchDatasetOptions.cs
```diff
@@ -21,8 +22,14 @@ namespace Google.Cloud.BigQuery.V2
...
     public sealed class PatchDatasetOptions
     {
-        internal void ModifyRequest(DatasetsResource.PatchRequest request)
-        {
-        }
+        /// <summary>
+        /// Optional action to perform after preparing the request. If this property is non-null,
+        /// the <see cref="PatchRequest"/> used for a request will be passed to the delegate
+        /// before the request is executed. This allows for fine-grained modifications which aren't
+        /// otherwise directly supported by the properties in this options type.
```
amanda-tarafa: Drop the `... by the properties in...` bit

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/CreateDatasetOptions.cs
```diff
@@ -22,7 +23,14 @@ namespace Google.Cloud.BigQuery.V2
     /// </summary>
     public sealed class CreateDatasetOptions
     {
-        internal void ModifyRequest(InsertRequest request)
-        { }
+        /// <summary>
+        /// Optional action to perform after preparing the request. If this property is non-null,
+        /// the <see cref="InsertRequest"/> used for a request will be passed to the delegate
+        /// before the request is executed. This allows for fine-grained modifications which aren't
+        /// otherwise directly supported by the properties in this options type.
```
amanda-tarafa: And same here, there are no other properties in this class.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
```diff
@@ -100,13 +100,21 @@ public void CreateDataset()
...
```
amanda-tarafa: Similar requests regarding tests for this commit than for the preious commit.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.IntegrationTests/JobsTest.cs
cy-yun: What do you mean "the format for the reservation value does not match what's specified on documentation"? When I look at the definition for reservation, it is a string.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.JobCrud.cs
cy-yun: I removed all of these and moved the ModifyJobConfiguration call to Create job method in line 200. I traced the references to make sure that all implementations of the JobCreationOptions class passes through the CreateJob method.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
```diff
@@ -59,5 +61,25 @@ public abstract class JobCreationOptions
...
+        /// Only use this modifier to configure aspects for which there are no properties available.
+        /// This modifier is applied to the request after all properties on this type have been applied.
+        /// The delegate is only called once per operation, even if the request is automatically retried.
+        /// </remarks>
+        public Action<JobConfiguration> JobConfigurationModifier { get; set; }
+
+        /// <summary>
+        /// Execute the JobConfigurationModifier, if it is not null.
+        /// </summary>
+        /// <param name="jobConfiguration"></param>
```
amanda-tarafa: We don't need docs on internal methods. But if we are going to have them let's not make it about the modifier only.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/JobCreationOptions.cs
```diff
@@ -59,5 +61,25 @@ public abstract class JobCreationOptions
...
+        internal virtual void ModifyJobConfiguration(JobConfiguration jobConfiguration) => JobConfigurationModifier?.Invoke(jobConfiguration);
```
amanda-tarafa: This method can now apply the labels and we can remove that code from elsewhere.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/QueryOptions.cs
```diff
@@ -146,6 +147,12 @@ public sealed class QueryOptions : JobCreationOptions
         /// </remarks>
         public Action<JobConfigurationQuery> ConfigurationModifier { get; set; }
 
+        internal override void ModifyJobConfiguration(JobConfiguration jobConfiguration)
+        {
+            jobConfiguration.DryRun = DryRun;
```
amanda-tarafa: You only set it on the configuration if it's set, same as we do for all others in ModifyRequest.
```suggestion
            if (DryRun is not null)
            {
                jobConfiguration.DryRun = DryRun;
            }
```

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/QueryOptions.cs
```diff
@@ -146,6 +147,12 @@ public sealed class QueryOptions : JobCreationOptions
...
+            JobConfigurationModifier?.Invoke(jobConfiguration);
```
amanda-tarafa: So that we guarantee we always do whatever the parent is doing.

```suggestion
            base.ModifyJobConfiguration(jobConfiguration);
```

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2/BigQueryClientImpl.JobCrud.cs
```diff
@@ -201,6 +201,7 @@ internal Job CreateJob(JobConfiguration configuration, JobCreationOptions option
         {
             GaxPreconditions.CheckNotNull(configuration, nameof(configuration));
 
+            options?.ModifyJobConfiguration(configuration);
```
amanda-tarafa: Apply labels in `ModifyJobConfiguration` instead of on this method (line 219 below). And move this call to where we are currently applying the labels.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/UploadParquetOptionsTest.cs
```diff
@@ -31,11 +31,15 @@ public void ModifyRequest()
...
                 DestinationEncryptionConfiguration = new EncryptionConfiguration { KmsKeyName = "projects/1/locations/us/keyRings/1/cryptoKeys/1" },
                 DestinationSchemaUpdateOptions = SchemaUpdateOption.AllowFieldAddition | SchemaUpdateOption.AllowFieldRelaxation,
-                ConfigurationModifier = options => options.AllowJaggedRows = true
+                ConfigurationModifier = options => options.AllowJaggedRows = true,
+                JobConfigurationModifier = config => config.Reservation = "Reservation"
             };
 
             JobConfigurationLoad config = new JobConfigurationLoad();
             options.ModifyConfiguration(config);
+            var jobConfiguration = new JobConfiguration();
```
amanda-tarafa: Let's keep this test for testing ModifyRequest and let's add a new one for testing ModifyJobConfiguration.

But, since we are adding this on the parent class we don't need to test it for all the options:

- We can create a FakeJobCreationOptions class that's not abstract and test that the action and the labels are applied with that.
- We need to keep the QueryOptions test because that has the DryRun part that we need to test explicitly. But there, split the ModifyRequest test from the ModifyJobConfiguration test.

amanda-tarafa (changes requested):
Generally good, a few change requests only, and mostly for simplifying further, both production code and tests.

PR #14902, opened 2025-06-20
tests: Add Docker-based tests for integration-test-library

jskeet:
It's unfortunate that we need quite a lot of real-repo files, and currently we have to specify those everywhere. We could potentially have some "named file collections" or similar, so you could just specify "copy this set to /repo" but that can come later.


tools/Google.Cloud.Tools.ReleaseManager.IntegrationTests/ContainerCommands/DockerTests/IntegrationTestLibrary-FailingTest/metadata.yaml
```diff
@@ -0,0 +1,35 @@
+command: integration-test-library
+args:
+- '--repo-root=/test/repo'
+- '--library-id=Google.Test.V2'
+commonFiles:
```
cy-yun, line 5: can you help me understand how the test interprets the source & target here?

amanda-tarafa (approved):

PR #14906, opened 2025-06-23
tests: Docker tests for Configure command

cy-yun:
https://github.com/googleapis/librarian/issues/307

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager.IntegrationTests/ContainerCommands/DockerCommandTest.cs
+++ b/tools/Google.Cloud.Tools.ReleaseManager.IntegrationTests/ContainerCommands/DockerCommandTest.cs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using Google.Cloud.ClientTesting;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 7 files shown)


tools/Google.Cloud.Tools.ReleaseManager.IntegrationTests/ContainerCommands/DockerCommandTest.cs
```diff
@@ -130,10 +135,11 @@ private class TestMetadata
         public FileExpectations FileExpectations { get; set; } = new();
     }
 
-    public class FileExpectations
+    private class FileExpectations
     {
         public List<string> Present { get; set; } = [];
         public List<string> Absent { get; set; } = [];
+        public List<ExpectedActual> AssertContentEquals { get; set; } = [];
```
jskeet: Rather than having this explicitly, can we have a convention of "if there's an `expected` folder under the test case, that should be expected files"?

tools/Google.Cloud.Tools.ReleaseManager.IntegrationTests/ContainerCommands/DockerCommandTest.cs
```diff
@@ -62,6 +61,17 @@ public void DockerTest(string directory)
...
         }
+        var expectedDirectory = Path.Combine(outputDirectory, "expected");
+        if (Directory.Exists(expectedDirectory))
+        {
+            foreach (var expectedFilePath in Directory.GetFiles(expectedDirectory, "*.*", SearchOption.AllDirectories))
+            {
+                var actualFilePath = Path.Combine(outputDirectory, Path.GetRelativePath(expectedDirectory, expectedFilePath));
+                var expectedContent = File.ReadAllText(expectedFilePath);
+                var actualContent = File.ReadAllText(actualFilePath);
+                Assert.Equal(expectedContent, actualContent);
```
jskeet, line 72: In a follow-up, I suspect we'll want to make the failure show what files were being compared - but that can definitely come later.

jskeet (dismissed):
This is great! Thanks so much :)

jskeet:
@cy-yun: Might it be worth removing the final commit and then merging, so we can have a fresh PR that's just for Configure?

PR #14911, opened 2025-06-25
tests: Docker test for PackageLibrary-Passing

cy-yun:
https://github.com/googleapis/librarian/issues/311

### DIFF
--- a/tools/Google.Cloud.Tools.ReleaseManager.IntegrationTests/ContainerCommands/DockerTests/PackageLibrary-Passing/metadata.yaml
+++ b/tools/Google.Cloud.Tools.ReleaseManager.IntegrationTests/ContainerCommands/DockerTests/PackageLibrary-Passing/metadata.yaml
@@ -1,0 +1,27 @@
+command: package-library
+args:
+- '--repo-root=/test/repo'
+- '--library-id=Google.Test.V1'
+- '--output=/test/output'
+commonFiles:
+  # This is the simplest way of creating an empty directory
+  - source: irrelevant-content
+    target: output/.ignored
+  - source: test-apis.json
+    target: repo/generator-input/apis.json
+  - source: Google.Test.V1
+    target: repo/apis/Google.Test.V1
+  - source: Google.Test.V1/Google.Test.V1.sln
+    target: repo/docs/Google.Cloud.DocTools.sln
+  - source: Google.Test.V1/Google.Test.V1/Google.Test.V1.csproj
+    target: repo/docs/Google.Test.V1/Google.Test.V1.csproj
+repoFiles:
+  - source: docs/builddocs.sh
+    target: repo/docs/builddocs.sh 
+  - source: toolversions.sh
+    target: repo/toolversions.sh
+fileExpectations:
+  present:
+  - output/Google.Test.V1.1.0.0.nupkg
+  - output/Google.Test.V1.1.0.0.nupkg-sbom.spdx.json
+  - output/package-owner.txt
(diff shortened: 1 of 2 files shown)


tools/Google.Cloud.Tools.ReleaseManager.IntegrationTests/ContainerCommands/DockerTests/PackageLibrary-Passing/metadata.yaml
```diff
@@ -0,0 +1,27 @@
...
+commonFiles:
+  # This is the simplest way of creating an empty directory
+  - source: irrelevant-content
+    target: output/.ignored
+  - source: test-apis.json
+    target: repo/generator-input/apis.json
+  - source: Google.Test.V1
+    target: repo/apis/Google.Test.V1
+  - source: Google.Test.V1/Google.Test.V1.sln
+    target: repo/docs/Google.Cloud.DocTools.sln
```
jskeet, line 15: I'd expect us to have to copy the code for the projects in this solution as well. I'm experimenting locally.

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/PackageLibraryCommand.cs
```diff
@@ -78,10 +80,17 @@ public int Execute(ContainerOptions options)
     private static void PackageDocumentation(RootLayout rootLayout, ApiMetadata api, string outputDirectory)
     {
         var docsDir = rootLayout.CreateDocsLayout(api.Id).OutputDirectory;
+        if (!Directory.Exists(docsDir))
```
jskeet, line 83: I think we'll always get docs, so we shouldn't have this conditionality - this is basically working round the test not being able to generate the docs.

What alarms me more is that it looks like we're not failing earlier if we fail to create the docs, which does sound like a bug. I'm looking into that.

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/PackageLibraryCommand.cs
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Apis.Auth.OAuth2;
```
jskeet, line 15: I don't think either of these new using directives is required.

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/PackageLibraryCommand.cs
```diff
@@ -78,10 +80,17 @@ public int Execute(ContainerOptions options)
...
         var docsDir = rootLayout.CreateDocsLayout(api.Id).OutputDirectory;
+        if (!Directory.Exists(docsDir))
+        {
+            return;
+        }
         // Note: a comment in uploaddocs.sh claims we don't generate documentation for all packages.
         // I believe this is no longer true.
         var siteDir = Path.Combine(docsDir, "site");
-        FixDocfxOutput(siteDir, api);
+        if (Directory.Exists(siteDir))
```
jskeet, line 90: I think this if statement is probably *not* needed - if we've got an output directory, we should definitely have a docs directory.

tools/Google.Cloud.Tools.ReleaseManager/ContainerCommands/PackageLibraryCommand.cs
jskeet: Ah - no, it's because we don't have a docs directory in this test, and `builddocs.sh` skips APIs with no docs directory.

That's probably fine, and so this change is good. But we should have two tests - one for an API with docs and one for an API without docs. I'm going to try that out and create my own branch (rather than amending your PR) - we can then decide what to do.

PR #14912, opened 2025-06-25
feat: Support for Spanner ReadLockMode options

purva9413:
Keeping the ReadLockMode getter `public`. Clients should be able to set the property via chaining through the `WithReadLockMode` method giving us better control over validation of the property.

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
@@ -607,6 +607,34 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
             }
         }
 
+        [Fact]
+        public async Task CustomTransactionOptions()
+        {
+            // This test currently tests ReadLockMode and IsolationLevel
+            // More options compatible with these options can be added to this same test
+            using (var connection = _fixture.GetConnection())
+            {
+                await connection.OpenAsync();
+                var options = SpannerTransactionCreationOptions.ReadWrite
+                    .WithReadLockMode(ReadLockMode.Pessimistic)
+                    .WithIsolationLevel(System.Data.IsolationLevel.Serializable);
+
+                using (var tx = await connection.BeginTransactionAsync(options, transactionOptions: null, cancellationToken: default))
+                {
+                    Assert.Equal(TransactionMode.ReadWrite, tx.Mode);
+
+                    var cmd = CreateSelectAllCommandForKey(connection);
+                    cmd.Transaction = tx;
+                    using (var reader = await cmd.ExecuteReaderAsync())
+                    {
+                        Assert.True(
+                            await reader.ReadAsync(),
+                            "no data should be here from yesterday!");
+                    }
+                }
+            }
+        }
+
         /// <summary>
         /// Simple extension of DefaultLogger that also keeps a reference to the last logged CommitResponse.
         /// </summary>
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
@@ -19,6 +19,8 @@ using static Google.Cloud.Spanner.V1.TransactionOptions.Types;
 using IsolationLevel = System.Data.IsolationLevel;
 using SpannerIsolationLevel = Google.Cloud.Spanner.V1.TransactionOptions.Types.IsolationLevel;
 
+using TransactionReadLockMode = Google.Cloud.Spanner.V1.TransactionOptions.Types.ReadWrite.Types.ReadLockMode;
+
 namespace Google.Cloud.Spanner.Data.Tests;
 public class SpannerTransactionCreationOptionsTests
 {
@@ -63,6 +65,7 @@ public class SpannerTransactionCreationOptionsTests
         Assert.False(readWrite.ExcludeFromChangeStreams);
         Assert.Equal(IsolationLevel.Unspecified, readWrite.IsolationLevel);
         Assert.Equal(new TransactionOptions { ReadWrite = new ReadWrite() }, readWrite.GetTransactionOptions());
+        Assert.Equal(TransactionReadLockMode.Unspecified, readWrite.ReadLockMode);
     }
 
     [Fact]
@@ -79,6 +82,7 @@ public class SpannerTransactionCreationOptionsTests
         Assert.False(partitionedDml.ExcludeFromChangeStreams);
         Assert.Equal(IsolationLevel.Unspecified, partitionedDml.IsolationLevel);
         Assert.Equal(new TransactionOptions { PartitionedDml = new PartitionedDml() }, partitionedDml.GetTransactionOptions());
+        Assert.Equal(TransactionReadLockMode.Unspecified, partitionedDml.ReadLockMode);
     }
 
     [Fact]
@@ -95,6 +99,7 @@ public class SpannerTransactionCreationOptionsTests
         Assert.False(readOnly.ExcludeFromChangeStreams);
         Assert.Equal(IsolationLevel.Unspecified, readOnly.IsolationLevel);
         Assert.Equal(TimestampBound.Strong.ToTransactionOptions(), readOnly.GetTransactionOptions());
+        Assert.Equal(TransactionReadLockMode.Unspecified, readOnly.ReadLockMode);
     }
 
     [Fact]
@@ -110,6 +115,7 @@ public class SpannerTransactionCreationOptionsTests
         Assert.False(options.ExcludeFromChangeStreams);
         Assert.Equal(IsolationLevel.Unspecified, options.IsolationLevel);
         Assert.Equal(TimestampBound.Strong.ToTransactionOptions(), options.GetTransactionOptions());
+        Assert.Equal(TransactionReadLockMode.Unspecified, options.ReadLockMode);
     }
 
     [Fact]
@@ -126,6 +132,7 @@ public class SpannerTransactionCreationOptionsTests
         Assert.False(options.ExcludeFromChangeStreams);
         Assert.Equal(IsolationLevel.Unspecified, options.IsolationLevel);
         Assert.Equal(timestampBound.ToTransactionOptions(), options.GetTransactionOptions());
+        Assert.Equal(TransactionReadLockMode.Unspecified, options.ReadLockMode);
     }
 
     [Fact]
@@ -147,6 +154,7 @@ public class SpannerTransactionCreationOptionsTests
         Assert.False(options.ExcludeFromChangeStreams);
         Assert.Equal(IsolationLevel.Unspecified, options.IsolationLevel);
         Assert.Null(options.GetTransactionOptions());
+        Assert.Equal(TransactionReadLockMode.Unspecified, options.ReadLockMode);
     }
 
     [Fact]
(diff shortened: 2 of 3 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -105,7 +107,15 @@ public sealed class SpannerTransactionCreationOptions
...
     public bool ExcludeFromChangeStreams { get; }
 
-    private SpannerTransactionCreationOptions(TimestampBound timestampBound, TransactionId transactionId, bool isDetached, bool isSingleUse, bool isPartitionedDml, bool excludeFromChangeStreams)
+
+    /// <summary>
+    /// The ReadLockMode should be set via chaining through <see cref="WithReadLockMode(TransactionReadLockMode)"/>.
+    /// This control allows us to validate the value via the constructor.
+    /// This is set to internal so that we can unit test it.
+    /// </summary>
+    internal TransactionReadLockMode ReadLockMode { get; }
```
purva9413: I will change this to a public getter instead of keeping it internal.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -16,6 +16,8 @@
 using Google.Cloud.Spanner.V1;
 using static Google.Cloud.Spanner.V1.TransactionOptions.Types;
```
amanda-tarafa: nit: Remove the blank line

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -105,7 +107,15 @@ public sealed class SpannerTransactionCreationOptions
     /// </remarks>
     public bool ExcludeFromChangeStreams { get; }
 
-    private SpannerTransactionCreationOptions(TimestampBound timestampBound, TransactionId transactionId, bool isDetached, bool isSingleUse, bool isPartitionedDml, bool excludeFromChangeStreams)
+
```
amanda-tarafa:
```suggestion
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -105,7 +107,15 @@ public sealed class SpannerTransactionCreationOptions
...
+    /// <summary>
```
amanda-tarafa, line 117: Update this documentation to explain what this field actually is, not how we are using it. Follow some of the other properties as an example.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -183,6 +200,11 @@ internal TransactionOptions GetTransactionOptions()
         if (options is not null)
         {
             options.ExcludeTxnFromChangeStreams = ExcludeFromChangeStreams;
+            if(options.ReadWrite is not null)
+            {
+                // Options only settable on ReadWrite transactions (excluding the partioned dml txn as they too have TransactionMode set to ReadWrite)
```
amanda-tarafa: I think we can remove this, it's more or less clear from the code that this is an option for ReadWrite options only.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -137,6 +147,11 @@ private SpannerTransactionCreationOptions(TimestampBound timestampBound, Transac
             nameof(excludeFromChangeStreams),
             "Only read-write and partioned DML transactions can be marked for change stream exclusion.");
         ExcludeFromChangeStreams = excludeFromChangeStreams;
+        GaxPreconditions.CheckArgument(
+            TransactionMode == TransactionMode.ReadWrite || readLockMode == TransactionReadLockMode.Unspecified,
```
amanda-tarafa: I think you also need to check here that it is not partitioned DML?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
```diff
@@ -150,6 +152,24 @@ public void WithIsDetached()
         Assert.False(options.IsDetached);
     }
 
+    [Fact]
+    public void WithReadLockMode()
```
amanda-tarafa: You need to check that the default is unspecified on a few of the tests above this line.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
```diff
@@ -150,6 +152,24 @@ public void WithIsDetached()
...
+        var options = SpannerTransactionCreationOptions.ReadWrite.WithReadLockMode(TransactionReadLockMode.Optimistic);
+        Assert.Equal(TransactionReadLockMode.Optimistic, options.ReadLockMode);
+        options = options.WithReadLockMode(TransactionReadLockMode.Pessimistic);
+        Assert.Equal(TransactionReadLockMode.Pessimistic, options.ReadLockMode);
+    }
+
+    [Fact]
+    public void WithReadLockMode_ReadOnly()
+    {
+        var options = SpannerTransactionCreationOptions.ReadOnly.WithReadLockMode(TransactionReadLockMode.Unspecified);
```
amanda-tarafa: And when you start checking for partitioned DML, you also need to add a test for that here.

amanda-tarafa (changes requested):

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
purva9413: I think having this explicit comment there clarifies why only checking `TransactionMode` set to `ReadWrite` is not sufficient in this case. For me, it wasn't obviously clear that Partioned DML is a 'special' kind of read-write txn, and keeping this comment there would clarify that for anyone else. Lmk your thoughts on that.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
purva9413: Good catch! Done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -105,7 +106,14 @@ public sealed class SpannerTransactionCreationOptions
     /// </remarks>
     public bool ExcludeFromChangeStreams { get; }
 
-    private SpannerTransactionCreationOptions(TimestampBound timestampBound, TransactionId transactionId, bool isDetached, bool isSingleUse, bool isPartitionedDml, bool excludeFromChangeStreams)
+
+    /// <summary>
+    /// Whether locks for reads in a read-write transaction will be acquired on read or on commit.
```
amanda-tarafa: tiny nit, but just in case they add more read lock modes:

```suggestion
    /// Read lock modes for the transaction.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
amanda-tarafa: I think it's fine if you add this comment on the constructor, where you are checking that the options are compatible with the transaction type.

But here, we assume that things are already compatible (for this and other options). In particular it was confusing to read about Partitioned DML on this comment but then see no actual check about it right after (because we are checking on the constructor.)

I think it's aslo fine to add a comment at the top of this method saying something like: "At this point we have checked that all options are compatible so we don't have to check here".

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
```diff
@@ -150,6 +158,33 @@ public void WithIsDetached()
...
+        Assert.NotNull(options);
```
amanda-tarafa:
```suggestion
        var oldOptions = SpannerTransactionCreationOptions.ReadOnly;
        var newOptions = oldOptions..WithReadLockMode(TransactionReadLockMode.Unspecified);
        
        Assert.Same(oldOptions, newOptions);
```

And the same for the one below.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
```diff
@@ -150,6 +158,33 @@ public void WithIsDetached()
...
+
+        Assert.Throws<ArgumentException>(() => options.WithReadLockMode(TransactionReadLockMode.Optimistic));
+    }
+
+    [Fact]
+    public void WithReadLockMode_PartitionedDml()
```
amanda-tarafa, line 189: Add a theory for ReadWrite non-partitioned DML that accepts and succeeds for every value.

amanda-tarafa (commented):
A few tiny things.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerTransactionCreationOptionsTests.cs
purva9413: Done, adding the Theory in test method `WithReadLockMode`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
purva9413: That sounds reasonable. I already have the error msg if the params do not satisfy the constraints (readWrite + non-partionedDml) on the transaction for the readLockMode. So I think I will just remove the comment from here.

amanda-tarafa (dismissed):
LGTM, thanks!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -15,6 +15,7 @@
 using Google.Api.Gax;
 using Google.Cloud.Spanner.V1;
 using static Google.Cloud.Spanner.V1.TransactionOptions.Types;
+using TransactionReadLockMode = Google.Cloud.Spanner.V1.TransactionOptions.Types.ReadWrite.Types.ReadLockMode;
```
purva9413: TODO: Create an easily accessible ENUM type on top for Spanner.Data layer.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
```diff
@@ -607,6 +607,32 @@ await reader.ReadAsync(),
...
 
+        [Fact]
+        public async Task TransactionOptionsSetCorrectly()
+        {
+            // This test currently tests ReadLockMode and IsolationLevel
+            // More options compatible with these options can be added to this same test
+            using (var connection = _fixture.GetConnection())
+            {
+                await connection.OpenAsync();
+                var options = SpannerTransactionCreationOptions.ReadWrite.WithReadLockMode(ReadLockMode.Pessimistic).WithIsolationLevel(System.Data.IsolationLevel.Serializable);
```
amanda-tarafa: tiny nit:

```suggestion
                var options = SpannerTransactionCreationOptions.ReadWrite
                    .WithReadLockMode(ReadLockMode.Pessimistic)
                    .WithIsolationLevel(System.Data.IsolationLevel.Serializable);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/TransactionTests.cs
```diff
@@ -607,6 +607,32 @@ await reader.ReadAsync(),
             }
         }
 
+        [Fact]
+        public async Task TransactionOptionsSetCorrectly()
```
amanda-tarafa:
```suggestion
        public async Task CustomTransactionOptions()
```

amanda-tarafa (dismissed):
LGTM, with two tiny nits.

Also rebase/squash all commits in a single commit before merging.

PR #14927, opened 2025-07-01
feat(BigQuery): Add RequestModifier to Create, Get, Patch, Update DatasetOptions

cy-yun:
[b/374156757](https://buganizer.corp.google.com/issues/374156757), [b/410889782](https://buganizer.corp.google.com/issues/410889782) and [b/410884845](https://buganizer.corp.google.com/issues/410884845)

### DIFF
--- a/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
+++ b/apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using static Google.Apis.Bigquery.v2.DatasetsResource;
 using Google.Apis.Bigquery.v2.Data;
 using Google.Apis.Requests;
 using System;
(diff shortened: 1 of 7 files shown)


apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
```diff
@@ -114,13 +124,27 @@ public void GetDataset()
...
                 new Dataset { DatasetReference = reference });
-            var result = client.GetDataset(reference);
+            var result = client.GetDataset(
+                reference,
+                new GetDatasetOptions
+                {
+                    RequestModifier = request =>
+                    {
+                        request.AccessPolicyVersion = accessPolicyVersion;
+                        request.DatasetView = dataSetViewEnum;
```
amanda-tarafa: I think we can test a single one of these fields, because what we want to test is that the modifier is applied, and for that we only need to change one field. Let's keep accessPolicyVersion as that's the one that's common to all the methods.

apis/Google.Cloud.BigQuery.V2/Google.Cloud.BigQuery.V2.Tests/BigQueryClientImplTest.cs
```diff
@@ -146,26 +170,75 @@ public void GetDataset_FailThenRetry()
...
+        }
+
         // Effectively testing BigQueryClient.RetryIfETagPresent, but it's hard to test that without
         // making a concrete request via BigQueryClientImpl.
         [Fact]
         public void UpdateDataset_RetryIfETagPresent()
         {
             var projectId = "project";
             var datasetId = "dataset";
+            var accessPolicyVersion = 1;
```
amanda-tarafa, line 205: This test is testing something very specifi, i.e. retries, so let's not add this here. If there's no plain UpdateDataset test, add one and test this there.

amanda-tarafa (changes requested):
Mostly fine, just a couple of changes in tests.

amanda-tarafa (approved):
Thanks!

PR #14932, opened 2025-07-01
chore: re-enable Google.Cloud.Compute.V1 generation

jskeet:
This now works as we've updated the GAPIC generator.

(We should only merge this after updating to the new container image, obviously - hence the do-not-merge label.)


amanda-tarafa (approved):

jskeet:
Thanks! Will wait until the new image is built (it's in progress) before removing the label and merging.

PR #14973, opened 2025-07-15
chore: update dotnet tool versions and remove .NET 6

jskeet:
The old tools were the last reason for needing .NET 6 in our container.


jskeet:
(run-container-tests.sh works locally, btw.)

PR #14987, opened 2025-07-25
feat: Spanner IsolationLevel through ConnectionString

purva9413:


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -69,6 +70,15 @@ public string DatabaseRole
             set => this[DatabaseRoleKeyword] = value;
         }
 
+        /// <summary>
+        ///
```
amanda-tarafa: Nit: missing docs

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -595,6 +596,19 @@ internal Task<SpannerTransaction> BeginTransactionAsyncImpl(
                 }, "SpannerConnection.BeginTransactionAsync", Logger);
         }
 
+        private SpannerTransactionCreationOptions EnhanceTransactionCreationOptions(SpannerTransactionCreationOptions transactionCreationOptions)
```
amanda-tarafa: I think the name we'd usually use would be `MaybeWithConnectionDefaults`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -567,7 +567,8 @@ public Task<SpannerTransaction> BeginTransactionAsync(
                 nameof(transactionCreationOptions.IsSingleUse),
                 $"Single-use transactions may only be used with the {nameof(SpannerCommand.ExecuteReader)} set  of methods.");
 
-            return BeginTransactionAsyncImpl(transactionCreationOptions, transactionOptions, cancellationToken);
+            SpannerTransactionCreationOptions enhancedTransactionCreationOptions = EnhanceTransactionCreationOptions(transactionCreationOptions);
```
amanda-tarafa: Do all the begin transaction and similar methods end up here? We need to make sure that we apply de connection default to all transaction options.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -69,6 +70,15 @@ public string DatabaseRole
             set => this[DatabaseRoleKeyword] = value;
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        public string IsolationLevel
+        {
+            get => GetValueOrDefault(IsolationLevelKeyword, "");
```
amanda-tarafa: Should this be the enum value, defaulting to `Unspecified`. Do we have other fields here that map back to enums and what are we doing with those?

amanda-tarafa (commented):
I just reviewed the last commit as we talked about.

The approach looks good, I left a few comments but none a big deal.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
purva9413: Hmm I think all these properties set on the connection string are dealt with in string format so far. My thought process is we keep that consistent with IsolationLevel as well, and we can convert it to its Enum representation when we actually use it to set it on the SpannerTransactionOptions.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
purva9413: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: Hmm good question. All the transactions (i.e. ReadWrite with non partioned DML) for this use case go through the `BeginTransactionAsync`. `BeginTransactionAsync()` eventually calls into `BeginTransactionAsyncImpl()`.

But we have a Partitioned DML use case in `EphemeralTransaction.cs`, `ExecutePartitionedDmlAsync()` which directy calls into `BeginTransactionAsyncImpl`. So if we want this to be extendable in the future such that defaults for Partioned DML are also supported through the ConnectionString, it might make sense moving this into `BeginTransactionAsyncImpl()` instead

purva9413: Moving the call to `MaybeWithConnectionDefaults` into `BeginTransactionAsyncImpl` because I think it is better for the overall design but lmk if you have any questions/comments.

amanda-tarafa: I think it needs to be on `SpannerConnection.AcquireSessionAsync`, that's the lowest level method before handing off to Spanner.V1 and it's used from RetriableTransaction and EphemeralTransaction.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
amanda-tarafa: > I think all these properties set on the connection string are dealt with in string format so far

Nope, we have several `bool`, several `int` and some more complex ones like `DatabaseName`. `SpannerConnectionStringBuilder` can be used via code as well, and in those cases it wouldn't be user friendly to have developers do the conversions themselves.

amanda-tarafa (changes requested):

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: That seems fine to me too, one thing which will need to be handled when calling methods into the `SpannerConnection.AcquireSessionAsync` create a `SpannerTransaction` with the 'unmodified' creationOptions as a result. [Example](https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs#L606).

SpannerTransaction only has getters for the options, so we will have to modify those getters to return the connection string based option value using the `SpannerConnection` instance linked to this transaction.

amanda-tarafa: I'm no sure I get what you are saying, but the idea is that we modify `TransactionCreationOptions` before acquiring the session for which we use the modified options. After acquiring the session we create the SpannerTransaction with the maybe modified transaction creation options. And I say maybe modified, because in the case where a transaction creation options has a transaction ID we don't modify the transaction creation options: we are not really creating a new transaction, we are creating a representation of a shared transaction.

Happy to chat more about this tomorrow.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
purva9413: Noted, changing this to return IsolationLevel instead. Thanks!

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: As discussed offline, added in a `out` var to deal with functions calling into `AcquireSessionAsync` and then creating a `SpannerTransaction` with the creation options.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -69,6 +71,16 @@ public string DatabaseRole
             set => this[DatabaseRoleKeyword] = value;
         }
 
+        /// <summary>
+        /// Option to set the default for <see cref="System.Data.IsolationLevel"/> for all transactions.
+        /// If this is set on the Transaction explicitly, that will take precedence over this default.
```
amanda-tarafa: On the override on a transaction basis, mention SpannerTransactionCreationOptions, something like.

```suggestion
        /// Option to set the default for <see cref="System.Data.IsolationLevel"/> for all transactions executed on this connection.
        /// Override on a per-transaction basis using <see cref="SpannerTransactionCreationOptions" />.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -69,6 +71,16 @@ public string DatabaseRole
...
+        /// </summary>
+        public IsolationLevel IsolationLevel
+        {
+            get => (IsolationLevel) Enum.Parse(typeof(IsolationLevel), GetValueOrDefault(IsolationLevelKeyword, "Unspecified"));
```
amanda-tarafa: Can you add a `TEnum GetEnumValueOrDefault<TEnum>(string key, TEnum default)`, it should be similar to the GetUint... method in that it checks the dictionary for the enum value or the string value, etc. You can use `Enum.TryParse<TEnum>` internally.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -69,6 +71,16 @@ public string DatabaseRole
             set => this[DatabaseRoleKeyword] = value;
         }
 
+        /// <summary>
+        /// Option to set the default for <see cref="System.Data.IsolationLevel"/> for all transactions.
+        /// If this is set on the Transaction explicitly, that will take precedence over this default.
+        /// </summary>
+        public IsolationLevel IsolationLevel
```
amanda-tarafa, line 81: We need some unit tests for this to document the default value for this property and that the conversion happens if the stored value is a string.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -578,7 +566,7 @@ public Task<SpannerTransaction> BeginTransactionAsync(
                 !transactionCreationOptions.IsSingleUse,
                 nameof(transactionCreationOptions.IsSingleUse),
                 $"Single-use transactions may only be used with the {nameof(SpannerCommand.ExecuteReader)} set  of methods.");
-
+
```
amanda-tarafa: Nit: revert

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -587,6 +575,8 @@ internal Task<SpannerTransaction> BeginTransactionAsyncImpl(
             SpannerTransactionOptions transactionOptions,
             CancellationToken cancellationToken)
         {
+            SpannerTransactionCreationOptions enhancedTransactionCreationOptions = transactionCreationOptions;
```
amanda-tarafa: I think we don't need to do this if we pass the parameter as `ref`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -1002,10 +1008,11 @@ public async Task WhenSessionPoolReady(CancellationToken cancellationToken = def
             await _sessionPool.WhenPoolReady(sessionPoolSegmentKey, cancellationToken).ConfigureAwait(false);
         }
 
-        internal Task<PooledSession> AcquireSessionAsync(SpannerTransactionCreationOptions creationOptions, CancellationToken cancellationToken)
+        internal Task<PooledSession> AcquireSessionAsync(SpannerTransactionCreationOptions creationOptions, CancellationToken cancellationToken, out SpannerTransactionCreationOptions enhancedTransactionCreationOptions)
```
amanda-tarafa: Something like this:

```suggestion
        internal Task<PooledSession> AcquireSessionAsync(ref SpannerTransactionCreationOptions creationOptions, CancellationToken cancellationToken)
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -1002,10 +1008,11 @@ public async Task WhenSessionPoolReady(CancellationToken cancellationToken = def
...
         {
             SessionPool pool;
             DatabaseName databaseName;
+            enhancedTransactionCreationOptions = MaybeWithConnectionDefaults(creationOptions);
```
amanda-tarafa:
```suggestion
            creationOptions = MaybeWithConnectionDefaults(creationOptions);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
```diff
@@ -78,7 +78,8 @@ public sealed class SpannerTransaction : SpannerTransactionBase, ISpannerTransac
         private int _lastDmlSequenceNumber = 0;
 
         /// <inheritdoc />
-        public override IsolationLevel IsolationLevel => IsolationLevel.Serializable;
+        /// The isolation level set in the transaction creation options
```
amanda-tarafa: I think this now should be a normal `/// <summary></summary>` instead of inherit docs. Not sure what the inherit docs say though, but if that matches what we are doing, then we should just keep it.

amanda-tarafa (changes requested):
A couple more adjustments but looking good.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransaction.cs
purva9413: Yes let's keep the summary same as the inerhitdoc

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: If we make this a field passed by reference, we need to consider the calls from `EphemeralTransaction` and `RetriableTransaction` where the `_creationOptions` are private and readonly currently. This would mean we will have to remove the `readonly` modifier on these fields in these classes.

That seems fine to me since neither the `SpannerClientCreationOption` defaults nor the `SpannerTransactionCreationOptions` already set on the txn should change once objects for these types of transactions are created. 

But removing the `readonly` could potentially open up these fields to potential misuse and unintended modifications in the future. Thoughts?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
purva9413: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
purva9413: Good idea, added in the generic method to convert to the Enum

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: As discussed offline, we decided to keep this as an `out` var due to the property being `readonly` in some classes.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
amanda-tarafa: We discussed this offline, and we agreed to going back to using an `out` parameter and refactor once we have done multiplexed sessions.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerTransactionCreationOptions.cs
```diff
@@ -92,6 +95,12 @@ public sealed class SpannerTransactionCreationOptions
     /// </summary>
     public bool IsPartitionedDml { get; }
 
+    /// <summary>
+    /// Snapshot Isolation Level set to <see cref="System.Data.IsolationLevel"/>.
+    /// Only meant to be set for read-write transactions.
```
amanda-tarafa: Sorry for not mentioning this before, but just to make it match the others in this type:

```suggestion
    /// Isolation level for a read-write transaction.
    /// This will always be <see cref="IsolationLevel.Unspecified"/> for red-only transactions.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -596,6 +608,19 @@ private string GetValueOrDefault(string key, string defaultValue = "")
...
+            if (Enum.TryParse<TEnum>(key, true, out TEnum result))
+            {
+                return result;
+            }
+            else
+            {
+                return defaultValue;
+            }
+        }
+
```
amanda-tarafa: I was thinking of something more like `GetInt32OrDefault` that looks directly on the dictionary, and the dictionary can have both a string or an actual enum value. Like this:
```suggestion
        private TEnum GetEnumValueOrDefault<TEnum>(string key, TEnum defaultValue)
            where TEnum : struct =>
            TryGetValue(key, out object value)
                && (value is TEnum parsed
                || (value is string text && Enum.TryParse<TEnum>(text, out parsed))))
            ? parsed
            : default;
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -593,20 +581,37 @@ internal Task<SpannerTransaction> BeginTransactionAsyncImpl(
...
+            if(effectiveCreationOptions?.IsolationLevel == IsolationLevel.Unspecified
+                && Builder.IsolationLevel != IsolationLevel.Unspecified
+                && effectiveCreationOptions?.TransactionMode == TransactionMode.ReadWrite
+                && effectiveCreationOptions?.IsPartitionedDml == false)
+            {
+                effectiveCreationOptions = transactionCreationOptions.WithIsolationLevel(Builder.IsolationLevel);
+            }
+
+            return effectiveCreationOptions;
+        }
```
amanda-tarafa: I think we don't need the extra variable here?

```suggestion
        private SpannerTransactionCreationOptions MaybeWithConnectionDefaults(SpannerTransactionCreationOptions transactionCreationOptions) =>
        // Set the transaction IsolationLevel if specified in the ConnectionString and not already set on the SpannerTransactionCreationOptions
        effectiveCreationOptions?.IsolationLevel == IsolationLevel.Unspecified
            && Builder.IsolationLevel != IsolationLevel.Unspecified
            && effectiveCreationOptions?.TransactionMode == TransactionMode.ReadWrite
            && !effectiveCreationOptions?.IsPartitionedDml
        ? transactionCreationOptions.WithIsolationLevel(Builder.IsolationLevel);
        : transactionCreationOptions;
```

amanda-tarafa: Also, move this method below AcquierSessionAsync now that we just use it there.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -593,20 +581,37 @@ internal Task<SpannerTransaction> BeginTransactionAsyncImpl(
                     await OpenAsync(cancellationToken).ConfigureAwait(false);
 
                     PooledSession session;
+                    SpannerTransactionCreationOptions effectiveCreationOptions = transactionCreationOptions;
```
amanda-tarafa: I think if we don't initialize this here is clearer when we are using modified options or not:
```suggestion
                    SpannerTransactionCreationOptions effectiveCreationOptions;
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
```diff
@@ -593,20 +581,37 @@ internal Task<SpannerTransaction> BeginTransactionAsyncImpl(
...
                     {
-                        session = await AcquireSessionAsync(transactionCreationOptions, cancellationToken).ConfigureAwait(false);
+                        session = await AcquireSessionAsync(transactionCreationOptions, cancellationToken, out effectiveCreationOptions).ConfigureAwait(false);
                     }
                     else
                     {
-                        SessionName sessionName = SessionName.Parse(transactionCreationOptions.TransactionId.Session);
-                        ByteString transactionIdBytes = ByteString.FromBase64(transactionCreationOptions.TransactionId.Id);
+                        SessionName sessionName = SessionName.Parse(effectiveCreationOptions.TransactionId.Session);
+                        ByteString transactionIdBytes = ByteString.FromBase64(effectiveCreationOptions.TransactionId.Id);
```
amanda-tarafa: And then:

```suggestion
                        SessionName sessionName = SessionName.Parse(transactionCreationOptions.TransactionId.Session);
                        ByteString transactionIdBytes = ByteString.FromBase64(transactionCreationOptions.TransactionId.Id);
                        effectiveCreationOptions = transactionCreationOptioins;
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
```diff
@@ -81,8 +81,9 @@ async Task<TResult> CommitAttempt()
 
                         try
                         {
-                            session = await (session?.RefreshedOrNewAsync(cancellationToken) ?? _connection.AcquireSessionAsync(_creationOptions, cancellationToken)).ConfigureAwait(false);
-                            transaction = new SpannerTransaction(_connection, session, _creationOptions, _transactionOptions, isRetriable: true);
+                            SpannerTransactionCreationOptions effectiveCreationOptions = _creationOptions;
+                            session = await (session?.RefreshedOrNewAsync(cancellationToken) ?? _connection.AcquireSessionAsync(_creationOptions, cancellationToken, out effectiveCreationOptions)).ConfigureAwait(false);
+                            transaction = new SpannerTransaction(_connection, session, effectiveCreationOptions, _transactionOptions, isRetriable: true);
```
amanda-tarafa:
```suggestion
                            session = await (session?.RefreshedOrNewAsync(cancellationToken) ?? _connection.AcquireSessionAsync(_creationOptions, cancellationToken, out var effectiveCreationOptions)).ConfigureAwait(false);
                            transaction = new SpannerTransaction(_connection, session, effectiveCreationOptions, _transactionOptions, isRetriable: true);
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: The explicit var `effectiveCreationOptions` would help when we want to check and set multiple default options from the connection string.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: The compiler then complains on any use of the `effectiveCreationOptions` that it is a use of an uninitialized var, and I did not want to initialize it to null since in any code line after we used to use the `transactionCreationOptions` so thaving `SpannerTransactionCreationOptions effectiveCreationOptions = transactionCreationOptions;` would effectively mimic that behaviour.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
purva9413: IMO using `effectiveCreationOptions` in some parts and `transactionCreationOptioins` makes the code a bit confusing. It is easy to miss which options are used and passed at which point in this function. That's the reason behind only using the `effectiveCreationOptions` further in the code once those are initialized via the `out` var. Happy to discuss more on this though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
purva9413: Interesting syntax but I get the same compiler complaint with the use of an unassigned var - [CS0165](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0165?f1url=%3FappId%3Droslyn%26k%3Dk(CS0165))

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnection.cs
amanda-tarafa: We've discussed offline and agreed on something that we both like.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/RetriableTransaction.cs
amanda-tarafa: Yes, it does not work. If we need to refresh the session we are not initializing the `effectiveCreationOptions`.

amanda-tarafa (approved):
Thanks!

PR #14992, opened 2025-07-28
feat: Support channel affinity from Spanner.V1

amanda-tarafa:
@jskeet and @purva9413 , see what you think of this approach. After this is in, adding support for universe domain in Spanner.Data is trivial.

As always, best to review one commit at a time.


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
```diff
@@ -100,6 +100,13 @@ public SpannerClientBuilder MaybeCreateEmulatorClientBuilder()
             return builder;
         }
 
+        private void CopySettingsForEmulator(SpannerClientBuilder client)
```
jskeet: I'd be tempted to rename this, to avoid confusion. It took me a minute to figure out how *just* adding a new private method, that isn't overriding anything, and without adding any new calls to it, could do anything.

(Arguably it should be virtual in the base class, basically... maybe something for a new major GAX version?)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
```diff
@@ -57,14 +113,67 @@ public partial class SpannerClientBuilder
...
+                ? await base.CreateCallInvokerAsync(cancellationToken).ConfigureAwait(false)
+                : new GcpCallInvoker(
+                    ServiceMetadata,
+                    EffectiveEndpoint,
+                    await GetChannelCredentialsAsync(cancellationToken).ConfigureAwait(false),
+                    GetChannelOptions(),
+                    GetApiConfing(),
+                    EffectiveGrpcAdapter);
+
+        private ApiConfig GetApiConfing() => new ApiConfig
```
jskeet: Nit: typo in the name.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
```diff
@@ -57,14 +113,67 @@ public partial class SpannerClientBuilder
...
+        }
+
         partial void InterceptBuild(ref SpannerClient client) => client = MaybeCreateEmulatorClientBuilder()?.Build();
 
         partial void InterceptBuildAsync(CancellationToken cancellationToken, ref Task<SpannerClient> task) =>
             task = MaybeCreateEmulatorClientBuilder()?.BuildAsync(cancellationToken);
 
+        /// <inheritdoc/>
+        protected override CallInvoker CreateCallInvoker() =>
+         AffinityChannelPoolConfiguration is null
```
jskeet: Odd indentation here. I'd indent everything another three spaces.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
```diff
@@ -57,14 +113,67 @@ public partial class SpannerClientBuilder
...
         public DirectedReadOptions DirectedReadOptions { get; set; }
 
+        /// <summary>
+        /// Specifies the configuration to use for the affinity channel pool.
+        /// When set, each session will be bound to a gRPC channel meaning that
+        /// all commands executed on that session will be executed on that gRPC channel.
+        /// May be null, in which case there will be no channel affinity.
+        /// Defaults to null.
+        /// </summary>
+        public ChannelPoolConfig AffinityChannelPoolConfiguration {  get; set; }
```
jskeet, line 128: Maybe document something about changes after construction?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -39,12 +33,6 @@ namespace Google.Cloud.Spanner.Data
     /// </summary>
     public sealed class SessionPoolManager
     {
-        // TODO: Should these be configurable?
-        private static readonly GrpcChannelOptions s_grpcChannelOptions = GrpcChannelOptions.Empty
```
jskeet, line 43: These are the defaults from elsewhere, right?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -252,83 +240,8 @@ public override string ToString()
...
-            var callInvoker = new GcpCallInvoker(SpannerClient.ServiceMetadata, clientCreationOptions.Endpoint, credentials, s_grpcChannelOptions, apiConfig, clientCreationOptions.GrpcAdapter);
-            return new SpannerClientBuilder
-            {
-                CallInvoker = callInvoker,
-                Settings = spannerSettings,
-                LeaderRoutingEnabled = clientCreationOptions.LeaderRoutingEnabled,
-                DirectedReadOptions = clientCreationOptions.DirectedReadOptions,
-            }.Build();
-        }
+        internal static Task<SpannerClient> CreateClientAsync(SpannerClientCreationOptions clientCreationOptions, SpannerSettings spannerSettings, Logger logger) =>
```
jskeet: Why are we accepting a logger if we don't use it?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
```diff
@@ -30,106 +27,102 @@ namespace Google.Cloud.Spanner.Data
     /// </summary>
     internal sealed class SpannerClientCreationOptions : IEquatable<SpannerClientCreationOptions>
     {
-        private static readonly Lazy<Task<ChannelCredentials>> s_defaultCredentialsTaskProvider = new Lazy<Task<ChannelCredentials>>(CreatedScopedDefaultCredentials);
-
-        private static async Task<ChannelCredentials> CreatedScopedDefaultCredentials()
+        private static Lazy<DatabaseAdminSettings> DatabaseAdminSettingsCache { get; } = new Lazy<DatabaseAdminSettings>(CreateDatabaseAdminSettings);
```
jskeet: Why does this need to be lazy at all? Can't we just create it immediately? (We used to have laziness to avoid acquiring default credentials eagerly, but that's not in DatabaseAdminSettings.)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
```diff
@@ -57,14 +113,67 @@ public partial class SpannerClientBuilder
...
+                    ServiceMetadata,
+                    EffectiveEndpoint,
+                    await GetChannelCredentialsAsync(cancellationToken).ConfigureAwait(false),
+                    GetChannelOptions(),
+                    GetApiConfing(),
+                    EffectiveGrpcAdapter);
+
+        private ApiConfig GetApiConfing() => new ApiConfig
+        {
+            ChannelPool = AffinityChannelPoolConfiguration.Clone(),
```
jskeet: I don't think we really need to clone here, if we're only passing this to the GcpCallInvoker constructor - that clones the ApiConfig anyway.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
```diff
@@ -30,106 +27,102 @@ namespace Google.Cloud.Spanner.Data
...
-            // If the client connects to the emulator, use its credentials (regardless of builder.CredentialOverride)
-            _credentialsOverride = emulatorBuilder?.ChannelCredentials ?? builder.CredentialOverride;
-            _googleCredential = builder.GoogleCredential;
-            MaximumGrpcChannels = builder.MaximumGrpcChannels;
-            MaximumConcurrentStreamsLowWatermark = (uint) builder.MaxConcurrentStreamsLowWatermark;
-            LeaderRoutingEnabled = builder.EnableLeaderRouting;
-            DirectedReadOptions = builder.DirectedReadOptions;
+            var emulatorBuilder = clientBuilder.MaybeCreateEmulatorClientBuilder();
 
-            // TODO: add a way of setting this from the SpannerConnectionStringBuilder.
```
jskeet, line 99: Have we deliberately lost this TODO?

jskeet (dismissed):
A few nits, but I generally like it a lot. It's entirely possible that there are aspects I didn't catch that need adjustment, but if all the tests pass, we're probably fine.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
amanda-tarafa: +1 to making some methods virtual on ClientBuilderBase, I've added a note to our major version document about this.

Instead of renaming, I'll leave a comment, to make sure that this method hides the inehrited one.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
amanda-tarafa: Yep, done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
amanda-tarafa: Done

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
amanda-tarafa: Oh, I didn't noticed the constructor cloning this. So yes, no need to do that ourselves, I did leave a note as to why we weren't cloning there.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
amanda-tarafa: Yes, from [ClientBuilderBase](https://github.com/googleapis/gax-dotnet/blob/81c6bc81cd2cc9f0b16bc4c9396b5bbfbc7ad590/Google.Api.Gax.Grpc/ClientBuilderBase.cs#L33) :)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
amanda-tarafa: Oh, I'll send a separate commit to remove it, thought about it earlier.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
amanda-tarafa: It's reflection after all, and the DatabaseAdmin client is likely to be less used so we won't always need this. Happy to change it though?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
amanda-tarafa: Yes, because this now falls in the category of "Add a way to configure eerything that's on the Client Builder" like scopes, quota project, channel options, etc.... It now seemed odd to keep a comment just for this?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
```diff
@@ -30,106 +27,102 @@ namespace Google.Cloud.Spanner.Data
...
-            GrpcAdapter = GrpcAdapter.GetFallbackAdapter(SpannerClient.ServiceMetadata);
+            UsesEmulator = emulatorBuilder is not null;
+            _clientBuilder = emulatorBuilder ?? clientBuilder;
         }
 
+        internal Task<SpannerClient> CreateSpannerClientAsync(SpannerSettings settings) =>
+            // TODO: Consider implementing clone in SpannerClientBuilder.
+            new SpannerClientBuilder
```
purva9413, line 61: Why do we need a new SpannerClientBuilder here? Why not just use the one created in the constructor and set as a private variable on this class? `_clientBuilder.BuildAsync();`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -248,18 +239,11 @@ private async Task<int> ExecuteDdlAsync(CancellationToken cancellationToken)
...
                 // (This is fairly unpleasant, but we'll try to improve this in the next version of GAX.)
-                var databaseAdminClientBuilder = new DatabaseAdminClientBuilder
-                {
-                    // Note: deliberately not copying EmulatorDetection, as that's handled in SpannerClientCreationOptions
-                    Settings = s_databaseAdminSettings,
-                    Endpoint = channelOptions.Endpoint,
-                    ChannelCredentials = credentials
-                };
-                var databaseAdminClient = databaseAdminClientBuilder.Build();
+                var databaseAdminClientBuilder = channelOptions.CreateDatabaseAdminClientBuilder();
```
purva9413: Given it is an object of the SpannerClientCreationOptions would it be better if we rename `channelOptions` to `clientOptions`?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
```diff
@@ -57,14 +113,74 @@ public partial class SpannerClientBuilder
...
+        /// all commands executed on that session will be executed on that gRPC channel.
+        /// May be null, in which case there will be no channel affinity.
+        /// Defaults to null.
+        /// </summary>
+        /// <remarks>
+        /// Each client built will hold itw own copy of <see cref="AffinityChannelPoolConfiguration"/>,
+        /// which means changes made to this value affect clients built after the change but not clients
+        /// already built when the change happens.
+        /// </remarks>
+        public ChannelPoolConfig AffinityChannelPoolConfiguration {  get; set; }
```
purva9413, line 128: iiuc with the `SpannerClientBuilder` created in the constructor of `SpannerClientCreationOptions` always having `AffinityChannelPoolConfiguration` set, we only expect this to be not set for folks not going through the ADO.NET layer?

purva9413 (approved):

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
amanda-tarafa: Yes, that's correct. And we need both aspects of that, it's always set from SpannerClientCreationOptions and never set by default from users using SpannerClientBuilder directly, else, that'd be a breaking change (in behaviour).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
amanda-tarafa: Because this method is called with specific SpannerSettings each time, that are set on the cliet builder. In principle, the same SpannerClientCreationOptions may be used to create several different Spanner clients with different SpannerSettings, even in parallel, so all of those may end up overlapping if we used the same client builder.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
amanda-tarafa: Good question, this circles back a little to the chat we were meant to have about the client builder building mostly the call invoker, etc.

Although SpannerClientCreationOptions truly has options that configure the Spanner client and not just the call invoker, those client specific options do not apply to the Database admin client, only the call invoker (channel) options from SpannerClientCreationOptions are relevant for building a Database admin client, and that's what the name of this variable is meant to convey.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
jskeet: Let's keep it simple (remove the laziness) - there's not a lot of reflection here (just finding the version number) and I'd expect it to be insignificant in the grand scheme of things.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
jskeet: Fair enough. I guess we're only likely to add things to SpannerConnectionStringBuilder if either they're Spanner-specific or we get asked to by a customer.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
jskeet: I nearly left the same comment when I first reviewed. How about `databaseAdminClientOptions`? Or maybe `databaseAdminConnectionOptions`? And add a comment above the declaration explaining what it's really for. (`channelOptions` sounds like it would be a `GrpcChannelOptions` or similar...)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/SpannerClientBuilderPartial.cs
```diff
@@ -57,14 +113,74 @@ public partial class SpannerClientBuilder
...
 
+        /// <summary>
+        /// Specifies the configuration to use for the affinity channel pool.
+        /// When set, each session will be bound to a gRPC channel meaning that
+        /// all commands executed on that session will be executed on that gRPC channel.
+        /// May be null, in which case there will be no channel affinity.
+        /// Defaults to null.
+        /// </summary>
+        /// <remarks>
+        /// Each client built will hold itw own copy of <see cref="AffinityChannelPoolConfiguration"/>,
```
jskeet: itw => its

amanda-tarafa: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
amanda-tarafa: Yep, done.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
amanda-tarafa: I'm using `connectionOptions` as that's really what it is? And which of those connection options is used to create what, it's irrelevant. I'll push this, but happy to reconsider to something else.

amanda-tarafa (commented):
Latest comment addressed in review commits (typos fixes are squashed in their corresponding commits). PTAL

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
```diff
@@ -27,7 +27,7 @@ namespace Google.Cloud.Spanner.Data
     /// </summary>
     internal sealed class SpannerClientCreationOptions : IEquatable<SpannerClientCreationOptions>
     {
-        private static Lazy<DatabaseAdminSettings> DatabaseAdminSettingsCache { get; } = new Lazy<DatabaseAdminSettings>(CreateDatabaseAdminSettings);
+        private static DatabaseAdminSettings DatabaseAdminSettingsCache{ get; } = CreateDatabaseAdminSettings();
```
jskeet: Nit: space after the name.

As noted on chat, we could potentially get rid of the field and extra method, and just inline this code into CreateDatabaseAdminClientBuilder, but that can come later.

amanda-tarafa: Inlined now :)

jskeet (approved):

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
purva9413: I see, so you are saying if clients are created parallelly and since there is no thread safety in the client creation, we may end up overwriting the settings for one client with another if we use the same builder. Is that understanding correct? That makes sense, otherwise I was thinking we could just set the setting on the same builder each time before calling BuildAsync, but that would not work if the settings for clients can be clobbered by creation of other clients in parallel.

amanda-tarafa: Yes, your understanding is correct.

And even if these were not to be used in parallel, we would be overriding settings each time. That might not look like a big problem now but imagine if we were to expose settings in the connection string. So then we'd have settings coming from the connection string and settings comming fro the session pool manager, and we'd need to merge both, but if we are every time merging on top of the same property the end result is a mix and match of settings...

In general, anything shared like this builder should be inmutable.

PR #14998, opened 2025-08-01
feat: Support UniverseDomain

purva9413:


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerClientCreationOptionsTest.cs
```diff
@@ -137,5 +137,14 @@ public async Task GetCredentialsAsync_GoogleCredential()
             var options = new SpannerClientCreationOptions(builder);
             Assert.NotNull(await options.CreateSpannerClientAsync(new Spanner.V1.SpannerSettings()));
         }
+
+        [Fact]
+        public async Task Equality_UniverseDomainAndEndpoint()
```
amanda-tarafa: nit: move this with the other equality tests a few lines up.

amanda-tarafa: And also complete the test similar to the ones up there including what the effects of setting universe domain have for the endpoint here.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
```diff
@@ -336,5 +336,17 @@ public void Clone_MaintainsGoogleCredential()
             var clone = builder.Clone();
             Assert.Same(credential, clone.GoogleCredential);
         }
+
+        [Fact]
+        public void UniverseDomainTest()
```
amanda-tarafa: We usually split this tests in tow, one for default and one for custom.

I'd also like to see tests regarding the effect that setting universe domain has on endpoint on the connection string.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
```diff
@@ -147,6 +150,10 @@ public override string ToString()
             {
                 builder.Append($"; GoogleCredential: True");
             }
+            if (_clientBuilder.UniverseDomain is not null)
+            {
+                builder.Append($"; UniverseDomain: {_clientBuilder.UniverseDomain}");
+            }
```
amanda-tarafa: Maybe something like what we are doing for directec read options etc., where if it is null, we say "Default". And do the same for Endpoint.

(The if makes more sense for credentials because each credential property is an alternative to the other).

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -140,6 +141,28 @@ public string DataSource
             set => this[DataSourceKeyword] = ValidatedDataSource(value);
         }
 
+        /// <summary>
```
amanda-tarafa: Move this to where Endpoint, Host and Port are.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -140,6 +141,28 @@ public string DataSource
...
+        /// The universe domain to connect to, or null to use the default universe domain <see cref="DefaultDomain"/>.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// <see cref="UniverseDomain"/> is used to build the endpoint to connect to, unless <see cref="EndPoint"/>
+        /// is explicitly set (i.e. without using the default host and port), in which case <see cref="EndPoint"/> will be used without further modification.
```
amanda-tarafa: I think the part in parenthesis is confusing. And this whole thing should be:

"unless EndPoint or Host are explicitly set"

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -140,6 +141,28 @@ public string DataSource
...
+        /// </para>
+        /// <para>
+        /// If default credentials or one of <see cref="GoogleCredential"/> or <see cref="CredentialFile"/>
+        /// is used, <see cref="GoogleCredential.GetUniverseDomain"/> should be:
+        /// <list type="bullet">
+        /// <item>The same as <see cref="UniverseDomain"/> if <see cref="UniverseDomain"/> has been set.</item>
```
amanda-tarafa: Or the default universe domain.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -31,7 +31,8 @@ namespace Google.Cloud.Spanner.Data
     /// </summary>
     public sealed class SpannerConnectionStringBuilder : DbConnectionStringBuilder
     {
-        internal const string DefaultHost = "spanner.googleapis.com";
+        internal const string DefaultDomain = "googleapis.com";
+        internal const string DefaultHost = $"spanner.{DefaultDomain}";
```
amanda-tarafa: The default host cannot be a constant now, it needs to be `$"spanner.{UniverseDomain}"`, because if the user sets UniverseDomain = "mydomain.com" then the host will be "spanner.mydomain.com".

Because none of these properties ever return null, Host and EndPoint need to all be in sync with UniverseDomain and Port.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
```diff
@@ -48,6 +48,7 @@ internal SpannerClientCreationOptions(SpannerConnectionStringBuilder builder)
                 },
                 LeaderRoutingEnabled = builder.EnableLeaderRouting,
                 DirectedReadOptions = builder.DirectedReadOptions,
+                UniverseDomain = builder.UniverseDomain == SpannerConnectionStringBuilder.DefaultDomain ? null : builder.UniverseDomain,
```
amanda-tarafa: I think we can now add a method to the builder (or use TryGetValue) to see if a value has been actually set or not instead of comparing to the defaults. The same for Host and Port.

amanda-tarafa (changes requested):
Happy to chat about these, but my requests are mostly because we never return null on the specific connection string builder properties, so, for the ones that are related, like EndPoint, UniverseDomain, Host, etc. we need to keep them consistent.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
purva9413: Ack, I think the [ContainsKey](https://learn.microsoft.com/en-us/dotnet/api/system.data.common.dbconnectionstringbuilder.containskey?view=net-9.0) makes more sense in this case though. Going ahead and using that but lmk if you think otherwise.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
purva9413: Removing the const string DefaultHost altogether since we do not need it anymore for checks in other classes. Instead for the `Host` getter adding in `get => GetValueOrDefault(nameof(Host), $"spanner.{UniverseDomain}");`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerClientCreationOptionsTest.cs
purva9413: Done, I made the `_clientBuilder` into an `internal` property. Even so, we can only test the setting of Endpoint and UniverseDomain on the client creation options, but we cannot test the `EffectiveEndpoint` which is finally used by the `ClientBuilderBase` to create the GcpCallInvoker.

We do have unit tests for that separately [in GAX](https://github.com/googleapis/gax-dotnet/blob/main/Google.Api.Gax.Grpc.Tests/ClientBuilderBaseTest.EffectiveEndpointTest.cs) so I think that should be covered though.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerClientCreationOptions.cs
purva9413: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
purva9413: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -281,6 +281,28 @@ public int Port
             set => SetInt32WithValidation(nameof(Port), 1, 65535, value);
         }
 
+        /// <summary>
+        /// The universe domain to connect to, or null to use the default universe domain <see cref="DefaultDomain"/>.
+        /// </summary>
+        /// <remarks>
+        /// <para>
+        /// <see cref="UniverseDomain"/> is used to build the endpoint to connect to, unless <see cref="EndPoint"/> or <see cref="Host"/>
+        /// is explicitly set, in which case <see cref="EndPoint"/> will be used without further modification.
```
amanda-tarafa:
```suggestion
        ///  are explicitly set, in which case <see cref="EndPoint"/> will be used without further modification.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerConnectionStringBuilder.cs
```diff
@@ -281,6 +281,28 @@ public int Port
...
+        /// </para>
+        /// <para>
+        /// If default credentials or one of <see cref="GoogleCredential"/> or <see cref="CredentialFile"/>
+        /// is used, <see cref="GoogleCredential.GetUniverseDomain"/> should be:
+        /// <list type="bullet">
+        /// <item>The same as <see cref="UniverseDomain"/> if <see cref="UniverseDomain"/> has been set otherwise should be same as <see cref="DefaultDomain"/></item>
+        /// </list>
```
amanda-tarafa: I don't think we need a list if it's going to have a single item.

```suggestion
        /// is used, <see cref="GoogleCredential.GetUniverseDomain"/> should be
        /// the same as <see cref="UniverseDomain"/> if <see cref="UniverseDomain"/> has been set otherwise should be same as <see cref="DefaultDomain"/>
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
```diff
@@ -336,5 +336,32 @@ public void Clone_MaintainsGoogleCredential()
             var clone = builder.Clone();
             Assert.Same(credential, clone.GoogleCredential);
         }
+
+        [Fact]
+        public void Default_UniverseDomain()
+        {
+            var builder = new SpannerConnectionStringBuilder();
+
+            Assert.Equal(SpannerConnectionStringBuilder.DefaultDomain, builder.UniverseDomain);
```
amanda-tarafa, line 345: Can you assert here host and endpoint as well?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
```diff
@@ -336,5 +336,32 @@ public void Clone_MaintainsGoogleCredential()
...
+        }
+
+        [Fact]
+        public void Custom_UniverseDomain()
+        {
+            string universeDomain = "test-domain.test.goog";
+            var builder = new SpannerConnectionStringBuilder($"UniverseDomain={universeDomain}");
+
+            Assert.Equal(universeDomain, builder.UniverseDomain);
```
amanda-tarafa, line 356: And here, host and endpoint?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
```diff
@@ -336,5 +336,32 @@ public void Clone_MaintainsGoogleCredential()
...
+        }
+
+        [Fact]
+        public void EffectOfUniverseDomainOnEndpoint()
+        {
+            string universeDomain = "test-domain.test.goog";
+            var builder = new SpannerConnectionStringBuilder($"UniverseDomain={universeDomain}");
+
+            Assert.Equal(universeDomain, builder.UniverseDomain);
+            Assert.Equal($"spanner.{universeDomain}:443", builder.EndPoint);
```
amanda-tarafa, line 358: And host?

amanda-tarafa (commented):
LGTM, just a couple of tiny nits.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerConnectionStringBuilderTests.cs
purva9413: I'll unify the asserts for custom universe domain and its effect on host and endpoint under the `Custom_UniverseDomain` test and remove this test altogether since then they will just be checking the same thing.

PR #15023, opened 2025-08-06
tests: Fix diagnostics integration test

jskeet:
Currently this is failing with:

> System.InvalidOperationException : Multiple constructors accepting
> all given argument types have been found in type
> 'Google.Cloud.Diagnostics.AspNetCore3.IntegrationTests.NoBufferWarningLoggerTestApplication'.
> There should only be one applicable constructor.

Making the constructor protected allows it to still be called from derived classes, but without confusing DI.


amanda-tarafa (approved):
Oh, interesting...

jskeet:
> Oh, interesting...

My guess is that some policy changed between versions of the DI framework, otherwise we'd have seen this before.

PR #15108, opened 2025-09-19
Multiplex sessions v1

purva9413:
Creating a draft PR for early review.

Idea here is SessionPool translates to the MultiplexSession class. And PooledSession translates to the Transaction wrapper in this new design.

Lifecycle of the MulitplexSession -- Each query/operation coming in as part of the transaction checks for the expiry of the session. There is a proactive 'refresh' of the session every 7 days. If an operation comes in when the session is past this 7 day period, it can still safely use the existing session and we fire a session refresh in the background asynchronously. If the session is past 28 days of expiry, this session can no longer be safely used since there is already cleanup on the backend. The transaction needs to block till we get a new Session.

TODO: 
1. Add in summary for public methods
2. Add in more Unit tests covering MultiplexSession class

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/CloudSpannerFixtureBase.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/CloudSpannerFixtureBase.cs
@@ -14,8 +14,10 @@
 
 using Google.Cloud.ClientTesting;
 using Google.Cloud.Spanner.Common.V1;
+using Google.Cloud.Spanner.V1;
 using Google.Cloud.Spanner.V1.Internal.Logging;
 using System;
+using System.Threading.Tasks;
 
 namespace Google.Cloud.Spanner.Data.CommonTesting;
 
@@ -39,4 +41,6 @@ public abstract class CloudSpannerFixtureBase<TDatabase> : CloudProjectFixtureBa
     protected CloudSpannerFixtureBase(Func<string, TDatabase> databaseFactory) => Database = databaseFactory(ProjectId);
 
     public SpannerConnection GetConnection(Logger logger = null, bool logCommitStats = false) => Database.GetConnection(logger, logCommitStats);
+
+    public async Task<MultiplexSession> GetMultiplexSession() => await Database.GetMultiplexSession().ConfigureAwait(false);
 }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.CommonTesting/SpannerTestDatabaseBase.cs
@@ -16,9 +16,11 @@ using Google.Api.Gax;
 using Google.Api.Gax.ResourceNames;
 using Google.Cloud.Spanner.Admin.Instance.V1;
 using Google.Cloud.Spanner.Common.V1;
+using Google.Cloud.Spanner.V1;
 using Google.Cloud.Spanner.V1.Internal.Logging;
 using Grpc.Core;
 using System;
+using System.Threading.Tasks;
 
 namespace Google.Cloud.Spanner.Data.CommonTesting;
 
@@ -27,6 +29,8 @@ namespace Google.Cloud.Spanner.Data.CommonTesting;
 /// </summary>
 public abstract class SpannerTestDatabaseBase
 {
+    private MultiplexSession _multiplexSession;
+
     /// <summary>
     /// The Spanner Host name to connect to. It is read from the environment variable "TEST_SPANNER_HOST".
     /// </summary>
@@ -178,4 +182,26 @@ public abstract class SpannerTestDatabaseBase
             SessionPoolManager = SessionPoolManager.Create(new V1.SessionPoolOptions(), logger),
             LogCommitStats = logCommitStats
         });
+
+    public async Task<MultiplexSession> GetMultiplexSession()
+    {
+        if (_multiplexSession != null)
+        {
+            return _multiplexSession;
+        }
+
+        var options = new MultiplexSessionOptions();
+
+        _multiplexSession = await CreateMultiplexSession(options).ConfigureAwait(false);
+
+        return _multiplexSession;
+    }
+
+    private async Task<MultiplexSession> CreateMultiplexSession(MultiplexSessionOptions options)
+    {
+        var poolManager = SessionPoolManager.Create(options);
+        var muxSession = await poolManager.AcquireMultiplexSessionAsync(SpannerClientCreationOptions, DatabaseName, null);
+
+        return muxSession;
+    }
 }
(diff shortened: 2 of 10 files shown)


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -0,0 +1,752 @@
...
+        /// <remarks>
+        /// Will be set iif a transaction has been started and <see cref="TransactionOptions"/> represents read-only options
+        /// with <see cref="Types.ReadOnly.ReturnReadTimestamp"/> set to true
+        /// or if this session was created using <see cref="SessionPool.CreateDetachedSession(SessionName, ByteString, ModeOneofCase, Timestamp)"/>
+        /// and a value was provided for the <see cref="Timestamp"/> parameter.
+        /// </remarks>
+        public Timestamp ReadTimestamp => Interlocked.CompareExchange(ref _transaction, null, null)?.ReadTimestamp;
+
+        // internal for testing
+        internal MultiplexedSessionPrecommitToken PrecommitToken { get; set; }
```
amanda-tarafa, line 112: Let's make this a variable of the class. I'll find tests that mock the service and check the request.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -0,0 +1,752 @@
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="multiplexSession"></param>
+        /// <param name="transactionId"></param>
+        /// <param name="transactionOptions"></param>
+        /// <param name="singleUseTransaction"></param>
+        /// <param name="readTimestamp"></param>
+        public ManagedTransaction(ManagedSession multiplexSession, ByteString transactionId, TransactionOptions transactionOptions, bool singleUseTransaction, Timestamp readTimestamp)
```
amanda-tarafa, line 122: Private constructors and factory methods, this is to protect us from future breaking changes.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -0,0 +1,752 @@
...
+                // We don't need to lock here, this is the constructor.
+                _transaction = new Transaction
+                {
+                    Id = transactionId,
+                    ReadTimestamp = readTimestamp,
+                };
+            }
+        }
+
+        internal ManagedTransaction WithTransaction(ByteString transactionId, TransactionOptions transactionOptions, bool singleUseTransaction, Timestamp readTimestamp = null)
```
amanda-tarafa, line 153: This should be moved to ManagedSession now that we have a single session.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -0,0 +1,752 @@
...
+        {
+            return new ManagedTransaction(_multiplexSession, transactionId, transactionOptions, singleUseTransaction, readTimestamp);
+        }
```
amanda-tarafa, line 156:
```suggestion
        internal ManagedTransaction WithTransaction(ByteString transactionId, TransactionOptions transactionOptions, bool singleUseTransaction, Timestamp readTimestamp = null) =>
        new ManagedTransaction(_multiplexSession, transactionId, transactionOptions, singleUseTransaction, readTimestamp);
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -0,0 +1,752 @@
...
+
+        private Mutation MaybeFetchMutationKey(Protobuf.Collections.RepeatedField<Mutation> mutations)
+        {
+            if(mutations.Count < 1)
+            {
+                return null;
+            }
+
+            // Fetch first mutation in list, we will return the first mutation by default if it meets all conditions
+            Mutation key = mutations.ElementAt(0);
```
amanda-tarafa, line 166: Something like this:

```suggestion
            mutations.Where(m => m.KeySet.Keys.Count > 0 || m.Delete?.KeySet.Kets.Count > 0)
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -0,0 +1,752 @@
...
+                };
+            }
+        }
+
+        internal ManagedTransaction WithTransaction(ByteString transactionId, TransactionOptions transactionOptions, bool singleUseTransaction, Timestamp readTimestamp = null)
+        {
+            return new ManagedTransaction(_multiplexSession, transactionId, transactionOptions, singleUseTransaction, readTimestamp);
+        }
+
+        private Mutation MaybeFetchMutationKey(Protobuf.Collections.RepeatedField<Mutation> mutations)
```
amanda-tarafa, line 158: Make this a local method to the CommitAsync method.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -0,0 +1,752 @@
...
+using static Google.Cloud.Spanner.V1.TransactionOptions;
+
+namespace Google.Cloud.Spanner.V1
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public partial class ManagedTransaction
+    {
+        private readonly ManagedSession _multiplexSession;
```
amanda-tarafa, line 33: I've been thinking of something that will benefit the following use cases:

- Detached transactions
- Transaction that are ongoing while a ManagedSessions refreshes.

Instead of keeping the ManagedSession reference here, let's keep just the Session (the Spanner session) refrerence and the SpannerClient reference.
Instead of checking/attempting to refresh the session from within the ManagedTransaction, we do it internally from the ManagedSession, we add there methods to obtain the ManagedTransaction and in those methods we check/refresh the session.

Let's chat about it.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,272 @@
...
+
+    private async Task<bool> UpdateMuxSession(bool needsRefresh, double intervalInDays)
+    {
+        Session oldSession = _session;
+        await CreateOrRefreshSessionsAsync(default).ConfigureAwait(false);
+
+        return _session != oldSession;
+    }
+
+    internal async Task MaybeRefreshWithTimePeriodCheck()
```
amanda-tarafa, line 120: I forgot to mention, I think it's best if we save this task and check on it with a lock instead of using a semaphore (the same we do for creating a transaction in ManagedTransaction). The semaphore is stopping several refreshes to be "running" at once, but it's not stopping several refreshes to be "triggered" at once, meaning we may have a succession of refreshes one after the other.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
purva9413: We'll move this to ManagedSession and there is a check on the Session refresh so we can no longer use this syntax. Resolving.

amanda-tarafa:
Closing in favor of #15322 and #15335

PR #15122, opened 2025-10-01
feat: Spanner.V1 BatchWrite

robertvoinescu-work:
User need access to the `BatchWrite` API through the Spanner SDK. This is important because `BatchWrite` allows users to submit a set of bulk mutations in batch for efficient processing. To address this, We introduce a `BatchWriteAsync` method in `PooledSession` exposing the API in Spanner.V1.  

### Details
* BatchWrite does not exist in a transaction managed by the client - all transaction logic is server side. So, the `BatchWriteAsync` method avoids transaction management.
* Integration tests have been added
* Unit testing has been avoided since the returned type from `SpannerClient.BatchWrite`, `AsyncStream<>`, is a sealed class (i.e. cannot be mocked) with no public constructor (i.e. cannot be explicitly created in our tests). We could introduce wrappers to allow for mocking, but that adds a lot of abstraction layers I'm not sure we want.

### Non Goals
* Incorporate spanner into the Spanner ADO layer, this will be handled by another PR.


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -850,6 +853,37 @@ public Task<ExecuteBatchDmlResponse> ExecuteBatchDmlAsync(ExecuteBatchDmlRequest
...
+        /// <param name="request">The batch write request. Must not be null.</param>
+        /// <param name="callSettings">If not null, applies overrides to this RPC call.</param>
+        /// <returns>An asynchronous stream of <see cref="BatchWriteResponse"/> messages.</returns>
+        public async IAsyncEnumerable<BatchWriteResponse> BatchWriteAsync(BatchWriteRequest request, CallSettings callSettings)
+        {
+            CheckNotDisposed();
+            GaxPreconditions.CheckNotNull(request, nameof(request));
+            request.SessionAsSessionName = SessionName;
+
+            SpannerClient.BatchWriteStream stream = Client.BatchWrite(request, callSettings);
```
purva9413: We need to check Session expiry here. We have a helper method to check that for us, see how `RecordSuccessAndExpiredSessions` is used.

purva9413: I am also guessing we will need most of the plumbing we have for all the other RPCs, basically the calls into `ExecuteMaybeWithTransactionSelectorAsync` to select the transaction selector etc. I will also go through your design doc to see if I am missing any pieces here.

robertvoinescu-work: Thanks, I've updated the code to use `StreamResponsesAndRecordSuccessAsync`. The current implementations checks the session expiry during iteration through the stream (like how [`ResultStream`](https://github.com/googleapis/google-cloud-dotnet/blob/da24ec04595e8f2b8e85bfcf34848feba9da747f/apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ResultStream.cs#L176) currently does).

robertvoinescu-work: I've gone through the code of `ExecuteMaybeWithTransactionSelectorAsync` and made sure there are no other plumbing bits that need to be added here. 

Only as a note to future viewers on why we don't use the `ExecuteMaybeWithTransactionSelectorAsync` directly here or even extend it. `BatchWrite` is a unique scenario since it doesn't work within an explicit transaction. Spanner opaquely batches them up on their end into transactions. No transaction is passed and no transaction is returned, so we can't we set the transaction in a pooled session with a `TransactionSelector`.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -850,6 +853,25 @@ public Task<ExecuteBatchDmlResponse> ExecuteBatchDmlAsync(ExecuteBatchDmlRequest
             Transaction GetInlinedTransaction(ExecuteBatchDmlResponse response) => response?.ResultSets?.FirstOrDefault()?.Metadata?.Transaction;
         }
 
+        /// <summary>
+        /// Executes a BatchWrite RPC asynchronously, returning a stream of responses.
+        /// </summary>
+        /// <remarks>
+        /// This operation does not participate in the session's active transaction.
```
amanda-tarafa: Is this true? Or does this operation requires a session with no active transaction?
Even if this is true, would this operation also work with a session with no active transaction?

The answer to these questions is fundamental to how we implement this, so I'll wait for the answers before continuing the review.

robertvoinescu-work: I will need to explicitly verify with spanner team, but I've tested this against my own spanner instance (non-emulated):
1. `BatchWrite` succeeds when working on a fresh session with no active transaction.
2. `BatchWrite` succeeds when working on a session on which I've begun a transaction.

Will verify with spanner team since transaction management is opaque to the caller for `BatchWrite`.

amanda-tarafa: For 2. what happens to that transaction, any changes made within it, etc.?

robertvoinescu-work: I tried to use that transaction for read-write and received the following. It looks like batch write invalidates the currently active transaction
```
Error Message: Grpc.Core.RpcException : Status(StatusCode="FailedPre
      condition", Detail="This transaction has been invalidated by a later transaction in the same session."
```

amanda-tarafa: OK, so if using a session with a transaction breaks that transaction, let's not allowed that. BatchWrite should only run on a session with no transaction, that is a pooled session with `TransactionMode.None`. The way to validate this is to go through `ExecuteMaybeWithTransaction` and fail on all the callbacks that attempt to either set or extract a transaction from the BatchWrite request/response. Let me know if you want to chat about that. Else, let me know when the code reflects these changes by re-requesting my review. Thanks.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -391,6 +394,35 @@ Task<Transaction> BeginTransactionAsync(CancellationToken cancellationToken)
...
+        /// <param name="command">The command (RPC) that will be executed after the transaction selector has maybe been set.
+        /// Callers may fail if command is executed but no transaction selector has been set.</param>
+        /// <param name="inlinedTransactionExtractor">Will extract transaction information from the command's response if
+        /// transaction inlining was succesful. May be null, which indicates that command does not support transaction inlining.</param>
+        /// <param name="skipTransactionCreation">If true, transaction creation may be skipped. This is used by commit and rollback
+        /// so that a transaction is not created just for inmediate commit or rollback. Note that if there are pending mutations, commit
+        /// should set this parameter to false.</param>
+        /// <param name="cancellationToken">The cancellation token for the operation.</param>
+        /// <returns>A task whose result will be the result from having executed <paramref name="command"/>.</returns>
+        internal TResponse ExecuteMaybeWithTransactionSelector<TResponse>(
```
amanda-tarafa: We don't need a sync version of this method. If the actual exection is syncrhonous, the async method will return synchronously anyway, and here, . Please undo and call the async version instead.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -850,6 +882,42 @@ public Task<ExecuteBatchDmlResponse> ExecuteBatchDmlAsync(ExecuteBatchDmlRequest
...
 
+        /// <summary>
+        /// Executes a BatchWrite RPC asynchronously, returning a stream of responses.
+        /// </summary>
+        /// <remarks>
+        /// This operation can only be run when the session is configured to use no transaction.
+        /// </remarks>
+        /// <param name="request">The batch write request. Must not be null.</param>
+        /// <param name="callSettings">If not null, applies overrides to this RPC call.</param>
+        /// <returns>An asynchronous stream of <see cref="BatchWriteResponse"/> messages.</returns>
```
amanda-tarafa:
```suggestion
        /// <returns>An asynchronous stream of <see cref="BatchWriteResponse"/> objects.</returns>
```

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -850,6 +882,42 @@ public Task<ExecuteBatchDmlResponse> ExecuteBatchDmlAsync(ExecuteBatchDmlRequest
...
+        public IAsyncEnumerable<BatchWriteResponse> BatchWriteAsync(BatchWriteRequest request, CallSettings callSettings)
+        {
+            CheckNotDisposed();
+            GaxPreconditions.CheckNotNull(request, nameof(request));
+            // Fail early to prevent the creation of an unused transaction on the session.
+            GaxPreconditions.CheckState(TransactionOptions.ModeCase == ModeOneofCase.None, $"{nameof(BatchWriteAsync)} can only be used when the session is configured to use no transaction.");
```
amanda-tarafa: We don't need this, the ExecuteMaybeWithTransactionSelector will fail early enough (i.e. without making any RPC) if we configure the right callbacks. See below.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -850,6 +882,42 @@ public Task<ExecuteBatchDmlResponse> ExecuteBatchDmlAsync(ExecuteBatchDmlRequest
...
+                callSettings?.CancellationToken ?? default);
+
+            IAsyncEnumerable<BatchWriteResponse> BatchWriteAsync()
+            {
+                var stream = Client.BatchWrite(request, callSettings);
+                return StreamResponsesAndRecordSuccessAsync(stream.GetResponseStream());
+            }
+
+            // Batch write gRPC manages the transaction itself, if we are trying to explicitly set a transaction or inline, this is a bug.
+            void SetCommandTransaction(TransactionSelector transactionSelector) => throw new InvalidOperationException($"{nameof(BatchWriteAsync)} must be executed with no transaction. If you see this there is a bug in the code.");
```
amanda-tarafa: This one is sufficient to fail early. This would be our validation point, BatchWrite has not caused any RPCs to be executed at this point.
Message, or similar:
"BatchWrite does not support explicitly set transactions, it won't be executed within an active transaction. The sessions most be acquired with nameof(ModeOneOfCase.TransactionNone) or BatchWrite should complete execution before any other commands that would create a transaction are executed."

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -850,6 +882,42 @@ public Task<ExecuteBatchDmlResponse> ExecuteBatchDmlAsync(ExecuteBatchDmlRequest
...
+            Transaction GetInlinedTransaction(IAsyncEnumerable<BatchWriteResponse> response) => throw new InvalidOperationException($"{nameof(BatchWriteAsync)} cannot be executed with an inlined transaction. If you see this there is a bug in the code.");
```
amanda-tarafa: I think this delegate may be set to null on the paremeter call. That's a signal to the "ExecuteMaybeWithTransactionAsync" that this method does not support inlining. Because we are also asking it to skip transaction creation, this will never be called.

SetCommandTransaction will be called is some other method created a transaction and that's when we want to fail.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
```diff
@@ -876,6 +944,22 @@ private async Task RecordSuccessAndExpiredSessions(Task task)
             UpdateRefreshTime();
         }
 
+        private async IAsyncEnumerable<TResponse> StreamResponsesAndRecordSuccessAsync<TResponse>(AsyncResponseStream<TResponse> responseStream)
```
amanda-tarafa, line 918: I'm not sure we want to do this on V1. AsyncResponseStream is a wrapper that works for both pre and post C# 8 users. And we don't want to loose that just so we can keep better track of the refresh time/expired session.

I think we should not wrap the response, and then, to make certain that we don't return a bad session to the pool, after the succesful call to BatchWrite, before returning the AsyncResponseStream, I'd update the refresh time to now, to make certain the session will be refreshed before releasing.

But, wrting about all of this, and knowing that we are about to release MUX, I think we should pause. Let's chat.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
robertvoinescu-work: Done.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
robertvoinescu-work: Good point, fixed.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
robertvoinescu-work: Done.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
robertvoinescu-work: This is a good point, I've updated the code to return `AsyncResponseStream` instead. Thanks for the info on best practice :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/PooledSession.cs
robertvoinescu-work: Removed.

Issue #15107, opened 2025-09-19
Google.Cloud.Firestore: Allow manually invoking object creation from raw fields to support deserializing nested polymorphic collections


Autoquark:
**Is your feature request related to a problem? Please describe.**
I'm trying to serialize and retrieve C# objects via firestore where the objects may be of various types derived from a common base class. This works great if I have a separate firestore document for each object I want to store - I can query for the object, get a DocumentSnapshot, examine a special field which acts as a type discriminator, then call ConvertTo and pass the appropriate type (side note, it would be nice if there was an overload that takes the Type as a plain parameter rather than a type parameter, at the moment I have to use reflection to specify the type parameter at runtime).

Example class:

```
[FirestoreData]
public abstract class PolymorphicFirestoreModel : FirestoreModel
{
    protected PolymorphicFirestoreModel()
    {
        SubtypeName = GetType().FullName!;
    }
    
    [FirestoreProperty]
    public string SubtypeName { get; set; }
}

class Derived1 : PolymorphicFirestoreModel { ... }
class Derived2 : PolymorphicFirestoreModel { ... }
```

However, what I would also like to be able to do is have a c# object containing a collection which has objects of multiple types, and be able to round trip that via firestore and restore each object in the nested collection as the appropriate type. Unfortunately that isn't so easy because the only way I can invoke the library's functionality to convert a firestore document into a c# object is via a DocumentSnapshot, I can't just give it a Dictionary<string, object> even though I can see that's basically what it operates on internally. Example of what I'd like to serialize:

```
[FirestoreData]
public abstract class ModelWithNestedCollection
{
    // Could contain objects of various different types
    [FirestoreProperty]
    public IList<PolymorphicFirestoreModel> NestedObjects { get; set; }
}
```

 **Describe the solution you'd like**
A way to invoke the library's functionality for converting a raw set of named fields into a c# object of a specified class without needing a DocumentSnapshot.

 **Describe alternatives you've considered**
At the moment I handle these subcollections by making them into a List<DocumentReference> referencing root documents in a separate firestore collection, which I can then deserialise into the appropriate type as described above. This works, but it means I'm forced to split out data into separate documents that should really all live in the same place as I always want to fetch it all at the same time, and makes it difficult to do things like duplicate a document as I have to duplicate all the referenced documents as well.

jskeet:
[...]

PR #15126, opened 2025-10-02
feat: implement support for Firestore type discriminators

jskeet:
Note that I figured this was a sufficiently advanced use case that it would be okay to pass in the map in "Firestore format" (Value) rather than creating a POCO.

This is a prototype for #15107

cc @Autoquark

### DIFF
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Snippets/DataModelSnippets.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Snippets/DataModelSnippets.cs
@@ -417,7 +417,9 @@ namespace Google.Cloud.Firestore.Snippets
         public abstract class Shape
         {
             // We happen to use the short name of the type in all
-            // implementations, but 
+            // implementations, but this is just a matter of convenience.
+            // The type discriminator is not restricted in what fields
+            // of the value are used, or their types, or their values.
             [FirestoreProperty]
             public abstract string Type { get; }
         }
--- a/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ConverterRegistry.cs
+++ b/apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ConverterRegistry.cs
@@ -42,7 +42,7 @@ namespace Google.Cloud.Firestore
         /// Adds the given converter to the registry.
         /// </summary>
         /// <typeparam name="T">The type that <paramref name="converter"/> converts to/from.</typeparam>
-        /// <param name="converter">The converter to add.</param>
+        /// <param name="converter">The converter to add. Must not be null.</param>
         /// <exception cref="ArgumentException">There is already a converter in the registry for the given type.</exception>
         public void Add<T>(IFirestoreConverter<T> converter)
         {
@@ -52,10 +52,11 @@ namespace Google.Cloud.Firestore
         }
 
         /// <summary>
-        /// 
+        /// Adds the given type discriminator to the registry.
         /// </summary>
-        /// <typeparam name="T"></typeparam>
-        /// <param name="discriminator"></param>
+        /// <typeparam name="T">The type for which <paramref name="discriminator"/> can determine a concrete type.</typeparam>
+        /// <param name="discriminator">The type discriminator to add. Must not be null.</param>
+        /// <exception cref="ArgumentException">There is already a type discriminator in the registry for the given type.</exception>
         public void Add<T>(IFirestoreTypeDiscriminator<T> discriminator)
         {
             GaxPreconditions.CheckNotNull(discriminator, nameof(discriminator));


amanda-tarafa:
@Autoquark, @jskeet and I have been chatting and there's an alternative to this proposal, that's very similar but will help with keeping your types clean of any discriminator indicator, which in turn may help users with hierarchies of types they don't control. It's a little bit more involved library side, so we'd still need to protoype it and see if it looks good.

It goes something like this:

- Instead of registering a type discriminator for a base type, you register the base type for "hierarchical serialization".
- When serializing, for all instances of types in such registered hierarchies the library will include a special named field with the type full name.
- On deserialization of these registered types we load (and cache) the type specified in the special field from any of the already loaded assemblies, if it's there, otherwise we just deserialize as we've done before.
- Calling code (you) need to register the base type and make sure assemblies with descendant types are loaded. We can even give you an interface for this, something like  `ConverterRegistry.RegisterHierarchy(Type baseType, params Type[] descendantTypes)`. When you call that with all known hierarchy types, that forces the assemblies to be loaded.

Let us know which approach seesm more suitable for you.

Autoquark:
Hi, thanks so much for looking into this. Both approaches would work fine for me as I'm only currently interested in serialization of types I control, but I can the advantage of the second in general.

I suppose one advantage of the first approach is that the decision of what type to deserialize as can be more flexible since you can put any logic you like in the type discriminator implementation - again, not something I currently need, but maybe nice to have? If nothing else, it would allow you to change your C# class hierarchy and then still be able to load old data from firestore by having your type discriminator map the old class names to whatever your new equivalent might be. With the second approach you don't have any control over this process so once you have a load of data in your DB that you need to maintain compatibility with, you are kind of locked in.

I should probably mention that this is just a small hobby project of mine so please don't take my requirements as representative of a typical massive corporate software project or anything :)

jskeet:
Okay, this is now ready for review, with docs/snippets and some more tests.

When writing up the docs, I found there's an ambiguity between the type discriminator being the data, or the code that decides what to do with that data. Happy to rename one of them if that would be useful.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ConverterRegistry.cs
```diff
@@ -47,6 +51,17 @@ public void Add<T>(IFirestoreConverter<T> converter)
             _converterList.Add(converter);
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="discriminator"></param>
```
amanda-tarafa: Missing docs.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ConverterRegistry.cs
```diff
@@ -47,6 +51,17 @@ public void Add<T>(IFirestoreConverter<T> converter)
...
+        public void Add<T>(IFirestoreTypeDiscriminator<T> discriminator)
```
amanda-tarafa, line 60: And overload that accepts the function as well?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Snippets/DataModelSnippets.cs
```diff
@@ -409,5 +411,103 @@ public class Company
             public (string city, string state, string country) Location { get; set; }
         }
         // End sample
+
+        // Sample: DiscriminatedType
+        [FirestoreData]
+        public abstract class Shape
+        {
+            // We happen to use the short name of the type in all
+            // implementations, but
```
amanda-tarafa: but what? :)

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ConverterRegistry.cs
jskeet: As a delegate? That's an interesting idea. It would certainly work... we could even remove the interface entirely, and just have the delegate. What do you think?

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ConverterRegistry.cs
jskeet: Fixed.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore.Snippets/DataModelSnippets.cs
jskeet: Fixed.

jskeet:
Added an extra comment for the docs, but I'm starting to wonder about the possibility of removing the interface entirely. There are definitely pros and cons either way.

apis/Google.Cloud.Firestore/Google.Cloud.Firestore/ConverterRegistry.cs
amanda-tarafa: For anyone needing to hold some sort of state used for discriminating, the interface is good, although implementing the delegate with an instance method is as good in terms of what it can achieve, but maybe not as evident?

Mind you, it doesn't even need to be a defined delegate type, you can just accept `Func<IDictionary<string, Value>, Type>`. That's possibly the looser approach? I'm fine with that, for the simple cases, users can then implement that in place.

jskeet:
Decided not to do the overload accepting a delegate right now. But we can do so later.

Issue #13815, opened 2024-11-13
Feature Request: Integrate Microsoft.Extensions.AI into Google.Cloud.AIPlatform


Krzysztof318:
## Feature Request: Integrate Microsoft.Extensions.AI with Google.Cloud.AIPlatform

**Is your feature request related to a problem? Please describe.**

While Google Cloud SDK for .NET and specifically the `Google.Cloud.AIPlatform` library provide robust tools for interacting with Google AI services, the current implementations lack a unified abstraction layer compatible with Microsoft.Extensions.AI.

**Describe the solution you'd like**

I propose integrating Microsoft.Extensions.AI with `Google.Cloud.AIPlatform` (v1/v1beta). This library offers a unified API abstraction for AI services, similar to logging and dependency injection abstractions in .NET. By leveraging Microsoft.Extensions.AI, we can achieve:

* Aligned development experience: Developers can utilize familiar abstractions, streamlining the integration of Google AI services.
* Enhanced flexibility: Developers can easily switch between Google AI services and other providers with minimal code changes.
* Improved code maintainability: A consistent abstraction layer promotes the creation of modular and reusable AI components.

**Additional context**

* Microsoft.Extensions.AI announcement: [Microsoft DevBlogs](https://devblogs.microsoft.com/dotnet/introducing-microsoft-extensions-ai-preview/)
* Real-world example: Implementation of Microsoft.Extensions.AI in AWS SDK for .NET [AWS Pull Request](https://github.com/aws/aws-sdk-net/pull/3545)
* Initial PR as a concept: #13814 

## Questions to the Google Team

1. **Potential challenges:** What are the potential challenges and trade-offs of adopting this approach in your repo?
2. **Preferred implementation:** What would be the preferred approach for implementing this integration? Would it involve creating new project or extend existing AIPlatform projects?
It would be best option to be consistent with Microsoft.Extensions.AI idea but I am open to suggestions from your side.

I'm eager to discuss this proposal further and go work on pull request when we will agree on the details.

I believe integrating Microsoft.Extensions.AI with `Google.Cloud.AIPlatform` would significantly benefit the .NET developer community and foster a more unified approach to AI development on Google Cloud.

cc: @stephentoub

amanda-tarafa:
@Krzysztof318 thanks! We need to figure out a few things internally, so we need a couple of days. I'll keep this issue updated.

amanda-tarafa:
Just here to say, we are still looking into this.

[...]

PR #15134, opened 2025-10-06
feat: Add AsIChatClient/EmbeddingGenerator/ImageGenerator for PredictionServiceClient

stephentoub:
e.g.

**Using IChatClient with function calling**:

https://github.com/user-attachments/assets/1a4dd179-f942-47e6-8a5d-bc1a65ac1d46

**Generating embeddings**:

image

**Generating images**:

image

Fixes https://github.com/googleapis/google-cloud-dotnet/issues/13815

cc: @jskeet, @amanda-tarafa, @Krzysztof318

### DIFF
--- a/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/Google.Cloud.VertexAI.Extensions.IntegrationTests.csproj
+++ b/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/Google.Cloud.VertexAI.Extensions.IntegrationTests.csproj
@@ -4,8 +4,7 @@
     <TargetFrameworks>net8.0;net462</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net8.0</TargetFrameworks>
     <IsPackable>false</IsPackable>
-    <NoWarn>1701;1702;1705;xUnit2004;xUnit2013;MEAI001</NoWarn>
-    <Nullable>enable</Nullable>
+    <NoWarn>$(NoWarn);1701;1702;1705;xUnit2004;xUnit2013</NoWarn>
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="..\..\..\tools\Google.Cloud.ClientTesting\Google.Cloud.ClientTesting.csproj" />
--- a/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/VertexAIExtensionsTest.cs
+++ b/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/VertexAIExtensionsTest.cs
@@ -13,8 +13,8 @@
 // limitations under the License.
 
 using Google.Cloud.AIPlatform.V1;
+using Google.Cloud.ClientTesting;
 using Microsoft.Extensions.AI;
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
@@ -24,15 +24,13 @@ namespace Google.Cloud.VertexAI.Extensions.Tests;
 
 public class VertexAIExtensionsTest
 {
-    private static readonly string? s_projectId = Environment.GetEnvironmentVariable("TEST_PROJECT");
-    private static readonly string? s_location = Environment.GetEnvironmentVariable("TEST_PROJECT_LOCATION");
+    private static readonly string? s_projectId = TestEnvironment.GetTestProjectId();
+    private static readonly string? s_location = TestEnvironment.GetProjectLocationId();
 
     [Fact]
     public async Task AsIChatClient_BasicRequestResponse_NonStreaming()
     {
-        (string projectId, string location) = EnsureConfiguration();
-
-        IChatClient client = (await CreateClientAsync()).AsIChatClient(EndpointName.FormatProjectLocationPublisherModel(projectId, location, "google", "gemini-2.5-pro"));
+        IChatClient client = (await CreateClientAsync()).AsIChatClient(EndpointName.FormatProjectLocationPublisherModel(s_projectId, s_location, "google", "gemini-2.5-pro"));
         Assert.NotNull(client);
 
         var response = await client.GetResponseAsync("Hello");
@@ -43,29 +41,18 @@ public class VertexAIExtensionsTest
     [Fact]
     public async Task AsIChatClient_BasicRequestResponse_Streaming()
     {
-        (string projectId, string location) = EnsureConfiguration();
-
-        IChatClient client = (await CreateClientAsync()).AsIChatClient(EndpointName.FormatProjectLocationPublisherModel(projectId, location, "google", "gemini-2.5-pro"));
+        IChatClient client = (await CreateClientAsync()).AsIChatClient(EndpointName.FormatProjectLocationPublisherModel(s_projectId, s_location, "google", "gemini-2.5-pro"));
         Assert.NotNull(client);
 
-        List<ChatResponseUpdate> updates = [];
-        await foreach (var update in client.GetStreamingResponseAsync("Hello"))
-        {
-            updates.Add(update);
-        }
-
-        Assert.NotEmpty(updates);
-
-        ChatResponse response = updates.ToChatResponse();
+        ChatResponse response = await client.GetStreamingResponseAsync("Hello").ToChatResponseAsync();
+        Assert.NotNull(response);
         Assert.NotEmpty(response.Text);
     }
 
     [Fact]
     public async Task AsIEmbeddingGenerator_EmbedSeveralInputs()
     {
-        (string projectId, string location) = EnsureConfiguration();
-
-        IEmbeddingGenerator<string, Embedding<float>> generator = (await CreateClientAsync()).AsIEmbeddingGenerator(EndpointName.FormatProjectLocationPublisherModel(projectId, location, "google", "gemini-embedding-001"));
+        IEmbeddingGenerator<string, Embedding<float>> generator = (await CreateClientAsync()).AsIEmbeddingGenerator(EndpointName.FormatProjectLocationPublisherModel(s_projectId, s_location, "google", "gemini-embedding-001"));
         Assert.NotNull(generator);
 
         GeneratedEmbeddings<Embedding<float>> embeddings = await generator.GenerateAsync(["Hello", "World"]);
@@ -81,9 +68,7 @@ public class VertexAIExtensionsTest
     [Fact]
     public async Task AsIImageGenerator_GenerateImage()
     {
-        (string projectId, string location) = EnsureConfiguration();
-
-        IImageGenerator generator = (await CreateClientAsync()).AsIImageGenerator(EndpointName.FormatProjectLocationPublisherModel(projectId, location, "google", "imagen-4.0-fast-generate-001"));
+        IImageGenerator generator = (await CreateClientAsync()).AsIImageGenerator(EndpointName.FormatProjectLocationPublisherModel(s_projectId, s_location, "google", "imagen-4.0-fast-generate-001"));
         Assert.NotNull(generator);
 
         ImageGenerationResponse response = await generator.GenerateImagesAsync("A cute baby sea otter");
@@ -92,16 +77,6 @@ public class VertexAIExtensionsTest
         DataContent image = Assert.Single(response.Contents.OfType<DataContent>());
         Assert.Equal("image/png", image.MediaType);
         Assert.InRange(image.Data.Length, 1, int.MaxValue);
-    }
-
-    private static (string ProjectId, string Location) EnsureConfiguration()
-    {
-        if (string.IsNullOrEmpty(s_projectId) || string.IsNullOrEmpty(s_location))
-        {
-            throw new InvalidOperationException("Set the TEST_PROJECT and TEST_PROJECT_LOCATION environment variables to run these tests.");
-        }
-
-        return (s_projectId!, s_location!);
     }
 
     private static Task<PredictionServiceClient> CreateClientAsync() =>
--- a/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
+++ b/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
@@ -14,6 +14,7 @@
 
 using Google.Cloud.AIPlatform.V1;
 using Google.Protobuf;
+using Google.Protobuf.WellKnownTypes;
 using Grpc.Core;
 using Microsoft.Extensions.AI;
 using System;
@@ -23,6 +24,7 @@ using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
+using Struct = Google.Protobuf.WellKnownTypes.Struct;
 
 namespace Google.Cloud.VertexAI.Extensions.Tests;
 
@@ -65,7 +67,7 @@ public class AsIChatClientTest
         Assert.NotNull(metadata);
         Assert.Equal("gcp.vertex_ai", metadata.ProviderName);
         Assert.Equal(new("https://aiplatform.googleapis.com:443"), metadata.ProviderUri);
-        Assert.Equal(defaultModelId, metadata.DefaultModelId);
+        Assert.Equal(defaultModelId?.Substring(defaultModelId.LastIndexOf('/') + 1), metadata.DefaultModelId);
     }
 
     [Fact]
@@ -86,17 +88,7 @@ public class AsIChatClientTest
     {
         DelegateCallInvoker invoker = new()
         {
-            OnGenerateContentRequest = request =>
-            {
-                GenerateContentResponse response = new();
-
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "Hello" } } }
-                });
-
-                return response;
-            }
+            OnGenerateContentRequest = request => CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "Hello" } } }),
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
@@ -118,13 +110,8 @@ public class AsIChatClientTest
                 Assert.Single(request.Contents[0].Parts);
                 Assert.Equal("What is the weather like?", request.Contents[0].Parts[0].Text);
 
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "It's sunny today." } } },
-                    FinishReason = Candidate.Types.FinishReason.Stop
-                });
-
+                GenerateContentResponse response = CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "It's sunny today." } } });
+                response.Candidates[0].FinishReason = Candidate.Types.FinishReason.Stop;
                 return response;
             }
         };
@@ -136,8 +123,7 @@ public class AsIChatClientTest
 
         Assert.NotNull(result);
         Assert.Equal(ChatRole.Assistant, result.Messages[0].Role);
-        Assert.Single(result.Messages[0].Contents);
-        Assert.IsType<TextContent>(result.Messages[0].Contents[0]);
+        Assert.IsType<TextContent>(Assert.Single(result.Messages[0].Contents));
         Assert.Equal("It's sunny today.", ((TextContent) result.Messages[0].Contents[0]).Text);
         Assert.Equal(ChatFinishReason.Stop, result.FinishReason);
     }
@@ -160,20 +146,14 @@ public class AsIChatClientTest
                 Assert.Equal("image/png", request.Contents[0].Parts[1].InlineData.MimeType);
                 Assert.True(request.Contents[0].Parts[1].InlineData.Data.ToByteArray().SequenceEqual(imageData));
 
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "I see an image." } } }
-                });
-
-                return response;
+                return CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "I see an image." } } });
             }
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
         ChatMessage[] messages =
         [
-            new ChatMessage(ChatRole.User,
+            new(ChatRole.User,
             [
                 new TextContent("Describe this image"),
                 new DataContent(imageData, "image/png")
@@ -204,20 +184,14 @@ public class AsIChatClientTest
                 Assert.Equal("https://example.com/image.jpg", request.Contents[0].Parts[1].FileData.FileUri);
                 Assert.Equal("image/jpeg", request.Contents[0].Parts[1].FileData.MimeType);
 
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "Image analysis complete." } } }
-                });
-
-                return response;
+                return CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "Image analysis complete." } } });
             }
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
         ChatMessage[] messages =
         [
-            new ChatMessage(ChatRole.User,
+            new(ChatRole.User,
             [
                 new TextContent("Analyze this image"),
                 new UriContent(imageUri, "image/jpeg")
@@ -235,30 +209,22 @@ public class AsIChatClientTest
     {
         DelegateCallInvoker invoker = new()
         {
-            OnGenerateContentRequest = request =>
-            {
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
+            OnGenerateContentRequest = request => CreateResponse(
+                new()
                 {
-                    Content = new Content()
+                    Role = "model",
+                    Parts =
                     {
-                        Role = "model",
-                        Parts =
+                        new Part()
                         {
-                            new Part()
+                            FileData = new()
                             {
-                                FileData = new FileData()
-                                {
-                                    FileUri = "not a valid uri",
-                                    MimeType = "image/jpeg"
-                                }
+                                FileUri = "not a valid uri",
+                                MimeType = "image/jpeg"
                             }
                         }
                     }
-                });
-
-                return response;
-            }
+                }),
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
@@ -272,30 +238,22 @@ public class AsIChatClientTest
     {
         DelegateCallInvoker invoker = new()
         {
-            OnGenerateContentRequest = request =>
-            {
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
+            OnGenerateContentRequest = request => CreateResponse(
+                new()
                 {
-                    Content = new Content()
-                    {
-                        Role = "model",
-                        Parts =
+                    Role = "model",
+                    Parts =
                         {
                             new Part()
                             {
-                                FunctionCall = new FunctionCall
+                                FunctionCall = new()
                                 {
                                     Name = "get_weather",
-                                    Args = Protobuf.WellKnownTypes.Struct.Parser.ParseJson("{\"location\": \"San Francisco\"}")
+                                    Args = Struct.Parser.ParseJson("{\"location\": \"San Francisco\"}")
                                 }
                             }
                         }
-                    }
-                });
-
-                return response;
-            }
+                }),
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
@@ -304,8 +262,7 @@ public class AsIChatClientTest
         ChatResponse result = await chatClient.GetResponseAsync(messages);
 
         Assert.NotNull(result);
-        Assert.Single(result.Messages[0].Contents);
-        Assert.IsType<FunctionCallContent>(result.Messages[0].Contents[0]);
+        Assert.IsType<FunctionCallContent>(Assert.Single(result.Messages[0].Contents));
 
         FunctionCallContent functionCall = (FunctionCallContent) result.Messages[0].Contents[0];
         Assert.Equal("get_weather", functionCall.Name);
@@ -327,20 +284,14 @@ public class AsIChatClientTest
                 Assert.NotNull(request.Contents[0].Parts[0].FunctionResponse);
                 Assert.Equal("call_123", request.Contents[0].Parts[0].FunctionResponse.Name);
 
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "Based on the weather data, it's sunny." } } }
-                });
-
-                return response;
+                return CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "Based on the weather data, it's sunny." } } });
             }
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
         ChatMessage[] messages =
         [
-            new ChatMessage(ChatRole.User,
+            new(ChatRole.User,
             [
                 new FunctionResultContent("call_123", new { temperature = 72, condition = "sunny" })
             ])
@@ -359,15 +310,11 @@ public class AsIChatClientTest
 
         DelegateCallInvoker invoker = new()
         {
-            OnGenerateContentRequest = request =>
-            {
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
+            OnGenerateContentRequest = request => CreateResponse(
+                new()
                 {
-                    Content = new Content()
-                    {
-                        Role = "model",
-                        Parts =
+                    Role = "model",
+                    Parts =
                         {
                             new Part()
                             {
@@ -375,11 +322,7 @@ public class AsIChatClientTest
                                 ThoughtSignature = ByteString.FromBase64(reasoningData)
                             }
                         }
-                    }
-                });
-
-                return response;
-            }
+                }),
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
@@ -388,8 +331,7 @@ public class AsIChatClientTest
         ChatResponse result = await chatClient.GetResponseAsync(messages);
 
         Assert.NotNull(result);
-        Assert.Single(result.Messages[0].Contents);
-        Assert.IsType<TextReasoningContent>(result.Messages[0].Contents[0]);
+        Assert.IsType<TextReasoningContent>(Assert.Single(result.Messages[0].Contents));
 
         TextReasoningContent reasoningContent = (TextReasoningContent) result.Messages[0].Contents[0];
         Assert.Equal(reasoningData, reasoningContent.ProtectedData);
@@ -404,30 +346,26 @@ public class AsIChatClientTest
             {
                 // Verify ChatOptions were applied correctly
                 Assert.Equal("projects/test/locations/us-central1/publishers/google/models/gemini-pro", request.Model);
-                Assert.NotNull(request.GenerationConfig);
-                Assert.Equal(0.7f, request.GenerationConfig.Temperature);
-                Assert.Equal(0.9f, request.GenerationConfig.TopP);
-                Assert.Equal(40, request.GenerationConfig.TopK);
-                Assert.Equal(100, request.GenerationConfig.MaxOutputTokens);
-                Assert.Equal(42, request.GenerationConfig.Seed);
-                Assert.Equal(0.5f, request.GenerationConfig.FrequencyPenalty);
-                Assert.Equal(0.3f, request.GenerationConfig.PresencePenalty);
-                Assert.Contains("STOP", request.GenerationConfig.StopSequences);
-
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
+                Assert.Equal(new()
                 {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "Response with options applied." } } }
-                });
+                    Temperature = 0.7f,
+                    TopP = 0.9f,
+                    TopK = 40,
+                    MaxOutputTokens = 100,
+                    Seed = 42,
+                    FrequencyPenalty = 0.5f,
+                    PresencePenalty = 0.3f,
+                    StopSequences = { "STOP" }
+                }, request.GenerationConfig);
 
-                return response;
+                return CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "Response with options applied." } } });
             }
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/default-model");
         ChatMessage[] messages = [new(ChatRole.User, "Test message")];
 
-        ChatOptions options = new ChatOptions
+        ChatOptions options = new()
         {
             ModelId = "projects/test/locations/us-central1/publishers/google/models/gemini-pro",
             Temperature = 0.7f,
@@ -463,21 +401,15 @@ public class AsIChatClientTest
                 Assert.Single(request.Contents);
                 Assert.Equal("user", request.Contents[0].Role);
 
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "I'm here to help!" } } }
-                });
-
-                return response;
+                return CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "I'm here to help!" } } });
             }
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
         ChatMessage[] messages =
         [
-            new ChatMessage(ChatRole.System, "You are a helpful assistant."),
-            new ChatMessage(ChatRole.User, "Hello")
+            new(ChatRole.System, "You are a helpful assistant."),
+            new(ChatRole.User, "Hello")
         ];
 
         ChatResponse result = await chatClient.GetResponseAsync(messages);
@@ -507,20 +439,14 @@ public class AsIChatClientTest
                 Assert.Contains(request.Tools, t => t.Retrieval != null);
                 Assert.Contains(request.Tools, t => t.GoogleSearch != null);
 
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "I can use tools to help you." } } }
-                });
-
-                return response;
+                return CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "I can use tools to help you." } } });
             }
         };
 
         IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
         ChatMessage[] messages = [new(ChatRole.User, "What tools do you have?")];
 
-        ChatOptions options = new ChatOptions
+        ChatOptions options = new()
         {
             Tools =
             [
@@ -549,13 +475,7 @@ public class AsIChatClientTest
                 Assert.NotNull(request.ToolConfig.FunctionCallingConfig);
                 Assert.Equal(FunctionCallingConfig.Types.Mode.Auto, request.ToolConfig.FunctionCallingConfig.Mode);
 
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "Auto mode enabled." } } }
-                });
-
-                return response;
+                return CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "Auto mode enabled." } } });
             }
         };
 
@@ -565,7 +485,7 @@ public class AsIChatClientTest
         ChatOptions options = new()
         {
             Tools = [new HostedCodeInterpreterTool()],
-            ToolMode = new AutoChatToolMode()
+            ToolMode = ChatToolMode.Auto,
         };
 
         ChatResponse result = await chatClient.GetResponseAsync(messages, options);
@@ -587,13 +507,7 @@ public class AsIChatClientTest
                 Assert.Equal(FunctionCallingConfig.Types.Mode.Any, request.ToolConfig.FunctionCallingConfig.Mode);
                 Assert.Contains("get_weather", request.ToolConfig.FunctionCallingConfig.AllowedFunctionNames);
 
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "Required mode with specific function." } } }
-                });
-
-                return response;
+                return CreateResponse(new() { Role = "model", Parts = { new Part() { Text = "Required mode with specific function." } } });
             }
         };
 
@@ -603,7 +517,7 @@ public class AsIChatClientTest
         ChatOptions options = new()
         {
             Tools = [AIFunctionFactory.Create((string location) => "the weather", "get_weather", "Gets weather information")],
-            ToolMode = new RequiredChatToolMode("get_weather")
+            ToolMode = ChatToolMode.RequireSpecific("get_weather")
         };
 
         ChatResponse result = await chatClient.GetResponseAsync(messages, options);
@@ -627,13 +541,7 @@ public class AsIChatClientTest
                     "\"properties\":{\"result\":{\"type\":\"string\"}}",
                     Regex.Replace(request.GenerationConfig.ResponseJsonSchema.ToString(), @"\s+", ""));
 
-                GenerateContentResponse response = new();
-                response.Candidates.Add(new Candidate()
-                {
-                    Content = new Content() { Role = "model", Parts = { new Part() { Text = "{\"result\": \"success\"}" } } }
-                });
-
-                return response;
+                return CreateResponse(new() { Role = "model", Parts = { new Part() { Text = """{"result": "success"}""" } } });
             }
         };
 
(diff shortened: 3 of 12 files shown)


apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
```diff
@@ -0,0 +1,546 @@
...
+        GenerateContentRequest request = CreateRequest(messages, options);
+
+        // Send it.
+        GenerateContentResponse generateResult = await _client.GenerateContentAsync(request, cancellationToken).ConfigureAwait(false);
+
+        // Create the response.
+        ChatResponse chatResponse = new(new ChatMessage(ChatRole.Assistant, []))
+        {
+            CreatedAt = generateResult.CreateTime?.ToDateTimeOffset(),
+            ModelId = VertexAIExtensions.GetModelIdFromEndpoint(request.Model),
```
verdie-g: Would `generateResult.ModelVersion` work here?

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceImageGenerator.cs
```diff
@@ -0,0 +1,179 @@
...
+
+    /// <inheritdoc />
+    public async Task<ImageGenerationResponse> GenerateAsync(ImageGenerationRequest request, ImageGenerationOptions? options = null, CancellationToken cancellationToken = default)
+    {
+        VertexAIExtensions.ThrowIfNull(request);
+
+        // Create the PredictRequest object. If the options contains a RawRepresentationFactory, try to use it to
+        // create the request instance, allowing the caller to populate it with Vertex-specific options. Otherwise, create
+        // a new instance directly.
+        PredictRequest predicRequest = options?.RawRepresentationFactory?.Invoke(this) as PredictRequest ?? new();
```
ericstj: nit:
```suggestion
        PredictRequest predictRequest = options?.RawRepresentationFactory?.Invoke(this) as PredictRequest ?? new();
```

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
```diff
@@ -0,0 +1,546 @@
...
+
+                case FunctionResultContent functionResultContent:
+                    parts.Add(new Part()
+                    {
+                        FunctionResponse = new FunctionResponse
+                        {
+                            Name = functionResultContent.CallId,
+                            Response = functionResultContent.Result is not null ?
+                            Protobuf.WellKnownTypes.Struct.Parser.ParseJson(JsonSerializer.Serialize(new { result = functionResultContent.Result })) :
+                            new Protobuf.WellKnownTypes.Struct()
```
verdie-g: According to the [doc](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/Content#FunctionResponse), `Response` supports output and error special fields. Is there a way with MEAI and is it desirable to set the error when the function failed? MEAI does mention that FunctionResultContent.Exception "is for informational purposes only".

stephentoub: For security reasons I wouldn't want to automatically send any such exception back to the LLM, as it could contain sensitive information. The FunctionInvokingChatClient by default will put the failure information from the Exception into the Result if and only if the developer has explicitly opted into that by setting IncludeDetailedErrors to true:
https://github.com/dotnet/extensions/blob/5858a796217aab02fe4bcabe67447987c2cd6530/src/Libraries/Microsoft.Extensions.AI/ChatCompletion/FunctionInvokingChatClient.cs#L1071-L1083
We could here look for whether the Exception is non-null and if it is put the result into the "error", but it doesn't appear like that would have a meaningful effect.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
stephentoub: For chat, it looks like it. For embeddings, it's always coming back as empty. I updated the PR to prefer the model version in the response if it exists.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// in conjunction with <paramref name="projectId"/> and <paramref name="location"/> to form the full resource name.
+    /// It may be "publisher/model", in which case <paramref name="publisher"/> is ignored. Or it may be the full resource name of the model,
+    /// in which case <paramref name="publisher"/>, <paramref name="projectId"/>, and <paramref name="location"/> will all be ignored.
+    /// </param>
+    /// <returns>An <see cref="IChatClient"/> that wraps the specified client.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
+    public static IChatClient AsIChatClient(
+        this PredictionServiceClient client, string projectId, string location, string? publisher = null, string? defaultModelId = null)
+    {
+        ThrowIfNull(client);
```
amanda-tarafa: We have ` GaxPreconditions.CheckNotNull(client, nameof(client));` and other similar methods. These usually return whatever they checked so you can probably turn these to expression-bodied implementations.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// <param name="projectId">The ID of the Google Cloud project that accesses the resource, e.g. "my-vertex-project".</param>
+    /// <param name="location">The region where the resource is hosted, e.g. "us-central1".</param>
+    /// <param name="publisher">The model publisher. Defaults to "google" if not specified.</param>
+    /// <param name="defaultModelId">
+    /// The default model ID to use for chat requests if not specified in <see cref="ChatOptions.ModelId"/>.
+    /// If this is not specified, the model ID must be provided in each <see cref="ChatOptions"/> instance.
+    /// The model ID may be just the model name, in which case <paramref name="publisher"/> (defaulting to "google") is used
+    /// in conjunction with <paramref name="projectId"/> and <paramref name="location"/> to form the full resource name.
+    /// It may be "publisher/model", in which case <paramref name="publisher"/> is ignored. Or it may be the full resource name of the model,
+    /// in which case <paramref name="publisher"/>, <paramref name="projectId"/>, and <paramref name="location"/> will all be ignored.
```
amanda-tarafa: I think this adds unnecesssary complexity and it's also different to what we usually offer: either use the full resource name i.e. "projects/project-id/location/us-central1/publisher/...." or the parts. 

I think we should keep ChatOptions.ModelId as just the model ID and allow users to add the other parts as ChatOptions.AdditonalProperties.

(I'm happy to make those changes myself later though)

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// </param>
+    /// <returns>An <see cref="IChatClient"/> that wraps the specified client.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
+    public static IChatClient AsIChatClient(
```
amanda-tarafa, line 35: I think we would have used `ToIChatClient` but if the convention for the rest of implementations is to use `As...` then that's fine.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// <param name="location">The region where the resource is hosted, e.g. "us-central1".</param>
+    /// <param name="publisher">The model publisher. Defaults to "google" if not specified.</param>
+    /// <param name="defaultModelId">
+    /// The default model ID to use for chat requests if not specified in <see cref="ImageGenerationOptions.ModelId"/>.
+    /// If this is not specified, the model ID must be provided in each <see cref="ImageGenerationOptions"/> instance.
+    /// The model ID must be the full resource name of the model, e.g. "projects/{projectId}/locations/{location}/publishers/{publisher}/models/{model}"
+    /// </param>
+    /// <returns>An <see cref="IImageGenerator"/> that wraps the specified client.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="client"/> is <see langword="null"/>.</exception>
+    [Experimental("MEAI001")]
```
amanda-tarafa: No need to add this for now, we'll be releasing a beta first.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
```diff
@@ -0,0 +1,546 @@
...
+    {
+        VertexAIExtensions.ThrowIfNull(serviceType);
+
+        if (serviceKey is null)
+        {
+            // If there's a request for metadata, lazily-initialize it and return it. We don't need to worry about race conditions,
+            // as there's no requirement that the same instance be returned each time, and creation is idempotent.
+            if (serviceType == typeof(ChatClientMetadata))
+            {
+                return _metadata ??= new("gcp.vertex_ai", defaultModelId: _defaultModelId);
```
amanda-tarafa: For URI we could use `PredictionServiceClient.DefaultEndpoint`, but with a sufficiently configured client, the actual URI may be different, and we can't know it easily.

Is the expectation that this is the actual URI or the default one would be good enough.

Also, I need to check if the name `gcp.vertex_ai` is good enough or another one is prefered, but we won't block this PR on that.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
```diff
@@ -0,0 +1,546 @@
...
+                }
+            }
+        }
+
+        // Set the model, preferring what was in GenerateContentRequest if one was provided via RawRepresentationFactory,
+        // then what was in ChatOptions.Model, and finally any default set at the IChatClient level.
+        model ??= _defaultModelId;
+        if (string.IsNullOrWhiteSpace(request.Model) && !string.IsNullOrWhiteSpace(model))
+        {
+            request.Model = VertexAIExtensions.GetModelEndpoint(_projectId, _location, _publisher, model);
```
amanda-tarafa: Maybe this method will go, but we would call this `GetModelName`. The thing you are building is the full resource name of the model.

One thing to note is that we cannot allow the location to be modified, as that's included in the effective URI (e.g. https://us-central1-aiplatform.googleapis.com) and we cannot change that after client creation (see this [example](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/aiplatform/api/AIPlatform.Samples/MultiTurnChatSample.cs)) . I think I'm getting more and more convinced that we allow publisher and modelId to be changed through ChatOptions but not project and location. I'm also wondering if we should add the extension methods on the `PredictionServiceClientBuilder` instead of on the `PredictionServiceClient`, and then we'd named them `BuildAsIChatClient` etc. (I'm happy to make these changes myself though).

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.csproj
```diff
@@ -6,10 +6,14 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <Description>Vertex AI implementation of Microsoft.Extensions.AI.</Description>
     <PackageTags>Vertex AI;AI Extensions;Google;Cloud</PackageTags>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="ConfigureAwaitChecker.Analyzer" PrivateAssets="All" />
     <ProjectReference Include="..\..\Google.Cloud.AIPlatform.V1\Google.Cloud.AIPlatform.V1\Google.Cloud.AIPlatform.V1.csproj" />
     <PackageReference Include="Microsoft.Extensions.AI" />
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" />
```
ericstj, line 13: It would be nice if this also multi-targeted a newer .NET to leverage newer features / API and reduce dependencies.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
stephentoub: Ok, I switched to using it. (For the future, you might consider updating it to use `CallerArgumentExpression` so that all call sites can be just `(client)` rather than `(client, nameof(client))`.)

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
stephentoub: It's your project and we can make this fit whatever norms you think are best, but FWIW allowing this actually feels simpler to me rather than more complex.

It seems a full resource name is common, in which case I would expect to be able to use one as ChatOptions.ModelId, which means allowing ChatOptions.ModelId to be a full resource name or just the model ID. And at that point, since the publisher is also closely tied to the model, it's desirable to be able to specify publisher/model, which is also just the suffix of the full resource name.

Plus, it seems the full resource name can also be for fine-tuned models, in which case the full resource can be composed of slightly different parts, so being able to just accept the full resource rather than individual parts is simpler: the caller can compose it and then we just use it.

ChatOptions.AdditionalOptions is more complicated, both for the caller who needs to know what key names to use and what the types of their values need to be, and for us, fishing out the various pieces. Seems simpler to just say that the model can be effectively any of the things you'd expect:
- the full resource name with project-id+location+publisher+model
- the full resource name with project-id+location+endpoint
- the publisher+model
- or just the model

... and they all "just work".

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
stephentoub: In some cases, there's no cast method at all because the client library just implements the interfaces directly on the relevant type (e.g. `PredictionServiceClient : IChatClient, IEmbeddingGenerator, ...`). But for most cases where there is such a method, it's with `As`.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
stephentoub: Ok, sounds good. I included it here because `IImageGenerator` itself is `[Experimental]` with the same diagnostic ID, but I've removed it and just suppressed that diagnostic for the library.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
stephentoub: > For URI we could use PredictionServiceClient.DefaultEndpoint

Ah, great, I'll use that.

> Is the expectation that this is the actual URI or the default one would be good enough.

It's primarily used for telemetry purposes. The default should be good enough. If there's an easy-ish way to do better, cool, but it's not worth a ton of effort.

> Also, I need to check if the name gcp.vertex_ai is good enough or another one is prefered, but we won't block this PR on that.

This is what the OpenTelemetry genai standard convention recommends:
https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/
`gcp.vertex_ai | Vertex AI`
so it's what I went with. But if you'd prefer something else, we can certainly change it.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
stephentoub: > we would call this GetModelName

Renamed it.

> I'm also wondering if we should add the extension methods on the PredictionServiceClientBuilder instead of on the PredictionServiceClient, and then we'd named them BuildAsIChatClient

We could have both? I've added those for now, too. Given that it's possible to get a PredictionServiceClient without going through a builder, it seems like you'd want them on the client as well, and having them off the "client" is also what's done in other libraries.

> One thing to note is that we cannot allow the location to be modified, as that's included in the effective URI (e.g. https://us-central1-aiplatform.googleapis.com/) and we cannot change that after client creation (see this [example](https://github.com/GoogleCloudPlatform/dotnet-docs-samples/blob/main/aiplatform/api/AIPlatform.Samples/MultiTurnChatSample.cs)) .

I'm not following how a developer successfully uses PredictionServiceClient. It can optionally include a location, and if it does the location can't be overridden via the full resource name? But there's no way to extract from the PredictionServiceClient whether it has a location, and it's valid to supply a full resource name as the model name and that full resource name can include a location?

If the location can't be changed, then it wouldn't be valid accepting a location to the AsIChatClient method either, no?

I'm confused how this is supposed to work.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.csproj
stephentoub: @amanda-tarafa, have you considered adding a net8.0 or net9.0 target to these libraries?

amanda-tarafa: Yes, we'll do, soonish. But independently from this work.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
amanda-tarafa: (There's history of how we got here and we now have to live with it, it's not in the control of client libraries).

Unfortunately users have to create a client per location, as the location ends up as part of the endpoint. And then they need to make certain to use the right client for the co-located resources (this is not exclusive of the Vertex AI API, many other Google APIs are similar).

We cannot change a client location after creation, nor there's a reliable/easy way to figure out the location it was created with, if any. So resources (e.g. models) still need to be specified with the location, which has to match whatever the client was created with. For models, the same client (with the right credentials) can be used to access different models "in" different projects from different publishers, but they all need to be in the same location.

That's why I was prefering:

- Different ChatOptions for each of the segments on the resource name. And we can't support location to be changed.
- To simplify, don't support project to be changed either, as that would be a less common use case: project here is used for billing, and you'd need a credential that has permissions in all the projects you are planning to use. So, it's OK if we have a client per project+location pair.
- (You really end up with a costum option for the publisher only)
- To avoid users creating a client with location A and using it for an IChatClient with location B, have the extension methods on the client builder, in that way we can manipulate the client builder endpoint to make it match the location we've gotten, or fail if things don't match, etc.
- The clients created through no-client-builders are the "default" clients, those will use the default endpoint, for which the matching location is `global`, so, for "extension" methods that match the default experience, we don't need to receive a client nor a location, just project, publisher and model, we internally build the default client and use `global` as the location.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
```diff
@@ -0,0 +1,547 @@
...
+                {
+                    switch (tool)
+                    {
+                        case AIFunctionDeclaration af:
+                            functionDeclarations ??= [];
+                            functionDeclarations.Add(new FunctionDeclaration
+                            {
+                                Name = af.Name,
+                                Description = af.Description ?? "",
+                                ParametersJsonSchema = Protobuf.WellKnownTypes.Value.Parser.ParseJson(af.JsonSchema.ToString()),
```
verdie-g: [FunctionDeclaration](https://cloud.google.com/vertex-ai/docs/reference/rest/v1beta1/FunctionDeclaration) should have a `responseJsonSchema` too. Can we set it with `AIFunctionDeclaration.ReturnJsonSchema`?

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
```diff
@@ -0,0 +1,547 @@
...
+            {
+                content = new TextContent(part.Text);
+            }
+            else if (part.InlineData is not null)
+            {
+                content = new DataContent(part.InlineData.Data.ToByteArray(), part.InlineData.MimeType);
+            }
+            else if (part.FileData is not null)
+            {
+                if (Uri.TryCreate(part.FileData.FileUri, UriKind.Absolute, out Uri? uri))
```
jskeet: If this fails, we end up with no content - is that better than an error?

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
stephentoub: Thanks!

> (You really end up with a costum option for the publisher only)

Publisher is tied tightly to model. If I specify a non-google model, I'm then forced to specify the publisher. Why not allow that to be done as part of the ModelID? To me as a user, it's simpler and more intuitive to do:
```csharp
options.ModelId = "mistralai/mistral-large-2411";
```
than to do:
```csharp
options.ModelId = "mistral-large-2411";
options.AdditionalProperties ??= [];
options.AdditionalProperties["publisher"] = "mistralai"; // after searching docs for what key to use
```

> Different ChatOptions for each of the segments on the resource name. And we can't support location to be changed.

I'm still struggling with this. It's _already_ the case with PredictionServiceClient that when someone creates a request and sets the Model, they can (need to?) specify the full resource name, right? Why should ChatOptions.ModelId be more restrictive than that?

If the push back is only against being able to use "publisher/model", I can more easily understand. But if the push back is against ChatOptions.ModelId being allowed to be the full resource name, then I still don't understand but would like to.

> To avoid users creating a client with location A and using it for an IChatClient with location B

Isn't this already a possibility with PredictionServiceClient directly? You create it with location A and then specify a Model with location B?

> have the extension methods on the client builder, in that way we can manipulate the client builder endpoint to make it match the location we've gotten, or fail if things don't match, etc.

Since you offered, could you push a commit that has the shape / behavior for this that you're hoping for?

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
stephentoub: It's subjective, of course, but in general we've erred on the side of these mappings being forgiving. Someone working with an IChatClient doesn't necessarily know what exact implementation they're working with, who's backing it, etc., especially in scenarios where either it's a library that's been parameterized with an IChatClient or a larger app where an IChatClient is being injected via DI, different backends might be used under different conditions or in different parts of the app, etc. On top of that, these systems are already non-deterministic. As such, we've preferred to just throw away data that can't be translated or understood rather than fail outright.

That said, that philosophy is typically more about handling of inputs from the caller before being passed to the service/model. In this particular case, this is about interpreting the results that the service is sending back. And presumably the service should always be sending back valid file URIs, and if it didn't, it'd be a bug in vertex? If so, yeah, making this an exception would be reasonable. No different than any exceptions that might be triggered by the underlying PredictionClientService failing to parse the wire data.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
verdie-g: From https://github.com/dotnet/extensions/discussions/6902#discussioncomment-14637170

> It was deliberate. I'd initially added it, but it failed for really common schemas, like string or arrays. I could not find any documentation about what else might be invalid schema according to gemini's handling, and thus opted to leave it out rather than risk it blowing up randomly based on some corner case I'm not aware of. The response schema doesn't affect what the model produces, it could only serve to inform it with a bit more information about how to interpret function results it gets back, so it did not seem like the right trade-off right now.

Fair enough.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
amanda-tarafa: > But if the push back is against ChatOptions.ModelId being allowed to be the full resource name, then I still don't understand but would like to.

The pushback is for allowing ChatOptions.ModelId to be anything from the "true" modelId to the full resource name. In PredictionServiceClient (and all other generated clients for all other APIs) we don't accept bits and pieces. The combination of having defaults that may be overridable with a magic string is what I think departs from our usual user experience. ChatOptions.ModelId being (always) the full resource name is OK, we just wouldn't need the defaults then.

But, given that location can't be changed, and chaging the project would be a corner case which I'm confortable not allowing, I think it's OK to have those two on (IChat)client creation and not modifiable.

That leaves with publisher and model only, and I see your point on using the ChatOptions.AdditionalProperties. I'm still worried about overloading the meaning of ModelId, where in Google world that would be only "mistral-large-2411". Would you say is a common use case to use the same IChatClient with different models? We have to support it because there's ChatOptions.ModelId, but if that's a rare ocurrence, then it's easier to say "go use the additional properties for the publisher override".
Ultimately, I think I'm more confortable with project and location out of the picture, where ChatOptions.ModelId could be either "mistral-large-2411" or "mistralai/mistral-large-2411". So yep, OK.

> Isn't this already a possibility with PredictionServiceClient directly? You create it with location A and then specify a Model with location B?

Yes, but with PredictionServiceClient there's no bits and pieces to decompose the model name in, and there are no defaults. The model name has to be specified in full, always. For colocated models the request succeeds and fails otherwise. But here, we'd be taking a default location that maybe doesn't even work with the client.

The more I write this, the more I think the pushback is just with splitting the resource name. What if we just take the full resource name of a model as the default model in extensions and the ChatOptions.ModelId needs to be also the full resource model. In that way it's all equivalent to PredictionServiceClient? And also, future proof in terms of model names supporting different segments?

`public static IChatClient AsIChatClient(
        this PredictionServiceClient client, string defaultModelName = null)`

> Since you offered, could you push a commit that has the shape / behavior for this that you're hoping for?

Yep, I'm happy to merge this PR as is and then I can make those changes if we end up needing them.

stephentoub: > What if we just take the full resource name of a model as the default model in extensions and the ChatOptions.ModelId needs to be also the full resource model

Ok. I reverted it back to what I had initially. Thanks.

stephentoub:
@amanda-tarafa, I believe I've addressed all feedback. When you get a chance, please let me know if there are any further changes you'd like to see. Thanks!

verdie-g:
@stephentoub in the traces it gets a little awkward because MEAI see the model as `projects/{projectId}/locations/{location}/publishers/{publisher}/models/{model}`. Is that expected?
image

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/Google.Cloud.VertexAI.Extensions.IntegrationTests.csproj
```diff
@@ -4,7 +4,8 @@
     <TargetFrameworks>net8.0;net462</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net8.0</TargetFrameworks>
     <IsPackable>false</IsPackable>
-    <NoWarn>1701;1702;1705;xUnit2004;xUnit2013</NoWarn>
+    <NoWarn>1701;1702;1705;xUnit2004;xUnit2013;MEAI001</NoWarn>
```
jskeet: This will be overwritten by our project generation code. We can tweak that after the fact though.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/VertexAIExtensionsTest.cs
```diff
@@ -0,0 +1,110 @@
...
+
+        ImageGenerationResponse response = await generator.GenerateImagesAsync("A cute baby sea otter");
+
+        Assert.NotNull(response);
+        DataContent image = Assert.Single(response.Contents.OfType<DataContent>());
+        Assert.Equal("image/png", image.MediaType);
+        Assert.InRange(image.Data.Length, 1, int.MaxValue);
+    }
+
+    private static (string ProjectId, string Location) EnsureConfiguration()
```
jskeet: @amanda-tarafa: For after this PR: we can use TestEnvironment.GetTestProjectId  and TestEnvironment.GetProjectLocationId for this.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/VertexAIExtensionsTest.cs
```diff
@@ -0,0 +1,110 @@
...
+
+    [Fact]
+    public async Task AsIChatClient_BasicRequestResponse_Streaming()
+    {
+        (string projectId, string location) = EnsureConfiguration();
+
+        IChatClient client = (await CreateClientAsync()).AsIChatClient(EndpointName.FormatProjectLocationPublisherModel(projectId, location, "google", "gemini-2.5-pro"));
+        Assert.NotNull(client);
+
+        List<ChatResponseUpdate> updates = [];
```
jskeet: Can we use System.Linq.Async to simplify this?
(Non-blocking.)

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/Google.Cloud.VertexAI.Extensions.Tests.csproj
```diff
@@ -4,7 +4,8 @@
     <TargetFrameworks>net8.0;net462</TargetFrameworks>
     <TargetFrameworks Condition=" '$(OS)' != 'Windows_NT' ">net8.0</TargetFrameworks>
     <IsPackable>false</IsPackable>
-    <NoWarn>1701;1702;1705;xUnit2004;xUnit2013</NoWarn>
+    <NoWarn>1701;1702;1705;xUnit2004;xUnit2013;MEAI001</NoWarn>
+    <Nullable>enable</Nullable>
```
jskeet, line 8: @amanda-tarafa More tweaks to add for project generation later.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
```diff
@@ -0,0 +1,1527 @@
...
+            throw new NotSupportedException();
+
+        public override AsyncDuplexStreamingCall<TRequest, TResponse> AsyncDuplexStreamingCall<TRequest, TResponse>(Method<TRequest, TResponse> method, string? host, CallOptions options) =>
+            throw new NotSupportedException();
+
+        public override TResponse BlockingUnaryCall<TRequest, TResponse>(Method<TRequest, TResponse> method, string? host, CallOptions options, TRequest request) =>
+            throw new NotSupportedException();
+
+        /// <summary>A test implementation of IAsyncStreamReader for mocking streaming responses.</summary>
+        private sealed class TestAsyncResponseStream<T>(T[] responses) : IAsyncStreamReader<T>
```
jskeet, line 1395: @amanda-tarafa We can use `AsyncStreamAdapter<T>` from the GAX testing package as an alternative to this.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
```diff
@@ -0,0 +1,1527 @@
...
+    }
+
+    [Fact]
+    public async Task IChatClient_GetResponseAsync_BasicRequest()
+    {
+        DelegateCallInvoker invoker = new()
+        {
+            OnGenerateContentRequest = request =>
+            {
+                GenerateContentResponse response = new();
```
jskeet: @amanda-tarafa I think this can be simplified to:

```csharp
return new()
{
    Candidates = {
      new Candidate() { Content = new Content() { Role = "model", Parts = { new Part() { Text = "Hello" } } } }
    }
}
```

(Or similar.)

jskeet: Looking at the other tests, creating a method to return a Candidate (or maybe just a Content) with the specified text would simplify things a bit.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
```diff
@@ -0,0 +1,1527 @@
...
+            }
+        };
+
+        IChatClient chatClient = CreateClient(invoker).AsIChatClient("projects/test-project/locations/us-central1/publishers/google/models/mymodel");
+        ChatMessage[] messages = [new(ChatRole.User, "What's the weather in San Francisco?")];
+
+        ChatResponse result = await chatClient.GetResponseAsync(messages);
+
+        Assert.NotNull(result);
+        Assert.Single(result.Messages[0].Contents);
```
jskeet: @amanda-tarafa I *think* Assert.Single returns the value and `Assert.IsType` returns it cast, so this can be:

```csharp
var content = Assert.Single(result.Messages[0].Contents);
var functionCall = Assert.IsType<FunctionCallContent>(content);
```

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
```diff
@@ -0,0 +1,1527 @@
...
+    [Fact]
+    public async Task IChatClient_GetResponseAsync_WithChatOptions()
+    {
+        DelegateCallInvoker invoker = new()
+        {
+            OnGenerateContentRequest = request =>
+            {
+                // Verify ChatOptions were applied correctly
+                Assert.Equal("projects/test/locations/us-central1/publishers/google/models/gemini-pro", request.Model);
+                Assert.NotNull(request.GenerationConfig);
```
jskeet: Can we construct the exact GenerationConfig we expect, and then just use Assert.Equal on that? (Protobuf messages implement equality.)

jskeet: (This might be an approach for some other tests... I don't know how much latitude we expect in terms of the requests.)

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceEmbeddingGenerator.cs
```diff
@@ -0,0 +1,162 @@
...
+
+        return result;
+    }
+
+    /// <inheritdoc />
+    public object? GetService(System.Type serviceType, object? serviceKey = null)
+    {
+        GaxPreconditions.CheckNotNull(serviceType, nameof(serviceType));
+
+        if (serviceKey is null)
```
jskeet, line 137: If we really want to always return null if the `serviceKey` is non-null, I think I'd reverse the logic to put an early-out here, for clarity. But is that definitely the desired behavior? This is outside my domain for sure, but it seems a little odd.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceEmbeddingGenerator.cs
```diff
@@ -0,0 +1,162 @@
...
+        string? model = options?.ModelId ?? _defaultModelId;
+        if (string.IsNullOrWhiteSpace(request.Endpoint) && !string.IsNullOrWhiteSpace(model))
+        {
+            request.Endpoint = model;
+        }
+
+        // Add all of the inputs.
+        foreach (string value in values)
+        {
+            request.Instances.Add(new Protobuf.WellKnownTypes.Value
```
jskeet: Would "using Value = Protobuf.WellKnownTypes.Value" mean we don't need any of these full qualifications?

jskeet: (If it's done deliberately to avoid any ambiguity, that's a different matter of course.)

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
```diff
@@ -0,0 +1,529 @@
...
+
+            // Yield the update.
+            yield return responseUpdate;
+        }
+    }
+
+    /// <inheritdoc />
+    public object? GetService(System.Type serviceType, object? serviceKey = null)
+    {
+        GaxPreconditions.CheckNotNull(serviceType, nameof(serviceType));
```
jskeet, line 118: Same comment as on PredictionServiceEmbeddingGenerator.GetService.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
```diff
@@ -0,0 +1,529 @@
...
+            request.Contents.Add(new Content() { Role = "user", Parts = { new Part() { Text = "" } } });
+        }
+
+        return request;
+    }
+
+    /// <summary>Creates <see cref="Part"/>s for <paramref name="contents"/> and adds them to <paramref name="parts"/>.</summary>
+    private static void AddPartsForAIContents(IEnumerable<AIContent> contents, RepeatedField<Part> parts)
+    {
+        foreach (AIContent content in contents)
```
jskeet, line 337: My initial temptation is to refactor this into a loop and a local function with a switch expression returning a Part. That's slightly tricky for TextReasoningContent where we want to ignore a base64 parsing failure. I might have a go after this is merged...

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
```diff
@@ -0,0 +1,529 @@
...
+    }
+
+    /// <summary>Creates <see cref="AIContent"/>s for <paramref name="parts"/> and adds them to <paramref name="contents"/>.</summary>
+    private static void AddAIContentsForParts(RepeatedField<Part> parts, IList<AIContent> contents)
+    {
+        foreach (var part in parts)
+        {
+            AIContent? content = null;
+
+            if (!string.IsNullOrEmpty(part.Text))
```
jskeet: We can switch on `part.DataOneofCase` potentially.

stephentoub:
> This is a far from complete review - and @stephentoub I'm not suggesting you make any changes

Thanks, @jskeet. I'll go through them this morning and address low hanging fruit. There's a new M.E.AI release coming out today, anyway, so I may as well bump to it. I can also address @verdie-g's feedback.

stephentoub:
> @stephentoub in the traces it gets a little awkward because MEAI see the model as `projects/{projectId}/locations/{location}/publishers/{publisher}/models/{model}`. Is that expected?

I could change it so that the ChatClientMetadata exposes just the model ID rather than the full resource name; that would help if you supply the defaultModelId to the As/Build methods. If you instead provide the full resource name in ChatOptions.ModelId, it'll show up that way in telemetry.

verdie-g:
It may have been discussed earlier with Amanda but can't we save the project/location/provider somewhere so I can just pass `gemini-2.5-pro` to ChatOptions.ModelId?

stephentoub:
> It may have been discussed earlier with Amanda but can't we save the project/location/provider somewhere so I can just pass `gemini-2.5-pro` to ChatOptions.ModelId?

There were extensive discussions, if you read back through the comments.

verdie-g:
Okay this is going to be a little awkward for our use-case as we use a single project for the entire company and the project/location should be hidden from our developers but I suppose we could have a middleware that appends the project/location to `ChatOptions.ModelId`.

stephentoub:
> Okay this is going to be a little awkward for our use-case as we use a single project for the entire company and the project/location should be hidden from our developers but I suppose we could have a middleware that appends the project/location to `ChatOptions.ModelId`.

@amanda-tarafa

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/Google.Cloud.VertexAI.Extensions.IntegrationTests.csproj
stephentoub: I've moved this to a Directory.Build.props.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/VertexAIExtensionsTest.cs
stephentoub: Done

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/VertexAIExtensionsTest.cs
stephentoub: We could (though moving forward preferably System.Linq.AsyncEnumerable instead). But there's a built-in ToChatResponseAsync method in M.E.AI that I've switched to using here.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/Google.Cloud.VertexAI.Extensions.Tests.csproj
stephentoub: I've moved this to a Directory.Build.props. However, the project generation will still need to be tweaked, as this `<NoWarn>` is overriding all other NoWarns set higher up. It should be augmented to start with `$(NoWarn);` so that it incorporates other NoWarns set earlier.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
stephentoub: I wrote the src implementations by hand, as I care a lot about ensuring that library code others will be using is as pristine as possible. But I let copilot / sonnet 4.5 write almost all of the tests :) I've now added a helper.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
stephentoub: Combined

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
stephentoub: Done (for two occurrences... there was a third I left as-is because it wasn't testing exact equality)

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceEmbeddingGenerator.cs
stephentoub: >  I think I'd reverse the logic to put an early-out here, for clarity

I have it like this to consolidate the `return null`, which should apply in any situation where we don't have something to return. It could of course be separated if you'd prefer.

> But is that definitely the desired behavior?

This maps to how DI is generally handled, e.g. if you ask for a named HttpClient and it doesn't exist, you don't get back the default, you get back null (or an exception if you asked for it as required). We could of course do whatever behavior we want here, but this is how it's handled in every other IChatClient implementation I've seen.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceEmbeddingGenerator.cs
stephentoub: Yup

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
stephentoub: Same answer :)

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
stephentoub: I'd looked at that, but it didn't have anything for Thought, so I just went with the cascade. But I've changed it to just accommodate Thought in the default case.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceEmbeddingGenerator.cs
jskeet: So would a named client be handled by something else earlier in the DI pipeline, effectively? That's the bit I'm missing - how the calling code gets a named client if it *does* exist.

stephentoub: > So would a named client be handled by something else earlier in the DI pipeline, effectively? That's the bit I'm missing - how the calling code gets a named client if it _does_ exist.

Yes, that would typically be done by registering a keyed instance in DI. The IChatClient you get back would then be based on the key you supplied to DI.

That's distinct from this GetService method, which is named the same to invoke parallels between them, but isn't use for the same purpose. IChatClient.GetService is more of a glorified cast. Let's say someone has a pipeline of IChatClient instances, e.g.
```csharp
IChatClient client = predictionServiceClient.AsIChatClient(...)
    .AsBuilder()
    .UseFunctionInvocation()
    .UseOpenTelemetry()
    .Build();
```
The IChatClient they get back isn't castable to a PredictionServiceChatClient (and even if it was that type is internal), so to extract information from it, they can't just cast. But they can call GetService, asking the instance to give it back the desired thing. In most cases, just the type is sufficient, so they'd do `GetService<PredictionServiceClient>()`, and each stage of the pipeline would either satisfy the request or pass it along to the next stage, until it gets to the leaf PredictionServiceChatClient, which can then return the desired instance or null if it can't. The key parameter is then a further way to disambiguate which instance is desired. And if someone provides a key, we assume they're looking for something more specific than just type, so if we have state that doesn't have a key, it doesn't match, following the same principles as in keyed DI.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
stephentoub: I couldn't find this. Presumably there's some additional testing package that needs to be referenced?

stephentoub:
> Approving - all my comments were really for "potential post-work" anyway :)

Thanks, understood, I'm just generally of the "why put off for tomorrow" mindset :)

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.Tests/AsIChatClientTest.cs
jskeet: Yup, it's Google.Api.Gax.Grpc.Testing IIRC - but please don't worry about it, we can make that change after merging.

amanda-tarafa:
> > Okay this is going to be a little awkward for our use-case as we use a single project for the entire company and the project/location should be hidden from our developers but I suppose we could have a middleware that appends the project/location to `ChatOptions.ModelId`.
> 
> @amanda-tarafa

@verdie-g You would have that problem with any Google .NET client, for instance if you were using the PredictionServiceClient diretly. In general we only take the full resource name for all resources involved in a RPC, and our service clients do not have default values for any of the resource name segments. Let's leave this is as for the moment as a baseline, it's easier to add more ways to specify the resource name later on than to remove something that we add now and then causes unexpected friction.

Also, @jskeet what do you think, not sure if you followed the previous discussion regarding full resource name vs. storing default segments, etc. Let's chat about it if you haven't.

jskeet:
> > > Okay this is going to be a little awkward for our use-case as we use a single project for the entire company and the project/location should be hidden from our developers but I suppose we could have a middleware that appends the project/location to `ChatOptions.ModelId`.
> > 
> > 
> > @amanda-tarafa
> 
> @verdie-g You would have that problem with any Google .NET client, for instance if you were using the PredictionServiceClient diretly. In general we only take the full resource name for all resources involved in a RPC, and our service clients do not have default values for any of the resource name segments. Let's leave this is as for the moment as a baseline, it's easier to add more ways to specify the resource name later on than to remove something that we add now and then causes unexpected friction.
> 
> Also, @jskeet what do you think, not sure if you followed the previous discussion regarding full resource name vs. storing default segments, etc. Let's chat about it if you haven't.

Discussed offline, and I'm basically keen on keeping it as simple as possible for now.

PR #15148, opened 2025-10-10
tools: reinstate the update-dependencies command

jskeet:
(This was deleted in May 2025 in a purge of obsolete tools. In this case, the tool isn't actually obsolete.)

This is now *only* a standalone command (not invoked automatically from anything else).


jskeet:
This has found a few more updates beyond #15146, but we'll address those later...

PR #15158, opened 2025-10-14
fix: Tweak a few things in Google.Cloud.VertexAI.Extensions

stephentoub:
- Changed dependency to be on M.E.AI.Abstractions rather than on M.E.AI. The latter contains additional middleware that's not used or needed by this implementation.
- Updated M.E.AI.Abstractions dependency from 9.9.1 to 9.10.0
- Fixed handling of null IServiceProvider in BuildXx methods (I didn't realize it wasn't optional in Build(IServiceProvider)).

Contributes to https://github.com/googleapis/google-cloud-dotnet/issues/13815

cc: @amanda-tarafa, @jskeet

### DIFF
--- a/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
+++ b/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
@@ -56,7 +56,11 @@ public static class VertexAIExtensions
     {
         GaxPreconditions.CheckNotNull(builder, nameof(builder));
 
-        return builder.Build(provider).AsIChatClient(defaultModelId);
+        PredictionServiceClient client = provider is not null ?
+            builder.Build(provider) :
+            builder.Build();
+
+        return client.AsIChatClient(defaultModelId);
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


global.json
```diff
@@ -1,7 +1,7 @@
 {
   "sdk": {
-    "version": "8.0.413",
-    "allowPrerelease": false,
+    "version": "10.0.100-rc.2.25429.103",
```
amanda-tarafa: Not yet :). Once it's stable, and then we first need to update our internal CI.

amanda-tarafa (changes requested):
Thanks!

global.json
stephentoub: Oops, that was a mistake; I didn't intend to include my own local global.json.

apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/VertexAIExtensions.cs
```diff
@@ -56,7 +56,11 @@ public static IChatClient BuildIChatClient(
     {
         GaxPreconditions.CheckNotNull(builder, nameof(builder));
 
-        return builder.Build(provider).AsIChatClient(defaultModelId);
+        PredictionServiceClient client = provider is not null ?
+            builder.Build(provider) :
```
stephentoub, line 60: FWIW, the vast majority of APIs that accept an IServiceProvider allow it to be null; you might consider updating `Build(IServiceProvider)` to behave like `Build()` when IServiceProvider is null rather than throwing.

PR #15172, opened 2025-10-18
ci: Build and test everything if Directory.Packages.props changes.

amanda-tarafa:
Also, make trivial changes in Directory.Packages.props that will test this change.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -36,8 +36,12 @@ while (( "$#" )); do
     runtests=false
   elif [[ "$1" == "--diff" ]]
   then
-    apis+=($(git diff main --name-only | grep -e 'apis/.*/' | cut -d/ -f 2 | uniq))
-    diff=true
+    if git diff main --name-only | grep -q "Directory.Packages.prop" | grep -q "Directory.Build.props" | grep -q "Directory.Build.targets" | grep -q "global.json"; then
+      secho "Top level dependencies changed. Building and testing everything. Note that --regex parameter is still respected, but it will be applied to the full API list and not just to those that changed."
+    else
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -36,8 +36,12 @@ while (( "$#" )); do
     runtests=false
   elif [[ "$1" == "--diff" ]]
   then
-    apis+=($(git diff main --name-only | grep -e 'apis/.*/' | cut -d/ -f 2 | uniq))
-    diff=true
+    if git diff main --name-only | grep -q "Directory.Packages.prop"; then
```
jskeet: Possibly add a comment (or modify the message) to make it clear that the regex can still filter this?

(My understanding is that build-pr.yml will still effectively split this into three build jobs, each of which will only build a subset, yes?)

jskeet (dismissed):

build.sh
amanda-tarafa: Modifying the message...

But yes, regex filtering happens after API list filtering. And regex is not really filtering out any APIs on the list, it's just splitting the set in three separate runs.

jskeet: Right - it's only that with the current message, we'll get three messages of "Building and testing everything" which makes it sound like it's building everything three times :)

amanda-tarafa: I've added the other "top level dependencies" file and changed the message. And yes, to clarify my las comment that was confusing, on build.sh the API list will still be filtered by regex. It is build-pr action that takes care to pass regexes that will end up building everything that's on the original API list. See if the comment is good enough now.

jskeet (approved):

PR #15213, opened 2025-11-04
chore: revert release batch 2 and restrict release size

jskeet:


jskeet:
We may have an alternative approach - marked this as "do not merge" for now...

PR #15216, opened 2025-11-05
refactor(storage): generalize move object feature to all types of buckets

mahendra-google:
This PR contains code changes to make move object feature generalized to all types of buckets earlier it was limited to hns bucket only. Please see[ b/429345064](https://b.corp.google.com/issues/429345064)


amanda-tarafa (commented):
Thanks, this looks good.

We won't merge just yet because we are in the middle of batch releasing everything, but we'll do so as soon as there are no clashes. (I'll rename the commit to feat though).

amanda-tarafa (approved):

PR #15230, opened 2025-11-06
tools: streamline running integration tests from ReleaseManager

jskeet:
This command is used to run integration tests on Windows. I believe we spend most of the time checking out the repo and then finding there are no tests to run.

Additionally, we now log with timestamps to make any further diagnosis easier.


jskeet:
I believe this will significantly speed up step 3 of the current release process. (It won't change anything else.)

Important caveats with this:

- Libraries which only have smoke tests were being tested, and will now be skipped; they'll still run in other steps.
- If integration tests or snippets are deleted between the release PR being created and the job running, the integration tests won't run (in step 3) - this seems like a negligible risk to me.

PR #15258, opened 2025-11-17
feat(Storage): Add ReturnPartialSuccess support to ListBuckets 

mahendra-google:
This PR introduces the `returnPartialSuccess` flag (via `ListBucketsOptions`). When set to true, the method will return buckets from reachable locations and populate the resource names of the buckets from `unreachable`  locations, rather than failing the entire request. Please see [b/459649174](https://b.corp.google.com/issues/401332989)

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/ListBucketsOptionsTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.Tests/ListBucketsOptionsTest.cs
@@ -31,6 +31,7 @@ namespace Google.Cloud.Storage.V1.Tests
             Assert.Null(request.MaxResults);
             Assert.Null(request.PageToken);
             Assert.Null(request.SoftDeleted);
+            Assert.Null(request.ReturnPartialSuccess);
         }
 
         [Fact]
@@ -45,6 +46,7 @@ namespace Google.Cloud.Storage.V1.Tests
                 PageToken = "nextpage",
                 Fields = "items(name),nextPageToken",
                 SoftDeletedOnly = true,
+                ReturnPartialSuccess = true
             };
             options.ModifyRequest(request);
             Assert.Equal(10, request.MaxResults);
@@ -53,6 +55,7 @@ namespace Google.Cloud.Storage.V1.Tests
             Assert.Equal("nextpage", request.PageToken);
             Assert.Equal("items(name),nextPageToken", request.Fields);
             Assert.True(request.SoftDeleted);
+            Assert.True(request.ReturnPartialSuccess);
         }
     }
 }
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.csproj
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.csproj
@@ -10,7 +10,7 @@
   <ItemGroup>
     <PackageReference Include="ConfigureAwaitChecker.Analyzer" PrivateAssets="All" />
     <PackageReference Include="Google.Api.Gax.Rest" />
-    <PackageReference Include="Google.Apis.Storage.v1" VersionOverride="[1.69.0.3707, 2.0.0.0)" />
+    <PackageReference Include="Google.Apis.Storage.v1" VersionOverride="[1.71.0.3920, 2.0.0.0)" />
   </ItemGroup>
   <ItemGroup>
     <Compile Update="StorageClient.*.cs">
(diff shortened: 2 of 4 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListBucketsTest.cs
```diff
@@ -75,6 +75,38 @@ public void PartialResponses()
             }
         }
 
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void ListBuckets_ReturnPartialSuccess(bool returnPartialSuccess)
```
amanda-tarafa: I don't see the point of having this integration test. We really cannot test anything, and even if we could it would be server behaviour that we usally don't test client side.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListBucketsTest.cs
```diff
@@ -75,6 +75,38 @@ public void PartialResponses()
...
+            // Get all reachable buckets from all pages.
+            var reachableBuckets = pages.SelectMany(page => page.Items ?? Enumerable.Empty<Bucket>());
+
+            // Check reachable buckets list contains the bucket we just created.
+            Assert.Contains(reachableBuckets, bucket => bucket.Name == bucketName);
+
+            // Get all unreachable buckets from all pages into a single list.
+            var unreachableBuckets = pages.SelectMany(page => page.Unreachable ?? Enumerable.Empty<string>());
+
+            if (!returnPartialSuccess)
```
amanda-tarafa: Avoid conditionals in tests, it's better to write separate tests.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListBucketsTest.cs
```diff
@@ -75,6 +75,38 @@ public void PartialResponses()
...
+            var options = new ListBucketsOptions { ReturnPartialSuccess = returnPartialSuccess };
+            var pagedResult = _fixture.Client.ListBuckets(_fixture.ProjectId, options);
+
+            var pages = pagedResult.AsRawResponses().ToList();
+
+            // Get all reachable buckets from all pages.
+            var reachableBuckets = pages.SelectMany(page => page.Items ?? Enumerable.Empty<Bucket>());
+
+            // Check reachable buckets list contains the bucket we just created.
+            Assert.Contains(reachableBuckets, bucket => bucket.Name == bucketName);
```
amanda-tarafa: How/what guarantees that this bucket is reachable, this test may be flaky.

amanda-tarafa (changes requested):

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListBucketsTest.cs
mahendra-google: Agreed. I should have refactored this logic into separate test methods.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListBucketsTest.cs
mahendra-google: I understand the concern regarding potential flakiness and bucket is being unreachable but I think that the multi-region buckets are designed which replicates metadata across regions to ensure availability even during regional downtime. WDYT?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListBucketsTest.cs
mahendra-google: Agreed. Fundamentally, this is testing server-side behavior rather than validating the client logic. Also we cannot deterministically trigger a partial regional failure in the real GCS environment during our integration test run and can't effectively test the `unreachable` vs `reachable` logic.  I am also inclined to remove this test case and and rely on the unit test for the `returnPartialSuccess` flag in ListBucketOptions. WDYT?

amanda-tarafa: I think I would prefer that we remove the test, for all the reasons discussed here and in the other related comments. Thanks!

mahendra-google: Integration test (**ListBuckets_ReturnPartialSuccess**) is removed for all the reasons discussed here and in the other related comments.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/ListBucketsTest.cs
mahendra-google: Integration test (**ListBuckets_ReturnPartialSuccess**) is removed for all the reasons discussed here and in the other related comments.

amanda-tarafa (approved):

Issue #15267, opened 2025-11-19
Integration test failure on Spanner emulator 1.5.44


jskeet:
Currently, the Spanner integration tests run in `.github/workflows/spanner-emulator-pr-push.yml` run against the "latest" emulator .

In emulator version 1.5.44, test InvalidDmlBatch_PartialSuccess_ExplicitTransaction fails with:

> Cannot read or query using a transaction after it has been committed or rolledback

I have not investigated whether the test is at fault, or the library, or the emulator... but it looks like the test succeeds with emulator version 1.5.43.

Proposed course of action:

- Pin the emulator version to 1.5.43 for now, and keep it pinned - but regularly update it. That gives us less risk of being surprised by tests starting to fail without any changes
- Investigate the failure and either report it to the Spanner team if we believe it's an emulator bug, or fix the test/library if it's a bug in this repo.

purva9413:
Created an issue on Spanner Emulator b/462134611. Updates will be on that bug, closing this out.

PR #15268, opened 2025-11-19
tests: pin the emulator version we use for Spanner testing

jskeet:
Towards #15267


jskeet:
Note: first try at this PR deliberately pinned to 1.5.44, as I wanted to see the tests fail. I then adjusted to 1.5.43, which is the last "known good for our tests" version.

PR #15271, opened 2025-11-21
feat(Log4Net)!: Support credential type configuration

robertvoinescu-work:
Added `CredentialType` to GoogleStackdriverAppender configuration to support credential specification.

BREAKING CHANGE: Credentials now default to service_account type instead of auto-detecting. If using
non-service-account credentials (e.g., authorized_user), you MUST explicitly set the CredentialType
property in your configuration.


apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.csproj
```diff
@@ -14,5 +14,10 @@
...
     <PackageReference Include="log4net" />
+    <!--
+      This should typically be pulled in as a transitive dependency of GAX, but for now
+      GAX doesn't reference the latest version.
+    -->
+    <PackageReference Include="Google.Apis.Auth" VersionOverride="1.73.0" />
   </ItemGroup>
-</Project>
\ No newline at end of file
+</Project>
```
robertvoinescu-work: This was missing a final new line, adding it now since it's required by the .editorconfig:

Image

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.csproj
```diff
@@ -14,5 +14,10 @@
     <PackageReference Include="Google.Cloud.Logging.V2" VersionOverride="[4.5.0, 5.0.0)" />
     <PackageReference Include="Grpc.Core" PrivateAssets="None" Condition="'$(TargetFramework)'=='net462'" />
     <PackageReference Include="log4net" />
+    <!--
```
robertvoinescu-work: TODO: Remove before merge, we will rely on the gax transitive dependency.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
```diff
@@ -36,6 +39,19 @@ namespace Google.Cloud.Logging.Log4Net.Tests
 {
     public class Log4NetTest
     {
+        private const string DummyServiceAccountCredentialJson = @"{
```
amanda-tarafa:
```suggestion
        private const string FakeServiceAccountCredentialJson = @"{
```

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
```diff
@@ -347,7 +363,7 @@ public async Task SingleLogEntryWithLocation()
                 Assert.True(string.IsNullOrEmpty(entry0.SourceLocation.File) || entry0.SourceLocation.File.EndsWith("Log4NetTest.cs"),
                 $"Actual 'entry0.SourceLocation.File' = '{entry0.SourceLocation.File}'");
                 // The exact line number seems to vary depending on framework (Core or desktop) and CI system.
-                Assert.True(entry0.SourceLocation.Line == 0L || entry0.SourceLocation.Line == 46L || entry0.SourceLocation.Line == 44L,
+                Assert.True(entry0.SourceLocation.Line == 0L || entry0.SourceLocation.Line == 46L || entry0.SourceLocation.Line == 44L || entry0.SourceLocation.Line == 62L,
```
amanda-tarafa: Oh my, I think we should remove this. I mean, all of it.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.csproj
amanda-tarafa: Equally marking as request changes while we release GAX.

amanda-tarafa (changes requested):
I have mostly the same questions here regarding the defaults. I know what you've done in code, but I want to make sure that what's done in configuration files is fine as well. Let's chat about it.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
robertvoinescu-work: Yeah I also don't like it. Removed!

generator-input/apis.json
```diff
@@ -3504,7 +3504,8 @@
         "Stackdriver"
       ],
       "dependencies": {
-        "Google.Api.Gax.Grpc": "default",
+        "Google.Api.Gax": "4.13.0",
```
amanda-tarafa, line 3567: Do we need this one explicitly?

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
```diff
@@ -233,10 +233,8 @@ private LoggingServiceV2Client BuildLoggingServiceClient() =>
         private GoogleCredential GetCredentialFromConfiguration()
         {
             var credential =
-#pragma warning disable CS0618 // Temporarily disable warnings for obsolete methods. See b/453009677 for more details.
-                !string.IsNullOrWhiteSpace(CredentialFile) ? GoogleCredential.FromFile(CredentialFile) :
-                !string.IsNullOrWhiteSpace(CredentialJson) ? GoogleCredential.FromJson(CredentialJson) :
-#pragma warning restore CS0618
+                !string.IsNullOrWhiteSpace(CredentialFile) ? CredentialFactory.FromFile(CredentialFile, CredentialType) :
```
amanda-tarafa, line 236: Curious here what happens if CredentialType is a bad value, including empty string? What's the error message we get from Auth and is it clear enough in thie context?

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.IntegrationTests/CredentialTests.cs
```diff
@@ -1,77 +0,0 @@
...
-using Google.Cloud.ClientTesting;
-using System;
-using System.IO;
-using Xunit;
-
-namespace Google.Cloud.Logging.Log4Net.IntegrationTests
-{
-    [SnippetOutputCollector]
-    [Collection(nameof(Log4NetFixture))]
-    public class CredentialTests
```
amanda-tarafa, line 24: Just to double check, we are able to remove these tests because we have equivalent unit tests, correct? These didn't need to be in the integration tests project to begin with? Please add a note on the commit description (not the commit name) explaining this.

You need to remove the entry for the IntegrationTests project in apis.json as well.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
amanda-tarafa: Although I see that none of the other fields are validated.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
```diff
@@ -327,32 +341,6 @@ public async Task SingleLogEntry()
             Assert.Equal("Message0", uploadedEntries[0].TextPayload.Trim());
         }
 
-        [Fact]
-        public async Task SingleLogEntryWithLocation()
```
amanda-tarafa, line 331: Can't remember why we removed this test?

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
```diff
@@ -694,5 +682,36 @@ public async Task GaePlatformMismatchedProjectId()
...
+            var tempFile = Path.GetTempFileName();
+            File.WriteAllText(tempFile, FakeServiceAccountCredentialJson);
+            try
+            {
+                var appender = new GoogleStackdriverAppender
+                {
+                    CredentialFile = tempFile,
+                    CredentialType = "invalid_type"
+                };
+                Assert.Throws<InvalidOperationException>(() => appender.ActivateOptions());
```
amanda-tarafa: Can we add similar tests but that succeeded here so we now that if things are right the credential is loaded?

amanda-tarafa: Or maybe a better question is, what's the fundamental difference between these tests here and the ones in CredentialTests?

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/CredentialTests.cs
```diff
@@ -0,0 +1,183 @@
...
+using log4net.Config;
+using System;
+using System.IO;
+using System.Linq;
+using System.Xml;
+using Xunit;
+
+namespace Google.Cloud.Logging.Log4Net.Tests
+{
+    public class CredentialTests
```
amanda-tarafa: These tests are great thanks!

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/CredentialTests.cs
```diff
@@ -0,0 +1,183 @@
...
+
+using log4net;
+using log4net.Config;
+using System;
+using System.IO;
+using System.Linq;
+using System.Xml;
+using Xunit;
+
+namespace Google.Cloud.Logging.Log4Net.Tests
```
amanda-tarafa: For new files, we try to use file scoped namespaces.

amanda-tarafa (changes requested):
A few things, but mostly questions. Can you update the documentation as well? Here: https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.Logging.Log4Net/docs/configuration.md

Also, apply the relevant changes from this one to the NLog PR, once we get this one merged and all relevant changes done in NLog, I'll review that one. Thanks.

generator-input/apis.json
robertvoinescu-work: Yes, there's some transitive dependencies pulling in a lower version that breaks the build. So we specify 4.13.0 as the lowest allowed version.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
robertvoinescu-work: Good point. The error handling would happen in auth.
1. Null CredentialType -> ArgumentNullException with parameter "credentialType" specified in the message. 
2. Empty CredentialType -> InvalidOperationException  with message "Error creating credential from JSON or Json Parameters. Unrecognized credential type ."

I don't like this and think we will want to treat the empty case as we do the null and throw earlier in the same check. Is it okay we address this as a follow-up?

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.IntegrationTests/CredentialTests.cs
robertvoinescu-work: Yes these are mostly translated and extended in the new unit tests (I added quite a few more). I've updated the commit description and thanks for the heads on on the other api.json entry! Fixed.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
robertvoinescu-work: It's very fragile due to checking for exact line numbers. But looking at it again I can just remove the part that just checks the line number and we can keep the test otherwise.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/Log4NetTest.cs
robertvoinescu-work: Looking at it again these cases are already covered by the credential tests, there we also test the success cases. I will just remove these tests and keep everything in the credential tests.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/CredentialTests.cs
robertvoinescu-work: Good catch, I moved this file from the integration tests as a starting point and forgot to update how we namespace. Fixed.

generator-input/apis.json
amanda-tarafa: This is adding the extra line again.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
amanda-tarafa: Definetely OK to do in a separate PR. But also, I'm fine with an empty string failing as it does now in Auth, I don't think we should change that there.

If that may present a problem here because the unset value defaults to empty string, then we maybe add the extra validation here, or convert the empty value to null before calling Auth.

Alternative: because we've said that CredentialType defaults to service account credential, would it make sense to check for null or empty here and send SA to Auth in those cases?

apis/Google.Cloud.Logging.Log4Net/docs/configuration.md
amanda-tarafa: Add a note here and below saying that credentialType defaults to SA. Something like: "If `credentialType` is not specified, then it will default to ...". Also mentione here where to find the values, linking to https://docs.cloud.google.com/dotnet/docs/reference/Google.Apis/latest/Google.Apis.Auth.OAuth2.JsonCredentialParameters

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/CredentialTests.cs
amanda-tarafa: This test would be better if the file path and the JSON content were of a valid credential. This test will continue to pass if we remove the check for the double credential, because the credentials are not good.

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net/GoogleStackdriverAppender.cs
robertvoinescu-work: Got it, I'll leave it as is then. There should be no problems when the type is unset. I've got validation for this in `CredentialTest.LoadCredentialFromJson_Succeeds` and `CredentialTest.LoadCredentialFromFile_Succeeds` which doesn't specify a credentialType.

generator-input/apis.json
robertvoinescu-work: The .editorconfig specifies adding a new line at the end of each file:
image

My editor must have picked it up and added it in. I might prefer to keep the new line so it's in line with the config, but I can also see a preferences for not introducing these formatting updates into older files. What would you prefer?

apis/Google.Cloud.Logging.Log4Net/Google.Cloud.Logging.Log4Net.Tests/CredentialTests.cs
robertvoinescu-work: Really good point, I've updated the test to use valid files and json content.

apis/Google.Cloud.Logging.Log4Net/docs/configuration.md
robertvoinescu-work: Done.

generator-input/apis.json
robertvoinescu-work: I've removed the new line for now.

amanda-tarafa: The only thing is apis.json is generated. I don't mind changing what generates it to add the extra blank line, but until we do that, it will be removed even if we add it.

PR #15280, opened 2025-11-28
feat: Spanner non-awaiting DDL

olavloite:
Adds a method to execute DDL on Spanner without having the client awaiting the long-running operation to finish. This makes it possible to for example start the creation of a secondary index using the client library, without having the library waiting for that operation to finish. Creating a secondary index on a table with a large amount of data can take a very long time (several days).

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AdminTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AdminTests.cs
@@ -12,7 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Api.Gax.Grpc;
+using Google.Cloud.Spanner.Admin.Database.V1;
 using Google.Cloud.Spanner.Data.CommonTesting;
+using Google.LongRunning;
+using Google.Protobuf;
+using Google.Protobuf.WellKnownTypes;
 using System;
 using System.Threading.Tasks;
 using Xunit;
@@ -239,19 +244,76 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
             using (var connection = new SpannerConnection(builder.WithDatabase(dbName)))
             {
                 var dropSingersCmd = connection.CreateDdlCommand("DROP TABLE Singers");
-                var dropBothTablesCmd = connection.CreateDdlCommand("DROP TABLE Albums", "DROP TABLE Singers");
+                var dropAlbumsCmd = connection.CreateDdlCommand("DROP TABLE Albums");
 
                 await Assert.ThrowsAsync<SpannerException>(() => dropSingersCmd.ExecuteNonQueryAsync());
-                var operationName = await dropBothTablesCmd.StartDdlAsync();
-                Assert.NotEqual("", operationName);
+                await dropAlbumsCmd.ExecuteNonQueryAsync();
+                await dropSingersCmd.ExecuteNonQueryAsync();
             }
 
             using (var connection = new SpannerConnection(builder))
             {
                 var dropCommand = connection.CreateDdlCommand($"DROP DATABASE {dbName}");
-                var operationName = await dropCommand.StartDdlAsync();
-                // DropDatabase does not return a long-running operation.
-                Assert.Equal("", operationName);
+                await dropCommand.ExecuteNonQueryAsync();
+            }
+        }
+
+        [Fact]
+        public async Task StartDdlReturnsOperationName()
+        {
+            string dbName = GenerateDatabaseName();
+            var builder = new SpannerConnectionStringBuilder(_fixture.Database.NoDbConnectionString);
+            var connectionOptions = new SpannerClientCreationOptions(builder);
+            var adminClientBuilder = connectionOptions.CreateDatabaseAdminClientBuilder();
+            var adminClient = await adminClientBuilder.BuildAsync();
+            var channel = adminClientBuilder.LastCreatedChannel;
+
+            try
+            {
+                using (var connection = new SpannerConnection(builder))
+                {
+                    var createDbCommand = connection.CreateDdlCommand($"CREATE DATABASE {dbName}");
+                    var operationName = await createDbCommand.StartDdlAsync();
+                    Assert.False(string.IsNullOrEmpty(operationName));
+
+                    await HandleLro<Database, CreateDatabaseMetadata>(
+                        adminClient.CreateDatabaseOperationsClient, operationName);
+                }
+
+                using (var connection = new SpannerConnection(builder.WithDatabase(dbName)))
+                {
+                    var createTableCommand = connection.CreateDdlCommand(
+                        "CREATE TABLE Singers (SingerId INT64 PRIMARY KEY, Name STRING(1024))");
+                    var operationName = await createTableCommand.StartDdlAsync();
+                    Assert.False(string.IsNullOrEmpty(operationName));
+
+                    await HandleLro<Empty, UpdateDatabaseDdlMetadata>(
+                        adminClient.UpdateDatabaseDdlOperationsClient, operationName);
+                }
+
+                using (var connection = new SpannerConnection(builder))
+                {
+                    var dropCommand = connection.CreateDdlCommand($"DROP DATABASE {dbName}");
+                    var operationName = await dropCommand.StartDdlAsync();
+                    // DropDatabase does not return a long-running operation.
+                    Assert.Null(operationName);
+                }
+            }
+            finally
+            {
+                channel?.Shutdown();
+            }
+
+            async Task HandleLro<TResponse, TMetadata>(OperationsClient client, string operationName)
+                where TResponse : class, IMessage<TResponse>, new()
+                where TMetadata : class, IMessage<TMetadata>, new()
+            {
+                var rawOperation = await client.GetOperationAsync(operationName);
+                var operation = new Operation<TResponse, TMetadata>(rawOperation, client);
+                var completedOperation = await operation.PollUntilCompletedAsync();
+
+                Assert.True(completedOperation.IsCompleted);
+                Assert.Null(completedOperation.Exception);
             }
         }
     }
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
@@ -244,11 +244,14 @@ namespace Google.Cloud.Spanner.Data
             /// <summary>
             /// Starts a DDL operation, but does not wait for the long-running operation to finish.
             /// </summary>
-            /// <returns>The name of the long-running operation that was created</returns>
+            /// <returns>
+            /// The name of the long-running operation that was created or null if the DDL statement did not
+            /// create a long-running operation.
+            /// </returns>
             internal async Task<string> StartDdlAsync(CancellationToken cancellationToken)
             {
                 var operation = await ExecuteDdlAsync(pollUntilCompleted: false, cancellationToken).ConfigureAwait(false);
-                return operation?.Name ?? "";
+                return operation?.Name;
             }
 
             private async Task<Operation> ExecuteDdlAsync(bool pollUntilCompleted, CancellationToken cancellationToken)
@@ -275,15 +278,8 @@ namespace Google.Cloud.Spanner.Data
                             ExtraStatements = { CommandTextBuilder.ExtraStatements ?? new string[0] },
                             ProtoDescriptors = CommandTextBuilder.ProtobufDescriptors?.ToByteString() ?? ByteString.Empty,
                         };
-                        var response = await databaseAdminClient.CreateDatabaseAsync(request).ConfigureAwait(false);
-                        if (pollUntilCompleted)
-                        {
-                            response = await response.PollUntilCompletedAsync().ConfigureAwait(false);
-                        }
-                        if (response.IsFaulted)
-                        {
-                            throw SpannerException.FromOperationFailedException(response.Exception);
-                        }
+                        var createDbOperation = await databaseAdminClient.CreateDatabaseAsync(request).ConfigureAwait(false);
+                        var response = await HandleLro(createDbOperation).ConfigureAwait(false);
                         operation = response.RpcMessage;
                     }
                     else if (CommandTextBuilder.IsDropDatabaseCommand)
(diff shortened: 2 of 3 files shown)


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -181,7 +182,7 @@ internal Task<int> ExecuteNonQueryAsync(CancellationToken cancellationToken)
                 switch (CommandTextBuilder.SpannerCommandType)
                 {
                     case SpannerCommandType.Ddl:
-                        return ExecuteDdlAsync(cancellationToken);
+                        return ExecuteDdlAsync(pollUntilCompleted: true, cancellationToken).ContinueWith(_ => 1, TaskContinuationOptions.ExecuteSynchronously);
```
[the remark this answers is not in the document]
olavloite: Reverted to returning 0 to remain compatible with the previous behavior.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -513,6 +514,39 @@ public long ExecutePartitionedUpdate() =>
...
+            Task.Run(() => ExecuteDdlAsync(pollUntilCompleted, _synchronousCancellationTokenSource.Token)).ResultWithUnwrappedExceptions();
+
+        /// <summary>
+        /// Executes this command as DDL. The command must contain one or more DDL statements;
+        /// <see cref="SpannerConnection.CreateDdlCommand(string, string[])"/> for details.
+        /// </summary>
+        /// <param name="pollUntilCompleted">
+        /// If true, the returned task will wait for the DDL operation to finish execution on Spanner.
+        /// Otherwise, the returned task will finish as soon as the DDL operation has successfully been started on
+        /// Spanner, and the caller can monitor the progress of the long-running operation.
```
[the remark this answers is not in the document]
olavloite: I disagree with that. The intention here is that the caller is free to choose whether they want/need to poll the operation until it is complete, or whether it is OK for it to be a shoot-and-forget operation. In the latter case, they can also monitor the progress using for example the Google Cloud Console.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -513,6 +514,39 @@ public long ExecutePartitionedUpdate() =>
...
             CreateExecutableCommand().ExecutePartitionedUpdateAsync(cancellationToken);
 
+        /// <summary>
+        /// Executes this command as DDL. The command must contain one or more DDL statements;
+        /// <see cref="SpannerConnection.CreateDdlCommand(string, string[])"/> for details.
+        /// </summary>
+        /// <param name="pollUntilCompleted">
+        /// If true, the returned task will wait for the DDL operation to finish execution on Spanner.
+        /// Otherwise, the returned task will finish as soon as the DDL operation has successfully been started on
+        /// Spanner, and the caller can monitor the progress of the long-running operation.
```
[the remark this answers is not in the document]
olavloite: I disagree with that. The intention here is that the caller is free to choose whether they want/need to poll the operation until it is complete, or whether it is OK for it to be a shoot-and-forget operation. In the latter case, they can also monitor the progress using for example the Google Cloud Console.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -258,7 +266,11 @@ private async Task<int> ExecuteDdlAsync(CancellationToken cancellationToken)
                             ProtoDescriptors = CommandTextBuilder.ProtobufDescriptors?.ToByteString() ?? ByteString.Empty,
                         };
                         var response = await databaseAdminClient.CreateDatabaseAsync(request).ConfigureAwait(false);
-                        response = await response.PollUntilCompletedAsync().ConfigureAwait(false);
+                        operation = response.RpcMessage;
+                        if (pollUntilCompleted)
+                        {
+                            response = await response.PollUntilCompletedAsync().ConfigureAwait(false);
+                        }
                         if (response.IsFaulted)
```
amanda-tarafa: This check should happen before you get the operation.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -258,7 +266,11 @@ private async Task<int> ExecuteDdlAsync(CancellationToken cancellationToken)
                             ProtoDescriptors = CommandTextBuilder.ProtobufDescriptors?.ToByteString() ?? ByteString.Empty,
                         };
                         var response = await databaseAdminClient.CreateDatabaseAsync(request).ConfigureAwait(false);
-                        response = await response.PollUntilCompletedAsync().ConfigureAwait(false);
+                        operation = response.RpcMessage;
+                        if (pollUntilCompleted)
```
amanda-tarafa: Instead of adding this if in two places, let's have this method return the un-polled operation and the wrapping ones poll until completion if they need to.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -513,6 +514,39 @@ public long ExecutePartitionedUpdate() =>
...
+        /// </param>
+        /// <returns>
+        /// A reference to the long-running operation that was started for the DDL statement(s).
+        /// Note: The operation is null for DropDatabase commands.
+        /// </returns>
+        public Operation ExecuteDdl(bool pollUntilCompleted = false) =>
```
amanda-tarafa: Let's keep all the methods in the library async, consumers can always block themselves.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -513,6 +514,39 @@ public long ExecutePartitionedUpdate() =>
...
+        /// If true, the returned task will wait for the DDL operation to finish execution on Spanner.
+        /// Otherwise, the returned task will finish as soon as the DDL operation has successfully been started on
+        /// Spanner, and the caller can monitor the progress of the long-running operation.
+        /// </param>
+        /// <param name="cancellationToken">An optional token for canceling the call.</param>
+        /// <returns>
+        /// A reference to the long-running operation that was started for the DDL statement(s).
+        /// Note: The operation is null for DropDatabase commands.
+        /// </returns>
+        public Task<Operation> ExecuteDdlAsync(bool pollUntilCompleted = false, CancellationToken cancellationToken = default) =>
```
amanda-tarafa: We usually don't expose the underlying types in the ADO.NET layer. Is it really needed to return an operation? And in that case what part of the Operation is needed?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
olavloite: Is that really possible in this case, considering the fact that the two methods return two different operation types:
1. `Operation<Database, CreateDatabaseMetadata>`
2. `Operation<Empty, UpdateDatabaseDdlMetadata>`

(Type non-generic `Operation` class cannot be used for polling)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
olavloite: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
olavloite: Replaced with `StartDdlAsync`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
olavloite: Replaced with `Task<string> StartDdlAsync`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
robertvoinescu-work: If the goal is to remove duplication might I instead suggest a a generic local function to centralize the polling and error-handling logic (at the bottom of this method):
```csharp
async Task<Operation> HandleLro<TResp, TMeta>(Operation<TResp, TMeta> operation)
{
    if (pollUntilCompleted)
    {
        operation = await operation.PollUntilCompletedAsync().ConfigureAwait(false);
    }
    if (operation.IsFaulted)
    {
        throw SpannerException.FromOperationFailedException(operation.Exception);
    }
    return operation;
}
```

If we go the other way and use the returned `Operation` to poll later, the calling method will add extra overhead because it needs to establish a new channel since this method shuts down the one it creates in the `finally` block.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -513,6 +514,21 @@ public long ExecutePartitionedUpdate() =>
...
+        /// used to cancel the request to start the execution of the DDL statements. It cannot be used to cancel the
+        /// long-running operation once it has been started.
+        /// The command must contain one or more DDL statements;
+        /// <see cref="SpannerConnection.CreateDdlCommand(string, string[])"/> for details.
+        /// </summary>
+        /// <returns>
+        /// The name of the long-running operation that was started for the DDL statement(s).
+        /// Note: The ID is empty for DropDatabase commands.
+        /// </returns>
+        public Task<string> StartDdlAsync(CancellationToken cancellationToken = default) =>
```
robertvoinescu-work, line 529: NIT: Can we add a param xml element, something like.
```csharp
/// <param name="cancellationToken">An optional token for canceling the call.</param>
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AdminTests.cs
robertvoinescu-work, line 42: Can we add a test to ensure an operation name is returned also, maybe something that extends this like so:
```csharp
[Fact]
public async Task StartDdlAsync_ReturnsOperationAndCompletes()
{
    string dbName = GenerateDatabaseName();
    var operationsClient = DatabaseAdminClient.Create().CreateDatabaseOperationsClient;
    using var connection = new SpannerConnection(_fixture.Database.NoDbConnectionString);

    try
    {
        var createCmd = connection.CreateDdlCommand($"CREATE DATABASE {dbName}");
        var operationName = await createCmd.StartDdlAsync();

        Assert.False(string.IsNullOrEmpty(operationName));

        // Wait for the database creation LRO to finish
        var rawOperation = await operationsClient.GetOperationAsync(operationName);
        var operation = new Operation<Database, CreateDatabaseMetadata>(rawOperation, operationsClient);
        var completedOperation = await operation.PollUntilCompletedAsync();

        Assert.True(completedOperation.IsCompleted);
        Assert.Null(completedOperation.Exception);
    }
    finally
    {
        // Always cleanup the created database
        var dropCommand = connection.CreateDdlCommand($"DROP DATABASE {dbName}");
        await dropCommand.ExecuteNonQueryAsync();
    }

    // No sessions created, so no session pool.
}

```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -235,10 +236,27 @@ private async Task<SpannerDataReader> ExecuteDmlReaderAsync(CommandBehavior beha
...
+            }
+
+            /// <summary>
+            /// Starts a DDL operation, but does not wait for the long-running operation to finish.
+            /// </summary>
+            /// <returns>The name of the long-running operation that was created</returns>
+            internal async Task<string> StartDdlAsync(CancellationToken cancellationToken)
+            {
+                var operation = await ExecuteDdlAsync(pollUntilCompleted: false, cancellationToken).ConfigureAwait(false);
+                return operation?.Name ?? "";
```
amanda-tarafa: Maybe return null instead of ""?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
amanda-tarafa: Yes +1 to Robert's comment, I hadn't noticed the channel being closed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AdminTests.cs
```diff
@@ -239,17 +239,19 @@ AlbumTitle STRING(MAX),
...
             {
                 var dropSingersCmd = connection.CreateDdlCommand("DROP TABLE Singers");
-                var dropAlbumsCmd = connection.CreateDdlCommand("DROP TABLE Albums");
+                var dropBothTablesCmd = connection.CreateDdlCommand("DROP TABLE Albums", "DROP TABLE Singers");
 
                 await Assert.ThrowsAsync<SpannerException>(() => dropSingersCmd.ExecuteNonQueryAsync());
-                await dropAlbumsCmd.ExecuteNonQueryAsync();
-                await dropSingersCmd.ExecuteNonQueryAsync();
+                var operationName = await dropBothTablesCmd.StartDdlAsync();
+                Assert.NotEqual("", operationName);
```
amanda-tarafa: Would you mind leaving the existing tests as is and creating a new one for this. The one we have is not great (re: name and what its testing) ,but if we also check that the operation name is returned, then it gets even more confusing. I'm fine if you duplicate this one and tweak it, or just use the create tables and drop database parts of it.

amanda-tarafa (changes requested):

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AdminTests.cs
olavloite: Done (We need to get the admin client from the connection builder in order to correctly connect to the emulator)

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
olavloite: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/AdminTests.cs
olavloite: Done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
olavloite: Done

Issue #15284, opened 2025-12-01
Figure out Microsoft.Bcl.AsyncInterfaces issues


jskeet:
Currently, GAX depends on Microsoft.Bcl.AsyncInterfaces v6.0.0.

We would like Google.Cloud.Firestore to depend on System.Linq.AsyncEnumerable, which in turn depends on Microsoft.Bcl.AsyncInterfaces v10.0.0. It's also worth noting that Google.Cloud.Firestore targets netstandard2.0 (as well as net462, but that's irrelevant).

This causes issues, not in Google.Cloud.Firestore itself, but in apps that depend on it. I've managed to isolate this from Google code:

## IndirectLibrary

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
  </ItemGroup>

</Project>
```

(No relevant C# code.)

## Library

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
    <LangVersion>latest</LangVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="System.Linq.AsyncEnumerable" Version="10.0.0" />
    <ProjectReference Include="..\IndirectLibrary\IndirectLibrary.csproj" />
  </ItemGroup>

</Project>

```

This time there does need to be C# code referencing `IAsyncEnumerable<T>`:

```csharp
using System.Collections.Generic;

namespace Library;

public class Class1
{
    public static void M<T>(IAsyncEnumerable<T> source)
    {
    }
}
```

## App

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="..\Library\Library.csproj" />
  </ItemGroup>

</Project>
```

C# code is irrelevant.

We get:

> warning MSB3277: Found conflicts between different versions of "Microsoft.Bcl.AsyncInterfaces" that could not be resolved.

## Workaround

We could:

[...]

PR #15285, opened 2025-12-01
feat: release new major versions of Firestore and Datastore with System.Linq.AsyncEnumerable

### LINKED ISSUE #15037: Use System.Linq.AsyncEnumerable instead of System.Linq.Async
Starting in .NET 10, there are conflicts due to a dependency on `System.Linq.Async`. This affects all downstream products.

While a workaround is available, it would be nice if this can get resolved for a .NET 10 release (maybe keep previous platform support with ifdefs)

Some more information is available in https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/10.0/asyncenumerable

jskeet:
Fixes #15037.

This takes the first option in #15284 and should only be merged if we're happy with that.

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/docs/index.md
+++ b/apis/Google.Cloud.Datastore.V1/docs/index.md
@@ -23,22 +23,6 @@ Several custom conversions, additional constructors,
 factory methods (particularly on [Filter](obj/api/Google.Cloud.Datastore.V1.Filter.yml)
 are provided to simplify working with the protobuf messages.
 
-## Conflicts with .NET 10
-
-.NET 10 introduces the `AsyncEnumerable` class which causes
-conflicts with the `System.Linq.Async` package that
-Google.Cloud.Datastore.V1 depends on.
-
-We will release a new major version with a dependency on
-`System.Linq.AsyncEnumerable` instead, but within the current major
-version (3.x) you can follow the workaround described on the [.NET compatibility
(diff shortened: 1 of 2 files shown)


generator-input/pipeline-state.json
```diff
@@ -1639,6 +1640,7 @@
         {
             "id": "Google.Cloud.Firestore.Admin.V1",
             "currentVersion": "3.16.0",
+            "nextVersion": "4.0.0",
```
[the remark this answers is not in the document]
jskeet: Will fix

jskeet:
(Note: we won't actually do a release until Dec 3rd.)

amanda-tarafa (approved):

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/message.txt
```diff
@@ -0,0 +1 @@
+Bump message
```
amanda-tarafa: Out of curiosity, shouldn't the changes in project file, etc. trigger the release? Or is it because the other commits are `chore`?

jskeet: I did it this way to make sure it was just the right libraries, with the right message. The change when *generating* the project files would have done it otherwise though, yes - I think. This just felt a bit easier to predict.

PR #15293, opened 2025-12-03
tests: fix snippets for GkeMultiCloud to use a non-deprecated service

jskeet:


jskeet:
This fixes b/465628184

PR #15308, opened 2025-12-08
feat(docs): add devsite-help and snippets for Retries, Update Masks, and OCC

bshaffer:
I followed the examples of other pages in `docs/devsite-help`, but it's quite possible I'm missing a critical aspect. I also am not sure how these will be tested. 

I ran all samples locally to verify they worked but please take a look and review 
1. Docs content
2. Sample code
3. How I've connected things up


tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,75 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Cloud.SecretManager.V1;
+using Google.Protobuf.WellKnownTypes;
+using Grpc.Core;
+
+namespace Google.Cloud.Tools.Snippets
```
amanda-tarafa: File scoped namespace (no curly brackets, see [tools/Google.Cloud.Docs.Snippets/LongRunningOperationsSnippets.cs](https://github.com/googleapis/google-cloud-dotnet/blob/main/tools/Google.Cloud.Docs.Snippets/LongRunningOperationsSnippets.cs)) for new files, here and in the others.

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,75 @@
...
+{
+    public class UpdateMaskSnippets
+    {
+        public async Task UpdateMasks()
```
amanda-tarafa: This won't run as a test. See https://github.com/googleapis/google-cloud-dotnet/blob/main/tools/Google.Cloud.Docs.Snippets/CallSettingsSnippets.cs for an example.

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            string projectId = "your-project-id";
+            string secretId = "test-secret";
```
amanda-tarafa: So that this is run, these need to come from the fixture.

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            // Sample: UpdateMasks
+            // Required using directives:
+            // using Google.Cloud.SecretManager.V1;
+            // using Google.Protobuf.WellKnownTypes;
+            // using Grpc.Core;
```
amanda-tarafa: We don't have these in the other product neutral samples. Let's keep the consistency.

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            // Setup Client
```
amanda-tarafa: We avoid this type of comment that does not add specific information. As well as the other "Prepare ..." etc.

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,75 @@
...
+            SecretManagerServiceClient client = await SecretManagerServiceClient.CreateAsync();
+
+            // Prepare the resource with NEW values
+            Secret secret = new Secret
+            {
+                // Set the full resource name
+                SecretName = SecretName.FromProjectSecret(projectId, secretId),
+
```
amanda-tarafa, line 29:
```suggestion
```

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,75 @@
...
+                // Populate ONLY the fields we intend to change
```
amanda-tarafa: Is this true? I thought that the best thing about field masks is that even if you have several values populated, only the ones in the field mask are updated.

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,75 @@
...
+                Labels = { { "env", "production" } }
+            };
+
+            // Create the FieldMask
+            FieldMask updateMask = new FieldMask();
+
+            // Add the Protobuf field name in snake_case
+            // Note: The C# property is 'Labels', but the proto field name is 'labels'.
+            updateMask.Paths.Add("labels");
```
amanda-tarafa:
```suggestion
            // Create the FieldMask
            FieldMask updateMask = new FieldMask
            {
                // Add the original protobuf field name.
                // In this example, the C# property is 'Labels', but the protobuf field name is 'labels'.
                Paths = { "labels" }
            };
```

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,75 @@
...
+
+            // Prepare the Request object
+            UpdateSecretRequest request = new UpdateSecretRequest
+            {
+                Secret = secret,
+                UpdateMask = updateMask
+            };
+
+            // Call the API
+            try
```
amanda-tarafa: We avoid showing how to handle generic errors. We only show error handling if it's relevant/specific to the sample.

tools/Google.Cloud.Docs.Snippets/RetriesSnippets.cs
```diff
@@ -0,0 +1,92 @@
...
+// limitations under the License.
+
+using Google.Cloud.SecretManager.V1;
+using Google.Api.Gax.Grpc;
+using Google.Api.Gax;
+using Grpc.Core;
+
+namespace Google.Cloud.Tools.Snippets
+{
+    public class RetriesSnippets
```
amanda-tarafa: We already have [retry documentation](https://docs.cloud.google.com/dotnet/docs/reference/help/call-settings#retry) and an [example for retry](https://github.com/googleapis/google-cloud-dotnet/blob/bb5d1eed2464f19ab904efb0a2823ce5d73c0b78/tools/Google.Cloud.Docs.Snippets/CallSettingsSnippets.cs#L135). The example is not currently showing in docs, but that's just because we need to add the reference to the sample [here](https://github.com/googleapis/google-cloud-dotnet/blob/main/docs/devsite-help/call-settings.md)

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
```diff
@@ -0,0 +1,105 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Api.Gax.ResourceNames;
+using Google.Cloud.Iam.V1;
+using Google.Cloud.ResourceManager.V3;
+using Grpc.Core;
+
+namespace Google.Cloud.Tools.Snippets
```
amanda-tarafa: All the same comments from the field mask snippet apply here.

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
```diff
@@ -0,0 +1,105 @@
...
+{
+    public class OccForIamSnippets
+    {
+        public async Task<int> OccForIam()
+        {
+            string projectId = "your-project-id";
+            string role = "roles/cloudkms.cryptoKeyEncrypterDecrypter";
+            string member = "user:betterbrent@google.com";
```
amanda-tarafa: All of these should come from the fixture, and definetely not harcoded.

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
```diff
@@ -0,0 +1,105 @@
...
+            // using Grpc.Core;
+
+            // Setup Client
+            ProjectsClient client = await ProjectsClient.CreateAsync();
+            ProjectName resourceName = ProjectName.FromProject(projectId);
+
+            int retries = 0;
+
+            // --- START OCC LOOP ---
+            while (retries < maxRetries)
```
amanda-tarafa: I don't think we need to show an actual retry loop. I'd prefer if we only show what errors we get if optimistic concurrency fails. Users should decide whether they want to retry or not, and how to do that.

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
```diff
@@ -0,0 +1,105 @@
...
+                    Console.WriteLine($"Attempt {retries + 1}: Writing modified policy...");
+                    Policy updatedPolicy = await client.SetIamPolicyAsync(resourceName, policy);
+
+                    // SUCCESS
+                    Console.WriteLine("Successfully updated IAM policy.");
+                    return 0;
+                }
+                catch (RpcException ex) when (
+                    ex.StatusCode == StatusCode.Aborted ||
+                    ex.StatusCode == StatusCode.FailedPrecondition)
```
amanda-tarafa: Are two distinct codes a signal of failed optimistic concurrency?

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
```diff
@@ -0,0 +1,105 @@
...
+                        throw;
+                    }
+
+                    // Simple backoff
+                    await Task.Delay(100 * retries);
+                }
+            }
+            // --- END OCC LOOP ---
+
+            return 1; // non-zero, failed
```
amanda-tarafa: This is not idiomatic. An error in .NET is an exception. But I don't think we need to include all the plumbing and the method within the sample. Look at the others in here, the sample portion is the minimum possible that shows the feature, but we are not trying to be prescriptive in whether that goes in a method etc.

docs/devsite-help/update-masks.md
```diff
@@ -0,0 +1,19 @@
...
+
+When updating resources (often corresponding to HTTP PATCH requests) in Google Cloud APIs, you generally use an **Update Mask** (represented by `Google.Protobuf.WellKnownTypes.FieldMask`).
+
+The purpose of the Update Mask is to tell the server exactly which fields you intend to modify, preventing accidental overwrites or resetting of other fields that you did not include in your request object.
+
+By providing a `FieldMask`, you explicitly declare the subset of fields in the resource object that should be updated.
+
+## Constructing and Applying the FieldMask
+
+The `FieldMask` object is part of the Google Protobuf library and contains a collection of strings representing the fields to be modified.
```
amanda-tarafa:
```suggestion
The `FieldMask` object is part of the `Google.Protobuf` library and contains a collection of strings representing the fields paths to be modified.
```

docs/devsite-help/update-masks.md
```diff
@@ -0,0 +1,19 @@
...
+
+**Crucial Point:** The strings in the `FieldMask.Paths` collection **must** correspond to the **snake\_case** field names defined in the Protocol Buffer (protobuf) schema, not the C\# property names (which are PascalCase).
```
amanda-tarafa:
```suggestion
**Crucial Point:** The strings in the `FieldMask.Paths` collection **must** correspond to the **original** field names defined in the Protocol Buffer (protobuf) schema, not the C\# property names (which are PascalCase).
```

docs/devsite-help/toc.yml
```diff
@@ -28,6 +28,10 @@
     href: long-running-operations.md
   - name: Pagination (page streaming)
     href: page-streaming.md
+  - name: Update masks
+    href: update-masks.md
+  - name: Retries
+    href: retries.md
```
amanda-tarafa: We have these in CallSettings already.

docs/devsite-help/retries.md
```diff
@@ -0,0 +1,32 @@
+# Configuring Retries and Timeouts
```
amanda-tarafa: Look at the one in call-settings.md and maybe modify it if there's something missing from there, plus, add the sample code that we already have?

docs/devsite-help/occ-for-iam.md
```diff
@@ -0,0 +1,26 @@
+# Optimistic Concurrency Control (OCC) Loop for IAM
```
amanda-tarafa:
```suggestion
# Optimistic Concurrency Control (OCC) for IAM
```

amanda-tarafa: And also, why does this need to be IAM specific? I think we can use IAM for examples but the same principles apply to many APIs that have Etags for their resources? (I should have left this commenton the initial content review, but it only occured to me now).

docs/devsite-help/occ-for-iam.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Introduction to OCC
+
+Optimistic Concurrency Control (OCC) is a strategy used to manage shared resources and prevent "lost updates" or race conditions when multiple users or processes attempt to modify the same resource simultaneously.
+
+In the context of Google Cloud .NET libraries, IAM Policy objects contain an `Etag` property. When calling `SetIamPolicy`, the client library includes this `Etag`. If the server detects that the `Etag` provided does not match the current version on the server, it throws an RPC exception with the status `Aborted` or `FailedPrecondition`.
```
amanda-tarafa: Is there any info about the difference between Aborted or FailedPrecondition?

docs/devsite-help/occ-for-iam.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Implementing the OCC Loop
```
amanda-tarafa: Same comment as before, I don't think we need to show or describe an actual retry loop. For instance, for someone building a UI to allow users to modify an OCCed resource, they might not want to retry, they might want to notify the user and let the user retry (by clicking the button again) etc.

amanda-tarafa (changes requested):
Let me know how I can help.

docs/devsite-help/occ-for-iam.md
bshaffer: Yes I agree! I think IAM is just the most common use-case, so it's the one we decided to use.

I can remove IAM from the title, and reword the sample to make it clearer that IAM is only one use of OCC, if you think that would be better

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
bshaffer: I've removed them from all the product neutral guides, but just FYI this was taken from here: 

https://github.com/googleapis/google-cloud-dotnet/blob/2837d44971610aa1995988f3ef103f568c0d25c3/tools/Google.Cloud.Docs.Snippets/FaqSnippets.cs#L25-L30

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
bshaffer: That makes sense to me! What do you think the sample should show? I can just place this in a comment

```
// Handle the etag mismatch. For example, retry the change or prompt the user to
// submit their changes again
```

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
bshaffer: There is no secret in the fixture, should I add a `_fixture.CreateSecret()`?

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
bshaffer: That's my understanding too. I'll remove "ONLY" here since that's confusing.

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
bshaffer: SGTM, I'll remove the int returns

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
bshaffer: This is what Gemini tells me, although we should probably verify:

   - ABORTED: This is the most common status code for a concurrency failure, such as an ETag mismatch during a read-modify-write cycle. It indicates the resource was modified by another writer after you read it.

   - FAILED_PRECONDITION: While less common for ETag mismatches, this status code can also be returned in certain race conditions or if the resource state is otherwise incompatible with the request.

See [Understanding API Errors](https://developers.google.com/google-ads/api/docs/best-practices/understand-api-errors)

docs/devsite-help/occ-for-iam.md
bshaffer: See https://github.com/googleapis/google-cloud-dotnet/pull/15308#discussion_r2676947621

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
bshaffer: So they should come from environment variables instead? Could you provide some guidance as to how I should define them?

docs/devsite-help/occ-for-iam.md
bshaffer: loop removed!

docs/devsite-help/retries.md
bshaffer: Added the retry sample to the call settings docs. That should be sufficient.

docs/devsite-help/occ-for-iam.md
```diff
@@ -0,0 +1,22 @@
+# Optimistic Concurrency Control (OCC) for IAM
```
amanda-tarafa: Can't find the other comment but yes, I think it'd be best to refer to IAM as an example only and remove it from anywhere that's general (the titles and file names etc.) and could imply that OCC only applies to IAM resources.

docs/devsite-help/occ-for-iam.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Introduction to OCC
+
+Optimistic Concurrency Control (OCC) is a strategy used to manage shared resources and prevent "lost updates" or race conditions when multiple users or processes attempt to modify the same resource simultaneously.
+
+In the context of Google Cloud .NET libraries, IAM Policy objects contain an `Etag` property. When calling `SetIamPolicy`, the client library includes this `Etag`. If the server detects that the `Etag` provided does not match the current version on the server, it throws an RPC exception with the status `Aborted` or `FailedPrecondition`.
```
amanda-tarafa: Something like this to make it less IAM centric?

```suggestion
Google API resources contain an `Etag` property used by the server to control for optimistic concurrency. If the `Etag` valud provided when modifying a resource does not match the current value on the server, it returns an error, usually with status `Aborted` (Check the documentation of the API you are using for specific OCC error codes).
```

docs/devsite-help/occ-for-iam.md
```diff
@@ -0,0 +1,22 @@
...
+
+### **OCC Steps**
+
+| Step | Action | C\# Implementation |
```
amanda-tarafa: I think the example is enoug vs. adding the implemnation steps on this table? We'd then need to remember to change the table if we ever change the example.

docs/devsite-help/occ-for-iam.md
```diff
@@ -0,0 +1,22 @@
+# Optimistic Concurrency Control (OCC) for IAM
+
+## Introduction to OCC
+
+Optimistic Concurrency Control (OCC) is a strategy used to manage shared resources and prevent "lost updates" or race conditions when multiple users or processes attempt to modify the same resource simultaneously.
+
+In the context of Google Cloud .NET libraries, IAM Policy objects contain an `Etag` property. When calling `SetIamPolicy`, the client library includes this `Etag`. If the server detects that the `Etag` provided does not match the current version on the server, it throws an RPC exception with the status `Aborted` or `FailedPrecondition`.
+
+### **OCC Steps**
```
amanda-tarafa:
```suggestion
## **OCC Steps** demonstrated with IAM
```

docs/devsite-help/occ-for-iam.md
```diff
@@ -0,0 +1,22 @@
...
+
+| Step | Action | C\# Implementation |
+| ----- | ----- | ----- |
+| **1\. Read** | Fetch the current IAM Policy. | `await client.GetIamPolicyAsync(name)` |
+| **2\. Modify** | Apply changes to the `Policy` object. | Modify `policy.Bindings` collection. |
+| **3\. Write** | Attempt to set the policy. | `await client.SetIamPolicyAsync(name, policy)` |
+| **4\. Catch** | Catch specific `RpcException`. | `catch (RpcException ex) when (ex.StatusCode == StatusCode.Aborted)` |
+
+## Examples
```
amanda-tarafa:
```suggestion
### Example
```

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
```diff
@@ -0,0 +1,89 @@
...
+            // SUCCESS
+            Console.WriteLine("Successfully updated IAM policy.");
+        }
+        catch (RpcException ex) when (
+            ex.StatusCode == StatusCode.Aborted ||
+            ex.StatusCode == StatusCode.FailedPrecondition)
+        {
+            // A concurrency conflict can manifest as either Aborted or FailedPrecondition,
+            // depending on the specifics of the API and the race condition.
+            // To ensure the retry mechanism is robust, we catch both.
```
amanda-tarafa: Maybe:?

```suggestion
            // A concurrency conflict usually manifests as Aborted but dependeing on the API
            // other error codes may be used, like FailedPrecondition.
            // To ensure the retry mechanism is robust, check the documentation for the API
            // you are implementing OCC for.
```

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
```diff
@@ -0,0 +1,89 @@
...
+            // The 'policy' object contains the original 'Etag' from Step 1.
+            Console.WriteLine("Writing modified policy...");
+            Policy updatedPolicy = await client.SetIamPolicyAsync(resourceName, policy);
+
+            // SUCCESS
+            Console.WriteLine("Successfully updated IAM policy.");
+        }
+        catch (RpcException ex) when (
+            ex.StatusCode == StatusCode.Aborted ||
+            ex.StatusCode == StatusCode.FailedPrecondition)
```
amanda-tarafa:
```suggestion
        catch (RpcException ex) when ( ex.StatusCode == StatusCode.Aborted )
```

tools/Google.Cloud.Docs.Snippets/OccSnippets.cs
```diff
@@ -0,0 +1,89 @@
...
+    {
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void OccForIam()
+    {
+        string projectId = _ficxture.ProjectId;
+        string role = "roles/cloudkms.cryptoKeyEncrypterDecrypter";
+        string member = "user:betterbrent@google.com";
```
amanda-tarafa, line 36: Thinking about this more, there are a couple of things that might be a problem here:

- We are changing permissions on the project that we run tests in, and we have automation for maintaining permissions on it, that will complain about these changes.
- It's OK to harcode these values, but we should use more inocuos roles and less specific members.

Because we can already create/destroy topics on the fixture, what about if we use the Pub/Sub Publisher client instead to update the IAM policy on the topic, and we can set [the viewer](https://docs.cloud.google.com/iam/docs/roles-permissions/pubsub#pubsub.viewer) role on the topic for `"domain:google.com"`?

Let me know if you'd prefer me to work on that one.

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+    private readonly SnippetFixture _fixture;
+
+    public CallSettingsSnippets(SnippetFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void UpdateMasks()
```
amanda-tarafa: This won't build, it should be `async Task`

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
```diff
@@ -0,0 +1,68 @@
...
+    {
+        string projectId = _fixture.ProjectId;;
+        string secretId = "test-secret";
```
amanda-tarafa, line 34: We don't have a `test-secret` so this will fail. Either we create and destroy a secret on the fixture, or we change this example to use a Pub/Sub topic and use the utilities that we already have on the fixture.

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    public CallSettingsSnippets(SnippetFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void OccForIam()
+    {
+        string projectId = _ficxture.ProjectId;
```
amanda-tarafa: There's a typo here, this won't build. If you rebase your PR on upstream/main there's a new check that will fail.
```suggestion
        string projectId = _fixture.ProjectId;
```

tools/Google.Cloud.Docs.Snippets/OccForIamSnippets.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    private readonly SnippetFixture _fixture;
+
+    public CallSettingsSnippets(SnippetFixture fixture)
+    {
+        _fixture = fixture;
+    }
+
+    [Fact]
+    public void OccForIam()
```
amanda-tarafa: Or won't build otherwise:

```suggestion
    public async Task OccForIam()
```

amanda-tarafa (changes requested):
Let me know if you prefer I do the finishing touches.

tools/Google.Cloud.Docs.Snippets/OccSnippets.cs
bshaffer: Without knowing more about the way these tests are set up, I can't say I follow what the issue is we're trying to solve or how to solve it. If these are changes you feel like you could do easily instead, please do so!

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
bshaffer: I changed these because of [this comment](https://github.com/googleapis/google-cloud-dotnet/pull/15308#discussion_r2621177050), which I misunderstood to mean "remove `async Task`"

tools/Google.Cloud.Docs.Snippets/UpdateMaskSnippets.cs
bshaffer: Either way works for me, just let me know which you prefer (and maybe some guidance for how to set it up)

amanda-tarafa (commented):
I'll take it from here for the changes in the tests. Thanks!!!

And I'll send it for you to review once I've done that.

amanda-tarafa:
Closing in favour of #15423 which just tweks the sample code. Thanks @bshaffer !

PR #15315, opened 2025-12-10
fix: NRE in PredictionServiceChatClient

### LINKED ISSUE #15314: PredictionServiceChatClient NullReferenceException
#### Environment details

  - OS: MacOS
  - .NET version: 8
  - Package name and version: Google.Cloud.VertexAI.Extensions (unreleased)

#### Steps to reproduce

```
Object reference not set to an instance of an object.. Exception was: System.NullReferenceException: Object reference not set to an instance of an object.
       at Google.Protobuf.ByteString.CopyFrom(Byte[] bytes)
       at Google.Cloud.VertexAI.Extensions.PredictionServiceChatClient.AddPartsForAIContents(IList`1 contents, RepeatedField`1 parts) in PredictionServiceChatClient.cs:line 419
       at Google.Cloud.VertexAI.Extensions.PredictionServiceChatClient.CreateRequest(IEnumerable`1 messages, ChatOptions options) in PredictionServiceChatClient.cs:line 324
       at Google.Cloud.VertexAI.Extensions.PredictionServiceChatClient.GetResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken) in PredictionServiceChatClient.cs:line 67
```

I understand that the problem is on [this line](https://github.com/googleapis/google-cloud-dotnet/blob/af3bf06bc3b88186d9c1ec52bb9ecedac9d22d37/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs#L420). It seems to be a rare problem as we didn't experience it in our evaluations but only after deploying it to production.

cc @stephentoub 

I'll try to share an input that generate this exception.

verdie-g:
Fix #15314

I added a minimal repro as a test. The interaction is a little unusual as the assistant answers with two messages, one with text, one with a tool with the same text as argument. The reason is probably that we ask the assistant to always end an interaction with a specific tool call.

### DIFF
--- a/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/VertexAIExtensionsTest.cs
+++ b/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/VertexAIExtensionsTest.cs
@@ -39,6 +39,33 @@ public class VertexAIExtensionsTest
     }
 
     [Fact]
+    public async Task AsIChatClient_TextReasoningContent_WithToolCall_DoesNotThrow()
+    {
+        IChatClient client = await new PredictionServiceClientBuilder()
+            .BuildIChatClientAsync(EndpointName.FormatProjectLocationPublisherModel(s_projectId, s_location, "google", "gemini-2.5-pro"));
+
+        ChatMessage[] messages =
+        [
+            new(ChatRole.User, "Tell me something I don't know"),
+            new(ChatRole.Assistant,
+            [
+                new TextContent("I'm not quite sure what you mean."),
+                new TextReasoningContent(null) { ProtectedData = "CqEDAY89a18A3AfzAjWZZvPxwakP5Fx0TRN6OH2t" },
+            ]),
+            new(ChatRole.Assistant,
+            [
+                new FunctionCallContent("final_result", "final_result"),
+            ]),
+            new(ChatRole.Tool,
+            [
+                new FunctionResultContent("final_result", "Success: Function completed."),
+            ]),
+        ];
+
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions.IntegrationTests/VertexAIExtensionsTest.cs
```diff
@@ -38,6 +38,33 @@ public async Task AsIChatClient_BasicRequestResponse_NonStreaming()
...
+                new FunctionCallContent("final_result", "final_result"),
+            ]),
+            new(ChatRole.Tool,
+            [
+                new FunctionResultContent("final_result", "Success: Function completed."),
+            ]),
+        ];
+
+        await client.GetResponseAsync(messages);
+    }
```
stephentoub, line 66: I just saw your PR after putting up mine. Sorry for the duplication.

Our src fix is effectively the same. For tests, though, this shouldn't need an integration test to verify. My PR validates the change in a unit test which benefits from not needing the live service.

verdie-g: Ok I'll let Amanda choose which one to kill :D

amanda-tarafa: I saw @stephentoub 's first, but thanks both!

amanda-tarafa:
Closing in favor of #15317

PR #15316, opened 2025-12-10
chore: remove pregeneration hack for Artifact Registry

jskeet:
We should only merge this when cl/842641187 has been submitted.


amanda-tarafa (approved):
Nice

PR #15322, opened 2025-12-11
feat: Multiplex session v1 implementation

purva9413:


apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -0,0 +1,751 @@
...
+
+namespace Google.Cloud.Spanner.V1
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    public partial class ManagedTransaction
+    {
+        // All these fields are derived from the ManagedSession
+        private readonly Session _multiplexSession;
```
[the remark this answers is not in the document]
purva9413: I think this is more expressive on which type of session (multiplex vs pooled) from the backend is expected in the ManagedSession. Plus it is a private variable.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,345 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using static Google.Cloud.Spanner.V1.TransactionOptions;
+
+namespace Google.Cloud.Spanner.V1;
+
+/// <summary>
+/// TODO: Add summary for mux sessions
+/// </summary>
```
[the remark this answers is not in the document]
amanda-tarafa: +1 :)

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+public class ManagedSession
+{
+    private readonly Logger _logger;
+    private readonly CreateSessionRequest _createSessionRequestTemplate;
```
amanda-tarafa: I think this one can be static?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+    private Task _sessionCreationTask;
+    private readonly object _sessionCreationTaskLock = new object();
+
+    internal Session _session;
```
amanda-tarafa: private?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+
+    private const double HardRefreshIntervalInDays = 28.0;
+    private const double SoftRefreshIntervalInDays = 7.0;
```
amanda-tarafa: const usually at the top of the class definition

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+
+    private readonly IClock _clock;
+
+    /// <summary>
+    /// The client used for all operations in this managed session.
+    /// </summary>
+    internal SpannerClient Client { get; }
+
+    internal Task<Session> CreateSessionTask { get; }
```
amanda-tarafa: There's a variable for this also

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+
+    /// <summary>
+    /// The name of the session. This is never null.
+    /// </summary>
+    internal SessionName SessionName => Session.SessionName;
```
amanda-tarafa, line 59: It can be null now, before we acquire the first session?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+
+    /// <summary>
+    /// The Spanner session resource associated to this pooled session.
+    /// Won't be null.
+    /// </summary>
+    internal Session Session
```
amanda-tarafa, line 65: Do we need a property?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+
+    /// <summary>
+    /// The database for this managed session
+    /// </summary>
+    public DatabaseName DatabaseName { get; }
+
+    /// <summary>
+    /// The database role of the managed session
+    /// </summary>
+    public string DatabaseRole { get; }
```
amanda-tarafa, line 84: Arent these two on the ManagedSessionOptions already, and/or could we put them there?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+
+    /// <summary>
```
amanda-tarafa, line 86: Missing docs

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+                CreatorRole = DatabaseRole ?? "",
+                Multiplexed = true
+            }
+        };
+    }
+
+    /// <summary>
+    /// Returns a ManagedTransaction for the same ManagedSession as this with the given <paramref name="transactionOptions"/>.
+    /// </summary>
+    public async Task<ManagedTransaction> CreateManagedTransactionWithOptions(TransactionOptions transactionOptions, bool singleUseTransaction)
```
amanda-tarafa: I think I would name these CreateTransaction or CreateManagedTransaction, the parameters are clear and we don't have to call them out in the name of the method.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+    {
+        await MaybeRefreshWithTimePeriodCheck().ConfigureAwait(false);
+        return new ManagedTransaction(this, transactionId: null, transactionOptions, singleUseTransaction, readTimestamp: null);
```
amanda-tarafa, line 122: Do we need to pass `this` or could we just pass `_session`?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,347 @@
...
+
+    private async Task<bool> UpdateMuxSession(bool needsRefresh, double intervalInDays)
+    {
+        Session oldSession = _session;
+        await CreateOrRefreshSessionsAsync(default).ConfigureAwait(false);
+
+        return _session != oldSession;
+    }
+
+    internal async Task MaybeRefreshWithTimePeriodCheck()
```
amanda-tarafa: Why do we need 3 methods for this? Can't we do we a single one that also updates the variable at the end?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
purva9413: Well, the `SessionBuilder.BuildAsync()` is the only way for the customer to create a ManagedSession. And we do not return prematurely in the sense we only return a `ManagedSession` object once the underlying `Session` of the `ManagedSession` is created. Does that make sense?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
purva9413: I wanted to keep ManagedSessionOptions as optional configurable parameters for the ManagedSession (Similar to the SessionPoolOptions). The DatabaseName and DatabaseRole are required parameters for creating a multiplex session in Spanner (so by relation also for creating a ManagedSession), and are set on the `SessionBuilder` initially when the customer wants to build a ManagedSession.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
purva9413: There are a couple of things needed from the ManagedSession object when we create the ManagedTransaction. These are set in the constructor of the ManagedTransaction through the `this` ManagedSession object:

1) The client
2) The underlying session
3) Configurable options from the `ManagedSession.Options` like the RPC timeouts.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
purva9413: Consolidated this into just the one method `CreateOrRefreshSessionsAsync` as we discussed instead of having wrapping methods returning true/false depending on whether the session refreshed successfully.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
purva9413: Let's do `CreateManagedTransaction` with its overloads

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
purva9413: Removed, this was leftover code with nothing refrencing it. Thanks for catching!

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,318 @@
...
+            var callSettings = Client.Settings.CreateSessionSettings
+                .WithExpiration(Expiration.FromTimeout(Options.Timeout))
+                .WithCancellationToken(cancellationToken);
+
+            Session multiplexSession;
+
+            try
+            {
+                // Non-blocking, fast check if the session is still "fresh" enough
+                if (_session != null && !SessionHasExpired(HardRefreshIntervalInDays))
```
robertvoinescu-work, line 178: The `SessionHasExpired` treats the null case as expired. The null check shouldn't be necessary.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,318 @@
...
+                                    {
+                                        // Very likely we cannot continue without getting a fresh session in this case.
+                                        throw;
+                                    }
+                                }
+                                finally
+                                {
+                                    // Clear the task after completion (success or failure)
+                                    // to allow the next refresh to run.
+                                    _ = Interlocked.Exchange(ref _sessionCreationTask, null);
```
robertvoinescu-work, line 239: NIT: I don't think you need an interlocked exchange here since it's already executing within the context of a lock. But It should be fine, if we want to be defensive.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -0,0 +1,736 @@
...
+        internal TransactionOptions.ModeOneofCase TransactionMode => TransactionOptions.ModeCase;
+
+        private SpannerClient Client => _client;
+
+        /// <summary>
+        /// Whether the transaction is single use or not.
+        /// If this is true then <see cref="TransactionOptions"/> will be <see cref="TransactionOptions.ModeOneofCase.ReadOnly"/>
+        /// and <see cref="TransactionId"/> will be null.
+        /// </summary>
+        internal bool SingleUseTransaction { get; }
```
robertvoinescu-work, line 76: NIT: How do you feel about`IsSingleUseTransaction` to make it clear it's a boolean. Otherwise I see that name and think it might be a transaction object.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -168,7 +129,7 @@ private PooledSession(SessionPool.ISessionPool pool, SessionName sessionName, By
             GaxPreconditions.CheckArgument(
                 transactionId is null || TransactionOptions.ModeCase != ModeOneofCase.None,
                 nameof(transactionOptions),
-                "No transaction options were specified for the given transasaction ID.");
+                $"No transaction options were specified for the given transasaction ID {transactionId is null}, {transactionId?.ToBase64()}, {TransactionOptions.ModeCase}.");
```
robertvoinescu-work, line 132: This message would be hard to interpret without looking at the code, could you update it to something more like this:
```
$"No transaction options were specified. TransactionId: {transactionId?.ToBase64() ?? "null"}, Mode: {TransactionOptions.ModeCase}."

File: apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs [hunk 18]
```diff
@@ -600,13 +402,63 @@ async Task<CommitResponse> CommitAsync()
...
-                var response = await RecordSuccessAndExpiredSessions(Client.CommitAsync(request, callSettings)).ConfigureAwait(false);
-                MarkAsCommittedOrRolledBack();
-                return response;
+                CommitResponse finalResponse;
+
+                do
+                {
+                    // This loop will keep executing as long as we are signaled by Spanner to retry the Commit
+                    // It only exits if we no longer receive a PrecommitToken based retry enum from backend
+                    finalResponse = await ExecuteCommitOnceAsync(request, callSettings).ConfigureAwait(false);
```
robertvoinescu-work: Do we need any delays between the retires?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -693,19 +528,18 @@ public Task<PartitionResponse> PartitionQueryAsync(PartitionQueryRequest request
         /// from this object.</param>
         /// <param name="callSettings">If not null, applies overrides to this RPC call.</param>
         /// <returns>A task representing the asynchronous operation. When the task completes, the result is the response from the RPC.</returns>
-        internal Task<PartitionResponse> PartitionReadOrQueryAsync(PartitionReadOrQueryRequest request, CallSettings callSettings)
+        internal async Task<PartitionResponse> PartitionReadOrQueryAsync(PartitionReadOrQueryRequest request, CallSettings callSettings)
```
robertvoinescu-work, line 531: Why add the `async` modifier? I think it's marginally more efficient the original way. Same question other places like `RollBackAsync`

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs
```diff
@@ -866,33 +706,30 @@ private void MaybeApplyDirectedReadOptions(IReadOrQueryRequest request)
...
+                    PrecommitToken = token;
+                }
+            }
         }
 
-        private void CheckNotDisposed()
+        internal MultiplexedSessionPrecommitToken FetchPrecommitToken()
         {
-            if (Interlocked.CompareExchange(ref _disposed, 0, 0) == 1)
+            lock (_precommitTokenUpdateLock)
```
robertvoinescu-work: Instead of adding this method I suggest for simplicity we just update the property to match what's already being done for others:
```
public Timestamp ReadTimestamp => Interlocked.CompareExchange(ref _transaction, null, null)?.ReadTimestamp;

public MultiplexedSessionPrecommitToken PrecommitToken => Interlocked.CompareExchange(ref _transaction, null, null)?.SomethingSomething;

Comment (robertvoinescu-work) [reply apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedTransaction.cs]: Unless, this object is expected to be larger than 64bits

File: apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ResultStream.cs [hunk 21]
```diff
@@ -173,7 +215,7 @@ async Task<Transaction> ExecuteStreamingAsync()
                         hasNext = await MoveNextAsync().ConfigureAwait(false);
                     }
 
-                    Task<bool> MoveNextAsync() => _grpcCall.ResponseStream.MoveNext(cancellationToken).WithSessionExpiryChecking(_pooledSession.Session);
+                    Task<bool> MoveNextAsync() => _grpcCall.ResponseStream.MoveNext(cancellationToken).WithSessionExpiryChecking(_transaction != null ? _transaction.Session : _pooledSession.Session);
```
robertvoinescu-work, line 218: We can avoid this conditional check on each `MoveNextAsync` call by determining the session on construction, maybe save it to a private field like `ResultStream.Session`.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -56,11 +58,21 @@ static SessionPoolManager()
         private readonly ConcurrentDictionary<SessionPool, TargetedPool> _poolReverseLookup =
             new ConcurrentDictionary<SessionPool, TargetedPool>();
 
+        // Dictionary to store the sessions managed by this manager
+        // Each session is keyed by a complex key, as each (multiplex) session is unique to a combination of client options and unique segment key
+        private readonly ConcurrentDictionary<(SpannerClientCreationOptions, SessionPoolSegmentKey), Task<ManagedSession>> _targetedSessions =
```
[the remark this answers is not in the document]
robertvoinescu-work: NIT: Do we want to consider a record instead of a tuple here?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SessionPoolManager.cs
```diff
@@ -126,6 +159,27 @@ internal Task<SessionPool> AcquireSessionPoolAsync(SpannerClientCreationOptions
...
+            GaxPreconditions.CheckNotNull(options, nameof(options));
+
+            return _targetedSessions.GetOrAdd((options, segmentKey), _ => CreateMultiplexSessionAsync());
+
+            async Task<ManagedSession> CreateMultiplexSessionAsync()
+            {
+                var client = await _clientFactory.Invoke(options, SpannerSettings).ConfigureAwait(false);
+                var managedSessionBuilder = new SessionBuilder(dbName, client)
+                {
+                    Options = ManagedSessionOptions,
```
robertvoinescu-work, line 174: ManagedSessionOptions could be null here depending on the static constructor used to create the `SessionPoolmanager` do we need another null check here?

robertvoinescu-work: I might prefer a seperate class for this logic with a base class.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,317 @@
...
+        /// </summary>
+        public DatabaseName DatabaseName { get; set; }
+
+        /// <summary>
+        /// The database role of the managed session. This will be used when creating the underlying Spanner session.
+        /// </summary>
+        public string DatabaseRole { get; set; }
+
+        /// <summary>
+        /// The client used for all operations in this managed session.
```
robertvoinescu-work, line 300: We should add a note here similar to  `DataBasename` saying it is required.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,317 @@
...
+            // But with the MUX session, we essentially only have 1 session we need to manage per client.
+            // So there is no case of the mux session going back and forth in terms of its healthiness.
+        }
+
+    }
+
+    /// <summary>
+    /// Builder to build a <see cref="ManagedSession"/>
+    /// </summary>
+    public sealed partial class SessionBuilder
```
robertvoinescu-work: Why partial?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,317 @@
...
+    public async Task<ManagedTransaction> CreateManagedTransaction(TransactionOptions transactionOptions, bool singleUseTransaction, CancellationToken cancellationToken = default)
+    {
+        await CreateOrRefreshSessionsAsync(cancellationToken).ConfigureAwait(false);
+        return new ManagedTransaction(this, transactionId: null, transactionOptions, singleUseTransaction, readTimestamp: null);
+    }
+
+    /// <summary>
+    /// Returns a ManagedTransaction for the same ManagedSession as this one with the given transaction related values.
+    /// </summary>
+    public async Task<ManagedTransaction> CreateManagedTransaction(ByteString transactionId, TransactionOptions transactionOptions, bool singleUseTransaction, Timestamp readTimestamp = null, CancellationToken cancellationToken = default)
```
robertvoinescu-work, line 128: The names need to be updated to `CreateMangedTransactionAsync` for all methods.

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,317 @@
...
+    {
+        GaxPreconditions.CheckNotNull(transactionId, nameof(transactionId));
+        await CreateOrRefreshSessionsAsync(cancellationToken).ConfigureAwait(false);
+        return new ManagedTransaction(this, transactionId, transactionOptions, singleUseTransaction, readTimestamp);
+    }
+
+    /// <summary>
+    /// Returns a ManagedTransaction for the same ManagedSession as this one with the given transaction mode.
+    /// </summary>
+    public async Task<ManagedTransaction> CreateManagedTransaction(ByteString transactionId, ModeOneofCase transactionMode, Timestamp readTimestamp = null)
```
robertvoinescu-work, line 138: Should we take `bool singleUseTransaction` here also?

apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs
```diff
@@ -0,0 +1,317 @@
...
+        DateTime? sessionCreateTime = _session?.CreateTime.ToDateTime(); // Inherent conversion into UTC DateTime
+        if (_session == null || _session.Expired || currentTime - sessionCreateTime >= TimeSpan.FromDays(intervalInDays))
+        {
+            return true;
+        }
+
+        return false;
+    }
+
+    private async Task CreateOrRefreshSessionsAsync(CancellationToken cancellationToken, bool needsRefresh = false)
```
robertvoinescu-work, line 165: This is more of a high level comment. The current code works, but how do we feel about using `SemaphoreSlim` instead? It provides a natural way to deal with async locking without the need to manage the lifecycle of a `_sessionCreationTask`.

robertvoinescu-work: High level it'd be like this:
```csharp
private readonly SemaphoreSlim _sessionLock = new(1, 1);

private async Task TriggerRefresh(double refreshInterval, CancellationToken ct)
{
    await _sessionLock.WaitAsync(ct);
    try
    {
        if (SessionHasExpired(refreshInterval))
        {
            _session = await Client.CreateSessionAsync(...);
        }
    }
    finally
    {
        _sessionLock.Release();
    }
}

File: apis/Google.Cloud.Spanner.V1/Google.Cloud.Spanner.V1/ManagedSession.cs [hunk 29]
```diff
@@ -0,0 +1,317 @@
...
+
+/// <summary>
+/// This class helps manage Spanner multiplex session creation and lifecycle.
+/// A <see cref="ManagedSession"/> should only be created through a <see cref="SessionBuilder"/>.
+/// It provides factory methods to create <see cref="ManagedTransaction"/> which will be created using the current underlying session in this ManagedSession.
+/// A check to see if the underlying session needs to be refreshed or created will be done everytime a method to create a ManagedTransaction is called.
+/// The class will do a soft, non-blocking refresh if the underlying session is greater than 7 days but less than 28 days old from its creation.
+/// A hard, blocking refresh is done if the session is greater than 28 days old from its creation. 
+/// </summary>
+public class ManagedSession
```
robertvoinescu-work, line 37: Leaving as a reminder to our discussion, we will also need to update the affinity configs in `SpannerClientBuilder.MethodConfig` so we use transactionId instead of session as the affinity key.

amanda-tarafa:
Closing, superseded by #15414

PR #15328, opened 2025-12-12
fix: Initialize ChatResponseUpdate.MessageId with ResponseId

### LINKED ISSUE #15327: PredictionServiceChatClient doesn't set a value for MessageId
MessageId is always null in ChatResponseUpdate when using the PredictionServiceChatClient. This s breaking for AG-UI consumers, or any other client that depended on that logic to group message updates.

I'm not sure if it's a Google.Cloud.VertexAI.Extensions or a Microsoft.Agents.AI.Hosting.AGUI.AspNetCore issue. If it's the latter feel free to close this ticket.

#### Environment details

  - OS: Win11
  - .NET version: 8
  - Package name and version: Google.Cloud.VertexAI.Extensions version 1.0.0-beta01

#### Steps to reproduce

  1. Create a GCP Chat Client
  2. Wrap it in an AG-UI endpoint (  Microsoft.Agents.AI.Hosting.AGUI.AspNetCore )
  3. Response will have messageId at null which breaks all client code
`{"messageId":null,"delta":"I am Gemini, a large language model built by Google.","type":"TEXT_MESSAGE_CONTENT"}`

Having  `MessageId = generateResult.ResponseId,` during ChatResponseUpdate creation fixes it. 

Happy to send a PR for it if needed.

Thanks!

amanda-tarafa:
Fixes #15327

### DIFF
--- a/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
+++ b/apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
@@ -104,6 +104,7 @@ internal sealed class PredictionServiceChatClient(PredictionServiceClient client
                 ModelId = !string.IsNullOrEmpty(generateResult.ModelVersion) ? generateResult.ModelVersion : request.Model,
                 RawRepresentation = generateResult,
                 ResponseId = generateResult.ResponseId,
+                MessageId = generateResult.ResponseId,
             };
 
             // Populate the response update contents.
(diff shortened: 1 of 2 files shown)


apis/Google.Cloud.VertexAI.Extensions/Google.Cloud.VertexAI.Extensions/PredictionServiceChatClient.cs
robertvoinescu-work, line 70: Do we need to set a default `MessageId` here also?

amanda-tarafa: Nope, ChatResponse is a single response (not a streamed "part") so it doesn't have a `MessageId`, it's a single response to all the messages if you whish.

PR #15354, opened 2026-01-02
chore: Regenerate all APIs to update copyright year

jskeet:
Validation that this is only copyright changes:

```sh
git diff --cached --unified=0 | grep -v Copyright | grep -v '\-\-\-' | grep -v '+++' | grep -v @@ | grep -v "diff --git" | grep -v -E '^index'
```


jskeet:
(It would be useful if we could get this PR in today, before the next scheduled generation on Monday.)

jskeet:
I'll close the generation PR today so that this can go in first.

jskeet:
@robertvoinescu-work: If you could have a look at this today, that would be really handy so that I can merge it before the next generation tomorrow. (So that the only generated changes tomorrow will be from API updates instead of copyright years.)

robertvoinescu-work (approved):

PR #15364, opened 2026-01-09
feat: Implement IAsyncDisposable on DatastoreTransaction

Nasicus:
---- 

I noticed that Rollback is performed synchronously on a Datastore transaction when it's disposed.

So i drafted a PR to support `IAsyncDisposable` which seemed easy enough.
Not sure if I missed something important or if this could work like this.

Feel free to take this PR over (or even just the idea of it) if this makes sense for you too.

### DIFF
--- a/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreTransaction.cs
+++ b/apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreTransaction.cs
@@ -666,8 +666,7 @@ namespace Google.Cloud.Datastore.V1
         /// </summary>
         public virtual ValueTask DisposeAsync()
         {
-            Dispose();
-            return default;
+            throw new NotImplementedException();
         }
     }
 }


apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreTransaction.cs
```diff
@@ -659,5 +659,15 @@ public virtual void Dispose()
...
+
+        /// <summary>
+        /// If the transaction has already been committed, this operation is a no-op.
+        /// Otherwise, it rolls back the transaction.
+        /// </summary>
+        public virtual ValueTask DisposeAsync()
+        {
+            Dispose();
+            return default;
+        }
```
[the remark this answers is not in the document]
Nasicus: not sure I'd agree with that 🤷🏿

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreTransaction.cs
```diff
@@ -659,5 +659,15 @@ public virtual void Dispose()
...
             throw new NotImplementedException();
         }
+
+        /// <summary>
+        /// If the transaction has already been committed, this operation is a no-op.
+        /// Otherwise, it rolls back the transaction.
+        /// </summary>
+        public virtual ValueTask DisposeAsync()
+        {
+            Dispose();
```
amanda-tarafa: I think I'd throw here the same as in the rest of the methods. Regarding backwards compatibility it should be fine, derived classes either implemented `IAsyncDisposable` and their implementation will hide this one, or didn't in which case they couldn't be used in the context of an `IAsyncDisposable` before.

amanda-tarafa (changes requested):

apis/Google.Cloud.Datastore.V1/Google.Cloud.Datastore.V1/DatastoreTransaction.cs
Nasicus: good point! For some reason I assumed that it will automatically use the async dispose, but of course you have to write `await using...` :)

amanda-tarafa (approved):

PR #15366, opened 2026-01-09
chore: update GAPIC generator to 1.4.36

jskeet:


jskeet:
I suspect this may become obsolete soon anyway

PR #15369, opened 2026-01-09
feat(spanner): Add support for UUID data type.

robertvoinescu-work:
Add support for UUID data type for Spanner.

b/327986023

NOTE: Emulator doesn't support UUID, but i've run the integration tests against an actual spanner isntance and all pass:
image

### DIFF
--- a/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
+++ b/apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
@@ -110,6 +110,8 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
                 { "ProtobufValueArrayValue", SpannerDbType.ArrayOf(SpannerDbType.FromClrType(typeof(Value))), null },
                 { "ProtobufPersonArrayValue", SpannerDbType.ArrayOf(SpannerDbType.FromClrType(typeof(Person))), null },
                 { "ProtobufValueWrapperArrayValue", SpannerDbType.ArrayOf(SpannerDbType.FromClrType(typeof(ValueWrapper))), null },
+                { "UuidValue", SpannerDbType.Uuid, null },
+                { "UuidArrayValue", SpannerDbType.ArrayOf(SpannerDbType.Uuid), null },
             };
 
             if (_fixture.RunningOnEmulator || !isDml)
@@ -209,6 +211,9 @@ namespace Google.Cloud.Spanner.Data.IntegrationTests
             ValueWrapper testValueWrapper = new ValueWrapper { OneValue = Value.ForString("Hello") };
             ValueWrapper[] vwArray = { testValueWrapper, null, new ValueWrapper() };
 
+            var testUuid = Guid.NewGuid();
+            Guid?[] testUuidArray = { Guid.NewGuid(), null, Guid.NewGuid() };
+
             var parameters = new SpannerParameterCollection
             {
                 { "BoolValue", SpannerDbType.Bool, true },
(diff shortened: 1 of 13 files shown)


apis/Google.Cloud.Spanner.Data/run-cloud-tests.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
amanda-tarafa: We probably don't want this on the repo?

apis/Google.Cloud.Spanner.Data/run-all-types-tests.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
amanda-tarafa: Nor this one?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerDbTypeTests.ValueConversions.cs
```diff
@@ -331,6 +338,17 @@ public static IEnumerable<object[]> GetValidValueConversions()
             // Interval tests
             yield return new object[] { "P0Y", SpannerDbType.Interval, Quote("P0Y") };
             yield return new object[] { Interval.Parse("P1Y2M3DT4H5M6S"), SpannerDbType.Interval, Quote("P1Y2M3DT4H5M6S") };
+            // UUID tests
+            yield return new object[] { "8f8c4746-17b1-4d9f-a634-58e11942095f", SpannerDbType.Uuid, Quote("8f8c4746-17b1-4d9f-a634-58e11942095f")};
+            yield return new object[] { Guid.Parse("8f8c4746-17b1-4d9f-a634-58e11942095f"), SpannerDbType.Uuid, Quote("8f8c4746-17b1-4d9f-a634-58e11942095f") };
+            yield return new object[] { Guid.Parse("8f8c4746-17b1-4d9f-a634-58e11942095f"), SpannerDbType.Uuid, Quote("8f8c474617b14d9fa63458e11942095f"), TestType.ValueToClr };
```
amanda-tarafa, line 344: The Guid conversion shouldn't be lossy, so we should always be able to test the round-trip conversion. Why are you specifying here `ValueToClr` only? Am I missing something?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
```diff
@@ -117,6 +117,11 @@ private async Task ExecuteWriteNullsTest(Func<SpannerParameterCollection, Task<i
                 // b/348711708
                 parameters.Add("ProtobufValueValue", SpannerDbType.FromClrType(typeof(Value)), null);
             }
+            if (!_fixture.RunningOnEmulator)
```
amanda-tarafa: The latest version of the emulator already supports UUID (presumably). We already updated to the latest version yesterday or around.

apis/Google.Cloud.Spanner.Data/run-cloud-tests.sh
robertvoinescu-work: Probably not.. this was pushed by mistake, thanks! Removed.

apis/Google.Cloud.Spanner.Data/run-all-types-tests.sh
robertvoinescu-work: Likewise removed.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.IntegrationTests/WriteTests.cs
robertvoinescu-work: Thanks for the heads up, looks like everything is now passing with the emulator :). I've updated the code to not skip UUID in the case of emulator integration tests.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data.Tests/SpannerDbTypeTests.ValueConversions.cs
robertvoinescu-work: As discussed there is no loss, but this is a result of formatting. In particular Spanner can take the following formats:
```
00000000-0000-0000-0000-000000000000 
00000000000000000000000000000000 
```

But there are more supported asci string formats for a guid than those two for example:
```
{00000000-0000-0000-0000-000000000000} 
(00000000-0000-0000-0000-000000000000)
```
The current implementations enables the client to understand all these formats as inputs, but when providing a guid as an ascii string it will choose to provide that in the standard hyphenated format.

PR #15381, opened 2026-01-14
chore: remove obsolete ForEachAsync utility method

jskeet:
We don't use this in generated snippets any more, and the handwritten snippets have been updated to use await foreach.


jskeet:
Will remove all the redundant semi-colons...

PR #15395, opened 2026-01-21
feat(Storage): Enable full object checksum validation for resumable uploads

mahendra-google:
This PR enables full object checksum validation (specifically CRC32C) for JSON-based resumable uploads, ensuring end-to-end data integrity for multi-chunk transfers. Please see [b/461996128](http://b/461996128)

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2017 Google Inc. All Rights Reserved.
+// Copyright 2017 Google Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -12,26 +12,220 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Apis.Http;
 using Google.Apis.Services;
+using Google.Apis.Upload;
 using System;
 using System.IO;
+using System.Linq;
 using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
 using static Google.Apis.Storage.v1.ObjectsResource;
-using Google.Apis.Upload;
 
 namespace Google.Cloud.Storage.V1
 {
     /// <summary>
     /// Upload subclass which allows us to modify headers, for customer-supplied encryption keys.
     /// </summary>
-    internal sealed class CustomMediaUpload : InsertMediaUpload
+    internal sealed class CustomMediaUpload : InsertMediaUpload, IDisposable
     {
+        private readonly Crc32cHashInterceptor _interceptor;
+        private readonly IClientService _service;
+        private readonly HashingStream _hashingStream;
+        private bool _disposed;
+
         public CustomMediaUpload(IClientService service, Apis.Storage.v1.Data.Object body, string bucket,
-            Stream stream, string contentType)
-            : base(service, body, bucket, stream, contentType)
+            Stream stream, string contentType, UploadObjectOptions options)
+            : base(service, body, bucket, options?.UploadValidationMode != UploadValidationMode.None ? new HashingStream(stream) : stream, contentType)
         {
+            _service = service;
+            var validationMode = options?.UploadValidationMode ?? UploadObjectOptions.DefaultValidationMode;
+            if (validationMode != UploadValidationMode.None)
+            {
+                _hashingStream = ContentStream as HashingStream;
+                _interceptor = new Crc32cHashInterceptor(this, _hashingStream, _service);
+                _service?.HttpClient?.MessageHandler?.AddExecuteInterceptor(_interceptor);
+            }
         }
 
         internal new ResumableUploadOptions Options => base.Options;
+
+        public void Dispose()
+        {
+            if (!_disposed)
+            {
+                _service?.HttpClient?.MessageHandler?.RemoveExecuteInterceptor(_interceptor);
+                _interceptor?.Dispose();
+                _disposed = true;
+            }
+        }
+
+        private sealed class Crc32cHashInterceptor : IHttpExecuteInterceptor, IDisposable
+        {
+            private const string GoogleHashHeader = "x-goog-hash";
+            private readonly IClientService _service;
+            private readonly CustomMediaUpload _mediaUpload;
+            private Uri _uploadUri;
+            private readonly HashingStream _hashingStream;
+
+            public Crc32cHashInterceptor(CustomMediaUpload mediaUpload, HashingStream hashingStream, IClientService service)
+            {
+                _hashingStream = hashingStream;
+                _service = service;
+                _mediaUpload = mediaUpload;
+                _mediaUpload.UploadSessionData += OnSessionData;
+                _mediaUpload.ProgressChanged += OnProgressChanged;
+            }
+
+            public Task InterceptAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+            {
+                if (_uploadUri == null || (!ReferenceEquals(_uploadUri, request.RequestUri) && !_uploadUri.Equals(request.RequestUri)))
+                {
+                    return Task.CompletedTask;
+                }
+
+                if (request.Method == System.Net.Http.HttpMethod.Put && request.Content?.Headers.Contains("Content-Range") is true)
+                {
+                    var rangeHeader = request.Content.Headers.GetValues("Content-Range").First();
+
+                    if (IsFinalChunk(rangeHeader))
+                    {
+                        var calculatedHash = _hashingStream.GetBase64Hash();
+                        request.Headers.TryAddWithoutValidation(GoogleHashHeader, $"crc32c={calculatedHash}");
+                    }
+                }
+                return Task.CompletedTask;
+            }
+
+            private void OnSessionData(IUploadSessionData data)
+            {
+                _uploadUri = data.UploadUri;
+                _mediaUpload.UploadSessionData -= OnSessionData;
+            }
+
+            private void OnProgressChanged(IUploadProgress progress)
+            {
+                if (progress.Status is UploadStatus.Completed or UploadStatus.Failed)
+                {
+                    // Clean up when upload is finished.
+                    _service?.HttpClient?.MessageHandler?.RemoveExecuteInterceptor(this);
+                    _mediaUpload.ProgressChanged -= OnProgressChanged;
+                }
+            }
+
+            public void Dispose()
+            {
+                _mediaUpload.UploadSessionData -= OnSessionData;
+                _mediaUpload.ProgressChanged -= OnProgressChanged;
+            }
+        }
+
+        private static bool IsFinalChunk(string rangeHeader)
+        {
+            // Expected format: "bytes {start}-{end}/{total}" or "bytes */{total}" for the final request.
+            // We are interested in the final chunk of a known-size upload.
+            const string prefix = "bytes ";
+            if (!rangeHeader.StartsWith(prefix, StringComparison.OrdinalIgnoreCase))
+            {
+                return false;
+            }
+
+            ReadOnlySpan<char> span = rangeHeader.AsSpan(prefix.Length);
+            int slashIndex = span.IndexOf('/');
+            if (slashIndex == -1)
+            {
+                return false;
+            }
+
+            var totalSpan = span.Slice(slashIndex + 1);
+            if (totalSpan.IsEmpty || totalSpan[0] == '*')
+            {
+                return false;
+            }
+
+            if (!long.TryParse(totalSpan.ToString(), System.Globalization.NumberStyles.Integer, System.Globalization.CultureInfo.InvariantCulture, out long totalSize))
+            {
+                return false;
+            }
+
+            var rangeSpan = span.Slice(0, slashIndex);
+            int dashIndex = rangeSpan.IndexOf('-');
+            if (dashIndex == -1)
+            {
+                return false;
+            }
+
+            var endByteSpan = rangeSpan.Slice(dashIndex + 1);
+            if (!long.TryParse(endByteSpan.ToString(), System.Globalization.NumberStyles.Integer, System.Globalization.CultureInfo.InvariantCulture, out long endByte))
+            {
+                return false;
+            }
+
+            // If endByte is the last byte of the file, it's the final chunk.
+            return (endByte + 1) == totalSize;
+        }
+
+        internal sealed class HashingStream : Stream
+        {
+            private readonly Stream _stream;
+            private readonly Crc32c _hasher;
+            private long _maxPositionHashed = 0;
+
+            public HashingStream(Stream stream)
+            {
+                _stream = stream;
+                _hasher = new Crc32c();
+            }
+
+            public override int Read(byte[] buffer, int offset, int count)
+            {
+                long startingPos = _stream.Position;
+                int bytesRead = _stream.Read(buffer, offset, count);
+                ProcessBytes(buffer, offset, bytesRead, startingPos);
+                return bytesRead;
+            }
+
+            public override async Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
+            {
+                long startingPos = _stream.Position;
+                int bytesRead = await _stream.ReadAsync(buffer, offset, count, cancellationToken).ConfigureAwait(false);
+                ProcessBytes(buffer, offset, bytesRead, startingPos);
+                return bytesRead;
+            }
+
+            private void ProcessBytes(byte[] buffer, int offset, int bytesRead, long startingPos)
+            {
+                if (bytesRead <= 0) return;
+
+                // Only hash bytes that are beyond the furthest point we've already hashed.
+                // This handles the rewind and re-read scenario during retries.
+                if (startingPos + bytesRead > _maxPositionHashed)
+                {
+                    long newBytesStart = Math.Max(startingPos, _maxPositionHashed);
+                    int actuallyNewCount = (int) ((startingPos + bytesRead) - newBytesStart);
+                    int bufferOffset = offset + (int) (newBytesStart - startingPos);
+
+                    _hasher.UpdateHash(buffer, bufferOffset, actuallyNewCount);
+                    _maxPositionHashed = startingPos + bytesRead;
+                }
+            }
+
+            public override long Position
+            {
+                get => _stream.Position;
+                set => _stream.Position = value;
+            }
+
+            public override long Seek(long offset, SeekOrigin origin) => _stream.Seek(offset, origin);
+            public string GetBase64Hash() => Convert.ToBase64String(_hasher.GetHash());
+            public override bool CanRead => _stream.CanRead;
+            public override bool CanSeek => _stream.CanSeek;
+            public override bool CanWrite => _stream.CanWrite;
+            public override long Length => _stream.Length;
+            public override void Flush() => _stream.Flush();
+            public override void SetLength(long value) => _stream.SetLength(value);
+            public override void Write(byte[] buffer, int offset, int count) => _stream.Write(buffer, offset, count);
+        }
     }
 }
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UploadObject.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UploadObject.cs
@@ -48,7 +48,7 @@ namespace Google.Cloud.Storage.V1
         {
             ValidateObject(destination, nameof(destination));
             GaxPreconditions.CheckNotNull(source, nameof(source));
-            var mediaUpload = new CustomMediaUpload(Service, destination, destination.Bucket, source, destination.ContentType);
+            var mediaUpload = new CustomMediaUpload(Service, destination, destination.Bucket, source, destination.ContentType, options);
             options?.ModifyMediaUpload(mediaUpload);
             ApplyEncryptionKey(options?.EncryptionKey, options?.KmsKeyName, mediaUpload);
             return mediaUpload;
@@ -143,9 +143,6 @@ namespace Google.Cloud.Storage.V1
         {
             private readonly StorageClient _client;
             private readonly ObjectsResource.InsertMediaUpload _mediaUpload;
-            private readonly Crc32c _crc;
-            private readonly Action<Object> _validationFailureAction;
-            private readonly Func<Object, CancellationToken, Task> _validationFailureAsyncAction;
 
             internal UploadHelper(
                 StorageClient client,
@@ -160,22 +157,6 @@ namespace Google.Cloud.Storage.V1
                 {
                     _mediaUpload.ProgressChanged += progress.Report;
                 }
-
-                var validationMode = options?.UploadValidationMode ?? UploadObjectOptions.DefaultValidationMode;
-                GaxPreconditions.CheckEnumValue(validationMode, nameof(UploadObjectOptions.UploadValidationMode));
-                switch (validationMode)
-                {
-                    case UploadValidationMode.DeleteAndThrow:
-                        _crc = new Crc32c();
-                        _mediaUpload.UploadStreamInterceptor += _crc.UpdateHash;
-                        _validationFailureAction = obj => client.DeleteObject(obj, new DeleteObjectOptions { Generation = obj.Generation });
-                        _validationFailureAsyncAction = (obj, token) => client.DeleteObjectAsync(obj, new DeleteObjectOptions { Generation = obj.Generation }, token);
-                        break;
-                    case UploadValidationMode.ThrowOnly:
-                        _crc = new Crc32c();
-                        _mediaUpload.UploadStreamInterceptor += _crc.UpdateHash;
-                        break;
-                }
             }
 
             internal Object Execute()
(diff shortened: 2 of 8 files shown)


apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
```diff
@@ -26,12 +30,151 @@ namespace Google.Cloud.Storage.V1
...
+                if (dashIndex == -1)
+                {
+                    return false;
+                }
+
+                var endByteSpan = rangeSpan.Slice(dashIndex + 1);
+                if (!long.TryParse(endByteSpan.ToString(), System.Globalization.NumberStyles.Integer, System.Globalization.CultureInfo.InvariantCulture, out long endByte))
+                {
+                    return false;
+                }
```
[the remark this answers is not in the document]
mahendra-google: Removal of `ToString()` from `endByteSpan.ToString()` gives compilation error. It looks like code suggestion does not work with existing target framework.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
```diff
@@ -26,12 +30,151 @@ namespace Google.Cloud.Storage.V1
...
+                var totalSpan = span.Slice(slashIndex + 1);
+                if (totalSpan.IsEmpty || totalSpan[0] == '*')
+                {
+                    return false;
+                }
+
+                if (!long.TryParse(totalSpan.ToString(), System.Globalization.NumberStyles.Integer, System.Globalization.CultureInfo.InvariantCulture, out long totalSize))
+                {
+                    return false;
+                }
```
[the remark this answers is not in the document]
mahendra-google: Removal of `ToString()` from `totalSpan.ToString()` gives a compilation error.  It looks like code suggestion does not work with existing target framework.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
krishnamd-jkp: Adding this globally to the http client which is not ideal. Every single request GET, LIST, DELETE made by this client must now run through the URI check in InterceptAsync. Also if there is an exception and if OnProgressChanged is not triggered, the interceptor stays in the client forever.

krishnamd-jkp: On the final chunk, you are calculating the hash for entire stream which could be in Gbs. This will cause  a performance regression for huge data. Incase of retries, this logic will be computed multiple times again

krishnamd-jkp: You're not checking if the user has disabled this (using UploadValidationMode.None) crc32c validation here.

krishnamd-jkp: I think you are modifying the global stream's position here which could be a side effect. If this stream is used concurrently somewhere, that could lead to an issue

krishnamd-jkp (changes requested):

krishnamd-jkp:
Can you use UploadStreamInterceptor which already calculates the hash and use it here instead of recomputing the hash on final chunk?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
mahendra-google: The `CustomMediaUpload` class is specialized for upload operations only , utilizing `POST` and `PUT`  HTTP requests only . In the event of an exception whether it is originating from the user , the network or the API , the `progress.Status` transitions to `uploadStatus.Failed`. At this point, the interceptor should be removed.

mahendra-google: The CRC32C hash is calculated by re-reading the entire stream asynchronously. In the event of a retry, this calculation will trigger only if the failure occurs during the finalization phase.

mahendra-google: We are modifying the passed stream's position which we are trying to upload.  
1. `_stream.Position = 0`
By the time the code reaches `IsFinalChunk` , the stream pointer is likely at the very end of the file . To calculate a checksum for the whole file, the hasher needs to see every single byte from the beginning. Setting the position to 0 allows the while loop to re-read the entire stream from start to finish.

2. `_stream.Position = originalPosition`
It saves the originalPosition before it starts and restores it afterward. This is an attempt to avoid the side effect of leaving the stream at the end of the file, assuming that some other part of the program might still expect the stream to be at its original state.

mahendra-google:
> Can you use UploadStreamInterceptor which already calculates the hash and use it here instead of recomputing the hash on final chunk?

> @krishnamd-jkp UploadStreamInterceptor does not calculate hash . Please see https://github.com/googleapis/google-api-dotnet-client/blob/main/Src/Support/Google.Apis/Upload/ResumableUpload.cs#L207

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
krishnamd-jkp: Even asynchronously re-reading the entire 100GB stream at the very end of the upload just to calculate the hash will cause a massive, unpredictable latency spike

mahendra-google: If user has disabled UploadValidation using (UploadValidationMode.None). I have added condition to skip adding crc32c hash in `x-goog-hash`  header.

mahendra-google: @krishnamd-jkp  I feel that there is a trade-off between integrity-guarantees at the server side before objects get fully uploaded and performance. Could you suggest something to reduce performance regression ?

mahendra-google: @krishnamd-jkp  Thank you for the feedback regarding the performance regression. You're absolutely right- re-reading a large stream at the final chunk is not scalable.

I have refactored the implementation to use a ***HashingStream*** wrapper (a decorator pattern).

Key changes in this update:

**Incremental Calculation:** The CRC32C hash is now updated byte-by-byte during the Read operations of the upload itself.

**Zero Latency Spike:** By the time the final chunk is reached, the hash is already fully calculated. We no longer re-read the stream or reset its **_Position_** to 0.

This should resolve the latency and retry concerns. **Please feel free to suggest any further tweaks!**

mahendra-google: @krishnamd-jkp  As per new approach we are no longer modifying the stream's position.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
```diff
@@ -26,12 +30,154 @@ namespace Google.Cloud.Storage.V1
...
     {
+        private readonly Crc32cHashInterceptor _interceptor;
+        private readonly IClientService _service;
+        private readonly HashingStream _hashingStream;
+
         public CustomMediaUpload(IClientService service, Apis.Storage.v1.Data.Object body, string bucket,
-            Stream stream, string contentType)
-            : base(service, body, bucket, stream, contentType)
+            Stream stream, string contentType, UploadObjectOptions options)
+            : base(service, body, bucket, new HashingStream(stream), contentType)
```
krishnamd-jkp: We are not checking if user has disabled checksumming when we are wrapping this HashingStream.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
```diff
@@ -26,12 +30,154 @@ namespace Google.Cloud.Storage.V1
...
+        }
+
+        private sealed class HashingStream : Stream
+        {
+            private readonly Stream _stream;
+            private readonly Crc32c _hasher = new();
+
+            public HashingStream(Stream stream) => _stream = stream;
+
+            public override int Read(byte[] buffer, int offset, int count)
```
krishnamd-jkp, line 181: I think without the state of previous reads, any retries in uploads will cause recalculation of hash and the final hash will be wrong. Have we validated this for retries also?

krishnamd-jkp, line 181: Also, should there be a ReadAsync method also? If not, I think calling from an Async method will be synchronous for the user.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
```diff
@@ -26,12 +30,154 @@ namespace Google.Cloud.Storage.V1
...
+
         public CustomMediaUpload(IClientService service, Apis.Storage.v1.Data.Object body, string bucket,
-            Stream stream, string contentType)
-            : base(service, body, bucket, stream, contentType)
+            Stream stream, string contentType, UploadObjectOptions options)
+            : base(service, body, bucket, new HashingStream(stream), contentType)
         {
+            _service = service;
+            var validationMode = options?.UploadValidationMode ?? UploadObjectOptions.DefaultValidationMode;
+            _hashingStream = (HashingStream) ContentStream;
```
krishnamd-jkp: We are not checking if user has disabled it before wrapping the stream with hashing stream.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
mahendra-google: @krishnamd-jkp  I have now overridden `ReadAsync` method of the base `Stream` class in `HashingStream`  wrapper.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
mahendra-google: I am now initializing the `Crc32cHashInterceptor` and injecting it into the HTTP request pipeline only when `UploadValidationMode` is set to value other than `UploadValidationMode.None`

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
mahendra-google: Added a conditional check to ensure we only instantiate `HashingStream` when `UploadValidationMode` is not `None`. This prevents unnecessary hashing overhead when the user has disabled checksumming.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
mahendra-google: You are absolutely right—great catch. In its previous state, the `HashingStream` was strictly forward only. If the client library encountered a transient error and sought the stream back to retry an upload, the Crc32c hasher would have continued accumulating data from the rewind point, leading to a corrupted final hash.

I have updated the implementation to make the `HashingStream` retry-aware. Stream now detects any backward seek or position reset and re-instantiates the hasher.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
```diff
@@ -26,12 +30,195 @@ namespace Google.Cloud.Storage.V1
...
+                }
+                return bytesRead;
+            }
+
+            public override long Position
+            {
+                get => _stream.Position;
+                set
+                {
+                    if (value < _stream.Position)
```
krishnamd-jkp, line 188: This need not always be reset to 0. So in cases where the position is not reset to 0, the hash will be incorrect.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
```diff
@@ -26,12 +30,195 @@ namespace Google.Cloud.Storage.V1
...
+                _hashingStream = hashingStream;
+                _service = service;
+                _mediaUpload = mediaUpload;
+                _mediaUpload.UploadSessionData += OnSessionData;
+                _mediaUpload.ProgressChanged += OnProgressChanged;
+            }
+
+            public Task InterceptAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+            {
+                if (_uploadUri != null && !_uploadUri.Equals(request.RequestUri))
```
krishnamd-jkp, line 72: What happens if the uploadUri is null? SHould we be returning Task.CompletedTask in that case also?

mahendra-google: @krishnamd-jkp   That’s a great question. The current logic uses `(_uploadUri != null && !_uploadUri.Equals(request.RequestUri)) ` as a guard clause to ensure we return `Task.CompletedTask` without allowing the `IsFinalChunk` check to evaluate , if we are certain the request is pointing elsewhere.

If `_uploadUri` is null . In this scenario, we choose to proceed with the method (eventually returning `Task.CompletedTask` at the end) to allow the `IsFinalChunk` check to evaluate.

It would be good idea to switch to `_uploadUri == null || !_uploadUri.Equals(request.RequestUri)
` which would return early for every request where the URI isn't yet set, which might prevent the checksum from being attached to the final chunk. I have modified condition from `(_uploadUri != null && !_uploadUri.Equals(request.RequestUri)) ` to `(_uploadUri == null || !_uploadUri.Equals(request.RequestUri)`

krishnamd-jkp (approved):

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
mahendra-google: Good catch! You're absolutely right—since the position isn't always reset to 0, the hash calculation would be incorrect in those scenarios. I have updated logic to account for the current position instead of assuming a zero-start. Thanks for pointing that out!

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
```diff
@@ -26,12 +30,185 @@ namespace Google.Cloud.Storage.V1
...
+            Stream stream, string contentType, UploadObjectOptions options)
+            : base(service, body, bucket, options?.UploadValidationMode != UploadValidationMode.None ? new HashingStream(stream) : stream, contentType)
         {
+            _service = service;
+            var validationMode = options?.UploadValidationMode ?? UploadObjectOptions.DefaultValidationMode;
+            if (validationMode != UploadValidationMode.None)
+            {
+                _hashingStream = ContentStream as HashingStream;
+                _interceptor = new Crc32cHashInterceptor(this, _hashingStream, _service);
+                _service?.HttpClient?.MessageHandler?.AddExecuteInterceptor(_interceptor);
```
amanda-tarafa, line 48: The same service may be used in parallel for many different requests, including for many different uploads. You'll have interceptors that will be applied to all requests???

Even when you check the URI, assuming that check is enough in all cases, you are adding latency to all requests.

What am I missing?

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
```diff
@@ -26,12 +30,185 @@ namespace Google.Cloud.Storage.V1
...
+            {
+                if (progress.Status is UploadStatus.Completed or UploadStatus.Failed)
+                {
+                    // Clean up when upload is finished.
+                    _service?.HttpClient?.MessageHandler?.RemoveExecuteInterceptor(this);
+                    _mediaUpload.ProgressChanged -= OnProgressChanged;
+                }
+            }
+
+            private bool IsFinalChunk(string rangeHeader)
```
amanda-tarafa: Why don't you store when you've read 0 bytes from the stream (you are already wrapping the stream) instead of all of this. The request that comes after that is the last request.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UploadObject.cs
amanda-tarafa: Why do we continue to check the hash client side? Which means, btw, we are hashing twice, see the `UploadHelper` class starting at around line 150.

amanda-tarafa (dismissed):

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UploadObject.cs
mahendra-google: Removed the client-side hash validation in `UploadHelper` (lines 150+) to eliminate redundant hashing based on `UploadValidationMode`. By removing this check, we avoid the overhead of hashing the object twice. Updated `Execute` and `ExecuteAsync` by removing the manual exception handling block for hash mismatches. We now rely entirely on the server-side validation for data integrity.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
mahendra-google: Great suggestion. I have refactored `HashingStream` to include an `EndOfStreamReached` property. This flag is set as soon as the internal `Read`/`ReadAsync` call returns 0 bytes.

This allowed me to remove the `IsFinalChunk` helper method and the string parsing of range headers. The upload logic now simply checks the stream's state to determine if it's the final request. Updated in the latest commit.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
mahendra-google: You make a very fair point regarding the service-wide scope of the interceptor.

While I have included logic to remove the interceptor upon upload `completion` or `failure` (to prevent long-term memory leaks on line `94` of the `CustomMediaUpload` class), I recognize that during high-concurrency periods, the interceptor chain could grow. This would indeed force every request—even those unrelated to this upload—to perform the `URI ` check.

I initially assumed the latency of a reference comparison would be negligible, but I see how this architectural choice could be `suboptimal` for shared service instances. Since I want to ensure this implementation aligns with our performance standards, I am very open to your suggestions.

mahendra-google: Updated the `URI` comparison to include a reference equality check (`ReferenceEquals`) before performing a value-based `.Equals()`. This introduces an $O(1)$ short-circuit optimization especially for requests other than upload . For unrelated requests , we can now `fail fast` at the CPU register level by comparing memory addresses rather than performing a character-by-character string or component validation.This bypasses the $O(n)$ complexity of a full `URI` equality check, significantly reducing CPU cycles and latency for the majority of requests where the object references do not match.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UploadValidationException.cs
```diff
@@ -22,6 +22,8 @@ namespace Google.Cloud.Storage.V1
     /// <summary>
     /// Exception thrown when an upload failed validation.
     /// </summary>
+    [Obsolete("This exception is no longer thrown. Server-side validation is now performed, " +
```
amanda-tarafa, line 25: Note that if you stop throwing the exception it is a breaking change. Users that were expecting this to be thrown won't know that the checksum failed unless they change their code. The right path here is to bump major version.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
amanda-tarafa: My recommendation here is that you add an event on the base upload type that triggers on the last request only. This would be in `Google.Apis`. (Note that this is different to the initial proposal in `Google.Apis` which meant changing upload logic for all Discovery based libraries).

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UploadObject.cs
```diff
@@ -160,43 +157,13 @@ internal UploadHelper(
                 {
                     _mediaUpload.ProgressChanged += progress.Report;
                 }
-
-                var validationMode = options?.UploadValidationMode ?? UploadObjectOptions.DefaultValidationMode;
-                GaxPreconditions.CheckEnumValue(validationMode, nameof(UploadObjectOptions.UploadValidationMode));
-                switch (validationMode)
-                {
-                    case UploadValidationMode.DeleteAndThrow:
```
amanda-tarafa, line 168: Does this implementation respects both `DeleteAndThrow` and `ThrowOnly`? I don't see where the service would know the difference, so this would be another breaking change, that warrants a new major version bump.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/UploadValidationException.cs
mahendra-google: You’re absolutely right—altering the error-handling flow would definitely break the contract for any users currently relying on that exception to trigger their own validation logic.

I agree that maintaining the integrity of the current version is the priority. I will pivot this update to a major version bump to ensure we follow proper semantic versioning and give our users a clear signal that the checksum behavior has changed.

Thank you for the guidance on this; it’s better to do it right than to break existing.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/StorageClientImpl.UploadObject.cs
mahendra-google: I appreciate you flagging that—ensuring we maintain backward compatibility is definitely a priority.

The implementation does respect both `DeleteAndThrow `and `ThrowOnly`. In the `CustomMediaUpload ` class, I am passing the `UploadObjectOptions ` instance directly into the constructor. The logic is scoped so that we only calculate the hash if the `UploadValidationMode` is set to something other than `None`. This allows the service to distinguish between the modes effectively without altering the existing contract.

amanda-tarafa: The hash is indeed only calculated if the value is set to other than `None`, that is correct.

The implementation does not respect the difference between `DeleteAndThrow` and `ThrowOnly`. These values are not being passed to the service, and they are not being enforced client side either as far as I can see.

- `ThrowOnly`: Throws an exception if the hashes do not match, but the object remains in storage.
- `DeleteAndThrow`: Throws an exception if the hashes do not match **and** removes the object from storage.

mahendra-google: You are completely right, and thank you for pointing that out. The legacy `DeleteAndThrow ` and `ThrowOnly ` behaviors weren't being properly enforced .

To address this comprehensively and lean into a more robust architecture, I have refactored the implementation to leverage server-side validation instead of relying on client-side cleanup.

Here is a summary of the changes made in this update:

Introduced `UploadValidationMode.RejectAndThrow`: This leverages server-side validation, ensuring that invalid uploads are rejected automatically before the object is ever created.

Updated Default Upload Validation Mode & Tests: `UploadObjectOptions` now uses this new mode as the default behavior  and `UploadObjectTest` has been updated to reflect this.

Deprecated Legacy Modes: We have deprecated `ThrowOnly` and `DeleteAndThrow`. Because server-side validation prevents the object from being created in the first place . By switching to server-side rejection, the object is blocked at the front door. 

1. It satisfies `DeleteAndThrow` because the final state of the bucket is perfectly clean—no invalid object exists.
2. It satisfies (and improves upon) `ThrowOnly` because it alerts the user via an exception, but removes the dangerous side effect of leaving a corrupted file behind.

amanda-tarafa: > It satisfies (and improves upon) ThrowOnly because it alerts the user via an exception, but removes the dangerous side effect of leaving a corrupted file behind.

This is a breaking change. Deprecation is a signal that the behavior will stop working on the next major version. But while an element is deprecated, its behavior must be enforced. And it's not a dangerous side effect, it's a behavior explicitly requested by user when they specify `ThrowOnly`.

amanda-tarafa:
When you are ready to merge, please use squash and merge so we don't get the 16 iteration commits on the repo history. As an alternative, rebase the PR on top of main into a single commit before merging.

amanda-tarafa:
Regarding `do not merge` see b/461996128#comment9

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/CustomMediaUpload.cs
mahendra-google: @amanda-tarafa  Recommended approach is being implemented.  I will be raising separate PRs for both `Google.Apis` and `storage library` soon.

amanda-tarafa: Thanks. Because that is a change in the core libraries the .NET Cloud SDK team will need to review thorughly. A 1-2 pages design document describing the proposed change, before the PR is sent for review, should help expediting things.

amanda-tarafa:
Closing in favour of #15769

PR #15399, opened 2026-01-21
feat(PubSub): Nack pending messages during shutdown.

robertvoinescu-work:
These changes ensure that pending messages are Nack'ed instead of dropped during shutdown.

b/427314526

### DIFF
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
@@ -2017,6 +2017,53 @@ namespace Google.Cloud.PubSub.V1.Tests
             });
         }
 
+        [Fact]
+        public void NackMessagesOnShutdown()
+        {
+            var msgs = new[] { new[] {
+                ServerAction.Data(TimeSpan.Zero, ["msg0"]),
+                ServerAction.Data(TimeSpan.Zero, ["msg1", "msg2", "msg3"]),
+                ServerAction.Data(TimeSpan.Zero, ["msg4", "msg5"]),
+                ServerAction.Inf()
+            } };
+            // Set flowMaxElements to 2 to ensure that "msg0" and "msg2" block flow control preventing
+            // "msg3","msg4","msg5" from being processed.
+            using (var fake = Fake.Create(msgs, flowMaxElements: 2, useMsgAsId: true, disposeTimeout: TimeSpan.FromSeconds(10)))
+            {
+                fake.Scheduler.Run(async () =>
+                {
+                    var handledMsgs = new List<string>();
+                    var doneTask = fake.Subscriber.StartAsync(async (msg, ct) =>
+                    {
+                        var data = msg.Data.ToStringUtf8();
+                        handledMsgs.Locked(() => handledMsgs.Add(data));
+                        if (data == "msg0" || data == "msg2")
+                        {
+                            // Delay handling so that StopAsync is called while the rest
+                            // are still pulled but waiting for a flow control slot.
+                            await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(5), ct));
+                        }
+                        return SubscriberClient.Reply.Ack;
+                    });
+
+                    // Wait for "msg0" and "msg2" to start being handled. "msg1" will also be handled, but not hold onto
+                    // flow control, releasing it so "msg2" can begin being handled.
+                    await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(1), CancellationToken.None));
+                    Assert.Equivalent(new [] {"msg0", "msg1", "msg2"}, handledMsgs);
+
+                    // Stop the subscriber. This should ensure pulled messages that haven't entered the user handler yet
+                    // will be NAck'ed. Specifically for messages already in flow control, they will have to wait for pending
+                    // messages to be processed before they are NAck'ed.
+                    await fake.TaskHelper.ConfigureAwait(fake.Subscriber.StopAsync(CancellationToken.None));
+
+                    // Verify that "msg0", "msg1" and "msg2" completed handling normally, while the rest were
+                    // automatically Nacked during shutdown.
+                    Assert.Equivalent(new [] {"msg0", "msg1", "msg2"}, fake.Subscribers[0].Acks.Select(x => x.Id));
+                    Assert.Equivalent(new [] {"msg3", "msg4", "msg5"}, fake.Subscribers[0].Nacks.Select(x => x.Id));
+                });
+            }
+        }
+
         /// <summary>
         /// Creates a sequence of failing streaming pull responses, optionally followed by one "never responds" streaming pull.
         /// The nested aspect is because each streaming pull can (theoretically) consist of multiple responses.
--- a/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
+++ b/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
@@ -496,12 +496,20 @@ public sealed partial class SubscriberClientImpl
             }
         }
 
-        private async Task ProcessPullMessagesAsync(List<ReceivedMessage> msgs, HashSet<string> msgIds)
+        private async Task ProcessPullMessagesAsync(List<ReceivedMessage> msgs, HashSet<string> leaseTracking)
         {
             // Running async. Common data needs locking
             for (int msgIndex = 0; msgIndex < msgs.Count; msgIndex++)
             {
-                _softStopCts.Token.ThrowIfCancellationRequested();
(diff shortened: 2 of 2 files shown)


apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
```diff
@@ -510,19 +518,25 @@ await _taskHelper.ConfigureAwait(_flow.Process(msg.CalculateSize(), _messageOrde
...
                     if (msg.DeliveryAttempt > 0)
                     {
                         msg.Message.Attributes[DeliveryAttemptAttrKey] = msg.DeliveryAttempt.ToString(CultureInfo.InvariantCulture);
                     }
                     // Call user message handler
-                    var reply = await _taskHelper.ConfigureAwaitHideErrors(() => _handler.HandleMessage(msg.Message, _hardStopCts.Token), Reply.Nack);
+                    var reply = await _taskHelper.ConfigureAwaitHideErrors(() =>
+                    {
+                        // If the subscriber shut down while waiting for flow control, skip the handler.
+                        // Throwing here triggers a Nack, releasing the message for redelivery.
```
amanda-tarafa, line 531: Why? And why is it safe to do it outside the lock?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -2017,6 +2017,50 @@ public void StreamingPullRetry_UnavailableAfterLongDelayTriggersRetryWithoutBack
             });
         }
 
+        [Fact]
+        public void NackMessagesOnShutdown()
+        {
+            var msgs = new[] { new[] {
+                ServerAction.Data(TimeSpan.Zero, ["msg1", "msg2", "msg3", "msg4", "msg5"]),
```
amanda-tarafa: What about messages that were pending from a previous pull?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
robertvoinescu-work: 1. `ConfigureAwaitHideErrors` accepts a `Func` callback and a fallback value (in this case, `Reply.Nack`) to be returned should an exception occur during execution. So on cancellation we will end up NAck'ing
2. ThrowIfCancellationRequested` i[s thread-safe](https://learn.microsoft.com/en-us/dotnet/api/system.threading.cancellationtoken?view=net-10.0) and can safely be moved out of the lock.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
robertvoinescu-work: I just added "msg0" in this test case into an earlier pull. This message will remain unhandled when shutdown is initiated at which point it will be allowed for finish processing. Then "msg3", "msg4" and "msg5" which didn't make it into flow control before shutdown will be Nack'ed. Does that clarify the concer, or did you mean something else by "pending".

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
amanda-tarafa: OK, but then we need to wait for flow control to allow us to get here, correct? We are not really nacking inmediately? I'm not saying we should do it differently or that we coud, I just want to make sure I understand.

Is there another way? Note I'm no so interested in this case as maybe how this plays out with the upcoming cases?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -2017,6 +2017,52 @@ public void StreamingPullRetry_UnavailableAfterLongDelayTriggersRetryWithoutBack
...
+                    // Wait for "msg0" and "msg2" to start being handled, "msg1" will also be handled, but not hold onto
+                    // flow control.
+                    await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(1), CancellationToken.None));
+                    Assert.Equivalent(new [] {"msg0", "msg1", "msg2"}, handledMsgs);
+
+                    // Stop the subscriber. This should immediately trigger Nacks for pulled messages
+                    // that haven't entered the user handler yet.
+                    await fake.TaskHelper.ConfigureAwait(fake.Subscriber.StopAsync(CancellationToken.None));
+
+                    // Verify that "msg1" completed handling normally, while the rest were
```
amanda-tarafa: nit: Update this comment.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -2017,6 +2017,52 @@ public void StreamingPullRetry_UnavailableAfterLongDelayTriggersRetryWithoutBack
             });
         }
 
+        [Fact]
+        public void NackMessagesOnShutdown()
+        {
+            var msgs = new[] { new[] {
+                ServerAction.Data(TimeSpan.Zero, ["msg0"]),
```
amanda-tarafa, line 2024: For some reason I don't see here our earlier comments around here.

My concern is more about a previous pull with several messages and this pull with with several messages, where several messages of the previous pull are not handled. I guess that's now the taken care off by the second and third message sets.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
```diff
@@ -2017,6 +2017,52 @@ public void StreamingPullRetry_UnavailableAfterLongDelayTriggersRetryWithoutBack
...
+                        }
+                        return SubscriberClient.Reply.Ack;
+                    });
+
+                    // Wait for "msg0" and "msg2" to start being handled, "msg1" will also be handled, but not hold onto
+                    // flow control.
+                    await fake.TaskHelper.ConfigureAwait(fake.Scheduler.Delay(TimeSpan.FromSeconds(1), CancellationToken.None));
+                    Assert.Equivalent(new [] {"msg0", "msg1", "msg2"}, handledMsgs);
+
+                    // Stop the subscriber. This should immediately trigger Nacks for pulled messages
```
amanda-tarafa: What is advancing the fake scheduler? Why does this says "immediately", shouldn't we wait for flow control to allow us in?

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.SingleChannel.cs
robertvoinescu-work: As discussed offline, this won't be a concern in the other cases of NackImmediately where all items in LeaseTracking will be NAck'ed or WaitForProcessing where nothing will be NAck'ed. We won't try to be more pro-active with the NAcking as this will be here just for the current shutdown flow while it is being deprecated.

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
robertvoinescu-work: I updated the wording based on our discussion. The wording is ambiguous shutdown does not immediately trigger the NAcks

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
robertvoinescu-work: Done

apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1.Tests/SubscriberClientTest.cs
robertvoinescu-work: As a note on how this works, the fake scheduler is being automatically advanced by `TestScheduler.Run`, which all of the test code is fed into. `TestScheduler.Delay` just creates the points in fake time where tasks resolve. We essentially schedule the following:

T=1s: `StopAsync` is called.
T=5s: `msg0` and `msg2` finish processing, unblocking the rest of the logic.

The `TestScheduler.Run` method iterates in order through these time steps until the test code returns. The `StopAsync` method is called at T=1s, but it won't resolve until those active handlers (msg0 and msg2) finish. Since the the main task is not resolved yet, `Run` will automatically jump to the next interesting point in time (T=5s). At that point, the handlers finish, flow control is unblocked, and the remaining messages are Nack'ed so the test can wrap up.

PR #15402, opened 2026-01-22
chore: bump gRPC versions and centralize them

jskeet:


jskeet:
@amanda-tarafa The first version of this PR had a commit which introduced a file in Bigtable just to provoke the conformance tests - they passed.

PR #15418, opened 2026-01-29
chore: remove generator-input/tweaks/global.json

### LINKED ISSUE #15394: Remove generator-input/tweaks/global.json
I *think* we only needed the file when we built post-generation code (e.g. BigtableClientGenerator) during generation. We now build it as part of the container (when running from Librarian).

It would be good to get rid of it if we can.

jskeet:
This is no longer necessary, but causes some confusion.

Fixes #15394

### DIFF
--- a/generator-input/tweaks/global.json
+++ b/generator-input/tweaks/global.json
@@ -1,7 +1,0 @@
-{
-  "sdk": {
-    "version": "8.0.417",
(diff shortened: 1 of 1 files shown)


jskeet:
(Tested by generating Bigtable with a new image created after this change - there's no diff. That's the only API where post-generation runs dotnet.)

PR #15424, opened 2026-01-30
chore: Unblock release of docs-help

amanda-tarafa:
@jskeet I think this should be right, so that we can publish the changes in #15423 .


jskeet (approved):
Yes, I *think* this is right.

PR #15451, opened 2026-02-12
chore: reset Dataform V1Beta last-generated commit

jskeet:
PR #15447 updated the pipeline state even after the build failed, which meant we then didn't regenerate.


jskeet:
(I'm going to regenerate everything locally to see if there's anything else in the same position.)

jskeet:
No other APIs need regenerating, but we do want a tweak for the VertexAI extensions dependencies.

PR #15457, opened 2026-02-12
chore: replace old dotnet teams with cloud-sdk-dotnet-team

suztomo:
b/479544839


suztomo:
There will be other team renaming in this CODEOWNERS file as a next wave. Not in this pull request.

PR #15472, opened 2026-02-18
chore: update image tag to v2026-02-18

jskeet:


jskeet:
Tested locally by running `update-apis` which worked.

PR #15476, opened 2026-02-23
fix: Fix PredictionServiceChatClient to skip empty TextReasoningContent

stephentoub:
When TextReasoningContent has no text and no protected data, the Part is now skipped entirely instead of setting Part.Text to null, which throws via Protobuf's CheckNotNull.

cc: @amanda-tarafa


amanda-tarafa:
Thanks!

PR #15483, opened 2026-02-24
fix(Log4Net): Skip snippets in restricted environement.

robertvoinescu-work:


amanda-tarafa:
I'll close this one as we are still figuring out what the failure is on CI.

PR #15492, opened 2026-03-02
feat(Storage): Add filter to ListObjectsOptions and integration tests for object context

mahendra-google:
This PR introduces the ability to filter ListObjects results with the introduction of `Filter` property in `ListObjectsOptions` and adds comprehensive integration tests for the `Object Context` feature. Please see [b/434726579](http:/b/434726579)

### DIFF
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/PatchObjectTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/PatchObjectTest.cs
@@ -12,10 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Apis.Storage.v1.Data;
 using Google.Cloud.ClientTesting;
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using Xunit;
+using static Google.Cloud.Storage.V1.IntegrationTests.TestHelpers;
 using Object = Google.Apis.Storage.v1.Data.Object;
 
 namespace Google.Cloud.Storage.V1.IntegrationTests
@@ -49,6 +53,36 @@ namespace Google.Cloud.Storage.V1.IntegrationTests
             var client = _fixture.Client;
             Assert.Throws<ArgumentException>(() => client.PatchObject(new Object { Bucket = _fixture.SingleVersionBucket }));
             Assert.Throws<ArgumentException>(() => client.PatchObject(new Object { Name = IdGenerator.FromGuid() }));
+        }
+
+        [Fact]
+        public void ClearAllObjectContexts()
+        {
+            var client = _fixture.Client;
+            var custom = new Dictionary<string, ObjectCustomContextPayload>
+            {
+                  { "A\u00F1\u03A9\U0001F680", new ObjectCustomContextPayload { Value = "Ab\u00F1\u03A9\U0001F680" } }
+            };
+
+            var destination = new Object
+            {
+                Bucket = _fixture.SingleVersionBucket,
+                Name = IdGenerator.FromGuid(),
+                ContentType = "test/type",
+                ContentDisposition = "attachment",
+                Metadata = new Dictionary<string, string> { { "x", "y" } },
+                Contexts = new Object.ContextsData { Custom = custom }
+            };
+            var source = GenerateData(100);
+            _fixture.Client.UploadObject(destination, source);
+
+            var modifiedCustom = new Dictionary<string, ObjectCustomContextPayload>
+            {
+            };
+
+            Object obj = new Object { Name = destination.Name, Bucket = destination.Bucket, ContentType = "text/plain", Contexts = new Object.ContextsData { Custom = modifiedCustom } };
+            var updated = client.PatchObject(obj);
+            Assert.Null(updated.Contexts);
         }
     }
 }
--- a/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UpdateObjectTest.cs
+++ b/apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UpdateObjectTest.cs
@@ -12,10 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+using Google.Apis.Storage.v1.Data;
 using Google.Cloud.ClientTesting;
 using System.Collections.Generic;
 using System.IO;
 using Xunit;
+using static Google.Cloud.Storage.V1.IntegrationTests.TestHelpers;
 
 namespace Google.Cloud.Storage.V1.IntegrationTests
 {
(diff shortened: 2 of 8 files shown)


amanda-tarafa:
The .NET SDK team will review once the GCS team has approved. `do-not-merge` until both approvals are in, thanks.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/ListObjectsOptions.cs
```diff
@@ -169,6 +175,10 @@ internal void ModifyRequest(ListRequest request)
             {
                 request.MatchGlob = MatchGlob;
             }
+            if (Filter != null)
```
krishnamd-jkp, line 178: Check for empty filter also.

mahendra-google: @krishnamd-jkp  Added check for empty filter also.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/ListObjectsOptions.cs
```diff
@@ -169,6 +175,10 @@ internal void ModifyRequest(ListRequest request)
             {
                 request.MatchGlob = MatchGlob;
             }
+            if (!string.IsNullOrWhiteSpace(Filter))
```
amanda-tarafa: Just check for null here. If a user explicitly set a value of white spaces, we shouldn't ignore that, the user should see whatever result the service returns in those cases, whether is an error or something else. FYI @krishnamd-jkp .

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UploadObjectTest.cs
```diff
@@ -91,6 +92,30 @@ public void UploadWithObject()
...
+                Bucket = _fixture.MultiVersionBucket,
+                Name = IdGenerator.FromGuid(),
+                Contexts = new Object.ContextsData { Custom = custom }
+            };
+            var source = GenerateData(100);
+            var result = _fixture.Client.UploadObject(destination, source);
+            Assert.Equal(destination.Name, result.Name);
+            Assert.Equal(destination.Bucket, result.Bucket);
+            Assert.Equal(destination.Contexts.Custom.Keys.Single(), result.Contexts.Custom.Keys.Single());
+            Assert.Equal(destination.Contexts.Custom.Values.Single().Value, result.Contexts.Custom[custom.Keys.Single()].Value);
```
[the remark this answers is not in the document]
amanda-tarafa: This is good advice, please apply to result only. Instead of comparing with destination, use a string variable for the key and value to add both in destination and result.

(Here and in the other tests where this may apply)

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UpdateObjectTest.cs
```diff
@@ -39,5 +41,58 @@ public void Success()
...
             Assert.Equal("value", updated.Metadata["key"]);
         }
+
+        [Fact]
+        public void UpdateObjectWithObjectContexts()
+        {
+            var client = _fixture.Client;
+            var source = GenerateData(100);
+
+            var custom = new Dictionary<string, ObjectCustomContextPayload>
```
amanda-tarafa, line 52: Please simplify this test by using a single entry on each of the dictionaries, that's enough to test that the update works and will make the test so much easy to read. Equally use string variables for the expected values.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/PatchObjectTest.cs
```diff
@@ -50,5 +54,131 @@ public void MissingProperties()
             Assert.Throws<ArgumentException>(() => client.PatchObject(new Object { Bucket = _fixture.SingleVersionBucket }));
             Assert.Throws<ArgumentException>(() => client.PatchObject(new Object { Name = IdGenerator.FromGuid() }));
         }
+
+        [Fact]
+        public void ClearAllObjectContexts()
```
amanda-tarafa, line 59: Similar comments as in the other tests.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/PatchObjectTest.cs
```diff
@@ -50,5 +54,131 @@ public void MissingProperties()
...
+            {
+            };
+
+            Object obj = new Object { Name = destination.Name, Bucket = destination.Bucket, ContentType = "text/plain", Contexts = new Object.ContextsData { Custom = modifiedCustom } };
+            var updated = client.PatchObject(obj);
+            Assert.Null(updated.Contexts);
+        }
+
+        [Fact]
+        public void RemoveIndividualObjectContext()
```
amanda-tarafa: And really, we don't need all these different tests that are only testing for server side behavior (which should be tested at the server side). A single test confirming that patching the context works is enough for testing client side behavior. Otherwise we are straining the shared CI infra, etc.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/ListObjectsOptions.cs
krishnamd-jkp: Yeah I think it makes sense since default value is null. I think we should return the server error when user explicitly sets empty filter.

mahendra-google: Thanks @amanda-tarafa  , @krishnamd-jkp .  I do agree with you. Then I will revert the last commit.

mahendra-google:
> A few change requests. Most changes I've requested in some tests will apply to all tests, even though I didn't leave comments in all tests. A few notes also:
> 
> * This should have been 2 PRs, one for the filter options, and one for the context tests. It's easier to reason about the changes and review, etc.
> * In general we don't test service behavior in client libraries. Service behavior should be tested by the service. A single test per operation testing that the context is created/updated/patched/copied etc. is more than enough to test client side behavior. Even that seems excessive as we already test that the library successfully creates/updates/patches ... objects, and the context property is no different than other object properties in that regard, from the point of view of the library. Otherwise we risk reaching quotas and increasing flakes on CI.

@amanda-tarafa   Thank you for the review and the clear guidance on the PR structure. 
I completely agree—splitting the filter options and the context tests into two PRs would have made the review   process much smoother and easier to reason about. I’ll ensure future changes are decoupled this way to respect the review flow.
 I will refine the tests to focus strictly on the client-side behavior as suggested.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1/ListObjectsOptions.cs
mahendra-google: Changes reverted to a null-check for `Filter`.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UploadObjectTest.cs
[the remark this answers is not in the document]
mahendra-google: Acknowledged. I've applied the suggested changes using a dedicated string variable for both the key and value.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/UpdateObjectTest.cs
mahendra-google: The test is simplified now to use a single entry of the dictionaries. Also created a dedicated string variables for the key and value.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/PatchObjectTest.cs
mahendra-google: Similar changes are made here as well by keeping one test per operation.

apis/Google.Cloud.Storage.V1/Google.Cloud.Storage.V1.IntegrationTests/PatchObjectTest.cs
mahendra-google: A single test confirming patching i.e `ClearAllObjectContexts` is preserved for testing client side behavior and other tests are  removed to prevent risk of reaching quotas and flakes on CI.

mahendra-google:
@amanda-tarafa  PR is rebased . All Checks are green now.

PR #15501, opened 2026-03-05
Revert "chore: Library release: 20260304T162729Z"

amanda-tarafa:
Reverts googleapis/google-cloud-dotnet#15499


amanda-tarafa:
@jskeet The release I triggered failed because #15499 was merged but didn't go out at all.

PR #15534, opened 2026-03-25
feat(spanner): add ClientContext to commands

olavloite:
Spanner supports a new ClientContext object for sending additional context values for a statement. This change adds support for that by adding a ClientContext property to the various Statement classes, and including that when a request is built.

This implementation is intentionally kept simple for now. In the future, we might consider adding support for setting default values at either the Connection level or globally for all connections.

cc @aseering


apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
```diff
@@ -293,6 +293,12 @@ public override UpdateRowSource UpdatedRowSource
         /// </remarks>
         public DirectedReadOptions DirectedReadOptions { get; set; }
 
+        /// <summary>
+        /// Client context to be attached to the request. This can be used to pass values that can be read 
+        /// by any SECURE_CONTEXT() calls invoked by this request (e.g., by queries against Parameterized Secure Views).
+        /// </summary>
+        public Google.Cloud.Spanner.V1.RequestOptions.Types.ClientContext ClientContext { get; set; }
```
amanda-tarafa: Pass by comment, I'll do a better review later. We've been trying hard not to expose the underlying generated types on Spanner.Data. Is there an expectation that ClientContext changes frequently?

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
```diff
@@ -114,6 +116,7 @@ private RequestOptions BuildRequestOptions() =>
                 {
                     Priority = PriorityConverter.ToProto(Priority),
                     RequestTag = Tag ?? "",
+                    ClientContext = ClientContext?.Clone()
```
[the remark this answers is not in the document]
olavloite: Sorry, but this is just plain and simple WRONG. `SpannerBatchCommand` __unconditionally__ calls `BuildRequestOptions`. So the `ClientContext` is __always__ included when it has been set.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -462,6 +464,7 @@ private RequestOptions BuildRequestOptions() =>
                 {
                     Priority = PriorityConverter.ToProto(Priority),
                     RequestTag = Tag ?? "",
+                    ClientContext = ClientContext?.Clone()
```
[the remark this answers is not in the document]
olavloite: This is also incorrect. `BuildRequestOptions` is invoked unconditionally, so `ClientContext` will always be included when it has been set. You seem to be confused with `SpannerBatchWriteCommand`

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchCommand.ExecutableCommand.cs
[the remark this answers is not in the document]
olavloite: done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchWriteCommand.ExecutableCommand.cs
```diff
@@ -80,14 +82,15 @@ private BatchWriteRequest CreateRequest()
...
+            if (string.IsNullOrEmpty(Tag) && Priority == Priority.Unspecified && ClientContext == null)
             {
                 return null;
             }
             return new RequestOptions
             {
                 RequestTag = Tag ?? "",
-                Priority = PriorityConverter.ToProto(Priority)
+                Priority = PriorityConverter.ToProto(Priority),
+                ClientContext = ClientContext?.Clone()
```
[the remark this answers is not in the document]
olavloite: done

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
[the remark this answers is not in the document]
olavloite: done

olavloite:
I'm not sure I'm reading it correctly, but I think that the build error is unrelated to this change:

```
Finding changes in Google.Cloud.NetApp.V1...
Comparing old and new versions (by source)
Major changes:
Enum 'StoragePoolType' entry 'UnifiedLargeCapacity' removed.
```

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
olavloite: @amanda-tarafa

> Is there an expectation that ClientContext changes frequently?

TLDR: Yes, very much so.

Most relational databases store session state in the server. 'Session state' in this case is for example something like 'default_timezone'. This session state is unique per database connection, and a client can change this by changing a session variable. In Spanner, we do not store this type of session state in the server, as it would create a lot of complexities with keeping it up to date on all servers etc. We have therefore traditionally included this type of 'session state' in the requests that are being sent to Spanner. These 'session variables' have been spread out in many different protos (e.g. `QueryOptions`, `RequestOptions`, `TransactionOptions` etc.). We want to centralize this into one proto, and that is `ClientContext`. So (almost) all new options will in the future be added to this proto.

Or as the documentation of the proto reads:

```
/// Container for various pieces of client-owned context attached to a request.
```

amanda-tarafa:
> I'm not sure I'm reading it correctly, but I think that the build error is unrelated to this change:
> 
> ```
> Finding changes in Google.Cloud.NetApp.V1...
> Comparing old and new versions (by source)
> Major changes:
> Enum 'StoragePoolType' entry 'UnifiedLargeCapacity' removed.
> ```

You need to rebase on HEAD.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.cs
amanda-tarafa: Ok, but before we would have changed whatever the command or handwritten option we had as an equivalent for the generated type where the new option was being added. And now, we'll only have to change this single place (if we handwrite its equivalent), which is an improvement in any case. It doesn't seem like the existence of `ClientContext` should bring on a deluge of new properties, just a new place to put whatever was coming before.

I'd still like to have control of how we expose this, whether some of the options may support immutability or not, etc. For instance, a few of the transaction options are needed on transaction creation, after which they must not be changed, but others may be changed until the first command has executed but not after, or on commit, etc. If we expose the proto then it's harder for us to control these things, and easier for users to shoot themselves on the foot. We've had to jump through some hoops before (see our handwritten `TransactionCreationOptions` vs. `TransactionOptions` as an example), especially since we are constrained by the ADO.NET surface in how we can expose things. Happy to chat more about this.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerCommand.ExecutableCommand.cs
```diff
@@ -75,6 +76,7 @@ public ExecutableCommand(SpannerCommand command)
                 Priority = command.Priority;
                 Tag = command.Tag;
                 DirectedReadOptions = command.DirectedReadOptions;
+                ClientContext = command.ClientContext?.Clone();
```
amanda-tarafa, line 79: Hmmm, maybe I should clone a few of these as well. I'll come back here and see whether we need it.

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchWriteCommand.ExecutableCommand.cs
```diff
@@ -80,14 +82,15 @@ private BatchWriteRequest CreateRequest()
 
         private RequestOptions BuildRequestOptions()
         {
-            if (string.IsNullOrEmpty(Tag) && Priority == Priority.Unspecified)
+            if (string.IsNullOrEmpty(Tag) && Priority == Priority.Unspecified && ClientContext == null)
```
amanda-tarafa, line 85: Interesting difference. @olavloite do you know if Batch Write would not accept a default `RequestOptions` in case none of these were set, the same as the rest of the commands seem to accept?

Otherwise I'll come back here and simplify these to match the others.

amanda-tarafa (approved):

apis/Google.Cloud.Spanner.Data/Google.Cloud.Spanner.Data/SpannerBatchWriteCommand.ExecutableCommand.cs
olavloite: @amanda-tarafa  (Sorry, I completely lost track of this PR)

`BatchWrite` should also just accept a default `RequestOptions` instance, so this does indeed seem to be redundant.

